PR #684, opened 2022-02-25
IGNITE-16453 Try to optimize IgniteLongList

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-16453


modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/reuse/LongListReuseBag.java
```diff
@@ -17,13 +17,13 @@
...
+import it.unimi.dsi.fastutil.longs.LongArrayList;
 import java.io.Externalizable;
-import org.apache.ignite.internal.util.IgniteLongList;
 
 /**
- * {@link IgniteLongList}-based reuse bag.
+ * {@link LongArrayList}-based reuse bag.
  */
-public final class LongListReuseBag extends IgniteLongList implements ReuseBag {
+public final class LongListReuseBag extends LongArrayList implements ReuseBag {
```
sashapolo, line 26: Why does this class have to extend `LongArrayList`? Doesn't it make sense to use composition here?

sashapolo: Also, would it make more sense to use `LongArrayFIFOQueue` here?

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
```diff
@@ -4550,10 +4551,10 @@ public long pollFreePage() {
                 return 0L;
             }
 
-            if (freePages.getClass() == IgniteLongList.class) {
-                IgniteLongList list = ((IgniteLongList) freePages);
+            if (freePages.getClass() == LongArrayList.class) {
```
sashapolo, line 4554: why not `instanceOf`? Also, maybe a queue is more suitable here?

tkalkirill: I think reference equality is used here for better performance, I need to think about the queue.

ibessonov: I doubt that the performance difference is even measurable in this case

tkalkirill: I fixed it, hope it's better.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/reuse/LongListReuseBag.java
tkalkirill: Fixed it.

tkalkirill:
@ibessonov @sashapolo Guys please look again.

sashapolo (approved):

ibessonov (approved):

PR #685, opened 2022-02-25
IGNITE-16534 Make BplusMetaIo abstract 

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-16534

### DIFF
--- a/modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
+++ b/modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
@@ -168,6 +168,9 @@ public abstract class BplusTree<L, T extends L> extends DataStructure implements
     // TODO: IGNITE-16350 Make it final.
     private IoVersions<? extends BplusLeafIo<L>> leafIos;
 
+    // TODO: IGNITE-16350 Make it final.
+    private IoVersions<? extends BplusMetaIo> metaIos;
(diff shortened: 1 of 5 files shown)


sashapolo (approved):

tkalkirill:
@ibessonov Could you review it?

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
```diff
@@ -893,11 +897,12 @@ protected BplusTree(
             long metaPageId,
             @Nullable ReuseList reuseList,
             IoVersions<? extends BplusInnerIo<L>> innerIos,
-            IoVersions<? extends BplusLeafIo<L>> leafIos
+            IoVersions<? extends BplusLeafIo<L>> leafIos,
```
ibessonov, line 900: I feel like passing a registry and types should be more convenient, what do you think?

Types can even be retrieved from protected methods to reduce a number of parameters in constructor. But if you think that it's a bad solution, just let me know. Invoking virtual methods in constructors is an anti-pattern, I know it

ibessonov: Ok, that's fair. I guess we can revisit it in the future, it's not like the code can't be changed

ibessonov (approved):

PR #690, opened 2022-02-25
IGNITE-16498 Add ability to add handlers to RestModule from other Ignite modules

rpuch:
https://issues.apache.org/jira/browse/IGNITE-16498


modules/rest-api/src/main/java/org/apache/ignite/rest/ErrorResult.java
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package org.apache.ignite.internal.rest;
+package org.apache.ignite.rest;
```
sashapolo, line 18: Why did you remove the "internal" package?

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestModule.java
```diff
@@ -17,136 +17,73 @@
...
-    /** Node configuration route. */
-    private static final String NODE_CFG_URL = "/management/v1/configuration/node/";
-
-    /** Cluster configuration route. */
-    private static final String CLUSTER_CFG_URL = "/management/v1/configuration/cluster/";
-
-    /** Path parameter. */
-    private static final String PATH_PARAM = "selector";
-
+public class RestModule implements RestHandlersRegister, IgniteComponent {
```
sashapolo, line 47: I would suggest to rename this class to `RestComponent` to be consistent with other components

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestModule.java
```diff
@@ -17,136 +17,73 @@
...
-    /** Presentation of cluster configuration. */
-    private final ConfigurationPresentation<String> clusterCfgPresentation;
-
     /** Netty bootstrap factory. */
     private final NettyBootstrapFactory bootstrapFactory;
 
+    private final SimpleRouter router = new SimpleRouter();
+
     /** Netty channel. */
     private volatile Channel channel;
```
sashapolo, line 60: I would suggest to re-write this logic using the busy lock approach

modules/rest/src/main/java/org/apache/ignite/internal/rest/routes/Router.java
```diff
@@ -17,113 +17,19 @@
...
-    public void addRoute(Route route) {
-        routes.add(route);
-    }
-
+public interface Router {
     /**
      * Finds the route by request.
      *
      * @param req Request.
      * @return Route if founded.
```
sashapolo, line 32:
```suggestion
     * @return Route if found.
```

modules/rest/src/main/java/org/apache/ignite/internal/rest/routes/Router.java
```diff
@@ -17,113 +17,19 @@
...
      */
-    public Optional<Route> route(HttpRequest req) {
-        return routes.stream().filter(r -> r.match(req)).findFirst();
-    }
+    Optional<Route> route(HttpRequest req);
```
sashapolo, line 34: we don't use Optionals anywhere, I would suggest to use `@Nullable` instead

modules/rest-api/src/main/java/org/apache/ignite/rest/Routes.java
```diff
@@ -0,0 +1,84 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.rest;
+
+import io.netty.handler.codec.http.HttpMethod;
+import io.netty.util.AsciiString;
+
+/**
+ * Allows to configure REST handlers vs routes.
```
sashapolo, line 24: vs? Did you mean `via`?

modules/rest-api/src/main/java/org/apache/ignite/rest/Routes.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+public interface Routes {
+    /**
+     * Adds the route to router chain.
```
sashapolo, line 28:
```suggestion
     * Adds the route to the router chain.
```

modules/rest-api/src/main/java/org/apache/ignite/rest/Routes.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * GET query helper.
+     *
+     * @param route      Route.
+     * @param acceptType Accept type.
+     * @param hnd        Actual handler of the request.
+     * @return Router
+     */
+    default Routes get(String route, AsciiString acceptType, RequestHandler hnd) {
```
sashapolo, line 42: `AsciiString` is netty-specific, do we want to expose that?

modules/rest-api/src/main/java/org/apache/ignite/rest/Routes.java
```diff
@@ -0,0 +1,84 @@
...
+     * @param hnd        Actual handler of the request.
+     * @return Router
+     */
+    default Routes put(String route, AsciiString acceptType, RequestHandler hnd) {
+        addRoute(new Route(route, HttpMethod.PUT, acceptType.toString(), hnd));
+        return this;
+    }
+
+    /**
+     * Defines a PATCH route.
```
sashapolo, line 73: This javadoc is inconsistent with other javadocs =)

modules/rest/src/main/java/org/apache/ignite/internal/rest/routes/SimpleRouter.java
```diff
@@ -0,0 +1,80 @@
...
+    /** Routes. */
+    private final List<Route> routes;
+
+    /**
+     * Creates a new router with the given list of {@code routes}.
+     *
+     * @param routes Routes.
+     */
+    public SimpleRouter(List<Route> routes) {
+        this.routes = List.copyOf(routes);
```
sashapolo, line 50: 1. This constructor is not used anywhere.
2. `List.copyOf` creates an immutable list, `addRoute` which breaks the contract of this class (e.g. `addRoute` will fail)

modules/rest-api/src/main/java/org/apache/ignite/rest/Routes.java
rpuch: Rephrased it

modules/rest-api/src/main/java/org/apache/ignite/rest/Routes.java
rpuch: Removed `AsciiString` usages

modules/rest-api/src/main/java/org/apache/ignite/rest/Routes.java
rpuch: Made the other javadocs consistent with this one

modules/rest/src/main/java/org/apache/ignite/internal/rest/routes/SimpleRouter.java
rpuch: Removed the constructor, it's easy to add it back if we need it in the future

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestModule.java
rpuch: This should probably be implemented in a separate issue.
I filed a ticket about this: https://issues.apache.org/jira/browse/IGNITE-16636

sashapolo (approved):

PR #692, opened 2022-02-28
IGNITE-16558 Sql schema synchronization

korlov42:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/SqlSchemaManagerImpl.java
```diff
@@ -72,36 +79,73 @@ public SchemaPlus schema(@Nullable String schema) {
     /** {@inheritDoc} */
     @Override
     @NotNull
```
tledkov-gridgain, line 74: - Two spaces between 'requiredVer={}' and 'latestKnownVer'
- Should we use different exceptions for cases when table not found (normal flow, e.g. table dropped) and invalid schema version (breaks internal invariant)

tledkov-gridgain, line 74: - Two spaces between 'requiredVer={}' and 'latestKnownVer'
- Should we use different exceptions for cases when table not found (normal flow, e.g. table dropped) and invalid schema version (breaks internal invariant)

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/SqlSchemaManagerImpl.java
```diff
@@ -72,33 +72,43 @@ public SchemaPlus schema(@Nullable String schema) {
...
         }
 
         if (table == null) {
             throw new IgniteInternalException(
                     IgniteStringFormatter.format("Table not found [tableId={}]", id));
         }
 
+        if (table.version() < ver) {
+            throw new IgniteInternalException(
+                    IgniteStringFormatter.format("Table version not found [tableId={}, requiredVer={},  latestKnownVer={}]",
```
tledkov-gridgain, line 92: - Two spaces between 'requiredVer={}' and 'latestKnownVer'
- Should we use different exceptions for cases when table not found (normal flow, e.g. table dropped) and invalid schema version (breaks internal invariant)

korlov42: > Should we use different exceptions

I agree that "Table version not found" is more like `IllegalStateException`, but currently they both will be handled in the same way. With that said, do we really need a separate exception right now?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/IgniteTableModify.java
```diff
@@ -117,6 +117,8 @@ public IgniteRel clone(RelOptCluster cluster, List<IgniteRel> inputs) {
     public RelWriter explainTerms(RelWriter pw) {
         return super.explainTerms(pw)
                 .itemIf("tableId", getTable().unwrap(InternalIgniteTable.class).id().toString(),
+                        pw.getDetailLevel() == ALL_ATTRIBUTES)
```
tledkov-gridgain, line 120: Why `ProjectableFilterableTableScan#explainTerms` uses condition `pw.getDetailLevel() != EXPPLAN_ATTRIBUTES` but here `pw.getDetailLevel() == ALL_ATTRIBUTES`?

tledkov-gridgain (approved):
OK with me.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/IgniteTableModify.java
korlov42: It's kinda mess here. In fact, we are using `ALL_ATTRIBUTES` in plan serialisation as well as in plan explanation. So, I'he changed `ProjectableFilterableTableScan#explainTerms` to be consistent with TableModify rel.

PR #700, opened 2022-03-03
IGNITE-16543

denis-chudov:


modules/core/src/main/java/org/apache/ignite/internal/causality/VersionedValue.java
```diff
@@ -300,6 +323,14 @@ private void onStorageRevisionUpdate(long causalityToken) {
         assert causalityToken > actualToken0 : IgniteStringFormatter.format(
                 "New token should be greater than current [current={}, new={}]", actualToken0, causalityToken);
 
+        synchronized (updateMutex) {
```
vldpyatkov, line 326: We assume the revision update method cannot invoke before any setting value.
I think, we may sure the assumption will work to updating similarly and no reason to prevent concurrency here.

denis-chudov: removed

PR #705, opened 2022-03-05
IGNITE-16393 Move network's user object serialization to the user thread

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-16393


modules/network/src/main/java/org/apache/ignite/network/NettyBootstrapFactory.java
```diff
@@ -139,6 +139,27 @@ public void start() {
...
 
+    /**
+     * Returns {@code true} if the current thread is a network thread, {@code false} otherwise.
+     *
+     * @return {@code true} if the current thread is a network thread, {@code false} otherwise.
+     */
+    public static boolean isInNetworkThread() {
+        String name = Thread.currentThread().getName();
+
+        if (name.contains("-srv-worker")) {
```
rpuch, line 150: It seems a bit scary to detect threads by name. It is possible to attach some thread-local flag to these threads and check using this flag?

rpuch:
Looks good to me, thanks!

PR #706, opened 2022-03-05
IGNITE-16654 add an ability to compare entry values in meta storage conditional updates

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-16654

### DIFF
--- a/modules/metastorage-client/src/integrationTest/java/org/apache/ignite/internal/metastorage/client/ItMetaStorageServiceTest.java
+++ b/modules/metastorage-client/src/integrationTest/java/org/apache/ignite/internal/metastorage/client/ItMetaStorageServiceTest.java
@@ -18,6 +18,7 @@
 package org.apache.ignite.internal.metastorage.client;
 
 import static java.util.stream.Collectors.toList;
+import static org.apache.ignite.internal.metastorage.client.CompoundCondition.and;
 import static org.apache.ignite.internal.metastorage.client.CompoundCondition.or;
 import static org.apache.ignite.internal.metastorage.client.Conditions.revision;
 import static org.apache.ignite.internal.metastorage.client.Conditions.value;
@@ -48,6 +49,7 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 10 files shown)


modules/metastorage-server/src/test/java/org/apache/ignite/internal/metastorage/server/ValueConditionTest.java
```diff
@@ -57,4 +61,54 @@ public void ne() {
...
+
+    /**
+     * Tests that value is greater than another one.
+     */
+    @Test
+    public void gt() {
+        Condition cond = new ValueCondition(GREATER, KEY, VAL_1);
+
+        // byte[]{22} > byte[]{11}.
+        assertTrue(cond.test(new Entry(KEY, VAL_2, 1, 1)));
```
kgusakov, line 73: I'd test some cases, when method returns false. At the moment, implementation `return true` will pass the test :)

modules/metastorage-client/src/integrationTest/java/org/apache/ignite/internal/metastorage/client/ItMetaStorageServiceTest.java
```diff
@@ -820,7 +823,80 @@ public void testMultiInvoke() throws Exception {
                 new StatementResult(false))));
 
         assertThat(resultIf.orElse().update(), upd(new Update(
-                Arrays.asList(new org.apache.ignite.internal.metastorage.server.Operation(OperationType.PUT, key2.bytes(), rval2)),
+                List.of(new org.apache.ignite.internal.metastorage.server.Operation(OperationType.PUT, key2.bytes(), rval2)),
+                new StatementResult(false))));
+    }
+
+    @Test
+    public void testMultiInvokeValueConditions() throws Exception {
```
kgusakov, line 831: Due to the integration nature of this suite - I think we should avoid creation of many test methods with different combination of Conditions here. Conditions should be tested in according unit tests suites.

But, I see, that you want to check integration of all parts for new value conditions - I think the good idea, just to extend current `testMultiInvoke` with the needed cases.

kgusakov (changes requested):

modules/metastorage-server/src/test/java/org/apache/ignite/internal/metastorage/server/ValueConditionTest.java
alievmirza: done

modules/metastorage-client/src/integrationTest/java/org/apache/ignite/internal/metastorage/client/ItMetaStorageServiceTest.java
alievmirza: done

kgusakov (approved):

kgusakov (approved):

PR #709, opened 2022-03-09
Ignite 16433

vldpyatkov:


modules/core/src/main/java/org/apache/ignite/internal/causality/VersionedValue.java
```diff
@@ -228,19 +276,21 @@ public void fail(long causalityToken, Throwable throwable) {
...
 
-        assert actualToken0 + 1 == causalityToken : IgniteStringFormatter.format("Token must be greater than actual by exactly 1 "
-                + "[token={}, actual={}]", causalityToken, actualToken0);
+        assert actualToken0 == NOT_INITIALIZED || actualToken0 + 1 == causalityToken : IgniteStringFormatter.format(
+                "Token must be greater than actual by exactly 1 [token={}, actual={}]", causalityToken, actualToken0);
 
         Entry<Long, CompletableFuture<T>> histEntry = history.floorEntry(actualToken0);
 
-        CompletableFuture<T> previousFuture = histEntry.getValue();
+        CompletableFuture<T> previousFuture = histEntry == null ? CompletableFuture.completedFuture(getDefault()) : histEntry.getValue();
```
denis-chudov, line 284: seems that we need `previousFuture` inly when `hasUpdatedValue == false`

PR #715, opened 2022-03-11
IGNITE-16613 [Compute Grid] IgniteCompute facade and ComputeJob interface

rpuch:
https://issues.apache.org/jira/browse/IGNITE-16613


modules/compute-api/src/main/java/org/apache/ignite/compute/ComputeJob.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.apache.ignite.compute;
+
+/**
+ * A Compute job that may be executed on an Ignite node (or a few nodes, or on the whole cluster).
+ *
+ * @param <R> job result type
+ */
+public interface ComputeJob<R> {
```
ibessonov, line 25: Do we have serializability requirement right now? Just asking

modules/compute-api/src/main/java/org/apache/ignite/compute/ComputeJob.java
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Executes the job on an Ignite node.
+     *
+     * @param context  context
+     * @param args     job arguments
```
ibessonov, line 30: This javadoc will have to be expanded in the future, right? Currently, I don't understand supported argument types, for example

modules/compute/src/test/java/org/apache/ignite/internal/compute/IgniteComputeImplTest.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.compute;
+
+import org.apache.ignite.compute.IgniteCompute;
+import org.junit.jupiter.api.Test;
+
+class IgniteComputeImplTest {
```
ibessonov, line 23: Do we really need this test?

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -411,6 +413,12 @@ public void setBaseline(Set<String> baselineNodes) {
         }
     }
 
+    /** {@inheritDoc} */
+    @Override
+    public IgniteCompute compute() {
+        return new IgniteComputeImpl();
```
ibessonov, line 419: I know that it's good enough for now, but why not field?

modules/compute-api/src/main/java/org/apache/ignite/compute/ComputeJob.java
rpuch: It's an interesting question. Here https://issues.apache.org/jira/browse/IGNITE-16613 the `Serializable` is omitted. The current `IgniteCompute` API (namely, the fact that it accepts job class, not job instance) makes me think that we do not need to pass job instances from node to node, instead it seems that each node can instantiate its own instance of a job. In such usage scenario, we don't need a job to be serializable.

rpuch: I think we could leave it as is for now, but ask the guys (Andrey, probably) about the serializability issue.

modules/compute-api/src/main/java/org/apache/ignite/compute/ComputeJob.java
rpuch: Yes, it should be extended in the future. Moreover, currently, there is only the execution method, so no ability to split one task to few jobs. Maybe I'm wrong, but I think this will be changed on next phases, so right now it doesn't make a lot of sense to write a lot in the Javadoc.

As for the supported types, this will also change. On early phases we will only support simple argument types (approximately, primitives, strings, byte arrays, maybe something else), but this restriction will probably be lifted (or weakened) in the future. So, for now we cannot tell a lot in the Javadoc on this matter.

modules/compute/src/test/java/org/apache/ignite/internal/compute/IgniteComputeImplTest.java
rpuch: Removed it

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
rpuch: It's still not clear how it will be created. To make it a little less worrying, I now changed it to lazy instantiation.

PR #717, opened 2022-03-11
IGNITE-16678 RaftGroupService#transferLeadership fixed.

sanpwc:
…ransferLeadership fixed.

### DIFF
--- a/modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItRaftGroupServiceTest.java
+++ b/modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItRaftGroupServiceTest.java
@@ -128,7 +128,7 @@ public class ItRaftGroupServiceTest {
 
     @Test
     @Timeout(10)
-    public void testTransferLeadership() throws Exception{
+    public void testTransferLeadership() throws Exception {
         RaftGroupService raftGroupService = raftGroups.get(clusterServices.get(0).topologyService().localMember()).get();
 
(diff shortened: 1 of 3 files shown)


modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItRaftGroupServiceTest.java
```diff
@@ -0,0 +1,155 @@
...
+            raftGroups.put(clusterServices.get(i).topologyService().localMember(), raftGroupServiceFuture);
+        }
+
+        raftGroups.get(clusterServices.get(0).topologyService().localMember()).get();
+        raftGroups.get(clusterServices.get(1).topologyService().localMember()).get();
+    }
+
+    @AfterAll
+    public static void afterAll() throws Exception {
+        for (Loza raftSrv : raftSrvs) {
```
kgusakov, line 121: We should shutdown clients from `raftGroups`, shouldn't we? I see, that `org.apache.ignite.raft.jraft.rpc.impl.RaftGroupServiceImpl#shutdown` is empty now, but it can be defined better in future.

kgusakov (changes requested):

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItRaftGroupServiceTest.java
sanpwc: Fixed.

PR #721, opened 2022-03-14
IGNITE-16594 Colocation key hash code calculation

tledkov-gridgain:


modules/schema/src/main/java/org/apache/ignite/internal/util/ColocationUtils.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Append value to the hash calculation.
+     *
+     * @param calc Hash calculator.
+     * @param v Value to update hash.
+     * @param typeSpec Value type.
+     */
+    public static void append(HashCalculator calc, Object v, NativeTypeSpec typeSpec) {
+        if (v == null) {
```
ygerzhedovich, line 68: the same case alreay implemented inside HashCalculator

modules/schema/src/main/java/org/apache/ignite/internal/util/ColocationUtils.java
```diff
@@ -0,0 +1,126 @@
...
+            calc.appendLong(0);
```
AMashenkov, line 69: Will it be better to make calculator responsible for null values hash?
```suggestion
            calc.appendNull();
```

PR #723, opened 2022-03-14
IGNITE-16688 Extract RocksDB snapshots into a separate class

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-16688

### DIFF
--- a/modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
+++ b/modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
@@ -190,6 +190,8 @@ public class RocksDbKeyValueStorage implements KeyValueStorage {
 
         List<ColumnFamilyDescriptor> descriptors = cfDescriptors();
 
+        assert descriptors.size() == 2;
+
         var handles = new ArrayList<ColumnFamilyHandle>(descriptors.size());
 
         options = new DBOptions()
@@ -198,9 +200,9 @@ public class RocksDbKeyValueStorage implements KeyValueStorage {
 
         db = RocksDB.open(options, dbPath.toAbsolutePath().toString(), descriptors, handles);
 
-        data = ColumnFamily.createExisting(db, handles.get(0));
+        data = ColumnFamily.wrap(db, handles.get(0));
 
-        index = ColumnFamily.createExisting(db, handles.get(1));
+        index = ColumnFamily.wrap(db, handles.get(1));
 
         snapshotManager = new RocksSnapshotManager(db, List.of(fullRange(data), fullRange(index)), snapshotExecutor);
     }
--- a/modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
+++ b/modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
@@ -27,20 +27,16 @@ import static org.apache.ignite.internal.metastorage.server.persistence.RocksSto
 import static org.apache.ignite.internal.metastorage.server.persistence.RocksStorageUtils.valueToBytes;
 import static org.apache.ignite.internal.metastorage.server.persistence.StorageColumnFamilyType.DATA;
 import static org.apache.ignite.internal.metastorage.server.persistence.StorageColumnFamilyType.INDEX;
-import static org.apache.ignite.internal.rocksdb.RocksUtils.createSstFile;
 import static org.apache.ignite.internal.rocksdb.RocksUtils.find;
 import static org.apache.ignite.internal.rocksdb.RocksUtils.forEach;
+import static org.apache.ignite.internal.rocksdb.snapshot.ColumnFamilyRange.fullRange;
 import static org.apache.ignite.internal.util.ArrayUtils.LONG_EMPTY_ARRAY;
 
-import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
@@ -63,6 +59,7 @@ import org.apache.ignite.internal.metastorage.server.Value;
 import org.apache.ignite.internal.metastorage.server.WatchEvent;
 import org.apache.ignite.internal.rocksdb.ColumnFamily;
 import org.apache.ignite.internal.rocksdb.RocksBiPredicate;
+import org.apache.ignite.internal.rocksdb.snapshot.RocksSnapshotManager;
 import org.apache.ignite.internal.util.Cursor;
 import org.apache.ignite.internal.util.IgniteUtils;
 import org.apache.ignite.lang.IgniteBiTuple;
@@ -74,13 +71,11 @@ import org.rocksdb.ColumnFamilyDescriptor;
 import org.rocksdb.ColumnFamilyHandle;
 import org.rocksdb.ColumnFamilyOptions;
 import org.rocksdb.DBOptions;
-import org.rocksdb.IngestExternalFileOptions;
 import org.rocksdb.Options;
 import org.rocksdb.ReadOptions;
 import org.rocksdb.RocksDB;
 import org.rocksdb.RocksDBException;
 import org.rocksdb.RocksIterator;
-import org.rocksdb.Snapshot;
 import org.rocksdb.WriteBatch;
 import org.rocksdb.WriteOptions;
 
@@ -96,9 +91,6 @@ import org.rocksdb.WriteOptions;
  * entry and the value is a {@code byte[]} that represents a {@code long[]} where every item is a revision of the storage.
  */
 public class RocksDbKeyValueStorage implements KeyValueStorage {
-    /** Suffix for the temporary snapshot folder. */
-    private static final String TMP_SUFFIX = ".tmp";
-
     /** A revision to store with system entries. */
     private static final long SYSTEM_REVISION_MARKER_VALUE = 0;
 
@@ -147,6 +139,9 @@ public class RocksDbKeyValueStorage implements KeyValueStorage {
     /** Index column family. */
     private volatile ColumnFamily index;
 
+    /** Snapshot manager. */
+    private volatile RocksSnapshotManager snapshotManager;
+
     /** Revision. Will be incremented for each single-entry or multi-entry update operation. */
     private volatile long rev;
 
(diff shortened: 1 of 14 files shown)


modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
```diff
@@ -179,25 +179,30 @@ public void start() {
 
         ColumnFamilyOptions indexFamilyOptions = new ColumnFamilyOptions(indexOptions);
 
-        List<ColumnFamilyDescriptor> descriptors = Arrays.asList(
+        return List.of(
                 new ColumnFamilyDescriptor(DATA.nameAsBytes(), dataFamilyOptions),
```
SammyVimes, line 183: I wonder if column family options should be closed too

modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
```diff
@@ -179,25 +179,30 @@ public void start() {
...
-        }
+        var handles = new ArrayList<ColumnFamilyHandle>(descriptors.size());
+
+        options = new DBOptions()
+                .setCreateMissingColumnFamilies(true)
+                .setCreateIfMissing(true);
+
+        db = RocksDB.open(options, dbPath.toAbsolutePath().toString(), descriptors, handles);
+
+        data = ColumnFamily.createExisting(db, handles.get(0));
```
SammyVimes, line 201: `recreateDb` is called inside `start`, but here you use `createExisting`. What if column family doesn't exist?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
```diff
@@ -405,7 +369,7 @@ private ColumnFamilyDescriptor cfDescriptorFromName(String cfName) {
      * Creates a descriptor of the "partition" Column Family.
      */
     private static ColumnFamilyDescriptor partitionCfDescriptor() {
-        return new ColumnFamilyDescriptor(PARTITION_CF_NAME.getBytes(StandardCharsets.UTF_8), new ColumnFamilyOptions());
+        return new ColumnFamilyDescriptor(PARTITION_CF_NAME.getBytes(UTF_8), new ColumnFamilyOptions());
```
SammyVimes, line 372: seems like this method is not used

modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
SammyVimes: Oh, I see, you use create missing column families

SammyVimes (changes requested):

modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
```diff
@@ -218,78 +223,33 @@ private void destroyRocksDb() throws RocksDBException {
     public void close() throws Exception {
         IgniteUtils.shutdownAndAwaitTermination(snapshotExecutor, 10, TimeUnit.SECONDS);
```
rpuch, line 224: This is not what you touch in this PR, just a sidenode. The executor uses an unbounded work queue and a bounded number of threads, and snapshotting operations can probably be pretty long. So, if there is an ongoing snapshot, the executor will be stopped forcibly in the midst of a snapshot. Two (possible?) problems:

1. A snapshot might not be finished?
2. Some tasks submitted to the executor could be enqueued in its work queue and have no chance to be executed, so they will be thrown on the floor due to stop. This will make the corresponding futures hang forever.

Should we file tickets about these problems?

modules/rocksdb-common/src/main/java/org/apache/ignite/internal/rocksdb/snapshot/RocksSnapshotManager.java
```diff
@@ -0,0 +1,212 @@
...
+
+                    // Create futures for capturing SST snapshots of the column families
+                    CompletableFuture<?>[] sstFutures = ranges.stream()
+                            .map(cf -> createSstFileAsync(cf, snapshot, tmpPath))
+                            .toArray(CompletableFuture[]::new);
+
+                    return CompletableFuture.allOf(sstFutures).thenApply(v -> snapshot);
+                }, executor)
+                .whenCompleteAsync((snapshot, e) -> {
+                    // Release the snapshot
```
rpuch, line 92: Do we need this comment at all? The following line talks for itself.

modules/rocksdb-common/src/main/java/org/apache/ignite/internal/rocksdb/snapshot/RocksSnapshotManager.java
```diff
@@ -0,0 +1,212 @@
...
+                    db.releaseSnapshot(snapshot);
+
+                    // Snapshot is not actually closed here, because a Snapshot instance doesn't own a pointer, the
+                    // database does. Calling close to maintain the AutoCloseable semantics
+                    snapshot.close();
```
rpuch, line 97: It seems that if `e` is not null, then `snapshot` is null, so this might throw an NPE.

modules/storage-rocksdb/src/test/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorageTest.java
```diff
@@ -206,4 +208,74 @@ void testRestart(
...
+
+        assertThat(partitionStorage.read(testData1), is(testData1));
+        assertThat(partitionStorage.read(testData2), is(nullValue()));
+    }
+
+    /**
+     * Tests that loading snapshots for one partition does not influence data in another.
+     */
+    @Test
+    void testSnapshotIndependence() {
```
rpuch, line 240: It looks like this test actually has three test scenarios:

1. When we restore one partition from a snapshot, other partitions are not affected
2. When we restore a partition from a snapshot, the data is reverted to the values that were in the snapshot (we don't see a key that was added after snapshot creation)
3. When we read by a row, actually its key gets used for addressing the value, and `value` part is ignored.

Item 3 does not seem to have anything with snapshotting, so my suggestion is to remove it (line 278).

As for items 1 and 2, I suggest to split the test scenario to two and test them independently.

rpuch (changes requested):

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
sashapolo: indeed, I wonder why my IDEA didn't complain about that

modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
sashapolo: I had a thought about that. I think that it is ok if options will be closed by the GC for the following reasons:
1. It is quite cumbersome to pass these options everywhere in order for them to be closed.
2. We create only a couple of these objects during the entire life of an Ignite node, so it's ok if they hang out in the heap for a while.
3. We actually very rarely get rid of these objects.

modules/rocksdb-common/src/main/java/org/apache/ignite/internal/rocksdb/ColumnFamily.java
```diff
@@ -44,53 +43,53 @@
...
+
+    /**
+     * Creates a wrapper around an already created Column Family.
+     *
+     * @param db RocksDB instance.
+     * @param handle Column Family handle.
+     * @return Column Family wrapper.
+     * @throws RocksDBException If an error has occurred during creation.
+     */
+    public static ColumnFamily createExisting(RocksDB db, ColumnFamilyHandle handle) throws RocksDBException {
```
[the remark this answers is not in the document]
sashapolo: cool, I like this approach

modules/rocksdb-common/src/main/java/org/apache/ignite/internal/rocksdb/snapshot/RocksSnapshotManager.java
sashapolo: nice catch!

modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
SammyVimes: Ok, true

modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
sashapolo: Let's imagine the scenario when this method is actually closed: we are stopping an Ignite node. In order to stop the node, we first need to stop the Meta Storage raft group. In which case it's the enclosing component's job to verify that all necessary futures have been resolved before allowing to close itself. What do you think?

SammyVimes (approved):
LGTM

modules/storage-rocksdb/src/test/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorageTest.java
sashapolo: > Item 3 does not seem to have anything with snapshotting, so my suggestion is to remove it (line 278).

I don't agree, it checks that loading a snapshot for a different partition does not overwrite existing keys in another partition.

I'll split the test

modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
rpuch: So, in our case, this would mean that `MetaStorageManager` (the class that now stops `RocksDbKeyValueStorage`) would need to make sure that all the futures are resolved? Currently, its `stop()` method does not seem to bother about it explicitly, but maybe I'm missing something here.

If we push this responsibility to the enclosing component, then this must be a part of the contract, so the contract grows a bit more, and it becomes a little bit less explicit (as the job to deal with such hanging futures is not encapsulated close to the executor). I'm not sure if this is bad or not, to be honest.

Also, we still need this to be done at the enclosing component side, and it does not seem that it's done there yet (again, maybe I just don't see it). Not in this PR, of course.

ibessonov (approved):

rpuch (approved):

modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
sashapolo: > So, in our case, this would mean that MetaStorageManager (the class that now stops RocksDbKeyValueStorage) would need to make sure that all the futures are resolved? 

yes

> Currently, its stop() method does not seem to bother about it explicitly, but maybe I'm missing something here.

because it's not entirely correct =)

I think I will create a ticket to properly implement the component's stop

PR #726, opened 2022-03-15
IGNITE-16689 Get rid of guava library in sql-engine module

zstan:


modules/configuration-annotation-processor/src/integrationTest/java/org/apache/ignite/internal/configuration/processor/AbstractProcessorTest.java
```diff
@@ -133,7 +132,7 @@ public BatchCompilation(List<ClassName> schemaClasses, Compilation compilation)
             generatedSources = compilation.generatedSourceFiles();
 
             generatedClasses = generatedSources.stream()
-                    .collect(Collectors.toMap(object -> fromGeneratedFilePath(object.getName()), Functions.identity()));
+                    .collect(Collectors.toMap(object -> fromGeneratedFilePath(object.getName()), o -> o));
```
yozaneaux, line 135: You could use `Function.identity()` from `import java.util.function.Function;` package.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItDataSchemaSyncTest.java
```diff
@@ -132,7 +131,7 @@ void beforeEach() throws Exception {
      */
     @AfterEach
     void afterEach() throws Exception {
-        IgniteUtils.closeAll(Lists.reverse(clusterNodes));
+        IgniteUtils.closeAll(clusterNodes);
```
yozaneaux, line 134: If you want to keep the reverse, you could use:
```
        Collections.reverse(clusterNodes);
        IgniteUtils.closeAll(clusterNodes);
```

yozaneaux (commented):
Hi @zstan ,

You could use Function.identity() from jdk.

Regards

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItDataSchemaSyncTest.java
zstan: reverse - was a kind of hack from past ) Now it`s not an issue any more.

AMashenkov (approved):

PR #727, opened 2022-03-15
IGNITE-16683 Ignite-3 enable jdbc tests

vldpyatkov:


modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/RaftGroupServiceImpl.java
```diff
@@ -613,7 +613,7 @@ else if (th instanceof SMFullThrowable)
      * @return {@code True} if this is a recoverable exception.
      */
     private boolean recoverable(Throwable t) {
-        return t.getCause() instanceof IOException || t.getCause() instanceof TimeoutException;
+        return t instanceof TimeoutException || (t.getCause() != null && t.getCause() instanceof IOException);
```
sk0x50, line 616: You don't need to check `t.getCause()` for `null` value before using `instanceof`. So, this expression can be simplified:
`return t instanceof TimeoutException || t.getCause() instanceof IOException;`

vldpyatkov: Fixed.

PR #728, opened 2022-03-15
IGNITE-16596 Use colocation key for affinity calculation

tledkov-gridgain:


modules/schema/src/main/java/org/apache/ignite/internal/schema/row/Row.java
```diff
@@ -894,8 +897,21 @@ private byte[] readBytes(ByteBuffer chunk, int off, int len) {
...
+    /** {@inheritDoc} */
     @Override
     public byte[] bytes() {
         return row.bytes();
     }
+
+    /** {@inheritDoc} */
+    @Override
+    public int colocationHash() {
+        HashCalculator hashCalc = new HashCalculator();
```
korlov42, line 909: do we really need to calculate it every time?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItColocationTest.java
```diff
@@ -0,0 +1,301 @@
...
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.EnumSource.Mode;
+import org.junit.jupiter.params.provider.MethodSource;
+
+/**
+ * Tests for the data colocation.
+ */
+@ExtendWith(WorkDirectoryExtension.class)
+public class ItColocationTest extends AbstractBasicIntegrationTest {
```
korlov42, line 61: I believe the sql engine tests is not the better place to test colocation. Let's do integration tests inside module `table`. For e2e tests it's better to create dedicated package.

modules/schema/src/main/java/org/apache/ignite/internal/schema/row/Row.java
tledkov-gridgain: Fixed

AMashenkov (approved):

korlov42 (approved):

PR #737, opened 2022-03-21
IGNITE-16532 Java thin: add heartbeats

ptupitsyn:
Implement heartbeats in Java thin client: https://cwiki.apache.org/confluence/display/IGNITE/IEP-83+Thin+Client+Keepalive

* Add `IgniteClientConfiguration.heartbeatInterval`, default 30 seconds.
* Add `ClientConnectorConfiguration.idleTimeout`, default 0 (disabled).
* Add `idleTimeout` to handshake.
* Set effective heartbeat interval to `Math.min(heartbeatInterval, idleTimeout / 3)`.


modules/client/src/main/java/org/apache/ignite/internal/client/TcpClientChannel.java
```diff
@@ -60,6 +63,9 @@
             ProtocolVersion.V3_0_0
     );
 
+    /** Minimum supported heartbeat interval. */
+    private static final long MIN_RECOMMENDED_HEARTBEAT_INTERVAL = 50;
```
isapego, line 67: Don't you think 50 ms is a too short for real network envirnment?

isapego (commented):
Is it possible to configure client/server so that connection will be closed if no user actions are performed on it?

modules/client/src/main/java/org/apache/ignite/internal/client/TcpClientChannel.java
ptupitsyn: @isapego agree, what would you suggest?

isapego: I don't know really. How about 500 or at least 200-300?

ptupitsyn: 500 sounds good, fixed.

isapego (approved):

PR #738, opened 2022-03-21
IGNITE-16719.

sanpwc:


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -542,6 +542,8 @@ public void subscribe(Subscriber<? super BinaryRow> subscriber) {
 
             private AtomicInteger scanCounter = new AtomicInteger(1);
 
+            private static final int INTERNAL_BATCH_SIZE = 10_000;
```
tledkov-gridgain, line 545: Should we use configured parameter to define internal batch size?

tledkov-gridgain, line 545: Should we use configured parameter to define internal batch size?

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItInternalTableScanTest.java
```diff
@@ -527,7 +543,7 @@ public void onSubscribe(Subscription subscription) {
             public void onNext(BinaryRow item) {
                 retrievedItems.add(item);
 
-                if (retrievedItems.size() % reqAmount == 0) {
+                if (reqAmount != Long.MAX_VALUE && retrievedItems.size() % reqAmount == 0) {
```
tledkov-gridgain, line 546: What for additional  condition `reqAmount != Long.MAX_VALUE` is added?
Is the condition `retrievedItems.size() % reqAmount` **true** in this case?

tledkov-gridgain, line 546: What for additional  condition `reqAmount != Long.MAX_VALUE` is added?
Is the condition `retrievedItems.size() % reqAmount` **true** in this case?

tledkov-gridgain (approved):
OK with me

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItInternalTableScanTest.java
sanpwc: Yep, it's possible to remove given check here. Fixed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -607,7 +617,7 @@ private void cancel(boolean closeCursor) {
             /**
              * Requests and processes n requested elements where n is an integer.
              *
-             * @param n Requested amount of items.
+             * @param n amount of items to request and process.
```
korlov42, line 620:
```suggestion
             * @param n Amount of items to request and process.
```

korlov42 (approved):

PR #739, opened 2022-03-21
IGNITE-16697 MV storage methods and reference implementation

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-16697


modules/core/src/main/java/org/apache/ignite/internal/util/Cursor.java
```diff
@@ -25,5 +25,38 @@
...
+            @Override
+            public void close() throws Exception {
+                // No-op.
+            }
 
+            /** {@inheritDoc} */
+            @Override
+            public Iterator<T> iterator() {
+                return this;
+            }
```
AMashenkov, line 47: I guess, `Cursor extends Iterable` to make Cursor usage easier, e.g. iteration via 'for' loop.
Does it make sense to make this implementation as a default in the Cursor interface?

I suggest just move method "as is" from static nested class to outer interface.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexStorage.java
```diff
@@ -58,12 +64,63 @@
...
+
+    /** Exclude upper bound. */
+    byte LESS = 0;
+
+    /** Include upper bound. */
+    byte LESS_OR_EQUAL = 1 << 1;
+
+    byte FORWARD = 0;
+
+    byte BACKWARDS = 1 << 2;
```
AMashenkov, line 81: Does you mean that every index storage MUST support Backward iteration?
If it is not mandatory requirement, then let's add smth like UnsupportedBackwardIndexScanException to method contract into "@throws" section.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexStorage.java
```diff
@@ -58,12 +64,63 @@
...
+     *      {@code null} means unbounded.
+     * @param upperBound Upper bound. Exclusivity is controlled by a {@link #LESS} or {@link #LESS_OR_EQUAL} flag.
+     *      {@code null} means unbounded.
+     * @param flags Control flags. {@link #GREATER} | {@link #LESS} | {@link #FORWARD} by default. Other available values
+     *      are {@link #GREATER_OR_EQUAL}, {@link #LESS_OR_EQUAL} and {@link #BACKWARDS}.
+     * @param timestamp Timestamp value for consistent multiversioned index scan.
+     * @param columnsProjection Bit set with column indexes to return. {@code null} means returning all available columns.
+     * @param partitionFilter Partition filter predicate. {@code null} means returning data from all partitions.
+     * @return Cursor with fetched index rows.
+     * @throws IllegalArgumentException If backwards flag is passed and backwards iteration is not supported by the storage.
```
AMashenkov, line 102: Let's add smth like UnsupportedBackwardIndexScanException exception for this

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexStorage.java
```diff
@@ -58,12 +64,63 @@
...
+            @Nullable BitSet columnsProjection,
+            @Nullable IntPredicate partitionFilter
+    ) {
+        throw new UnsupportedOperationException("scan");
+    }
+
+    /**
+     * The sole purpose of this class is to avoid massive refactoring while changing the original IndexRow.
+     */
+    interface IndexRowEx {
```
AMashenkov, line 118: Is it temporary class or you are going to fix it within this PR?
Would you mind to add a TODO with a ticket number to rework this?

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestMvSortedIndexStorage.java
```diff
@@ -0,0 +1,253 @@
...
+                })
+                .takeWhile(binaryRow -> {
+                    int cmp = upperCmp.applyAsInt(binaryRow);
+
+                    return includeUpper0 && cmp >= 0 || !includeUpper0 && cmp > 0;
+                })
+                .filter(binaryRow -> {
+                    int partition = binaryRow.hash() % partitions;
+
+                    // This code has been copy-pasted.
```
AMashenkov, line 146:
```suggestion
```

AMashenkov (approved):

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/PartitionStorage.java
```diff
@@ -128,6 +143,66 @@
      */
     Cursor<DataRow> scan(Predicate<SearchRow> filter) throws StorageException;
 
+    /**
+     * Scans the partition and returns a cursor of values in at the given timestamp.
+     *
+     * @param keyFilter Key filter. Binary rows passed to the filter may or may not have a value, filter should only check keys.
+     * @param timestamp Timestamp
```
AMashenkov, line 150:
```suggestion
     * @param timestamp Timestamp.
```

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/PartitionStorage.java
```diff
@@ -128,6 +143,66 @@
...
+     */
+    default Cursor<BinaryRow> scan(Predicate<BinaryRow> keyFilter, @Nullable Timestamp timestamp) {
+        throw new UnsupportedOperationException("scan");
+    }
+
+    /**
+     * Exception class that describes the situation where two independant transactions attempting to write values for the same key.
+     */
+    class TxIdMismatchException extends RuntimeException {
+    }
```
AMashenkov, line 161: Let's move to a separate high-level class (or make is static at least).

tledkov-gridgain: Why does not inherit `IgniteException`?

AMashenkov (approved):

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorage.java
```diff
@@ -0,0 +1,260 @@
...
+        this.tableCfg = tableCfg;
+
+        this.descriptor = descriptor;
+
+        this.pk = pk;
+
+        partitions = tableCfg.partitions();
+
+        index = new ConcurrentSkipListSet<>((l, r) -> {
+            int cmp = compareColumns(l, r);
```
sashapolo, line 80: can be replaced with `((Comparator<BinaryRow>) this::compareColumns).thenComparing(BinaryRow::keySlice)`

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/PrefixComparator.java
```diff
@@ -76,7 +74,18 @@ int compare(BinaryRow binaryRow) {
...
+
+        NativeTypeSpec typeSpec = column.type().spec();
+
+        return compareColumns(row, schemaIndex, typeSpec, value);
+    }
+
+    /**
+     * Compares a particular column of a {@code row} with the given value.
+     */
+    public static int compareColumns(Row row, int schemaIndex, NativeTypeSpec typeSpec, Object value) {
```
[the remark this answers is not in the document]
sashapolo: This is also a huge abstraction leak, can we get rid of it somehow?

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorage.java
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public boolean supportsIndexOnlyScan() {
+        return false;
+    }
+
+    private int compareColumns(BinaryRow l, BinaryRow r) {
+        Row leftRow = new Row(descriptor, l);
```
sashapolo, line 132: Can we reuse `BinaryRowComparator` logic here? For example, there's a `binaryRowComparator` method there...

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Multi-versioned partition storage.
+ * POC version, that represents a combination between a replicated TX-aware MV storage and physical MV storage. Their API are very similar,
+ * althouh there are very important differences that will be addressed in the future.
+ */
+public interface MvPartitionStorage {
+    /**
+     * Exception class that describes the situation where two independant transactions attempting to write values for the same key.
+     */
+    class TxIdMismatchException extends IgniteException {
```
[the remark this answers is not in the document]
ibessonov: Sure

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Reads the value from the storage as it was at the given timestamp.
```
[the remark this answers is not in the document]
ibessonov: Ok

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
```diff
@@ -0,0 +1,96 @@
...
+     * @param key Key.
+     * @param timestamp Timestamp to associate with commited value.
+     * @throws StorageException If failed to write data to the storage.
+     */
+    void commitWrite(BinaryRow key, Timestamp timestamp) throws StorageException;
+
+    /**
+     * Removes data associated with old timestamps.
+     *
+     * @param from Start of hashes range to process. Inclusive.
```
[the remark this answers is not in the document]
ibessonov: You know what, I'll drop this method completely. Things change too quickly, this signature makes no sense most likely

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
```diff
@@ -0,0 +1,96 @@
...
+     * @param to End of hashes range to process. Inclusive.
+     * @param timestamp Timestamp to remove all the data with a lesser timestamp.
+     * @return Future for the operation.
+     */
+    CompletableFuture<?> cleanup(int from, int to, Timestamp timestamp);
```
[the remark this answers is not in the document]
ibessonov: That's right, it's neither implemented nor ever used

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexDescriptor.java
```diff
@@ -166,7 +166,7 @@ public SortedIndexDescriptor(String name, TableView tableConfig) {
     /**
      * Creates a {@link SchemaDescriptor} from a list of index key columns.
      */
-    private static SchemaDescriptor createSchemaDescriptor(List<ColumnView> indexKeyColumnViews) {
+    public static SchemaDescriptor createSchemaDescriptor(List<ColumnView> indexKeyColumnViews) {
```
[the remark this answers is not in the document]
ibessonov: I guess it's an artifact, I made it private again

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexMvStorage.java
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Storage for a sorted index.
+ * POC version, that represents a combination between a replicated TX-aware MV storage and physical MV storage. Real future implementation
+ * will be defined later. Things to notice here: TX-aware implementation should have a projection bitset instead of full row reading.
+ * Physical storage API will be enriched with append/remove methods, like in reference implementation.
+ */
+public interface SortedIndexMvStorage {
+    /** Exclude lower bound. */
+    int GREATER = 0;
```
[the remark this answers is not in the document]
ibessonov: These are default values, I didn't use them in tests explicitly, maybe I should fix that

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexMvStorage.java
```diff
@@ -0,0 +1,95 @@
...
+        /**
+         * Returns indexed column value.
+         *
+         * @param idx PK column index.
+         * @return Indexed column value.
+         */
+        Object value(int idx);
+    }
+
+    boolean supportsBackwardsScan();
```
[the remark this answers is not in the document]
ibessonov: Guys asked for them, there's a chance that we won't support backwards scan in B+Tree, for example, but who knows

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexMvStorage.java
```diff
@@ -0,0 +1,95 @@
...
+
+    boolean supportsIndexOnlyScan();
+
+    /**
+     * Returns a range of index values between the lower bound and the upper bound, consistent with the passed timestamp.
```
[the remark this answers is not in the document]
ibessonov: It means the state of the data as it was at the given time

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    @AfterAll
+    static void afterAll() {
+        kvMarshaller = null;
+        schemaDescriptor = null;
+        marshallerFactory = null;
+    }
+
+    protected BinaryRow binaryKey(TestKey key) {
```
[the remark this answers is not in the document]
ibessonov: Correct, I missed that somehow

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestMvPartitionStorage.java
```diff
@@ -0,0 +1,172 @@
...
+        this.indexes = indexes;
+    }
+
+    private static class VersionChain {
+        final BinaryRow row;
+        final Timestamp begin;
+        final UUID txId;
+        final VersionChain next;
+
+        VersionChain(BinaryRow row, Timestamp begin, UUID txId, VersionChain next) {
```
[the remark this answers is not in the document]
ibessonov: Ok

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestMvPartitionStorage.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        index.remove(aborted);
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void commitWrite(BinaryRow key, Timestamp timestamp) {
+        map.compute(key.keySlice(), (keyBuf, versionChain) -> {
+            assert versionChain != null;
```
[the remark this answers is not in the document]
ibessonov: This invariant must be guaranteed by external lock manager in the future

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorage.java
```diff
@@ -0,0 +1,260 @@
...
+import org.apache.ignite.internal.storage.index.SortedIndexMvStorage;
+import org.apache.ignite.internal.tx.Timestamp;
+import org.apache.ignite.internal.util.Cursor;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Test implementation of MV sorted index storage.
+ */
+public class TestSortedIndexMvStorage implements SortedIndexMvStorage {
+    private final TableView tableCfg;
```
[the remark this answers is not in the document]
ibessonov: Indeed

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorage.java
```diff
@@ -0,0 +1,260 @@
...
+
+    private final int partitions;
+
+    private final IndexColumnView[] indexColumns;
+
+    private final int[] columnIndexes;
+
+    private final NativeType[] nativeTypes;
+
+    protected TestSortedIndexMvStorage(
```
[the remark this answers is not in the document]
ibessonov: There are no reasons for this

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorage.java
ibessonov: Thank you

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorage.java
```diff
@@ -0,0 +1,260 @@
...
+
+        assert idxCfg instanceof SortedIndexView;
+
+        SortedIndexView sortedIdxCfg = (SortedIndexView) idxCfg;
+
+        NamedListView<? extends IndexColumnView> columns = sortedIdxCfg.columns();
+
+        int length = columns.size();
+
+        this.indexColumns = new IndexColumnView[length];
```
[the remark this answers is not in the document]
ibessonov: There must have been a reason for this, I think SortedIndexDescriptor uses different indexes for columns. This code addresses the original BinaryRow, not the "compressed" one that only has the required columns

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorage.java
ibessonov: Again, there's a mismatch with columns indexes, as far as I know

ibessonov: It's weird that we have two almost identical comparison methods

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorage.java
```diff
@@ -0,0 +1,260 @@
...
+            lowerBound = upperBound;
+            upperBound = tempBound;
+        }
+
+        ToIntFunction<BinaryRow> lowerCmp = lowerBound == null ? row -> 1 : boundComparator(lowerBound, direction, includeLower ? 0 : -1);
+        ToIntFunction<BinaryRow> upperCmp = upperBound == null ? row -> -1 : boundComparator(upperBound, direction, includeUpper ? 0 : 1);
+
+        Iterator<IndexRowEx> iterator = index.stream()
+                .dropWhile(binaryRow -> {
+                    return lowerCmp.applyAsInt(binaryRow) < 0;
```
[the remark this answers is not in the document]
ibessonov: Ok

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        public TestValue(Integer intVal, String strVal) {
+            this.intVal = intVal;
+            this.strVal = strVal;
+        }
+
+        @Override
+        public int compareTo(TestValue o) {
+            //TODO Compare nuppable values.
```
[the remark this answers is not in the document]
ibessonov: No, nulls are not used in tests at this moment

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
```diff
@@ -138,6 +139,7 @@ public void start() throws StorageException {
         DBOptions dbOptions = new DBOptions()
                 .setCreateIfMissing(true)
                 .setCreateMissingColumnFamilies(true)
+                .setWalRecoveryMode(WALRecoveryMode.SkipAnyCorruptedRecords)
```
sashapolo, line 142: Why is this needed?

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
```diff
@@ -0,0 +1,206 @@
...
+        assertEquals(List.of(value1), convert(pk.scan(row -> true, ts2)));
+        assertEquals(List.of(value1), convert(pk.scan(row -> true, ts3)));
+
+        assertEquals(List.of(value1, value2), convert(pk.scan(row -> true, ts4)));
+        assertEquals(List.of(value1, value2), convert(pk.scan(row -> true, ts5)));
+    }
+
+    private List<TestValue> convert(Cursor<BinaryRow> cursor) throws Exception {
+        try (cursor) {
+            List<TestValue> list = StreamSupport.stream(cursor.spliterator(), false)
```
sashapolo, line 197: can be written shorter, for example:
```
return StreamSupport.stream(cursor.spliterator(), false)
        .map(this::value)
        .sorted(Comparator.nullsFirst(Comparator.naturalOrder()))
```

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
```diff
@@ -0,0 +1,206 @@
...
+    /**
+     * Creates a storage instance for testing.
+     */
+    protected abstract MvPartitionStorage partitionStorage();
+
+    /**
+     * Tests that reads and scan from empty storage return empty results.
+     */
+    @Test
+    public void testEmpty() throws Exception {
```
sashapolo, line 48: I would suggest to split this test into two

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
```diff
@@ -0,0 +1,206 @@
...
+
+        // Read with timestamp returns null.
+        assertNull(pk.read(binaryKey(key), Timestamp.nextVersion()));
+    }
+
+    /**
+     * Tests basic invariants of {@link MvPartitionStorage#abortWrite(BinaryRow)}.
+     */
+    @Test
+    public void testAbortWrite() throws Exception {
```
sashapolo, line 90: `abortWrite` also influences the state of the indexes, that's not tested anywhere

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
```diff
@@ -0,0 +1,206 @@
...
+
+        assertEquals(newValue, value(pk.read(binaryRow, null)));
+
+        assertEquals(value, value(pk.read(binaryRow, tsExact)));
+        assertEquals(value, value(pk.read(binaryRow, tsAfter)));
+        assertEquals(value, value(pk.read(binaryRow, Timestamp.nextVersion())));
+
+        pk.commitWrite(binaryKey(key), Timestamp.nextVersion());
+
+        assertEquals(newValue, value(pk.read(binaryRow, null)));
```
sashapolo, line 148: should we also check that previous value can still be extracted after a commit?

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
```diff
@@ -0,0 +1,206 @@
...
+        pk.addWrite(binaryRow, UUID.randomUUID());
+
+        // Attempt to write from another transaction.
+        assertThrows(TxIdMismatchException.class, () -> pk.addWrite(binaryRow, UUID.randomUUID()));
+
+        // Read without timestamp returns uncommited row.
+        assertEquals(value, value(pk.read(binaryKey(key), null)));
+
+        // Read with timestamp returns null.
+        assertNull(pk.read(binaryKey(key), Timestamp.nextVersion()));
```
sashapolo, line 83: addWrite also influences the index, that is not tested

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractSortedIndexMvStorageTest.java
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * Creates a storage instanc efor testing.
+     */
+    protected abstract SortedIndexMvStorage createIndexStorage(String name, TableView tableCfg);
+
+    @Test
+    public void testEmpty() throws Exception {
+        SortedIndexMvStorage index1 = createIndexStorage(INDEX1, tableCfg.value());
+        SortedIndexMvStorage index2 = createIndexStorage(INDEX2, tableCfg.value());
```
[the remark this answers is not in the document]
ibessonov: They have different orderings, just to be more confident

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractSortedIndexMvStorageTest.java
```diff
@@ -0,0 +1,238 @@
...
+        pk.commitWrite(binaryRow, ts == null ? Timestamp.nextVersion() : ts);
+    }
+
+    protected IndexRowPrefix prefix(String val) {
+        return () -> new Object[]{val};
+    }
+
+    protected List<TestValue> convert(Cursor<IndexRowEx> cursor) throws Exception {
+        try (cursor) {
+            return StreamSupport.stream(cursor.spliterator(), false)
```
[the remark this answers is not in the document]
ibessonov: We should wait for your PR to me merged, seems like a valuable addition

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractSortedIndexMvStorageTest.java
```diff
@@ -0,0 +1,238 @@
...
+
+        assertEquals(List.of(val9010, val8010), convert(index2.scan(
+                null, prefix("20"), FORWARD, null, null
+        )));
+
+        assertEquals(List.of(val8010, val9010), convert(index2.scan(
+                null, prefix("20"), BACKWARDS, null, null
+        )));
+
+        // Lower bound inclusive.
```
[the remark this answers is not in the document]
ibessonov: Probably, let me see

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractSortedIndexMvStorageTest.java
```diff
@@ -0,0 +1,238 @@
...
+    public void testEmpty() throws Exception {
+        SortedIndexMvStorage index1 = createIndexStorage(INDEX1, tableCfg.value());
+        SortedIndexMvStorage index2 = createIndexStorage(INDEX2, tableCfg.value());
+
+        assertEquals(List.of(), convert(index1.scan(null, null, (byte) 0, null, null)));
+        assertEquals(List.of(), convert(index2.scan(null, null, (byte) 0, null, null)));
+    }
+
+    @Test
+    public void testBounds() throws Exception {
```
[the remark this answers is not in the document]
ibessonov: Yes, I'll add more tests soon

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
ibessonov: Oopsie

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
ibessonov: Thank you!

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
ibessonov: Why? It's, like, 6 lines of code

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
ibessonov: It should be tested in tests for indexes

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
ibessonov: There is a test for that

ibessonov: Why not

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
ibessonov: It's tested in another test

modules/core/src/main/java/org/apache/ignite/internal/util/Cursor.java
```diff
@@ -25,5 +25,38 @@
...
+     * @param it Iterator.
+     * @param <T> Type of elements in iterator.
+     * @return Cursor.
+     */
+    static <T> Cursor<T> fromIterator(Iterator<? extends T> it) {
+        return new Cursor<T>() {
+            /** {@inheritDoc} */
+            @Override
+            public void close() throws Exception {
+                // No-op.
```
korlov42, line 46: Perhaps, it would be better to propagate `close()` to the iterator in case the latter implements an interface `AutoCloseable`

ibessonov: Nice, I'll do it

modules/core/src/main/java/org/apache/ignite/internal/util/Cursor.java
```diff
@@ -25,5 +26,40 @@
...
+     * @param it Iterator.
+     * @param <T> Type of elements in iterator.
+     * @return Cursor.
+     */
+    static <T> Cursor<T> fromIterator(Iterator<? extends T> it) {
+        return new Cursor<T>() {
+            /** {@inheritDoc} */
+            @Override
+            public void close() throws Exception {
+                if (it instanceof Closeable) {
```
korlov42, line 47: I believe `AutoCloseable` fits better here. The logic is simple: since `Cursor` extends `AutoCloseable`, I would expect that wrapped iterator which extends `AutoCloseable` as well will be closed by invocation of 'close()' on the cursor

ibessonov: Closeable is a high-level interface, why would we restrict ourselves with just one of its implementations? I don't see a connection

ibessonov: Sorry, I'm just stupid and thought that AutoCloseable extends Closeable, which is wrong

sashapolo (approved):

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestMvPartitionStorageTest.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * MV partition storage test implementation for {@link TestMvPartitionStorage} class.
+ */
+public class TestMvPartitionStorageTest extends AbstractMvPartitionStorageTest {
+    /** Test partition storage instance. */
+    private TestMvPartitionStorage storage;
+
+    @BeforeEach
+    void setUp() {
+        storage = new TestMvPartitionStorage(List.of());
```
rpuch, line 34: The initialization of `storage` can be moved to the field initializer, making the field `final` and also making the `@BeforeEach` unnecessary (so the code as a whole simpler).

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorage.java
```diff
@@ -0,0 +1,249 @@
...
+     */
+    public TestSortedIndexMvStorage(
+            String name,
+            TableView tableCfg,
+            SchemaDescriptor descriptor,
+            Map<Integer, TestMvPartitionStorage> pk
+    ) {
+        this.descriptor = descriptor;
+
+        this.pk = pk;
```
rpuch, line 76: A defensive copy?

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorage.java
```diff
@@ -0,0 +1,249 @@
...
+
+    public void append(BinaryRow row) {
+        index.add(row);
+    }
+
+    public void remove(BinaryRow row) {
+        index.remove(row);
+    }
+
+    public boolean matches(BinaryRow aborted, BinaryRow existing) {
```
rpuch, line 149: Why are the parameters called the way they are? Is it only used in the 'write abort' scenario? It just seems weird that the naming of the parameters is so highly specialized, while the name of the method is pretty generic.

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorageTest.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public class TestSortedIndexMvStorageTest extends AbstractSortedIndexMvStorageTest {
+
+    private List<TestSortedIndexMvStorage> indexes;
+
+    private TestMvPartitionStorage partitionStorage;
+
+    @BeforeEach
+    void setUp() {
+        indexes = new CopyOnWriteArrayList<>();
```
rpuch, line 40: Initialization of these fields can be moved to field initiailzers

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
```diff
@@ -0,0 +1,78 @@
...
+     * Commits a pending update of the ongoing transaction. Invoked during commit. Committed value will be versioned by the given timestamp.
+     *
+     * @param key Key.
+     * @param timestamp Timestamp to associate with committed value.
+     * @throws StorageException If failed to write data to the storage.
+     */
+    void commitWrite(BinaryRow key, Timestamp timestamp) throws StorageException;
+
+    /**
+     * Scans the partition and returns a cursor of values in at the given timestamp.
```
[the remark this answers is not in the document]
ibessonov: Done

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Creates an uncommitted version, assigned to the given transaction id.
+     *
+     * @param row Binary row to update. Key only row means value removal.
+     * @param txId Transaction id.
+     * @throws TxIdMismatchException If there's another pending update associated with different transaction id.
+     * @throws StorageException If failed to write data to the storage.
+     */
+    void addWrite(BinaryRow row, UUID txId) throws TxIdMismatchException, StorageException;
```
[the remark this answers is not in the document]
ibessonov: I don't think that replacing UUID is necessary right now, it would only bloat the code. As far as I know, Alexey is looking forward to implementing newer UUID standards for transaction ids. So there's a high likelihood of having them forever

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
```diff
@@ -0,0 +1,233 @@
...
+        // Scan.
+        assertEquals(List.of(), convert(pk.scan(row -> true, null)));
+        assertEquals(List.of(), convert(pk.scan(row -> true, Timestamp.nextVersion())));
+    }
+
+    /**
+     * Tests basic invariants of {@link MvPartitionStorage#addWrite(BinaryRow, UUID)}.
+     */
+    @Test
+    public void testAddWrite() throws Exception {
```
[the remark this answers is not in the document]
ibessonov: I'm afraid that tests will become 3 times bigger then they already are. Can we leave these tests?

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
```diff
@@ -0,0 +1,233 @@
...
+        assertEquals(List.of(value1, value2), convert(pk.scan(row -> true, ts4)));
+        assertEquals(List.of(value1, value2), convert(pk.scan(row -> true, ts5)));
+    }
+
+    private List<TestValue> convert(Cursor<BinaryRow> cursor) throws Exception {
+        try (cursor) {
+            return StreamSupport.stream(cursor.spliterator(), false)
+                    .map(BaseMvStoragesTest::value)
+                    .sorted(Comparator.nullsFirst(Comparator.naturalOrder()))
+                    .collect(Collectors.toList());
```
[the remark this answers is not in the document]
ibessonov: I'll do it, thank you!

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
```diff
@@ -0,0 +1,179 @@
...
+import org.apache.ignite.internal.schema.row.Row;
+import org.apache.ignite.lang.IgniteException;
+import org.jetbrains.annotations.Nullable;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+
+/**
+ * Base test for MV storages, contains pojo classes, their descriptor and a marshaller instance.
+ */
+public class BaseMvStoragesTest {
```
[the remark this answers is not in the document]
ibessonov: Done

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
```diff
@@ -0,0 +1,179 @@
...
+    /** Default reflection marshaller factory. */
+    protected static MarshallerFactory marshallerFactory;
```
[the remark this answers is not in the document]
ibessonov: Here's my reasoning:
- these fields are the same for every test case, that's why I made them static
- initialization is in BeforeAll for consistency with AfterAll
- fields are not final because I clean them afterwards to avoid any potential memory leaks

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestMvPartitionStorage.java
```diff
@@ -0,0 +1,167 @@
...
+public class TestMvPartitionStorage implements MvPartitionStorage {
+    private final ConcurrentHashMap<ByteBuffer, VersionChain> map = new ConcurrentHashMap<>();
+
+    private final List<TestSortedIndexMvStorage> indexes;
+
+    public TestMvPartitionStorage(List<TestSortedIndexMvStorage> indexes) {
+        this.indexes = indexes;
+    }
+
+    private static class VersionChain {
```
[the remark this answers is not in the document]
ibessonov: I don't know what to say, technically it is the chain, because it has a head and a link to the tail, right?

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestMvPartitionStorage.java
```diff
@@ -0,0 +1,167 @@
...
+            this.next = next;
+        }
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void addWrite(BinaryRow row, UUID txId) throws TxIdMismatchException {
+        map.compute(row.keySlice(), (keyBuf, versionChain) -> {
+            if (versionChain != null && versionChain.begin == null && !txId.equals(versionChain.txId)) {
+                throw new TxIdMismatchException();
```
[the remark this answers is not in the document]
ibessonov: This exception should in impossible in practice, so I wouldn't bother

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestMvPartitionStorage.java
```diff
@@ -0,0 +1,167 @@
...
+            }
+
+            return new VersionChain(row, null, txId, versionChain);
```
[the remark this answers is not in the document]
ibessonov: Done

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestMvPartitionStorageTest.java
ibessonov: Ok, why not

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorage.java
ibessonov: No, we don't need it

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorage.java
ibessonov: Yes, it's only used for abort right now. Should I use longer name for it?

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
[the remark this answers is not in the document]
rpuch: Yes, this was just a suggestion

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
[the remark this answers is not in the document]
rpuch: I think memory leaks in tests are usually not something that we should be concerned about :) Also, if the fields are not static, but instance fields, their values are garbage as soon as the tests are finished in the given test class, so nullifying the fields seems to be unnecessary, hence `@AfterAll` goes away, etc.

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestMvPartitionStorage.java
```diff
@@ -0,0 +1,167 @@
...
+/**
+ * Test implementation of MV partition storage.
+ */
+public class TestMvPartitionStorage implements MvPartitionStorage {
+    private final ConcurrentHashMap<ByteBuffer, VersionChain> map = new ConcurrentHashMap<>();
+
+    private final List<TestSortedIndexMvStorage> indexes;
+
+    public TestMvPartitionStorage(List<TestSortedIndexMvStorage> indexes) {
+        this.indexes = indexes;
```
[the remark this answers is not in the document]
rpuch: Would an explicit method to add an index be better? Otherwise, addition through a live list is implicit, so it's easier to confuse a reader.

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestMvPartitionStorage.java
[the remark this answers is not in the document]
rpuch: Maybe we talk about different definitions of a chain. In my understanding (I mean physical chains, not some possible mathematical abstractions), a chain is something that consists of links. Each link (but the last one) links to next link. According to this description, `VersionChain` in the code describes a link, not the whole chain. The chain can be traversed from the first link; so the first link is in some sense equivalent to the whole chain. But it's not literally the chain.

It's probably a gray zone. For example, in the Java standard library, `LinkedList` and its `Node` are different classes. This separation allows to make sure that we understand clearly what aspect of 'the thing' we are addressing now. On the other hand, if we keep them combined, we can argue that 'the thing' has properties of both chain and a link, but in the real world a chain and a link a different entities, that's what made me nag here.

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorage.java
rpuch: How about generalizing the method (renaming the parameters and make them symmetrical, that is, check both for non-nullability)?

rpuch (approved):

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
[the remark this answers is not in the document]
ibessonov: I see. In Ignite 2.x we had several major leaks in a very similar scenario, so for me it's a good habit to nullify static fields when tests don't need them anymore

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestMvPartitionStorage.java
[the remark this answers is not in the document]
ibessonov: You are right, this is my laziness I guess. I don't really expect anyone to think much about constructors in these classes. Everything else is much more important.

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestMvPartitionStorage.java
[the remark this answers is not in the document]
ibessonov: I see :)
So, my perspective - "chain" in this context is a data structure that's known as the simplest way to implement stacks, nothing more. I drew no parallels with chains in real world

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/basic/TestSortedIndexMvStorage.java
ibessonov: There's no use for this right now. It's POC, it will be heavily modified in the nearest future, that's where we'll make these changes

PR #741, opened 2022-03-22
IGNITE-16721 Temporary solution for native memory usage reducing

zstan:


modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
```diff
@@ -1167,36 +1167,37 @@ private void initPools(final NodeOptions opts) {
...
 
+
         if (opts.getfSMCallerExecutorDisruptor() == null) {
-            opts.setfSMCallerExecutorDisruptor(new StripedDisruptor<FSMCallerImpl.ApplyTask>(
-                NamedThreadFactory.threadPrefix(opts.getServerName(), "JRaft-FSMCaller-Disruptor"),
-                opts.getRaftOptions().getDisruptorBufferSize(),
-                () -> new FSMCallerImpl.ApplyTask(),
-                opts.getStripes()));
+            opts.setfSMCallerExecutorDisruptor(new StripedDisruptor<>(
+                    NamedThreadFactory.threadPrefix(opts.getServerName(), "JRaft-FSMCaller-Disruptor"),
```
SammyVimes, line 1173: Is it a good idea to alter formatting here? It’s jraft code that we won’t be able to automatically merge later

SammyVimes (changes requested):

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
zstan: ok ok understand

SammyVimes (approved):
LGTM!

PR #743, opened 2022-03-22
IGNITE-16705 Resolve Compute futures that had no chance to be executed on node stop

rpuch:
https://issues.apache.org/jira/browse/IGNITE-16705

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/future/InFlightFutures.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/future/InFlightFutures.java
@@ -17,7 +17,6 @@
 
 package org.apache.ignite.internal.future;
 
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
@@ -29,7 +28,7 @@ import java.util.concurrent.ConcurrentHashMap;
  * never be completed).
  */
 public class InFlightFutures implements Iterable<CompletableFuture<?>> {
-    private final Set<CompletableFuture<?>> inFlightFutures = Collections.newSetFromMap(new ConcurrentHashMap<>());
+    private final Set<CompletableFuture<?>> inFlightFutures = ConcurrentHashMap.newKeySet();
 
     /**
      * Registers a future in the in-flight futures collection. When it completes (either normally or exceptionally),
--- a/modules/core/src/main/java/org/apache/ignite/internal/future/InFlightFutures.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/future/InFlightFutures.java
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package org.apache.ignite.internal.compute;
+package org.apache.ignite.internal.future;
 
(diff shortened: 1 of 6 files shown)


SammyVimes (approved):
LGTM

modules/core/src/main/java/org/apache/ignite/internal/future/InFlightFutures.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Maintains a collection of in-flight {@link CompletableFuture}s (i.e. futures that are not yet completed) to later
+ * have a possibility to process them somehow (for example, cancel them all if we know for sure that they will
+ * never be completed).
+ */
+public class InFlightFutures implements Iterable<CompletableFuture<?>> {
+    private final Set<CompletableFuture<?>> inFlightFutures = Collections.newSetFromMap(new ConcurrentHashMap<>());
```
sanpwc, line 32: ConcurrentHashMap.newKeySet() WDYT?

modules/core/src/main/java/org/apache/ignite/internal/future/InFlightFutures.java
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * Registers a future in the in-flight futures collection. When it completes (either normally or exceptionally),
+     * it will be removed from the collection.
+     *
+     * @param future the future to register
+     */
+    public void registerFuture(CompletableFuture<?> future) {
+        future.whenComplete((result, ex) -> inFlightFutures.remove(future));
```
sanpwc, line 41: And what about pipelining? The ordering of independent completion stages isn't guaranteed, so it's possible to remove the future while one of it's stages is still in progress. E.g.
```
CompletableFuture<R> future = startLocalExecution(jobClass, args);
inFlightFutures.registerFuture(future);
...
future.thenAcceptAsync()
```

sanpwc (commented):
All in all, the patch looks fine to me. It's however an optimization, so it's still required to both protect critical sections inside futures with busy locks and handle NodeStoppingException.

modules/core/src/main/java/org/apache/ignite/internal/future/InFlightFutures.java
rpuch: Thanks, it's better this way, switched to `ConcurrentHashMap#newKeySet()`

modules/core/src/main/java/org/apache/ignite/internal/future/InFlightFutures.java
rpuch: This PR is about making sure that a future does not get stuck due to our thread pool being stopped. It looks like the consequtive stages should be responsibility of those who add them.

There is also another problem: consequtive stages might be computationally heavy, and they can potentially block our 'job execution thread pool'. This needs further investigation, I've created an issue: https://issues.apache.org/jira/browse/IGNITE-16747

PR #745, opened 2022-03-23
IGNITE-16729 Fix flaky test

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-16729

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/testframework/WorkDirectoryExtension.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/testframework/WorkDirectoryExtension.java
@@ -176,7 +176,7 @@ public class WorkDirectoryExtension
                 .map(Method::getName)
                 .orElse(STATIC_FOLDER_NAME);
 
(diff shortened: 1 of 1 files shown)


modules/core/src/test/java/org/apache/ignite/internal/testframework/WorkDirectoryExtension.java
```diff
@@ -176,7 +176,9 @@ private static Path createWorkDir(ExtensionContext context) throws IOException {
                 .map(Method::getName)
                 .orElse(STATIC_FOLDER_NAME);
 
-        Path workDir = getTestClassDir(context).resolve(testMethodDir + '_' + System.currentTimeMillis());
+        // not using currentTimeMillis because some tests can have the same name (e.g. repeated tests) and execute in less than a
```
tkalkirill, line 179:
```suggestion
        // Not using currentTimeMillis because some tests can have the same name (e.g. repeated tests) and execute in less than a
```

tkalkirill (approved):

PR #748, opened 2022-03-23
IGNITE-16727 Add an option to turn off metadata synchronization

vldpyatkov:


modules/configuration/src/main/java/org/apache/ignite/internal/configuration/util/ConfigurationUtil.java
```diff
@@ -1127,6 +1128,12 @@ public static boolean containsNameAnnotation(Field schemaField) {
 
         assert property instanceof ConfigurationNode;
 
+        // If the property is set as {@code true} the attempt to get configuration property direct from Metastorage will be skipped, and
+        // the local one local will be returned. Otherwise, when the property is set as {@code false}, the direct property will be returned.
+        if (IgniteSystemProperties.getBoolean("IGNITE_GET_METADATA_LOCALLY_ONLY")) {
```
sk0x50, line 1133: We should not read this system property through `IgniteSystemProperties` because this call ends up with `Hashtable.get()` which is `synchronized`.

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/util/ConfigurationUtil.java
```diff
@@ -1127,6 +1128,12 @@ public static boolean containsNameAnnotation(Field schemaField) {
...
+            return property;
```
sk0x50, line 1134: It would be nice to add a comment (perhaps, Jira link to an issue) that this code should be removed later.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -114,6 +116,14 @@
 
     private static final int INITIAL_SCHEMA_VERSION = 1;
 
+    /**
+     * If the property is set as {@code true} the attempt to get configuration property direct from Metastorage will be skipped, and the
```
sk0x50, line 120: ```If this property is et to {@code true} then an attempt got get the configuration property directly from the meta storage will be skipped, and the local property will be returned.```

vldpyatkov: Done.

PR #749, opened 2022-03-24
IGNITE-16725 Thin client: Fix Netty buffer leaks

ptupitsyn:


modules/client/src/main/java/org/apache/ignite/internal/client/TcpClientChannel.java
```diff
@@ -141,7 +146,12 @@ private void close(Exception cause) {
     /** {@inheritDoc} */
     @Override
     public void onMessage(ByteBuf buf) {
-        processNextMessage(buf);
+        try {
+            processNextMessage(buf);
+        } catch (Throwable t) {
+            buf.release();
+            throw t;
+        }
```
isapego, line 154: `finally`?

isapego (approved):
Looks good to me in general. See my comment though

modules/client/src/main/java/org/apache/ignite/internal/client/TcpClientChannel.java
isapego: Sorry, finally will not work this way here. Ignore

ptupitsyn: When everything goes well, the buffer is released by the `try` block on line 247 after decoding the message. The decoding happens later on `asyncContinuationExecutor`, not on a current thread, so we can't use `finally` here.

But when there is an exception (unexpected message type, unexpected response id, etc) we should release the buffer here to avoid the leak, because full message decoding won't happen.

isapego: Yeah, I undestand it now. Thanks

PR #759, opened 2022-04-04
IGNITE-16011 Start rebalance, when pending assignments updated

kgusakov:


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1317,152 +1379,153 @@ private RuntimeException convertThrowable(Throwable th) {
...
+                int part = extractPartitionNumber(evt.entryEvent().newEntry().key());
+                UUID tblId = extractTableId(evt.entryEvent().newEntry().key());
 
-        for (String nodeName : nodes) {
-            if (!currClusterMemberNames.contains(nodeName)) {
-                throw new IgniteException("Node '" + nodeName + "' not in current network cluster membership. "
-                        + " Adding not alive nodes is not supported yet.");
-            }
-        }
+                TableImpl tbl = tablesByIdVv.latest().get(tblId);
```
vldpyatkov, line 1456: Why the table should always here?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1317,152 +1379,153 @@ private RuntimeException convertThrowable(Throwable th) {
...
-                            }
-                        });
-                    }
-                })
-                .thenCompose((v) -> {
-                    CompletableFuture<?>[] changePeersFutures = new CompletableFuture<?>[changePeersQueue.size()];
+                    raftGrpSvc.refreshLeader().join();
 
-                    int i = 0;
+                    if (new Peer(raftGrpSvc.clusterService().topologyService().localMember().address()).equals(raftGrpSvc.leader())) {
```
vldpyatkov, line 1487: How do we can sure that the leader won't change just after?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1317,152 +1379,153 @@ private RuntimeException convertThrowable(Throwable th) {
...
-                                    change.partitions(),
-                                    change.replicas());
+                try {
+                    var deltaPeers = newPeers.stream()
+                            .filter(p -> !assignments.get(part).contains(p))
+                            .collect(Collectors.toList());
 
-                            if (!recalculatedAssignments.equals(ByteUtils.fromBytes(currAssignments))) {
-                                change.changeAssignments(ByteUtils.toBytes(recalculatedAssignments));
+                    raftGrpSvc = raftMgr.updateRaftGroup(grpId, assignments.get(part), deltaPeers, raftGrpLsnrSupplier,
```
vldpyatkov, line 1482: Here is excess invocation, because we do nothing with peers in RaftGroupSercive, although recreate them (I know it is to start raft server if it is required, but maybe we can avoid of restarts of RaftGroupSercive).

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1317,152 +1379,153 @@ private RuntimeException convertThrowable(Throwable th) {
...
+    }
 
-        return CompletableFuture.allOf(futures);
+    /**
+     * Extract partition number from the pending key of partition.
+     *
+     * @param key Key.
+     * @return Partition number.
+     */
+    private int extractPartitionNumber(ByteArray key) {
```
vldpyatkov, line 1527: What do you think about moving the both methods to a utility class?

modules/api/src/main/java/org/apache/ignite/configuration/schemas/runner/ClusterConfigurationSchema.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.ignite.configuration.annotation.Value;
+
+/**
+ * Configuration schema for cluster endpoint subtree.
+ */
+@ConfigurationRoot(rootName = "cluster", type = ConfigurationType.DISTRIBUTED)
+public class ClusterConfigurationSchema {
+    /** List of unique names of those cluster nodes that will host distributed metastorage instances. */
+    @Value(hasDefault = true)
+    public String[] metastorageNodes = new String[0];
```
sanpwc, line 31: It seems to be outdated. MetastorageNodes are now specified as part of init command 
`void init(String name, Collection<String> metaStorageNodeNames, Collection<String> cmgNodeNames);`

modules/api/src/main/java/org/apache/ignite/configuration/schemas/runner/ClusterConfigurationSchema.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /** List of unique names of cluster nodes, which form a baseline. */
+    @Value(hasDefault = true)
+    public String[] baselineNodes = new String[0];
```
sanpwc, line 35: It also seems to be outdated, currently it's possible to consider all nodes from logic topology as baseline nodes. 

All-in-all, I don't think that we need ClusterConfigurationSchema.

modules/core/src/main/java/org/apache/ignite/internal/causality/VersionedValue.java
```diff
@@ -276,8 +276,8 @@ public void set(long causalityToken, T value) {
             history.put(causalityToken, initFut);
         }
 
-        assert actualToken0 == NOT_INITIALIZED || actualToken0 + 1 == causalityToken : IgniteStringFormatter.format(
-                "Token must be greater than actual by exactly 1 [token={}, actual={}]", causalityToken, actualToken0);
+        assert actualToken0 == NOT_INITIALIZED || actualToken0 <= causalityToken : IgniteStringFormatter.format(
```
sanpwc, line 279: Here and bellow. Seems that according to current logic of assignments key registration on start it's not possible to have `actualToken0 == causalityToken`, only `actualToken0 < causalityToken` is valid.
Also `Token must be greater than or equal to actual ->  Token must be greater than`
WDYT?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
```diff
@@ -858,7 +858,12 @@ private void storeEntries(Collection<IgniteBiTuple<ByteArray, byte[]>> entries,
 
                     entries.forEach(e -> batch.put(e.getKey(), e.getValue()));
 
-                    return vaultMgr.putAll(batch);
+                    // skip already processed revision
```
sanpwc, line 861: Seems that it's also not required within given solution because all watches are registered on start only.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
```diff
@@ -845,7 +845,7 @@ private CompletableFuture<IgniteUuid> updateAggregatedWatch() {
     private void storeEntries(Collection<IgniteBiTuple<ByteArray, byte[]>> entries, long revision) {
         appliedRevision()
                 .thenCompose(appliedRevision -> {
-                    if (revision <= appliedRevision) {
```
sanpwc, line 848: Seems that `<=` is valid within given solution because all watches are registered on start only.

sanpwc (changes requested):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
```diff
@@ -0,0 +1,124 @@
...
+
+    /** {@inheritDoc} */
+    @Override
+    public void onLeaderElected() {
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void onNewPeersConfigurationApplied(List<PeerId> peers) {
+        Map<ByteArray, Entry> keys = metaStorageMgr.getAll(
```
sanpwc, line 86: Please use busy lock here and in other corresponding places.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
```diff
@@ -0,0 +1,124 @@
...
+                Set.of(partAssignmentsPlannedKey(partId), partAssignmentsPendingKey(partId))).join();
+
+        Entry plannedEntry = keys.get(partAssignmentsPlannedKey(partId));
+        Entry pendingEntry = keys.get(partAssignmentsPendingKey(partId));
+
+        tblConfiguration.change(ch -> {
+            List<List<ClusterNode>> assignments =
+                    (List<List<ClusterNode>>) ByteUtils.fromBytes(((ExtendedTableChange) ch).assignments());
+            assignments.set(partNum, ((List<ClusterNode>) ByteUtils.fromBytes(pendingEntry.value())));
```
sanpwc, line 95: According to rebalance doc, instead of `pendingEntry` we should use applied peers.
```
**Trigger**: When leader applied new Configuration with list of resulting peers `<applied peer>`, it calls `onChangePeersCommitted(<applied peers>)`

**Pseudocode**:
```
metastoreInvoke: \\ atomic
    partition.assignments.stable = appliedPeers
```

File: modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java [hunk 12]
```diff
@@ -0,0 +1,124 @@
...
+            ((ExtendedTableChange) ch).changeAssignments(ByteUtils.toBytes(assignments));
+        });
+
+        if (plannedEntry.value() != null) {
```
sanpwc, line 99: Why do we need external plannedEntry null checking in addition to in-invoke
`notExists(partAssignmentsPlannedKey(partId)),` one? I'd rather say that invoke should be considered as an atomic operation and it's actually **is** because handling invoke command is linearized with raft.

All-in-all I suggest considering following operation instead

        /**
         * metastoreInvoke: \\ atomic
         *     partition.assignments.stable = appliedPeers
         *     if empty(partition.assignments.planned):
         *         partition.assignments.pending = empty
         *     else:
         *         partition.assignments.pending = partition.assignments.planned
         *         remove(partition.assignments.planned)
         *
         */
        if (!metaStorageMgr.invoke(If.iif(
                notExists(partAssignmentsPlannedKey(partId)),
                ops(
                        put(partAssignmentsStableKey(partId), appliedAssignments),
                        remove(partAssignmentsPendingKey(partId))
                )
                        .yield(true),
                ops(
                        put(partAssignmentsStableKey(partId), appliedAssignments),
                        put(partAssignmentsPendingKey(partId), get(partAssignmentsPlannedKey(partId))),
                        remove(partAssignmentsPlannedKey(partId))
                )
                        .yield(false))
        ).join().getAsBoolean()) {
            onNewPeersConfigurationApplied(peers);
        }

please pay attention that 

- instead of pre-readed plannedEntry of keys I used get operation that I suggest to add in addition to put, remove, noop ones.
- instead of pending keys resolved appliedAssignments are used.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
```diff
@@ -427,14 +430,31 @@ void reset(final Status st) {
...
             this.version++;
             this.stage = Stage.STAGE_NONE;
             this.nchanges = 0;
+
+            Closure oldDoneClosure = done;
+
             if (this.done != null) {
-                Utils.runClosureInThread(this.node.getOptions().getCommonExecutor(), this.done, st != null ? st :
+                this.done = (Status status) -> {
+                    if (node.getOptions().getRaftGrpEvtsLsnr() != null) {
```
sanpwc, line 448: I'd rather consider default no-op implementation in order not to check for null

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/DistributedConfigurationStorage.java
```diff
@@ -306,7 +306,12 @@ public boolean onUpdate(@NotNull WatchEvent events) {
 
                     long newChangeId = masterKeyEntry.revision();
 
-                    assert newChangeId > changeId.get();
+                    assert newChangeId >= changeId.get();
+
+                    // skip already processed update
+                    if (changeId.get() == newChangeId) {
```
sanpwc, line 312: Ass was mentioned above it seems that there's nothing to skip because all watches are resisted on start.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -195,16 +211,47 @@ public TableManager(
     /** {@inheritDoc} */
     @Override
     public void start() {
+        tablesCfg.tables().any().replicas().listen(replicasCtx -> {
+            if (!busyLock.enterBusy()) {
```
sanpwc, line 215: Let's extract it to explicit onReplicasChange() like it's done with onTableCreate(), onTableDelete(), etc. WDYT?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sanpwc: > Here is excess invocation, because we do nothing with peers in RaftGroupSercive, although recreate them (I know it is to start raft server if it is required, but maybe we can avoid of restarts of RaftGroupSercive).

Yep, I believe that, we should update internalTableImpl with new raft group services.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sanpwc: > How do we can sure that the leader won't change just after?

It's actually not a problem. In worst case we will have attempt to start several rebalances to the same set of target peers, because each onLeaderElected() will also start one. @kgusakov As far as I remember we've planned to propagate term to chagnePeers to skip such not-a-leader anymore rebalances.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1317,152 +1370,77 @@ private RuntimeException convertThrowable(Throwable th) {
...
+                        var newNodes = newPeers.stream().map(n -> new Peer(n.address())).collect(Collectors.toList());
 
-                    return CompletableFuture.allOf(changePeersFutures);
-                })
-                .whenComplete((res, th) -> {
-                    if (th != null) {
-                        setBaselineFut.completeExceptionally(th);
-                    } else {
-                        setBaselineFut.complete(null);
+                        raftGrpSvc.changePeersAsync(newNodes).join();
```
sanpwc, line 1426: Well, the core idea was - not to start changePeers process for stale notifications. Seems that it's not true right now, am I wrong?

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -0,0 +1,166 @@
...
+                        ).yield(),
+                        If.iif(value(partAssignmentsPendingKey).ne(partAssignmentsBytes),
+                                ops(
+                                        put(partAssignmentsPlannedKey, partAssignmentsBytes),
+                                        put(partChangeTriggerKey, ByteUtils.longToBytes(revision))
+                                ).yield(),
+                                ops(remove(partAssignmentsPlannedKey)).yield())),
+                ops().yield());
+
+        return metaStorageMgr.invoke(iif);
```
sanpwc, line 97: Let's also wrap it with busy lock.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1317,152 +1379,153 @@ private RuntimeException convertThrowable(Throwable th) {
...
-            throw new NodeStoppingException();
-        }
-        try {
-            setBaselineInternal(nodes);
-        } finally {
-            busyLock.leaveBusy();
-        }
+    private RaftGroupEventsListener raftGroupEventsListener(String tblName, int partNum,
+            String partId) {
+        return new RaftGroupEventsListener() {
```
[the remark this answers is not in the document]
kgusakov: Done

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
kgusakov: Guys, we must update raft group services, only after successful rebalance and configuration updates, but not in the listener of pending assignments - it's too early, this rebalance can be failed.

But, I agree, that we don't need to recreate the raft services here, I changed this code to:
- on the non-leader nodes just start new raft group nodes
- on the leader node run changePeersAsync on the already created raft group service from internalTable

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
kgusakov: Done

modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
```diff
@@ -165,7 +190,7 @@ public CompletableFuture<RaftGroupService> prepareRaftGroup(
         }
 
         try {
-            return prepareRaftGroupInternal(groupId, nodes, lsnrSupplier);
+            return prepareRaftGroupInternal(groupId, nodes, lsnrSupplier, () -> null);
```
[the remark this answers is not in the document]
kgusakov: Done

examples/src/main/java/org/apache/ignite/example/rebalance/RebalanceExample.java
```diff
@@ -1,201 +0,0 @@
...
- *     <li>Run the example in the IDE.</li>
- *     <li>
- *         When requested, start another two nodes using the CLI tool:
- *         {@code ignite node start --config=$IGNITE_HOME/examples/config/ignite-config.json my-first-additional-node}<br>
- *         {@code ignite node start --config=$IGNITE_HOME/examples/config/ignite-config.json my-second-additional-node}
- *     </li>
- *     <li>Press {@code Enter} to resume the example.</li>
- * </ol>
- */
-public class RebalanceExample {
```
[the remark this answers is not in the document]
kgusakov: Ooops, I thought these files generated from md

modules/core/src/main/java/org/apache/ignite/internal/causality/VersionedValue.java
kgusakov: Fixed

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
kgusakov: Calls of this util methods already covered by TableManager busy lock.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
kgusakov: updated according to our discussion

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
kgusakov: fixed

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
kgusakov: This call states inside the watch for partition pending assignments.

As far as I understand, according to the nature of ordered metastore events:
- this event can't be triggered before table creation (rebalance can't be triggered before it)
- this event can't be triggered after table drop (rebalance for partition can't be triggered for removed table)

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
```diff
@@ -0,0 +1,168 @@
...
+    private final String partId;
+
+    /** Partition number. */
+    private final int partNum;
+
+    /** Busy lock of parent component for synchronous stop. */
+    private IgniteSpinBusyLock busyLock;
+
+    /** Resolver that resolves a network address to cluster node. */
+    private final Function<NetworkAddress, ClusterNode> clusterNodeRslvr;
```
sanpwc, line 72: Unfortunately, it's illegal to use TopologyService-based cluster node resolver from TableManager here. Probably it's illegal to use it even in table manager, but it's out of our scope. Let's use union(assignments.stable, assignments.pending) in order to resolve peers.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
```diff
@@ -0,0 +1,168 @@
...
+
+        Entry plannedEntry = keys.get(partAssignmentsPlannedKey(partId));
+
+        List<ClusterNode> appliedPeers = peers
+                .stream()
+                .map(p -> clusterNodeRslvr.apply(
+                        NetworkAddress.from(p.getEndpoint().getIp() + ":" + p.getEndpoint().getPort())))
+                .collect(Collectors.toList());
+
+        tblConfiguration.change(ch -> {
```
sanpwc, line 140: Let's add a test, **within separate ticket**, that will check that everything works fine, if node files after tblConfiguration.change but before ms.invoke. I expect that it should work properly, but nevertheless the test will be useful.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
sanpwc: As was discussed, it'll be fixed later within separate ticket.

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
sanpwc: Agreed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
```diff
@@ -0,0 +1,168 @@
...
+    @Override
+    public void onLeaderElected() {
+        // TODO: IGNITE-16800 implement this method
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void onNewPeersConfigurationApplied(List<PeerId> peers) {
+        if (!busyLock.enterBusy()) {
+            throw new IgniteInternalException(new NodeStoppingException());
```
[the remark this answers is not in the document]
kgusakov: Fixed by no-op reaction.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
kgusakov: Fixed

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
kgusakov: Thanks, added this case to https://issues.apache.org/jira/browse/IGNITE-14873 in comments

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -195,16 +213,18 @@ public TableManager(
...
             @Override
             public CompletableFuture<?> onCreate(ConfigurationNotificationEvent<SchemaView> schemasCtx) {
                 return onSchemaCreate(schemasCtx);
             }
         });
 
-        ((ExtendedTableConfiguration) tablesCfg.tables().any()).assignments().listen(assignmentsCtx -> {
-            return onUpdateAssignments(assignmentsCtx);
-        });
+        ((ExtendedTableConfiguration) tablesCfg.tables().any()).assignments().listen(this::onUpdateAssignments);
```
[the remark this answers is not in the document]
kgusakov: Thanks, added this case to https://issues.apache.org/jira/browse/IGNITE-14873 in comments

sanpwc (approved):

kgusakov:
Merged to integration branch https://github.com/gridgain/apache-ignite-3/tree/ignite-14209

PR #769, opened 2022-04-09
IGNITE-16802 Shared RocksDB instance for the Raft Log Storage

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-16802

### DIFF
--- a/modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
+++ b/modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
@@ -43,6 +43,7 @@ import org.apache.ignite.internal.network.serialization.SerializationService;
 import org.apache.ignite.internal.network.serialization.UserObjectSerializationContext;
 import org.apache.ignite.internal.network.serialization.marshal.DefaultUserObjectMarshaller;
 import org.apache.ignite.lang.IgniteInternalException;
+import org.apache.ignite.lang.IgniteLogger;
 import org.apache.ignite.network.AbstractClusterService;
 import org.apache.ignite.network.ClusterLocalConfiguration;
 import org.apache.ignite.network.ClusterService;
@@ -56,6 +57,9 @@ import org.apache.ignite.network.NodeFinderFactory;
  * Cluster service factory that uses ScaleCube for messaging and topology services.
  */
 public class ScaleCubeClusterServiceFactory {
+    /** Logger. */
+    private static final IgniteLogger LOG = IgniteLogger.forClass(ScaleCubeClusterServiceFactory.class);
+
     /**
      * Creates a new {@link ClusterService} using the provided context. The created network will not be in the "started" state.
      *
@@ -78,6 +82,7 @@ public class ScaleCubeClusterServiceFactory {
         var messagingService = new DefaultMessagingService(messageFactory, topologyService, userObjectSerialization);
 
         return new AbstractClusterService(context, topologyService, messagingService) {
+
             private volatile ClusterImpl cluster;
 
             private volatile ConnectionManager connectionMgr;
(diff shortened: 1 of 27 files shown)


modules/raft/src/main/java/org/apache/ignite/raft/jraft/util/Utils.java
```diff
@@ -123,6 +126,29 @@ public static Future<?> runClosureInThread(ExecutorService executor, final Closu
         return runClosureInThread(executor, done, Status.OK());
     }
 
+    private static final long pid = ProcessHandle.current().pid();
```
sk0x50, line 129: This looks like a debug piece of code. Should it be removed?

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
```diff
@@ -157,10 +157,10 @@ public void stop() {
                     Thread.currentThread().interrupt();
 
                     throw new IgniteInternalException("Interrupted while waiting for the ClusterService to stop", e);
-                } catch (TimeoutException e) {
-                    throw new IgniteInternalException("Timeout while waiting for the ClusterService to stop", e);
                 } catch (ExecutionException e) {
                     throw new IgniteInternalException("Unable to stop the ClusterService", e.getCause());
+                } catch (TimeoutException ignored) {
+                    // Failed to leave gracefully
```
sk0x50, line 163: Perhaps, it makes sense to log a warning message here.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/JRaftServiceFactory.java
```diff
@@ -31,11 +31,11 @@ public interface JRaftServiceFactory {
     /**
      * Creates a raft log storage.
      *
-     * @param uri The log storage uri from {@link NodeOptions#getSnapshotUri()}
+     * @param groupId The name of the raft group.
      * @param raftOptions the raft options.
      * @return storage to store raft log entires.
      */
-    LogStorage createLogStorage(final String uri, final RaftOptions raftOptions);
+    LogStorage createLogStorage(final String groupId, final RaftOptions raftOptions);
```
sk0x50, line 38: Everything else works with the URI. Why is this change needed? It doesn't look very consistent.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/util/Utils.java
SammyVimes: You're right, I completely forgot about this

modules/raft/src/main/java/org/apache/ignite/raft/jraft/JRaftServiceFactory.java
SammyVimes: All log storages have the same URI in this PR. The only way to distinguish data in a storage is by raft group id. Hence the change

modules/raft/src/test/java/org/apache/ignite/raft/jraft/storage/impl/SharedVsNonSharedLogStorageBenchmark.java
```diff
@@ -0,0 +1,271 @@
...
+import org.apache.ignite.raft.jraft.option.RaftOptions;
+import org.apache.ignite.raft.jraft.storage.LogStorage;
+import org.apache.ignite.raft.jraft.storage.LogStorageProvider;
+import org.apache.ignite.raft.jraft.test.TestUtils;
+import org.apache.ignite.raft.jraft.util.Utils;
+
+/**
+ * Benchmark for shared versus non-shared log storage.
+ */
+public class SharedVsNonSharedLogStorageBenchmark {
```
sashapolo, line 46: please fix all code style issues in this class

sashapolo (approved):

PR #771, opened 2022-04-11
IGNITE-16823 .NET: Add cluster awareness to Compute

ptupitsyn:
* Include node id and name into handshake response.
* Establish connections to all known endpoints in background.
* During Compute calls, match target node name against active connections and send request directly to the correct node when possible.


isapego (approved):
Looks good

PR #775, opened 2022-04-13
IGNITE-16848 InternalTuple interface extracted from Row.

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-16848


modules/schema/src/main/java/org/apache/ignite/internal/schema/row/InnerTuple.java
```diff
@@ -0,0 +1,223 @@
...
+     */
+    Instant timestampValue(int col);
+
+    /**
+     * Checks whether the given column contains a null value.
+     *
+     * @param col Column index.
+     * @return {@code true} if this column contains a null value, {@code false} otherwise.
+     */
+    boolean hasNullValue(int col);
```
ademakov, line 222: Just a matter of personal taste, I would move this up, just after the count() method.

ibessonov: Why not

modules/schema/src/main/java/org/apache/ignite/internal/schema/row/BinaryTuple.java
```diff
@@ -0,0 +1,223 @@
...
+    int count();
+
+    /**
+     * Checks whether the given column contains a null value.
+     *
+     * @param col Column index.
+     * @return {@code true} if this column contains a null value, {@code false} otherwise.
+     */
+    boolean hasNullValue(int col);
+
```
ademakov, line 47: I'd suggest to add another method:

`ByteBuffer getRawData(int col)` -- it will return `null` for NULL and real sub-buffer to the underlying binary field otherwise.

ibessonov: So basically you imply that there's information about element size in the internal structure of the tuple. I disagree with this statement, this can only be applied to some specific implementations. And, I don't get why we would need such method in  the first place

ademakov: This interface already requires that there is information on where each element starts. It seems natural that you can also figure out where an element ends given its start and knowledge about tuple internals.

Also for varlen elements the size information must be present in any case.

In principle this method could be defined to return valid data only for varlens while for fixed-size elements it could be allowed to throw.

This method might be needed for:
  1. schema-less copying of individual columns to other tuples (e.g. by SQL-engine when doing projections, etc);
  2. cheap calculation of hash value for a byte field, etc.

ibessonov: > schema-less copying of individual columns to other tuples (e.g. by SQL-engine when doing projections, etc);

> cheap calculation of hash value for a VARBINARY element, etc.

Both of these statements imply that data encoding algorithm is known and the same for every implementation, which is simply not true.
Plus, I would want to have `Object[]`-based implementation, for which it would be impossible to implement a method like this.

ademakov: > Plus, I would want to have Object[]-based implementation, for which it would be impossible to implement a method like this.

Then this interface shouldn't be named `BinaryTuple`.

ibessonov: Yeah, about that... let's rename it. I can't think of a good name. We already have "org.apache.ignite.table.Tuple", so simply calling it Tuple is a bad idea. SchemalessTuple maybe? I don't know

ibessonov: > Then this interface shouldn't be named BinaryTuple.

Yes, I agree

SammyVimes (approved):

PR #781, opened 2022-04-15
IGNITE-16861 .NET: Fix TestCorrectHandshakeReturnsSuccess

ptupitsyn:


modules/platforms/dotnet/Apache.Ignite.Tests/RawSocketConnectionTests.cs
```diff
@@ -83,7 +83,8 @@ private static unsafe void ReadHandshake(NetworkStream stream)
 
             var str = Encoding.UTF8.GetString(msg);
 
-            Assert.AreEqual(110, msgSize, str);
+            // Different size is caused by different node names in response (PlatformTestNodeRunner or PlatformTestNodeRunner_2).
+            Assert.LessOrEqual(110, msgSize, str);
```
SammyVimes, line 87:
```suggestion
            StringAssert.Contains(str, "PlatformTestNodeRunner");
```
Makes more sense as there is no hard-coded byte length

SammyVimes (changes requested):

PR #782, opened 2022-04-17
IGNITE-16868 Fix build in IDEA

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-16868

### DIFF
--- a/modules/raft/pom.xml
+++ b/modules/raft/pom.xml
@@ -146,37 +146,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <id>default-compile</id>
-                        <configuration>
-                            <annotationProcessorPaths>
-                                <path>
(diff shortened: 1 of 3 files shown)


modules/raft/pom.xml
```diff
@@ -189,6 +158,20 @@
...
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>org.apache.ignite</groupId>
+                            <artifactId>ignite-network-annotation-processor</artifactId>
+                            <version>${project.version}</version>
+                        </path>
+                        <path>
+                            <groupId>org.openjdk.jmh</groupId>
+                            <artifactId>jmh-generator-annprocess</artifactId>
```
sashapolo, line 170: why is this change needed? The previous configuration was used to avoid the `jmh` processor during non-test build

SammyVimes: IDEA doesn't know about different executions, so it doesn't execute annotation processor at all for raft module. It's also inconsistent with other modules, which use jmh.

sashapolo (approved):

PR #784, opened 2022-04-18
IGNITE-16472 Incorporate CMG Manager into the node lifecycle

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-16472


modules/runner/src/main/java/org/apache/ignite/internal/app/LifecycleManager.java
```diff
@@ -0,0 +1,128 @@
...
+     * @throws NodeStoppingException If node stopping intention was detected.
+     */
+    synchronized void startComponents(IgniteComponent... components) throws NodeStoppingException {
+        for (IgniteComponent component : components) {
+            startComponent(component);
+        }
+    }
+
+    /**
+     * Callback that should be called after all components, that comprise an Ignite node, has been started. After this method completes,
```
rpuch, line 88: has been started -> have been started

rpuch (approved):

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -396,90 +390,83 @@ public CompletableFuture<Ignite> start(@Language("HOCON") @Nullable String cfg)
...
-                    baselineMgr,
-                    dataStorageMgr,
-                    distributedTblMgr,
-                    qryEngine,
-                    restComponent,
-                    clientHandlerModule,
-                    nettyBootstrapFactory
-            ));
-        }
+        lifecycleManager.stopNode();
```
sanpwc, line 469: We should only stop node if the status was STARTED. 
So let's use 
```
        if (status.getAndSet(Status.STOPPING) == Status.STARTED) {
            stopAllComponents();
        }
```
as stopNode impementation.

sanpwc (changes requested):

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
sashapolo: I've fixed this comment, but then I remembered that I actually did this on purpose: it is possible that a node may hang during join in which case it will never acquire the STARTED state. This will make it impossible to stop it in "afterEach" in tests, for example

PR #787, opened 2022-04-19
IGNITE-16611 Basic RocksDB-based MV-storage implementation & refactored interfaces

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-16611


SammyVimes (changes requested):

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/UuidIgniteRowId.java
```diff
@@ -0,0 +1,98 @@
...
+    public static IgniteRowId randomRowId() {
+        return new UuidIgniteRowId(UUID.randomUUID());
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void writeTo(ByteBuffer buf, boolean signedBytesCompare) {
+        assert buf.order() == ByteOrder.LITTLE_ENDIAN;
+
+        long mask = signedBytesCompare ? 0x0080808080808080L : 0x8000000000000000L;
```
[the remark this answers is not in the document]
ibessonov: Yeah

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/UuidIgniteRowId.java
```diff
@@ -0,0 +1,98 @@
...
+
+        buf.putLong(Long.reverseBytes(mask ^ uuid.getMostSignificantBits()));
+        buf.putLong(Long.reverseBytes(mask ^ uuid.getLeastSignificantBits()));
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public int compare(ByteBuffer buf, boolean signedBytesCompare) {
+        assert buf.order() == ByteOrder.LITTLE_ENDIAN;
```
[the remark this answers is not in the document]
ibessonov: Sure

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/UuidIgniteRowId.java
```diff
@@ -0,0 +1,98 @@
...
+    /** {@inheritDoc} */
+    @Override
+    public int compareTo(@NotNull IgniteRowId o) {
+        if (!(o instanceof UuidIgniteRowId)) {
+            throw new IllegalArgumentException(o.getClass().getName());
+        }
+
+        UuidIgniteRowId that = (UuidIgniteRowId) o;
+
+        int cmp = Long.compare(uuid.getMostSignificantBits(), that.uuid.getMostSignificantBits());
```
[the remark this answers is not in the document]
ibessonov: Here we use comparison as it's intended. The goal it to have "compare" method that's consistent with this one.
BTW, current implementation is consistent with UUID#compareTo

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,489 @@
...
+    /** Timestamp size in bytes. */
+    private static final int TIMESTAMP_SIZE = 2 * Long.BYTES;
+
+    /** Maximum possible size of the key. */
+    private static final int MAX_KEY_SIZE = /* partId */ ROW_ID_OFFSET + /* rowId */ MAX_ROW_ID_SIZE + /* timestamp */ TIMESTAMP_SIZE;
+
+    /** Threadlocal direct buffer instance to read keys from RocksDB. */
+    private static final ThreadLocal<ByteBuffer> MV_KEY_BUFFER = withInitial(() -> allocateDirect(MAX_KEY_SIZE).order(BIG_ENDIAN));
+
+    /** Threadlocal ob-heap byte buffer instance to use for key manipulations. */
```
rpuch, line 77: on-heap

modules/rocksdb-common/src/main/java/org/apache/ignite/internal/rocksdb/ColumnFamily.java
```diff
@@ -41,7 +41,8 @@ public class ColumnFamily {
     private final String cfName;
 
     /** Column family handle. */
-    private final ColumnFamilyHandle cfHandle;
+    // Temporarily made public until the integration with MV-store is completed.
+    public final ColumnFamilyHandle cfHandle;
```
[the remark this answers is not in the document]
ibessonov: I will change it in the future. This little change really simplified my test

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
```diff
@@ -48,31 +47,30 @@ public interface MvPartitionStorage {
      * @throws TxIdMismatchException If there's another pending update associated with different transaction id.
      * @throws StorageException If failed to write data to the storage.
      */
-    void addWrite(BinaryRow row, UUID txId) throws TxIdMismatchException, StorageException;
+    void addWrite(IgniteRowId rowId, @Nullable BinaryRow row, UUID txId) throws TxIdMismatchException, StorageException;
```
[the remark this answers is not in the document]
ibessonov: It's a full row, key+value

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/UuidIgniteRowId.java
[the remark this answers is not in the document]
ibessonov: I extracted this statement into a method and added few coments

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/IgniteRowId.java
```diff
@@ -0,0 +1,51 @@
...
+    void writeTo(ByteBuffer buf, boolean signedBytesCompare);
+
+    /**
+     * Compares row id with a byte buffer, previously ritten by a {@link #writeTo(ByteBuffer, boolean)} method.
+     *
+     * @param buf Input byte buffer with {@link java.nio.ByteOrder#BIG_ENDIAN} byte order.
+     * @param signedBytesCompare Defines properties of a binary comparator. {@code true} if bytes are compared as signed values,
+     *      {@code false} if unsigned.
+     */
+    int compare(ByteBuffer buf, boolean signedBytesCompare);
```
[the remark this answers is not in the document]
ibessonov: Ok

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/UuidIgniteRowId.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        return Long.compare(uuid.getLeastSignificantBits(), mask ^ buf.getLong());
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public int compareTo(@NotNull IgniteRowId o) {
+        if (!(o instanceof UuidIgniteRowId)) {
+            throw new IllegalArgumentException(o.getClass().getName());
```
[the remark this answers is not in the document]
ibessonov: Ok

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,496 @@
...
+    private final int partId;
+
+    /** RocksDb instance. */
+    private final RocksDB db;
+
+    /** Partitions column family. */
+    private final ColumnFamilyHandle cf;
+
+    /** Write options. */
+    private final WriteOptions writeOpts = new WriteOptions().setDisableWAL(true);
```
rpuch, line 98: Why is WAL being disabled here? I think the reason is worth mentioning in a comment.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,494 @@
...
+            // Here we prepare direct buffer to read key without timestamp. Shared direct buffer is used to avoid extra memory allocations.
+            ByteBuffer directBuffer = MV_KEY_BUFFER.get().position(0).limit(ROW_ID_OFFSET + igniteRowIdSize);
+
+            it.key(directBuffer);
+
+            // Comparison starts from the position of the row id.
+            directBuffer.position(ROW_ID_OFFSET);
+
+            // Return null if seek found a wrong key.
+            if (rowId.compareTo(directBuffer, false) != 0) {
```
rpuch, line 229: Let's introduce a constant with value `false` and name like `UNSIGNED_COMPARISION`

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,494 @@
...
+                return null;
+            }
+
+            // Get binary row from the iterator. It has the exact payload that we need.
```
rpuch, line 233: Too many comments, to my taste. This line seems obvious :)

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,494 @@
...
+            byte[] valueBytes = it.value();
+
+            assert valueBytes != null;
+
+            // Empty array means a tombstone.
+            if (valueBytes.length == 0) {
```
rpuch, line 239: I suggest to introduce a method called `isTombstone(byte[] bytes)` with `return bytes.length == 0;`, which would make the comment unneeded, but the code would become a bit more readable.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,494 @@
...
+                // Prepare direct buffer slice to read keys from the iterator. It's goind to be used to read row ids without timestamps.
+                // To enforce this, we limit it to a proper capacity.
+                ByteBuffer directBuffer = MV_KEY_BUFFER.get()
+                        .position(0).limit(ROW_ID_OFFSET + igniteRowIdSize).slice().order(BIG_ENDIAN);
+
+                while (true) {
+                    // This flag is used to skip row ids that were created after required timestamp.
+                    boolean found = true;
+
+                    // We should do it all the time. Here in particular it means one of two things:
```
rpuch, line 301: 'all the time'? Wouldn't 'after each seek' be clearer?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,494 @@
...
+                    //  - partition is empty;
+                    //  - iterator exhausted all the data in partition.
+                    if (invalid(it)) {
+                        return false;
+                    }
+
+                    // At this point, seekKeyBuf should contain row id that's above the one we already scanned, but not greater than any
+                    // other row id in partition. When we start, row id is filled with zeroes. Value during the iteration is described later
+                    // in this code. Now let's descrie what we'll find, assuming that iterator found something:
```
rpuch, line 310: describe

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,494 @@
...
+                    // Finish scan if nothing was found.
+                    if (invalid(it)) {
+                        return false;
+                    }
+
+                    // Read the actual key into a direct buffer.
+                    it.key(directBuffer);
+
+                    // To understand this condition please read that huge comment above.
+                    if (timestamp == null || !directBuffer.equals(seekKeyBufSliceWithoutTimestamp)) {
```
rpuch, line 333: The structure now is

```
if (timestamp == null || COND) {
    ACTION1
    if (timestamp != null) {
        ACTION2
    }
}
```

I suggest to transform it to

```
if (timestamp == null || COND) {
    ACTION1
}
if (timestamp != null) {
    ACTION2
}
```

This seems (arguably) easier to comprehend.

Further, `COND` can be extracted to a boolean variable with a clarifying name like `didNotFindExactRowId` (if I understood it correctly) to facilitate comprehension even more.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,494 @@
...
+                            // At this point, "seekKeyBuf" has row id that exists in partition.
+                            it.seek(seekKeyBuf.array());
+
+                            // Iterator may not be valid if that row was created after required timestamp.
+                            if (invalid(it)) {
+                                return false;
+                            }
+
+                            // Or iterator may still be valid even if there's no version for required timestamp. In this case row id
+                            // itself will be different and must check it.
```
rpuch, line 352: 'and WE must check it', otherwise, it reads is if row id had to check it

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,494 @@
...
+                    // To understand this condition please read that huge comment above.
+                    if (timestamp == null || !directBuffer.equals(seekKeyBufSliceWithoutTimestamp)) {
+                        // Copy actual row id into a "seekKeyBuf" buffer.
+                        GridUnsafe.copyMemory(
+                                null, GridUnsafe.bufferAddress(directBuffer) + ROW_ID_OFFSET,
+                                seekKeyBuf.array(), GridUnsafe.BYTE_ARR_OFF + ROW_ID_OFFSET,
+                                igniteRowIdSize
+                        );
+
+                        // Perform additional "seek" if timestamp is not null. Motivation for it is described in comments above.
```
rpuch, line 341: Maybe it would make sense to say that we are here because we ran out of the 'previous' row-related versions, so we must looks for a timestamp-matching version of the **next** row? I mean, when scanning, we need to find a version in each row. Reading the existing comments, I did not see this statement explicitly (or maybe I missed it).

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,494 @@
...
+                    // This one might look tricky. We finished processing next row. There are three options:
+                    //  - "found" flag is false - there's no fitting version of the row. We'll continue to next iteration;
+                    //  - value is empty, we found a tombstone. We'll continue to next iteration as well;
+                    //  - value is not empty and everything's good. We'll cache it and return from method.
+                    // In all three cases we need to prepare the value of "seekKeyBuf" so that it has not-yet-scanned row id in it.
+                    // the only valid way to do so is to treat row id payload as one big unsigned integer in Big Endian and increment it.
+                    // It's important to note that increment may overflow. In this case "carry flag" will go into incrementing partition id.
+                    // This is fine for three reasons:
+                    //  - iterator has an upper bound, following "seek" will result in invalid iterator state.
+                    //  - partition id iself cannot be overflown, because it's limited with a constant less than 0xFFFF. It's something like
```
rpuch, line 370: I suggest adding the following assertion (on the mentioned constant) to the static initializer of this class:

```
assert MAX_PARTITION_ID < 0xFFFF;
```

To make sure that this invariant is kept in the future.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,494 @@
...
+                    //    65500, I think.
+                    //  - "seekKeyBuf" buffer value will not be used after that, so it's ok if we corrupt its data (in every other instance,
+                    //    buffer starts with a valid partition id, which is set during buffer's initialization).
+                    for (int i = ROW_ID_OFFSET + igniteRowIdSize - 1;; i--) {
```
rpuch, line 374: Let's extract this incrementation to a method of its own

rpuch:
I think we need a test that makes sure that `hasNext()` of the Cursor returned by `scan()` is idempotent (I mean, if you called it twice, it only advances the cursor once).

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,496 @@
...
+
+    /** Size of the {@link IgniteRowId} identifiers for this particular partition storage. */
+    private final int igniteRowIdSize;
+
+    /**
+     * Partition ID (should be treated as an unsigned short).
+     *
+     * <p/>Partition IDs are always stored in the big endian order, since they need to be compared lexicographically.
+     */
+    private final int partId;
```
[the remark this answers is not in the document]
ibessonov: Sure

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
ibessonov: I'll remove this call for now. Storage state recovery procedure is not yet specified properly, so it's too early to disable WAL

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,496 @@
...
+                        .order(BIG_ENDIAN)
+                        .putShort((short) partId)
+        );
+
+        upperBound = new Slice(partitionEndPrefix());
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void addWrite(IgniteRowId rowId, @Nullable BinaryRow row, UUID txId) throws TxIdMismatchException, StorageException {
```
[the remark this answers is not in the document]
ibessonov: As of right now, it's ok. We have not yet started implementing transactions, there's a chance that behavior or API will change slightly in the future

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,496 @@
...
+        // Prepare a buffer with partition id and row id.
+        ByteBuffer keyBuf = prepareHeapKeyBuf(rowId);
+
+        try {
+            // Check concurrent transaction data.
+            if (RocksDB.NOT_FOUND != db.get(cf, keyBuf.array(), 0, keyBuf.position(), BYTE_EMPTY_ARRAY, 0, 0)) {
```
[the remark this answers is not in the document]
ibessonov: There is a test for this case. Version is created during commit, and there can be no two pending writes for the same key in different transactions

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,494 @@
...
+            db.put(cf, writeOpts, keyBuf.array(), 0, keyBuf.position(), valueBytes, 0, valueBytes.length);
+        } catch (RocksDBException e) {
+            throw new StorageException(e);
+        }
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public @Nullable BinaryRow read(IgniteRowId rowId, @Nullable Timestamp timestamp) {
+        // Prepare a buffer with partition id and row id.
```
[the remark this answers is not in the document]
ibessonov: I'll remove the comment :)

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
ibessonov: I don't know, IDEA shows you names of parameters in such cases, and you barely ever read code using different software

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
ibessonov: My idea was - the more comments, the better. This adds consistency even if comment is obvious - method is split into small digestible sections

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
ibessonov: Ok

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -0,0 +1,494 @@
...
+
+        // Size of the buffer to seek values. Should fit partition id, row id and maybe timestamp, if it's not null.
+        int seekKeyBufSize = ROW_ID_OFFSET + igniteRowIdSize + (timestamp == null ? 0 : TIMESTAMP_SIZE);
+
+        // Here's seek buffer itself. Originally it contains a valid partition id, row id payload that's filled with zeroes, and maybe
+        // a timestamp value. Zero row id guarantees that it's lexicographically less then or equal to any other row id stored in the
+        // partition.
+        // Byte buffer from a thread-local field can't be used here, because of two reasons:
+        //  - no one guarantees that there will only be a single cursor;
+        //  - no one guarantees that returned cursor will not be used other threads.
```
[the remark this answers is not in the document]
ibessonov: Thank you!

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
ibessonov: Maybe

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
ibessonov: I don't think that you suggest equivalent change. "timestamp != null" is not enough for ACTION2. I believe that two consecutive if-s will be more complicated, especially with the necessity to introduce a variable, which value only makes sense for the first "if"

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
ibessonov: Sure

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
ibessonov: Really? There's no point in it, it's a core invariant of the system

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
ibessonov: Ok

SammyVimes (approved):

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
ibessonov: I don't know what to reply. Here we're looking for the matching version of "current" (known) row id, not a new (next) one. It's the last clause in that huge comment with many different options

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
rpuch: My bad. I meant `if (COND)` in the second `if`. In this case the variable is used in both `if`s (so twice), and also it adds some explanation to its own meaning.

To understand two embedded ifs, you need to think about them at the same time. To understand two consequtive ifs, you may think about them in sequence, independently, so consecutive ifs are usually easier to comprehend. In this case, I had to mentally apply the suggested transformation to the 'plain' structure to actually understand what is going on in this piece of code.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
rpuch: I feel really unconfortable when a core invariant is not verified with some test or production code :) Because it is so easy to break an invariant that is not verified automatically.

If it is already verified, then ok.

ibessonov: We have a specific constant, INDEX_PARTITION which is 0xFFFF. Regular partitions can't reach it, I see no point in testing it. It's not obvious, I get it, but this is a part of the design

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
ibessonov: Ok, I give up, let's see if it gets better

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/UuidIgniteRowId.java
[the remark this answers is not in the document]
ibessonov: I changed this code completely, there's no point in proving this at this point.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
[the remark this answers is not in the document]
ibessonov: txId is used now

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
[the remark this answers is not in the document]
ibessonov: I was wrong. Now I added few checks in AbstractMvPartitionStorageTest#testAddWrite to test this case.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/UuidIgniteRowId.java
```diff
@@ -0,0 +1,155 @@
...
+     * The least significant 64 bits.
+     */
+    private final long leastSigBits;
+
+    /**
+     * Constructor.
+     *
+     * @param uuid UUID.
+     */
+    public UuidIgniteRowId(UUID uuid) {
```
ascherbakoff, line 44: I think we need to pass a partition here.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/IgniteRowId.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Interface that represents row id in primary index of the table.
+ *
+ * @see MvPartitionStorage
+ */
+public interface IgniteRowId extends Comparable<IgniteRowId> {
+    /**
+     * Returns a partition id for current row id.
+     */
+    int partitionId();
```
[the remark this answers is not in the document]
ibessonov: It's not mandatory by our code style, and usually it contains the same text as the description. I think we can ignore it in such cases

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/IgniteRowId.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package org.apache.ignite.internal.storage;
+
+/**
+ * Interface that represents row id in primary index of the table.
+ *
+ * @see MvPartitionStorage
+ */
+public interface IgniteRowId extends Comparable<IgniteRowId> {
```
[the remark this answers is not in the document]
ibessonov: Ok

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/UuidIgniteRowId.java
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Writes row id into a byte buffer. Binary row representation should match natural order defined by {@link #compareTo(Object)} when
+     * comparing lexicographically.
+     *
+     * @param buf Output byte buffer with {@link java.nio.ByteOrder#BIG_ENDIAN} byte order.
+     * @param signedBytesCompare Defines properties of a target binary comparator. {@code true} if bytes are compared as signed values,
+     *      {@code false} if unsigned.
+     */
+    public void writeTo(ByteBuffer buf, boolean signedBytesCompare) {
```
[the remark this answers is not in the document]
ibessonov: Right now it's more like an artifact of comparability. I'll change it

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/UuidIgniteRowId.java
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Compares row id with a byte buffer, previously written by a {@link #writeTo(ByteBuffer, boolean)} method.
+     *
+     * @param buf Input byte buffer with {@link java.nio.ByteOrder#BIG_ENDIAN} byte order.
+     * @param signedBytesCompare Defines properties of a binary comparator. {@code true} if bytes are compared as signed values,
+     *      {@code false} if unsigned.
+     * @return A negative integer, zero, or a positive integer as this row id is less than, equal to, or greater than the specified row id.
+     */
+    public int compareTo(ByteBuffer buf, boolean signedBytesCompare) {
```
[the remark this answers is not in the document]
ibessonov: If I understand you correctly, we don't. In this specific case, the parameter will be a thread-local direct buffer, that we use to read values from RocksDB without extra allocations. It's not permanently stored anywhere.
And you're right, it's kinda a copy of UUID class, just adapted a little bit to our API.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/UuidIgniteRowId.java
```diff
@@ -0,0 +1,155 @@
...
+        this.leastSigBits = leastSigBits;
+    }
+
+    /**
+     * Returns {@link UuidIgniteRowId} instance based on {@link UUID#randomUUID()}. Highest two bytes of the most significant bits part will
+     * encode partition id.
+     *
+     * @param partitionId Partition id.
+     */
+    public static IgniteRowId randomRowId(int partitionId) {
```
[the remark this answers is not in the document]
ibessonov: I this I'll add TODO for it. Do we have JIRA issue for adapting newer UUID formats?
I'd say that it is out of scope of current PR.

Anyways, why are we so concerned about collisions in UUID v4? Are there any real world examples of this ever happening? Like, ever in the history of using secure random?

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/UuidIgniteRowId.java
ibessonov: Ok, not a big deal

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
```diff
@@ -29,43 +29,56 @@
...
  * although there are very important differences that will be addressed in the future.
  */
-public interface MvPartitionStorage {
+public interface MvPartitionStorage extends AutoCloseable {
     /**
      * Reads the value from the storage as it was at the given timestamp. {@code null} timestamp means reading the latest value.
      *
-     * @param key Key.
+     * @param rowId Row id.
      * @param timestamp Timestamp.
```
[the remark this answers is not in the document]
ibessonov: Not nullable anymore

ascherbakoff (approved):

PR #790, opened 2022-04-22
IGNITE-16890 Fix JVM crash

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-16890

### DIFF
--- a/modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/impl/RocksDbSharedLogStorage.java
+++ b/modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/impl/RocksDbSharedLogStorage.java
@@ -115,11 +115,11 @@ public class RocksDbSharedLogStorage implements LogStorage, Describer {
     /** RW lock. */
     private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
 
-    /** Read lock. */
-    private final Lock readLock = this.readWriteLock.readLock();
+    /** Storage use lock. Non-exclusive. */
+    private final Lock useLock = this.readWriteLock.readLock();
 
-    /** Write lock. */
-    private final Lock writeLock = this.readWriteLock.writeLock();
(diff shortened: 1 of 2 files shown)


modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/impl/RocksDbSharedLogStorage.java
```diff
@@ -532,11 +548,13 @@ private void addConfBatch(LogEntry entry, WriteBatch batch) throws RocksDBExcept
      */
     private boolean executeBatch(WriteBatchTemplate template) {
         this.readLock.lock();
-        if (this.db == null) {
+
+        if (stopped) {
```
sashapolo, line 552: I would prefer moving this part under the try-finally block

SammyVimes: Done

sashapolo (approved):

modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/impl/RocksDbSharedLogStorage.java
```diff
@@ -123,8 +121,11 @@ private static class EmptyWriteContext implements WriteContext {
     /** Write lock. */
     private final Lock writeLock = this.readWriteLock.writeLock();
```
rpuch, line 122: It seems that `readLock` and `writeLock` are not optimal names here, because they are not about reading and writing (both for read/write operations in this class, readLock is taken), but for 'managing the service' (start/stop/reset) and 'using' it. How about renaming them to `manageLock` and `useLock`, or `exclusiveLock` and `nonExclusiveLock`?

Just a suggestion :)

rpuch (approved):

PR #811, opened 2022-05-18
IGNITE-16999 Sql. Refactor IgniteBuiltInMethod and IgniteMethod classes

zstan:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/AsyncRootNode.java
```diff
@@ -80,7 +80,7 @@
      *
      * @see #lastRow
      */
-    private boolean firstRequest = true;
+    private volatile boolean firstRequest = true;
```
korlov42, line 83: do we really need this volatile?

korlov42:
Why don't you delete IgniteBuiltInMethod?

korlov42 (approved):

PR #814, opened 2022-05-18
IGNITE-16933 PageMemory-based MV storage implementation

rpuch:
https://issues.apache.org/jira/browse/IGNITE-16933


modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/LinkRowId.java
```diff
@@ -0,0 +1,45 @@
...
+    private final long rowLink;
+
+    public LinkRowId(long rowLink) {
+        this.rowLink = rowLink;
+    }
+
+    @Override
+    public int partitionId() {
+        long pageId = PageIdUtils.pageId(rowLink);
+        return PageIdUtils.partitionId(pageId);
```
ibessonov, line 39: You can pass rowLink directly into "partitionId", it's fine. We don't abuse data format here, it's well documented. Or, if you wish to leave it this way, please add a separating line :)

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PageMemoryMvPartitionStorage.java
```diff
@@ -0,0 +1,515 @@
...
+    }
+
+    @Nullable
+    private ByteBufferRow findRowVersionByTimestamp(VersionChain versionChain, Timestamp timestamp) {
+        long nextRowPartitionlessLink = versionChain.headLink();
+
+        while (true) {
+            RowVersion rowVersion = findRowVersion(nextRowPartitionlessLink);
+            Timestamp rowVersionTs = rowVersion.timestamp();
+            if (rowVersionTs != null && rowVersionTs.beforeOrEquals(timestamp)) {
```
ibessonov, line 220: I would actually move a comparison method into RowVersion. First reason - it bloats the API of Timestamp. Seconds - it allows you to store timestamp as two longs inside of RowVersion, which is nice. Third - no null-checks and stuff.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PageMemoryMvPartitionStorage.java
```diff
@@ -0,0 +1,515 @@
...
+    private void insertRowVersion(RowVersion rowVersion) {
+        try {
+            rowVersionFreeList.insertDataRow(rowVersion);
+        } catch (IgniteInternalCheckedException e) {
+            throw new StorageException("Cannot store a row version", e);
+        }
+    }
+
+    @Override
+    public @Nullable BinaryRow addWrite(RowId rowId, @Nullable BinaryRow row, UUID txId) throws TxIdMismatchException, StorageException {
```
ibessonov, line 273: As I already mentioned, this method is not effective

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PageMemoryMvPartitionStorage.java
```diff
@@ -0,0 +1,515 @@
...
+        try {
+            versionChainFreeList.removeDataRowByLink(currentVersionChain.link());
+        } catch (IgniteInternalCheckedException e) {
+            throw new StorageException("Cannot remove chain version", e);
+        }
+    }
+
+    @Override
+    public void commitWrite(RowId rowId, Timestamp timestamp) throws StorageException {
+        VersionChain currentVersionChain = findVersionChainForModification(rowId);
```
ibessonov, line 348: I don't think that "VersionChain" is a good name for what we have. It's more like a descriptor, a head. Chain is what it refers to. Also, it's easy to confuse with "Versioned Chain", the name that's used in many sources about MV storages.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersion.java
```diff
@@ -0,0 +1,137 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Represents row version inside row version chain.
+ */
+public class RowVersion extends StorableBase {
+    /**
+     * A 'timestamp' representing absense of a timestamp.
+     */
+    public static final Timestamp NULL_TIMESTAMP = new Timestamp(Long.MIN_VALUE, Long.MIN_VALUE);
```
ibessonov, line 35: I proposed using MAX_VALUE, what's the reasoning here?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersion.java
```diff
@@ -0,0 +1,137 @@
...
+    boolean isCommitted() {
+        return timestamp != null;
+    }
+
+    @Override
+    public int size() {
+        return timestampStoreSize() + nextLinkStoreSize() + valueStoreSize();
+    }
+
+    private int timestampStoreSize() {
```
ibessonov, line 116: Can these be constants instead of methods? It's very unusual for Ignite to see code like this.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersion.java
```diff
@@ -0,0 +1,137 @@
...
+        return PartitionlessLinks.PARTITIONLESS_LINK_SIZE_BYTES;
+    }
+
+    private int valueStoreSize() {
+        return Integer.BYTES + value.limit();
+    }
+
+    @Override
+    public int headerSize() {
+        return 0;
```
ibessonov, line 130: I specifically mentioned header sizes in JIRA. Please put "size()" here. This is a bug right now

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersionFreeList.java
```diff
@@ -0,0 +1,120 @@
...
+
+        this.statHolder = statHolder;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected long allocatePageNoReuse() throws IgniteInternalCheckedException {
+        return pageMem.allocatePage(grpId, INDEX_PARTITION, defaultPageFlag);
```
ibessonov, line 86: I believe other structures store partitionId in a field, please check it. INDEX_PARTITION is only applicable in volatile case, persistent storage will need a valid partition number here.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/VersionChain.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    public IoVersions<? extends AbstractDataPageIo> ioVersions() {
+        return VersionChainDataIo.VERSIONS;
+    }
+
+    @Override
+    public String toString() {
+        return "VersionChain{"
```
ibessonov, line 105: Is this a auto-generated toString? Usually we use S.toString or manual formatting with brackets []

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/VersionChainFreeList.java
```diff
@@ -0,0 +1,120 @@
...
+
+        this.statHolder = statHolder;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected long allocatePageNoReuse() throws IgniteInternalCheckedException {
+        return pageMem.allocatePage(grpId, INDEX_PARTITION, defaultPageFlag);
```
ibessonov, line 86: Same here

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/VersionChain.java
```diff
@@ -0,0 +1,110 @@
...
+        return 2 * Long.BYTES;
+    }
+
+    private int headLinkStoreSize() {
+        return PartitionlessLinks.PARTITIONLESS_LINK_SIZE_BYTES;
+    }
+
+    @Override
+    public int headerSize() {
+        return 0;
```
ibessonov, line 95: This is not a bug, but I'd recommend using meaningful value here. It'll help you later with optimal "read" implementation.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/io/AbstractDataPageIo.java
```diff
@@ -611,24 +626,73 @@ protected int getDataOffset(long pageAddr, int itemId, int pageSize) {
 
         assert directCnt > 0 : "itemId=" + itemId + ", directCnt=" + directCnt + ", page=" + printPageLayout(pageAddr, pageSize);
 
+        final int directItemId;
```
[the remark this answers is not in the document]
rpuch: It makes sense to declare this variable `final` because the code that follows it is pretty long (longer than 3-5 lines). It makes the code more easily comprehensible (the reader easily sees that the variable never gets reassigned). This allows to understand the code more quickly (and spare time).

In the original code, for example, `itemId` was reassigned, and you had to read really carefully (=slowly) to spot it.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/LinkRowId.java
rpuch: Why do we need a separating line here? Separating lines separate blocks of code, like '3 lines doing a coherent job, then 2 lines doing another coherent job'. Here, both lines do the same coherent job, so they form one 'block', so no separating lines seems to be needed.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PageMemoryMvPartitionStorage.java
```diff
@@ -0,0 +1,515 @@
...
+import org.apache.ignite.lang.IgniteInternalCheckedException;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Implementation of {@link MvPartitionStorage} using Page Memory.
+ *
+ * @see MvPartitionStorage
+ */
+public class PageMemoryMvPartitionStorage implements MvPartitionStorage {
```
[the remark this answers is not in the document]
rpuch: It is modeled after `PageMemoryPartitionStorage` which, for now, only works with volatile page memory, but has no `Volatile` in its name. It does not seem coherent to rename just one of them. Are you sure you want just `PageMemoryMvPartitionStorage` to be renamed?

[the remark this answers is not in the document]
rpuch: Also, it contains TODOs about the persistent case, so it will probably handle both volatile and persistent cases (at least, it's how `PageMemoryPartitionStorage` (no MV) looks like)

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PageMemoryMvPartitionStorage.java
```diff
@@ -0,0 +1,515 @@
...
+    private static final byte[] TOMBSTONE_PAYLOAD = new byte[0];
+    private static final Predicate<BinaryRow> MATCH_ALL = row -> true;
+
+    private final int partitionId;
+    private final int groupId;
+
+    private final VersionChainFreeList versionChainFreeList;
```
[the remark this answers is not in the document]
rpuch: Why do I need to separate them with empty lines?

It makes sense to separate fields with blank lines if you want to logicallly group them. 2 fields are strongly connected by their meaning, ok, we put them in the same block. Fields inside a block are declared without blank lines between them. Blocks are separated with blank lines.

If we just ALWAYS separate field declarations with blank lines, we lose the ability to group them logically, but we gain nothing in return.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PageMemoryMvPartitionStorage.java
rpuch: It's a very natural operation to have on `Timestamp`, it basically pertains to the abstraction (note that it uses the domain-specific 'before' instead of the general 'less'). We bloat an API when we add an operation that is not too cohesive with the type, but here it's not the case. Making a user writing in terms of `compareTo(...) <=> x` produce less readable code. I am sure that we will need to compare timestamps in '<=' sense a lot, so this method seems really useful.

Storing a timestamp as two longs means breaking encapsulation of Timestamp (who knows how will its internals change in the future?). It's an optimization, and at this point it's premature for sure.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PageMemoryMvPartitionStorage.java
rpuch: What do you mean 'not effective'?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PageMemoryMvPartitionStorage.java
rpuch: Well, it can be regarded as a lazy chain, or like a reference to a chain (and we usually do not distinguish references to objects and the objects themselves). When in Java you have a `LinkedList` object, its fields actually only contain reference to next items, not the whole list content, as `ArrayList` does. With `VersionChain`, the only difference is that it does not 'lazy-load' itself, but the 'dereference' is done using an external (to `VersionChain`) code. So it seems acceptable.

As 'Versioned Chain', I did not find such a term, could you please point me to the corresponding documents? But, googling by 'Version Chain MVCC', I found pages talking about Version Chains, which seem to be the same thing that is denoted by `VersionChain` in this PR. So it does not seem to be too confusing...

Anyway, if you still don't like it, what alternative name would you propose?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersion.java
rpuch: A local timestamp cannot be negative, so a negative value seems a natural fit for a role of a 'special value'

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersion.java
rpuch: I introduced the methods solely as an aid to the reader. When you see '4 + 8 + 8', you have to somehow lookup the information (by looking at other code) that will help you see WHY it's '4 + 8 + 8'. With the methods in place, it's self-explanatory, you don't need to read other code, you spend less time understanding the `size()` method.

I know that in Ignite people don't do that, and it's a pity. My opinion is that this needs to be improved. The performance should not suffer as the JIT compiler will easily inline it all, but the readability becomes a bit better.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersion.java
rpuch: `headerSize()` is only used for fragmented case to make sure that the 'header' is not torn apart and it always remains on the first page. Why is it a bug if we return 0 here?

Please note that in Ignite 2, the only implementation that returns something non-zero from this method is `MvccDataRow`. Other implementations either return 0, or just throw an exception.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersionFreeList.java
rpuch: I copied it from `TableFreeList`, it uses the index partition. But in Ignite 2 the production code never uses index partition. I will talk to Kirill about this.

rpuch: Kirill told me the following: when a free list is one per region, we use index partition; if it's per partition, we use the data partition. In this case, it's global, so index partition seems fine.

If you have other concerns, please let me know.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/PageMemoryDataRegion.java
```diff
@@ -33,4 +34,11 @@ public interface PageMemoryDataRegion {
      */
     @Nullable
     PageMemory pageMemory();
+
+    /**
+     * Returns page memory or throws an exception if not started.
+     */
+    default PageMemory requiredPageMemory() {
```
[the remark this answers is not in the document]
rpuch: Removed the added method and moved null-check to `pageMemory()` so that it is not `@Nullable` anymore (as from the analysis of the current code, no caller of `pageMemory()` tolerates null value)

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/StorableBase.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.pagememory;
+
+/**
+ * Base of {@link Storable} containing useful trivial implementations of some methods.
+ */
+public abstract class StorableBase implements Storable {
```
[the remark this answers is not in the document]
rpuch: Removed the class. It seemed to be useful because originally it was used by 2 classes, but now it is not needed at all

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersion.java
rpuch: Please disregard the previous comment, I re-read your comment and understood that I was fighting windmills. I changed it to constants (where it's possible).

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/io/AbstractDataPageIo.java
```diff
@@ -673,6 +737,18 @@ public DataPagePayload readPayload(final long pageAddr, final int itemId, final
...
 
+    /**
+     * Returns {@code true} iff an item with given ID exists in a page at the given address.
+     *
+     * @param pageAddr address where page data begins in memory
+     * @param itemId   item ID to check
+     * @param pageSize size of the page in bytes
+     * @return {@code true} iff an item with given ID exists in a page at the given address
+     */
+    public boolean itemExists(long pageAddr, int itemId, final int pageSize) {
```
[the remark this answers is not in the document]
rpuch: We discussed this, you suggested to add the following validation: 'if this is a direct itemId, but some indirect itemId references it, we should treat this itemId as non-existent'. This is enforced by the following line:

    if (anyIndirectItemPointsAtDirectItem(pageAddr, itemId, directCnt, indirectCnt)) {

Is this what you mention in your comment now? If not, let's discuss it deeper.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersion.java
rpuch: It was decided to postpone this till https://issues.apache.org/jira/browse/IGNITE-17008 is implemented

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/VersionChain.java
rpuch: Fixed

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersionFreeList.java
rpuch: I now create these free lists one per MvStorage instance and each of them gets the corresponding partitionId.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersion.java
rpuch: After a conversation, it was decided that a meaningful header size is better as it makes reading code simpler. I implemented the method as suggested.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PageMemoryMvPartitionStorage.java
rpuch: We had a conversation about this. There are pros and cons in each of the proposed naming schemes (VersionChain/RowVersion, VersionHistory/VersionChain, etc), so for now I leave it as is, we'll be able to revisit the naming later. I added a short clarification to the class javadoc.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/VolatilePageMemoryDataRegion.java
```diff
@@ -62,32 +68,77 @@ public void start() {
...
-                    grpId,
-                    pageMemory,
-                    PageLockListenerNoOp.INSTANCE,
-                    metaPageId,
-                    true,
-                    null,
-                    PageEvictionTrackerNoOp.INSTANCE,
-                    IoStatisticsHolderNoOp.INSTANCE
-            );
+            this.tableFreeList = createTableFreeList(pageMemory, FREE_LIST_GROUP_ID);
```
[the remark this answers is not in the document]
rpuch: Thanks for noting this, I did not know about this feature of page memory free lists. Implemented the suggested change.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/io/AbstractDataPageIo.java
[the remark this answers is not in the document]
rpuch: From our conversation, it turns out that it was actually the additional validation we agreed to add.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/datapage/DataPageReader.java
```diff
@@ -0,0 +1,182 @@
...
+import org.apache.ignite.internal.pagememory.io.AbstractDataPageIo;
+import org.apache.ignite.internal.pagememory.io.DataPagePayload;
+import org.apache.ignite.internal.pagememory.metric.IoStatisticsHolder;
+import org.apache.ignite.lang.IgniteInternalCheckedException;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Contains logic for reading values from page memory data pages (this includes handling multy-page values).
+ */
+public abstract class DataPageReader<T> {
```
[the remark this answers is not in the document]
rpuch: I added `DataPageReader#traversal()` operation which allows to avoid code duplication when using different traversals. (The traversals themselves are stateful, but it's possible to cache their instances in thread locals, which I did).

Using this new abstraction, I optimized search for version by a timestamp and added ability to only load version value when needed.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PageMemoryMvPartitionStorage.java
rpuch: This is optimized now. BTW, I also optimized commit (only timestamp is written instead of all the version data).

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PageMemoryMvPartitionStorage.java
```diff
@@ -72,8 +83,17 @@ public PageMemoryMvPartitionStorage(int partitionId, TableView tableConfig, Page
 
         groupId = StorageUtils.groupId(tableConfig);
 
-        versionChainFreeList = ((VolatilePageMemoryDataRegion) dataRegion).versionChainFreeList();
-        rowVersionFreeList = ((VolatilePageMemoryDataRegion) dataRegion).rowVersionFreeList();
+        try {
+            versionChainFreeList = createVersionChainFreeList(dataRegion.pageMemory(), groupId, partitionId);
```
ibessonov, line 87: Why did you decide to create these lists per partition? It would be better to turn them into parameters. Maybe there was a slight misunderstanding between us -  volatile region should have shared freelists to better utilize memory, it's an optimization that we should do. Persistent regions must have separate freelists for each partition. Right now we have separate lists in volatile region, which is not what I expect.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/ReadLatestRowVersion.java
```diff
@@ -0,0 +1,99 @@
...
+
+        timestamp = Timestamps.readTimestamp(pageAddr, payload.offset() + RowVersion.TIMESTAMP_OFFSET);
+        nextLink = PartitionlessLinks.readFromMemory(pageAddr, payload.offset() + RowVersion.NEXT_LINK_OFFSET);
+
+        if (!loadValue.test(timestamp)) {
+            result = new RowVersion(partitionIdFromLink(link), firstFragmentLink, timestamp, nextLink, null);
+            return STOP_TRAVERSAL;
+        }
+
+        return readRowVersionValue.consumePagePayload(link, pageAddr, payload, null);
```
ibessonov, line 69: Same problem might be present here, I don't see where you skip the header

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersion.java
```diff
@@ -38,10 +41,24 @@ public class RowVersion extends StorableBase {
...
+    private static final int TIMESTAMP_STORE_SIZE_BYTES = 2 * Long.BYTES;
+    private static final int NEXT_LINK_STORE_SIZE_BYTES = PartitionlessLinks.PARTITIONLESS_LINK_SIZE_BYTES;
+    private static final int VALUE_SIZE_STORE_SIZE_BYTES = Integer.BYTES;
+
+    public static final int TIMESTAMP_OFFSET = 0;
+    public static final int NEXT_LINK_OFFSET = TIMESTAMP_STORE_SIZE_BYTES;
+    public static final int VALUE_SIZE_OFFSET = NEXT_LINK_OFFSET + NEXT_LINK_STORE_SIZE_BYTES;
+    public static final int VALUE_OFFSET = VALUE_SIZE_OFFSET + VALUE_SIZE_STORE_SIZE_BYTES;
+
+    private final int partitionId;
```
ibessonov, line 53: Can you please explain, why we need two fields if partition id can be derived from the link value?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersion.java
```diff
@@ -38,10 +41,24 @@ public class RowVersion extends StorableBase {
...
+    private long link;
+
     @Nullable
     private final Timestamp timestamp;
     private final long nextLink;
     private final int valueSize;
+    @IgniteToStringExclude
```
ibessonov, line 60: I know that you are against these comments, but please use line separators more often, with these changes code becomes more and more inconsistent.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/ScanVersionChainByTimestamp.java
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * First it's {@code true} (this means that we traverse first slots of versions of the Version Chain using NextLink);
+     * then it's {@code false} (when we found the version we need and we read its value).
+     */
+    private boolean lookingForVersion = true;
+
+    private final ReadRowVersionValue readRowVersionValue = new ReadRowVersionValue();
+
+    {
```
ibessonov, line 49: Why can't we have regular constructors? This construction is barely ever used in Ignite code, and you invoke virtual method in it. Why?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/ScanVersionChainByTimestamp.java
```diff
@@ -0,0 +1,124 @@
...
+        if (lookingForVersion) {
+            Timestamp rowVersionTs = Timestamps.readTimestamp(pageAddr, payload.offset() + RowVersion.TIMESTAMP_OFFSET);
+
+            if (rowTimestampMatches(rowVersionTs, timestamp)) {
+                return readFullyOrStartReadingFragmented(link, pageAddr, payload);
+            } else {
+                return advanceToNextVersion(pageAddr, payload, partitionIdFromLink(link));
+            }
+        } else {
+            // we are continuing reading a fragmented row
```
ibessonov, line 64: I can't link you to the source, but according to our codestyle, comments start with capital letter and end with the dot.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/Timestamps.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Writes a {@link Timestamp} to memory starting at the given address + offset.
+     *
+     * @param addr      memory address
+     * @param offset    offset added to the address
+     * @param timestamp the timestamp to write
+     * @return number of bytes written
+     */
+    public static int writeTimestamp(long addr, int offset, @Nullable Timestamp timestamp) {
```
ibessonov, line 58: Ok, some methods are public, some are not. Why?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/TransactionIds.java
```diff
@@ -0,0 +1,57 @@
...
+            txIdLow = VersionChain.NULL_UUID_COMPONENT;
+        }
+
+        putLong(addr, offset, txIdHigh);
+        putLong(addr, offset + Long.BYTES, txIdLow);
+
+        return 2 * Long.BYTES;
+    }
+
+    private TransactionIds() {
```
ibessonov, line 54: Why is this necessary?

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/datapage/DataPageReader.java
```diff
@@ -54,129 +49,49 @@ public DataPageReader(PageMemory pageMemory, int groupId, IoStatisticsHolder sta
     }
 
     /**
-     * Returns a row by link. To get the row bytes, more than one pages may be traversed (if the corresponding row
-     * was fragmented when stored).
+     * Traverses page memory starting at the given link. At each step, reads the current data row and feeds it to the given
```
ibessonov, line 52: I would avoid "data row" mentioning in this context. I tried to abstract from pages content.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/datapage/DataPageReader.java
```diff
@@ -54,129 +49,49 @@ public DataPageReader(PageMemory pageMemory, int groupId, IoStatisticsHolder sta
...
+     * {@link PageMemoryTraversal} object which updates itself (usually) and returns next link to continue traversal
+     * (or {@link PageMemoryTraversal#STOP_TRAVERSAL} to stop.
      *
      * @param link Row link
-     * @return row object assembled from the row bytes
+     * @param traversal object consuming payloads and controlling the traversal
```
ibessonov, line 57: Params descriptions usually start with a capital letters. Please try to adapt your code style to what you see in the project, not what technically counts as appropriate according to TC. We should all try to write code in similar manner, it's very important

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/datapage/DataPageReader.java
```diff
@@ -54,129 +49,49 @@ public DataPageReader(PageMemory pageMemory, int groupId, IoStatisticsHolder sta
     }
 
     /**
-     * Returns a row by link. To get the row bytes, more than one pages may be traversed (if the corresponding row
-     * was fragmented when stored).
+     * Traverses page memory starting at the given link. At each step, reads the current data row and feeds it to the given
+     * {@link PageMemoryTraversal} object which updates itself (usually) and returns next link to continue traversal
+     * (or {@link PageMemoryTraversal#STOP_TRAVERSAL} to stop.
```
ibessonov, line 54: You forgot a closing parenthesis.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/datapage/DataPageReader.java
```diff
@@ -54,129 +49,49 @@ public DataPageReader(PageMemory pageMemory, int groupId, IoStatisticsHolder sta
...
-                    byte[] bytes = new byte[data.payloadSize()];
-                    dataBuf.get(bytes);
-                    try {
-                        baos.write(bytes);
-                    } catch (IOException e) {
-                        throw new IllegalStateException("Should not happen", e);
-                    }
-
-                    nextLink = data.nextLink();
+                    currentLink = traversal.consumePagePayload(currentLink, pageAddr, data, argument);
```
ibessonov, line 86: Ok, it took me a while to understand this.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/Timestamps.java
rpuch: Methods used from other packages are public, others are package local.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/TransactionIds.java
rpuch: To avoid instantiation

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/RowVersion.java
rpuch: Because, when we create a new `RowVersion`, there is no yet link for it. That's why we need to provide the partition ID explicitly.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/ReadLatestRowVersion.java
rpuch: It gets skipped in the method you point to

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/ScanVersionChainByTimestamp.java
rpuch: Removed the calls, they are not needed now

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/ScanVersionChainByTimestamp.java
rpuch: Fixed

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/datapage/DataPageReader.java
rpuch: How would you replace it?

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/datapage/DataPageReader.java
rpuch: Fixed

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PageMemoryMvPartitionStorage.java
rpuch: Ok, I think I see what you mean. Applied the change.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/datapage/DataPageReader.java
rpuch: Any actionable suggestions? :)

ibessonov: I mean, this code is valid. The fact that I would do it differently doesn't imply that there's something wrong with the code. Although I do believe that it will be pretty tough to debug chain traversal. Let's hope that we won't have to do this

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/TransactionIds.java
ibessonov: I know, but in general I don't understand why it's necessary. It's not like somebody would want to instantiate this class.

rpuch: It's a widely-used practice, for example, omittance of such a denying constructor causes a warning in SonarCube with default Java analysis rules. It's not a super-important thing, but it is just one more hint for a developer to not try adding instance methods to this class.

ibessonov: Honestly, I never understood this. This is one of those practices to everyone does, while at the same time there are absolutely no practical benefits from it.

PR #816, opened 2022-05-19
IGNITE-16943: Implement Micronaut REST

PakhomovAlexander:
- Introduce micronaut-based rest server
- Generate openapi.yaml file from server implementation
- Add rest-client module that produces auto-generated rest client
- Use micronaut-serde library in order to avoid jackson dependency


parent/pom.xml
```diff
@@ -53,19 +53,21 @@
...
         <netty.version>4.1.70.Final</netty.version>
         <javapoet.version>1.13.0</javapoet.version>
         <javax.annotation.api.version>1.3.2</javax.annotation.api.version>
         <jetbrains.annotations.version>20.1.0</jetbrains.annotations.version>
         <jmh.framework.version>1.35</jmh.framework.version>
         <junit.version>5.8.1</junit.version>
-        <micronaut.version>2.1.2</micronaut.version>
-        <micronaut.test.junit5.version>2.3.1</micronaut.test.junit5.version>
+        <micronaut.version>3.4.1</micronaut.version>
+        <micronaut.test.junit5.version>3.0.5</micronaut.test.junit5.version>
```
sashapolo, line 65: I think we can get rid of this property if we use a bom

parent/pom.xml
```diff
@@ -954,6 +963,7 @@
                             <ignorePackages>
                                 <package>org.apache.ignite.internal.sql.engine</package>
                                 <package>org.apache.ignite.internal.generated</package>
+                                <package>org.apache.ignite.rest.client</package>
```
sashapolo, line 966: Do I understand correctly that this package contains clients generated by Swagger? If yes, please add a comment, if no - why did you add it here?

modules/rest-client/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+    <version>1</version>
+    <relativePath>../../parent/pom.xml</relativePath>
+  </parent>
+
+  <artifactId>ignite-rest-client</artifactId>
+  <version>3.0.0-SNAPSHOT</version>
+
+  <dependencies>
+
+    <!--client specific dependencies, versions cannot be moved to the parent pom-->
```
sashapolo, line 37: Is this module (and this file) auto-generated?

modules/cluster-management/src/integrationTest/java/org/apache/ignite/internal/cluster/management/ItClusterManagerTest.java
```diff
@@ -107,13 +107,13 @@ void testInit(TestInfo testInfo) throws Exception {
 
         initCluster(metaStorageNodes, cmgNodes);
 
-        assertThat(cluster.get(0).clusterManager().metaStorageNodes(), will(containsInAnyOrder(metaStorageNodes)));
-        assertThat(cluster.get(1).clusterManager().metaStorageNodes(), will(containsInAnyOrder(metaStorageNodes)));
+        assertThat(cluster.get(0).getClusterManager().metaStorageNodes(), will(containsInAnyOrder(metaStorageNodes)));
+        assertThat(cluster.get(1).getClusterManager().metaStorageNodes(), will(containsInAnyOrder(metaStorageNodes)));
```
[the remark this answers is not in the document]
PakhomovAlexander: Thanks for the note, I think this was a mistake. I'll revert this back.

modules/cli/src/main/java/org/apache/ignite/cli/builtins/config/ConfigurationClient.java
```diff
@@ -81,7 +81,7 @@ public String get(
     ) {
         var req = HttpRequest
                 .newBuilder()
-                .header("Content-Type", "application/json");
+                .header("Content-Type", "text/plain");
```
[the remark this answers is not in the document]
PakhomovAlexander: I've changed the Content-Type from applicaiton/json to text/plain because the HOCON configuration, in fact, is represented as a text (string) but not as a json object.

modules/configuration/src/main/java/org/apache/ignite/internal/rest/configuration/ClusterConfigurationController.java
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * Updates cluster configuration in HOCON format.
+     *
+     * @param updatedConfiguration the cluster configuration to update.
+     */
+    @Operation(operationId = "updateClusterConfiguration")
+    @ApiResponse(responseCode = "200", description = "Configuration updated")
+    @Consumes(MediaType.TEXT_PLAIN)
+    @Patch
+    public void updateConfiguration(@Body String updatedConfiguration) throws Throwable {
```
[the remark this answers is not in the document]
PakhomovAlexander: Good point, will try.

modules/rest-api/src/main/java/org/apache/ignite/internal/rest/api/RestFactory.java
```diff
@@ -15,8 +15,10 @@
...
 
+package org.apache.ignite.internal.rest.api;
+
 /**
- * This package contains classes that represent inbound requests, outbound responses and main REST handler.
+ * Factory that produces all beans that is necessary for the controller class.
  */
-
-package org.apache.ignite.internal.rest.netty;
+public interface RestFactory {
```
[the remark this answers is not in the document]
PakhomovAlexander: This interface represents the factory that is needed for a controller, that is defined in a module. RestComponent takes the list of this factories and puts to the micronaut context.

modules/rest/pom.xml
```diff
@@ -45,52 +45,41 @@
...
         </dependency>
 
         <dependency>
             <groupId>org.apache.ignite</groupId>
-            <artifactId>ignite-network</artifactId>
+            <artifactId>ignite-api</artifactId>
         </dependency>
 
         <!-- 3rd party dependencies -->
+        <!-- micronaut dependencies, use dependencyManagement when ignite-cli is migrated to micronaut 3 -->
```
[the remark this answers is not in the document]
PakhomovAlexander: I am going to apply this change in the current pull request. In case I face some problem I'll create one.

parent/pom.xml
PakhomovAlexander: Yes, you are right. I'll add a comment.

modules/cli/src/main/java/org/apache/ignite/cli/builtins/config/ConfigurationClient.java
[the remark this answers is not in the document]
sashapolo: Makes sense, can you please leave a comment?

modules/rest-client/pom.xml
PakhomovAlexander: Actually, this is a valuable note. 

This module (just module, not code) and pom are not generated. The plugin section in this pom configures the source code generation. This is done for integration tests for the generated client in the runner module.  

From one side, the client should be generated where it is really needed. In our case, it is inside the cli module. But the Open API spec is generated by the rest module and I think that there should be an integration test for the client that is generated from this Open API spec. And these checks have to be done by the server.

I can add a comment to the pom or drop this module and create a ticket for integration tests that will generate the client somewhere inside and not include it as a module.

modules/cluster-management/pom.xml
```diff
@@ -49,17 +49,26 @@
...
         <dependency>
             <groupId>org.apache.ignite</groupId>
-            <artifactId>ignite-vault</artifactId>
+            <artifactId>ignite-rest-api</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>io.micronaut</groupId>
+            <artifactId>micronaut-http</artifactId>
+            <version>${micronaut.version}</version>
```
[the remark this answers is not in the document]
PakhomovAlexander: As for micronaut-bom, it might not pass vulnerabilities checks https://mvnrepository.com/artifact/io.micronaut/micronaut-bom/3.5.0. 

One more thing is that it declares a huge set of dependencies that are not only micronaut-related. That could be a problem in dependency management.
https://repo1.maven.org/maven2/io/micronaut/micronaut-bom/3.5.0/micronaut-bom-3.5.0.pom

modules/rest-client/pom.xml
sashapolo: > drop this module and create a ticket for integration tests that will generate the client somewhere inside and not include it as a module.

That would be very much preferable, thank you

modules/configuration/src/main/java/org/apache/ignite/internal/rest/configuration/ClusterConfigurationController.java
```diff
@@ -0,0 +1,94 @@
...
+     * Returns cluster configuration in HOCON format.
+     *
+     * @return the whole cluster configuration in HOCON format.
+     */
+    @Operation(operationId = "getClusterConfiguration")
+    @ApiResponse(
+            responseCode = "200",
+            content = @Content(mediaType = MediaType.TEXT_PLAIN, schema = @Schema(type = "string")),
+            description = "Get cluster configuration")
+    @Produces(MediaType.TEXT_PLAIN)
```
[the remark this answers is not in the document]
PakhomovAlexander: I'll add a comment.

modules/cluster-management/pom.xml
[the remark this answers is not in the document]
sashapolo: > As for micronaut-bom, it might not pass vulnerabilities checks

Why? There should be no difference between using a bom and declaring dependencies directly

> That could be a problem in dependency management.

Can you provide an example please?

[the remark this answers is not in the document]
PakhomovAlexander: > Why? There should be no difference between using a bom and declaring dependencies directly

Yes, you are right, there is no difference.

> Can you provide an example please?

Micronaut-bom includes a lot of other libraries' version definitions, for example, neo4j.version, h2.version, and many more. So, Ignite 3 will have them included to dependency management too.  Maybe that is not a problem but if Iginite 3 would like to support neo4j with a version that is different from micronaut-bom then we might face an issue.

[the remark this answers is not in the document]
sashapolo: AFAIK, if you declare a dependency directly in the dependency management section, it will override any artifacts declared in imported boms

modules/configuration/src/main/java/org/apache/ignite/internal/rest/configuration/ClusterConfigurationController.java
```diff
@@ -0,0 +1,95 @@
...
+    @Get
+    public String getConfiguration() {
```
sashapolo, line 62: missing `@Override` here and on other methods

modules/configuration/src/test/java/org/apache/ignite/internal/rest/configuration/ConfigurationControllerBaseTest.java
```diff
@@ -0,0 +1,147 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/**
+ * The base test for configuration controllers.
+ */
+@MicronautTest
+public abstract class ConfigurationControllerBaseTest {
+    @Inject
+    EmbeddedServer server;
```
sashapolo, line 48: Same question: can these fields be `private`?

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
```diff
@@ -17,151 +17,147 @@
...
+import org.apache.ignite.internal.rest.api.RestFactory;
+import org.apache.ignite.internal.rest.configuration.ClusterConfigurationController;
+import org.apache.ignite.internal.rest.configuration.NodeConfigurationController;
 import org.apache.ignite.lang.IgniteInternalException;
 import org.apache.ignite.lang.IgniteLogger;
-import org.apache.ignite.network.NettyBootstrapFactory;
 
 /**
- * Rest component is responsible for starting REST endpoints.
+ * Rest module is responsible for starting a REST endpoints for accessing and managing configuration.
```
sashapolo, line 43:
```suggestion
 * Rest module is responsible for starting REST endpoints for accessing and managing configuration.
```

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
```diff
@@ -17,151 +17,147 @@
...
-
-    private final SimpleRouter router = new SimpleRouter();
-
-    /** Netty channel. */
-    private volatile Channel channel;
+    /** Factories that produce beans needed for REST controllers. */
+    private final List<RestFactory> restFactories;
+    private final RestConfiguration restConfiguration;
+    /** Server host. */
+    private final String host = "localhost";
```
sashapolo, line 61: Why is this not a static constant?

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
```diff
@@ -17,151 +17,147 @@
...
-                ch = bindRes.channel();
+            try {
+                context = buildMicronautContext(portCandidate).start();
                 port = portCandidate;
+                log.info("REST protocol started successfully");
                 break;
-            } else if (!(bindRes.cause() instanceof BindException)) {
-                throw new RuntimeException(bindRes.cause());
+            } catch (ApplicationStartupException e) {
+                log.error("Got exception " + e.getCause() + " during node start on port " + portCandidate + " , trying again");
```
sashapolo, line 90: That's a bit strange. Are we sure that if we get this exception then the only reason for it is that the target port is busy?

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
```diff
@@ -17,151 +17,147 @@
...
-
-    /** Netty channel. */
-    private volatile Channel channel;
+    /** Factories that produce beans needed for REST controllers. */
+    private final List<RestFactory> restFactories;
+    private final RestConfiguration restConfiguration;
+    /** Server host. */
+    private final String host = "localhost";
+    /** Micronaut application context. */
+    private ApplicationContext context;
```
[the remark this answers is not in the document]
sashapolo: Actually, `port` can be left as-is if you carefully guard it with `context` accesses

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/rest/exception/handler/IgniteInternalExceptionHandler.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+@Produces
+@Singleton
+@Requires(classes = {IgniteInternalException.class, ExceptionHandler.class})
+public class IgniteInternalExceptionHandler implements ExceptionHandler<IgniteInternalException, HttpResponse<ErrorResult>> {
+
+    @Override
+    public HttpResponse<ErrorResult> handle(HttpRequest request, IgniteInternalException exception) {
+        final ErrorResult errorResult = new ErrorResult("ALREADY_INITIALIZED", exception.getMessage());
+        return HttpResponseFactory.INSTANCE.status(HttpStatus.CONFLICT).body(errorResult);
```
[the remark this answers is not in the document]
PakhomovAlexander: Since I've named it INITIALIZATION_EXCEPTION now this question is out of date.

modules/rest/pom.xml
```diff
@@ -151,7 +128,21 @@
...
+                            <artifactId>micronaut-inject-java</artifactId>
+                            <version>${micronaut.version}</version>
+                        </path>
+                        <path>
+                            <groupId>io.micronaut.openapi</groupId>
+                            <artifactId>micronaut-openapi</artifactId>
+                            <version>${maven.micronaut.openapi.plugin.version}</version>
+                        </path>
                     </annotationProcessorPaths>
+                    <fork>true</fork>
```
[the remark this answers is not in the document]
PakhomovAlexander: Runnin the compiler in separate thread but I'll remove it in order not to confuse engineers.

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
PakhomovAlexander: Yes, this is strange. I'll rewrite it.

modules/configuration/src/main/java/org/apache/ignite/internal/rest/configuration/ClusterConfigurationController.java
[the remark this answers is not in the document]
rpuch: `application/hocon`?

modules/configuration/src/main/java/org/apache/ignite/internal/rest/configuration/exception/handler/ConfigPathUnrecognizedExceptionHandler.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+@Produces(MediaType.APPLICATION_JSON)
+@Singleton
+@Requires(classes = {ConfigPathUnrecognizedException.class, ExceptionHandler.class})
+public class ConfigPathUnrecognizedExceptionHandler implements
+        ExceptionHandler<ConfigPathUnrecognizedException, HttpResponse<ErrorResult>> {
+
+    @Override
+    public HttpResponse<ErrorResult> handle(HttpRequest request, ConfigPathUnrecognizedException exception) {
+        final ErrorResult errorResult = new ErrorResult("CONFIG_PATH_UNRECOGNIZED", exception.getMessage());
```
rpuch: `final` seems redundant here as the method is just 2 lines

modules/configuration/src/test/java/org/apache/ignite/internal/rest/configuration/ClusterConfigurationControllerTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Inject
+    @Client("/management/v1/configuration/cluster/")
+    HttpClient client;
+
+    @Override
+    HttpClient client() {
+        return client;
+    }
+
```
rpuch, line 45: An excessive blank line?

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
```diff
@@ -17,151 +17,147 @@
...
+    private void setFactories(Micronaut micronaut) {
+        for (var factory : restFactories) {
+            micronaut.singletons(factory);
         }
+    }
 
-        return ch.closeFuture();
+    private int mapServerStartupException(ServerStartupException exception) {
+        if (exception.getCause() instanceof BindException) {
+            return -1;
```
rpuch, line 136: What do these 1 and -1 mean in this case?

modules/cluster-management/src/integrationTest/java/org/apache/ignite/internal/cluster/management/rest/ItClusterManagementControllerTest.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    @Factory
+    @Bean
+    @Replaces(ClusterManagementRestFactory.class)
+    public ClusterManagementRestFactory clusterManagementRestFactory() {
+        return new ClusterManagementRestFactory(clusterService);
+    }
+
+    @NotNull
```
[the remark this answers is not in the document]
PakhomovAlexander: > It's not a JSON, but it's not just any plain text. Could we use a specific content type here? For example, here https://www.w3schools.io/file/hocon-introduction/ they suggest application/hocon; it does not seem to be registered with IANA, but it looks logical.

Yes, you are right, that sounds logical. But client generators get confused when you set content type that is not registered with IANA. I tried this.

modules/cli/src/integrationTest/java/org/apache/ignite/cli/ItConfigCommandTest.java
```diff
@@ -113,9 +109,7 @@ public void setAndGetWithManualHost() {
 
         assertEquals(0, exitCode);
 
-        DocumentContext document = JsonPath.parse(removeTrailingQuotes(unescapeQuotes(out.toString(UTF_8))));
-
-        assertEquals(1, document.read("$.network.shutdownQuietPeriod", Integer.class));
+        assertTrue(out.toString(UTF_8).contains("\"shutdownQuietPeriod\" : 1"));
```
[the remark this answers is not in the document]
PakhomovAlexander: > Is this caused by changing the content type from application/json to text/plain?  
Yes

modules/cli/src/integrationTest/java/org/apache/ignite/cli/ItConfigCommandTest.java
```diff
@@ -169,11 +166,9 @@ public void partialGet() {
 
         assertEquals(0, exitCode);
 
-        JSONObject outResult = (JSONObject) JSONValue.parse(removeTrailingQuotes(unescapeQuotes(out.toString(UTF_8))));
-
-        assertTrue(outResult.containsKey("inbound"));
+        assertTrue(out.toString(UTF_8).contains("\"inbound\""));
```
sashapolo, line 169: It's a bit strange that you decided to use `assertThat` above, but not here)

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/rest/exception/handler/IgniteInternalExceptionHandler.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+@Produces
+@Singleton
+@Requires(classes = {IgniteInternalException.class, ExceptionHandler.class})
+public class IgniteInternalExceptionHandler implements ExceptionHandler<IgniteInternalException, HttpResponse<ErrorResult>> {
+
+    @Override
+    public HttpResponse<ErrorResult> handle(HttpRequest request, IgniteInternalException exception) {
+        ErrorResult errorResult = new ErrorResult("INITIALIZATION_EXCEPTION", exception.getMessage());
+        return HttpResponse.badRequest().body(errorResult);
```
sashapolo, line 40: Are you sure it should be a "bad request` errror?

modules/configuration/src/main/java/org/apache/ignite/internal/rest/configuration/exception/handler/IgniteExceptionHandler.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+@Produces
+@Singleton
+@Requires(classes = {IgniteException.class, ExceptionHandler.class})
+public class IgniteExceptionHandler implements ExceptionHandler<IgniteException, HttpResponse<ErrorResult>> {
+
+    @Override
+    public HttpResponse<ErrorResult> handle(HttpRequest request, IgniteException exception) {
+        ErrorResult errorResult = new ErrorResult("APPLICATION_EXCEPTION", exception.getMessage());
+        return HttpResponse.badRequest().body(errorResult);
```
sashapolo, line 41: Same here, I don't think it should be a "bad request"

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
```diff
@@ -17,151 +17,161 @@
...
-                .childHandler(new RestApiInitializer(router));
-
         for (int portCandidate = desiredPort; portCandidate <= desiredPort + portRange; portCandidate++) {
-            ChannelFuture bindRes = bootstrap.bind(portCandidate).awaitUninterruptibly();
-
-            if (bindRes.isSuccess()) {
-                ch = bindRes.channel();
+            try {
+                context = buildMicronautContext(portCandidate).start();
                 port = portCandidate;
```
sashapolo, line 90: unfortunately this is not correct. If you want to avoid making `port` variable `volatile` and guard it with `context` accesses, you must always write to the `port` variable **before** writing to the `context` variable. Same applies to reading, but in reverse order

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
```diff
@@ -17,151 +17,161 @@
...
+                log.info("REST protocol started successfully");
+                return;
+            } catch (ApplicationStartupException e) {
+                if (isBindExceptionInside(e)) {
```
sashapolo, line 94: I would suggest to introduce a `findBindException` method that will return an exception or `null`. `e.getCause().getCause().getCause()` looks really weird

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
```diff
@@ -17,151 +17,161 @@
...
+                    log.error("Got exception " + e.getCause().getCause().getCause() + " during node start on port "
+                            + portCandidate + " , trying again");
+                    continue;
+                }
                 break;
```
sashapolo, line 99: Looks like in case of `e` not being a `BindException` you will lose the exception and will not log it. I think it makes sense to simply re-throw it

parent/pom.xml
```diff
@@ -473,6 +475,55 @@
                 <version>${micronaut.version}</version>
             </dependency>
 
+            <dependency>
+                <groupId>io.micronaut</groupId>
```
sashapolo, line 479: So, what about using a `bom`?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/rest/exception/handler/IgniteInternalExceptionHandler.java
PakhomovAlexander: Not it is not obvious what exactly happend on IgniteInternalException. `IgniteInternalException` looks like 500 and I think it is better to mark it as an internal server error.

PakhomovAlexander: In future we have to introduce specific exceptions for situations like: already initialized, initializing, etc.

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
PakhomovAlexander: Thanks, you are absolutelly right.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/rest/exception/handler/IgniteInternalExceptionHandler.java
sashapolo: > IgniteInternalException looks like 500 and I think it is better to mark it as an internal server error.

Exactly

parent/pom.xml
PakhomovAlexander: micronaut-bom downgrades <junit5.version>5.7.2</junit5.version> , but ignite 3 uses 5.8.1 version. It actually breaks the compilation. Are we ok to downgrade the junit version?

modules/cluster-management/pom.xml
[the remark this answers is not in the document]
PakhomovAlexander: What about two boms have different versions?

parent/pom.xml
sashapolo: I don't understand how this can happen. Shouldn't our explicit dependency override what is stated in the bom?

PakhomovAlexander: junit-bom does not override this

sashapolo (approved):

ptupitsyn (approved):

PR #819, opened 2022-05-23
IGNITE-16856 Sql. Ability to create table without specifying PK

korlov42:


tledkov-gridgain (approved):
OK with me, please take a look at the my comment of the Jira issue.

PR #828, opened 2022-05-26
IGNITE-17038 Improve SQL API

ptupitsyn:
* Merge `AsyncSession` and `ReactiveSession` into `Session`: consistent with other APIs where sync and async methods reside in a single interface.
* Add missing async/reactive methods (`executeScript`) and missing parameters (`arguments`).


modules/api/src/main/java/org/apache/ignite/sql/Session.java
```diff
@@ -121,11 +217,18 @@ int[] executeBatch(
...
      */
     @Override
     void close();
 
+    /**
+     * Invalidates session, cleans up remote session resources, and stops all queries that are running within the current session.
+     *
+     * @return Operation future.
+     */
+    CompletableFuture<Void> closeAsync();
```
AMashenkov, line 230:
```suggestion
    CompletableFuture<Void> closeAsync();
    
     /**
     * Invalidates session, cleans up remote session resources, and stops all queries that are running within the current session.
     *
     * @return Publisher.
     */
    Flow.Publisher<Void> closeReactive();
```

AMashenkov (approved):

tledkov-gridgain (approved):
Ok with me.

PR #832, opened 2022-05-30
IGNITE-16962: SQL API: Implement query metadata

AMashenkov:


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlAsynchronousApiTest.java
```diff
@@ -192,6 +197,40 @@ public void select() throws ExecutionException, InterruptedException {
...
+
+        // Validata columns metadata.
+        ResultSetMetadata meta = rs.metadata();
+
+        assertNotNull(meta);
+        assertEquals(-1, meta.indexOf("COL"));
+        assertEquals(0, meta.indexOf("COL1"));
+        assertEquals(1, meta.indexOf("COL0"));
+
+        //TODO: Fix nullability issue
```
ygerzhedovich, line 218: What does it mean?

modules/api/src/main/java/org/apache/ignite/sql/ColumnMetadata.java
```diff
@@ -40,16 +42,71 @@ public interface ColumnMetadata {
...
      *
      * @return Value type.
      */
-    Object type();
+    SqlColumnType type();
+
+    /**
+     * Returns SQL column precision.
+     *
+     * @return Value precision.
```
korlov42, line 53: it is worth mentioning in the javadoc the value that will be returned for precision/scale in case those characteristics are not applicable

modules/api/src/main/java/org/apache/ignite/sql/async/AsyncResultSet.java
```diff
@@ -50,13 +49,12 @@
...
 public interface AsyncResultSet {
     /**
-     * Returns metadata for the results if the result contains rows ({@link #hasRowSet()} returns {@code true}), or {@code null} if
-     * inapplicable.
+     * Returns metadata for the results.
      *
      * @return ResultSet metadata.
      * @see ResultSet#metadata()
      */
-    @Nullable ResultSetMetadata metadata();
```
korlov42, line 59: I believe, metadata should remain nullable for the public result set, and here is why: in the internal API any execution produces a result represented by a list of lists. This is true for DML and DDL statements as well. In order to be able to consume the result, there is a metadata which describes the type of the rows returned. On the other hand, in public API, results of DDL and DML statements are unwrapped to a particular fields in the ResultSet (affectedRows, wasApplied), hence metadata is only needed to process a result set returned by a query

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/DdlPlan.java
```diff
@@ -45,7 +45,7 @@ public Type type() {
     /** {@inheritDoc} */
     @Override
     public ResultSetMetadata metadata() {
-        return List::of;
+        return ResultSetMetadataImpl.NO_METADATA;
```
korlov42, line 48: let's introduce a constant for DDL meta with a single column of type Bool and name "Applied"

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/FragmentPlan.java
```diff
@@ -53,7 +53,7 @@ public Type type() {
     /** {@inheritDoc} */
     @Override
     public ResultSetMetadata metadata() {
-        return List::of;
+        return ResultSetMetadataImpl.NO_METADATA;
```
korlov42, line 56: I wonder if anyone asking fragment about its meta. Probably it would be better to throw IllegaleState here, WDYT?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/ResultSetMetadataImpl.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param columns Columns metadata.
+     */
+    public ResultSetMetadataImpl(List<ColumnMetadata> columns) {
+        this.columns = Collections.unmodifiableList(columns);
+
+        columnsIndices = new HashMap<>(columns.size());
```
korlov42, line 51: well, default load factor is .75, so you will get one more allocation here.

BTW, should we consider to switch to Object2IntMap ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/ResultSetMetadataImpl.java
```diff
@@ -0,0 +1,79 @@
...
+    public List<ColumnMetadata> columns() {
+        return columns;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public int indexOf(String columnName) {
+        Integer idx = columnsIndices.get(columnName);
+
+        return idx == null ? -1 : idx;
```
korlov42, line 71: why don't you use `columnsIndices.getOrDefault(columnName, -1)` here?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/DdlPlan.java
AMashenkov: Got it, thanks.

AMashenkov: > let's introduce a constant for DDL meta.
Seems, we can do the same with DML, isn't it?

korlov42: Yes, indeed

korlov42: Well, as I could see, it's already derived from a query plan... So, it's up to you to change this to predefined meta or leave it as is

korlov42 (approved):

PR #837, opened 2022-05-30
IGNITE-14972 Java thin: Implement SQL API

ptupitsyn:


modules/client/src/main/java/org/apache/ignite/internal/client/sql/ClientAsyncResultSet.java
```diff
@@ -0,0 +1,196 @@
...
+    public CompletionStage<? extends AsyncResultSet> fetchNextPage() {
+        requireResultSet();
+
+        if (closed) {
+            return CompletableFuture.failedFuture(new IgniteClientException("Cursor is closed."));
+        }
+
+        if (!hasMorePages || resourceId == null) {
+            return CompletableFuture.failedFuture(new IgniteClientException("No more pages."));
+        }
```
isapego, line 140: The check does not match the one in `hasMorePages()`, which may result in user calling `hasMorePages()`, getting `true`, trying fetch next page and getting an exception as a result (if `resourceId == null`). I know that maybe this is not possible scenario, but anyway, lets use `hasMorePages()` call here, and move `resourceId == null` check to this method.

isapego (changes requested):

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/sql/ClientSqlCursorNextPageRequest.java
```diff
@@ -0,0 +1,69 @@
...
+                .thenCompose(r -> {
+                    packCurrentPage(out, r);
+                    out.packBoolean(r.hasMorePages());
+
+                    if (!r.hasMorePages()) {
+                        try {
+                            resources.remove(resourceId);
+                        } catch (IgniteInternalCheckedException e) {
+                            throw new IgniteException(e);
+                        }
```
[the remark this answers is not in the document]
ptupitsyn: Good point. Actually, this checked exception only occurs when resource is already removed, or the registry is closed. There is no leak in any case. We can simply ignore the exception.

modules/client/src/main/java/org/apache/ignite/internal/client/sql/ClientAsyncResultSet.java
ptupitsyn: Agree, fixed.

isapego (approved):

PR #842, opened 2022-06-01
Ignite 17062

denis-chudov:


modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
```diff
@@ -481,8 +481,7 @@ private static class Node {
                     metaStorageManager,
                     schemaManager);
 
-            //TODO: Get rid of it after IGNITE-17062.
-            sqlQueryProcessor = new SqlQueryProcessor(registry, clusterService, tableManager, dataStorageMgr, Map::of);
+            sqlQueryProcessor = new SqlQueryProcessor(registry, clusterService, tableManager, schemaManager, dataStorageMgr, Map::of);
```
vldpyatkov, line 484: Why did you not remove sqlQueryProcessor?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/SqlSchemaManagerImpl.java
```diff
@@ -79,11 +85,15 @@ public SqlSchemaManagerImpl(
         });
 
         schemasVv.whenComplete((token, stringIgniteSchemaMap, throwable) -> {
-            rebuild(token, stringIgniteSchemaMap);
+            if (throwable != null) {
+                throw new IgniteInternalException("Couldn't evaluate sql schemas for causality token: " + token, throwable);
```
sanpwc, line 89: Why we need this? Who and how will handle such exception?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -226,10 +232,20 @@ public TableManager(
...
 
         tablesByIdVv = new VersionedValue<>(null, HashMap::new);
 
+        registry.accept(token -> {
+            List<CompletableFuture<?>> futures = new ArrayList<>(beforeTablesVvComplete);
+
+            beforeTablesVvComplete.clear();
+
+            return CompletableFuture.allOf(futures.toArray(new CompletableFuture[] {}))
+                    .thenRun(() -> tablesByIdVv.complete(token));
```
sanpwc, line 246: Well, not sure, but I'd also add orTimeout(some rather big amount of time) terminator for diagnostic purposes that will fail an exception explicitly with some extra output about futures and context. I believe, that we should create corresponding ticket for removing such diagnostic timeouts and left such ticket within todo here. It's up to you though.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/SqlSchemaManagerImpl.java
denis-chudov: moved this exception into `calciteSchemaVv.completeExceptionally`

PR #848, opened 2022-06-03
IGNITE-16971: New CLI MVP

PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-16971


modules/cli/src/integrationTest/java/org/apache/ignite/cli/commands/status/ItStatusReplCommandTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests for {@link StatusReplCommand}.
+ */
+class ItStatusReplCommandTest extends CliCommandTestIntegrationBase {
+
+    @Test
+    @DisplayName("Should print status when valid cluster url is given")
+    @Disabled // TODO: https://issues.apache.org/jira/browse/IGNITE-17091
```
kgusakov, line 34: `@Disabled` has the special form with the reason of disable `@Disabled(https://issues.apache.org/jira/browse/IGNITE-17091)`. It will be better, than sepparate comment, I think. Use it here and in the other disables from PR, pls, if you don't mind.

modules/cli/src/main/java/org/apache/ignite/cli/commands/sql/SqlMetaData.java
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Constants for SQL auto-completion.
+ */
+
+public class SqlMetaData {
+
+    /**
+     * SQL reserved keywords.
+     */
+    public static String[] KEYWORDS = {
```
kgusakov, line 29: Do we have any plans to share the real actual list of keywords from the ignite sql module? It looks like we will have many custom keywords about data layouts and etc.

modules/cli/src/main/java/org/apache/ignite/cli/core/exception/WrappedException.java
```diff
@@ -0,0 +1,33 @@
...
+public class WrappedException extends RuntimeException {
+    /**
+     * Constructor.
+     *
+     * @param cause cause exception.
+     */
+    public WrappedException(Throwable cause) {
+        super(cause);
+    }
+}
```
kgusakov, line 33: No newline - pls check it everywhere

kgusakov (approved):

modules/cli/src/integrationTest/java/org/apache/ignite/cli/IntegrationTestBase.java
```diff
@@ -0,0 +1,336 @@
...
+/**
+ * Integration test base. Setups ignite cluster per test class and provides useful fixtures and assertions.
+ */
+@ExtendWith(WorkDirectoryExtension.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@MicronautTest
+public class IntegrationTestBase extends BaseIgniteAbstractTest {
+    public static final int DEFAULT_NODES_COUNT = 3;
+    /** Correct ignite cluster url. */
+    protected static final String NODE_URL = "http://localhost:10300";
```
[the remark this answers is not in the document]
valepakh: @kgusakov It seems that this document is outdated, according to [IGNITE-15371](https://issues.apache.org/jira/browse/IGNITE-15371) Ignite 3 uses [Google codestyle ](https://google.github.io/styleguide/javaguide.html#s4.6-whitespace) which states that the blank line between two consecutive fields is optional.

PR #851, opened 2022-06-03
IGNITE-17013 added changePeersAsync retry logic for exceptional cases

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-17013


modules/core/src/main/java/org/apache/ignite/internal/util/IgniteSpinBusyLock.java
```diff
@@ -26,7 +26,7 @@
  *
  * <p>For example, there may be a manager that have different threads for some purposes and the manager must not be stopped while at least a
  * single thread is in "busy" state. In this situation each thread must enter to "busy" state calling method {@link #enterBusy()} in
- * critical pieces of code which, i.e. use grid kernal functionality, notifying that the manager and the whole grid kernal cannot be stopped
+ * critical pieces of code, notifying that the manager and the whole ignite node cannot be stopped
```
sanpwc, line 29: Why? Seems that IgniteSpingBusyLock is a general-purpose lock, so it's not only about node stopping.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/RaftGroupServiceImpl.java
```diff
@@ -645,7 +656,7 @@ else if (th instanceof SMFullThrowable)
      * @param t The throwable.
      * @return {@code True} if this is a recoverable exception.
      */
-    private boolean recoverable(Throwable t) {
+    public static boolean recoverable(Throwable t) {
```
sanpwc, line 659: According to our new approach of splitting sendWithRetry inner retries and outer rebalance related it seems incorrect to expose raft inner retry criterion.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1362,6 +1364,49 @@ public void onError(@NotNull Throwable e) {
...
 
+    /**
+     * Performs {@link RaftGroupService#changePeersAsync(java.util.List, long)} on a provided raft group service.
+     * Retry mechanism is applied to repeat {@link RaftGroupService#changePeersAsync(java.util.List, long)} if previous one
+     * failed with some exception.
+     *
+     * @param raftGroupServiceSupplier Raft groups service.
+     * @return Function which performs {@link RaftGroupService#changePeersAsync(java.util.List, long)}.
+     */
+    public BiFunction<List<Peer>, Long, CompletableFuture<Void>> changePeersAsync(Supplier<RaftGroupService> raftGroupServiceSupplier) {
```
sanpwc, line 1375: private?

alievmirza: made package-private, it's needed to make it possible to call it from `org.apache.ignite.internal.table.distributed.TableManagerTest`

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/RaftGroupServiceImpl.java
alievmirza: I've extracted the same method for the TableManager.

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteSpinBusyLock.java
alievmirza: I've just removed words about `grid kernal` because this is a legacy from Ignite 2) 

Talking about node stopping, this section is just an example of usage `IgniteSpinBusyLock`, so I think it's useful.

sanpwc: ok

PR #853, opened 2022-06-03
IGNITE-17100 Make JoinReady command idempotent

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-17100

### DIFF
--- a/modules/cluster-management/src/integrationTest/java/org/apache/ignite/internal/cluster/management/raft/ItCmgRaftServiceTest.java
+++ b/modules/cluster-management/src/integrationTest/java/org/apache/ignite/internal/cluster/management/raft/ItCmgRaftServiceTest.java
@@ -37,6 +37,8 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import org.apache.ignite.internal.cluster.management.ClusterState;
 import org.apache.ignite.internal.cluster.management.ClusterTag;
+import org.apache.ignite.internal.cluster.management.raft.commands.JoinReadyCommand;
+import org.apache.ignite.internal.cluster.management.raft.commands.JoinRequestCommand;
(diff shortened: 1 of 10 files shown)


modules/cluster-management/src/integrationTest/java/org/apache/ignite/internal/cluster/management/raft/ItCmgRaftServiceTest.java
```diff
@@ -471,4 +468,31 @@ void testClusterStateValidation() {
...
+
+        assertThat(service.startJoinCluster(state.clusterTag()), willCompleteSuccessfully());
+
+        assertThat(service.startJoinCluster(state.clusterTag()), willCompleteSuccessfully());
+
+        assertThat(service.completeJoinCluster(), willCompleteSuccessfully());
+
+        assertThat(service.completeJoinCluster(), willCompleteSuccessfully());
+
+        assertThat(service.completeJoinCluster(), willCompleteSuccessfully());
```
rpuch, line 496: Why is completion tried 3 times, while start is only tried 2 times?

modules/cluster-management/src/integrationTest/java/org/apache/ignite/internal/cluster/management/raft/ItCmgRaftServiceTest.java
```diff
@@ -471,4 +468,31 @@ void testClusterStateValidation() {
...
+    }
```
rpuch, line 497: Does it make sense to also check how `startJoinCluster()` works AFTER `completeJoinCluster()`?

sashapolo: I don't know, it shouldn't happen in a normal case scenario...

rpuch (approved):

PR #856, opened 2022-06-06
IGNITE-17098 - Added info on pluggable storage

IgGusev:


docs/_docs/sql-reference/ddl.adoc
```diff
@@ -33,10 +33,12 @@ Parameters:
 * `tableColumn` - name and type of a column to be created in the new table.
 * `DEFAULT` - specifies a default value for the column. Only constant values are accepted.
 * `IF NOT EXISTS` - create the table only if a table with the same name does not exist.
```
ygerzhedovich, line 35:
```suggestion
* `IF NOT EXISTS` - create the table only if a table with the same name does not exist.
* `COLOCATED BY` - XXXXXXXXXXXXXX.
```

ygerzhedovich: ok

ygerzhedovich (approved):

PR #865, opened 2022-06-08
IGNITE-17057 Implement synchronous SQL API for java client

ptupitsyn:
* Move `ResultSetImpl` to `ignite-api` module and reuse to implement sync API on the client.
* Refactor `ClientSqlExecuteRequest` to use sessions because statements are not fully supported on the server side.
* Use the same tests for server and client SQL APIs to ensure consistent behavior: `ItSqlClientSynchronousApiTest`, `ItSqlClientAsynchronousApiTest`.


modules/client/src/main/java/org/apache/ignite/internal/client/sql/ClientAsyncResultSet.java
```diff
@@ -145,6 +145,11 @@ public CompletionStage<? extends AsyncResultSet> fetchNextPage() {
                     readRows(r.in());
                     hasMorePages = r.in().unpackBoolean();
 
+                    if (!hasMorePages) {
+                        // When last page is fetched, server closes the cursor.
+                        closed = true;
+                    }
+
```
isapego, line 152: Good catch

isapego (approved):

modules/client/src/main/java/org/apache/ignite/internal/client/sql/ClientSession.java
```diff
@@ -82,15 +83,17 @@ public ClientSession(
     /** {@inheritDoc} */
     @Override
     public ResultSet execute(@Nullable Transaction transaction, String query, @Nullable Object... arguments) {
-        // TODO IGNITE-17057.
-        throw new UnsupportedOperationException("Not implemented yet.");
+        Objects.requireNonNull(query);
+
+        return new ResultSetImpl(sync(executeAsync(transaction, query, arguments)));
```
tledkov-gridgain, line 88: Different methods are used to await CompletableFuture at three places: 
- sync (here)
- await (embedded SessionImpl)
- join (ResultSetImpl)

What do you think about unification?

tledkov-gridgain: What do you think about default implementation at the Session interface?

tledkov-gridgain: I guess `CompletionException` is not expected by user at the synchronous API.
Could we make temporary sync/await method with convert to SqlException?

ptupitsyn: Agree, added a temporary fix.

PR #873, opened 2022-06-09
IGNITE-17144 Fix parsing of alpha versions of Ignite

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-17144

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/properties/IgniteProductVersion.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/properties/IgniteProductVersion.java
@@ -22,13 +22,14 @@ import java.util.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.apache.ignite.internal.util.StringUtils;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Class representing an Ignite version.
  */
 public class IgniteProductVersion implements Serializable {
     private static final Pattern VERSION_PATTERN =
-            Pattern.compile("(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<maintenance>\\d+)(?<snapshot>-SNAPSHOT)?");
+            Pattern.compile("(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<maintenance>\\d+)((?<snapshot>-SNAPSHOT)|-(?<alpha>alpha\\d+))?");
 
     /**
      * Version of the current node.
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/ignite/internal/properties/IgniteProductVersion.java
```diff
@@ -96,6 +102,10 @@ public boolean snapshot() {
         return isSnapshot;
     }
 
+    public String alphaVersion() {
```
tkalkirill, line 105: Missing javadoc

modules/core/src/main/java/org/apache/ignite/internal/properties/IgniteProductVersion.java
```diff
@@ -47,11 +48,15 @@ public class IgniteProductVersion implements Serializable {
     /** Flag indicating if this is a snapshot release. */
     private final boolean isSnapshot;
 
-    private IgniteProductVersion(byte major, byte minor, byte maintenance, boolean isSnapshot) {
+    /** Alpha version part or an empty string if this is not an alpha release. */
+    private final String alphaVersion;
```
rpuch, line 52: What if at some point we switch to beta versions? Should we generalize over 'alpha|beta|rc|...'?

tkalkirill, line 52: Why not byte?

sashapolo: Because I don't know if we actually need to change it a little bit in order to accommodate beta version, for example. We can switch to a byte later if needed.

tkalkirill: Add https://issues.apache.org/jira/browse/IGNITE-17146

tkalkirill: Ok fix it in https://issues.apache.org/jira/browse/IGNITE-17146

tkalkirill (approved):

SammyVimes:
The patch looks good to me, merged to:

`main` branch – 30a2d9fffe91b60fd8a73f82c9dbfa6c04c33989
`ignite-3.0.0-alpha5` branch – ab60e76755558a70efddf6f88742e2b309429b8e
Thank you for the contribution

PR #874, opened 2022-06-10
session management draft

korlov42:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionBuilderImpl.java
```diff
@@ -103,20 +106,32 @@ public SessionBuilder defaultPageSize(int pageSize) {
     /** {@inheritDoc} */
     @Override
     public SessionBuilder property(String name, @Nullable Object value) {
-        props.put(name, value);
+        var prop = QueryProperty.byName(name);
```
AMashenkov, line 109: What will be if server supports a property, but client doesn't?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/session/Session.java
```diff
@@ -0,0 +1,168 @@
...
+
+    /** Checks whether the given session has expired or not. */
+    public boolean expired() {
+        var last = lastTouched.get();
+
+        if (last == EXPIRED) {
+            return true;
+        }
+
+        return currentTimeProvider.now() - last > idleTimeoutMs && lastTouched.compareAndSet(last, EXPIRED);
```
AMashenkov, line 99: May CAS happened concurrently?
If so, then current thread may return false from this method, however `lastTouched` is already equals EXPIRED.
```
        return currentTimeProvider.now() - last > idleTimeoutMs && 
        (lastTouched.compareAndSet(last, EXPIRED) || lastTouched.get() == EXPIRED);
```
or even just
```
        if (currentTimeProvider.now() - last > idleTimeoutMs)
            lastTouched.set(EXPIRED);
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/property/PropertiesHelperTest.java
```diff
@@ -0,0 +1,61 @@
...
+            for (var prop : omittedProps) {
+                assertNull(propMap.get(prop.name));
+            }
+        }
+
+        {
+            assertNull(propMap.get(NON_STATIC_PROP_NAME));
+        }
+    }
+}
```
AMashenkov, line 61:
```suggestion
}

```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
```diff
@@ -364,6 +318,12 @@ public static <T> T await(CompletionStage<T> stage) {
         }
     }
 
+    private void closeInternal() {
+        if (closed.compareAndSet(false, true)) {
+            busyLock.block();
```
AMashenkov, line 323: Second thread will not wait for `busyLock.block()` and proceeds with `qryProc.closeSession(sessionId)`.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionBuilderImpl.java
korlov42: This is a sessionBuilder for embedded session, so the set of properties will always be the same.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/session/Session.java
korlov42: Good catch! Fixed

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
korlov42: It's ok. A busyLock is used to prevent the session being used after cancellation. Subsequent call to qryProc.cancel() will not have any effect according to the javadoc

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryProcessor.java
```diff
@@ -35,7 +55,8 @@ public interface QueryProcessor extends IgniteComponent {
      * @return List of sql cursors.
      *
      * @throws IgniteException in case of an error.
-     * */
+     */
```
ygerzhedovich, line 58: Let's rewrite tests to remove deprecated methods

korlov42: Deprecated methods can't be removed right now because the JDBC event handler still relies on them.  I would suggest to return to this when we will be implementing a script support

just removed the deprecation for now

AMashenkov (approved):

PR #876, opened 2022-06-14
IGNITE-17093 Map error codes for cli commands

valepakh:
https://issues.apache.org/jira/browse/IGNITE-17093


valepakh:
This PR also fixes some of the problems in the https://issues.apache.org/jira/browse/IGNITE-17109, see first comment.

modules/cli/src/integrationTest/java/org/apache/ignite/cli/deprecated/ItConfigCommandTest.java
```diff
@@ -123,10 +123,10 @@ public void setWithWrongData() {
                 "network.foo=\"bar\""
         );
 
-        //assertEquals(1, exitCode); // TODO
+        assertEquals(1, exitCode);
         assertThat(
                 err.toString(UTF_8),
-                both(startsWith("Command node config update failed with reason: Got error while updating the node configuration."))
+                both(startsWith("An error occurred when calling \"update node configuration\""))
```
PakhomovAlexander, line 129: 'update node configuration' is not the name of existing command

modules/cli/src/integrationTest/java/org/apache/ignite/cli/deprecated/ItConfigCommandTest.java
```diff
@@ -141,10 +141,10 @@ public void setWithWrongData() {
                 "network.shutdownQuietPeriod=asd"
         );
 
-        //assertEquals(1, exitCode); // TODO
+        assertEquals(1, exitCode);
         assertThat(
                 err.toString(UTF_8),
-                both(containsString("Command node config update failed with reason: Got error while updating the node configuration."))
+                both(containsString("An error occurred when calling \"update node configuration\""))
```
PakhomovAlexander, line 147: and here

modules/cli/src/test/java/org/apache/ignite/cli/commands/cliconfig/CliConfigSubCommandTest.java
```diff
@@ -37,11 +36,12 @@ void noKey() {
         // When executed without arguments
         execute();
 
-        // Then
```
PakhomovAlexander, line 40: If you remove '// Than', so remove '// When' as well

modules/cli/src/test/java/org/apache/ignite/cli/commands/sql/SqlCommandTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+class SqlCommandTest extends CliCommandTestBase {
+
+    @Override
+    protected Class<?> getCommandClass() {
+        return SqlCommand.class;
+    }
+
+    @Test
+    @DisplayName("Should throw error if executed without execute or script-file options")
```
PakhomovAlexander, line 34: Should throw error if executed without --execute or --script-file

modules/cli/src/test/java/org/apache/ignite/cli/commands/CliCommandsErrorTest.java
```diff
@@ -0,0 +1,152 @@
...
+        sout = new StringWriter();
+        serr = new StringWriter();
+        cmd.setOut(new PrintWriter(sout));
+        cmd.setErr(new PrintWriter(serr));
+    }
+
+    private void execute(Class<?> cmdClass, String urlOptionName, String urlOptionValue, List<String> additionalOptions) {
+        setUp(cmdClass);
+        List<String> options = new ArrayList<>();
+        options.add("--" + urlOptionName + "=" + urlOptionValue);
```
PakhomovAlexander, line 67: I believe explicit option passing is obvious in the context of testing CLI. Could we just use "--node-url=something"?

modules/cli/src/test/java/org/apache/ignite/cli/commands/CliCommandsErrorTest.java
```diff
@@ -0,0 +1,152 @@
...
+        options.addAll(additionalOptions);
+        exitCode = cmd.execute(options.toArray(new String[0]));
+    }
+
+    static List<Arguments> cmdClassAndOptionsProvider() {
+        return List.of(
+                Arguments.arguments(NodeConfigShowSubCommand.class, "node-url", List.of()),
```
PakhomovAlexander, line 74: connect, status, cluster init also have --cluster-url or --node-url options. Shall we add tests for them as well?

modules/cli/src/test/java/org/apache/ignite/cli/commands/CliCommandsErrorTest.java
valepakh: There are different option names and different option values which are passed from the parameters for tests from the cmdClassAndOptionsProvider method.

modules/cli/src/test/java/org/apache/ignite/cli/commands/CliCommandsErrorTest.java
```diff
@@ -0,0 +1,152 @@
...
+                Arguments.arguments(NodeConfigUpdateSubCommand.class, "node-url", List.of("config")),
+                Arguments.arguments(ClusterConfigShowSubCommand.class, "cluster-url", List.of()),
+                Arguments.arguments(ClusterConfigUpdateSubCommand.class, "cluster-url", List.of("config"))
```
PakhomovAlexander, line 77: Could you use more real world value like '{key: value}'. Than it is obvious what is beeng tested here.

modules/cli/src/main/java/org/apache/ignite/cli/deprecated/spec/ClusterReplCommandSpec.java
```diff
@@ -84,7 +84,7 @@ public static class InitClusterCommandSpec extends BaseCommand {
 
         /** {@inheritDoc} */
         @Override
-        public void run() {
+        public Integer call() {
```
PakhomovAlexander, line 87: Does exit code make any sense in REPL? If no then I don't think we even have to return it.

modules/cli/src/main/java/org/apache/ignite/cli/core/exception/IgniteCliApiException.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param url endpoint URL.
+     */
+    public IgniteCliApiException(Throwable cause, String call, String url) {
+        super(cause);
+        this.call = call;
+        this.url = url;
+    }
+
+    /**
+     * Gets the REST API call.
```
PakhomovAlexander, line 42: Call name or what?

modules/cli/src/main/java/org/apache/ignite/cli/core/exception/handler/UnknownCommandExceptionHandler.java
```diff
@@ -28,8 +28,9 @@
 public class UnknownCommandExceptionHandler implements ExceptionHandler<UnknownCommandException> {
 
     @Override
-    public void handle(ExceptionWriter err, UnknownCommandException e) {
+    public int handle(ExceptionWriter err, UnknownCommandException e) {
         err.write(e.getMessage());
+        return 1;
```
PakhomovAlexander, line 33: Are we shure that it is not 2?

modules/cli/src/main/java/org/apache/ignite/cli/core/call/CallExecutionPipeline.java
```diff
@@ -90,24 +90,25 @@ public static <I extends CallInput, T> CallExecutionPipelineBuilder<I, T> builde
...
 
         if (callOutput.hasError()) {
-            exceptionHandlers.handleException(ExceptionWriter.fromPrintWriter(errOutput), callOutput.errorCause());
-            return;
+            return exceptionHandlers.handleException(ExceptionWriter.fromPrintWriter(errOutput), callOutput.errorCause());
         }
 
-        if (callOutput.isEmpty()) {
-            return;
+        if (!callOutput.isEmpty()) {
```
PakhomovAlexander, line 107: Why did you decided to revert the expresion? Now it is less readable but I can't get the reason.

modules/cli/src/main/java/org/apache/ignite/cli/deprecated/spec/ClusterReplCommandSpec.java
valepakh: No it doesn't. I'll split the commands so the REPL commands will stay Runnables.

modules/cli/src/main/java/org/apache/ignite/cli/core/exception/IgniteCliApiException.java
valepakh: It was intented to be a call name, but probably it would be better to pass a real command name used.

modules/cli/src/main/java/org/apache/ignite/cli/core/exception/handler/UnknownCommandExceptionHandler.java
valepakh: This is only for REPL mode so this it irrelevant, I'll change to 2 and add a comment.

modules/cli/src/main/java/org/apache/ignite/cli/core/call/CallExecutionPipeline.java
valepakh: Because otherwize two return statements will be needed and this would be less readable than negative condition IMO.

modules/cli/src/test/java/org/apache/ignite/cli/commands/CliCommandsErrorTest.java
```diff
@@ -0,0 +1,152 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import picocli.CommandLine;
+
+/**
+ * Base class for testing error handling in CLI commands that use REST.
+ */
+@MicronautTest
+public class CliCommandsErrorTest {
```
PakhomovAlexander, line 45: I would name this test 'UrlOptionsNegativeTest'. Then it is clear what is the purpuse of this test and when a developer has to add more cases here.

modules/cli/src/test/java/org/apache/ignite/cli/commands/CliCommandsErrorTest.java
```diff
@@ -0,0 +1,152 @@
...
+import org.apache.ignite.cli.commands.configuration.node.NodeConfigShowSubCommand;
+import org.apache.ignite.cli.commands.configuration.node.NodeConfigUpdateSubCommand;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import picocli.CommandLine;
+
+/**
+ * Base class for testing error handling in CLI commands that use REST.
```
PakhomovAlexander, line 42: It is not a base class.

modules/cli/src/main/java/org/apache/ignite/cli/commands/configuration/node/NodeConfigShowReplSubCommand.java
```diff
@@ -59,42 +56,23 @@ public class NodeConfigShowReplSubCommand extends BaseCommand {
...
                 .inputProvider(input::build)
                 .output(spec.commandLine().getOut())
                 .errOutput(spec.commandLine().getErr())
                 .decorator(new JsonDecorator())
-                .exceptionHandler(new ShowConfigReplExceptionHandler())
                 .build()
                 .runPipeline();
     }
-
-    private static class ShowConfigReplExceptionHandler implements ExceptionHandler<ApiException> {
```
PakhomovAlexander, line 83: Whey did you remove this handler?

modules/cli/src/main/java/org/apache/ignite/cli/call/status/StatusReplCall.java
```diff
@@ -40,9 +39,7 @@
 public class StatusReplCall implements Call<EmptyCallInput, Status> {
 
     private final NodeManager nodeManager;
-
     private final CliPathsConfigLoader cliPathsCfgLdr;
```
PakhomovAlexander, line 42: I don't think that removing the empty lines is a good idea.

modules/cli/src/main/java/org/apache/ignite/cli/call/status/StatusCall.java
```diff
@@ -55,32 +53,27 @@ public StatusCall(NodeManager nodeManager, CliPathsConfigLoader cliPathsCfgLdr)
...
+            return DefaultCallOutput.success(
+                    Status.builder()
+                            .connected(connected != null)
+                            .connectedNodeUrl(input.getClusterUrl())
+                            .initialized(connected != null && canReadClusterConfig(input))
+                            .nodeCount(nodeManager.getRunningNodes(paths.logDir, paths.cliPidsDir()).size())
+                            .build()
+            );
+        } catch (ApiException | IllegalArgumentException e) {
+            return DefaultCallOutput.failure(new IgniteCliApiException(e, "get node configuration", input.getClusterUrl()));
```
PakhomovAlexander, line 67: 'get node configuration' -> 'status'

modules/cli/src/main/java/org/apache/ignite/cli/commands/configuration/node/NodeConfigShowReplSubCommand.java
valepakh: Running node returns its config regardless of whether the cluster is initialized or not.

modules/cli/src/main/java/org/apache/ignite/cli/call/status/StatusReplCall.java
valepakh: Code style states that it is optional and I think it's better to use libe breaks between groups of fields, rather than between each field.

modules/cli/src/test/java/org/apache/ignite/cli/commands/CliCommandsErrorTest.java
PakhomovAlexander: What stops you from passing '--cluster-url' instead of 'cluster-url' in cmdClassAndOptionsProvider?

modules/cli/src/test/java/org/apache/ignite/cli/commands/CliCommandsErrorTest.java
PakhomovAlexander: Agreed, disabled test with the link to jira would be a good solution.

PakhomovAlexander: > Connect and status commands currently rely on the local installation so they should be tested in the integration tests. ,

I can pass `--cluster-url` option to them, right? Why is local installation required?

modules/cli/src/main/java/org/apache/ignite/cli/call/status/StatusReplCall.java
PakhomovAlexander: The whole Ignite 3 source code uses line breaks. I believe it is better not to break the convention even if something looks more readable.

modules/cli/src/test/java/org/apache/ignite/cli/commands/CliCommandsErrorTest.java
valepakh: Ah, sorry, connect command doesn't use the --cluster-url option so some changes in the tests are required to test it.

valepakh: As to why the local installation is required for the status command - because the REST endpoint is not implemented yet?

modules/cli/src/main/java/org/apache/ignite/cli/Main.java
```diff
@@ -54,11 +54,12 @@ public class Main {
     public static void main(String[] args) {
         initJavaLoggerProps();
 
+        int exitCode = 0;
```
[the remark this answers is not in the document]
valepakh: Throwing exceptions from main method results in non-zero exit code.
Although I couldn't find where it's specified, the comment in the launcher code indicates that
> The launcher's exit code (in the absence of calls to System.exit) will be non-zero if main threw an exception.

modules/cli/src/test/java/org/apache/ignite/cli/commands/UrlOptionsNegativeTest.java
```diff
@@ -64,28 +69,45 @@ private void setUp(Class<?> cmdClass) {
...
+        //                Arguments.arguments(ClusterCommandSpec.InitClusterCommandSpec.class, "---cluster-url=",
+        //                        List.of("--cluster-name=cluster", "--meta-storage-node=test"))
         );
     }
 
     @ParameterizedTest
     @MethodSource("cmdClassAndOptionsProvider")
     @DisplayName("Should display error when wrong port is given")
-    public void incorrectPortTest(Class<?> cmdClass, String urlOptionName, List<String> additionalOptions) {
+    void incorrectPort(Class<?> cmdClass, String urlOptionName, List<String> additionalOptions, boolean isReplCommand) {
```
PakhomovAlexander, line 103: Just split this test into two repli and non-repl. Please, don't use flags like `isReplCommand`

PakhomovAlexander (approved):
LGTM

kgusakov:
Shouldn't TODOs with the 17093 issue number be removed (or maybe recreated with another number) or fixed in this PR?

modules/cli/src/main/java/org/apache/ignite/cli/core/exception/handler/IgniteCliApiExceptionHandler.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Exception handler for {@link IgniteCliApiException}.
+ */
+public class IgniteCliApiExceptionHandler implements ExceptionHandler<IgniteCliApiException> {
+    private static final IgniteLogger log = IgniteLogger.forClass(IgniteCliApiExceptionHandler.class);
+
+    @Override
+    public int handle(ExceptionWriter err, IgniteCliApiException e) {
+        String message;
```
kgusakov, line 36:
```suggestion
        String message;
```
```suggestion
        String message;

```

modules/cli/src/main/java/org/apache/ignite/cli/core/exception/handler/IgniteCliApiExceptionHandler.java
```diff
@@ -0,0 +1,61 @@
...
+                message = "Could not connect to URL: " + e.getUrl();
+            } else if (apiCause != null) {
+                message = apiCause.getMessage();
+            } else {
+                message = "An error occurred, error code: " + cause.getCode() + ", response: " + cause.getResponseBody();
+            }
+        } else {
+            message = e.getCause() != e ? e.getCause().getMessage() : e.getMessage();
+        }
+        log.error(message, e);
```
kgusakov, line 52:
```suggestion
        log.error(message, e);
```
```suggestion
        log.error(message, e);

```

modules/cli/src/main/java/org/apache/ignite/cli/core/exception/handler/TimeoutExceptionHandler.java
```diff
@@ -29,9 +29,10 @@ public class TimeoutExceptionHandler implements ExceptionHandler<TimeoutExceptio
     private static final IgniteLogger log = IgniteLogger.forClass(TimeoutExceptionHandler.class);
 
     @Override
-    public void handle(ExceptionWriter err, TimeoutException e) {
+    public int handle(ExceptionWriter err, TimeoutException e) {
         log.error("Timeout exception ", e);
         err.write("Command failed with timeout.");
+        return 1;
```
kgusakov, line 35: Please, add empty lines according to style guide.

modules/cli/src/main/java/org/apache/ignite/cli/deprecated/spec/BootstrapIgniteCommandSpec.java
```diff
@@ -44,7 +45,8 @@ public class BootstrapIgniteCommandSpec extends BaseCommand implements IgniteCom
 
     /** {@inheritDoc} */
     @Override
-    public void run() {
+    public Integer call() {
         cmd.init(urls, spec.commandLine().getOut(), spec.commandLine().getColorScheme());
+        return 0;
```
kgusakov, line 50: Empty line before, pls.

modules/cli/src/test/java/org/apache/ignite/cli/commands/CliCommandTestBase.java
```diff
@@ -17,39 +17,100 @@
...
  * Base class for testing CLI commands.
  */
 @MicronautTest
-public class CliCommandTestBase {
+public abstract class CliCommandTestBase {
     @Inject
     private ApplicationContext context;
 
-    private CommandLine commandLine;
+    private CommandLine cmd;
```
kgusakov, line 39: Empty lines here and below pls

modules/cli/src/test/java/org/apache/ignite/cli/commands/UrlOptionsNegativeTest.java
```diff
@@ -0,0 +1,232 @@
...
+ * Tests error handling with various invalid URLs in CLI commands that use REST.
+ */
+@MicronautTest
+public class UrlOptionsNegativeTest {
+    private static final String NODE_URL = "http://localhost:10300";
+
+    @Inject
+    private ApplicationContext context;
+
+    private CommandLine cmd;
```
kgusakov, line 56: Empty lines

kgusakov (changes requested):

PR #881, opened 2022-06-15
IGNITE-16955 Improve logging for rebalance process

kgusakov:


modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -42,9 +42,33 @@
  */
 public class RebalanceUtil {
 
+    /** Logger. */
+    private static IgniteLogger LOG = IgniteLogger.forClass(RebalanceUtil.class);
+
+    /** Return code of metastore multi-invoke which identify,
```
alievmirza, line 48: `identifies` here and below

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -366,6 +366,9 @@ private CompletableFuture<?> onUpdateReplicas(ConfigurationNotificationEvent<Int
             if (replicasCtx.oldValue() != null && replicasCtx.oldValue() > 0) {
                 TableConfiguration tblCfg = replicasCtx.config(TableConfiguration.class);
 
+                LOG.info("Received update for replicas number for table={} from replicas={} to replicas={}",
```
alievmirza, line 369: If we run sone rebalance test, we can see in the logs the following:

```
2022-06-23 13:18:27:798 +0300 [INFO][Thread-242][TableManager] Received update for replicas number for table=PUBLIC.TBL1 from replicas=1 to replicas=2
2022-06-23 13:18:27:799 +0300 [INFO][Thread-244][TableManager] Received update for replicas number for table=PUBLIC.TBL1 from replicas=1 to replicas=2
2022-06-23 13:18:27:802 +0300 [INFO][Thread-241][TableManager] Received update for replicas number for table=PUBLIC.TBL1 from replicas=1 to replicas=2
```
Lines looks the same and could be more informative, if we add info about the node where such update has happened

alievmirza: Maybe we should think about a general solution, where all lines of logs will contain info about the node which logged that line

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
```diff
@@ -140,6 +140,10 @@ public void onLeaderElected(long term) {
                     if (!pendingEntry.empty()) {
                         List<ClusterNode> pendingNodes = (List<ClusterNode>) ByteUtils.fromBytes(pendingEntry.value());
 
+                        LOG.info("New leader elected for the raft group={} "
```
alievmirza, line 143:
```suggestion
                        LOG.info("New leader was elected for the raft group={} "
```

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/RaftGroupServiceImpl.java
```diff
@@ -372,6 +372,9 @@ public static CompletableFuture<RaftGroupService> start(
 
         CompletableFuture<ChangePeersAsyncResponse> fut = new CompletableFuture<>();
 
+        LOG.info("Sending request to changePeers of group={} to peers={} with leader term={}",
```
alievmirza, line 375: changePeersAsync

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
```diff
@@ -140,6 +140,10 @@ public void onLeaderElected(long term) {
...
+                                        + "of partition={}, table={} and pending reconfiguration to peers={} discovered",
+                                partId, partNum, tblConfiguration.name().value(), pendingNodes);
+
```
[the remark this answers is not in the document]
kgusakov: Are you sure that we really need it? I mean - we already have logging inside the changePeersAsync on the server/client side and before it in listeners, also separate messages about problems and retries in movePartition.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
kgusakov: Yes, I thought about it. I think it must be the general solution, because this point connected with all current logs.

PR #890, opened 2022-06-20
IGNITE-17193: Return application/problem+json in REST

PakhomovAlexander:
Previously it was ErrorResult. Now the standard problem
is returned in all REST methods.


PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-17193

modules/rest-api/src/main/java/org/apache/ignite/internal/rest/exception/handler/IgniteExceptionHandler.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.ignite.internal.rest.constants.HttpCode;
+import org.apache.ignite.internal.rest.problem.HttpProblemResponse;
+import org.apache.ignite.lang.IgniteException;
+
+/**
+ * Handles {@link IgniteException} and represents it as an application/problem+json response.
+ */
+@Singleton
+@Requires(classes = {IgniteException.class, ExceptionHandler.class})
+public class IgniteExceptionHandler implements ExceptionHandler<IgniteException, HttpResponse<? extends Problem>> {
```
sk0x50, line 40: I think it will be needed to add a handler for `IgniteCheckedException` in nearest future.
Please take a look at https://issues.apache.org/jira/browse/IGNITE-14931
By the way, traceId will be introduced by IGNITE-14931 as well

PakhomovAlexander: Thank you for the note, I've created a ticket for that https://issues.apache.org/jira/browse/IGNITE-17281

PR #895, opened 2022-06-23
IGNITE-17162 Fix init cluster command options

valepakh:
In addition to changing --node-endpoint to the --cluster-url option in the cluster init command, this PR actually rewrites this command to use CLI execution pipeline and generated REST client. Since we can't easily mock the HttpClient for the generated client, mock-server was used to test this and other commands. mock-server has an ability to test against the generated OpenAPI spec but it was found that it's not enough, for example, it's impossible to verify the request payload in this case.


modules/cli/src/main/java/org/apache/ignite/cli/commands/cluster/ClusterInitReplSubCommand.java
```diff
@@ -0,0 +1,107 @@
...
+ * Initializes an Ignite cluster.
+ */
+@Command(name = "init", description = "Initializes an Ignite cluster.")
+public class ClusterInitReplSubCommand extends BaseCommand implements Runnable {
+
+    /**
+     * Node url option.
+     */
+    @Option(
+            names = {"--cluster-url"}, description = "Url to ignite node.",
```
sk0x50, line 43: IMHO, `cluster-url` is not obvious parameter/name. I would like to rename it to `cluster-endpoint-url` or something like that. Cluster has multiple nodes/endpoints that have its own url, there is no one dedicated URL for whole cluster. What do you think?

modules/cli/src/main/java/org/apache/ignite/cli/commands/cluster/ClusterInitReplSubCommand.java
```diff
@@ -0,0 +1,107 @@
...
+    )
+    private String clusterUrl;
+
+    /**
+     * List of names of the nodes (each represented by a separate command line argument) that will host the Metastorage Raft group. If the
+     * "--cmg-nodes" parameter is omitted, the same nodes will also host the Cluster Management Raft group.
+     */
+    @Option(names = "--meta-storage-node", required = true, description = {
+            "Name of the node (repeat like '--meta-storage-node node1 --meta-storage-node node2' to specify more than one node)",
+            "that will host the Meta Storage Raft group.",
```
sk0x50, line 54: Let's remove `raft` from the description. I would prefer `Meta Storage` ( or `Meta Storage service`) instead of `Meta Storage Raft group`. The same for the `CMG` -> `Cluster Management Service`.

valepakh: Maybe we should use the same terminology as in the [IEP-77](https://cwiki.apache.org/confluence/display/IGNITE/IEP-77%3A+Node+Join+Protocol+and+Initialization+for+Ignite+3#IEP77:NodeJoinProtocolandInitializationforIgnite3-MetaStorage)? That is, `Meta Storage` and `Cluster Management Group`.

PR #896, opened 2022-06-23
IGNITE-17214 A Hybrid Logical Clock implementation.

sergeyuttsel:


modules/core/src/test/java/org/apache/ignite/hlc/HybridClockTestUtils.java
```diff
@@ -0,0 +1,47 @@
...
+ * Utils for a Hybrid Logical Clock testing.
+ */
+public class HybridClockTestUtils {
+    /**
+     * Creates a mocked system clock.
+     *
+     * @param expected Expected value which returns by system clock.
+     * @return The mocked clock.
+     */
+    public static MockedStatic<Clock> mockToEpochMilli(long expected) {
```
ascherbakoff, line 38: Here is the citation from the internet on statick mocks:
"Generally speaking, some might say that when writing clean code, we shouldn't need to mock static classes. This could typically hint at a design issue or code smell in the application.

Why? First, a class depending on a static method has tight coupling, and second, it nearly always leads to code that is difficult to test. Ideally, a class should not be responsible for obtaining its dependencies, and if possible, they should be externally injected.

So, it's always worth investigating if we can refactor our code to make it more testable."
I suggest to take it into a consideration.

modules/core/src/main/java/org/apache/ignite/hlc/HybridTimestamp.java
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+package org.apache.ignite.hlc;
+
+import org.apache.ignite.internal.tostring.S;
+
+/**
+ * A hybrid timestamp that combines physical clock and logical clock.
+ */
+public class HybridTimestamp implements Comparable<HybridTimestamp> {
```
ascherbakoff, line 25: The implementation seems correct, but not optimal. It produces new objects and does unnecessary operations. I suggest to look at https://github.com/eventreducer/hlc/blob/master/src/main/java/org/eventreducer/hlc/HybridTimestamp.java as a good reference. It also has a nice abstaction of physical time provider.

modules/core/src/main/java/org/apache/ignite/hlc/HybridTimestamp.java
```diff
@@ -0,0 +1,111 @@
...
+        this.logical = logical;
+    }
+
+    /**
+     * Compares hybrid timestamps.
+     *
+     * @param times Times for comparing.
+     * @return The highest hybrid timestamp.
+     */
+    public static HybridTimestamp max(HybridTimestamp... times) {
```
ascherbakoff, line 49: Why is this method declared _public_ ?

modules/core/src/main/java/org/apache/ignite/hlc/HybridTimestamp.java
```diff
@@ -0,0 +1,111 @@
...
+        return logical;
+    }
+
+    /**
+     * Returns a new hybrid timestamp with incremented logical component.
+     *
+     * @return The hybrid timestamp.
+     */
+    public HybridTimestamp addTicks(int ticks) {
+        return new HybridTimestamp(physical, this.logical + ticks);
```
ascherbakoff, line 83: Creating an object on tick is not good

modules/core/src/main/java/org/apache/ignite/hlc/HybridTimestamp.java
```diff
@@ -0,0 +1,111 @@
...
+     * Compares hybrid timestamps.
+     *
+     * @param times Times for comparing.
+     * @return The highest hybrid timestamp.
+     */
+    public static HybridTimestamp max(HybridTimestamp... times) {
+        HybridTimestamp maxTime = times[0];
+
+        for (int i = 1; i < times.length; i++) {
+            maxTime = maxTime.compareTo(times[i]) > 0 ? maxTime : times[i];
```
ascherbakoff, line 53: The assignment is needed only for new max.

modules/core/src/main/java/org/apache/ignite/hlc/HybridClock.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+public class HybridClock {
+    /** Latest timestamp. */
+    private HybridTimestamp latestTime;
+
+    /**
+     * The constructor which initializes the latest time to current time by system clock.
+     */
+    public HybridClock() {
+        this.latestTime = new HybridTimestamp(Clock.systemUTC().instant().toEpochMilli(), 0);
```
ascherbakoff, line 34: Default value for logicalTime is 0, according to paper

modules/core/src/main/java/org/apache/ignite/hlc/HybridTimestamp.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+     * Returns a new hybrid timestamp with incremented logical component.
+     *
+     * @return The hybrid timestamp.
+     */
+    public HybridTimestamp addTicks(int ticks) {
+        return new HybridTimestamp(physical, this.logical + ticks);
+    }
```
ascherbakoff, line 84: What about overflow ? I suggest to make logical counter of type long (or propagate overflow to logicalTime)

modules/core/src/main/java/org/apache/ignite/hlc/HybridTimestamp.java
sergeyuttsel: Fixed

modules/core/src/main/java/org/apache/ignite/hlc/HybridClock.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * Creates a timestamp for a received event.
+     *
+     * @param requestTime Timestamp from request.
+     * @return The hybrid timestamp.
+     */
+    public synchronized HybridTimestamp tick(HybridTimestamp requestTime) {
+        HybridTimestamp now = new HybridTimestamp(Clock.systemUTC().instant().toEpochMilli(), -1);
```
sanpwc, line 61: Do we really need to create HybridTimestamp instance in order to only compare it's physical?
I believe should postpone new gerHybridTIme instantiation and do it only if now.physical is bigger than request.physical and latestTime.physical.

modules/core/src/main/java/org/apache/ignite/hlc/HybridTimestamp.java
sanpwc: Agreed, it seems to be an extremely frequent operation. I'd rather check it with JMH: whether it's better to use thread-safe mutable instances or such new-on-tick approach.

PR #898, opened 2022-06-24
IGNITE-17192 Binary Tuple Implementation 

ademakov:


modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuple.java
```diff
@@ -0,0 +1,58 @@
...
+    /** {@inheritDoc} */
+    @Override
+    public int count() {
+        return elementCount();
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public BigDecimal decimalValue(int index) {
+        return decimalValue(index, schema.element(index).decimalScale);
```
AMashenkov, line 46: Why a scale is not a part of decimal value?

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleParser.java
```diff
@@ -0,0 +1,215 @@
...
+     */
+    public interface Sink {
+        /**
+         * Provides the location of the next tuple value.
+         *
+         * @param index Value index.
+         * @param begin Start offset of the value, 0 for NULL.
+         * @param end End offset of the value, 0 for NULL.
+         */
+        void nextElement(int index, int begin, int end);
```
AMashenkov, line 39: It looks like this method is used for instance initialization.
The index param is never used. Can Sink have a package visibility?

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleParser.java
```diff
@@ -0,0 +1,215 @@
...
+        return buffer.asReadOnlyBuffer().position(sink.begin).limit(sink.end).order(ByteOrder.LITTLE_ENDIAN);
+    }
+
+    /**
+     * Locate the specified tuple element.
+     *
+     * @param index Index of the element.
+     * @param sink Receiver.
+     */
+    public void fetch(int index, Sink sink) {
```
AMashenkov, line 136: Sink param is used here to pass more than one parameter to the outsize the method.
Why can't it be a return type?

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleSchema.java
```diff
@@ -0,0 +1,247 @@
...
+     * @param colBegin First columns in the range.
+     * @param colEnd Last column in the range (exclusive).
+     * @return Tuple schema.
+     */
+    private static BinaryTupleSchema createSchema(SchemaDescriptor descriptor, int colBegin, int colEnd) {
+        int numCols = colEnd - colBegin;
+
+        Element[] elements = new Element[numCols];
+
+        for (int i = 0; i < numCols; i++) {
```
AMashenkov, line 115: Do you mean that?
```suggestion
        for (int i = colBegin; i < numCols; i++) {
```

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleSchema.java
```diff
@@ -0,0 +1,247 @@
...
+     * Create a schema for binary tuples with selected row columns.
+     *
+     * @param descriptor Row schema.
+     * @param indexes List of columns.
+     * @return Tuple schema.
+     */
+    public static BinaryTupleSchema createValueSchema(SchemaDescriptor descriptor, int[] indexes) {
+        Element[] elements = new Element[indexes.length];
+
+        for (int i : indexes) {
```
AMashenkov, line 173: Let's rewrite to indexed-for-loop

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleSchema.java
ademakov: Not quite but there was a bug indeed, thanks for pointing out.

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleParser.java
```diff
@@ -0,0 +1,215 @@
...
+            if (offset == nextOffset && hasNullMap()) {
+                int nullIndex = BinaryTupleSchema.HEADER_SIZE + i / 8;
+                byte nullMask = (byte) (1 << (i % 8));
+                if ((buffer.get(nullIndex) & nullMask) != 0) {
+                    sink.nextElement(i, 0, 0);
+                    continue;
+                }
+            }
+
+            sink.nextElement(i, offset, nextOffset);
```
AMashenkov, line 184: In the fetch method you use it to return a pair of `begin` and `end` offsets.
Here, Sink semantic is different. You use it as a Consumer.

I think Sink could be a top-level public class with the consumer semantic,
and fetch() method can be rewritten to return an object with pair of values, e.g. Element is a good name.

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuple.java
ademakov: Is is specified this way in IEP-92: https://cwiki.apache.org/confluence/display/IGNITE/IEP-92%3A+Binary+Tuple+Format#IEP92:BinaryTupleFormat-NumberandDecimalrepresentation
If you don't agree with this approach let's discuss it at some IEP-92 related place. On slack, wiki, dev-list, or whatever works best.

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuple.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * Reads value for specified element.
+     *
+     * @param index Element index.
+     * @return Element value.
+     */
+    public Object value(int index) {
```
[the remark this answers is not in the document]
ademakov: It is not used now but the whole thing is not yet used now. I implemented functionally-complete API so it could be used in different situations in the future.

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuple.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /** {@inheritDoc} */
+    @Override
+    public int count() {
+        return elementCount();
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public BigDecimal decimalValue(int index) {
```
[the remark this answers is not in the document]
ademakov: Currently such a change would conflict with the NativeTypeSpec.objectValue() mechanism. Let postpone this until this mechanism is improved.

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleBuilder.java
```diff
@@ -0,0 +1,966 @@
...
+     * Creates a builder with direct buffer.
+     *
+     * @param numElements Number of tuple elements.
+     * @param allowNulls True if NULL values are possible, false otherwise.
+     * @param totalValueSize Total estimated length of non-NULL values, -1 if not known.
+     * @return Tuple builder.
+     */
+    public BinaryTupleBuilder createWithDirectBuffer(int numElements, boolean allowNulls, int totalValueSize) {
+        var builder = new BinaryTupleBuilder(numElements, allowNulls, totalValueSize);
+        builder.allocateDirect(totalValueSize);
```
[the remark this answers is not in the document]
ademakov: Direct buffers are removed.

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleBuilder.java
```diff
@@ -0,0 +1,966 @@
...
+    }
+
+    /**
+     * Append a value for the current element.
+     *
+     * @param value Element value.
+     * @return {@code this} for chaining.
+     */
+    public BinaryTupleBuilder appendShort(short value) {
+        if (value <= Byte.MAX_VALUE && value >= Byte.MIN_VALUE) {
```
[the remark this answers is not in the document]
ademakov: Okay. I did it though for me it looks like utter crap.

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleReader.java
```diff
@@ -0,0 +1,537 @@
...
+    }
+
+    /**
+     * Reads value of specified element.
+     *
+     * @param index Element index.
+     * @return Element value.
+     */
+    public Byte byteValueBoxed(int index) {
+        var element = seek(index);
```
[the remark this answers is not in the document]
ibessonov: I see, this one got ignored

ibessonov:
@ademakov ok, I have a question - will other comments be addressed? Do you need any assistance with them?

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleSchema.java
```diff
@@ -0,0 +1,346 @@
...
+
+        final int decimalScale;
+
+        final boolean nullable;
+
+        /**
+         * Constructor.
+         *
+         * @param type Element data type.
+         * @param nullable True for nullable elements, false for non-nullable,
```
korlov42, line 75:
```suggestion
         * @param nullable True for nullable elements, false for non-nullable.
```

ademakov: Thanks, fixed.

PR #901, opened 2022-06-28
IGNITE-17230 Support splt-file page store

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-17230


modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/checkpoint/CheckpointWorkflow.java
```diff
@@ -219,15 +226,17 @@ public Checkpoint markCheckpointBegin(
...
-
             for (DataRegion<PersistentPageMemory> dataRegion : dataRegions) {
                 dataRegion.pageMemory().finishCheckpoint();
             }
         }
 
         if (chp.hasDelta()) {
             checkpointMarkersStorage.onCheckpointEnd(chp.progress.id());
+
+            chp.progress.pagesToWrite(null);
```
SammyVimes, line 237: Do I understand correctly that if `hasDelta == false`, then dirty pages will be null, right? So we don't need to set null if hasDelta == false

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/PersistentPageMemory.java
```diff
@@ -1295,6 +1295,9 @@ public class Segment extends ReentrantReadWriteLock {
         /** Padding to read from word beginning. */
         private static final int ACQUIRED_PAGES_PADDING = 4;
 
+        /** Page memory. */
+        private final PersistentPageMemory pageMemory;
```
SammyVimes, line 1299: Segment class is inner, so you can just use `PersistentPageMemory.this`

SammyVimes (changes requested):

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/PersistentPageMemory.java
tkalkirill: fix it

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/checkpoint/CheckpointWorkflow.java
tkalkirill: If **hasDelta == false** then we set **CheckpointDirtyPages#EMPTY**, i can change to **null**, WDYT?

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/checkpoint/CheckpointManager.java
```diff
@@ -221,4 +233,56 @@ static boolean safeToUpdateAllPageMemories(Collection<? extends DataRegion<Persi
...
+        CheckpointProgress lastCheckpointProgress = lastCheckpointProgress();
+
+        assert lastCheckpointProgress != null : "Checkpoint has not happened yet";
+        assert lastCheckpointProgress.inProgress() : "Checkpoint must be in progress";
+
+        CheckpointDirtyPages pagesToWrite = lastCheckpointProgress.pagesToWrite();
+
+        assert pagesToWrite != null : "Dirty pages must be sorted out";
+
+        CompletableFuture<DeltaFilePageStoreIo> deltaFilePageStoreFuture = filePageStore.getOrCreateNewDeltaFile(
```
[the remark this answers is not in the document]
tkalkirill: Are you suggesting just return **DeltaFilePageStoreIo**, **join** execute inside and not outside??

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/checkpoint/CheckpointWorkflow.java
SammyVimes: No, I was just curious, current approach is good

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/checkpoint/CheckpointManager.java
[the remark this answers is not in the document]
tkalkirill: Discussed personally, leave as is.

SammyVimes (approved):
lgtm

PR #904, opened 2022-06-28
IGNITE-14931 Added error groups

sk0x50:


modules/api/src/main/java/org/apache/ignite/lang/TableNotFoundException.java
```diff
@@ -27,6 +30,6 @@ public class TableNotFoundException extends IgniteException {
      * @param name Table name.
      */
     public TableNotFoundException(String name) {
-        super(IgniteStringFormatter.format("Table does not exist [name={}]", name));
+        super(TABLE_ERR_GROUP.name(), TABLE_ALREADY_EXISTS_ERR, "Table does not exist [name=" + name + ']');
```
ibessonov, line 33: Wrong error code

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroup.java
```diff
@@ -0,0 +1,189 @@
...
+     * Registers a new error code within this error group.
+     *
+     * @param errorCode Error code to be registered.
+     * @return Full error code which includes group code and specific error code.
+     * @throws IllegalArgumentException If the given {@code errorCode} is already registered
+     *      or {@code errorCode} is greater than 0xFFFF or less than or equal to 0.
+     */
+    public int registerErrorCode(int errorCode) {
+        if (errorCode < 0 || errorCode > 0xFFFF) {
+            throw new IllegalArgumentException("Error code should be greater than 0 and less than or equal to 0xFFFF");
```
ibessonov, line 84: Message contradicts the condition, I think it must be "<= 0" instead of strict "<"

sk0x50: Yep, you are right.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroup.java
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * Creates a new error group with the given {@code groupName} and {@code groupCode}.
+     *
+     * @param groupName Group name to be created.
+     * @param groupCode Group code to be created.
+     * @return New error group.
+     * @throws IllegalArgumentException If the specified name or group code already registered.
+     *      or {@code groupCode} is greater than 0xFFFF or less than or equal to 0.
+     */
+    public static ErrorGroup newGroup(String groupName, int groupCode) {
```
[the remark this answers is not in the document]
sk0x50: Do we really need this? In my understanding, this method should be called from static initializer block or initializing a static variable, and so there is no need to synchronize it.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroup.java
```diff
@@ -0,0 +1,189 @@
...
+        String grpName = groupName.toUpperCase(Locale.ENGLISH);
+
+        Optional<ErrorGroup> grp = registeredGroups.stream().filter(g -> g.name().equalsIgnoreCase(grpName)).findFirst();
```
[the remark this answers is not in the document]
sk0x50: yep, you are right. will fix

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -0,0 +1,47 @@
...
+        public static final ErrorGroup UNKNOWN_ERR_GROUP = ErrorGroup.newGroup("UNK", 1);
+
+        /** Unknown error. */
+        public static int UNKNOWN_ERR = UNKNOWN_ERR_GROUP.registerErrorCode(1);
+    }
+
+    /** Tables error group. */
+    public static class Table {
+        /** Table error group. */
+        public static final ErrorGroup TABLE_ERR_GROUP = ErrorGroup.newGroup("TBL", 3);
```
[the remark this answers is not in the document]
sk0x50: It was assigned to the `COMMON_ERR_GROUP`, but I decided to remove it later. will fix.

modules/client-common/src/test/java/org/apache/ignite/internal/client/proto/ClientMessageDecoderTest.java
```diff
@@ -51,8 +51,8 @@ void testInvalidMagicThrowsException() {
 
         var t = assertThrows(IgniteException.class, () -> decode(buf));
 
-        assertEquals("Invalid magic header in thin client connection. Expected 'IGNI', but was 'BEEF'.",
-                t.getMessage());
+        String expected = "Invalid magic header in thin client connection. Expected 'IGNI', but was 'BEEF'.";
+        assertTrue(t.getMessage().contains(expected), "Expected: " + expected + ", actual: " + t.getMessage());
```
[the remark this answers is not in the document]
sk0x50: for some reasons `harmcrest` library is not included to dependencies for this module. I will add it and check.

modules/api/src/main/java/org/apache/ignite/lang/ColumnAlreadyExistsException.java
```diff
@@ -27,6 +30,6 @@ public class ColumnAlreadyExistsException extends IgniteException {
      * @param name Column name.
      */
     public ColumnAlreadyExistsException(String name) {
```
ibessonov, line 32: It's weird that exceptions like these are in "lang" module. Makes no sense. We should fix it in the future.

modules/client/src/test/java/org/apache/ignite/client/ClientTupleTest.java
```diff
@@ -94,16 +96,16 @@ public void testEmptySchemaThrows() {
     @Test
     public void testSetThrowsWhenColumnIsNotPresent() {
         var ex = assertThrows(IgniteException.class, () -> getBuilder().set("x", "y"));
```
ibessonov, line 98: BTW, should we assert a more specific type here? We do have a right subclass here

modules/core/src/main/java/org/apache/ignite/lang/IgniteInternalCheckedException.java
```diff
@@ -26,29 +34,182 @@ public class IgniteInternalCheckedException extends Exception {
...
+    /** Unique identifier of this exception that should help locating the error message in a log file. */
+    private final UUID traceId;
+
+    /**
+     * Creates a new exception with the given group and error code.
+     *
+     * @param groupName Group name.
+     * @param code Full error code.
+     */
+    public IgniteInternalCheckedException(String groupName, int code) {
```
ibessonov, line 60: I believe that internal exception could have ErrorGroup as a parameter, what do you think?

modules/api/src/main/java/org/apache/ignite/lang/ColumnAlreadyExistsException.java
sk0x50: I'm inclined to agree with you.

modules/core/src/main/java/org/apache/ignite/lang/IgniteInternalCheckedException.java
sk0x50: Well, that was my first thought, however, it does not seem to be working in case of RU. Let's assume that node (that have a new error group) wants to send an error to the old node. Perhaps, we need something as follows:
```
public class IgniteInternalCheckedException extends Exception {
    String groupName;
    int code;
    public IgniteInternalCheckedException(int fullErrorCode) {
        this.groupName = ErrorGroup.groupByCode(ErrorGroup.extractGroupCode(fullErrorCode));
        this.code = fullErrorCode;
    }
    ...
}
public class ModuleSpecificCheckedException extends IgniteInternalCheckedException {
    public ModuleSpecificCheckedException(int fullErrorCode) {
        super(fullErrorCode);
        
        // check that error code relates to the specific group
        assert ErrorGroups.ModuleGroup.CUSTOM_ERR_GRP.code() == ErrorGroup.extractGroupCode(fullErrorCode); 
    }
    ...
}
```

ibessonov: All I wanted is to simplify a developers job, so that they don't have to call "name()" method every time. A set of fields will be the same, this doesn't affect RU.

ibessonov (approved):

PR #905, opened 2022-06-28
IGNITE-17091: Implement status command

PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-17091


modules/cli/src/main/java/org/apache/ignite/cli/commands/node/status/NodeStatusSubCommand.java
```diff
@@ -0,0 +1,63 @@
...
+@Singleton
+public class NodeStatusSubCommand extends BaseCommand implements Callable<Integer> {
+
+    /**
+     * Node url option.
+     */
+    @SuppressWarnings("PMD.UnusedPrivateField")
+    @Option(
+            names = {"--node-url"}, description = "Url to node.",
+            descriptionKey = "ignite.node-url", defaultValue = "http://localhost:10300"
```
Pochatkin, line 45: Default value not needed anymore after IGNITE-17130

modules/cli/src/main/java/org/apache/ignite/cli/call/cluster/status/ClusterStatusCall.java
```diff
@@ -0,0 +1,81 @@
...
+        IgnitePaths paths = cliPathsCfgLdr.loadIgnitePathsOrThrowError();
+        ClusterStatusBuilder clusterStatusBuilder = ClusterStatus.builder()
+                .nodeCount(nodeManager.getRunningNodes(paths.logDir, paths.cliPidsDir()).size());
+
+        try {
+            ClusterState clusterState = fetchClusterState(input.getClusterUrl());
+            clusterStatusBuilder
+                    .initialized(true)
+                    .name(clusterState.getClusterTag().getClusterName())
+                    .cmgNodes(clusterState.getCmgNodes());
```
Pochatkin, line 64: As I see from org.apache.ignite.internal.cluster.management.rest.ClusterManagementController.clusterState clusterState can be null. NPE possible here. But moreover I think that return null as value from REST call its bad practice. Please do not add null check here, better change sematic of org.apache.ignite.internal.cluster.management.rest.ClusterManagementController.clusterState to not return null as value.

Pochatkin (changes requested):

modules/cli/src/main/java/org/apache/ignite/cli/commands/node/status/NodeStatusSubCommand.java
PakhomovAlexander: in case IGNITE-17130 is merged, I wil rebase

modules/node-management/pom.xml
```diff
@@ -0,0 +1,156 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kgusakov, line 1: It looks like it is not 'node-management' module, but just module with rest controller and some interfaces for it. Actual node management is doing under 'runner' module. 
If we create the separate top-level module for this kind of job - I fear we will be drowned in one-controller modules and become the REST factory instead of database :)

I agree, that maybe the runner is not the best place for it, but at the moment it is the real place where node lifecycle manages.

Maybe, if we have other controllers (which can't be placed in any existent modules), we'll should return to this topic and maybe create the separate place or package inside current 'rest*' modules. WDYT?

kgusakov (changes requested):

modules/node-management/pom.xml
PakhomovAlexander: The problem here is that `rest` module should know about `NodeManagementController` in order to generate the openapi specification at compile time. `NodeManagementController` depends on `LifecycleManager` from `runner` module.  And `runner` depends on `rest` module. Here is a cyclic dependency. So, I've added `node-management` in order to move all REST-related stuff out of `runner` module and solve the cyclic dependency problem. 

Your comment makes sense and I think there is a couple of ways we could go:
- choose a better name for the new module and/or move it under /rest/ path (make it a submodule of rest)
- move all stuff from `node-management` back to `runner` and try to solve the compile-time dependency between `runner` and `rest` in a different way. 

Any thoughts?

modules/cli/src/main/java/org/apache/ignite/cli/call/cluster/status/ClusterStatusCall.java
```diff
@@ -0,0 +1,82 @@
...
+                    .metadataStorageNodes(clusterState.getMsNodes())
+                    .cmgNodes(clusterState.getCmgNodes());
+        } catch (ApiException e) {
+            if (e.getCode() == 404) { // NOT_FOUND means the cluster is not initialized yet
+                clusterStatusBuilder.initialized(false);
+            } else {
+                return DefaultCallOutput.failure(new IgniteCliApiException(e, input.getClusterUrl()));
+            }
+        } catch (IllegalArgumentException e) {
+            clusterStatusBuilder.initialized(false);
```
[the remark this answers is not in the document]
PakhomovAlexander: Good note, I suppose this was added by mistake. REST client throws `ApiExeption` and that's enough.

kgusakov (approved):

PR #908, opened 2022-06-29
IGNITE-17198 Minimal implementation of pure in-memory storage (with in-memory RAFT)

rpuch:
https://issues.apache.org/jira/browse/IGNITE-17198

### DIFF
--- a/.idea/codeStyles/codeStyleConfig.xml
+++ b/.idea/codeStyles/codeStyleConfig.xml
@@ -1,5 +1,5 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
-    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Ignite 3" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="GoogleStyle" />
   </state>
 </component>
--- a/.idea/codeStyles/codeStyleConfig.xml
+++ b/.idea/codeStyles/codeStyleConfig.xml
@@ -1,5 +1,5 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
-    <option name="PREFERRED_PROJECT_CODE_STYLE" value="GoogleStyle" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Ignite 3" />
   </state>
 </component>
--- a/modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLozaTest.java
+++ b/modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLozaTest.java
@@ -33,6 +33,7 @@ import java.nio.file.Path;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
+import org.apache.ignite.internal.raft.server.RaftGroupOptions;
 import org.apache.ignite.internal.testframework.WorkDirectory;
 import org.apache.ignite.internal.testframework.WorkDirectoryExtension;
 import org.apache.ignite.network.ClusterNode;
@@ -66,7 +67,7 @@ public class ItLozaTest {
      */
     private RaftGroupService startClient(String groupId, ClusterNode node, Loza loza) throws Exception {
         return loza.prepareRaftGroup(groupId,
-                List.of(node), () -> mock(RaftGroupListener.class)
+                List.of(node), () -> mock(RaftGroupListener.class), RaftGroupOptions.defaults()
         ).get(10, TimeUnit.SECONDS);
     }
 
--- a/modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLozaTest.java
+++ b/modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLozaTest.java
@@ -34,7 +34,6 @@ import java.nio.file.Path;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
-import org.apache.ignite.internal.raft.server.RaftGroupOptions;
 import org.apache.ignite.internal.testframework.WorkDirectory;
 import org.apache.ignite.internal.testframework.WorkDirectoryExtension;
 import org.apache.ignite.network.ClusterNode;
--- a/modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLozaTest.java
+++ b/modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLozaTest.java
@@ -17,6 +17,7 @@
 
 package org.apache.ignite.internal.raft;
 
+import static org.apache.ignite.internal.raft.server.RaftGroupOptions.defaults;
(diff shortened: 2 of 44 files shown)


modules/raft/src/main/java/org/apache/ignite/internal/raft/server/RaftGroupOptions.java
```diff
@@ -0,0 +1,52 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.raft.server;
+
+/**
+ * Options specific to a Raft group that is being started.
+ */
+public class RaftGroupOptions {
+    private final boolean volatileStores;
```
SammyVimes, line 24: I always forget: we can omit comments for private members, right?

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/VolatileJRaftServiceFactory.java
```diff
@@ -0,0 +1,63 @@
...
+
+        return new LocalSnapshotStorage(uri, raftOptions);
+    }
+
+    @Override public RaftMetaStorage createRaftMetaStorage(final String uri, final RaftOptions raftOptions) {
+        return new VolatileRaftMetaStorage();
+    }
+
+    @Override public LogEntryCodecFactory createLogEntryCodecFactory() {
+        return LogEntryV1CodecFactory.getInstance();
```
SammyVimes, line 57: Probably should be default or something. I don't see that we are ever going to use different codecs (probably just the implementation of this one)

modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/impl/VolatileRaftMetaStorage.java
```diff
@@ -0,0 +1,75 @@
...
+package org.apache.ignite.raft.jraft.storage.impl;
+
+import org.apache.ignite.raft.jraft.entity.PeerId;
+import org.apache.ignite.raft.jraft.option.RaftMetaStorageOptions;
+import org.apache.ignite.raft.jraft.storage.RaftMetaStorage;
+import org.apache.ignite.raft.jraft.storage.VolatileStorage;
+
+/**
+ * Volatile (in-memory) implementation of {@link RaftMetaStorage}. Used for Raft groups storing partition data of
+ * volatile (in-memory) storages.
```
SammyVimes, line 27: I think this `(in-memory)` can be omitted

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -500,6 +502,10 @@ private void updateAssignmentInternal(ConfigurationNotificationEvent<byte[]> ass
         CompletableFuture.allOf(futures).join();
     }
 
+    private RaftGroupOptions groupOptionsForInternalTable(InternalTable internalTbl) {
+        return new RaftGroupOptions(internalTbl.storage().isVolatile());
```
SammyVimes, line 506: Is it possible that there is going to be more options inside the RaftGroupOptions? Maybe it's not a good idea to use constructor here and you already have factory methods. Maybe you can add some new method like "RaftGroupOptions#optionsForTable(isVolatile)"? Does it make sense?

SammyVimes (changes requested):
Looks good overall, just a couple of minor things

.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -1,5 +1,5 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
-    <option name="PREFERRED_PROJECT_CODE_STYLE" value="GoogleStyle" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Ignite 3" />
```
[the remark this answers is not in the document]
rpuch: Oops, good catch!

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLozaTest.java
```diff
@@ -66,7 +67,7 @@ public class ItLozaTest {
      */
     private RaftGroupService startClient(String groupId, ClusterNode node, Loza loza) throws Exception {
         return loza.prepareRaftGroup(groupId,
-                List.of(node), () -> mock(RaftGroupListener.class)
+                List.of(node), () -> mock(RaftGroupListener.class), RaftGroupOptions.defaults()
```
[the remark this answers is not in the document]
rpuch: I tried to import it statically in tests (to make it less verbose), but I did not import it statically in production code (to make it more visible). Now, I made all the code to conform to this rule.

WDYT?

modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/JraftServerImpl.java
```diff
@@ -339,13 +351,20 @@ public synchronized boolean startRaftGroup(String groupId, @NotNull RaftGroupEve
...
+            nodeOptions.setRaftMetaUri(serverDataPath.resolve("meta").toString());
+        }
         nodeOptions.setSnapshotUri(serverDataPath.resolve("snapshot").toString());
 
         nodeOptions.setFsm(new DelegatingStateMachine(lsnr));
 
         nodeOptions.setRaftGrpEvtsLsnr(evLsnr);
 
+        if (groupOptions.volatileStores()) {
+            // override to volatile stores
```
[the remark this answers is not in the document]
rpuch: Sure, removed it

modules/raft/src/main/java/org/apache/ignite/internal/raft/server/RaftGroupOptions.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public static RaftGroupOptions forPersistentStores() {
+        return new RaftGroupOptions(false);
+    }
+
+    public static RaftGroupOptions forVolatileStores() {
+        return new RaftGroupOptions(true);
+    }
+
+    public RaftGroupOptions(boolean volatileStores) {
```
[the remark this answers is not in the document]
rpuch: Made it private

modules/raft/src/main/java/org/apache/ignite/internal/raft/server/RaftGroupOptions.java
rpuch: In Ignite3 we can, but in this case I preferred to add Javadocs

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/VolatileJRaftServiceFactory.java
rpuch: Moved the methods' implementations to `default` methods

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLozaTest.java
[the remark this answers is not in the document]
SammyVimes: 👍

PR #913, opened 2022-07-01
IGNITE-17240 Provide an ability to configure logging backend through IgniteClient.Builder

korlov42:


ptupitsyn (commented):
Client-related changes look good to me.
However, there are many changes that are not related to the client, is that intentional?

korlov42:
These changes were made because of a [test](https://github.com/apache/ignite-3/pull/913/files#diff-19d884e5e1433125614c8ade207d111107eb5445748b1cdbf59cbdd0344cfe28R44). I've added a validation that no components in modules which the client module depends on are using static logger, otherwise configuring a logger won't make any sense. This test reveals that NamedThreadFactory which is widelly used throughout a codebase does actually use a static logger, so I've forced to rework all those places.

rpuch (approved):

modules/core/src/main/java/org/apache/ignite/internal/components/LongJvmPauseDetector.java
```diff
@@ -107,9 +105,7 @@ public void run() {
                     lastWakeUpTime = System.currentTimeMillis();
                 }
 
-                if (LOG.isDebugEnabled()) {
-                    LOG.debug(getName() + " has been started.");
-                }
+                log.debug(getName() + " has been started");
```
ygerzhedovich, line 108: should we use parameter here instead of concatenation?

parent/pom.xml
```diff
@@ -771,6 +772,18 @@
                 <scope>test</scope>
             </dependency>
 
+            <dependency>
+                <groupId>com.tngtech.archunit</groupId>
+                <artifactId>archunit</artifactId>
```
ygerzhedovich, line 777: should we use test scope for the library?

korlov42: if my understanding is correct, we use parent/pom just to declare possible dependencies and theirs versions. The scope should be defined in the child module this dep will be used

ygerzhedovich: Yes, you are right. However the lib is for testing purposes, isn't it? If so, let's say test scope here, otherwise it required to duplicate it for downstream modules

ygerzhedovich (approved):

modules/client/src/main/java/org/apache/ignite/client/IgniteClient.java
```diff
@@ -88,6 +89,8 @@ class Builder {
         /** Retry policy. */
         private RetryPolicy retryPolicy = new RetryReadPolicy();
 
+        private LoggerFactory loggerFactory;
```
AMashenkov, line 92: javadoc

AMashenkov (approved):

modules/client/src/main/java/org/apache/ignite/client/IgniteClient.java
korlov42: afaik, javdocs on a private members are not mandatory in AI3

modules/client/src/main/java/org/apache/ignite/internal/client/IgniteClientConfigurationImpl.java
```diff
@@ -69,7 +72,9 @@ public IgniteClientConfigurationImpl(
             int reconnectThrottlingRetries,
             Executor asyncContinuationExecutor,
             long heartbeatInterval,
-            RetryPolicy retryPolicy) {
+            RetryPolicy retryPolicy,
+            LoggerFactory loggerFactory
```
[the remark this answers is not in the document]
korlov42: fixed

PR #918, opened 2022-07-01
IGNITE-16760

denis-chudov:


modules/metastorage-server/src/test/java/org/apache/ignite/internal/metastorage/server/AbstractKeyValueStorageTest.java
```diff
@@ -2043,6 +2043,43 @@ public void rangeCursor() {
...
+
+        storage.put(key2, val2);
+
+        assertEquals(3, storage.revision());
+        assertEquals(3, storage.updateCounter());
+
+        // Range that includes tombstones.
+        Cursor<Entry> cur = storage.range(key1, null, true);
+
+        assertEquals(2, cur.stream().count());
```
sanpwc, line 2075: I'd rather add extra check not only for size, but for actual returned data set. Otherwise given test will pass if, for example, only tombstone is returned as a result for `storage.range(key1, null, false);`

denis-chudov: done

PR #922, opened 2022-07-05
IGNITE-17241 Introduce utility class Loggers

korlov42:


modules/core/src/main/java/org/apache/ignite/internal/logger/Loggers.java
```diff
@@ -0,0 +1,70 @@
+package org.apache.ignite.internal.logger;
```
AMashenkov, line 1: License is missed

korlov42: fixed

PR #924, opened 2022-07-05
IGNITE-17303 RocksDB snapshots might include writes added after snapshot creation start

rpuch:
https://issues.apache.org/jira/browse/IGNITE-17303


modules/rocksdb-common/src/main/java/org/apache/ignite/internal/rocksdb/snapshot/RocksSnapshotManager.java
```diff
@@ -77,18 +78,22 @@ public RocksSnapshotManager(RocksDB db, Collection<ColumnFamilyRange> ranges, Ex
...
                             .map(cf -> createSstFileAsync(cf, snapshot, tmpPath))
                             .toArray(CompletableFuture[]::new);
 
-                    return CompletableFuture.allOf(sstFutures).thenApply(v -> snapshot);
+                    return CompletableFuture.allOf(sstFutures);
                 }, executor)
-                .whenCompleteAsync((snapshot, e) -> {
+                .thenCompose(Function.identity())
+                .whenCompleteAsync((ignored, e) -> {
                     if (e != null) {
```
ibessonov, line 97: Can you please also move this statement below the closing of the snapshot? Thanks!

ibessonov (approved):

modules/rocksdb-common/src/main/java/org/apache/ignite/internal/rocksdb/snapshot/RocksSnapshotManager.java
rpuch: Thanks for catching this

PR #926, opened 2022-07-05
IGNITE-17281 Added support error codes and trace ids to rest components

sk0x50:


PakhomovAlexander (approved):

PakhomovAlexander:
LGTM

modules/rest-api/src/test/java/org/apache/ignite/internal/rest/exception/handler/IgniteExceptionHandlerTest.java
```diff
@@ -64,7 +68,7 @@ static Stream<Arguments> igniteExceptions() {
                         Problem.builder()
                                 .status(500)
                                 .title("Internal Server Error")
-                                .code(UNKNOWN_ERR)
+                                .code(humanReadableCode)
                                 .detail(errorMessage + " Ooops")
```
PakhomovAlexander, line 72: As far as I can see, `errorMessage` contains `traceId` and a human-readable code.  But this information is already included in the problem json. Shall we duplicate the information in the API response? WDYT?

sk0x50: Yep, `errorMessage` contains `traceId`, human-readable code and detailed message. However, I think it would be useful to get additional field that provide an error code explicitly and so, the end-user doesn't need to parse `errorMessage`.

PR #928, opened 2022-07-05
IGNITE-14986 Re-work error handling in meta storage component in accordance with error groups

sk0x50:


modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/CursorImpl.java
```diff
@@ -159,19 +171,25 @@ public T next() {
...
+                if (cause instanceof NoSuchElementException) {
+                    throw (NoSuchElementException) cause;
                 }
 
                 LOG.debug("Unable to evaluate cursor hasNext command", e);
 
-                throw new IgniteInternalException(e);
+                if (e.getCause() instanceof IgniteInternalException) {
+                    IgniteInternalException exception = (IgniteInternalException) e.getCause();
+                    throw new IgniteInternalException(exception.traceId(), exception.code(), exception);
```
ibessonov, line 186: It's interesting, why don't we use `CURSOR_EXECUTION_ERR` here as a code? Original code will be preserved in the cause.

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/MetaStorageServiceImpl.java
```diff
@@ -498,7 +499,7 @@ private void stopWatch(IgniteUuid watchId) {
 
                                 v.cursor.close();
                             } catch (InterruptedException e) {
-                                throw new IgniteInternalException(e);
+                                throw new IgniteInternalException(WATCH_STOPPING_ERR, e);
```
ibessonov, line 502: We may create a MetaStorageException for the module, this would make code easier to read I believe

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/CursorImpl.java
```diff
@@ -159,19 +171,25 @@ public T next() {
...
+                }
+
+                if (cause instanceof NoSuchElementException) {
+                    throw (NoSuchElementException) cause;
                 }
 
                 LOG.debug("Unable to evaluate cursor hasNext command", e);
 
-                throw new IgniteInternalException(e);
+                if (e.getCause() instanceof IgniteInternalException) {
```
ibessonov, line 184: I remember you were asking about what to do in such cases. All I can recommend is to have a set of utilitary methods.
I'd go as far as writing it as `throw withCause(MetaStorageException::new, CURSOR_EXECUTION_ERR, e.getCause())`

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -36,12 +43,48 @@ public static class Table {
...
         /** Column already exists. */
-        public static int COLUMN_ALREADY_EXISTS_ERR = TABLE_ERR_GROUP.registerErrorCode(3);
+        public static final int COLUMN_ALREADY_EXISTS_ERR = TABLE_ERR_GROUP.registerErrorCode(3);
         /** Column not found. */
-        public static int COLUMN_NOT_FOUND_ERR = TABLE_ERR_GROUP.registerErrorCode(4);
+        public static final int COLUMN_NOT_FOUND_ERR = TABLE_ERR_GROUP.registerErrorCode(4);
+    }
+
+    /** Meta storage error group. */
+    public static class MetaStorage {
```
ibessonov, line 56: Why is this a part of public API?

ibessonov:
I also took some liberties while merging main branch and resolving conflicts. I hope it's not a problem

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -36,12 +43,48 @@ public static class Table {
...
+        /** Meta storage error group. */
+        public static final ErrorGroup META_STORAGE_ERR_GROUP = ErrorGroup.newGroup("MTS", 3);
```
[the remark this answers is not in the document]
sk0x50: Yes, it makes sense to me. Will change.

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/CursorImpl.java
sk0x50: Well, it makes sense to me. Will change.

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/MetaStorageServiceImpl.java
sk0x50: MetaStorageException has been added

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
sk0x50: I think that all error code should be available for the end user

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/CursorImpl.java
```diff
@@ -125,13 +129,13 @@ public boolean hasNext() {
...
             } catch (InterruptedException | ExecutionException e) {
-                if (e.getCause() != null && e.getCause().getClass().equals(NodeStoppingException.class)) {
+                if (e.getCause() instanceof NodeStoppingException) {
                     return false;
                 }
 
                 LOG.debug("Unable to evaluate cursor hasNext command", e);
 
-                throw new IgniteInternalException(e);
+                throw withCause(MetaStorageException::new, CURSOR_EXECUTION_ERR, e);
```
ibessonov, line 138: `e` can never be an Ignite*Exception. Do we determine its code by reading the message? Why don't we check the cause of ExecutionException?

sk0x50: This cause is checked inside the `withCause` implementation:
```
    Throwable unwrapped = unwrapCause(t);
```
where `unwrap` if the following:
```
    public static Throwable unwrapCause(Throwable e) {
        while ((e instanceof CompletionException || e instanceof ExecutionException) && e.getCause() != null) {
            e = e.getCause();
        }

        return e;
    }
```

modules/core/src/main/java/org/apache/ignite/internal/util/ExceptionUtils.java
```diff
@@ -324,4 +330,104 @@ public static Throwable unwrapCause(Throwable e) {
 
         return e;
     }
+
+    /**
+     * Creates a new exception, which type is defined by the provided {@code supplier}, with the specified {@code t} as a cause.
+     * In the case when the provided cause {@code t} is an instance of {@link IgniteInternalException}
+     * or {@link IgniteInternalCheckedException}, the original trace identifier and full error code are preserved.
+     * Otherwise, a newly generated trace identifier and {@link Common#UNKNOWN_ERR} are used.
```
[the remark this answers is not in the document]
sk0x50: Yes, it might be a problem. I will take a look at it. At first glance, we need a new one supplier which does not expect an error code.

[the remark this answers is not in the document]
sk0x50: Fixed as follows:
 - `withCause` accepts a new parameter - `defaultCode` which is used when the provided cause is not an instance of ignite exception
 - `withCauseAndCode` which always uses the provided error code.

ibessonov (approved):

PR #931, opened 2022-07-07
IGNITE-17291

denis-chudov:


modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/CursorImpl.java
```diff
@@ -50,19 +52,41 @@
...
+     * @param fn                    Function transforming the element of type {@link M} to the type of {@link T}.
+     */
+    public <M> CursorImpl(
+            RaftGroupService metaStorageRaftGrpSvc,
+            CompletableFuture<IgniteUuid> initOp,
+            Function<M, T> fn
+    ) {
+        this.metaStorageRaftGrpSvc = metaStorageRaftGrpSvc;
+        this.initOp = initOp;
+        this.it = new InnerIterator(e -> singleton(e), fn);
```
sanpwc, line 69: Do you really like an idea of wrapping every next result with SingletonSet?

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/CursorImpl.java
```diff
@@ -50,19 +52,41 @@
...
      * @param initOp                Future that runs meta storage service operation that provides cursor.
+     * @param internalCacheFn       Function transforming the result of {@link CursorNextCommand} to the
+     *                              {@link Iterable} of elements of type {@link M}.
+     * @param fn                    Function transforming the element of type {@link M} to the type of {@link T}.
      */
-    CursorImpl(RaftGroupService metaStorageRaftGrpSvc, CompletableFuture<IgniteUuid> initOp, Function<Object, T> fn) {
+    public <M> CursorImpl(
+            RaftGroupService metaStorageRaftGrpSvc,
+            CompletableFuture<IgniteUuid> initOp,
+            Function<Object, Iterable<M>> internalCacheFn,
```
sanpwc, line 84: I don't think that we need two explicit transformation functions. Please consider approach that is used within
`new CursorImpl<>(metaStorageRaftGrpSvc, watchRes, MetaStorageServiceImpl::watchResponse),`.

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/EntryImpl.java
```diff
@@ -48,7 +48,7 @@ public final class EntryImpl implements Entry {
      * @param rev     Revision.
      * @param updCntr Update counter.
      */
-    EntryImpl(@NotNull ByteArray key, @Nullable byte[] val, long rev, long updCntr) {
+    public EntryImpl(@NotNull ByteArray key, @Nullable byte[] val, long rev, long updCntr) {
```
sanpwc, line 51: package-private still seems fine here.

modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/MetaStorageRangeCursorTest.java
```diff
@@ -0,0 +1,182 @@
...
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+/**
+ * Test for cursor created by {@link MetaStorageService#range(ByteArray, ByteArray)} commands.
+ */
+@ExtendWith(MockitoExtension.class)
+public class MetaStorageRangeCursorTest {
```
sanpwc, line 66: I'd rather move given range test to org.apache.ignite.internal.metastorage.client.ItMetaStorageServiceTest, WDYT?

denis-chudov: `ItMetaStorageServiceTest` are more slow due to `@BeforeEach` method, this is unacceptable for the tests in `MetaStorageRangeCursorTest` that has several dozens of tests.

PR #936, opened 2022-07-11
IGNITE-16676  Parallel start of raft groups

kgusakov:


modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/JraftServerImpl.java
```diff
@@ -456,6 +485,25 @@ public void stopBlockMessages(String groupId) {
         client.stopBlock();
     }
 
+    /**
+     * Initialize (if needed) and receive per-group lock,
```
alievmirza, line 489: let's mention here that this is a raft group

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -260,6 +268,14 @@ public TableManager(
...
         rebalanceScheduler = new ScheduledThreadPoolExecutor(REBALANCE_SCHEDULER_POOL_SIZE,
                 new NamedThreadFactory("rebalance-scheduler", LOG));
+
+        ioExecutor = new ThreadPoolExecutor(
+                25,
+                Integer.MAX_VALUE,
+                100,
+                TimeUnit.MILLISECONDS,
+                new LinkedBlockingQueue<>(),
+                new NamedThreadFactory("tableManager-io", LOG));
```
alievmirza, line 278: seems like LOG should also be on a new line

kgusakov: This LOG is a part of one-line NamedThreadFactory constructor, not the previous one multiline ThreadPoolExecutor

modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/JraftServerImpl.java
```diff
@@ -466,6 +495,25 @@ public void stopBlockMessages(String groupId) {
...
 
+    /**
+     * Initialize (if needed) and receive per raft group lock,
+     * to prevent concurrent start of the same raft group.
+     *
+     * @param grpId Group id.
+     * @return lock
+     */
+    private Lock groupLock(String grpId) {
+        var newGroupLock = new ReentrantLock();
```
alievmirza, line 506: It seems to me that this is inefficient to create new ReentrantLock every time we call this method. Let's create it only if `startGroupInProgressLocks` doesn't contain lock for grpId

modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/JraftServerImpl.java
```diff
@@ -326,89 +331,113 @@ public synchronized boolean startRaftGroup(
...
+        } finally {
+            lock.unlock();
+        }
     }
 
     /** {@inheritDoc} */
     @Override
-    public boolean stopRaftGroup(String groupId) {
-        RaftGroupService svc = groups.remove(groupId);
+    public boolean stopRaftGroup(String grpId) {
```
alievmirza, line 421: If you call this method twice sequentially, `startGroupInProgressLocks` will contain lock for a stopped group. This may cause OOM, if such groups wont be started anymore.

kgusakov: Reworked the whole locking mechanism - now it is simplier and without these issues

modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/JraftServerImpl.java
kgusakov: Reworked the whole locking mechanism - now it is simplier and without these issues

PR #943, opened 2022-07-15
IGNITE-17133 User flow for CLI

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-17133


modules/cli/src/main/java/org/apache/ignite/cli/call/configuration/JsonString.java
```diff
@@ -0,0 +1,37 @@
...
+package org.apache.ignite.cli.call.configuration;
+
+/**
+ * Wrapper class for String instance with Json content.
+ */
+public class JsonString {
+    private final String value;
+
+    private JsonString(String value) {
+        this.value = value;
```
PakhomovAlexander, line 27: Should the value be validated? I think it might be helpful.

modules/cli/src/main/java/org/apache/ignite/cli/core/flow/builder/FlowBuilderImpl.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * Constructor.
+     *
+     * @param flow flow instance.
+     * @param exceptionHandlers exception handlers.
+     * @param decoratorRegistry decorator registry.
+     */
+    public FlowBuilderImpl(Flow<I, O> flow, ExceptionHandlers exceptionHandlers, DecoratorRegistry decoratorRegistry) {
```
PakhomovAlexander, line 58: Don't think it should be public.

modules/cli/src/main/java/org/apache/ignite/cli/core/flow/builder/FlowBuilderImpl.java
```diff
@@ -0,0 +1,126 @@
...
+        this.flow = flow;
+        this.exceptionHandlers.addExceptionHandlers(exceptionHandlers);
+        this.decoratorRegistry.addAll(decoratorRegistry);
+    }
+
+    @Override
+    public <OT> FlowBuilder<I, OT> appendFlow(Flow<O, OT> flow) {
```
PakhomovAlexander, line 65: I think `then` here is more consistent name. Looking at the next method it is logical to have `then` and `ifThen` instead of `appendFlow` and `ifThen`.

modules/cli/src/test/java/org/apache/ignite/cli/commands/flow/FlowTest.java
```diff
@@ -0,0 +1,98 @@
...
+                        List.of(new QuestionAnswer<>("yes"::equals, (a, i) -> Integer.parseInt(i) * 2),
+                                new QuestionAnswer<>("no"::equals, (a, i) -> Integer.parseInt(i))))
+                .ifThen(num -> num == 1, Flows.fromCall(new IntCall(), IntCallInput::new))
+                .ifThen(num -> num > 1, Flows.fromCall(new StrCall(), integer -> new StrCallInput(String.valueOf(integer))))
+                .build();
+
+
+        bindAnswers("yes", "yes");
+        Flowable<Integer> call = flow.call(Flowable.empty());
+        Assertions.assertEquals(2, call.value());
```
PakhomovAlexander, line 92: As far as I understood, the result has to be 4.

PakhomovAlexander (changes requested):

modules/cli/src/test/java/org/apache/ignite/cli/commands/flow/FlowTest.java
Pochatkin: The first yes gives you the number 1, and the second yes multiplies it by 2. 1 * 2 = 2. Why 4?

modules/cli/src/main/java/org/apache/ignite/cli/commands/questions/ConnectToClusterQuestion.java
```diff
@@ -0,0 +1,74 @@
...
+                        .build())
+                .then(prevUrl -> Flowable.success(clusterUrlOrSessionNode(clusterUrl)));
+    }
+
+    private String clusterUrlOrSessionNode(String clusterUrl) {
+        return clusterUrl != null ? clusterUrl : session.nodeUrl();
+    }
+}
+
+
```
SammyVimes, line 74: Odd newline

modules/cli/src/main/java/org/apache/ignite/cli/core/repl/context/CommandLineContextProvider.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.io.PrintWriter;
+import picocli.CommandLine;
+
+/**
+ * Provider of {@link CommandLineContext}.
+ */
+public class CommandLineContextProvider {
+
+    private static volatile CommandLine cmd;
```
SammyVimes, line 28: I wonder if having a static context can somehow mess up tests

modules/cli/src/main/java/org/apache/ignite/cli/core/flow/question/QuestionAskerFactory.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.cli.core.flow.question;
+
+/**
+ * Factory of {@link QuestionAsker}.
+ */
+public class QuestionAskerFactory {
+    private static volatile QuestionWriterReader questionWriterReader;
```
SammyVimes, line 24: Static context isn't very good for tests I think

SammyVimes (changes requested):
Ok overall, just a couple of questions

modules/cli/src/main/java/org/apache/ignite/cli/core/repl/context/CommandLineContextProvider.java
Pochatkin: Yeah, I agree with you, but I specially leave it here as tech debt. I will add link to ticket for with this debt, are you against it?

modules/cli/src/main/java/org/apache/ignite/cli/core/flow/question/QuestionAskerFactory.java
Pochatkin: Also as above

modules/cli/src/main/java/org/apache/ignite/cli/core/repl/context/CommandLineContextProvider.java
SammyVimes: It's fine with me

Pochatkin: Ok, cool. Added link to issue

SammyVimes (approved):
LGTM

PR #944, opened 2022-07-15
IGNITE-16870 Extend Schema with ability to specify function as default value generator

korlov42:


modules/api/src/main/java/org/apache/ignite/schema/definition/DefaultValueDefinition.java
```diff
@@ -0,0 +1,110 @@
...
+ * Definition of value provider to use as default.
+ */
+@SuppressWarnings("PublicInnerClass")
+public class DefaultValueDefinition {
+    /**
+     * Defines value provider as functional provider.
+     *
+     * @param name Name of the function to invoke to generate the value
+     * @return Default value definition.
+     * @see DefaultValueGenerators
```
ygerzhedovich, line 32:
```suggestion
     * @see DefaultValueGenerator
```

ygerzhedovich: also DefaultValueGenerator is internal

modules/schema/src/main/java/org/apache/ignite/internal/schema/Column.java
```diff
@@ -30,7 +29,7 @@
  */
 public class Column {
     /** Default "default value supplier". */
-    private static final Supplier<Object> NULL_SUPPLIER = () -> null;
+    private static final DefaultValueProvider NULL_SUPPLIER = DefaultValueProvider.constantProvider(null);
```
ygerzhedovich, line 32: Could we add 'nullProvider' into DefaultValueProvider? I think it common case and we could save a memory.

modules/api/src/main/java/org/apache/ignite/schema/definition/ColumnDefinition.java
```diff
@@ -43,10 +43,9 @@ public interface ColumnDefinition {
...
     /**
-     * Returns column default value expression.
+     * Returns default value definition.
      *
-     * @return Default column value expression.
+     * @return Default value definition.
      */
-    //TODO: IGNITE-14479: Rename to defaultValueExpr or allow constants? How to distinct expression from string constant???
-    Object defaultValue();
+    DefaultValueDefinition defaultValueDefinition();
```
ygerzhedovich, line 50:
```suggestion
<T extends DefaultValueDefinition> T defaultValueDefinition();
```

WDYT, will be better to use such approach here?

modules/schema/src/main/java/org/apache/ignite/internal/schema/definition/builder/ColumnDefinitionBuilderImpl.java
```diff
@@ -78,6 +79,9 @@ public ColumnDefinitionBuilderImpl withHints(Map<String, String> hints) {
     /** {@inheritDoc} */
     @Override
     public ColumnDefinition build() {
-        return new ColumnDefinitionImpl(colName, colType, nullable, defValExpr);
+        var defaultSupp = defValExpr == null
```
ygerzhedovich, line 82: should we assert that nullable false for defValExpr = null?
May be need distinguish explicit and implicit null for the case

modules/schema/src/test/java/org/apache/ignite/internal/schema/configuration/SchemaConfigurationConverterTest.java
```diff
@@ -332,32 +345,11 @@ private static Iterable<DefaultValueArg> generateTestArguments() {
...
-
                 paramList.add(
                         new DefaultValueArg(specToType(entry.getKey()), adjust(defaultValue))
                 );
             }
         }
         return paramList;
     }
-
-    // TODO: IGNITE-17370 remove this
```
ygerzhedovich, line 347: should we close IGNITE-17370 at all?

modules/schema/src/main/java/org/apache/ignite/internal/schema/Column.java
```diff
@@ -80,15 +79,15 @@ public Column(
...
      * @param nullable  If {@code false}, null values will not be allowed for this column.
-     * @param defValSup Default value supplier.
+     * @param defaultValueProvider Default value supplier.
      */
     public Column(
             String name,
             NativeType type,
             boolean nullable,
-            @NotNull Supplier<Object> defValSup
+            @NotNull DefaultValueProvider defaultValueProvider
```
AMashenkov, line 88:
```suggestion
            DefaultValueProvider defaultValueProvider
```

modules/schema/src/main/java/org/apache/ignite/internal/schema/Column.java
```diff
@@ -115,16 +114,16 @@ public Column(
...
-     * @param defValSup   Default value supplier.
+     * @param defaultValueProvider   Default value supplier.
      */
     public Column(
             int columnOrder,
             String name,
             NativeType type,
             boolean nullable,
-            @NotNull Supplier<Object> defValSup
+            @NotNull DefaultValueProvider defaultValueProvider
```
AMashenkov, line 124:
```suggestion
            DefaultValueProvider defaultValueProvider
```

modules/schema/src/main/java/org/apache/ignite/internal/schema/Column.java
```diff
@@ -135,22 +134,22 @@ public Column(
...
+     * @param defaultValueProvider   Default value supplier.
      */
     private Column(
             int schemaIndex,
             int columnOrder,
             String name,
             NativeType type,
             boolean nullable,
-            @NotNull Supplier<Object> defValSup
+            @NotNull DefaultValueProvider defaultValueProvider
```
AMashenkov, line 145:
```suggestion
            DefaultValueProvider defaultValueProvider
```

modules/schema/src/main/java/org/apache/ignite/internal/schema/DefaultValueGenerator.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+public enum DefaultValueGenerator {
+    /** Generator that generates random UUID string. */
+    GEN_RANDOM_UUID("genRandomUuid");
+
+    private final Method method;
+
+    DefaultValueGenerator(String methodName) {
+        try {
+            this.method = DefaultValueGenerator.class.getMethod(methodName);
```
AMashenkov, line 35: Can we use MethodHandle here?
From a performance standpoint, the MethodHandles API can be much faster than the Reflection API since the access checks are made at creation time rather than at execution time.

I think. In case of user-defined function, Reflection API looks preferable due to security reasons.
For standard functions like 'randomUUID' MethodHandle API is acceptable
WDYT?

AMashenkov (approved):

modules/api/src/main/java/org/apache/ignite/configuration/schemas/table/ColumnDefaultConfigurationSchema.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.ignite.configuration.annotation.PolymorphicConfig;
+import org.apache.ignite.configuration.annotation.PolymorphicId;
+
+/**
+ * Configuration of default value for table column.
+ */
+@PolymorphicConfig
+public class ColumnDefaultConfigurationSchema {
+    /** Default value is not specified or specified as null explicitly. */
+    public static final String NULL_VALUE_TYPE = "NULL";
```
zstan, line 29: no need public access here

modules/schema/src/main/java/org/apache/ignite/internal/schema/definition/builder/ColumnDefinitionBuilderImpl.java
korlov42: > should we assert that nullable false for defValExpr = null?

I don't think so. `nullable==false && defValExpr==null` means that the user MUST provide value for every such column in insert tuple.

> May be need distinguish explicit and implicit null for the case

Could you please elaborate on why do you think this is important to distinguish between implicit and explicit null?

modules/schema/src/test/java/org/apache/ignite/internal/schema/configuration/SchemaConfigurationConverterTest.java
korlov42: Currently the mentioned issue is still actual in master, hence we need to close this after current patch will be merged to master

modules/schema/src/main/java/org/apache/ignite/internal/schema/definition/builder/ColumnDefinitionBuilderImpl.java
ygerzhedovich: I would say the following situation:
ColumnDefinitionBuilderImpl builder = ......;
builder.asNullable(false);
builder.withDefaultValueExpression(null);

What does it mean? For me it user mistake and we should say about it. But maybe you are right and we just should interpret it as the column value should be define during insert.

modules/api/src/main/java/org/apache/ignite/schema/definition/ColumnDefinition.java
korlov42: Oh, I've got the idea: to avoid manual casting every time. Make sense actually

modules/schema/src/main/java/org/apache/ignite/internal/schema/DefaultValueGenerator.java
ygerzhedovich: It could be different from one JDK to another JDK

modules/api/src/main/java/org/apache/ignite/schema/definition/DefaultValueDefinition.java
korlov42: introduce public enumeration

modules/api/src/main/java/org/apache/ignite/schema/definition/DefaultValueDefinition.java
```diff
@@ -0,0 +1,110 @@
...
+            this.functionName = functionName;
+        }
+
+        /** Returns name of the function to use as value generator. */
+        public String functionName() {
+            return functionName;
+        }
+    }
+
+    /** Defines default value provider as a function call. */
```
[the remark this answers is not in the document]
korlov42: fixed

modules/schema/src/main/java/org/apache/ignite/internal/schema/Column.java
korlov42: done

modules/api/src/main/java/org/apache/ignite/schema/definition/ColumnDefinition.java
korlov42: fixed

modules/schema/src/main/java/org/apache/ignite/internal/schema/definition/builder/ColumnDefinitionBuilderImpl.java
korlov42: This `create table t (c1 int, c2 int not null default null)` is a valid CREATE TABLE statement. Your example is the very same, but via Java API

modules/api/src/main/java/org/apache/ignite/schema/definition/DefaultValueDefinition.java
```diff
@@ -0,0 +1,110 @@
...
+        return new ConstantValue(Objects.requireNonNull(value, "value"));
+    }
+
+    /**
+     * Defines value provider as constant null provider.
+     *
+     * @return Default value definition.
+     */
+    public static DefaultValueDefinition nullValue() {
+        return new DefaultValueDefinition(DefaultValueType.NULL);
```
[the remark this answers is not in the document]
korlov42: fixed

modules/schema/src/main/java/org/apache/ignite/internal/schema/Column.java
korlov42: fixed

modules/schema/src/main/java/org/apache/ignite/internal/schema/Column.java
korlov42: fixed

modules/api/src/main/java/org/apache/ignite/configuration/schemas/table/ColumnDefaultConfigurationSchema.java
korlov42: yep, but it's ok I suppose

ygerzhedovich (approved):

modules/schema/src/main/java/org/apache/ignite/internal/schema/DefaultValueGenerator.java
AMashenkov: The results may depends on security provider. Reflection method will check access rules on each call.

Why don't you use Blackhole to consume the results?
```suggestion
            @Benchmark
    public void methodHandler(Blackhole bh) {
        bh.consume(DefaultValueGenerator.GEN_RANDOM_UUID.nextMethodHandle());
    }
```

korlov42: > The results may depends on security provider. Reflection method will check access rules on each call.

make sense. Fixed

> Why don't you use Blackhole to consume the results?

Because JMH will use blackhole implicitly for returned values

AMashenkov (approved):

zstan (approved):

PR #945, opened 2022-07-16
IGNITE-17328 Introduce implicit transactions from SQL

zstan:


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlAsynchronousApiTest.java
```diff
@@ -173,16 +188,79 @@ public void ddl() throws ExecutionException, InterruptedException {
     public void dml() throws ExecutionException, InterruptedException {
         sql("CREATE TABLE TEST(ID INT PRIMARY KEY, VAL0 INT)");
 
-        IgniteSql sql = CLUSTER_NODES.get(0).sql();
+        IgniteSql sql = igniteSql();
         Session ses = sql.createSession();
 
+        SqlQueryProcessor qryProc = (SqlQueryProcessor) IgniteTestUtils.getFieldValue(sql, IgniteSqlImpl.class, "qryProc");
```
ygerzhedovich, line 194: it's a bit mess, let's introduce separate tests

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/AsyncSqlCursorImpl.java
```diff
@@ -71,6 +77,10 @@ public CompletionStage<BatchedResult<T>> requestNextAsync(int rows) {
                     t = t.getCause();
                 }
 
+                if (tx != null) {
```
ygerzhedovich, line 80: why we rollback the TX?

zstan: we need to somehow handle tx here, this is exceptionally situation, i agree that probably not in all cases we need rollback, but this need to be discussed somehow and probably under differ issue.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlAsynchronousApiTest.java
zstan: done

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/AsyncSqlCursorImpl.java
AMashenkov: What are the cases when transaction must or shouldn't be rolled back?

One the one side. Here, as we already have a cursor, the query is already parsed and mapped to nodes.
So, we can't be sure whether any lock was obtained or not.

On the other side, it is a client API part and server errors are processed on the server-side.
Errors on the client side (e.g. disconnect and marshaller failures) can be handle by the user,
However, I doubt any error is recoverable here and TX will be likely aborted by the user.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/client/ItThinClientSqlTest.java
```diff
@@ -121,6 +121,11 @@ void testExecuteAsyncDdlDml() {
             assertThrows(NoRowSetExpectedException.class, insertRes::currentPage);
         }
 
+        Transaction tx = client().transactions().begin();
```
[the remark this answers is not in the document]
zstan: done

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlSynchronousApiTest.java
```diff
@@ -159,10 +167,20 @@ public void dml() {
         IgniteSql sql = igniteSql();
```
korlov42, line 167: Could you please add the following case for both Sync and Async embedded API implementation:
- start explicit transaction
- insert some rows to a table
- rollback transaction
- verify rows were not inserted

korlov42: Oh, I see there is something similar in implicitTransactionsWithDml test. Let's split this onto several smaller tests

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -366,7 +370,7 @@ public CompletableFuture<Void> closeAsync() {
                     );
                 });
 
-        stage.whenComplete((cur, ex) -> {
+        stage = stage.whenComplete((cur, ex) -> {
```
korlov42, line 373: the idea behind is that you return future with attached cancellation handler to the client. Now you return the future of the handler to the client, and this doesn't make any sense

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionContext.java
```diff
@@ -332,6 +339,18 @@ public interface RunnableX {
         void run() throws Throwable;
     }
 
+    /** Transaction for current context. */
+    public InternalTransaction transaction() {
+        return tx;
+    }
+
+    /** Transaction. */
+    public void transaction(InternalTransaction tx) {
```
korlov42, line 348: do we really need such setter? Looks like it is not used anywhere

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -424,7 +436,7 @@ private void onError(RemoteException ex) {
 
         private void onNodeLeft(String nodeId) {
             remoteFragmentInitCompletion.entrySet().stream().filter(e -> nodeId.equals(e.getKey().nodeId()))
-                    .forEach(e -> e.getValue().completeExceptionally(new IgniteInternalException("asddd")));
+                    .forEach(e -> e.getValue().completeExceptionally(new IgniteInternalException("Node left, nodeId=" + nodeId)));
```
korlov42, line 439:
```suggestion
                    .forEach(e -> e.getValue().completeExceptionally(new IgniteInternalException("Node left the cluster [nodeId=" + nodeId + ']')));
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/MockedStructuresTest.java
```diff
@@ -264,6 +273,48 @@ void before() throws Exception {
                 .get(1, TimeUnit.SECONDS);
     }
 
+    /**
+     * Checks inner transactions are initialized correctly.
+     */
+    @Test
+    public void testInnerTxInitiated() throws Exception {
+        SessionId sesId = queryProc.createSession(new PropertiesHolder() {
+            @Override
```
korlov42, line 282: `PropertiesHolder.holderFor(Map.of())`

zstan:
@korlov42 is it ok for now ?

korlov42 (approved):

PR #948, opened 2022-07-19
IGNITE-17092: Implement topology command

PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-17092

### DIFF
--- a/modules/cli/src/integrationTest/java/org/apache/ignite/cli/commands/cluster/status/ItClusterStatusReplCommandInitializedTest.java
+++ b/modules/cli/src/integrationTest/java/org/apache/ignite/cli/commands/cluster/status/ItClusterStatusReplCommandInitializedTest.java
@@ -15,31 +15,41 @@
  * limitations under the License.
  */
 
-package org.apache.ignite.cli.commands.cluster;
+package org.apache.ignite.cli.commands.cluster.status;
 
 import static org.junit.jupiter.api.Assertions.assertAll;
 
-import org.apache.ignite.cli.commands.CliCommandTestIntegrationBase;
-import org.apache.ignite.cli.commands.cluster.status.ClusterStatusReplSubCommand;
-import org.junit.jupiter.api.Disabled;
+import org.apache.ignite.cli.commands.CliCommandTestInitializedIntegrationBase;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
 
 /**
- * Tests for {@link ClusterStatusReplSubCommand}.
+ * Tests for {@link ClusterStatusReplSubCommand} for the cluster that is initialized.
  */
-class ItClusterStatusReplCommandTest extends CliCommandTestIntegrationBase {
+class ItClusterStatusReplCommandInitializedTest extends CliCommandTestInitializedIntegrationBase {
+
+    String metaStorageNodeName;
+
+    @BeforeAll
+    void setUpMsNodeName(TestInfo testInfo) {
+        metaStorageNodeName = metaStorageNodeName(testInfo);
+    }
 
     @Test
-    @DisplayName("Should print status when valid cluster url is given")
-    @Disabled("Use topology call to get the number of nodes https://issues.apache.org/jira/browse/IGNITE-17092")
+    @DisplayName("Should print status when valid cluster url is given but cluster is initialized")
     void printStatus() {
         execute("cluster", "status", "--cluster-url", NODE_URL);
 
         assertAll(
                 this::assertExitCodeIsZero,
                 this::assertErrOutputIsEmpty,
-                () -> assertOutputContains("Cluster status:")
+                () -> assertOutputContains("name: cluster"),
+                () -> assertOutputContains("nodes: 3"),
+                () -> assertOutputContains("status: active"),
+                () -> assertOutputContains("cmgNodes: [" + metaStorageNodeName + "]"),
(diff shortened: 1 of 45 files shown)


modules/cli/src/integrationTest/java/org/apache/ignite/cli/commands/cluster/status/ItClusterStatusReplCommandInitializedTest.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Tests for {@link ClusterStatusReplSubCommand} for the cluster that is initialized.
+ */
+class ItClusterStatusReplCommandInitializedTest extends CliCommandTestInitializedIntegrationBase {
+
+    String metaStorageNodeName;
+
+    @BeforeAll
+    void setUpMsNodeName(TestInfo testInfo) {
+        metaStorageNodeName = metaStorageNodeName(testInfo);
```
SammyVimes, line 37: It seems like metastorage node name is always `testNodeName(testInfo, 0)`, so probably we may rewrite initializeCluster method to use that by default

modules/cli/src/integrationTest/java/org/apache/ignite/rest/ItGeneratedRestClientTest.java
```diff
@@ -291,6 +296,20 @@ void nodeState() {
         });
     }
 
+    @Test
+    void logicalTopology() {
+        assertDoesNotThrow(() -> {
```
SammyVimes, line 301: `assertDoesNotThrow` seems redundant here as there is `assertThat` inside it, so if it throws, the test fails anyway

modules/cli/src/integrationTest/java/org/apache/ignite/rest/ItGeneratedRestClientTest.java
```diff
@@ -291,6 +296,20 @@ void nodeState() {
...
+            assertThat(topologyApi.logical(), hasSize(3));
+        });
+    }
+
+    @Test
+    void physicalTopology() {
+        assertDoesNotThrow(() -> {
```
SammyVimes, line 308: same thing about assertDoesNotThrow

modules/rest-api/src/main/java/org/apache/ignite/internal/rest/api/cluster/NetworkAddressDto.java
```diff
@@ -0,0 +1,90 @@
...
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.io.Serializable;
+import org.apache.ignite.network.NetworkAddress;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * REST representation of {@link NetworkAddress}.
+ */
+@Schema(name = "NetworkAddress")
+public class NetworkAddressDto implements Serializable {
```
SammyVimes, line 32: Should it be `Serializable`?

modules/rest-api/src/main/java/org/apache/ignite/internal/rest/api/cluster/TopologyApi.java
```diff
@@ -0,0 +1,65 @@
...
+@Controller("/management/v1/cluster/topology")
+@Tag(name = "topology")
+public interface TopologyApi {
+
+    /**
+     * Cluster physical topology.
+     */
+    @Get("physical")
+    @Operation(operationId = "physical")
+    @ApiResponses({
```
SammyVimes, line 45: ApiResponses annotation should not be used directly because ApiResponse is a `@Repeatable` annotation. So it should be just

```
@ApiResponse(responseCode = "200", ...),
@ApiResponse(responseCode = "500", ...)
@Produces(MediaType.APPLICATION_JSON)
Collection<ClusterNodeDto> physicalTopology();
```

modules/rest/openapi/openapi.yaml
```diff
@@ -61,6 +61,48 @@ paths:
             application/problem+json:
               schema:
                 $ref: '#/components/schemas/Problem'
+  /management/v1/cluster/topology/logical:
```
SammyVimes, line 64: just out of curiosity: why do we need this file if we use Swagger that generates schema?

SammyVimes (changes requested):

modules/rest-api/src/main/java/org/apache/ignite/internal/rest/api/cluster/NetworkAddressDto.java
PakhomovAlexander: I think It is a copy-paste issue.

modules/rest-api/src/main/java/org/apache/ignite/internal/rest/api/cluster/TopologyApi.java
PakhomovAlexander: Thanks, makes sense.

modules/rest/openapi/openapi.yaml
PakhomovAlexander: Got your point. This specification is autogenerated by micronaut-openapi generator. It is placed here to make access easier for third-party clients.

SammyVimes: So in theory if someone changed java code but forgot to run generator we can end up with two different schemas, right?

SammyVimes (approved):
LGTM

SammyVimes:
Merged to main https://github.com/apache/ignite-3/commit/7fffd484b3d55ac7428d970e051bba1fd9d92e20

PR #952, opened 2022-07-22
IGNITE-16665 Introduced integer table id.

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-16665

### DIFF
--- a/modules/api/src/main/java/org/apache/ignite/configuration/schemas/table/TableConfigurationSchema.java
+++ b/modules/api/src/main/java/org/apache/ignite/configuration/schemas/table/TableConfigurationSchema.java
@@ -40,7 +40,7 @@ public class TableConfigurationSchema {
     @Immutable
     @Range(min = 1, max = Integer.MAX_VALUE)
     @Value(hasDefault = true)
-    public int intId = 0;
+    public int intId = 1;
 
     /** Table partitions. */
     @Range(min = 0, max = 65_000)
--- a/modules/api/src/main/java/org/apache/ignite/configuration/schemas/table/TableConfigurationSchema.java
+++ b/modules/api/src/main/java/org/apache/ignite/configuration/schemas/table/TableConfigurationSchema.java
@@ -36,7 +36,7 @@ public class TableConfigurationSchema {
     @InjectedName
     public String name;
 
-    /** Small integer id. */
+    /** Integer table id. */
     @Immutable
     @Range(min = 1, max = Integer.MAX_VALUE)
     @Value(hasDefault = true)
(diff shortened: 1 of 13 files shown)


SammyVimes (approved):
looks good to me

modules/api/src/main/java/org/apache/ignite/configuration/schemas/table/TableConfigurationSchema.java
```diff
@@ -35,6 +36,12 @@ public class TableConfigurationSchema {
     @InjectedName
     public String name;
 
+    /** Integer table id. */
+    @Immutable
+    @Range(min = 1, max = Integer.MAX_VALUE)
+    @Value(hasDefault = true)
+    public int intId = 1;
```
tkalkirill, line 43: Why such a name? seems to be more correct just **id**.
Why such a range?
Can we use UUID as an identifier?

modules/api/src/main/java/org/apache/ignite/configuration/schemas/table/TablesConfigurationSchema.java
```diff
@@ -29,6 +29,10 @@
 @SuppressWarnings("PMD.UnusedPrivateField")
 @ConfigurationRoot(rootName = "table", type = ConfigurationType.DISTRIBUTED)
 public class TablesConfigurationSchema {
+    /** Global integer id counter. */
+    @Value(hasDefault = true)
+    public int globalIdCounter = 0;
```
tkalkirill, line 34: Is this the global table identifier? why is it not clearly named? what happens if it overflows?

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/store/FilePageStoreManager.java
```diff
@@ -168,30 +168,30 @@ public long allocatePage(int grpId, int partId, byte flags) throws IgniteInterna
     /**
      * Initializing the file page stores for a group.
      *
-     * @param grpName Group name.
-     * @param grpId Group ID.
+     * @param tableName Table name.
+     * @param groupId Integer table id.
```
tkalkirill, line 172: **tableId** ?

modules/api/src/main/java/org/apache/ignite/configuration/schemas/table/TableConfigurationSchema.java
ibessonov: UUID is already a main "globally unique" identifier. This one - I'm not sure about the name, honestly. Field named "id" already exists.

modules/api/src/main/java/org/apache/ignite/configuration/schemas/table/TablesConfigurationSchema.java
ibessonov: I'l expand the comment

tkalkirill (approved):

PR #956, opened 2022-07-26
IGNITE-17135 Fix inconsistent SQL exceptions on server and client

ptupitsyn:
* Add `ErrorGroups.Sql` with a set of codes for existing cases.
* Fix deprecated constructor calls in SQL exception classes.
* Fix thrown exception types and codes in SQL internals.
* Fix exception wrapping in `Session`.
* Enable and update shared client/server SQL exception tests.


modules/api/src/main/java/org/apache/ignite/internal/sql/ResultSetImpl.java
```diff
@@ -83,7 +85,7 @@ public void close() {
     @Override
     public boolean hasNext() {
         if (it == null) {
-            throw new SqlException("There are no results");
+            throw new SqlException(QUERY_NO_RESULT_SET_ERR, "There are no results");
```
isapego, line 88: This is mostly a question to SQL guys, but are we sure we want to throw exception in this case at all? Why won't we just return `false`? How throwing an exception can be helpful for a user in this case? What do you think?

isapego (approved):

zstan (approved):

modules/api/src/main/java/org/apache/ignite/internal/sql/ResultSetImpl.java
ptupitsyn: We should throw an exception. Returning `false` will produce misleading behavior:

```java
        ResultSet rs = ses.execute(null, "INSERT INTO TEST VALUES (?, ?)", 20, 30);
        rs.forEachRemaining(System.out::println);
```

This will work without any errors if we return `false` from `hasNext`, but iterating over DML/DDL result set does not make sense. I think it's better to let the user know instead of silently allowing this.

isapego: 👍

PR #957, opened 2022-07-27
Ignite 17354

denis-chudov:


modules/metrics/src/main/java/org/apache/ignite/internal/metrics/composite/DistributionMetric.java
```diff
@@ -0,0 +1,144 @@
...
+     * Bounds of the buckets of distribution.
+     *
+     * @return Bounds of the buckets of distribution.
+     */
+    public long[] bounds() {
+        return bounds;
+    }
+
+    /** {@inheritDoc} */
+    @Override public List<Metric> asScalarMetrics() {
```
agura, line 127: A lot of actions. You can do it only once and always return this list just because you use gauges that always will return actual values from measurements.
 
Also, return list should be immutable.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricSource.java
```diff
@@ -0,0 +1,50 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.metrics;
+
+/**
+ * Interface for all metrics source.
```
agura, line 21: I think we should describe the purpose of the metrics source in java doc.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricSource.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+public interface MetricSource {
```
agura, line 23: This entity is source for set of metrics, so right name is `MetricsSource`. Please fix also java docs where this name is wrong.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricSource.java
```diff
@@ -0,0 +1,50 @@
...
+     */
+    String name();
+
+    /**
+     * Enables metrics for metric source. Creates and returns {@link MetricSet} built during enabling. Nothing happens if
+     * the metrics are already enabled for this source.
+     *
+     * @return Newly created {@link MetricSet} instance or {@code null} if metrics are already enabled.
+     */
+    MetricSet enable();
```
agura, line 37: `@Nullable`

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricSet.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.Map;
+import org.apache.ignite.internal.metrics.composite.CompositeMetric;
+import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.TestOnly;
+
+/**
+ * The Metric set that consists of set of metrics identified by a metric name.
+ * The metrics set is immutable.
+ */
+public class MetricSet implements Iterable<Metric> {
```
agura, line 32: This entity is a set of metrics, so right name is `MetricsSet`. Usually we don't use such names in Java (e.g. for variables), but it is needed to symmetry with `MetricsSource` and `MetricsRegistry` entities.

agura (changes requested):
Review is not finished. Just post the first set of comments.

Will continue from `MetricsSet`.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/scalar/HitRateMetric.java
```diff
@@ -0,0 +1,208 @@
...
+    private static byte getTag(long taggedLong) {
+        return (byte) (taggedLong >> TAG_OFFSET);
+    }
+
+    /**
+     * Rate time interval in milliseconds.
+     *
+     * @return Rate time interval in milliseconds.
+     */
+    public long rateTimeInterval() {
```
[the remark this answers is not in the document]
denis-chudov: It could be used later by exporters.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/scalar/HitRateMetric.java
```diff
@@ -0,0 +1,208 @@
...
+        lastHitTimes = new AtomicLongArray(size);
+    }
+
+    /**
+     * Adds x to the metric.
+     *
+     * @param hits Value to be added.
+     */
+    public void add(long hits) {
+        long curTs = currentTimeMillis();
```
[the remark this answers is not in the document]
denis-chudov: I checked both approaches and `System.currentTimeMillis` is slightly better. Are you sure that we need a separate ticket for investigation?

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/composite/DistributionMetric.java
```diff
@@ -0,0 +1,144 @@
...
+
+        return true;
+    }
+
+    /**
+     * Sets value.
+     *
+     * @param x Value.
+     */
+    public void value(long x) {
```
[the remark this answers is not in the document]
denis-chudov: how increment/decrement operations are applicable to the distribution, could you explain? Just add(1) doesnt seem to be useful

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricSet.java
```diff
@@ -0,0 +1,145 @@
...
+        return metrics.values().iterator();
+    }
+
+    /**
+     * The iterator that considers composite metrics as a group of scalar ones, see {@link CompositeMetric#asScalarMetrics()},
+     * and enumerates composite metrics in order to enumerate every scalar metric.
+     *
+     * @return Iterator.
+     */
+    public Iterator<Metric> scalarMetricsIterator() {
```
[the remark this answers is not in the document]
denis-chudov: matbe make CompositeAwareIterator public?

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricSet.java
denis-chudov: it's acceptable to create an adjective from the noun in singular form and it says nothing about the number of things described by the adjective. We agreed to leave it without 's'.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricRegistry.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import org.jetbrains.annotations.TestOnly;
+
+/**
+ * Metric registry. Metrics source (see {@link MetricSource} must be registered in this metrics registry after initialization
+ * of corresponding component and must be unregistered in case of component is destroyed or stopped. Metrics registry also
+ * provides access to all enabled metrics through corresponding metrics sets. Metrics registry lifetime is equal to the node lifetime.
+ */
+public class MetricRegistry {
```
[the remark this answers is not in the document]
denis-chudov: Added iterator for metric sets.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/CompositeAwareIterator.java
```diff
@@ -0,0 +1,81 @@
...
+package org.apache.ignite.internal.metrics;
+
+import java.util.Iterator;
+
+/**
+ * The iterator that considers composite metrics as a group of scalar ones, see {@link CompositeMetric#asScalarMetrics()},
+ * and enumerates composite metrics in order to enumerate every scalar metric.
+ */
+public class CompositeAwareIterator implements Iterator<Metric> {
+    private final Iterator<Metric> iterator;
```
kgusakov, line 27: Javadoc for fields?

modules/metrics/src/test/java/org/apache/ignite/internal/metrics/DistributionMetricTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test for {@link DistributionMetric}.
+ */
+public class DistributionMetricTest {
+    @Test
+    public void test() {
```
kgusakov, line 31: Could you group the tests to smaller named parts?

modules/metrics/src/test/java/org/apache/ignite/internal/metrics/AbstractScalarMetricsTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Abstract test for scalar metrics.
+ *
+ * @param <M> Metric type.
+ * @param <V> Type of metric value.
+ */
+public abstract class AbstractScalarMetricsTest<M extends Metric, V> {
+    @Test
+    public void metricTest() {
```
kgusakov, line 33: Again, could you separate it to smaller logical tests, it will be great to see, which part of logic failed from the test name on TC, for example.

modules/metrics/src/test/java/org/apache/ignite/internal/metrics/ThreadPoolMetricTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+        // Register source after the component created.
+        registry.registerSource(src);
+
+        // ------------------------------------------------------------------------
+
+        // Enable metrics by signal (or because configuration)
+        MetricSet metricSet = registry.enable(src.name());
+
+        LongMetric completedTaskCount = metricSet.get("CompletedTaskCount");
```
kgusakov, line 53: Why do we have test only for 1 metric from ThreadPoolMetricSource?

denis-chudov: Mainly, ThreadPoolMetricSource serves as an example that is a bit less abstract than MetricEntitiesTest, and functionality itself should be checked in other tests, so it seems not to be a problem, as for me

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/scalar/HitRateMetric.java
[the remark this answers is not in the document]
agura: How did you check these approaches? Are there any benchmarks?

> Are you sure that we need a separate ticket for investigation?

No, I don't. As I wrote earlier I suspect a problem here but I'm not sure.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/scalar/HitRateMetric.java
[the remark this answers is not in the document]
agura: Exporters don't need this method.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
```diff
@@ -0,0 +1,169 @@
...
+    private volatile List<Metric> scalarMetrics = null;
+
+    /**
+     * The constructor.
+     *
+     * @param name Name.
+     * @param desc Description.
+     * @param bounds Bounds of the buckets.
+     */
+    public DistributionMetric(String name, @Nullable String desc, @NotNull long[] bounds) {
```
agura, line 51: We don't use `@NotNull` annotation.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
```diff
@@ -0,0 +1,169 @@
...
+        this.measurements = new AtomicLongArray(bounds.length + 1);
+    }
+
+    /**
+     * Check whether given array is sorted.
+     *
+     * @param arr Array to check.
+     * @return {@code True} if array sorted, {@code false} otherwise.
+     */
+    private static boolean isSorted(@NotNull long[] arr) {
```
agura, line 67: We don't use `@NotNull` annotation.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
```diff
@@ -0,0 +1,169 @@
...
+        if (arr.length < 2) {
+            return true;
+        }
+
+        for (int i = 1; i < arr.length; i++) {
+            if (arr[i - 1] > arr[i]) {
```
agura, line 73: What if the comparable elements are equal?

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/composite/DistributionMetric.java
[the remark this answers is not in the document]
agura: My fault. Sorry.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricRegistry.java
```diff
@@ -0,0 +1,300 @@
...
+            IgniteBiTuple<MetricSource, MetricSet> s = new IgniteBiTuple<>(src, null);
+
+            IgniteBiTuple<MetricSource, MetricSet> old = sources.putIfAbsent(src.name(), s);
+
+            if (old != null) {
+                throw new IllegalStateException("Metrics source with given name already exists: " + src.name());
+            }
+
+            // Now we sure that this metric source wasn't registered before.
+            assert !src.enabled() : "Metric source shouldn't be enabled before registration in registry.";
```
agura, line 69: Why should not metric source enabled before registering? It is too strict limitation because the `MetricSource`'s contract doesn't prohibit this.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricRegistry.java
```diff
@@ -0,0 +1,300 @@
...
+
+        try {
+            Map<String, IgniteBiTuple<MetricSource, MetricSet>> sources0 = new TreeMap<>(sources);
+
+            boolean modified = modifier.apply(sources0);
+
+            if (modified) {
+                sources = sources0;
+
+                version++;
```
agura, line 276: It is incorrect behavior because, for example, registering of metric source doesn't change the metrics schema until will not be enabled.

The previous implementation was better because metrics are versioned while sources are not.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricRegistry.java
```diff
@@ -0,0 +1,300 @@
...
+     *
+     * @return Version.
+     */
+    public long version() {
+        return version;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public Iterator<MetricSet> iterator() {
```
agura, line 294: An iterator doesn't provide repeatable behavior. You can't reuse iterator. A new iterator could be returned for new version. So we need to provide a method that returns version and map with metrics atomically.

agura (changes requested):

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
denis-chudov: this condition will be `false` and the method will not return `false` on current check.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
```diff
@@ -0,0 +1,169 @@
...
+     * Bounds of the buckets of distribution.
+     *
+     * @return Bounds of the buckets of distribution.
+     */
+    public long[] bounds() {
+        return bounds;
+    }
+
+    /** {@inheritDoc} */
+    @Override public List<Metric> asScalarMetrics() {
```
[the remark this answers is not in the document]
denis-chudov: this method implements the method of interface so it can be only public

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
```diff
@@ -0,0 +1,169 @@
...
+        if (scalarMetrics == null) {
+            List<Metric> metrics = new ArrayList<>();
+
+            for (int i = 0; i < measurements.length(); i++) {
+                String from = i == 0 ? "0" : String.valueOf(bounds[i - 1]);
```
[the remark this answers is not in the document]
denis-chudov: it has, but implementation differs.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricRegistry.java
denis-chudov: I fixed this and added the tests for checking the correctness of version. Registering the new source seems to be the only case when the map changes and version doesnt, and the old implementation didnt provide copy-on-write semantics so I will leave the current one, in general.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/scalar/HitRateMetric.java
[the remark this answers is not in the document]
denis-chudov: I just wrote simple test with several threads concurrently calling `add` on one metric.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
agura: I mean that equal elements must not be allowed because it will be always empty bucket (e.g. 5_5).

agura: I mean that such situation must be impossible. The current implementation allows buckets like this `new long[] {10, 10, 10}` but it is doesn't make sense.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
```diff
@@ -0,0 +1,169 @@
...
+
+                String name = new StringBuilder(from).append('_').append(to).toString();
+
+                final int index = i;
+                LongGauge gauge = new LongGauge(name, "Single distribution bucket", () -> measurements.get(index));
+
+                metrics.add(gauge);
+            }
+
+            AtomicReferenceFieldUpdater<DistributionMetric, List> updater =
```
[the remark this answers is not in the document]
agura: But it will be created for every instance of `DoubleMetric` while it could be created only once for all instances of `DoubleMetric`.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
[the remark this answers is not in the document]
agura: But this bucket is not needed. If a user wants zero-bucket he will define it explicitly.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricRegistry.java
agura: Once again, `MetricSource`'s contract doesn't protect a developer from this error. Moreover, metrics source could be disabled by configuration but it is possible to register not enabled metrics source.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricRegistry.java
```diff
@@ -0,0 +1,332 @@
...
+            lock.unlock();
+        }
+    }
+
+    /**
+     * Metric set schema.
+     *
+     * @return Metric set schema.
+     */
+    public MetricSetSchema metricSetSchema() {
```
agura, line 306: There is no need to make a copy every time when exporter gets the metrics. It's enough to make metrics copy on modifying the metrics set collection and this copy could be reused multiple times. It is what I mean when talk about copy-on-write semantic. 

Also, the pair of `version` and `metricSets` is not a schema. 

Please, back to the previous implementation with different collections for sources and metric sets. It is simpler and more clear.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
denis-chudov: fixed, method is remaned to `isSortedAndUnique`.

PR #958, opened 2022-07-28
IGNITE-17433 Fix SQL exceptions when cursor is closed or has no more pages

ptupitsyn:
* Remove internal `SessionNotFoundException` - use the error code for internal logic.
* Move `CursorClosedException` to public API.
* Fix thrown exceptions on server and client.
* Add shared client/server tests to ensure consistent behavior.


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
```diff
@@ -163,7 +162,7 @@ public CompletableFuture<AsyncResultSet> executeAsync(@Nullable Transaction tran
             );
 
             result.whenComplete((rs, th) -> {
-                if (th instanceof SessionNotFoundException) {
+                if (th instanceof IgniteException && ((IgniteException) th).code() == SESSION_NOT_FOUND_ERR) {
```
isapego, line 165: We may want to have an utility method for this in future I believe.

isapego (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
ptupitsyn: Done

PR #959, opened 2022-07-28
IGNITE-17082: Rename ignite-config.json to ignite-config.hocon

PakhomovAlexander:
Also, related todo's are removed. The reason why I decided to
keep using `text/plain` for HOCON configuration is that
tooling (swagger-codegen, for example) do not support this
media type that is suggested to use for HOCON
(application/hocon). The reason of weak support is that
application/hocon is not the official MIME type.

We have integration tests guaranties that the returned
String can be deserialized as HOCON config.


PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-17082

ibessonov (approved):
Please correct checkstyle violations in tests

ibessonov (approved):

PR #961, opened 2022-07-29
IGNITE-17347 Add port parameters to Ignite3 CLI node start command

valepakh:
This will add --port, --rest-port and --join parameters to simplify starting nodes especially for local testing.

https://issues.apache.org/jira/browse/IGNITE-17347


modules/cli/src/main/java/org/apache/ignite/cli/deprecated/spec/NodeCommandSpec.java
```diff
@@ -100,12 +123,36 @@ public Integer call() {
...
+        private Path getConfigPath() {
+            return configOptions != null ? configOptions.configPath : null;
+        }
+
+        private String getConfigStr() {
+            if (configOptions == null || configOptions.configPath != null) {
+                return null;
+            }
+            // port is required
+            StringBuilder sb = new StringBuilder();
```
PakhomovAlexander, line 136: I think it is better to use `Config` class from `typesafe` lib to construct a configuration representation.

modules/cli/src/main/java/org/apache/ignite/cli/deprecated/spec/NodeCommandSpec.java
```diff
@@ -62,12 +65,31 @@ public static class StartNodeCommandSpec extends BaseCommand implements Callable
...
         public String nodeName;
 
-        /** Path to node config. */
-        @CommandLine.Option(names = "--config", description = "Configuration file to start the node with")
-        private Path configPath;
+        @ArgGroup
+        private ConfigOptions configOptions;
+
+        private static class ConfigOptions {
+            @ArgGroup(exclusive = false)
```
PakhomovAlexander, line 75: Could you add an unit test to verify that it is truly exclusive?

modules/cli/src/main/java/org/apache/ignite/cli/deprecated/builtins/node/NodeManager.java
```diff
@@ -136,6 +137,9 @@ public RunningNode start(
             if (srvCfg != null) {
                 cmdArgs.add("--config");
                 cmdArgs.add(srvCfg.toAbsolutePath().toString());
+            } else if (srvCfgStr != null) {
+                cmdArgs.add("--configStr");
```
sk0x50, line 141: `--configStr` does not look descriptive to be honest :) By the way, do we use camel case for commands? Perhaps it should be named `config-str`, for instance see at `work-dir`.

modules/cli/src/main/java/org/apache/ignite/cli/deprecated/spec/NodeCommandSpec.java
```diff
@@ -102,12 +130,32 @@ public Integer call() {
             out.println(tbl);
             return 0;
         }
+
+        private Path getConfigPath() {
+            return configOptions != null ? configOptions.configPath : null;
+        }
+
+        private String getConfigStr() {
```
sk0x50, line 138: Well, looks like these parameters (port, rest-port etc) cannot be provided with a configuration file. Right?
It seems to me, that ability to start two nodes with the same configuration file and overridden value of `port` would be very convenient. I mean the following:
`node start nodeName --config \path_to_configuration_file --port 12345`

What do you think?

modules/cli/src/main/java/org/apache/ignite/cli/deprecated/builtins/node/NodeManager.java
valepakh: `--config-file` and `--config-string` maybe?

modules/cli/src/main/java/org/apache/ignite/cli/deprecated/spec/NodeCommandSpec.java
valepakh: Yes I think this is possible and will be helpful.

valepakh:
I changed CLI runner parameters to `--config-path` and `--config-string` and implemented overriding config file with parameters.
Also I improved the IgniteCliInterfaceTest so that the mockserver only starts once, this significantly improved performance when running on my local machine in IDEA.

modules/cli/src/main/java/org/apache/ignite/cli/deprecated/builtins/node/NodeManager.java
sk0x50: Perhaps, `config-overrides` would cleaner. What do you think?

sk0x50: Looks like I was wrong. My last comment does not make sense. Please ignore.

PR #964, opened 2022-07-29
IGNITE-14487: Table view bulk methods refactoring.

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-14487

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/KeyValueBinaryViewImpl.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/KeyValueBinaryViewImpl.java
@@ -21,7 +21,6 @@ import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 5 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/KeyValueBinaryViewImpl.java
```diff
@@ -475,8 +476,8 @@ private Tuple unmarshalValue(BinaryRow row) {
      * @param rows Binary rows.
      * @return Key-value pairs of tuples.
      */
-    public Map<Tuple, Tuple> unmarshalValue(Collection<BinaryRow> rows) {
-        Map<Tuple, Tuple> pairs = new HashMap<>();
+    private Map<Tuple, Tuple> unmarshalValue(Collection<BinaryRow> rows) {
+        Map<Tuple, Tuple> pairs = new HashMap<>(rows.size());
```
zstan, line 480: Seems we still obtain reallocations here, check [1]

[1] https://github.com/apache/ignite/blob/master/modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java#L10140

zstan (approved):

PR #965, opened 2022-07-29
IGNITE-17439 Small enhancements for BinaryTuple

ademakov:


AMashenkov (approved):

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleSchema.java
```diff
@@ -289,10 +289,23 @@ public static int nullMapSize(int numElements) {
     }
 
     /**
-     * Returns the null map size in bytes if there are nullable elements, zero otherwise.
+     * Get offset of the byte that contains null-bit of a given tuple element.
```
korlov42, line 292: I believe we should use 'Returns' verb here and below. You can _get_ the offset by calling this method, whereas the method _returns_ this value to you

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleSchema.java
```diff
@@ -289,10 +289,23 @@ public static int nullMapSize(int numElements) {
...
+     *
+     * @param index Tuple element index.
+     * @return Offset of the required byte relative to the tuple start.
+     */
+    public static int getNullOffset(int index) {
```
korlov42, line 297: lets rename this to `nullByteOffset`, because that name is better reflects the essence

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleSchema.java
```diff
@@ -289,10 +289,23 @@ public static int nullMapSize(int numElements) {
...
+
+    /**
+     * Get a null-bit mask corresponding to a given tuple element.
+     *
+     * @param index Tuple element index.
+     * @return Mask to extract the required null-bit.
      */
-    public int nullMapSize() {
-        return hasNullableElements() ? nullMapSize(elementCount()) : 0;
+    public static byte getNullMask(int index) {
```
korlov42, line 307: as far as I can see, vast majority of the methods within this class omit the `get` prefix. Lets follow this rule to make code more consistent

korlov42 (approved):

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleSchema.java
ademakov: Well, are you talking about style or grammar? I know that sun/oracle recommends using 3rd person indicative in the docs. So this should go as "Gets ...". I don't understand why you urge to use "Returns ...", there is no basis for this neither in oracle recommendations nor in English grammar.

As for grammar, I'd personally prefer to use imperative mood. It is quite often used in documentation elsewhere and IMHO more correct than the 3rd person indicative. "Gets ..." is not a full sentence. A full sentence in indicative mood requires a subject. So to stay grammatically correct it should be written like this "This method gets ...". It is strange why oracle promotes bad English in their recommendations. But so be it.

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleSchema.java
ademakov: Perhaps the current name is not very good. But I think `nullByteOffset` would be confusing. People could misread it as "Offset of a null byte" and wrongfully think that this is about some byte with zero value. So to be entirely clear it should be something like `byteOffsetOfNullBitOctet`. What do you think?

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleSchema.java
ademakov: Ok.

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTupleSchema.java
korlov42: Well, this is more like suggestion.

The reasons I prefer "Returns" over "Gets" are follow:

- with "Returns" it sounds more natural to me
- built in java library uses "Returns" instead of "Gets" (please take a look at any class/interface at java.util package, for instance; honesty speaking, I've never seen "Gets" anywhere but in ignite codebase)
- "Returns" is imperative as well

Once again, it's just a suggestion. Feel free to ignore this (then just resolve the conversation)

ademakov: * The word get is perfectly natural. You could find a ton of cases where "get" is used in docs. E.g. `man getenv`:

```
GETENV(3)                                                        Linux Programmer's Manual                                                       GETENV(3)

NAME
       getenv, secure_getenv - get an environment variable
```

Methods are called like "getX()" instead of "returnX()" for a reason. The Java docs use `returns` more often perhaps to escape tautology. I guess Java doc authors think that "getX() -- returns X" looks better than "getX() -- gets X".

* Still it does not take a long time to find some examples of using "Gets" there:

https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/lang/System.html
```
getProperty()     Gets the system property indicated by the specified key.
```

* "Returns" is not imperative. It's 3rd person indicative. "Return" would be imperative. In python world their require to use imperative: 

https://peps.python.org/pep-0257/#one-line-docstrings

```
The docstring is a phrase ending in a period. It prescribes the function or method’s effect as a command (“Do this”, “Return that”), not as a description; e.g. don’t write “Returns the pathname …”.
```

It appears that Python people use English more carefully than Java people.

Regardless, I edited the comments the way you asked.

PR #966, opened 2022-07-29
IGNITE-17429 Sql. Extend sql grammar to support CREATE/DROP INDEX statements

korlov42:


AMashenkov (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/sql/IgniteSqlCreateIndex.java
```diff
@@ -41,6 +41,8 @@ public class IgniteSqlCreateIndex extends SqlCreate {
     /** Table name.  */
     private final SqlIdentifier tblName;
 
+    private final IgniteSqlIndexType type;
```
AMashenkov, line 44: Javadoc?

korlov42: According to coding guidelines, it's allowed to omit the javadoc for private members in case they are self-describing. I consider this is the case

PR #972, opened 2022-08-02
IGNITE-16657 [Native Persistence 3.0] Implement partitions chunks merger

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-16657


modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/compaction/Compactor.java
```diff
@@ -0,0 +1,372 @@
...
+        assert !queue.isEmpty();
+
+        updateHeartbeat();
+
+        int threads = threadPoolExecutor == null ? 1 : threadPoolExecutor.getMaximumPoolSize();
+
+        CompletableFuture<?>[] futures = new CompletableFuture[threads];
+
+        for (int i = 0; i < threads; i++) {
+            CompletableFuture<?> future = futures[i] = new CompletableFuture<>();
```
SammyVimes, line 208: nice

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/compaction/Compactor.java
```diff
@@ -0,0 +1,372 @@
...
+        ByteBuffer buffer = threadBuf.get();
+
+        for (long pageOffset : deltaFilePageStore.pageOffsets()) {
+            updateHeartbeat();
+
+            if (isCancelled()) {
+                return;
+            }
+
+            boolean read = deltaFilePageStore.readWithMergedToFilePageStoreCheck(pageOffset, pageOffset, buffer.rewind(), false);
```
SammyVimes, line 329: why page offset twice?

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/store/DeltaFilePageStoreIo.java
```diff
@@ -110,4 +149,21 @@ public long pageOffset(long pageId) {
     public int fileIndex() {
         return header.index();
     }
+
+    /**
+     * Returns page offsets within the store file.
+     */
+    public long[] pageOffsets() {
```
SammyVimes, line 156: why copy int array to a long array? Seems a bit like overkill

SammyVimes (changes requested):

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/compaction/Compactor.java
tkalkirill: Thanks

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/compaction/Compactor.java
```diff
@@ -0,0 +1,372 @@
...
+            }
+
+            if (threadPoolExecutor == null) {
+                merger.run();
+            } else {
+                try {
+                    threadPoolExecutor.execute(merger);
+                } catch (RejectedExecutionException ignore) {
+                    // Run the task synchronously.
+                    merger.run();
```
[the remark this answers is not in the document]
tkalkirill: This is a copy of the code from 2.0, the rejection can be if it was not possible to add a worker due to a stop or in the process of shutting down. 

In theory, we won’t have this, because at the stop of the **Compactor** we say that we are canceling, we wait for all threads to complete and only then shutdown the thread pool. 

I will remove this code here and at the Checkpointer.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/store/DeltaFilePageStoreIo.java
tkalkirill: Fix it

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/compaction/Compactor.java
tkalkirill: Added a comment.

SammyVimes (approved):
LGTM

PR #974, opened 2022-08-03
IGNITE-17233 Clarify cluster URL parameter name

valepakh:
https://issues.apache.org/jira/browse/IGNITE-17233

--cluster-url option could be misleading. Let's change it to --cluster-endpoint-url.


Pochatkin:
LGTM

PR #977, opened 2022-08-03
IGNITE-17286 Add missed busy locks to get rid of resources leaking during table creation

alievmirza:


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1063,17 +1107,33 @@ public CompletableFuture<List<Table>> tablesAsync() {
...
+                return directTableIds();
+            } finally {
+                busyLock.leaveBusy();
+            }
+        }).thenCompose(tableIds -> {
+            if (!busyLock.enterBusy()) {
+                throw new IgniteException(NODE_STOPPING_ERR, new NodeStoppingException());
+            }
+            try {
+                var tableFuts = new CompletableFuture[tableIds.size()];
```
denis-chudov, line 1124: shouldn't we make some wrapper, like `withBusyLock(busyLock, lambda)` - maybe it will make the code more compact

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/SqlSchemaManagerImpl.java
```diff
@@ -214,50 +195,58 @@ public synchronized CompletableFuture<?> onTableCreated(
...
+                                            }
+
+                                            Map<UUID, IgniteTable> resTbls = new HashMap<>(tables);
+
+                                            return igniteTableFuture
+                                                    .thenApply(igniteTable -> {
+                                                        resTbls.put(igniteTable.id(), igniteTable);
+
+                                                        return resTbls;
+                                                    });
```
denis-chudov, line 230: this will be asynchronous part in this update, possibly needs to be wrapped with busy lock

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -598,7 +622,23 @@ public void stop() {
...
+
+        try {
+            // Waiting for tablesByIdVv generally can be unbounded, because we don't have time limits for receiving storage revision update,
+            // where tablesByIdVv is completed, so we add timeout here.
+            tables = tablesByIdVv.waitForLatest(TABLES_COMPLETE_TIMEOUT, TimeUnit.SECONDS);
+        } catch (ExecutionException | InterruptedException | TimeoutException e) {
+            throw new IgniteException(TABLE_NOT_COMPLETED_ERR);
+        }
+
+        cleanUpTablesResources(tables);
```
denis-chudov, line 635: shouldnt we save some context of the tables that are being created, we already have the futures for them (beforeTablesVvComplete) - they can be enriched

modules/schema/src/main/java/org/apache/ignite/internal/schema/SchemaManager.java
```diff
@@ -336,11 +345,20 @@ private SchemaDescriptor getSchemaDescriptorLocally(int schemaVer, ExtendedTable
...
         }
 
         try {
-            return registriesVv.get(causalityToken).thenApply(regs -> tableId == null ? null : regs.get(tableId));
+            return registriesVv.get(causalityToken).thenApply(regs -> {
+                if (!busyLock.enterBusy()) {
+                    throw new IgniteException(NODE_STOPPING_ERR, new NodeStoppingException());
+                }
+                try {
+                    return tableId == null ? null : regs.get(tableId);
```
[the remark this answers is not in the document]
alievmirza: I think it's out of the context of this task

PR #979, opened 2022-08-04
IGNITE-17308 Rework SortedIndexMvStorage interface

sashapolo:
* Indices are no longer stored in PartitionMvStorage
* Indices are no longer aware of the MVCC properties
* Merge functionality with existing sorted indices implementation

https://issues.apache.org/jira/browse/IGNITE-17308


modules/core/src/test/java/org/apache/ignite/internal/testframework/VariableArgumentsProvider.java
```diff
@@ -45,7 +45,7 @@ public void accept(VariableSource variableSource) {
     public Stream<? extends Arguments> provideArguments(ExtensionContext context) throws Exception {
         Class<?> testClass = context.getRequiredTestClass();
 
-        Field source = testClass.getDeclaredField(variableName);
+        Field source = testClass.getField(variableName);
```
ibessonov, line 48: Very interesting, but what was wrong with non-public fields?

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexRow.java
```diff
@@ -17,26 +17,26 @@
...
 public interface IndexRow {
     /**
-     * Returns the serialized presentation of this row as a byte array.
+     * Returns the serialized presentation of indexed columns as a byte array in
+     * <a href="https://cwiki.apache.org/confluence/display/IGNITE/IEP-92%3A+Binary+Tuple+Format">IEP-92</a> format.
      *
      * @return Serialized byte array value.
      */
-    byte[] rowBytes();
+    ByteBuffer indexBytes();
```
ibessonov, line 34: It's weird that here we operate with bytes instead of InternalTuple or something. What's the reason of such decision? I find it non-intuitive

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexRowPrefix.java
```diff
@@ -23,7 +23,7 @@
 public interface IndexRowPrefix {
```
ibessonov, line 23: I believe that we should use InternalTuple for prefixes as well, there's no point in having additional interface now

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexRowSerializer.java
```diff
@@ -17,17 +17,17 @@
...
- * specified by the {@link SortedIndexDescriptor#indexRowColumns()}.
+ * specified by the {@link SortedIndexDescriptor#indexColumns()}.
  */
-public interface IndexRowFactory {
+public interface IndexRowSerializer {
     /**
      * Creates an Index row from a list of column values.
      */
-    IndexRow createIndexRow(Object[] columnValues, SearchRow primaryKey);
+    IndexRow createIndexRow(Object[] columnValues, RowId rowId);
```
ibessonov, line 30: This approach is inherently ineffective, most of column values could be copied directly from BinaryRow's payload. Serialization to Object and back is redundant

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexDescriptor.java
```diff
@@ -17,72 +17,58 @@
...
 
         private final boolean nullable;
 
-        ColumnDescriptor(Column column, boolean asc, boolean indexedColumn, boolean nullable) {
-            this.column = column;
-            this.asc = asc;
-            this.indexedColumn = indexedColumn;
-            this.nullable = nullable;
-        }
+        private final boolean asc;
```
ibessonov, line 51: Since we're here, you can add "nullFirst" flag, but only if it's easy and won't take much time.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexMvStorage.java
```diff
@@ -51,68 +50,47 @@ public interface SortedIndexMvStorage {
...
-            @Nullable IndexRowPrefix upperBound,
-            @MagicConstant(flagsFromClass = SortedIndexStorage.class) int flags,
-            UUID txId,
-            @Nullable IntPredicate partitionFilter
-    );
+    void remove(IndexRow row);
 
     /**
-     * Returns a range of index values between the lower bound and the upper bound, consistent with the passed timestamp.
+     * Returns a range of index values between the lower bound and the upper bound, consistent with the passed transaction id.
```
ibessonov, line 80: I'd remove the "consistent with" part, there's no such parameter now

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexMvStorage.java
```diff
@@ -51,68 +50,47 @@ public interface SortedIndexMvStorage {
...
+    /**
+     * Returns a factory for creating index rows for this storage.
+     */
+    IndexRowSerializer indexRowSerializer();
 
-    boolean supportsBackwardsScan();
+    /**
+     * Returns a class deserializing index columns.
+     */
+    IndexRowDeserializer indexRowDeserializer();
```
ibessonov, line 65: Again, we should discuss the decision of having separate serializer/deserializer in index. It's not clear, maybe we should remove them

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/BinaryTupleComparator.java
```diff
@@ -0,0 +1,149 @@
...
+import org.apache.ignite.internal.schema.BinaryTuple;
+import org.apache.ignite.internal.schema.BinaryTupleSchema;
+import org.apache.ignite.internal.schema.NativeType;
+import org.apache.ignite.internal.storage.index.SortedIndexDescriptor;
+import org.apache.ignite.internal.storage.index.SortedIndexDescriptor.ColumnDescriptor;
+
+/**
+ * Comparator implementation for comparting {@link BinaryTuple}s on a per-column basis.
+ */
+class BinaryTupleComparator implements Comparator<ByteBuffer> {
```
ibessonov, line 39: Please add TODO for https://issues.apache.org/jira/browse/IGNITE-17325

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/BinaryTupleComparator.java
```diff
@@ -0,0 +1,149 @@
...
+    /**
+     * Actual comparator implementation.
+     */
+    private final Comparator<ByteBuffer> delegate;
```
ibessonov, line 43: What's the point of having a delegate if we could just implement compareTo? What are the benefits?

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/BinaryTupleComparator.java
```diff
@@ -0,0 +1,149 @@
...
+        this.indexDescriptor = descriptor;
+
+        this.delegate = comparing(bytes -> new BinaryTuple(schema, bytes), binaryTupleComparator());
+    }
+
+    /**
+     * Creates a comparator that compares two {@link BinaryTuple}s by comparing individual columns.
+     */
+    private Comparator<BinaryTuple> binaryTupleComparator() {
+        return IntStream.range(0, indexDescriptor.indexColumns().size())
```
ibessonov, line 61: I expect comparison to be a very hot method. Let's avoid all these streams and lambdas, ok?

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/PrefixComparator.java
```diff
@@ -15,77 +15,67 @@
...
+import org.apache.ignite.internal.storage.index.SortedIndexDescriptor;
 import org.apache.ignite.internal.storage.index.SortedIndexDescriptor.ColumnDescriptor;
 import org.jetbrains.annotations.Nullable;
 
 /**
- * Class for comparing a {@link BinaryRow} representing an Index Key with a given prefix of index columns.
+ * Class for comparing a {@link BinaryTuple} representing an Index Row with a given prefix of index columns.
  */
-public class PrefixComparator {
+class PrefixComparator {
```
ibessonov, line 32: I expect this class to disappear in the future, is this a reasonable expectation? There should be a single place with comparison implementation

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexRowSerializer.java
sashapolo: This API is temporary and should be removed as soon as we understand how it will be integrated with the partition storage

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexMvStorage.java
sashapolo: Yes, I also don't like them, they should be removed when indices get integrated with the storage

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/BinaryTupleComparator.java
sashapolo: This is a test class, should we bother?

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/BinaryTupleComparator.java
sashapolo: you can't use all those functional things, like `comparing` and stuff, because they return a Comparator

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/BinaryTupleComparator.java
sashapolo: It's a test class, why should we bother? I expect the production implementation to be different

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/PrefixComparator.java
sashapolo: yes, I left it for simplicity purposes

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/BinaryTupleComparator.java
sashapolo: I believe there also was a ticket about implementing a naturally sorted byte order format

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexRow.java
sashapolo: Because it is impossible to deserialize an `InternalTuple`, while a `ByteBuffer` can be wrapped into a `BinaryTuple`. Returning a `BinaryTuple` might be a good alternative, but I'm not sure if it's actually better

modules/core/src/test/java/org/apache/ignite/internal/testframework/VariableArgumentsProvider.java
sashapolo: `getDeclaredField` doesn't look into fields that are declared in superclasses. It could be fixed by manually traversing the hierarchy, but I don't see anything wrong with simply declaring these variables as `public`

ibessonov: Ok, I realized that after commenting. It's fine

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexRow.java
ibessonov: I think it would be better. I'd prefer InternalTuple for convenience.
Anyway, there should be a common way of converting a InternalTuple into IEP-92 data format, otherwise such API wouldn't work. I realize that, and I believe that we should provide this "common way"

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/BinaryTupleComparator.java
ibessonov: I didn't realize it's a test class, sorry

ibessonov: Naturally sorted byte order format will not be implemented, as far as I know, we abandoned this idea

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/BinaryTupleComparator.java
ibessonov: Yes, but they're slow

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/BinaryTupleComparator.java
ibessonov: Ok, again, I didn't realize it was in a "test" folder

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexMvStorage.java
ibessonov: Ok, I guess we'll do it in https://issues.apache.org/jira/browse/IGNITE-17310

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/BinaryTupleComparator.java
sashapolo: Yes, but this is a test class, so I don't think it matters. This comparator should be heavily refactored for production use anyway

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexRow.java
sashapolo: Replaced with BinaryTuple

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexRowPrefix.java
sashapolo: Removed this class

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/BinaryTupleComparator.java
sashapolo: I removed the delegate

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/BinaryTupleComparator.java
sashapolo: I removed these streams

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/index/impl/PrefixComparator.java
sashapolo: I removed this class

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexMvStorage.java
sashapolo: oops

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexDescriptor.java
sashapolo: I'd rather not, because it's Friday and I'm too tired

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexRow.java
```diff
@@ -26,12 +26,12 @@
...
     /**
-     * Returns the serialized presentation of indexed columns as a byte array in
+     * Returns the serialized presentation of indexed columns as a {@link BinaryTuple} in
      * <a href="https://cwiki.apache.org/confluence/display/IGNITE/IEP-92%3A+Binary+Tuple+Format">IEP-92</a> format.
      *
-     * @return Serialized byte array value.
+     * @return Binary tuple representation of indexed columns.
      */
-    ByteBuffer indexBytes();
+    BinaryTuple indexColumns();
```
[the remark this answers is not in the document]
ibessonov: You see, the problem with BinaryTuple is that it contains its own schema for no good reason.
Well, there is a reason, but it's artificial and can be worked around

PR #980, opened 2022-08-04
IGNITE-14937 Introduce index management

korlov42:


AMashenkov (approved):

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -0,0 +1,263 @@
...
+                                .orElse(null);
+
+                        if (createdIndex != null) {
+                            LOG.info("Index created [schemaName={}, tableName={}, indexName={}, indexId={}]",
+                                    schemaName, tableName, indexName, createdIndex.id());
+
+                            future.complete(createdIndex);
+                        } else {
+                            var exception = new IgniteInternalException(
+                                    Common.UNEXPECTED_ERR, "Looks like the index was concurrently deleted");
```
ygerzhedovich, line 192: is it possible? I thought that all changer run sequentially and can't be run concurrently

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -0,0 +1,263 @@
...
+        if (!busyLock.enterBusy()) {
+            return CompletableFuture.failedFuture(new NodeStoppingException());
+        }
+
+        LOG.debug("Going to create index [schemaName={}, tableName={}, indexName={}]", schemaName, tableName, indexName);
+
+        try {
+            validateName(indexName);
+        } catch (Exception ex) {
+            return CompletableFuture.failedFuture(ex);
```
ygerzhedovich, line 129: for the case you don't leave busy lock

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -0,0 +1,263 @@
...
+            String schemaName,
+            String indexName,
+            String tableName,
+            Consumer<TableIndexChange> indexChange
+    ) {
+        if (!busyLock.enterBusy()) {
+            return CompletableFuture.failedFuture(new NodeStoppingException());
+        }
+
+        LOG.debug("Going to create index [schemaName={}, tableName={}, indexName={}]", schemaName, tableName, indexName);
```
ygerzhedovich, line 124: maybe short version of parameters name will be better here and below? Like a schema={}, table={}, index={}

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -0,0 +1,263 @@
...
+                    indexListChange.create(indexName, indexChange);
+
+                    TableIndexView indexView = indexListChange.get(indexName);
+
+                    Set<String> columnNames = Set.copyOf(tableChange.columns().namedListKeys());
+
+                    validateColumns(indexView, columnNames);
+                })).whenComplete((index, th) -> {
+                    if (th != null) {
+                        LOG.info("Unable to create index [schemaName={}, tableName={}, indexName={}]",
```
zstan, line 170: invalid params count ?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -0,0 +1,263 @@
...
+
+    private void validateColumns(TableIndexView indexView, Set<String> tableColumns) {
+        if (indexView instanceof SortedIndexView) {
+            var sortedIndexView = (SortedIndexView) indexView;
+
+            validateColumns(sortedIndexView.columns().namedListKeys(), tableColumns);
+        } else if (indexView instanceof HashIndexView) {
+            validateColumns(Arrays.asList(((HashIndexView) indexView).columnNames()), tableColumns);
+        } else {
+            throw new IllegalStateException("Unknown index type [type=" + (indexView != null ? indexView.getClass() : null) + ']');
```
zstan, line 242: why Illegal**State** ?

modules/index/src/main/java/org/apache/ignite/internal/index/SortedIndex.java
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * An object describing a sorted index.
+ *
+ * <p>Extends a basic index with operations related to sorted structures.
+ */
+public interface SortedIndex extends Index {
+    /** A flag denotes that left bound should be included into result. */
+    byte INCLUDE_LEFT = 0b01;
+
+    /** A flag denotes that left bound should be included into result. */
```
zstan, line 33: right

modules/index/src/main/java/org/apache/ignite/internal/index/SortedIndex.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.BitSet;
+import org.apache.ignite.internal.schema.BinaryTuple;
+import org.apache.ignite.internal.util.Cursor;
+
+/**
+ * An object describing a sorted index.
+ *
+ * <p>Extends a basic index with operations related to sorted structures.
+ */
+public interface SortedIndex extends Index {
```
zstan, line 29: Just pure interface for now ? Do we need some kind of TODO ? here ?

korlov42: Why do we need a TODO here?

modules/index/src/main/java/org/apache/ignite/internal/index/SortedIndex.java
korlov42: good catch! thanks

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
korlov42: I don't think so. Please take a look at `org.apache.ignite.internal.logger.IgniteLogger#info(java.lang.String, java.lang.Throwable, java.lang.Object...)`

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
korlov42: >  I thought that all changer run sequentially and can't be run concurrently

This is true. But the line you are pointing to is not a part of a changer closure. This actually a callback that would be invoked _after_ the changer closure is applied to the configuration

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
korlov42: Here is description of the IllegalState:
```
 * Signals that a method has been invoked at an illegal or
 * inappropriate time.  In other words, the Java environment or
 * Java application is not in an appropriate state for the requested
 * operation.
```

Look like it really doesn't fit very well. Do you have any suggestions?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
korlov42: good catch! thanks

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
korlov42: ok, fixed

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
zstan: Why IgniteException is not acceptable here? And IllegalArgumentException more close to problem i hope ?

modules/index/src/main/java/org/apache/ignite/internal/index/SortedIndex.java
zstan: And why do we need pure interface?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
korlov42: IgniteException is about violation business rules, and IllegalArgumentException is about passing wrong arguments (for instance, javadoc said "Argument should be greater than zero" and someone passed -1). Both of these are exception of valid codebase, but line in question is about someone just added new index types, but failed to update all required places, hence we got inconsistency in our codebase

Is it OK to throw AssertionError here?

modules/index/src/main/java/org/apache/ignite/internal/index/SortedIndex.java
korlov42: Well, actually it's OK to commit an interface without any implementation if you need to provide a contract. But this not the case. So removed it

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
zstan: burn it !)

zstan (approved):

PR #986, opened 2022-08-08
IGNITE-17349 Add common UI components

PakhomovAlexander:
Now everything that we are going to print should go through
UI component. The list of UI components decide how to render
errors, tables and more.

ValidationProblem is removed because it is impossible to
understand from the http status code to what class body
should be deserialized: Problem or ValidationProblem (both
of them have 400 http code). So, now there is only one
Problem that in case of validation issue provides additional
information in `invalidParams` field.

And Problem now does not include the trace id and error code
into the `datail` field because it makes it hard to render
on client side. The detail field should contain useful
explanation of what happened and how to fix this but not
a debug information.

Also, removed all dots from commands output and description.


PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-17349

modules/cli/src/main/java/org/apache/ignite/cli/core/style/component/MessageUiComponent.java
```diff
@@ -0,0 +1,95 @@
...
+
+package org.apache.ignite.cli.core.style.component;
+
+import static org.apache.ignite.cli.core.style.AnsiStringSupport.ansi;
+
+import org.apache.ignite.cli.core.style.element.UiElement;
+import org.apache.ignite.cli.core.style.element.UiString;
+
+/**
+ * UI component that represent a message.
```
valepakh, line 26:
```suggestion
 * UI component that represents a message.
```

modules/cli/src/main/java/org/apache/ignite/cli/core/style/component/QuestionUiComponent.java
```diff
@@ -0,0 +1,72 @@
...
+
+package org.apache.ignite.cli.core.style.component;
+
+import static org.apache.ignite.cli.core.style.AnsiStringSupport.ansi;
+
+import org.apache.ignite.cli.core.style.element.UiElement;
+import org.apache.ignite.cli.core.style.element.UiString;
+
+/**
+ * UI component that represent a question.
```
valepakh, line 26:
```suggestion
 * UI component that represents a question.
```

modules/cli/src/main/java/org/apache/ignite/cli/deprecated/ui/ProgressBar.java
```diff
@@ -107,7 +107,7 @@ private String render() {
 
         var completedPart = ((double) curr / (double) max);
 
-        // Space reserved for '||Done!'
+        // Space reserved for '||Done'
         var reservedSpace = 7;
```
valepakh, line 111: Should the resevred space be 6 now that the exclamation mark is removed?

modules/cli/src/main/java/org/apache/ignite/cli/commands/cluster/status/ClusterStatusReplSubCommand.java
```diff
@@ -58,7 +60,7 @@ public void run() {
         } else if (session.isConnectedToNode()) {
             inputUrl = session.nodeUrl();
         } else {
-            spec.commandLine().getErr().println("You are not connected to node. Run 'connect' command or use '--cluster-url' option.");
+            spec.commandLine().getErr().println(CONNECT_OR_USE_CLUSTER_URL_MESSAGE.render());
```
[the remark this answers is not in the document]
PakhomovAlexander: Agreed, but I think this is out of the scope of this ticket.

modules/cli/src/main/java/org/apache/ignite/cli/deprecated/ui/ProgressBar.java
PakhomovAlexander: I set it to 6 first, and it broke ProgressBarTest. This number makes no differense on what the user sees.

modules/cli/src/main/java/org/apache/ignite/cli/core/style/component/MessageUiComponent.java
```diff
@@ -0,0 +1,95 @@
...
+        this.messageUiElements = messageUiElements;
+        this.hint = hint;
+        this.hintUiElements = hintUiElements;
+    }
+
+    @Override
+    public String render() {
+        return ansi(
+                UiString.format(message, messageUiElements)
+                        + (UiString.format(hint, hintUiElements) == null ? ""
```
valepakh, line 53: Maybe it's worth to extract this to a variable? It's hard to follow the logic.

modules/cli/src/main/java/org/apache/ignite/cli/deprecated/spec/NodeCommandSpec.java
```diff
@@ -60,45 +63,22 @@ public class NodeCommandSpec {
...
     @Singleton
     public static class StartNodeCommandSpec extends BaseCommand implements Callable<Integer> {
 
+        /** Consistent id, which will be used by new node. */
+        @Parameters(paramLabel = "name", description = "Name of the new node")
+        public String nodeName;
         /** Loader for Ignite distributive paths. */
         @Inject
         private CliPathsConfigLoader cliPathsCfgLdr;
-
```
valepakh, line 70: Looks like some of the line breaks got removed during the merge and ConfigOptions class is again at the bottom.

modules/configuration-api/src/main/java/org/apache/ignite/configuration/validation/ConfigurationValidationException.java
```diff
@@ -54,4 +55,9 @@ public ConfigurationValidationException(List<ValidationIssue> issues) {
     public List<ValidationIssue> getIssues() {
         return issues;
     }
+
+    private static String createMessageFromIssues(List<ValidationIssue> issues) {
```
sk0x50, line 59: I don't think that this change is fully correct. It just skips error message from a validation issue.
In case of catching a such exception you need to add additional code to get all information. I mean that the following code will not be enough:
```
    try {
        ....
    }
    catch (Exception e) {
        log.("unexpected exception", e);
```
The code above will print: ```Validation did not pass for keys: Key 1, Key 2``. The problem itself will be lost.

PR #988, opened 2022-08-09
IGNITE-17400 Basic C++ binary tuple support

ademakov:
This provides partial C++ support for development of storage engines. Another PR will provide missing pieces under IGNITE-17401 ticket. After that it will not be absolutely complete again. E.g. Windows support will be absent for a while. Some changes might be needed to re-use it for client development. And so on. Additional tickets will be created and PRs provided on demand.


modules/platforms/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
+#
```
ibessonov, line 1: Two general questions:
- Is this a common practice to use `*.txt` files for CMake? I never used it, but it looks weird to me.
- `platforms` modules should probably contain code related to thin clients. I'd rather have a different module for this code to be honest.

modules/platforms/cpp/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+                        <goals>
+                            <goal>copy-resources</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>${project.build.directory}/classes/</outputDirectory>
+                            <resources>
+                                <resource>
+                                    <directory>${project.build.directory}/cpp/</directory>
+                                    <includes>
+                                        <include>**/*.so</include>
```
ibessonov, line 105: Two sets of questions here as well:
- Do we only support `.so` binaries? Why not include `.dll`? Has this been discussed?
- Usually binaries go to specific folders for specific architectures. Is this the case here? Has the list of supported architectures been discussed?

modules/platforms/cpp/schema/BinaryTupleBuilder.cpp
```diff
@@ -0,0 +1,178 @@
...
+        case DATA_TYPE::UUID:
+        case DATA_TYPE::DATE:
+        case DATA_TYPE::TIME:
+        case DATA_TYPE::DATETIME:
+        case DATA_TYPE::TIMESTAMP:
+        case DATA_TYPE::BITMASK:
+        case DATA_TYPE::NUMBER:
+        case DATA_TYPE::DECIMAL:
+        default:
+            throw std::logic_error("Unsupported type " + std::to_string(static_cast<int>(type)));
```
ibessonov, line 174: Do we need a TODO here? Why are they unsupported?

modules/platforms/cpp/schema/BinaryTupleBuilder.h
```diff
@@ -0,0 +1,378 @@
...
+
+    /**
+     * @brief starts a new tuple
+     */
+    void start() noexcept;
+
+    /**
+     * @brief assigns a null value for the next element
+     */
+    void claim(std::nullopt_t /*null*/) noexcept {
```
ibessonov, line 81: Right now the convention is very non-intuitive: sometimes you have implementation in a proper `cpp` file and sometimes it is in the `.h` file. Is there a rule here? I don't get it.

modules/platforms/cpp/schema/BinaryTupleBuilder.h
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    /**
+     * @brief computes required binary size for a given value
+     *
+     * @param value Actual element value.
+     * @return Required size.
+     */
+    static SizeT sizeOfInt64(std::int64_t value) noexcept {
+        if (std::uint64_t(value) <= UINT16_MAX) {
```
ibessonov, line 265: How does this work? In Java code I see:
```        
if (Short.MIN_VALUE <= value && value <= Short.MAX_VALUE) {
    return appendShort((short) value);
}
```
But here I see a conversion to unsigned type. I really doubt that `std::uint64_t` returns an absolute value, or does it? If it does, then I'd rather see an explicit `abs` call. If it doesn't, then this code doesn't seem safe, or even correct at all.

modules/platforms/cpp/schema/BinaryTupleParser.h
```diff
@@ -0,0 +1,161 @@
...
+class BinaryTupleParser {
+    BytesView binaryTuple; /**< The binary tuple to parse. */
+
+    IntT elementCount; /**< Total number of elements. */
+
+    IntT elementIndex; /**< Index of the next element to parse. */
+
+    bool hasNullmap; /**< Flag that indicates if the tuple contains a nullmap. */
+
+    unsigned int entrySize; /**< Size of an offset table entry. */
```
ibessonov, line 39: Why not `SizeT`?

modules/platforms/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+
+if (ENABLE_ADDRESS_SANITIZER)
+    add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-fsanitize=address>)
+    add_link_options($<$<COMPILE_LANGUAGE:CXX>:-fsanitize=address>)
+endif()
+
+if (ENABLE_UB_SANITIZER)
+    add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-fsanitize=undefined>)
+    add_link_options($<$<COMPILE_LANGUAGE:CXX>:-fsanitize=undefined>)
+endif()
```
isapego, line 38: Seems GNU-specific. Again, if we are going to use this code in C++ client we need more cross-platform solution.

modules/platforms/cpp/common/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+
+project(ignite-common)
+
+set(TARGET ${PROJECT_NAME})
+
+set(HEADERS Types.h)
+
+add_library(${TARGET} INTERFACE)
+
+#target_sources(${TARGET} INTERFACE ${HEADERS})
```
isapego, line 26: Should be removed, I guess?

modules/platforms/cpp/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<!--
+    POM file.
+-->
```
isapego, line 22: Possibly, will break compilation on Windows if that's something we'd like to support. But I guess, that's OK for now.

modules/platforms/cpp/CMakeLists.txt
ademakov: * It's weird for me too but this is how CMake works.
* Probably you are right. I'd appreciate any advise on module placement. If client developers would re-use this module, what is the best location for them?

modules/platforms/cpp/pom.xml
ademakov: Support of other platforms will be done later. Let's start with something. Ignite-3 still has time to mature.

isapego (changes requested):
We should definitely at least decide on how we move with C++ code base, do we want it to be a common for storage/client and then what are our coding guidelines. I'm not sure we can continue until these are decided.

modules/platforms/cpp/pom.xml
ademakov: This will be addressed under IGNITE-17401 ticket.

modules/platforms/cpp/schema/BinaryTupleBuilder.cpp
ademakov: This will be addressed under IGNITE-17401 ticket.

modules/platforms/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+cmake_minimum_required(VERSION 3.10)
+project(Ignite.C++ VERSION 3 LANGUAGES CXX)
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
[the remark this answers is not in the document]
ademakov: It's possible that we will have some portability issues. Actually right now I don't even have a windows environment for development. Let's start with something and then improve upon it step by step.

modules/platforms/cpp/CMakeLists.txt
ademakov: Understood.

modules/platforms/cpp/CMakeLists.txt
ibessonov: We should ask them probably, I don't know the answer

modules/platforms/cpp/schema/BinaryTupleBuilder.h
ademakov: Usually small functions go to headers for take advantage of inlining. Function templates go to headers even if they are not  small. In some cases you might want to use a header-only library for very basic stuff.

Here some code is copied from another (private) repo and not yet fully converted to a uniform style. I'm going to fix a couple  of such issues now.

modules/platforms/cpp/schema/BinaryTupleParser.h
ademakov: Thanks for the suggestion.

modules/platforms/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+cmake_minimum_required(VERSION 3.10)
+project(Ignite.C++ VERSION 3 LANGUAGES CXX)
```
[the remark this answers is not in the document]
ademakov: Yes, I think we need to discuss this.

modules/platforms/cpp/CMakeLists.txt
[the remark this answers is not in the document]
isapego: I agree, but we need to make some decisions first so we won't need to re-write too many code in future. One thing is a carefully isolated platform-dependent code and another thing is a compiler and platform specific calls all over the code base.

modules/platforms/cpp/CMakeLists.txt
isapego: Well, maybe it's Ok to store all C++ code in a single place for easier re-use. We just need to make sure that we have decent modularity here and that we can easily configure compilation so that we won't need to wait to compile all the code when we only work on our specific client or store code.

modules/platforms/cpp/schema/BinaryTupleBuilder.h
ademakov: Fixed, thanks.

modules/platforms/cpp/schema/BinaryTupleParser.cpp
```diff
@@ -0,0 +1,203 @@
...
+    TupleView tuple;
+    tuple.reserve(num);
+    while (elementIndex < num) {
+        tuple.emplace_back(getNext());
+    }
+
+    return tuple;
+}
+
+KeyView BinaryTupleParser::parseKey(IntT num) {
```
korlov42, line 107: Why do we need this? I consider BinaryTuple as a representation of a very abstract tuple in a binary from.

modules/platforms/cpp/schema/BinaryTupleBuilder.cpp
isapego: Let's add a TODO comment for that then.

modules/platforms/cpp/schema/BinaryTupleSchema.cpp
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "BinaryTupleSchema.h"
+#include <cassert>
+
+namespace ignite {
+
+} // namespace ignite
```
isapego, line 23: Empty file.

modules/platforms/cpp/schema/DataType.cpp
```diff
@@ -0,0 +1,79 @@
...
+ * limitations under the License.
+ */
+
+#include "DataType.h"
+
+#include <stdexcept>
+
+namespace ignite {
+
+bool isFixedSizeType(DATA_TYPE t) {
```
isapego, line 24: Here and below: `constexpr`?

isapego (approved):

modules/platforms/cpp/schema/BinaryTupleBuilder.h
```diff
@@ -0,0 +1,397 @@
...
+ * 1. Initialize the builder with the @ref start call.
+ * 2. Supply all elements with one or more @ref claim calls in the order that
+ *    corresponds to the tuple schema.
+ * 3. Determine tuple layout with the @ref layout call.
+ * 4. Supply all elements again with one or more @ref append calls in the same
+ *    order with the same values.
+ * 5. Finally, the resulting binary tuple is obtained with the @ref build call.
+ */
+class BinaryTupleBuilder {
+    IntT elementCount; /**< Total number of elements. */
```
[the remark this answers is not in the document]
ademakov: Done.

modules/platforms/cpp/common/CMakeLists.txt
ademakov: Done.

modules/platforms/cpp/schema/BinaryTupleSchema.cpp
ademakov: Removed.

modules/platforms/cpp/schema/BinaryTupleBuilder.h
```diff
@@ -0,0 +1,397 @@
...
+    void claim(const BinaryTupleSchema &schema, const std::vector<std::optional<BytesT>> &tuple) noexcept {
+        for (IntT i = 0; i < schema.numElements(); i++) {
+            claim(schema.getElement(i).dataType, tuple[i]);
+        }
+    }
+
+    /**
+     * @brief performs binary tuple layout
+     */
+    void layout();
```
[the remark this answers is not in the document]
ademakov: I think I'll remove it in IGNITE-17401 by using the same technique as in Java version.

modules/platforms/cpp/schema/BinaryTupleParser.cpp
ademakov: This is a micro-optimization for primary key tuples that cannot have nulls by definition. This feature is not strictly needed but some C++ folks might want it in certain scenarios. Anyway it is isolated and does not harm anything.

PR #990, opened 2022-08-09
IGNITE-17434 Fix partially correct previous implementation of sql implicit transactions.

zstan:
…licit transactions


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlSynchronousApiTest.java
```diff
@@ -181,13 +184,100 @@ public void dml() {
...
+
+        CompletableFuture<?> fut1 = runAsync(() -> {
+            Session ses = sql.createSession();
+
+            StringBuilder insert = new StringBuilder("INSERT INTO TEST VALUES");
+
+            boolean first = true;
+
+            for (int j : arr1) {
+                insert.append(String.format(first ? " (%d, %d)" : ", (%d, %d)", j, j));
```
ygerzhedovich, line 224: maybe StringJoiner will be convinient for the case

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlAsynchronousApiTest.java
```diff
@@ -313,10 +314,12 @@ public void checkTransactionsWithDml() throws Exception {
 
         checkDml(2 * ROW_COUNT, ses, "DELETE FROM TEST WHERE VAL0 >= 0");
 
-        assertEquals(ROW_COUNT + 1 + 1 + 1 + 1, txManagerInternal.finished() - txPrevCnt);
+        assertEquals(ROW_COUNT + 1 + 1 + 1 + 1 + 1, txManagerInternal.finished() - txPrevCnt);
 
         var states = (Map<UUID, TxState>) IgniteTestUtils.getFieldValue(txManagerInternal, TxManagerImpl.class, "states");
 
+        assertTrue(waitForCondition(() -> states.values().stream().noneMatch(e -> e.equals(TxState.PENDING)), 5_000));
```
ygerzhedovich, line 321: I don't understand why a transaction could be in a PENDING STATE here. Could you please elaborate this? Also even if so true the next assert looks strange.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlSynchronousApiTest.java
```diff
@@ -181,13 +184,100 @@ public void dml() {
 
         var states = (Map<UUID, TxState>) IgniteTestUtils.getFieldValue(txManagerInternal, TxManagerImpl.class, "states");
 
-        states.forEach((k, v) -> assertNotSame(v, TxState.PENDING));
+        assertTrue(waitForCondition(() -> states.values().stream().noneMatch(e -> e.equals(TxState.PENDING)), 5_000));
```
ygerzhedovich, line 187: the same question as above

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlSynchronousApiTest.java
```diff
@@ -181,13 +184,100 @@ public void dml() {
...
 
         checkDml(ROW_COUNT, ses, "UPDATE TEST SET VAL0 = VAL0 + ?", 1);
 
         checkDml(ROW_COUNT, ses, "DELETE FROM TEST WHERE VAL0 >= 0");
     }
 
+    /**
+     * Execute concurrent insertion batch of equal pk`s and different val`s.
+     * Check sum of arithmetic progression as a confirmation that only one batch has been commited.
```
ygerzhedovich, line 196:
```suggestion
     * Check sum of arithmetic progression as a confirmation that only one batch has been committed.
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlSynchronousApiTest.java
```diff
@@ -181,13 +184,100 @@ public void dml() {
...
+
+            boolean first = true;
+
+            for (int i = arr1.length - 1; i == 0; --i) {
+                insert.append(String.format(first ? " (%d, %d)" : ", (%d, %d)", arr1[i], arr2[i]));
+
+                first = false;
+            }
+
+            ses.execute(null, insert.toString()).close();
```
ygerzhedovich, line 249: java.util.concurrent.ExecutionException: java.lang.Exception: org.apache.ignite.sql.SqlException: IGN-SQL-3 TraceId:281f2b85-c768-4eea-851d-c03b31d1a3c4 IGN-SQL-3 TraceId:281f2b85-c768-4eea-851d-c03b31d1a3c4 Failed to parse query

ygerzhedovich: Let's don't ignore an exceptions, we must chech that exception is expected

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -356,12 +362,16 @@ class DistributedQueryManager {
 
         private volatile Long rootFragmentId = null;
 
-        private InternalTransaction transaction;
+        private @Nullable InternalTransaction transaction;
```
ygerzhedovich, line 365: Maybe I don't fully understand, but in my picture transaction should present always, it could be implicite, explicite or transaction sended in fragment to remote node.

ygerzhedovich: My bad - it should be implemented in separate ticket

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDmlTest.java
```diff
@@ -86,6 +88,56 @@ public void mergeOpChangePrimaryKey() {
...
+    public void batchWithConflictShouldBeRejectedEntirely() {
+        sql("CREATE TABLE test (id int primary key, val int)");
+
+        sql("INSERT INTO test values (1, 1)");
+
+        assertQuery("SELECT count(*) FROM test")
+                .returns(1L)
+                .check();
+
+        assertThrows(
```
ygerzhedovich, line 101: let's check that exception has right SQL code

ygerzhedovich: Also I think it should be SqlException instead of IgniteException

korlov42: fixed

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDmlTest.java
```diff
@@ -86,6 +88,56 @@ public void mergeOpChangePrimaryKey() {
...
+
+        values.add(values.get(0)); // add conflict entry from the first chunk
+
+        sql("CREATE TABLE test (id int primary key, val int default 1)");
+
+        String insertStatement = "INSERT INTO test (id) VALUES " + values.stream()
+                .map(Object::toString)
+                .collect(Collectors.joining("), (", "(", ")"));
+
+        assertThrows(
```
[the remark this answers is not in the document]
korlov42: fixed

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlAsynchronousApiTest.java
```diff
@@ -592,13 +591,15 @@ public void batchIncomplete() {
             }
         }
 
-        Throwable ex = assertThrowsWithCause(
-                () -> ses.executeBatchAsync(null, "INSERT INTO TEST VALUES (?, ?)", args).join(),
-                SqlBatchException.class
+        CompletionException ex = assertThrows(
```
ygerzhedovich, line 594: Should we change CompletionException to SqlExeption?

korlov42: This is not an option because `assertThrows` verifies the topmost exception only. `CompletionException` in this particular case is caused by `join()` within the closure rather then by internals of sql engine

PR #994, opened 2022-08-09
IGNITE-17500 Fix minor issues in RocksDB-based partition storages

sashapolo:
* "Destroy" method returns a future that completes after the data is flushed
* Storage is closed after "destroy" is called
* FlushOptions are cached in RocksDbTableStorage so that they do not get GC-ed

https://issues.apache.org/jira/browse/IGNITE-17500

### DIFF
--- a/modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
+++ b/modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
@@ -62,11 +62,11 @@ import org.rocksdb.WriteOptions;
 
 /**
  * Multi-versioned partition storage implementation based on RocksDB. Stored data has the following format:
- * <pre><code>
- * | rowId (16 bytes, BE) |</code></pre>
+ * <pre>{@code
+ * | rowId (16 bytes, BE) |}</pre>
  * or
- * <pre><code>
- * | rowId (16 bytes, BE) | timestamp (16 bytes, DESC) |</code></pre>
+ * <pre>{@code
+ * | rowId (16 bytes, BE) | timestamp (16 bytes, DESC) |}</pre>
  * depending on transaction status. Pending transactions' data doesn't have a timestamp assigned.
  *
  * <p/>BE means Big Endian, meaning that lexicographical bytes order matches a natural order of partitions.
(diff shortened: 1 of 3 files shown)


modules/storage-rocksdb/src/test/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorageTest.java
```diff
@@ -246,7 +248,8 @@ void storageAdvertisesItIsPersistent() {
         assertThat(storage.isVolatile(), is(false));
     }
 
-    private IgniteBiTuple<TestKey, TestValue> unwrap(BinaryRow binaryRow) {
+    @Nullable
+    private static IgniteBiTuple<TestKey, TestValue> unwrap(@Nullable BinaryRow binaryRow) {
```
tkalkirill, line 252:
```suggestion
    private static @Nullable IgniteBiTuple<TestKey, TestValue> unwrap(@Nullable BinaryRow binaryRow) {
```

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -193,11 +193,9 @@ public <V> V runConsistently(WriteClosure<V> closure) throws StorageException {
     /** {@inheritDoc} */
     @Override
     public CompletableFuture<Void> flush() {
-        CompletableFuture<Void> flushFuture = new CompletableFuture<>();
-
-        CompletableFuture<Void> oldFuture = flushFuturesByAppliedIndex.put(lastAppliedIndex, flushFuture);
-
-        assert oldFuture == null;
+        CompletableFuture<Void> flushFuture = flushFuturesByAppliedIndex.computeIfAbsent(
```
tkalkirill, line 196: Why did you remove the assert?

modules/storage-rocksdb/src/test/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorageTest.java
sashapolo: Why?

tkalkirill: This is just a suggestion: null can be a return value, not a modifier.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
sashapolo: Because `destroy` could have been called after a Raft command has been applied. This means that we will call `flush` twice: once for the command and once for the destroy, but they will have the same applied index

tkalkirill: ok

tkalkirill (approved):

SammyVimes (approved):
LGTM

PR #1001, opened 2022-08-11
IGNITE-17469 cli profile commands unification

valepakh:
https://issues.apache.org/jira/browse/IGNITE-17469

This unifies the `cli config profile` commands to use parameters rather than options for profile name and uses profile options for all node/cluster commands which use `--node-url` or `--cluster-endpoint-url` to take the default values from the profile specified by `--profile` option.


modules/cli/src/main/java/org/apache/ignite/cli/commands/cliconfig/CliConfigGetReplCommand.java
```diff
@@ -24,27 +24,23 @@
...
-    private String profileName;
-
     @Inject
     private CliConfigGetCall call;
 
     @Override
     public Integer call() {
         return CallExecutionPipeline.builder(call)
-                .inputProvider(() -> new CliConfigGetCallInput(key, profileName))
+                .inputProvider(() -> new CliConfigGetCallInput(key, null))
```
PakhomovAlexander, line 43: I would suggest to create one more constructor instead of passing the null.

modules/cli/src/main/java/org/apache/ignite/cli/commands/cliconfig/CliConfigShowReplCommand.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+@Command(name = "show", description = "Shows currently activated config")
+public class CliConfigShowReplCommand extends BaseCommand implements Callable<Integer> {
+    @Inject
+    private CliConfigShowCall call;
+
+    @Override
+    public Integer call() {
+        return CallExecutionPipeline.builder(call)
+                .inputProvider(() -> new StringCallInput(null))
```
PakhomovAlexander, line 40: Tha same for the constructor.

modules/cli/src/main/java/org/apache/ignite/cli/commands/cliconfig/profile/CliConfigProfileShowCommand.java
```diff
@@ -28,11 +28,11 @@
 /**
  * Show current profile command.
  */
-@CommandLine.Command(name = "show", description = "Show current default profile.")
-public class CliConfigShowProfileCommand extends BaseCommand implements Callable<Integer> {
+@CommandLine.Command(name = "show", description = "Shows current default profile")
```
PakhomovAlexander, line 31: "current default" sounds strange, maybe just "current"?

modules/cli/src/main/java/org/apache/ignite/cli/commands/cluster/ClusterUrlOptions.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Inject
+    private ConfigManagerProvider configManagerProvider;
+
+    /**
+     * Gets cluster URL from either the command line or from the config with specified or default profile.
+     *
+     * @return cluster URL
+     */
+    public String get() {
```
PakhomovAlexander, line 56: getClusterUrl

modules/cli/src/main/java/org/apache/ignite/cli/commands/node/NodeUrlOptions.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Inject
+    private ConfigManagerProvider configManagerProvider;
+
+    /**
+     * Gets node URL from either the command line or from the config with specified or default profile.
+     *
+     * @return node URL
+     */
+    public String get() {
```
PakhomovAlexander, line 56: getNodeUrl

modules/cli/src/test/java/org/apache/ignite/cli/commands/ProfileMixinTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+    /**
+     * Cluster URL from test profile in integration_tests.ini.
+     */
+    private static final String PROFILE_URL = "http://localhost:10301";
+
+    /**
+     * Cluster URL override from command line.
+     */
+    private static final String TEST_URL = "http://localhost:10302";
```
PakhomovAlexander, line 76: URL_FROM_CMD

modules/cli/src/test/java/org/apache/ignite/cli/commands/ProfileMixinTest.java
```diff
@@ -0,0 +1,244 @@
...
+public class ProfileMixinTest {
+    /**
+     * Cluster URL from default profile in integration_tests.ini.
+     */
+    private static final String DEFAULT_URL = "http://localhost:10300";
+
+    /**
+     * Cluster URL from test profile in integration_tests.ini.
+     */
+    private static final String PROFILE_URL = "http://localhost:10301";
```
PakhomovAlexander, line 71: URL_FROM_PROFILE

modules/cli/src/test/java/org/apache/ignite/cli/commands/cliconfig/CliConfigProfileListCommandTest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.jupiter.api.Test;
+
+class CliConfigProfileListCommandTest extends CliConfigCommandTestBase {
+    @Override
+    protected Class<?> getCommandClass() {
+        return CliConfigProfileListCommand.class;
+    }
+
+    @Test
+    public void testWithInternalSection() {
```
PakhomovAlexander, line 32: Could you add two more test cases?  
- for the single profile configured
- if there is no profile configured

PakhomovAlexander (changes requested):

modules/cli/src/main/java/org/apache/ignite/cli/call/cliconfig/profile/CliConfigProfileCreateCallInput.java
```diff
@@ -50,35 +53,4 @@ public String getCopyFrom() {
     public boolean isActivate() {
         return activate;
     }
-
-    /**
-     * Builder of {@link CliConfigCreateProfileCallInput}.
-     */
-    public static class CliConfigCreateProfileCallInputBuilder {
```
PakhomovAlexander, line 57: I am quite confused with the fact that you removed the builder only for this input? Why dont remove all builders for CallInput? I think it is ok no to have a builder for the input that has only one field but if the class has more that one field it is quite useful to hava a builder. Bulder helps not to confuse parameter names.

modules/cli/src/main/java/org/apache/ignite/cli/call/cliconfig/profile/CliConfigProfileListCall.java
```diff
@@ -0,0 +1,45 @@
...
+@Singleton
+public class CliConfigProfileListCall implements Call<EmptyCallInput, String> {
+
+    @Inject
+    private ConfigManagerProvider provider;
+
+    @Override
+    public CallOutput<String> execute(EmptyCallInput input) {
+        Collection<String> profileNames = provider.get().getProfileNames();
+        String result = profileNames.stream().collect(Collectors.joining(System.lineSeparator()));
```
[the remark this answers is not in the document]
valepakh: I decided to not mark the active profile when I found an interesting use-case - pipe the output of this command to change some option in all profiles. We might want to add the mark only when the output is not piped though.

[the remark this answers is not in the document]
YuryYudin: We're not supposed to pipe  anything in the REPL mode. It would make sense to still mark it there.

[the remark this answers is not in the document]
valepakh: I added the mark for current profile in case output is not redirected.

modules/cli/src/main/java/org/apache/ignite/cli/commands/decorators/ProfileListDecorator.java
```diff
@@ -0,0 +1,62 @@
...
+            return () -> data.getProfileNames().stream().collect(Collectors.joining(System.lineSeparator()));
+        }
+    }
+
+    private TerminalOutput decorateCurrentProfileName(ProfileList data) {
+        String currentProfileName = data.getCurrentProfileName();
+        return () -> data.getProfileNames().stream()
+                .map(p -> {
+                    if (p.equals(currentProfileName)) {
+                        return ansi(Style.BOLD.mark("* " + p));
```
PakhomovAlexander, line 46: In case we merge the UI Components it is better to use them here.

modules/cli/src/test/java/org/apache/ignite/cli/commands/cliconfig/CliConfigProfileListCommandTest.java
```diff
@@ -0,0 +1,67 @@
...
+                this::assertErrOutputIsEmpty
+        );
+    }
+
+    @Test
+    public void testSingleProfile() {
+        configManagerProvider.configManager = new IniConfigManager(TestConfigManagerHelper.createClusterUrlNonDefault());
+        execute();
+
+        String expectedResult = "default" + System.lineSeparator();
```
PakhomovAlexander, line 49: Could you rename `default` to `nonDefault`?

PakhomovAlexander: As far as I can see we set up non default profile but check the default one.

valepakh: It's a cluster URL which is non-default, not the profile. I'll create a separate single profile ini file.

SammyVimes (approved):
LGTM

PR #1006, opened 2022-08-14
IGNITE-17428 WIP

alievmirza:


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1294,37 +1295,17 @@ public CompletableFuture<TableImpl> tableAsyncInternal(UUID id, boolean checkCon
...
-
-                return completedFuture(true);
-            }
-
-            @Override
-            public void remove(@NotNull Throwable e) {
-                getTblFut.completeExceptionally(e);
+        IgniteTriConsumer<Long, Map<UUID, TableImpl>, Throwable> tablesListener = (token, tables, th) -> {
+            if (th == null) {
+                tablesByIdVv.get(token).thenRun(() -> getTblFut.complete(tablesByIdVv.latest().get(id)));
```
denis-chudov, line 1300:
```suggestion
                getTblFut.complete(tables.get(id));
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1294,37 +1295,27 @@ public CompletableFuture<TableImpl> tableAsyncInternal(UUID id, boolean checkCon
...
-
-                return completedFuture(true);
-            }
-
-            @Override
-            public void remove(@NotNull Throwable e) {
-                getTblFut.completeExceptionally(e);
+        IgniteTriConsumer<Long, Map<UUID, TableImpl>, Throwable> tablesListener = (token, tables, th) -> {
+            if (th == null) {
+                getTblFut.complete(tables.get(id));
```
vldpyatkov, line 1300: You need to check that the completed change is applied to the table.
I mean, you should be sure that tables.get(id) != null.

alievmirza: good catch, thank you

PR #1008, opened 2022-08-15
IGNITE-17528 Use parameter instead of option in config show commands

valepakh:
https://issues.apache.org/jira/browse/IGNITE-17528

This drops the `--selector` option in `config show` commands and allow similar UX as `config update` commands so that the user can write `node config show network` rather than `node config show --selector network`.


PakhomovAlexander (approved):
LGTM

SammyVimes (approved):
LGTM

PR #1010, opened 2022-08-16
IGNITE-17529 Override CLI config file path via environment variable

valepakh:
https://issues.apache.org/jira/browse/IGNITE-17529

Allow changing config file path, which is now determined by the XDG spec.


SammyVimes (approved):
LGTM

PR #1016, opened 2022-08-16
IGNITE-17196 In-memory partition rebalance

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-17196

### DIFF
--- a/modules/affinity/src/main/java/org/apache/ignite/internal/affinity/AffinityUtils.java
+++ b/modules/affinity/src/main/java/org/apache/ignite/internal/affinity/AffinityUtils.java
@@ -22,7 +22,7 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.function.Function;
+import java.util.function.IntFunction;
 import org.apache.ignite.network.ClusterNode;
 import org.jetbrains.annotations.NotNull;
 
@@ -34,7 +34,7 @@ public class AffinityUtils {
      * Calculates affinity assignments.
      *
      * @param partitions Partitions count.
-     * @param replicas   Replicas count.
+     * @param replicas Replicas count.
      * @param aggregator Function that creates a collection for the partition assignments.
      * @return List nodes by partition.
      */
@@ -42,7 +42,7 @@ public class AffinityUtils {
             @NotNull Collection<ClusterNode> baselineNodes,
             int partitions,
             int replicas,
-            Function<Integer, T> aggregator
+            IntFunction<T> aggregator
     ) {
         return RendezvousAffinityFunction.assignPartitions(
                 baselineNodes,
@@ -58,7 +58,7 @@ public class AffinityUtils {
      * Calculates affinity assignments.
      *
      * @param partitions Partitions count.
-     * @param replicas   Replicas count.
+     * @param replicas Replicas count.
      * @return List nodes by partition.
      */
     public static List<List<ClusterNode>> calculateAssignments(
--- a/modules/affinity/src/main/java/org/apache/ignite/internal/affinity/AffinityUtils.java
+++ b/modules/affinity/src/main/java/org/apache/ignite/internal/affinity/AffinityUtils.java
@@ -17,8 +17,12 @@
 
 package org.apache.ignite.internal.affinity;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.function.Function;
 import org.apache.ignite.network.ClusterNode;
 import org.jetbrains.annotations.NotNull;
 
@@ -31,6 +35,30 @@ public class AffinityUtils {
      *
      * @param partitions Partitions count.
      * @param replicas   Replicas count.
+     * @param aggregator Function that creates a collection for the partition assignments.
+     * @return List nodes by partition.
+     */
+    public static <T extends Collection<ClusterNode>> List<T> calculateAssignments(
+            @NotNull Collection<ClusterNode> baselineNodes,
+            int partitions,
+            int replicas,
+            Function<Integer, T> aggregator
+    ) {
+        return RendezvousAffinityFunction.assignPartitions(
+                baselineNodes,
+                partitions,
+                replicas,
+                false,
+                null,
+                aggregator
+        );
+    }
+
+    /**
+     * Calculates affinity assignments.
+     *
+     * @param partitions Partitions count.
+     * @param replicas   Replicas count.
      * @return List nodes by partition.
      */
     public static List<List<ClusterNode>> calculateAssignments(
@@ -38,12 +66,35 @@ public class AffinityUtils {
             int partitions,
             int replicas
     ) {
-        return RendezvousAffinityFunction.assignPartitions(
+        return calculateAssignments(
                 baselineNodes,
                 partitions,
                 replicas,
+                ArrayList::new
+        );
+    }
+
+    /**
+     * Calculates affinity assignments for single partition.
+     *
+     * @param baselineNodes Nodes.
+     * @param partition Partition id.
+     * @param replicas Replicas count.
+     * @return List of nodes.
+     */
+    public static Set<ClusterNode> calculateAssignmentForPartition(
+            Collection<ClusterNode> baselineNodes,
+            int partition,
+            int replicas
+    ) {
+        return RendezvousAffinityFunction.assignPartition(
+                partition,
+                new ArrayList<>(baselineNodes),
+                replicas,
+                null,
                 false,
-                null
+                null,
+                HashSet::new
         );
     }
 }
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
@@ -123,12 +123,12 @@ public class ByteUtils {
      * @param bytes Byte array.
      * @return Object.
      */
-    public static Object fromBytes(byte[] bytes) {
+    public static <T> T fromBytes(byte[] bytes) {
         try (
                 var bis = new ByteArrayInputStream(bytes);
                 var in = new ObjectInputStream(bis)
         ) {
-            return in.readObject();
+            return (T) in.readObject();
         } catch (IOException | ClassNotFoundException e) {
             throw new IgniteInternalException("Could not deserialize an object", e);
         }
--- a/modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/MetaStorageServiceImpl.java
+++ b/modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/MetaStorageServiceImpl.java
@@ -532,6 +532,7 @@ public class MetaStorageServiceImpl implements MetaStorageService {
             /**
              * Constructor.
              *
+             * @param nodeId Node id.
              * @param cursor Watch event cursor.
              * @param lsnr   The listener which receives and handles watch updates.
              */
--- a/modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/MetaStorageServiceImpl.java
+++ b/modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/MetaStorageServiceImpl.java
@@ -91,16 +91,21 @@ public class MetaStorageServiceImpl implements MetaStorageService {
     /** Local node id. */
     private final String localNodeId;
 
+    /** Local node name. */
+    private final String localNodeName;
+
     /**
      * Constructor.
      *
      * @param metaStorageRaftGrpSvc Meta storage raft group service.
      * @param localNodeId           Local node id.
+     * @param localNodeName         Local node name.
      */
-    public MetaStorageServiceImpl(RaftGroupService metaStorageRaftGrpSvc, String localNodeId) {
+    public MetaStorageServiceImpl(RaftGroupService metaStorageRaftGrpSvc, String localNodeId, String localNodeName) {
         this.metaStorageRaftGrpSvc = metaStorageRaftGrpSvc;
         this.watchProcessor = new WatchProcessor();
         this.localNodeId = localNodeId;
+        this.localNodeName = localNodeName;
     }
 
     /** {@inheritDoc} */
@@ -268,7 +273,6 @@ public class MetaStorageServiceImpl implements MetaStorageService {
 
         watchRes.thenAccept(
                 watchId -> watchProcessor.addWatch(
-                        localNodeId,
                         watchId,
                         new CursorImpl<>(metaStorageRaftGrpSvc, watchRes, MetaStorageServiceImpl::watchResponse),
                         lsnr
@@ -300,7 +304,6 @@ public class MetaStorageServiceImpl implements MetaStorageService {
 
         watchRes.thenAccept(
                 watchId -> watchProcessor.addWatch(
-                        localNodeId,
                         watchId,
                         new CursorImpl<>(metaStorageRaftGrpSvc, watchRes, MetaStorageServiceImpl::watchResponse),
                         lsnr
@@ -471,13 +474,12 @@ public class MetaStorageServiceImpl implements MetaStorageService {
          * Starts exclusive thread per watch that implement watch pulling logic and calls {@link WatchListener#onUpdate(WatchEvent)}} or
          * {@link WatchListener#onError(Throwable)}.
          *
-         * @param nodeId  Node id.
          * @param watchId Watch id.
          * @param cursor  Watch Cursor.
          * @param lsnr    The listener which receives and handles watch updates.
          */
-        private void addWatch(String nodeId, IgniteUuid watchId, CursorImpl<WatchEvent> cursor, WatchListener lsnr) {
-            Watcher watcher = new Watcher(nodeId, cursor, lsnr);
+        private void addWatch(IgniteUuid watchId, CursorImpl<WatchEvent> cursor, WatchListener lsnr) {
+            Watcher watcher = new Watcher(cursor, lsnr);
 
             watchers.put(watchId, watcher);
 
@@ -532,12 +534,11 @@ public class MetaStorageServiceImpl implements MetaStorageService {
             /**
              * Constructor.
              *
-             * @param nodeId Node id.
              * @param cursor Watch event cursor.
              * @param lsnr   The listener which receives and handles watch updates.
              */
-            Watcher(String nodeId, Cursor<WatchEvent> cursor, WatchListener lsnr) {
-                setName(nodeId);
+            Watcher(Cursor<WatchEvent> cursor, WatchListener lsnr) {
+                setName("ms-watcher-" + localNodeName);
                 this.cursor = cursor;
                 this.lsnr = lsnr;
             }
--- a/modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/MetaStorageServiceImpl.java
+++ b/modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/MetaStorageServiceImpl.java
@@ -268,6 +268,7 @@ public class MetaStorageServiceImpl implements MetaStorageService {
 
         watchRes.thenAccept(
                 watchId -> watchProcessor.addWatch(
+                        localNodeId,
                         watchId,
                         new CursorImpl<>(metaStorageRaftGrpSvc, watchRes, MetaStorageServiceImpl::watchResponse),
                         lsnr
@@ -299,6 +300,7 @@ public class MetaStorageServiceImpl implements MetaStorageService {
 
         watchRes.thenAccept(
                 watchId -> watchProcessor.addWatch(
+                        localNodeId,
                         watchId,
                         new CursorImpl<>(metaStorageRaftGrpSvc, watchRes, MetaStorageServiceImpl::watchResponse),
                         lsnr
@@ -469,12 +471,13 @@ public class MetaStorageServiceImpl implements MetaStorageService {
          * Starts exclusive thread per watch that implement watch pulling logic and calls {@link WatchListener#onUpdate(WatchEvent)}} or
          * {@link WatchListener#onError(Throwable)}.
          *
+         * @param nodeId  Node id.
          * @param watchId Watch id.
          * @param cursor  Watch Cursor.
          * @param lsnr    The listener which receives and handles watch updates.
          */
-        private void addWatch(IgniteUuid watchId, CursorImpl<WatchEvent> cursor, WatchListener lsnr) {
-            Watcher watcher = new Watcher(cursor, lsnr);
+        private void addWatch(String nodeId, IgniteUuid watchId, CursorImpl<WatchEvent> cursor, WatchListener lsnr) {
+            Watcher watcher = new Watcher(nodeId, cursor, lsnr);
 
             watchers.put(watchId, watcher);
 
@@ -532,7 +535,8 @@ public class MetaStorageServiceImpl implements MetaStorageService {
              * @param cursor Watch event cursor.
              * @param lsnr   The listener which receives and handles watch updates.
              */
-            Watcher(Cursor<WatchEvent> cursor, WatchListener lsnr) {
+            Watcher(String nodeId, Cursor<WatchEvent> cursor, WatchListener lsnr) {
+                setName(nodeId);
                 this.cursor = cursor;
                 this.lsnr = lsnr;
             }
@@ -554,6 +558,8 @@ public class MetaStorageServiceImpl implements MetaStorageService {
                                 watchEvt = watchEvtsIter.next();
                             } catch (Throwable e) {
                                 lsnr.onError(e);
+
+                                throw e;
                             }
 
                             assert watchEvt != null;
--- a/modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/SimpleCondition.java
+++ b/modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/SimpleCondition.java
@@ -77,6 +77,8 @@ public final class SimpleCondition implements Condition {
          * @throws IllegalStateException In the case when the condition is already defined.
          */
         public SimpleCondition eq(long rev) {
+            assert rev > 0 : "Revision must be positive.";
+
             validate(type());
 
             type(ConditionType.REV_EQUAL);
@@ -95,6 +97,8 @@ public final class SimpleCondition implements Condition {
          * @throws IllegalStateException In the case when the condition is already defined.
          */
         public SimpleCondition ne(long rev) {
+            assert rev > 0 : "Revision must be positive.";
+
             validate(type());
 
             type(ConditionType.REV_NOT_EQUAL);
@@ -113,6 +117,8 @@ public final class SimpleCondition implements Condition {
          * @throws IllegalStateException In the case when the condition is already defined.
          */
         public SimpleCondition gt(long rev) {
+            assert rev > 0 : "Revision must be positive.";
+
             validate(type());
 
             type(ConditionType.REV_GREATER);
@@ -131,6 +137,8 @@ public final class SimpleCondition implements Condition {
          * @throws IllegalStateException In the case when the condition is already defined.
          */
         public SimpleCondition ge(long rev) {
+            assert rev > 0 : "Revision must be positive.";
+
             validate(type());
 
             type(ConditionType.REV_GREATER_OR_EQUAL);
@@ -149,6 +157,8 @@ public final class SimpleCondition implements Condition {
          * @throws IllegalStateException In the case when the condition is already defined.
          */
         public SimpleCondition lt(long rev) {
+            assert rev > 0 : "Revision must be positive.";
+
             validate(type());
 
             type(ConditionType.REV_LESS);
@@ -167,6 +177,8 @@ public final class SimpleCondition implements Condition {
          * @throws IllegalStateException In the case when the condition is already defined.
          */
         public SimpleCondition le(long rev) {
+            assert rev > 0 : "Revision must be positive.";
+
             validate(type());
 
             type(ConditionType.REV_LESS_OR_EQUAL);
--- a/modules/network-annotation-processor/src/main/java/org/apache/ignite/internal/network/processor/messages/MessageImplGenerator.java
+++ b/modules/network-annotation-processor/src/main/java/org/apache/ignite/internal/network/processor/messages/MessageImplGenerator.java
@@ -49,6 +49,8 @@ import javax.tools.Diagnostic;
 import org.apache.ignite.internal.network.processor.MessageClass;
 import org.apache.ignite.internal.network.processor.MessageGroupWrapper;
 import org.apache.ignite.internal.network.processor.TypeUtils;
+import org.apache.ignite.internal.tostring.IgniteToStringInclude;
+import org.apache.ignite.internal.tostring.S;
 import org.apache.ignite.network.NetworkMessage;
 import org.apache.ignite.network.annotations.Marshallable;
 
@@ -105,6 +107,7 @@ public class MessageImplGenerator {
             String getterName = getter.getSimpleName().toString();
 
             FieldSpec.Builder fieldBuilder = FieldSpec.builder(getterReturnType, getterName)
+                    .addAnnotation(IgniteToStringInclude.class)
                     .addModifiers(Modifier.PRIVATE);
 
             boolean isMarshallable = getter.getAnnotation(Marshallable.class) != null;
@@ -162,6 +165,16 @@ public class MessageImplGenerator {
                 .build();
 
         messageImpl.addMethod(groupTypeMethod);
+
+        MethodSpec toStringMethod = MethodSpec.methodBuilder("toString")
+                .addAnnotation(Override.class)
+                .addModifiers(Modifier.PUBLIC)
+                .returns(String.class)
+                .addStatement("return $T.toString($T.class, this)", S.class, messageImplClassName)
+                .build();
+
+        messageImpl.addMethod(toStringMethod);
+
 
         // message type constant and getter
         FieldSpec messageTypeField = FieldSpec.builder(short.class, "TYPE")
--- a/modules/network-annotation-processor/src/main/java/org/apache/ignite/internal/network/processor/messages/MessageImplGenerator.java
+++ b/modules/network-annotation-processor/src/main/java/org/apache/ignite/internal/network/processor/messages/MessageImplGenerator.java
@@ -166,6 +166,7 @@ public class MessageImplGenerator {
 
         messageImpl.addMethod(groupTypeMethod);
 
+        // TODO: https://issues.apache.org/jira/browse/IGNITE-17591
         MethodSpec toStringMethod = MethodSpec.methodBuilder("toString")
                 .addAnnotation(Override.class)
                 .addModifiers(Modifier.PUBLIC)
@@ -174,7 +175,6 @@ public class MessageImplGenerator {
                 .build();
 
         messageImpl.addMethod(toStringMethod);
-
 
         // message type constant and getter
         FieldSpec messageTypeField = FieldSpec.builder(short.class, "TYPE")
--- a/modules/network/src/main/java/org/apache/ignite/internal/network/netty/NettyUtils.java
+++ b/modules/network/src/main/java/org/apache/ignite/internal/network/netty/NettyUtils.java
@@ -20,8 +20,11 @@ package org.apache.ignite.internal.network.netty;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.GenericFutureListener;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
+import org.jetbrains.annotations.Async.Execute;
+import org.jetbrains.annotations.Async.Schedule;
 
 /**
  * Netty utilities.
@@ -38,18 +41,21 @@ public class NettyUtils {
      * @return CompletableFuture.
      */
     public static <T, R, F extends Future<R>> CompletableFuture<T> toCompletableFuture(
-            F nettyFuture,
+            @Schedule F nettyFuture,
             Function<F, T> mapper
     ) {
         var fut = new CompletableFuture<T>();
 
-        nettyFuture.addListener((F future) -> {
-            if (future.isSuccess()) {
-                fut.complete(mapper.apply(future));
-            } else if (future.isCancelled()) {
-                fut.cancel(true);
-            } else {
-                fut.completeExceptionally(future.cause());
+        nettyFuture.addListener(new GenericFutureListener<F>() {
+            @Override
+            public void operationComplete(@Execute F future) throws Exception {
+                if (future.isSuccess()) {
+                    fut.complete(mapper.apply(future));
+                } else if (future.isCancelled()) {
+                    fut.cancel(true);
+                } else {
+                    fut.completeExceptionally(future.cause());
+                }
             }
         });
 
--- a/modules/network/src/main/java/org/apache/ignite/internal/network/netty/NettyUtils.java
+++ b/modules/network/src/main/java/org/apache/ignite/internal/network/netty/NettyUtils.java
@@ -20,7 +20,6 @@ package org.apache.ignite.internal.network.netty;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.util.concurrent.Future;
-import io.netty.util.concurrent.GenericFutureListener;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
 import org.jetbrains.annotations.Async.Execute;
@@ -46,16 +45,13 @@ public class NettyUtils {
     ) {
         var fut = new CompletableFuture<T>();
 
-        nettyFuture.addListener(new GenericFutureListener<F>() {
-            @Override
-            public void operationComplete(@Execute F future) throws Exception {
-                if (future.isSuccess()) {
-                    fut.complete(mapper.apply(future));
-                } else if (future.isCancelled()) {
-                    fut.cancel(true);
-                } else {
-                    fut.completeExceptionally(future.cause());
-                }
+        nettyFuture.addListener((@Execute F future) -> {
+            if (future.isSuccess()) {
+                fut.complete(mapper.apply(future));
+            } else if (future.isCancelled()) {
+                fut.cancel(true);
+            } else {
+                fut.completeExceptionally(future.cause());
             }
         });
 
--- a/modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/RaftGroupServiceImpl.java
+++ b/modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/RaftGroupServiceImpl.java
@@ -580,8 +580,8 @@ public class RaftGroupServiceImpl implements RaftGroupService {
                 if (err != null) {
                     if (recoverable(err)) {
                         executor.schedule(() -> {
-//                            LOG.warn("Recoverable error during the request type={} occurred (will be retried on the randomly selected node): ",
-//                                    err, req.getClass().getSimpleName());
+                            LOG.warn("Recoverable error during the request type={} occurred (will be retried on the randomly selected node): ",
+                                    err, req.getClass().getSimpleName());
 
                             sendWithRetry(randomNode(peer), req, stopTime, fut);
 
@@ -608,7 +608,7 @@ public class RaftGroupServiceImpl implements RaftGroupService {
 
                             if (resp0.errorCode() == (RaftError.ENOENT.getNumber())) {
                                 // If changing peers or requesting a leader and something is not found
-                                // probably target peer is doing rebalancing, try another peer
+                                // probably target peer is doing rebalancing, try another peer.
                                 if (req instanceof GetLeaderRequest || req instanceof ChangePeersAsyncRequest) {
                                     targetPeer = randomNode(peer);
                                 }
@@ -695,7 +695,11 @@ public class RaftGroupServiceImpl implements RaftGroupService {
     }
 
     /**
-     * @return Random node.
+     * Returns a random peer. Tries returning peer different from the peer passed as an argument.
+     * If peer is null, just returns a random peer.
+     *
+     * @param peer Previous peer.
+     * @return Random peer.
      */
     private Peer randomNode(Peer peer) {
         List<Peer> peers0 = peers;
@@ -710,12 +714,12 @@ public class RaftGroupServiceImpl implements RaftGroupService {
 
         int retries = 0;
 
-        ThreadLocalRandom current = current();
+        ThreadLocalRandom random = current();
 
         int newIdx = 0;
 
         while (retries < 5) {
-            newIdx = current.nextInt(peers0.size());
+            newIdx = random.nextInt(peers0.size());
 
             if (newIdx != lastPeerIndex) {
                 break;
--- a/modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/RaftGroupServiceImpl.java
+++ b/modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/RaftGroupServiceImpl.java
@@ -513,7 +513,7 @@ public class RaftGroupServiceImpl implements RaftGroupService {
 
         ActionRequest req = factory.actionRequest().command(cmd).groupId(groupId).readOnlySafe(true).build();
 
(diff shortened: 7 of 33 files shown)


modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
```diff
@@ -123,12 +123,12 @@ public static byte[] toBytes(Object obj) {
...
      * @return Object.
      */
-    public static Object fromBytes(byte[] bytes) {
+    public static <T> T fromBytes(byte[] bytes) {
         try (
                 var bis = new ByteArrayInputStream(bytes);
                 var in = new ObjectInputStream(bis)
         ) {
-            return in.readObject();
+            return (T) in.readObject();
```
sashapolo, line 131: This is an unsafe cast, I would prefer to leave the current approach with explicit casting

modules/network/src/main/java/org/apache/ignite/internal/network/netty/NettyUtils.java
```diff
@@ -38,18 +41,21 @@ public class NettyUtils {
...
-        nettyFuture.addListener((F future) -> {
-            if (future.isSuccess()) {
-                fut.complete(mapper.apply(future));
-            } else if (future.isCancelled()) {
-                fut.cancel(true);
-            } else {
-                fut.completeExceptionally(future.cause());
+        nettyFuture.addListener(new GenericFutureListener<F>() {
+            @Override
+            public void operationComplete(@Execute F future) throws Exception {
```
sashapolo, line 51: Is it impossible to add annotations to lambda parameters?

modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
SammyVimes: Well it's less code and the runtime implications are the same, are they not?

modules/network/src/main/java/org/apache/ignite/internal/network/netty/NettyUtils.java
SammyVimes: TIL that it is possible

modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
sashapolo: They are not exactly the same, because with the explicit cast the ClassCastException may be thrown earlier

ibessonov: I'm against such code, but technically, both approaches are absolutely the same here. This method will return Object without cast, because no type information is available to it. It's all just sugar.

sashapolo: ok, I did a little bit of thinking and this code looks safe and is more convenient to use. Please ignore my comment

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -606,6 +708,32 @@ private void updateAssignmentInternal(ConfigurationNotificationEvent<byte[]> ass
...
+     * @return A future that will hold the quantity of data nodes.
+     */
+    private CompletableFuture<Long> queryDataNodesCount(UUID tblId, int partId, List<ClusterNode> partAssignments) {
+        HasDataRequestBuilder requestBuilder = TABLE_MESSAGES_FACTORY.hasDataRequest().tableId(tblId).partitionId(partId);
+
+        //noinspection unchecked
+        CompletableFuture<Boolean>[] requestFutures = partAssignments.stream().map(node -> {
+            HasDataRequest request = requestBuilder.build();
+
+            return raftMgr.messagingService().invoke(node, request, TimeUnit.SECONDS.toMillis(3)).thenApply(response -> {
```
ibessonov, line 726: Hard-coded timeout, dude

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -606,6 +708,32 @@ private void updateAssignmentInternal(ConfigurationNotificationEvent<byte[]> ass
...
 
+    /**
+     * Calculates the quantity of the data nodes for the partition of the table.
+     *
+     * @param tblId Table id.
+     * @param partId Partition id.
+     * @param partAssignments Partition assignments.
+     * @return A future that will hold the quantity of data nodes.
+     */
+    private CompletableFuture<Long> queryDataNodesCount(UUID tblId, int partId, List<ClusterNode> partAssignments) {
```
ibessonov, line 719: `Long`. Really?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -801,6 +929,12 @@ private void dropTableLocally(long causalityToken, String name, UUID tblId, List
         }
     }
 
+    private List<ClusterNode> calculateAssignments(TableConfiguration tableCfg, int partNum) {
+        Integer partitions = tableCfg.partitions().value();
```
ibessonov, line 933: Quick node, it's better when you use "tableCfg.value()" and then access its properties. It's faster and it doesn't involve boxing.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -801,6 +929,12 @@ private void dropTableLocally(long causalityToken, String name, UUID tblId, List
...
+        Integer replicas = tableCfg.replicas().value();
+        return AffinityUtils.calculateAssignments(baselineMgr.nodes(), partitions, replicas).get(partNum);
```
ibessonov, line 935: Oof, this smells so badly. If we only need assignments for a single table, why do we calculate them for potentially hundreds or thousands irrelevant partitions?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -801,6 +929,12 @@ private void dropTableLocally(long causalityToken, String name, UUID tblId, List
         }
     }
 
+    private List<ClusterNode> calculateAssignments(TableConfiguration tableCfg, int partNum) {
```
ibessonov, line 932: Usually it's `partId` rather then `partNum`, or maybe I'm wrong, I don't know

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1562,6 +1697,40 @@ public void onError(@NotNull Throwable e) {
...
+            @Override
+            public boolean onUpdate(@NotNull WatchEvent evt) {
+                ByteArray key = evt.entryEvent().newEntry().key();
+
+                int partitionNumber = extractPartitionNumber(key);
+                UUID tblId = extractTableId(key, ASSIGNMENTS_SWITCH_REDUCE_PREFIX);
+                String partitionId = partitionRaftGroupName(tblId, partitionNumber);
+
+                TableImpl tbl = tablesByIdVv.latest().get(tblId);
+                ExtendedTableConfiguration tblCfg = (ExtendedTableConfiguration) tablesCfg.tables().get(tbl.name());
```
ibessonov, line 1711: What's the point of casting it to extended tale configuration if you don't use any of the extension properties?

ibessonov, line 1711: I wonder if there's a race possible with configuration update. Why getting the latest value is ok?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
```diff
@@ -252,92 +294,189 @@ private void scheduleChangePeers(List<PeerId> peers, long term) {
...
-                            plannedPartAssignmentsKey(partId),
-                            pendingPartAssignmentsKey(partId),
-                            stablePartAssignmentsKey(partId))).get();
+                            plannedPartAssignmentsKey,
+                            pendingPartAssignmentsKey,
+                            stablePartAssignmentsKey,
+                            switchReduceKey,
+                            switchAppendKey
+                    )
+            ).get();
```
ibessonov, line 311: This deserves a TODO. Another synchronous wait. You're not the author, right?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
```diff
@@ -252,92 +294,189 @@ private void scheduleChangePeers(List<PeerId> peers, long term) {
...
-    }
+            // For further addition
+            List<ClusterNode> calculatedSwitchAppend = union(retrievedSwitchAppend, reducedNodes);
+            calculatedSwitchAppend = subtract(calculatedSwitchAppend, addedNodes);
+            calculatedSwitchAppend = intersect(calculatedAssignments, calculatedSwitchAppend);
 
-    private static List<ClusterNode> resolveClusterNodes(
-            List<PeerId> peers, byte[] pendingAssignments, byte[] stableAssignments) {
-        Map<NetworkAddress, ClusterNode> resolveRegistry = new HashMap<>();
+            var calculatedPendingReduction = subtract(stable, retrievedSwitchReduce);
```
ibessonov, line 341: What's up with all these "var"? We shouldn't use them here

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
```diff
@@ -252,92 +294,189 @@ private void scheduleChangePeers(List<PeerId> peers, long term) {
...
+                successCase = ops(
+                        put(stablePartAssignmentsKey, stableByteArray),
+                        put(pendingPartAssignmentsKey, reductionByteArray),
+                        put(switchReduceKey, switchReduceByteArray),
+                        put(switchAppendKey, switchAppendByteArray)
+                ).yield(SWITCH_REDUCE_SUCCESS);
+                failCase = ops().yield(SWITCH_REDUCE_FAIL);
             } else {
-                throw new IgniteInternalException("Can't find appropriate cluster node for raft group peer: " + p);
+                Condition con5;
```
ibessonov, line 396: Variable names look odd

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
```diff
@@ -252,92 +294,189 @@ private void scheduleChangePeers(List<PeerId> peers, long term) {
...
+
+                    failCase = ops().yield(FINISH_REBALANCE_FAIL);
+                }
+
+                retryPreconditions = and(retryPreconditions, con5);
             }
-        }
 
-        return resolvedNodes;
+            int res = metaStorageMgr.invoke(If.iif(retryPreconditions, successCase, failCase)).get().getAsInt();
```
ibessonov, line 423: Another synchronous wait. No timeout, no comments, nothing

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -157,6 +171,12 @@ public class RebalanceUtil {
     /** Key prefix for stable assignments. */
     public static final String STABLE_ASSIGNMENTS_PREFIX = "assignments.stable.";
 
+    /** Key prefix for switch reduce assignments. */
+    public static final String ASSIGNMENTS_SWITCH_REDUCE_PREFIX = "assignments.switch.reduce.";
```
ibessonov, line 175: I guess it's not my business, but these names are too long. We will pollute metastorage with key names, it's horrible. I'm not laming you, of course

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -202,13 +222,46 @@ public static ByteArray stablePartAssignmentsKey(String partId) {
...
 
     /**
-     * Extract table id from pending key of partition.
+     * Key that is needed for the rebalance algorithm.
+     *
+     * @param partId Unique identifier of a partition.
+     * @return Key for a partition.
+     * @see <a href="https://github.com/apache/ignite-3/blob/main/modules/table/tech-notes/rebalance.md">Rebalnce documentation</a>
+     */
+    public static ByteArray switchReduceKey(String partId) {
```
ibessonov, line 231: Is this a partition id or an id of corresponding raft group?

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -202,13 +222,46 @@ public static ByteArray stablePartAssignmentsKey(String partId) {
...
      *
      * @param key Key.
+     * @param prefix Key prefix.
      * @return Table id.
      */
     public static UUID extractTableId(ByteArray key, String prefix) {
-        var strKey = key.toString();
+        String strKey = key.toString();
 
         return UUID.fromString(strKey.substring(prefix.length(), strKey.indexOf("_part_")));
```
ibessonov, line 266: Again, it has nothing to do with your code, but don't UUIDs have a fixed length? What the hell is going on with these classes

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -202,13 +222,46 @@ public static ByteArray stablePartAssignmentsKey(String partId) {
...
+    /**
+     * Extract table id from a metastorage key of partition.
      *
      * @param key Key.
+     * @param prefix Key prefix.
      * @return Table id.
      */
     public static UUID extractTableId(ByteArray key, String prefix) {
-        var strKey = key.toString();
+        String strKey = key.toString();
```
ibessonov, line 264: No encoding here

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -235,4 +288,209 @@ public static boolean recoverable(Throwable t) {
         // As long as we don't have a general failure handler, we assume that all errors are recoverable.
         return true;
     }
+
+    /**
+     * Starts the process of removing peer from raft group if that peer has in-memory storage or if its
+     * storage was cleared.
```
ibessonov, line 294: "has been", maybe?

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -235,4 +288,209 @@ public static boolean recoverable(Throwable t) {
...
+     *
+     * @param partId Partition's raft group id.
+     * @param clusterNode Cluster node to be removed from peers.
+     * @param metaStorageMgr MetaStorage manager.
+     * @return Completable future that signifies the completion of this operation.
+     */
+    public static CompletableFuture<Void> startPeerRemoval(String partId, ClusterNode clusterNode, MetaStorageManager metaStorageMgr) {
```
ibessonov, line 301: Last reminder about "partId", I hope you'll rename it everywhere

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -235,4 +288,209 @@ public static boolean recoverable(Throwable t) {
...
+        ByteArray key = switchReduceKey(partId);
+
+        return metaStorageMgr.get(key)
+                .thenCompose(retrievedAssignmentsSwitchReduce -> {
+                    byte[] prevValue = retrievedAssignmentsSwitchReduce.value();
+
+                    boolean prevValueEmpty = true;
```
ibessonov, line 308: Why can't it be one huge "if" with `prevValue == null` condition? Is it really simpler with a bunch of smaller ifs and extra variables?

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -235,4 +288,209 @@ public static boolean recoverable(Throwable t) {
...
+     * @param partId Partition's raft group id..
+     * @param event Assignments switch reduce change event.
+     * @return Completable future that signifies the completion of this operation.
+     */
+    public static CompletableFuture<Void> handleReduceChanged(MetaStorageManager metaStorageMgr, Collection<ClusterNode> baselineNodes,
+            int partitions, int replicas, int partNum, String partId, WatchEvent event) {
+        Entry entry = event.entryEvent().newEntry();
+        byte[] eventData = entry.value();
+
+        List<ClusterNode> assignments = AffinityUtils.calculateAssignments(baselineNodes, partitions, replicas).get(partNum);
```
ibessonov, line 361: Same here. Assignments calculation is a computationally complicated process and it might use a lot of memory. There has to be a workaround here.

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -235,4 +288,209 @@ public static boolean recoverable(Throwable t) {
...
+        List<ClusterNode> switchReduce = (List<ClusterNode>) ByteUtils.fromBytes(eventData);
```
ibessonov, line 362: Every usage of ByteUtils is a red flag for me. That's just a reminder for the future, it's very slow and produces a lot of boilerplate data. Assignments in Ignite 2.x are stored much more efficiently.

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -235,4 +288,209 @@ public static boolean recoverable(Throwable t) {
...
+        byte[] assignmentsByteArray = ByteUtils.toBytes(assignments);
+
+        if (switchReduce.isEmpty()) {
+            return CompletableFuture.completedFuture(null);
+        }
+
+        ByteArray changeTriggerKey = partChangeTriggerKey(partId);
+        byte[] rev = ByteUtils.longToBytes(event.entryEvent().newEntry().revision());
+
+        // if ((notExists(changeTriggerKey) || value(changeTriggerKey) < revision) && (notExists(pendingKey) && notExists(stableKey)) {
```
ibessonov, line 378: What's this?

ibessonov: BTW, I'm not validating the logic behind all these computation, otherwise it would take me two more days

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -235,4 +288,209 @@ public static boolean recoverable(Throwable t) {
...
+        //     put(pendingKey, pending)
+        //     put(stableKey, assignments)
+        //     put(changeTriggerKey, revision)
+        // } else if ((notExists(changeTriggerKey) || value(changeTriggerKey) < revision) && (notExists(pendingKey))) {
+        //     put(pendingKey, pending)
+        //     put(changeTriggerKey, revision)
+        // }
+
+        If iif = If.iif(
+                        and(
```
ibessonov, line 388: Padding look too big, is this ok?

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -235,4 +288,209 @@ public static boolean recoverable(Throwable t) {
...
+        // if ((notExists(changeTriggerKey) || value(changeTriggerKey) < revision) && (notExists(pendingKey) && notExists(stableKey)) {
+        //     put(pendingKey, pending)
+        //     put(stableKey, assignments)
+        //     put(changeTriggerKey, revision)
+        // } else if ((notExists(changeTriggerKey) || value(changeTriggerKey) < revision) && (notExists(pendingKey))) {
+        //     put(pendingKey, pending)
+        //     put(changeTriggerKey, revision)
+        // }
+
+        If iif = If.iif(
```
ibessonov, line 387: Variable name isn't the best. And why don't you use static import for `iif`?

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -235,4 +288,209 @@ public static boolean recoverable(Throwable t) {
...
+                        and(
+                                or(notExists(changeTriggerKey), value(changeTriggerKey).lt(rev)),
+                                and(notExists(pendingKey), (notExists(stablePartAssignmentsKey(partId))))
+                        ),
+                        Operations.ops(
```
ibessonov, line 392: ... or a static import for `ops`

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -235,4 +288,209 @@ public static boolean recoverable(Throwable t) {
...
+                        If.iif(
+                            and(
+                                    or(notExists(changeTriggerKey), value(changeTriggerKey).lt(rev)),
+                                    notExists(pendingKey)
+                            ),
+                            Operations.ops(
+                                    put(pendingKey, pendingByteArray),
+                                    put(changeTriggerKey, rev)
+                            ).yield(),
+                            ops().yield()
```
ibessonov, line 406: Hm, what does this do?

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -235,4 +288,209 @@ public static boolean recoverable(Throwable t) {
...
+    }
+
+    /**
+     * Removes nodes from collection of nodes.
+     *
+     * @param minuend Collection to remove nodes from.
+     * @param subtrahend Collection of nodes to be removed.
+     * @return Result of the subtraction.
+     */
+    public static List<ClusterNode> subtract(Collection<ClusterNode> minuend, Collection<ClusterNode> subtrahend) {
```
ibessonov, line 471: Had to google it. These are real words, damn!

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -235,4 +288,209 @@ public static boolean recoverable(Throwable t) {
...
+    }
+
+    /**
+     * Adds nodes to the collection of nodes.
+     *
+     * @param op1 First operand.
+     * @param op2 Second operand.
+     * @return Result of the addition.
+     */
+    public static List<ClusterNode> union(Collection<ClusterNode> op1, Collection<ClusterNode> op2) {
```
ibessonov, line 482: Why do you have collections as parameters?

ibessonov: How slow can this be btw? How big are these collections?
I know that "ususally" contains is invoked on a smaller collection. Should we apply such optimization here?

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -235,4 +288,209 @@ public static boolean recoverable(Throwable t) {
...
+    }
+
+    /**
+     * Returns an intersection of two collections of nodes.
+     *
+     * @param op1 First operand.
+     * @param op2 Second operand.
+     * @return Result of the intersection.
+     */
+    public static List<ClusterNode> intersect(Collection<ClusterNode> op1, Collection<ClusterNode> op2) {
```
ibessonov, line 493: Same here. Keep in mind that `contains` on lists is slow. That's one of reasons why I'm asking about parameter types. They may hide performance issues.

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/SimpleCondition.java
```diff
@@ -77,6 +77,8 @@ public long revision() {
          * @throws IllegalStateException In the case when the condition is already defined.
          */
         public SimpleCondition eq(long rev) {
+            assert rev > 0 : "Revision must be positive.";
```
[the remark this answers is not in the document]
SammyVimes: 1) It's basically the same check as on the metastorage's server side, see `RevisionCondition`'s constructor. 
2) What do you mean? Something like `Requires#requireNonNull`?

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/MetaStorageServiceImpl.java
```diff
@@ -529,10 +532,12 @@ private final class Watcher extends Thread {
             /**
              * Constructor.
              *
+             * @param nodeId Node id.
              * @param cursor Watch event cursor.
              * @param lsnr   The listener which receives and handles watch updates.
              */
-            Watcher(Cursor<WatchEvent> cursor, WatchListener lsnr) {
+            Watcher(String nodeId, Cursor<WatchEvent> cursor, WatchListener lsnr) {
+                setName(nodeId);
```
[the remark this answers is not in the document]
SammyVimes: How about `ms-watcher-${nodeName}`?

modules/network-annotation-processor/src/main/java/org/apache/ignite/internal/network/processor/messages/MessageImplGenerator.java
```diff
@@ -105,6 +107,7 @@ public TypeSpec generateMessageImpl(MessageClass message, TypeSpec builderInterf
             String getterName = getter.getSimpleName().toString();
 
             FieldSpec.Builder fieldBuilder = FieldSpec.builder(getterReturnType, getterName)
+                    .addAnnotation(IgniteToStringInclude.class)
```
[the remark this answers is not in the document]
SammyVimes: Let's add a ticket for this issue

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteInMemoryNodeRestartTest.java
```diff
@@ -0,0 +1,354 @@
...
+
+        return cfg == null ? IgniteStringFormatter.format(NODE_BOOTSTRAP_CFG, port, connectAddr) : cfg;
+    }
+
+    /**
+     * Stop the node with given index.
+     *
+     * @param idx Node index.
+     */
+    private static void stopNode(int idx) {
```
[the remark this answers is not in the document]
SammyVimes: No reason, it's actually a copy-paste from a different test

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -365,6 +380,50 @@ public CompletableFuture<Boolean> notify(@NotNull SchemaEventParameters paramete
...
+
+    /**
+     * Adds a table manager message handler.
+     *
+     * @param messagingService Messaging service.
+     */
+    private void addMessageHandler(MessagingService messagingService) {
+        var messageHandler = new NetworkMessageHandler() {
+            @Override
+            public void onReceived(NetworkMessage message, NetworkAddress senderAddr, @Nullable Long correlationId) {
```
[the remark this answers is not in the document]
SammyVimes: We only have listeners for groups ATM

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteInMemoryNodeRestartTest.java
```diff
@@ -0,0 +1,354 @@
...
+     *
+     * @param idx Node index.
+     * @param cfg Optional configuration string.
+     * @param predefinedPort Predefined port.
+     * @return Configuration string.
+     */
+    private static String configurationString(int idx, @Nullable String cfg, @Nullable Integer predefinedPort) {
+        int port = predefinedPort == null ? DEFAULT_NODE_PORT + idx : predefinedPort;
+        int connectPort = predefinedPort == null ? DEFAULT_NODE_PORT : predefinedPort;
+        String connectAddr = "[\"localhost:" + connectPort + "\"]";
```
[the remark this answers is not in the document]
SammyVimes: Practically the same but looks worse imho, because the semicolon should be escaped
`String connectAddr = "[localhost\":\"" + DEFAULT_NODE_PORT + "]";`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -365,6 +380,50 @@ public CompletableFuture<Boolean> notify(@NotNull SchemaEventParameters paramete
...
+                    assert correlationId != null;
+
+                    HasDataRequest msg = (HasDataRequest) message;
+
+                    UUID tableId = msg.tableId();
+                    int partitionId = msg.partitionId();
+
+                    boolean contains = false;
+
+                    TableImpl table = tablesByIdVv.latest().get(tableId);
```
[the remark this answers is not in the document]
SammyVimes: Network thread

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -365,6 +380,50 @@ public CompletableFuture<Boolean> notify(@NotNull SchemaEventParameters paramete
...
+
+                    if (table != null) {
+                        MvTableStorage storage = table.internalTable().storage();
+
+                        MvPartitionStorage mvPartition = storage.getMvPartition(partitionId);
+
+                        if (mvPartition != null) {
+                            // If applied index of a storage is greater than 0,
+                            // then there is data
+                            contains = mvPartition.lastAppliedIndex() > 0;
```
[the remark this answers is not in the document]
SammyVimes: It works as intended because we can't wait for recovery as recovery sends requests like this :)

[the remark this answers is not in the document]
SammyVimes: Added a comment

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -423,7 +482,7 @@ private CompletableFuture<?> onTableDelete(ConfigurationNotificationEvent<TableV
                     ctx.storageRevision(),
                     ctx.oldValue().name(),
                     ((ExtendedTableView) ctx.oldValue()).id(),
-                    (List<List<ClusterNode>>) ByteUtils.fromBytes(((ExtendedTableView) ctx.oldValue()).assignments())
+                    ByteUtils.fromBytes(((ExtendedTableView) ctx.oldValue()).assignments())
```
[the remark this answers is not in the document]
SammyVimes: But then it will be both embarassing and eye-bleeding

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
SammyVimes: Thanks!

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
SammyVimes: Right, I'll rewrite it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
SammyVimes: Well in this code partId is the id of the partition raft group (its name)

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
SammyVimes: You're right. I usually relies on IDEA to tell me that cast is redundant :(

SammyVimes: We need to calculate the latest target assignments to proceed with rebalance, I don't think something can go wrong here 🤔🤔

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
SammyVimes: You're right

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
SammyVimes: Right, forgot to clean them up

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
SammyVimes: well they're just parts of one big IF statement's condition basically, I don't think that they deserve a proper naming. Each of them have a comment that describes the condition

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
SammyVimes: Added a comment

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: It's possible to change values to something like "a.s.r", wdyt about it?

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: The latter one

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: They do have fixed length, but what do you mean?

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: There is no method that accepts encoding

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: Right

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: Fixed

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: What can I say, we must refactor the hell out of it

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: It's just the comment explaining the logic of the metastorage invoke

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: oops

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: If we could do metastorage operations as code, this would be just an `if`. I don't know how else I should call it. Any other name won't grasp the meaning of it.

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: Right

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: Just nothing

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/SimpleCondition.java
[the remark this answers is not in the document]
ibessonov: 2. Yes, I mean something like that. Is this even possible? I guess asserts are ok

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/MetaStorageServiceImpl.java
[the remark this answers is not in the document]
ibessonov: Looks good to me

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteInMemoryNodeRestartTest.java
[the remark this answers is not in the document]
ibessonov: Now I see, thank you!

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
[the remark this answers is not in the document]
ibessonov: > We only have listeners for groups ATM

Did not expect that...

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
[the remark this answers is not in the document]
ibessonov: Is this a operation fast enough to not stall the network thread?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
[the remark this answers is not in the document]
ibessonov: Is it possible to falsely return "false" here while `lastAppliedIndex` is still 0 and it won't be in a moment?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -552,33 +618,69 @@ private void updateAssignmentInternal(ConfigurationNotificationEvent<byte[]> ass
...
+                                            // with current partition.
+                                            String msg = "Unable to start partition " + partId + ". Majority not available.";
+                                            throw new IgniteInternalException(msg);
+                                        }
+                                    });
+                                } else {
+                                    fut = CompletableFuture.completedFuture(true);
                                 }
+
+                                return fut.thenComposeAsync(startGroup -> {
```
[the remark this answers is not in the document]
ibessonov: Can we use "composeAsync" or something?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
ibessonov: Ok, never mind then

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
ibessonov: We can do it later, because it is unrelated. It's just funny how we compare "writeObject" with native UOS, but then send these big ass strings over the network

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
ibessonov: I mean that calling `indexOf` makes no sense! It can be derived from `prefix.length()`. Author clearly didn't have time to think about performance

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
ibessonov: My bad, sorry.

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
ibessonov: You should comment this comment so that it won't feel like a real commented code

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
ibessonov: Calling it `if` is like writing `int number;` and `String string`, it doesn't explain the purpose of the value. It may be anything like `resultingOperation` or I don't know

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
ibessonov: Oh, it's an `else` branch! That is SO hard to read.
You know what? We can do what Mockito does, in a way. Add methods like `then` and `els` just for sugar.

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/AffinityUtils.java
```diff
@@ -17,15 +17,43 @@
...
+     * @param partitions Partitions count.
+     * @param replicas   Replicas count.
+     * @param aggregator Function that creates a collection for the partition assignments.
+     * @return List nodes by partition.
+     */
+    public static <T extends Collection<ClusterNode>> List<T> calculateAssignments(
+            @NotNull Collection<ClusterNode> baselineNodes,
+            int partitions,
+            int replicas,
+            Function<Integer, T> aggregator
```
ibessonov, line 45: There's an `IntFunction` in Java

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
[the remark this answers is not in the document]
SammyVimes: Well if it is 0 at this moment, then it should be false. We shouldn't wait for recovery, it should be something else, I think

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: I don't like this method at all, extracting table id from the key seems odd

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: I think it's a good idea, yes

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/SimpleCondition.java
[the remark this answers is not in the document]
SammyVimes: Weeeell, there are no methods for value comparison in `Requires`...

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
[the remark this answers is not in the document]
SammyVimes: It MAY be, but it's used all over, so I think we can leave it for the `Future#get` refactoring ticket that I've created

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
SammyVimes: Sure

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/SimpleCondition.java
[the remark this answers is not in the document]
ibessonov: I mean, you can write your own? Anyways, let's leave assertions, I don't really mind it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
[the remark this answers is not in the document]
ibessonov: ok

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
ibessonov: Remember when metastorage was in the design phase and people were thinking "we can store keys effectively, like storing integers as 4 bytes instead of their decimal representation...". Everyone forgot about it. But I digress.
Storing tableId in keys is absolutely fine

ibessonov (approved):

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/RaftGroupServiceImpl.java
```diff
@@ -603,7 +604,17 @@ else if (resp0.errorCode() == RaftError.EBUSY.getNumber() ||
                         resp0.errorCode() == (RaftError.EAGAIN.getNumber()) ||
                         resp0.errorCode() == (RaftError.ENOENT.getNumber())) { // Possibly a node has not been started.
                         executor.schedule(() -> {
-                            sendWithRetry(peer, req, stopTime, fut);
+                            Peer targetPeer = peer;
+
+                            if (resp0.errorCode() == RaftError.ENOENT.getNumber()) {
```
sk0x50, line 609: Does this change belong to the ticket itself? It seems to me, this is a separate issue that should be addressed by a separate ticket.

modules/compute/src/main/java/org/apache/ignite/internal/compute/ComputeMessageTypes.java
```diff
@@ -24,7 +24,7 @@
 /**
  * Message types for the Compute module.
  */
-@MessageGroup(groupName = "ComputeMessages", groupType = 6)
+@MessageGroup(groupType = 6, groupName = "ComputeMessages")
```
sk0x50, line 27: What is the purpose of this change? :)

SammyVimes: groupType is first everywhere else, just a little style change

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/MetaStorageServiceImpl.java
```diff
@@ -533,6 +538,7 @@ private final class Watcher extends Thread {
              * @param lsnr   The listener which receives and handles watch updates.
              */
             Watcher(Cursor<WatchEvent> cursor, WatchListener lsnr) {
+                setName("ms-watcher-" + localNodeName);
```
sk0x50, line 541: It would be better to use `IgniteThread` instead of `Watcher extends Thread` I think. Will file a ticket for this.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/RaftGroupServiceImpl.java
SammyVimes: I don't see how I can test this separately so it will be just a little commit with 3 lines of code and this will also delay the merging of this ticket, so...

modules/metastorage-client/src/main/java/org/apache/ignite/internal/metastorage/client/MetaStorageServiceImpl.java
SammyVimes: I agree, we don't need to extend java's basic threads when we have ignite thread

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/RaftGroupServiceImpl.java
sk0x50: I still insist that separate issues should go under separate tickets regardless their size.

PR #1024, opened 2022-08-18
IGNITE-17090 Show user-friendly message in sql command

PakhomovAlexander:
Previously sql command showed an abstract error message like
`SQL query parsing error` and it was hard to understand what
was wrong. Now it shows the root cause of the issue.

The format of the error message is still being different from
time to time, and we have to think about the error message
formatting in sql-engine.

https://issues.apache.org/jira/browse/IGNITE-17090


modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -272,10 +275,21 @@ public CompletableFuture<JdbcMetaPrimaryKeysResult> primaryKeysMetaAsync(JdbcMet
...
      */
     private StringWriter getWriterWithStackTrace(Throwable t) {
+        String message = null;
+        if (t instanceof CompletionException) {
+            if (t.getCause() instanceof IgniteException
+                    || t.getCause() instanceof CalciteContextException
+                    || t.getCause() instanceof QueryValidationException) {
+                message = t.getCause().getMessage();
+            }
+        }
```
AMashenkov, line 285:
```suggestion
        if (t instanceof CompletionException) {
            message = t.getCause().getMessage();
        }
```

modules/cli/src/main/java/org/apache/ignite/cli/core/exception/handler/SqlExceptionHandler.java
```diff
@@ -39,8 +48,48 @@ public class SqlExceptionHandler implements ExceptionHandler<SQLException> {
 
     public static final String CONNECTION_BROKE_MESSAGE = "Connection error";
 
+    private final Map<Integer, Function<IgniteException, ErrorComponentBuilder>> sqlExceptionMappers = new HashMap<>();
+
+    private final Function<IgniteException, ErrorComponentBuilder> defaultErrorMapper =
+            (e) -> ErrorUiComponent.builder().header("Unrecognized error while process SQL query" + e.getMessage());
```
valepakh, line 54: Maybe use a method similar to invalidQueryErrUiComponent to extract code, traceId and cause message?
Also "process"->"processing"

modules/cli/src/main/java/org/apache/ignite/cli/core/exception/handler/SqlExceptionHandler.java
```diff
@@ -39,8 +48,48 @@ public class SqlExceptionHandler implements ExceptionHandler<SQLException> {
...
+                    .errorCode(clientConnectionException.codeAsString())
+                    .traceId(clientConnectionException.traceId())
+                    .details(ErrorGroup.extractCauseMessage(clientConnectionException.getMessage()));
+        }
+
+        return ErrorUiComponent.builder()
+                .header(CLIENT_CONNECTION_FAILED_MESSAGE)
+                .errorCode(e.codeAsString())
+                .traceId(e.traceId())
+                .details(ErrorGroup.extractCauseMessage(e.getMessage()));
```
valepakh, line 84: Looks like there is a possibility to extract a method for IgniteCliException with custom header.

modules/cli/src/main/java/org/apache/ignite/cli/core/exception/handler/SqlExceptionHandler.java
```diff
@@ -60,13 +109,24 @@ public int handle(ExceptionWriter err, SQLException e) {
...
         }
 
         err.write(errorComponentBuilder.build().render());
         return 1;
     }
 
+    /** Handles IgniteException that has more information like error code and trace id. */
+    public int handleIgniteException(ExceptionWriter err, IgniteException e) {
+        Function<IgniteException, ErrorComponentBuilder> errorComponentBuilder = sqlExceptionMappers.getOrDefault(
+                e.code(), defaultErrorMapper);
```
valepakh, line 122:
```suggestion
        var errorComponentBuilder = sqlExceptionMappers.getOrDefault(e.code(), defaultErrorMapper);
```

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -272,10 +273,17 @@ public CompletableFuture<JdbcMetaPrimaryKeysResult> primaryKeysMetaAsync(JdbcMet
      * @return StringWriter filled with exception.
      */
     private StringWriter getWriterWithStackTrace(Throwable t) {
+        String message = null;
+        if (t instanceof CompletionException) {
```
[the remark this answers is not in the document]
Pochatkin: Done

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -272,10 +273,14 @@ public CompletableFuture<JdbcMetaPrimaryKeysResult> primaryKeysMetaAsync(JdbcMet
...
     private StringWriter getWriterWithStackTrace(Throwable t) {
+        String message = ExceptionUtils.unwrapCause(t).getMessage();
         StringWriter sw = new StringWriter();
         PrintWriter pw = new PrintWriter(sw);
 
-        t.printStackTrace(pw);
+        pw.print(message);
+
+        // check if stacktrace is needed
+        // t.printStackTrace(pw);
```
zstan, line 283: remove it plz

modules/client/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
```diff
@@ -734,16 +733,6 @@ public final void timeout(int timeout) throws SQLException {
...
-            String message = cause.getMessage();
-
-            if (message != null) {
-                if (message.contains("Failed to parse query")) {
-                    return new SQLException("Sql query execution failed.", SqlStateCode.PARSING_EXCEPTION, e);
-                }
-            }
-        }
-        return new SQLException("Internal server error.", SqlStateCode.INTERNAL_ERROR, e);
+        return new SQLException(e.getCause());
```
ygerzhedovich, line 736: it looks bad for me, we lose the code

sk0x50, line 736: It does not seem correct. You should not lose original `CompletionException` because it will lead to the fact that its stack frame will be lost as well. Let's change this line to `return new SQLException(e));`

modules/cli/src/main/java/org/apache/ignite/cli/core/exception/handler/SqlExceptionHandler.java
```diff
@@ -38,9 +47,46 @@ public class SqlExceptionHandler implements ExceptionHandler<SQLException> {
...
+        return ErrorUiComponent.builder()
+                .header(header)
+                .errorCode(e.codeAsString())
+                .traceId(e.traceId())
+                .details(ErrorGroup.extractCauseMessage(e.getMessage()));
+    }
 
     @Override
     public int handle(ExceptionWriter err, SQLException e) {
+        if (e.getCause() instanceof IgniteException) {
```
sk0x50, line 86: In order to properly handle `ExecutionException` and `CompletionException` I would propose:
```
        Throwable unwrappedCause = ExceptionUtils.unwrapCause(e.getCause());
        if (unwrappedCause instanceof IgniteException) {
            return handleIgniteException(err, (IgniteException) unwrappedCause);
        }
```

By the way, do we need to take care of `IgniteCheckedException`'s?

PR #1026, opened 2022-08-19
IGNITE-16860 Add info on get_random_uuid

IgGusev:


docs/_docs/sql-reference/ddl.adoc
```diff
@@ -23,17 +23,17 @@ Creates a new table.
 [source,sql]
 ----
 CREATE TABLE [IF NOT EXISTS] [simpleName | schemaName.simpleName] (tableColumn [, tableColumn]...)
-[COLOCATE [BY] (tableColumn [, tableColumn]...)]
+[COLOCATE [BY] (columnName [, columnName]...)]
 [ENGINE engineName]
 [WITH paramName=paramValue [,paramName=paramValue]...]
-tableColumn = columnName columnType [[NOT] NULL] [DEFAULT defaultValue] [PRIMARY KEY]
+tableColumn = columnName columnType [[NOT] NULL] [DEFAULT defaultValue | gen_random_uuid] [PRIMARY KEY]
```
korlov42, line 29: I would rather add the explanation of a 'defaultValue':
```
tableColumn = columnName columnType [[NOT] NULL] [DEFAULT defaultValue] [PRIMARY KEY]
defaultValue = literal | systemFunctionName
```

docs/_docs/sql-reference/ddl.adoc
```diff
@@ -23,17 +23,17 @@ Creates a new table.
...
 ----
 
 Parameters:
 
 * `tableName` - name of the table.
 * `tableColumn` - name and type of a column to be created in the new table.
-* `DEFAULT` - specifies a default value for the column. Only constant values are accepted.
+* `DEFAULT` - specifies a default value for the column. You can specify a constant value, or use the `gen_random_uuid` function to generate a random value.
```
korlov42, line 36: I would prefer to use more abstract form here, because the list of supported functions may change over the time, and such an approach seems a little bit error prone to me, since we need to find all the place where the particular name of the function is used. 

Let's re-phrase this as "You can specify a constant value or name of the system function to generate value" (or something similar to). Also we need to attach link to "system function" that redirect us to a list of supported functions (is it possible?)

PR #1027, opened 2022-08-19
IGNITE-17444 Introduce MetricExporter interface and service loading.

kgusakov:


modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -828,4 +828,20 @@ public static void inBusyLock(IgniteSpinBusyLock busyLock, Runnable fn) {
...
         }
     }
+
+    /**
+     * Gets simple class name taking care of empty names.
+     *
+     * @param cls Class to get the name for.
+     * @return Simple class name.
+     */
+    public static String getSimpleName(Class<?> cls) {
```
sk0x50, line 838: I cannot find usages of this method. Do we really need it?

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/exporters/PushMetricExporter.java
```diff
@@ -0,0 +1,102 @@
...
+     */
+    public long getPeriod() {
+        return period;
+    }
+
+    /**
+     * A heart of the push exporter.
+     * Inside this method all needed operations to send the metrics outside must be implemented.
+     */
+    public abstract void report();
```
sk0x50, line 101: It will be nice to clearly define assumptions related to exception handling.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/exporters/PushMetricExporter.java
```diff
@@ -0,0 +1,102 @@
...
+                Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory("metrics-exporter", log));
+
+        fut = scheduler.scheduleWithFixedDelay(() -> {
+            try {
+                report();
+            } catch (Throwable th) {
+                log.error("Metrics export error. "
+                        + "This exporter will be stopped [class=" + getClass() + ",name=" + name() + ']', th);
+
+                throw th;
```
sk0x50, line 63: Well, in this case the thread pool will be terminated and metric updates will not be pushed anymore. Anyway, I think we need clarify this moment and update javadoc in accordance with that.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricProvider.java
```diff
@@ -0,0 +1,46 @@
...
+    public MetricProvider(MetricRegistry metricRegistry) {
+        this.metricRegistry = metricRegistry;
+    }
+
+    /**
+     * Returns a map of (metricSetName -> metricSet) pairs with available metrics from {@link MetricRegistry}.
+     *
+     * @return map of metrics
+     */
+    public Map<String, MetricSet> metrics() {
```
agura, line 43: We lose metrics schema version here. It makes any optimization related to compact data packets impossible.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricProvider.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package org.apache.ignite.internal.metrics;
+
+import java.util.Map;
+
+/**
+ * Read-only metrics registry.
+ */
+public class MetricProvider {
```
agura, line 25: Actually, I don't understand the purpose of this entity. It just delegate calls to the `MetricRegistry`. It's also could be usefull to have access to the metrics sources because the class of metrics source could be used as metrics set schema identifier.

agura (changes requested):
Please, take a look to my comments.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricProvider.java
kgusakov: This wrapper is needed only for `MetricExporter` usage to prodive the clear read-only interface for all available metrics.

I think, that exporters shouldn't have direct access to MetricRegistry itself, it will produce the mess in design and give exporters dangerous possibilities to register new sources and etc.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricProvider.java
kgusakov: Added version

sk0x50:
LGTM

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricProvider.java
agura: Okay. But exporters will need to have access to RO collection of metric sources.

PR #1032, opened 2022-08-25
IGNITE-17485: Gradle build support

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-17485

* Added initial gradle project structure.
* Created test fixtures with all shared test classes.
* Extracted all dependencies to toml.
* Temporary disabled flaky tests and Windows broken tests. (IGNITE-17601)


PakhomovAlexander (changes requested):

modules/storage-api/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+    testFixturesImplementation project(':ignite-api')
+    testFixturesImplementation libs.jetbrains.annotations
+    testFixturesImplementation libs.hamcrest.core
+    testFixturesImplementation libs.junit5.api
+    testFixturesImplementation libs.junit5.impl
+}
+
+description = 'ignite-storage-api'
+
+//tasks.register('testsJar', Jar) {
```
YuryYudin, line 57: Do we need these lines here?

Pochatkin: Removed

PR #1033, opened 2022-08-25
IGNITE-17527 Add short aliases for options

valepakh:
https://issues.apache.org/jira/browse/IGNITE-17527

In addition to adding short options, this also adds additional checking when parsing URL options and provide meaningful messages when passing invalid URLs or network addresses.


modules/cli/src/main/java/org/apache/ignite/cli/commands/connect/ConnectCommand.java
```diff
@@ -44,7 +46,7 @@ public class ConnectCommand extends BaseCommand implements Runnable {
     @Override
     public void run() {
         CallExecutionPipeline.builder(connectCall)
-                .inputProvider(() -> new ConnectCallInput(nodeUrl))
+                .inputProvider(() -> new ConnectCallInput(nodeUrl.toString()))
```
kgusakov, line 49: Shouldn't we handle the `nodeUrl==null` case here? Doesn't `NodeUrlMixin` suit well here?

kgusakov (changes requested):

modules/cli/src/main/java/org/apache/ignite/cli/commands/connect/ConnectCommand.java
valepakh: In this command `nodeUrl` is a parameter, not an option, and it can't be null - it will taken from either the parameter or from the config via the `descriptionKey`, and will throw an error in case the config doesn't have this parameter defined.
I added a test which shows that.

kgusakov: Oh, I see, thanks.

kgusakov (approved):

SammyVimes (approved):
LGTM

PR #1034, opened 2022-08-25
IGNITE-17546 Disable autocomplete for questions in REPL

valepakh:
https://issues.apache.org/jira/browse/IGNITE-17546

This disables completion and history saving for question in REPL mode and also puts cursor at the end of the question.


SammyVimes (approved):
LGTM

PR #1035, opened 2022-08-25
IGNITE-17575 Create a Hash Index Storage per partition

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-17575


modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
```diff
@@ -70,38 +85,65 @@ void setUp() {
...
+    void testCreatePartition() {
+        MvPartitionStorage absentStorage = tableStorage.getMvPartition(0);
+
+        assertThat(absentStorage, is(nullValue()));
+
+        MvPartitionStorage partitionStorage = tableStorage.getOrCreateMvPartition(0);
+
+        assertThat(partitionStorage, is(notNullValue()));
+
+        var testData = binaryRow(new TestKey(1, "1"), new TestValue(10, "10"));
```
SammyVimes, line 105: illegal `var`

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
```diff
@@ -70,38 +85,65 @@ void setUp() {
...
+        partitionStorage.runConsistently(() -> {
+            UUID txId = UUID.randomUUID();
 
-        assertThat(tableStorage.getOrCreateMvPartition(partitionId), is(notNullValue()));
+            RowId rowId = partitionStorage.insert(testData, txId);
 
-        assertThat(tableStorage.getMvPartition(partitionId), is(notNullValue()));
+            assertThat(partitionStorage, is(sameInstance(tableStorage.getMvPartition(0))));
+
+            assertThat(unwrap(partitionStorage.read(rowId, txId)), is(equalTo(unwrap(testData))));
```
SammyVimes, line 114: I thought we are only testing the returnal of the partition storage

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/HashIndices.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.concurrent.ConcurrentMap;
+import org.apache.ignite.internal.rocksdb.ColumnFamily;
+import org.apache.ignite.internal.storage.index.HashIndexDescriptor;
+import org.apache.ignite.internal.storage.index.HashIndexStorage;
+import org.apache.ignite.internal.storage.rocksdb.index.RocksDbHashIndexStorage;
+
+class HashIndices {
+    private final HashIndexDescriptor descriptor;
+
+    private final ConcurrentMap<Integer, HashIndexStorage> storages = new ConcurrentHashMap<>();
```
SammyVimes, line 30: I wonder if we actually need concurrent map here. Chances are, we only create indices in one thread and access in other thread strictly after the creation.

SammyVimes (changes requested):

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/HashIndices.java
sashapolo: Do you propose to use a synchronised method instead?

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
sashapolo: This code was copied from other class. Will fix

SammyVimes (approved):
LGTM

PR #1037, opened 2022-08-26
IGNITE-17510 NPE in cluster configuration REST calls

valepakh:
https://issues.apache.org/jira/browse/IGNITE-17510

This improves exceptions handling in REST controllers. In cluster configuration endpoints we produce special exception telling the user that the cluster is not initialized. In all other cases we produce generic problem json.

### DIFF
--- a/modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ComponentNotStartedException.java
+++ b/modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ComponentNotStartedException.java
@@ -15,13 +15,12 @@
  * limitations under the License.
  */
 
-package org.apache.ignite.internal.cluster.management.rest.exception;
+package org.apache.ignite.internal.configuration;
 
 /**
- * Exception that is thrown when the cluster is not initialized.
+ * Exception thrown when component is not started.
  */
-public class ClusterNotInitializedException extends RuntimeException {
-    public ClusterNotInitializedException() {
-        super();
+public class ComponentNotStartedException extends RuntimeException {
+    public ComponentNotStartedException() {
     }
 }
--- a/modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ComponentNotStartedException.java
+++ b/modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ComponentNotStartedException.java
@@ -17,10 +17,14 @@
 
 package org.apache.ignite.internal.configuration;
 
+import org.apache.ignite.lang.ErrorGroups.Common;
+import org.apache.ignite.lang.IgniteInternalException;
+
 /**
  * Exception thrown when component is not started.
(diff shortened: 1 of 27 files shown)


modules/rest-api/src/main/java/org/apache/ignite/internal/rest/exception/handler/AllExceptionHandler.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.ignite.internal.rest.constants.HttpCode;
+import org.apache.ignite.internal.rest.problem.HttpProblemResponse;
+
+/**
+ * Handles {@link Exception} and represents it as an application/problem+json response.
+ * This will catch all unhandled exceptions since all REST endpoints are marked as producing problem json.
+ */
+@Singleton
+@Requires(classes = {Exception.class, ExceptionHandler.class})
+public class AllExceptionHandler implements
```
PakhomovAlexander, line 35: I would suggest logging the exception here because it might be a bug that has to be fixed. And the name of this Exception handler might be JavaExceptionHandler (IgniteExceptionHandler alredy exists).

modules/rest/openapi/openapi.yaml
```diff
@@ -410,7 +442,6 @@ components:
     Problem:
       required:
       - code
-      - detail
       - invalidParams
```
PakhomovAlexander, line 445: Do we reallly need to mark `detail` as not requered? This is not the only field that might be not requered here.

PakhomovAlexander (changes requested):

modules/rest/openapi/openapi.yaml
valepakh: The only reason for now is IDEA inspection which tells me that this is `Nonnull` but I compare it to `null`.

PakhomovAlexander: I would recommend rather to disable an inspection then introducing inconsistency to the public API spec.

valepakh: Probably it's worth investigatin how to generate proper spec, because now it treats all fields as required, which is not the case.

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ComponentNotStartedException.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.configuration;
+
+/**
+ * Exception thrown when component is not started.
+ */
+public class ComponentNotStartedException extends RuntimeException {
```
sk0x50, line 23: In my understanding this exception should be extendsd from `IgniteInternalException`

PR #1040, opened 2022-08-28
IGNITE-17583 Extract BinaryTuple core parts into a separate module

ptupitsyn:
Decouple `BinaryTupleParser`, `BinaryTupleBuilder`, `BinaryTupleReader` from internal schema functionality (`NativeTypeSpec`, etc), move to `ignite-binary-tuple` module.


isapego (approved):

modules/binary-tuple/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+
+        <!-- Test dependencies -->
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
```
sashapolo, line 55: I think these plugins are not needed, I don't see any configuration in this module

ptupitsyn: Good catch. Fixed in IGNITE-17297 branch: https://github.com/ptupitsyn/ignite-3/commit/2376298b767b27397bd53b5d25ab29e3e3a2089b

PR #1047, opened 2022-08-30
IGNITE-17473 Support transactional scan for RW transaction

korlov42:


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/AbstractBasicIntegrationTest.java
```diff
@@ -293,16 +295,22 @@ protected static void checkData(Table table, String[] columnNames, Object[]... t
         }
     }
 
-    protected static List<List<Object>> sql(String sql, Object... args) {
+    protected static  List<List<Object>> sql(String sql, Object... args) {
```
ygerzhedovich, line 298: extra space

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDmlTest.java
```diff
@@ -350,6 +351,33 @@ public void testMergeKeysConflict() {
         assertEquals(Sql.DUPLICATE_KEYS_ERR, ex.code());
     }
 
+    /**
+     * Test verifies that scan is executed within provided transaction.
+     */
+    @Test
+    @Disabled("https://issues.apache.org/jira/browse/IGNITE-17309")
```
ygerzhedovich, line 358: the ticket is resolved, should be there another one?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDmlTest.java
```diff
@@ -350,6 +351,33 @@ public void testMergeKeysConflict() {
...
+        sql("CREATE TABLE test (id int primary key, val int)");
+
+        Transaction tx = CLUSTER_NODES.get(0).transactions().begin();
+
+        sql(tx, "INSERT INTO test VALUES (0, 0)");
+
+        // just inserted row should be visible within the same transaction
+        assertEquals(1, sql(tx, "select * from test").size());
+
+        Transaction anotherTx = CLUSTER_NODES.get(0).transactions().begin();
```
ygerzhedovich, line 369: I suggest to add third tx which started before the first transaction and do the same checks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -372,53 +362,70 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
...
-                                public CompletableFuture<Void> closeAsync() {
-                                    session.touch();
-
-                                    return dataCursor.closeAsync();
-                                }
-                            }
-                    );
+                .thenCompose(sqlNode -> {
+                    BaseQueryContext ctx = BaseQueryContext.builder()
+                            .cancel(new QueryCancel())
```
ygerzhedovich, line 375: It looks strange to use new QueryCncel object here

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/AbstractBasicIntegrationTest.java
korlov42: fixed

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDmlTest.java
korlov42: This ticked is merged to a feature branch, but not main. Well, probably it does make sense to change this link to a link to epic...

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
korlov42: Good catch! Fixed

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDmlTest.java
korlov42: I believe the test should be minimal yet enough to verify the part it's supposed to verify. This particular test is aimed to verify that external transaction is used for scan operation.

ygerzhedovich (approved):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDmlTest.java
```diff
@@ -350,6 +351,33 @@ public void testMergeKeysConflict() {
...
```
zstan, line 369: 1. plz wrap CLUSTER_NODES.get(0).transactions() into different func, as it made for sql()
2. seems it helpful to check all nodes in cluster ? That all nodes not detect uncommited changes ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/TableScanNode.java
```diff
@@ -206,14 +208,15 @@ private void requestNextBatch() {
         }
 
         if (waiting == 0) {
-            waiting = inBufSize;
+            // we must not request rows more than inBufSize
+            waiting = inBufSize - inBuff.size();
```
zstan, line 212: why we can request next batch while **inBuff** is still not empty ?

korlov42: that's how it works right now

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDmlTest.java
korlov42: > plz wrap CLUSTER_NODES.get(0).transactions() into different func

ok, I'll do it next time, because this patch has already been merged

> seems it helpful to check all nodes in cluster ?

why do you think this will be helpful in context of current ticket?

PR #1051, opened 2022-09-02
IGNITE-17593 .NET: Thin 3.0: Implement BinaryTuple serialization

ptupitsyn:
* Port `BinaryTupleBuilder` and `BinaryTupleReader` from Java to .NET.
* Refactor `PooledArrayBufferWriter` to allow variable size prefix.
* Change UUID part order in `BinaryTuple` to be consistent with other implementations (client, BinaryRow).


modules/platforms/dotnet/Apache.Ignite/Internal/Proto/BinaryTuple/BinaryTupleBuilder.cs
```diff
@@ -0,0 +1,407 @@
...
+            }
+
+            _hasNullValues = true;
+
+            int nullIndex = BinaryTupleCommon.NullOffset(_elementIndex);
+            byte nullMask = BinaryTupleCommon.NullMask(_elementIndex);
+
+            var span = _buffer.GetSpan(nullIndex, 1);
+
+            span[0] |= nullMask;
```
isapego, line 108: Why not just 
```suggestion
            _buffer[nullIndex] |= nullMask;
```

isapego (approved):

modules/platforms/dotnet/Apache.Ignite/Internal/Proto/BinaryTuple/BinaryTupleBuilder.cs
ptupitsyn: `_buffer` is not an array, so we can't do that.

PR #1052, opened 2022-09-04
IGNITE-17595 Add test for required constructors in Ignite exceptions

ptupitsyn:
* Add an ArchUnit test for required `(UUID traceId, int code, String message, Throwable cause)` constructor in all classes derived from `IgniteException`.
* Add missing constructors.

### DIFF
--- a/modules/api/pom.xml
+++ b/modules/api/pom.xml
@@ -62,6 +62,16 @@
             <artifactId>hamcrest-optional</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>com.tngtech.archunit</groupId>
+            <artifactId>archunit</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.tngtech.archunit</groupId>
+            <artifactId>archunit-junit5</artifactId>
+        </dependency>
     </dependencies>
 
     <build>
--- a/modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionArchTest.java
+++ b/modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionArchTest.java
@@ -17,18 +17,38 @@
 
 package org.apache.ignite.lang;
 
+import com.tngtech.archunit.core.domain.JavaClass;
 import com.tngtech.archunit.junit.AnalyzeClasses;
 import com.tngtech.archunit.junit.ArchTest;
+import com.tngtech.archunit.lang.ArchCondition;
 import com.tngtech.archunit.lang.ArchRule;
+import com.tngtech.archunit.lang.ConditionEvents;
+import com.tngtech.archunit.lang.SimpleConditionEvent;
 import com.tngtech.archunit.lang.syntax.ArchRuleDefinition;
+
+import java.util.UUID;
 
 /**
  * Tests that all public Ignite exceptions have correct definitions.
  */
 @AnalyzeClasses
 public class IgniteExceptionArchTest {
+    @SuppressWarnings("unused")
     @ArchTest
     public static final ArchRule IGNITE_EXCEPTIONS_HAVE_REQUIRED_CONSTRUCTORS = ArchRuleDefinition.classes()
             .that().areAssignableTo(IgniteException.class)
-            .should().haveOnlyPrivateConstructors();
+            .should(new ArchCondition<>("Have standard IgniteException constructor") {
+                @Override
+                public void check(JavaClass javaClass, ConditionEvents conditionEvents) {
+                    var ctor = javaClass.tryGetConstructor(UUID.class, int.class, String.class, Throwable.class);
+
+                    SimpleConditionEvent event = new SimpleConditionEvent(
+                            javaClass,
+                            ctor != null,
+                            javaClass.getName() + " has a standard constructor with " +
+                                    "(UUID traceId, int code, String message, Throwable cause) signature.");
+
+                    conditionEvents.add(event);
+                }
+            });
 }
--- a/modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionArchTest.java
+++ b/modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionArchTest.java
@@ -55,6 +55,10 @@ public class IgniteExceptionArchTest {
             .should(new ArchCondition<>("have standard IgniteException constructor") {
                 @Override
                 public void check(JavaClass javaClass, ConditionEvents conditionEvents) {
+                    if (javaClass.getName().contains(".internal.")) {
+                        return;
+                    }
+
                     System.out.println("Checking class: " + javaClass.getName());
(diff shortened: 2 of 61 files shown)


ptupitsyn:
Test failure example: https://ci.ignite.apache.org/buildConfiguration/ignite3_Test_RunAllTests/6758803?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryCancelledException.java
```diff
@@ -33,4 +34,16 @@ public class QueryCancelledException extends IgniteException {
...
+
+    /**
+     * Creates a new exception with the given trace id, error code, detail message and cause.
+     *
+     * @param traceId Unique identifier of this exception.
+     * @param code Full error code.
+     * @param message Detail message.
+     * @param cause Optional nested exception (can be {@code null}).
+     */
+    public QueryCancelledException(UUID traceId, int code, String message, Throwable cause) {
```
ygerzhedovich, line 46: I suppose we shouldn't have a rule to have specific constructors for internal exceptions, just for public one.
In case we throws internal exception throught public API, we should fix it instead of cover up the problem

ptupitsyn: Good point, fixed: 
* Reverted internal class changes.
* Updated the test to skip classes in internal packages.

modules/api/pom.xml
```diff
@@ -62,6 +62,16 @@
             <artifactId>hamcrest-optional</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>com.tngtech.archunit</groupId>
+            <artifactId>archunit</artifactId>
```
ibessonov, line 68: `test` scope is missing

modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionArchTest.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+@AnalyzeClasses(
+        packages = "org.apache.ignite",
+        locations = IgniteExceptionArchTest.RootLocationProvider.class)
+public class IgniteExceptionArchTest {
+    static class RootLocationProvider implements LocationProvider {
+        @Override
+        public Set<Location> get(Class<?> testClass) {
+            // ignite-3/modules
+            var modulesRoot = Path.of("").toAbsolutePath().getParent();
```
ibessonov, line 45: We don't usually use `var` if type is not explicitly stated, like, for example, in a `new` statement

modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionArchTest.java
```diff
@@ -0,0 +1,76 @@
...
+    public static final ArchRule IGNITE_EXCEPTIONS_HAVE_REQUIRED_CONSTRUCTORS = ArchRuleDefinition.classes()
+            .that().areAssignableTo(IgniteException.class)
+            .should(new ArchCondition<>("have standard IgniteException constructor") {
+                @Override
+                public void check(JavaClass javaClass, ConditionEvents conditionEvents) {
+                    if (javaClass.getName().contains(".internal.")) {
+                        return;
+                    }
+
+                    System.out.println("Checking class: " + javaClass.getName());
```
ibessonov, line 62: Why do we need these logs?

modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionArchTest.java
```diff
@@ -0,0 +1,76 @@
...
+            var modulesRoot = Path.of("").toAbsolutePath().getParent();
+
+            return Set.of(Location.of(modulesRoot));
+        }
+    }
+
+    @SuppressWarnings("unused")
+    @ArchTest
+    public static final ArchRule IGNITE_EXCEPTIONS_HAVE_REQUIRED_CONSTRUCTORS = ArchRuleDefinition.classes()
+            .that().areAssignableTo(IgniteException.class)
```
ibessonov, line 54: What about IgniteCheckedException?

modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionArchTest.java
```diff
@@ -0,0 +1,76 @@
...
+                public void check(JavaClass javaClass, ConditionEvents conditionEvents) {
+                    if (javaClass.getName().contains(".internal.")) {
+                        return;
+                    }
+
+                    System.out.println("Checking class: " + javaClass.getName());
+                    var ctor = javaClass.tryGetConstructor(UUID.class, int.class, String.class, Throwable.class);
+
+                    if (!ctor.isPresent()) {
+                        var event = SimpleConditionEvent.violated(
```
ibessonov, line 66: `var` is misused here as well. Idk if that's a big problem or not, honestly

modules/api/pom.xml
ptupitsyn: It is set in parent pom:
https://github.com/apache/ignite-3/blob/0dc1b5852531c2e7f06dc1091c21b94c454fedd3/parent/pom.xml#L802

modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionArchTest.java
ptupitsyn: Fixed.

modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionArchTest.java
ptupitsyn: Good catch, fixed.

modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionArchTest.java
ptupitsyn: To see that ArchUnit actually finds what we expect.

modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionArchTest.java
ptupitsyn: Fixed.

modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionArchTest.java
ibessonov: So that's a test for the test, but it requires a person looking into logs. That's weird

ibessonov (approved):
I'm cool with these changes, thank you!

modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionArchTest.java
ptupitsyn: Ok, removed it.

PR #1066, opened 2022-09-08
IGNITE-17445 RocksDbKeyValueStorage recreates DB on start, so data ca…

vldpyatkov:
…n't be found until Raft log is replayed


modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/ItJraftCounterServerTest.java
```diff
@@ -193,20 +214,15 @@ protected void after() throws Exception {
...
-        LOG.info(">>>>>>>>>>>>>>> End test method: {}", testInfo.getTestMethod().orElseThrow().getName());
+        servers.clear();
     }
 
     /**
      * Starts server.
      *
-     * @param idx The index.
-     * @param clo Init closure.
+     * @param idx  The index.
```
ibessonov, line 223: Our code style doesn't force us to have padding for parameters, as far as I remember

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/ItJraftCounterServerTest.java
```diff
@@ -270,21 +286,25 @@ private void startCluster() throws Exception {
             startServer(i, raftServer -> {
                 raftServer.startRaftGroup(COUNTER_GROUP_0, listenerFactory.get(), INITIAL_CONF, defaults());
                 raftServer.startRaftGroup(COUNTER_GROUP_1, listenerFactory.get(), INITIAL_CONF, defaults());
-            }, opts -> {});
+            }, opts -> {
```
ibessonov, line 289: Old code looked better. You probably used auto-formatting for the entire file, right?

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/ItJraftCounterServerTest.java
```diff
@@ -709,6 +733,85 @@ public void testTimerThreadsCount() {
...
 
+    /**
+     * The test shows that all committed updates are applied after a RAFT group restart automatically.
+     * Actual data be available to read from state storage (not a state machine) directly just after the RAFT node started.
+     *
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testApplyUpdatesOutOfSnapshot() throws Exception {
+        HashMap<Integer, AtomicInteger> counters = new HashMap<>(3);
```
ibessonov, line 744: Why do you declare variables like `HashMap` and not `Map`? This looks very out of place

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/ItJraftCounterServerTest.java
```diff
@@ -709,6 +733,85 @@ public void testTimerThreadsCount() {
...
+        HashMap<Path, Integer> snapshotDataStorage = new HashMap<>(3);
+        HashMap<String, SnapshotMeta> snapshotMetaStorage = new HashMap<>(3);
+
+        for (int i = 0; i < 3; i++) {
+            var counter = counters.computeIfAbsent(i, (key) -> new AtomicInteger());
```
ibessonov, line 749: This can be a simple `put`, right?

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/ItJraftCounterServerTest.java
```diff
@@ -709,6 +733,85 @@ public void testTimerThreadsCount() {
...
+        raftClient.run(new TestWriteCommand());
+
+        for (AtomicInteger counter : counters.values()) {
+            assertTrue(TestUtils.waitForCondition(() -> counter.get() == 3, 10_000));
+        }
+
+        shutdownCluster();
+
+        Path peer0SnapPath = dataPath.resolve("node" + 0)
+                .resolve("test_raft_group" + "_" + peers.get(0).address().toString().replace(':', '_')).resolve("snapshot");
```
ibessonov, line 784: Is there a helper method somewhere to mask special characters?

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/ItJraftCounterServerTest.java
```diff
@@ -709,6 +733,85 @@ public void testTimerThreadsCount() {
...
     }
 
+    /**
+     * The test shows that all committed updates are applied after a RAFT group restart automatically.
+     * Actual data be available to read from state storage (not a state machine) directly just after the RAFT node started.
+     *
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testApplyUpdatesOutOfSnapshot() throws Exception {
```
ibessonov, line 743: Is it possible to write a unit test instead of integration one?

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/snasphot/UpdateCountRaftListener.java
```diff
@@ -0,0 +1,105 @@
...
+public class UpdateCountRaftListener implements RaftGroupListener {
+
+    /** The logger. */
+    private static IgniteLogger LOG = Loggers.forClass(UpdateCountRaftListener.class);
+
+    /** Counter of received updates. */
+    private final AtomicInteger counter;
+
+    /** Storage to persist a state of the listener on snapshot. */
+    private final HashMap<Path, Integer> snapshotDataStorage;
```
ibessonov, line 50: Same question here, what's the point of declaring it as `HashMap`?

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/snasphot/UpdateCountRaftListener.java
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.ignite.raft.client.service.RaftGroupListener;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * The RAFT state machine counts applied write commands and stores the result into {@link java.util.concurrent.atomic.AtomicLong} that is
+ * passed through constructor.
+ * {@see org.apache.ignite.raft.locked.TestWriteCommand}
+ */
+public class UpdateCountRaftListener implements RaftGroupListener {
+
```
ibessonov, line 42: This empty line shouldn't be here probably

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
```diff
@@ -1033,6 +1034,33 @@ public boolean init(final NodeOptions opts) {
         // Adds metric registry to RPC service.
         this.options.setMetricRegistry(this.metrics.getMetricRegistry());
 
+        //Wait committed
```
ibessonov, line 1037:
```suggestion
        // Wait committed.
```

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
```diff
@@ -1033,6 +1034,33 @@ public boolean init(final NodeOptions opts) {
...
+                    applyCommitLatch.countDown();
+                }
+            };
+
+            fsmCaller.addLastAppliedLogIndexListener(lnsr);
+
+            fsmCaller.onCommitted(commitIdx);
+
+            try {
+                applyCommitLatch.await();
```
ibessonov, line 1054: How long would it take? Did it affect the duration of current tests?
I hope we start these nodes in parallel at least.

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/ItJraftCounterServerTest.java
vldpyatkov: It might, but in the case, I will be needed to copy methods to start/stop raft nodes and raft clients.

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/ItJraftCounterServerTest.java
vldpyatkov: Because I use the implementation here.
Using interfaces is helpfully for the case when you receive an implementation from other code.

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/ItJraftCounterServerTest.java
vldpyatkov: Fixed.

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/ItJraftCounterServerTest.java
vldpyatkov: I did not find a method, but I can write it int the test class.

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/snasphot/UpdateCountRaftListener.java
vldpyatkov: removed.

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/snasphot/UpdateCountRaftListener.java
vldpyatkov: It is a place where is used interface better than implementation.
But I thought it is a test and fewer imports is better...
Ok, I fixed it.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
vldpyatkov: I hope this change will not have significant effect, but I understand the pure RAFT test become longer.
We can do this feature is configurable if it will be required in the future.

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/ItJraftCounterServerTest.java
vldpyatkov: It is true.

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/ItJraftCounterServerTest.java
ibessonov: I don't understand this reply. Do you use specific methods of HashMap that don't exist in Map?

ibessonov (approved):
Looks good to me, please proceed with review from someone that understand raft internals.

PR #1068, opened 2022-09-09
IGNITE-17628 Get rid of Statement#prepared in public API

korlov42:


ptupitsyn (approved):

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/IgniteSqlApiTest.java
```diff
@@ -105,11 +105,10 @@ public void statementApi() {
                 .property("memoryQuota", 10 * Constants.MiB) // Set default quota.
                 .build();
 
-        // Statement with params. Prepared statement.
+        // Statement with params.
         Statement preparedStatement = igniteSql.statementBuilder()
```
ygerzhedovich, line 109: shoud we rename the var?

ygerzhedovich: and may be delete the case at all

korlov42: fixed

ygerzhedovich (approved):

PR #1070, opened 2022-09-09
IGNITE-17621 Fixed registration already completed future in InFlightF…

sk0x50:
…utures

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/future/InFlightFuturesTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/future/InFlightFuturesTest.java
@@ -42,7 +42,7 @@ class InFlightFuturesTest {
 
 
     @Test
-    void addsAlreadyCompletedFutureToInFlightSet() {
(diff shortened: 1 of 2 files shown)


modules/core/src/test/java/org/apache/ignite/internal/future/InFlightFuturesTest.java
```diff
@@ -40,6 +40,16 @@ void addsFutureToInFlightSetOnRegistration() {
         assertThat(currentFutures(inFlightFutures), is(singleton(incompleteFuture)));
     }
 
+
+    @Test
+    void addsAlreadyCompletedFutureToInFlightSet() {
```
rpuch, line 45: The idea of the test seems to be to check that the future actually gets removed (not added). How about renaming the test method to `removesAlreadyCompletedFutureOnRegistration()`?

sk0x50: Yes, this makes sense to me. Will change the name.

rpuch (approved):

PR #1071, opened 2022-09-12
IGNITE-17667 Divide long test into suitable parts

zstan:


zstan:
@ygerzhedovich can you check it plz ?

ygerzhedovich:
@zstan could you please explain why you split test to so small pices? Results on my local machine
p1 - 16 sec
p2 - 14 sec
p3 - 1 min 45 sec
p4 - 1 min 3 sec
p5 - 1 min 5 sec
p6 - 20 sec
p7 - 46 sec
p8 - 59 sec

Also all these tests is green, so I don't undestand why they were ignored.

ygerzhedovich (approved):

PR #1076, opened 2022-09-13
IGNITE-17318 Implement RocksDB based sorted index storage

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-17318


modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.ignite.internal.binarytuple.BinaryTuplePrefixBuilder;
+import org.apache.ignite.internal.binarytuple.BinaryTupleReader;
+import org.apache.ignite.internal.schema.row.InternalTuple;
+
+/**
+ * Class that represents a Binary Tuple Prefix.
+ *
+ * @see BinaryTuplePrefixBuilder BinaryTuplePrefixBuilder for information about the Binary Tuple Prefix format.
+ */
+public class BinaryTuplePrefix extends BinaryTupleReader implements InternalTuple {
```
tkalkirill, line 31: I think it's better to make it the successor of the BinaryTuple. 

In fact, in the SortedIndexStorage, we will also work with BinaryTuple and not just with prefixes, and this confuses when used.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/BinaryTupleBuilder.java
```diff
@@ -567,6 +554,15 @@ public int elementIndex() {
         return elementIndex;
     }
 
+    /**
+     * Gets the expected number of tuple elements.
+     *
+     * @return Expected number of tuple elements.
+     */
```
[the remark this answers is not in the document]
sashapolo: All other methods of this class have a `@return` tag

modules/rocksdb-common/src/main/java/org/apache/ignite/internal/rocksdb/RocksUtils.java
```diff
@@ -83,28 +83,28 @@ public static void checkIterator(RocksIterator it) {
      * <p>This method tries to increment the least significant byte (in BE order) that is not equal to 0xFF (bytes are treated as
      * unsigned values).
      *
-     * @param rangeStart Start of a range of keys (prefix) in RocksDB.
+     * @param array Start of a range of keys (prefix) in RocksDB.
      * @return End of a range of keys in RocksDB or {@code null} if all bytes of the prefix are equal to 0xFF.
      */
-    public static byte @Nullable [] rangeEnd(byte[] rangeStart) {
-        byte[] rangeEnd = rangeStart.clone();
+    public static byte @Nullable [] incrementArray(byte[] array) {
```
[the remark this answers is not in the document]
sashapolo: Unfortunately this is not an equivalent transformation

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
ibessonov: Won't that break the hierarchy? I'd even go as far as not extending it from BinaryTupleReader tbh. Current classes already seem a little bit broken

sashapolo: > I'd even go as far as not extending it from BinaryTupleReader tbh

How would you read data from the prefix then?

modules/client/src/main/java/org/apache/ignite/internal/client/table/ClientKeyValueView.java
```diff
@@ -458,7 +458,7 @@ private void writeKeyValue(ClientSchema s, PayloadOutputChannel w, @Nullable Tra
     }
 
     private void writeKeyValueRaw(ClientSchema s, PayloadOutputChannel w, @NotNull K key, V val) {
-        var builder = BinaryTupleBuilder.create(s.columns().length, true);
+        var builder = new BinaryTupleBuilder(s.columns().length, true);
```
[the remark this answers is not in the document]
sashapolo: Because I needed to inherit from the Builder. I could have made the constructor `protected`, but I really don't see any reason in having a factory method that simply delegates to a constructor

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
sashapolo: Yes, how I would use it in the comparator then?

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
```diff
@@ -0,0 +1,71 @@
...
+        return elementCount();
+    }
+
+    @Override
+    public BigDecimal decimalValue(int index) {
+        return decimalValue(index, schema.element(index).decimalScale);
+    }
+
+    @Override
+    public int elementCount() {
```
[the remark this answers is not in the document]
sashapolo: As far as I understand, there's no difference, they simply come from different interfaces

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/SortedIndexStorage.java
```diff
@@ -74,8 +74,13 @@ public interface SortedIndexStorage {
...
-            @Nullable BinaryTuple upperBound,
+            @Nullable BinaryTuplePrefix lowerBound,
+            @Nullable BinaryTuplePrefix upperBound,
             @MagicConstant(flagsFromClass = SortedIndexStorage.class) int flags
     );
+
+    /**
+     * Removes all data from this index.
+     */
+    void destroy();
```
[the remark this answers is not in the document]
sashapolo: Hm, I had a look at the HashIndexStorage and you've actually deprecated that method. Looks like we need to simply remove this method as well

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/SortedIndices.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.ignite.internal.storage.index.SortedIndexStorage;
+import org.apache.ignite.internal.storage.rocksdb.index.RocksDbSortedIndexStorage;
+import org.rocksdb.RocksDBException;
+
+class SortedIndices {
+    private final SortedIndexDescriptor descriptor;
+
+    private final ColumnFamily columnFamily;
+
+    private final ConcurrentMap<Integer, SortedIndexStorage> storages = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
sashapolo: No particular reason. I just think that this array may be quite sparse and creating a sparse array for every index may be wasteful

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/ColumnFamilyUtils.java
```diff
@@ -58,9 +64,35 @@ static ColumnFamilyType fromCfName(String cfName) {
...
+
+    /**
+     * Extracts a Sorted Index ID from the given Column Family name.
+     *
+     * @param cfName Column Family name.
+     * @return Sorted Index ID.
+     *
+     * @see #sortedIndexCfName
+     */
+    static UUID sortedIndexId(String cfName) {
```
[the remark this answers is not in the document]
sashapolo: We can, but what's the point? I think that having human readable CF names may be useful

ibessonov (approved):

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/BinaryTupleComparator.java
```diff
@@ -15,80 +15,66 @@
...
-     * @param equals Value that should be returned if the provided tuple exactly matches the prefix.
-     * @return the value {@code 0} if the given row starts with the configured prefix;
-     *         a value less than {@code 0} if the row's prefix is smaller than the prefix; and
-     *         a value greater than {@code 0} if the row's prefix is larger than the prefix.
-     */
-    public int compare(BinaryTuple tuple1, BinaryTuple tuple2, int direction, int equals) {
-        for (int i = 0; i < prefixLength; i++) {
+        BinaryTupleSchema schema = descriptor.binaryTupleSchema();
+
+        InternalTuple tuple1 = isPrefix(buffer1) ? new BinaryTuplePrefix(schema, buffer1) : new BinaryTuple(schema, buffer1);
```
SammyVimes, line 54: I think `isPrefix` can be extracted into a variable (you already called this method in the assert statement)

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/ColumnFamilyUtils.java
```diff
@@ -58,9 +64,35 @@ static ColumnFamilyType fromCfName(String cfName) {
                 return PARTITION;
             } else if (HASH_INDEX_CF_NAME.equals(cfName)) {
                 return HASH_INDEX;
+            } else if (cfName.startsWith(SORTED_INDEX_CF_PREFIX)) {
+                return SORTED_INDEX;
```
SammyVimes, line 68: So we have a column family for every index? Why not one cf for all indexes?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
```diff
@@ -378,7 +394,28 @@ public CompletableFuture<Void> destroyPartition(int partitionId) throws StorageE
     /** {@inheritDoc} */
     @Override
     public SortedIndexStorage getOrCreateSortedIndex(int partitionId, UUID indexId) {
-        throw new UnsupportedOperationException("Not implemented yet");
+        SortedIndices storages = sortedIndices.computeIfAbsent(indexId, id -> {
+            var indexDescriptor = new SortedIndexDescriptor(indexId, tableCfg.value());
```
SammyVimes, line 398: I think this lambda deserves its own method

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
```diff
@@ -488,8 +533,26 @@ private ColumnFamilyDescriptor cfDescriptorFromName(String cfName) {
...
             default:
                 throw new StorageException("Unidentified column family [name=" + cfName + ", table=" + tableCfg.name() + ']');
         }
     }
+
+    /**
+     * Creates a Column Family descriptor for a Sorted Index.
+     */
+    private static ColumnFamilyDescriptor sortedIndexCfDescriptor(SortedIndexDescriptor descriptor) {
+        String cfName = sortedIndexCfName(descriptor.id());
```
SammyVimes, line 550: You just wrapped cfName into index id and now you're unwrapping it

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/SortedIndices.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import org.apache.ignite.internal.rocksdb.ColumnFamily;
+import org.apache.ignite.internal.storage.StorageException;
+import org.apache.ignite.internal.storage.index.SortedIndexDescriptor;
+import org.apache.ignite.internal.storage.index.SortedIndexStorage;
+import org.apache.ignite.internal.storage.rocksdb.index.RocksDbSortedIndexStorage;
+import org.rocksdb.RocksDBException;
+
+class SortedIndices {
```
SammyVimes, line 29: I think javadoc would be nice here.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/CursorUtils.java
```diff
@@ -0,0 +1,259 @@
...
+
+        @Override
+        public void close() throws Exception {
+            cursor.close();
+        }
+
+        @Override
+        public boolean hasNext() {
+            if (predicate == null) {
+                return skippedElement != null || cursor.hasNext();
```
SammyVimes, line 59: So it's not skipped. Probably "lastElement"?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbBinaryTupleComparator.java
```diff
@@ -0,0 +1,102 @@
...
+        // Binary Tuple Prefixes don't have row IDs, so they can't be compared.
+        if (compareTuples != 0 || isPrefix(firstBinaryTupleBuffer) || isPrefix(secondBinaryTupleBuffer)) {
+            return compareTuples;
+        }
+
+        return compareRowIds(a, b);
+    }
+
+    private static int compareRowIds(ByteBuffer a, ByteBuffer b) {
+        long firstMostSignBits = a.getLong(a.remaining() - Long.BYTES * 2);
```
SammyVimes, line 81: probably should be `a.limit()`

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbBinaryTupleComparator.java
```diff
@@ -0,0 +1,102 @@
...
+        long secondMostSignBits = b.getLong(b.remaining() - Long.BYTES * 2);
```
SammyVimes, line 82: same about limit

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbBinaryTupleComparator.java
```diff
@@ -0,0 +1,102 @@
...
+
+        int compare = Long.compare(firstMostSignBits, secondMostSignBits);
+
+        if (compare != 0) {
+            return compare;
+        }
+
+        long firstLeastSignBits = a.getLong(a.remaining() - Long.BYTES);
```
SammyVimes, line 90: also here

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbBinaryTupleComparator.java
```diff
@@ -0,0 +1,102 @@
...
+        long secondLeastSignBits = b.getLong(b.remaining() - Long.BYTES);
```
SammyVimes, line 91: and here

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbSortedIndexStorage.java
```diff
@@ -0,0 +1,240 @@
...
+
+            // Finally, compare the remaining parts of the tuples.
+            // TODO: This part may be optimized by comparing binary tuple representations directly. However, currently BinaryTuple prefixes
+            //  are not binary compatible with regular tuples. See https://issues.apache.org/jira/browse/IGNITE-17711.
+            return comparator.compare(prefix.byteBuffer(), binaryTupleSlice(key)) == 0;
+        };
+    }
+
+    private static ByteBuffer binaryTupleSlice(ByteBuffer key) {
+        return key.duplicate()
```
SammyVimes, line 232: Why duplicate instead of just slice?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbSortedIndexStorage.java
```diff
@@ -0,0 +1,240 @@
...
+                // Discard partition ID.
+                .position(Short.BYTES)
+                // Discard row ID.
+                .limit(key.limit() - Long.BYTES * 2)
+                .slice()
```
SammyVimes, line 237: And a slice here? Seems off

SammyVimes (changes requested):

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/ColumnFamilyUtils.java
sashapolo: The original idea was that you can only specify a single comparator per CF. But now I gave it a thought, and I think that we can get away with a single CF if we include the index ID in the index row and propagate a map from Index ID to its schema into the comparator. I will try to implement that.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbBinaryTupleComparator.java
sashapolo: This code is technically still correct, since position is always 0, but I will using `limit` is "more" correct. Nice catch!

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbSortedIndexStorage.java
sashapolo: In order to create a slice I need to change the original buffer's position and limit. Do you have a better approach in mind?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/ColumnFamilyUtils.java
sashapolo: I've had a little discussion and we decided to keep a single CF for performance reasons (so that you don't have to lookup index schema for every comparison). I will create a ticket to test the performance of the single CF approach

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbSortedIndexStorage.java
SammyVimes: I thought ByteBuffer#slice(start, length) exists. But it is available only in java13+ :(

SammyVimes (approved):
LGTM

PR #1077, opened 2022-09-14
IGNITE-17676 Fixed race in AwaitTasksCompletionExecutor that causes tests to be flaky

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-17676

### DIFF
--- a/modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/checkpoint/AwaitTasksCompletionExecutor.java
+++ b/modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/checkpoint/AwaitTasksCompletionExecutor.java
@@ -50,9 +50,9 @@ class AwaitTasksCompletionExecutor implements Executor {
     public void execute(Runnable command) {
         CompletableFuture<?> future = new CompletableFuture<>();
 
-        future.whenComplete((o, throwable) -> updateHeartbeat.run());
+        CompletableFuture<?> heartbeatFuture = future.whenComplete((o, throwable) -> updateHeartbeat.run());
 
-        pendingTaskFutures.add(future);
+        pendingTaskFutures.add(heartbeatFuture);
 
         executor.execute(() -> {
             try {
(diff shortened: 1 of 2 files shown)


modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/checkpoint/AwaitTasksCompletionExecutor.java
```diff
@@ -50,9 +50,9 @@ class AwaitTasksCompletionExecutor implements Executor {
     public void execute(Runnable command) {
         CompletableFuture<?> future = new CompletableFuture<>();
 
-        future.whenComplete((o, throwable) -> updateHeartbeat.run());
+        CompletableFuture<?> heartbeatFuture = future.whenComplete((o, throwable) -> updateHeartbeat.run());
 
-        pendingTaskFutures.add(future);
+        pendingTaskFutures.add(heartbeatFuture);
```
tkalkirill, line 55:
```suggestion
        pendingTaskFutures.add(future.whenComplete((o, throwable) -> updateHeartbeat.run()));
```

tkalkirill (approved):

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/checkpoint/AwaitTasksCompletionExecutor.java
ibessonov: Why do you think I should inline this statement? We would have 3 methods call in a single line, that's a lot and kinda unnecessary

tkalkirill: It's just an offer, you can reject it

ibessonov: Ok, I just wanted to know your motivation

tkalkirill: Just fewer lines of code

PR #1081, opened 2022-09-15
IGNITE-17355 CLI management of metrics

valepakh:
https://issues.apache.org/jira/browse/IGNITE-17355

This partially implements metric management - only for nodes pending discussion and decision about cluster-wide metric management.


Pochatkin:
LGTM

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/rest/NodeMetricController.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public NodeMetricController(MetricManager metricManager) {
+        this.metricManager = metricManager;
+    }
+
+    @Override
+    public void enable(String srcName) {
+        try {
+            metricManager.enable(srcName);
+        } catch (IllegalStateException e) {
```
kgusakov, line 41: It can produce `IllegalArgumentException` also

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/rest/NodeMetricController.java
```diff
@@ -0,0 +1,61 @@
...
+            throw new MetricNotFoundException(e);
+        }
+    }
+
+    @Override
+    public void disable(String srcName) {
+        try {
+            metricManager.disable(srcName);
```
kgusakov, line 49: It can produce `IllegalArgumentException` also

kgusakov (changes requested):

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/rest/NodeMetricController.java
valepakh: Looks like only `MetricRegistry#enable(MetricSource)` throws IAE, while `MetricRegistry#enable(String)` does not.

kgusakov: Yep, you are right

kgusakov (approved):

SammyVimes (approved):
LGTM

PR #1088, opened 2022-09-16
IGNITE-17696 Provide initial C++ style guide and clang-format

ademakov:


modules/platforms/cpp/client/include/ignite/ignite_client_configuration.h
```diff
@@ -85,10 +80,7 @@ class IgniteClientConfiguration
      *
      * @return Current logger.
      */
-    [[nodiscard]]
-    const std::shared_ptr<IgniteLogger>& getLogger() const {
-        return m_logger;
-    }
+    [[nodiscard]] const std::shared_ptr<IgniteLogger> &getLogger() const { return m_logger; }
```
isapego, line 83: I'm not sure about putting function attributes and body on the same line as other declaration by default - it feels to me as there is too much on a single line and it's harder to read such a code. Also, this leaves even less space for arguments.

modules/platforms/cpp/StyleGuide.md
```diff
@@ -0,0 +1,141 @@
...
+
+* The prototype/interface header for this implementation (ie, the .h/.hh file that
+  corresponds to this .cpp/.cc file).
+* Other headers from the same project, as needed.
+* Headers from other non-system libraries (for example, Boost.)
+* Standard C++, C, and system headers.
+* Conditionally included headers.
+
+These groups are separated by an empty line. Inside each group files are sorted in alphabetical
+order.
```
isapego, line 141: What is the benefits of this approach compared to putting system and std headers first (which I used to)?

modules/platforms/cpp/client/include/ignite/ignite_client_configuration.h
ademakov: Unfortunately clang-format doesn't have an option for this. Or at least I was not able to find it. We either use it as is or give up and do everything manually. I hope they will add such an option in future.

modules/platforms/cpp/StyleGuide.md
ademakov: There are two opposite views on this and neither is inherently superior. This is a matter of taste. I personally find that standard files are less important. They are just like noise. More important are dependencies between project modules and ыщ they should go first.

isapego: Sounds good.

isapego (approved):

PR #1095, opened 2022-09-19
IGNITE-17714: Separate task for tests

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-17714


Pochatkin:
Done in https://github.com/apache/ignite-3/pull/1199

PR #1102, opened 2022-09-19
IGNITE-17627 Extend MvPartitionStorage read API with write intent resolution capabilities

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-17627


modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/AbstractPageMemoryMvPartitionStorage.java
```diff
@@ -213,7 +218,11 @@ private PageMemoryHashIndexStorage createOrRestoreHashIndex(IndexMeta indexMeta)
 
     /** {@inheritDoc} */
     @Override
-    public @Nullable BinaryRow read(RowId rowId, HybridTimestamp timestamp) throws StorageException {
+    public @Nullable ReadResult read(RowId rowId, HybridTimestamp timestamp) throws StorageException {
```
ibessonov, line 221: btw, is it ok to return nulls now? I mean, what if you have a tombstone? There would be two ways to return empty value then, not cool. `ReadResult` itself should never be null.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/AbstractPageMemoryMvPartitionStorage.java
```diff
@@ -240,9 +249,23 @@ private PageMemoryHashIndexStorage createOrRestoreHashIndex(IndexMeta indexMeta)
...
+
+            assert chainTxId != null;
+
+            throwIfChainBelongsToAnotherTx(versionChain, txId);
+
+            return ReadResult.createFromWriteIntent(
+                    row,
+                    versionChain.transactionId(),
+                    versionChain.commitTableId(),
+                    null,
```
ibessonov, line 263: It looks weird that there's a gap between table id and partition id, I'd love to see them side by side

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/AbstractPageMemoryMvPartitionStorage.java
```diff
@@ -280,38 +305,65 @@ private void throwIfChainBelongsToAnotherTx(VersionChain versionChain, UUID txId
...
-
-        long newestCommittedLink = versionChain.newestCommittedLink();
+    private @Nullable ReadResult findRowVersionByTimestamp(VersionChain versionChain, HybridTimestamp timestamp) {
+        long headLink = versionChain.headLink();
 
         ScanVersionChainByTimestamp scanByTimestamp = new ScanVersionChainByTimestamp(partitionId);
 
         try {
-            rowVersionDataPageReader.traverse(newestCommittedLink, scanByTimestamp, timestamp);
+            rowVersionDataPageReader.traverse(headLink, scanByTimestamp, timestamp);
```
ibessonov, line 334: Maybe we should rename `scanByTimestamp`

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
```diff
@@ -106,12 +106,12 @@ private static HybridTimestamp convertTimestamp(Timestamp timestamp) {
...
      * @param rowId Row id.
      * @param txId Transaction id.
-     * @return Binary row that corresponds to the key or {@code null} if value is not found.
+     * @return Read result that corresponds to the key or {@code null} if value is not found.
      * @throws TxIdMismatchException If there's another pending update associated with different transaction id.
      * @throws StorageException If failed to read data from the storage.
      */
     @Nullable
-    BinaryRow read(RowId rowId, UUID txId) throws TxIdMismatchException, StorageException;
+    ReadResult read(RowId rowId, UUID txId) throws TxIdMismatchException, StorageException;
```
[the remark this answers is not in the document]
SammyVimes: Right

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
```diff
@@ -142,28 +151,30 @@ default BinaryRow read(RowId rowId, Timestamp timestamp) throws StorageException
...
     @Deprecated
     RowId insert(BinaryRow binaryRow, UUID txId) throws StorageException;
 
     /**
-     * Creates (or replaces) an uncommitted (aka pending) version, assigned to the given transaction id.
-     * In details:
-     * - if there is no uncommitted version, a new uncommitted version is added
-     * - if there is an uncommitted version belonging to the same transaction, it gets replaced by the given version
-     * - if there is an uncommitted version belonging to a different transaction, {@link TxIdMismatchException} is thrown
+     * Creates (or replaces) an uncommitted (aka pending) version, assigned to the given transaction id. In details: - if there is no
```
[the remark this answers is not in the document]
SammyVimes: How did this happen Oo

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/AbstractPageMemoryMvPartitionStorage.java
SammyVimes: Tombstone just returns null. Are you sure we should wrap null values? Seems a bit like overkill, it is an allocation after all. Basically ReadResult can be null, but binary row inside it can't be null.

SammyVimes: Hm, probably a good idea will be to return something like ReadResult.EMPTY

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/AbstractPageMemoryMvPartitionStorage.java
SammyVimes: I don't follow)

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
```diff
@@ -165,14 +185,14 @@ public void testAddWrite() {
...
         addWrite(rowId, binaryRow, txId);
 
-        // Read without timestamp returns uncommited row.
+        // Read without timestamp returns uncommitted row.
         assertRowMatches(read(rowId, txId), binaryRow);
 
         // Read with wrong transaction id should throw exception.
         assertThrows(TxIdMismatchException.class, () -> read(rowId, newTransactionId()));
 
         // Read with timestamp returns null.
```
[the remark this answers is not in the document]
ibessonov: I guess github shows the wrong line, I should have marked it properly. I meant this one:
```
// Read with timestamp returns null.
assertRowMatches(read(rowId, clock.now()), binaryRow);
```

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/AbstractPageMemoryMvPartitionStorage.java
ibessonov: > Basically ReadResult can be null, but binary row inside it can't be null.

This is a faulty logic. You should write such tests. Imagine if you already have a value and you remove it in transaction. This means returning a write intent with null value and non-null timestamp. Is that correct? I think it is.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/AbstractPageMemoryMvPartitionStorage.java
ibessonov: We could explicitly mention write intents or something

SammyVimes: In the name of the class?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -859,19 +1020,60 @@ private static boolean invalid(RocksIterator it) {
...
+     * @param valueBytes Value bytes as read from the storage.
+     * @param newestCommitTs Commit timestamp of the most recent committed write of this value.
+     * @return Read result instance or {@code null} if value is a tombstone.
+     */
+    private static ReadResult wrapUncommittedValue(byte[] valueBytes, @Nullable HybridTimestamp newestCommitTs) {
+        if (isTombstone(valueBytes, true)) {
+            return ReadResult.empty();
+        }
+
+        long txIdMsb = GridUnsafe.getLong(valueBytes, GridUnsafe.BYTE_ARR_OFF + TX_ID_OFFSET);
```
ibessonov, line 1049: You could extract `TxStateRocksDbStorage#bytesToUuid` into utils and introduce offset to it.
i believe there's also similar code in hash index storage for RocksDB

ibessonov (changes requested):

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
```diff
@@ -746,4 +809,168 @@ void testAppliedIndex() {
...
+
+    /**
+     * Performs a decrement operation on a 128-bit unsigned value that is represented by two longs.
+     *
+     * @param msb Most significant bytes of 128-bit unsigned integer.
+     * @param lsb Least significant bytes of 128-bit unsigned integer.
+     * @return Less by one value.
+     */
+    private Pair<Long, Long> getPrevious128Uint(long msb, long lsb) {
+        assert (msb | lsb) != 0L : "Cheer up! That was very unlikely";
```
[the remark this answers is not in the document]
SammyVimes: Good idea

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -335,17 +373,21 @@ public RowId insert(BinaryRow row, UUID txId) throws StorageException {
...
-        putTransactionId(value.array(), 0, txId);
+        putUuid(array, TX_ID_OFFSET, txId);
+        putUuid(array, TABLE_ID_OFFSET, commitTableId);
+        putShort(array, PARTITION_ID_OFFSET, (short) commitPartitionId);
 
-        value.position(TX_ID_SIZE).put(rowBytes);
+        value.position(VALUE_OFFSET).put(rowBytes);
 
         // Write binary row data as a value.
         writeBatch.put(cf, copyOf(keyArray, ROW_PREFIX_SIZE), copyOf(value.array(), value.capacity()));
```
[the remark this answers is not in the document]
SammyVimes: Right

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/chm/TestConcurrentHashMapMvPartitionStorage.java
```diff
@@ -199,51 +218,103 @@ private static BinaryRow read(
...
+
+            return ReadResult.createFromWriteIntent(binaryRow, chainHead.txId, chainHead.commitTableId, latestCommitTs,
+                    chainHead.commitPartitionId);
+        }
+
+        VersionChain cur = firstCommit;
+
         while (cur != null) {
-            if (timestamp.compareTo(cur.begin) >= 0) {
+            int compareResult = timestamp.compareTo(cur.ts);
```
ibessonov, line 293: Might be picky, but I truly believe that inlining this expression would make code simipler. When I see a pattern `x.compareTo(y) >= 0`, I subconsciously replace it with `x >= y`. Splitting evaluation and comparison makes it harder.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/AbstractPageMemoryMvPartitionStorage.java
```diff
@@ -280,38 +302,129 @@ private void throwIfChainBelongsToAnotherTx(VersionChain versionChain, UUID txId
...
-        } catch (IgniteInternalCheckedException e) {
-            throw new StorageException("Cannot search for a row version", e);
+        boolean hasWriteIntent = chainHead.isUncommitted();
+
+        RowVersion firstCommit;
+
+        if (hasWriteIntent) {
+            firstCommit = readRowVersion(chainHead.nextLink(), rowTimestamp -> timestamp.compareTo(rowTimestamp) >= 0);
+        } else {
+            firstCommit = readRowVersion(chainHead.headLink(), ALWAYS_LOAD_VALUE);
```
ibessonov, line 363: I don't know about ALWAYS_LOAD_VALUE, is this necessary?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/AbstractPageMemoryMvPartitionStorage.java
```diff
@@ -280,38 +302,129 @@ private void throwIfChainBelongsToAnotherTx(VersionChain versionChain, UUID txId
...
-        try {
-            rowVersionDataPageReader.traverse(newestCommittedLink, scanByTimestamp, timestamp);
-        } catch (IgniteInternalCheckedException e) {
-            throw new StorageException("Cannot search for a row version", e);
+        boolean hasWriteIntent = chainHead.isUncommitted();
+
+        RowVersion firstCommit;
+
+        if (hasWriteIntent) {
+            firstCommit = readRowVersion(chainHead.nextLink(), rowTimestamp -> timestamp.compareTo(rowTimestamp) >= 0);
```
ibessonov, line 361: I also have bad feelings about this condition

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/AbstractPageMemoryMvPartitionStorage.java
ibessonov: I think you get it, there may be unnecessary value loads

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -443,56 +487,166 @@ public void commitWrite(RowId rowId, HybridTimestamp timestamp) throws StorageEx
...
+                    // Wrong row id.
+                    return ReadResult.empty();
+                }
+
+                boolean isWriteIntent = keyLength == ROW_PREFIX_SIZE;
 
-                // This seek will either find a key with timestamp that's less or equal than required value, or a different key whatsoever.
-                // It is guaranteed by descending order of timestamps.
-                seekIterator.seek(keyBuf.array());
+                byte[] valueBytes = seekIterator.value();
```
ibessonov, line 514: This value load might be unnecessary as well. Is is cheaper then having to do an extra seek? I don't know. Was that your reasoning?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -822,17 +976,24 @@ private static void putTimestamp(ByteBuffer buf, HybridTimestamp ts) {
...
+    }
+
+    private static void putShort(byte[] array, int off, short value) {
+        GridUnsafe.putShort(array, GridUnsafe.BYTE_ARR_OFF + off, value);
     }
 
     private static void validateTxId(byte[] valueBytes, UUID txId) {
-        long msb = GridUnsafe.getLong(valueBytes, GridUnsafe.BYTE_ARR_OFF);
-        long lsb = GridUnsafe.getLong(valueBytes, GridUnsafe.BYTE_ARR_OFF + Long.BYTES);
+        UUID readTxId = bytesToUuid(valueBytes, 0);
```
ibessonov, line 993: Extra allocations, aaa

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
SammyVimes: I feel like doing another seek is going to be less efficient, yes

ibessonov (approved):

PR #1104, opened 2022-09-20
IGNITE-17401 Add classes for complex binary tuple types

ademakov:


modules/platforms/cpp/schema/BinaryTupleBuilder.cpp
```diff
@@ -104,45 +106,45 @@ void BinaryTupleBuilder::putBytes(BytesView bytes) {
 void BinaryTupleBuilder::putInt8(BytesView bytes) {
     SizeT size = sizeOfInt8(BinaryTupleParser::getInt8(bytes));
     assert(size <= bytes.size());
-    putBytes(BytesView{bytes.data(), size});
+    putBytes(BytesView {bytes.data(), size});
```
isapego, line 109: I'm not sure there should be spaces according to new coding style.

modules/platforms/cpp/schema/BinaryTupleParser.cpp
```diff
@@ -18,22 +18,61 @@
...
+    std::int32_t month = (date >> 5) & 15;
+    std::int32_t year = (date >> 9); // Sign matters.
+
+    return Date(year, month, day);
+}
+
+Time loadBytesAsTime(ignite::BytesView bytes) {
+    uint64_t time = loadBytesAs<std::uint32_t>(bytes);
+
+    int nano;
```
isapego, line 50: Why is it `int`? Can it be negative? And why not to specify size which is enough to fit the max value (1'000'000'000 I believe)?

modules/platforms/cpp/schema/BinaryTupleParser.cpp
```diff
@@ -18,22 +18,61 @@
...
+            time >>= 20;
+            break;
+        case 6:
+            time |= std::uint64_t(loadBytesAs<std::uint16_t>(bytes, 4)) << 32;
+            nano = ((int)time & ((1 << 30) - 1));
+            time >>= 30;
+            break;
+    }
+
+    int second = ((int)time) & 63;
```
isapego, line 68: I believe hex (i.e. `0x3F` or `0x40 - 1`) for bit masks is more common approach.

Also, applying this kind of mask to `int` value implies there can be no negative values. Why use `int` then?

modules/platforms/cpp/schema/BinaryTupleParser.cpp
```diff
@@ -18,22 +18,61 @@
...
+    int minute = ((int)time >> 6) & 63;
+    int hour = ((int)time >> 12) & 31;
```
isapego, line 70: Using bit shift operation on `int` value implies it have a certain size in bits. And while it's probably true for the most modern systems, why not use sized type here just to be safe?

isapego (approved):
Overall looks good, but I believe we need tests for `BinaryTupleBuilder` and `BinaryTupleParcer`. At least a simple "serialize, deserialize, check values are the same" test for different values of basic types.

modules/platforms/cpp/schema/BinaryTupleParser.cpp
ademakov: Here we have 60 seconds max so for me the closest mask to the value of 60 in decimal is more intuitive.

modules/platforms/cpp/schema/ignite_time.h
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param hour Hour-of-day, from 0 to 23.
+     * @param minute Minute-of-hour, from 0 to 59.
+     * @param second Second-of-minute, from 0 to 59.
+     * @param nano Nano-of-second, from 0 to 999,999,999.
+     */
+    constexpr ignite_time(int hour, int minute, int second = 0, int nano = 0)
```
isapego, line 44: Pretty weird they are all `int`s

isapego (approved):

modules/platforms/cpp/schema/ignite_time.h
ademakov: Just like in similar java class,

PR #1108, opened 2022-09-21
IGNITE-17730 Fix issues in SQL doc

IgGusev:


ygerzhedovich (approved):

ygerzhedovich:
LGTM, parts related to links in diagram and documentations itself will be covered separately. Schemas should be visible for any possible screen resolution

PR #1110, opened 2022-09-21
IGNITE-17562 Sql. Implement cascade drop for indexes

zstan:


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1234,42 +1232,47 @@ private CompletableFuture<Void> dropTableAsyncInternal(String name) {
...
-                                    .anyMatch(idx -> idxView.get(idx).tableId().equals(tbl.tableId()));
+                            List<String> indexes = indicesNames.stream().filter(idx ->
+                                    tablesCfg.indexes().get(idx).tableId().value().equals(tbl.tableId())).collect(Collectors.toList());
 
-                            //TODO: https://issues.apache.org/jira/browse/IGNITE-17562
-                            // Let's drop orphaned indices instantly.
-                            if (idxFound) {
-                                throw new IgniteException("Can't drop table with indices.");
+                            for (String indexName : indexes) {
+                                idxsChange.add(tablesCfg.indexes().change(chg -> chg.delete(indexName)));
```
AMashenkov, line 1254: What if external changer will be called more than once?
Should `idxsChange` be cleaned at first?

AMashenkov: Does it make sense to move the loop into the changer?
`tablesCfg.indexes().change(chg ->indexes.stream().foreach(indexName -> chg.delete(indexName)))`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1234,27 +1232,26 @@ private CompletableFuture<Void> dropTableAsyncInternal(String name) {
...
-                            NamedListView<TableIndexView> idxView = tablesCfg.indexes().value();
+                            tblChg.delete(name);
+                        }
+                        ).changeIndexes(idxChg -> {
+                            List<String> indicesNames = tablesCfg.indexes().value().namedListKeys();
 
-                            boolean idxFound = idxView.namedListKeys().stream()
-                                    .anyMatch(idx -> idxView.get(idx).tableId().equals(tbl.tableId()));
+                            List<String> indexes = indicesNames.stream().filter(idx ->
+                                    tablesCfg.indexes().get(idx).tableId().value().equals(tbl.tableId())).collect(Collectors.toList());
```
AMashenkov, line 1249:
```suggestion
                            List<String> indexes = indicesNames.stream()
                                    .filter(idx -> tablesCfg.indexes().get(idx).tableId().value().equals(tbl.tableId()))
                                    .foreach(name -> idxChg.delete(name);
```
there is no need to collect results.

AMashenkov (approved):

PR #1112, opened 2022-09-21
IGNITE-17725 Implement Partition Awareness in Java client

ptupitsyn:
* Add `PARTITION_ASSIGNMENT_GET` operation to the protocol.
* Add flags to every server -> client response. First flag indicates changed partition assignment.
* Add partition awareness to `ClientRecordBinaryView.get` (other APIs to be added in IGNITE-17739)

https://issues.apache.org/jira/browse/IGNITE-17725
https://cwiki.apache.org/confluence/display/IGNITE/IEP-95+Client+Partition+Awareness


modules/client/src/main/java/org/apache/ignite/internal/client/ReliableChannel.java
```diff
@@ -606,6 +621,21 @@ private void initAllChannelsAsync() {
         );
     }
 
+    private void onTopologyAssignmentChanged(ClientChannel clientChannel) {
+        // NOTE: Multiple channels will send the same update to us, resulting in multiple cache invalidations.
+        // This could be solved with a cluster-wide AssignmentVersion, but we don't have that.
```
isapego, line 626: Let's select a single server that will be used to monitor changes in assignment. We have heartbeats anyway so we will get updates from it even if we do not use it for communication.

isapego (changes requested):

modules/client/src/main/java/org/apache/ignite/internal/client/ReliableChannel.java
ptupitsyn: That's a great idea! Done. `testClientReceivesPartitionAssignmentUpdates` updated to check heartbeat scenario.

ptupitsyn: Updated the IEP too https://cwiki.apache.org/confluence/display/IGNITE/IEP-95+Client+Partition+Awareness

isapego (approved):

PR #1114, opened 2022-09-22
IGNITE-17687 Get rid of using deprecated exception constructors in the sql-engine module.

xtern:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
```diff
@@ -333,9 +334,9 @@ public static <T> T await(CompletionStage<T> stage) {
         try {
             return stage.toCompletableFuture().get();
         } catch (ExecutionException e) {
-            throw new IgniteException(e.getCause());
+            throw new IgniteException(OPERATION_INTERRUPTED_ERR, e.getCause());
```
ygerzhedovich, line 337: I think we should use SqlException here and below

xtern: Done, thanks

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -344,7 +347,8 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
         SchemaPlus schema = sqlSchemaManager.schema(schemaName);
 
         if (schema == null) {
-            return CompletableFuture.failedFuture(new IgniteInternalException(format("Schema not found [schemaName={}]", schemaName)));
+            return CompletableFuture.failedFuture(
+                    new IgniteInternalException(SCHEMA_NOT_FOUND_ERR, format("Schema not found [schemaName={}]", schemaName)));
```
ygerzhedovich, line 351: I think it should be public exception

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -361,7 +365,7 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
         try {
             session.registerResource(closeableResource);
         } catch (IllegalStateException ex) {
-            return CompletableFuture.failedFuture(new IgniteInternalException(
+            return CompletableFuture.failedFuture(new IgniteInternalException(SESSION_EXPIRED_ERR,
```
ygerzhedovich, line 368: It looks also as SqlException

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -449,7 +453,7 @@ private List<CompletableFuture<AsyncSqlCursor<List<Object>>>> query0(
         SchemaPlus schema = sqlSchemaManager.schema(schemaName);
 
         if (schema == null) {
-            throw new IgniteInternalException(format("Schema not found [schemaName={}]", schemaName));
+            throw new IgniteInternalException(SCHEMA_NOT_FOUND_ERR, format("Schema not found [schemaName={}]", schemaName));
```
ygerzhedovich, line 456: the same

PR #1117, opened 2022-09-23
IGNITE-17745 Fix race condition in CMG Manager

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-17745


PakhomovAlexander:
LGTM

PR #1120, opened 2022-09-26
IGNITE-17493 Add 'isLeader' check to CMG topology listener

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-17493

### DIFF
--- a/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
+++ b/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
@@ -457,7 +457,7 @@ public class ClusterManagementGroupManager implements IgniteComponent {
                                         + "Re-creating the CMG Raft service [reason={}]", e, e.getMessage());
                             } else {
                                 LOG.debug("CMG service started, but the cluster state is different. "
-                                        + "Re-creating the CMG Raft service [localState={}, clusterState={}]",
+                                                + "Re-creating the CMG Raft service [localState={}, clusterState={}]",
                                         service.nodeNames(), state.cmgNodes());
 
                                 destroyCmg();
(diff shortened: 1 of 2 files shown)


modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -526,17 +526,10 @@ private TopologyEventHandler cmgLeaderTopologyEventHandler(CmgRaftService raftSe
         return new TopologyEventHandler() {
             @Override
             public void onAppeared(ClusterNode member) {
-                raftService.readClusterState()
-                        .thenAccept(state -> {
-                            if (state != null) {
-                                sendClusterState(state, member)
-                                        .whenComplete((v, e) -> {
-                                            if (e != null) {
```
rpuch, line 534: This has been dropped. Should it be useful for debugging problems (maybe even as a WARNING)?

rpuch: `sendClusterState()` might throw something that is not covered by the code on line 593 (like an exception when creating a message). This is currently highly unlikely, but saw such cases in other projects (when due to a stupid coding error an NPE is thrown, for instance). If such an un unforseen exception is thrown, we will still not see it, and it will be pretty difficult to figure out what's going on.

IMO, better safe than sorry :)

rpuch (approved):

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
sashapolo: But otherwise we will log the same error twice, I don't like it

PR #1121, opened 2022-09-26
IGNITE-17673 Extended MV partition storage API with methods to help cleaning SQL indices

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-17673

### DIFF
--- a/modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
+++ b/modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
@@ -675,7 +675,7 @@ public class RocksDbMvPartitionStorage implements MvPartitionStorage {
             public void close() throws Exception {
                 super.close();
 
-                IgniteUtils.closeAll(it, options, upperBound);
+                IgniteUtils.closeAll(options, upperBound);
             }
         };
     }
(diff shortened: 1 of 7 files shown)


modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -648,6 +649,37 @@ private static boolean matches(RowId rowId, ByteBuffer keyByf) {
...
+            @Override
+            protected BinaryRow decodeEntry(byte[] key, byte[] value) {
+                return wrapValueIntoBinaryRow(value, key.length == ROW_PREFIX_SIZE);
+            }
+
+            @Override
+            public void close() throws Exception {
+                super.close();
+
+                IgniteUtils.closeAll(it, options, upperBound);
```
sashapolo, line 678: `it` is also closed in the `super.close` call. I wonder why it doesn't fail...

SammyVimes: Because it checks whether pointer still exists.

SammyVimes (changes requested):

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
ibessonov: Oopsie

SammyVimes (approved):
LGTM!

PR #1126, opened 2022-09-27
IGNITE-17711 Append element count to the end of BinaryTuplePrefix

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-17711


modules/schema/src/test/java/org/apache/ignite/internal/schema/BinaryTuplePrefixTest.java
```diff
@@ -71,6 +70,25 @@ public void testPrefix() {
         assertThat(prefix.doubleValueBoxed(4), is(nullValue()));
     }
 
+    /**
+     * Tests a corner case when a new internal buffer needs to be allocated to add the count value.
```
rpuch, line 74: How about renaming the method? It looks like it's about 'a reallocation is needed' case and not just 'small prefix' case

rpuch (approved):

rpuch (approved):

PR #1131, opened 2022-09-28
IGNITE-17726 Packaging: Docker image

valepakh:
https://issues.apache.org/jira/browse/IGNITE-17726

This adds gradle tasks allowing to build docker image. Also it replaces `application` plugin with `distribution` plugin since former is not actually used as we create our own `CreateStartScripts` task.

Default `./gradlew docker` task builds a `apacheignite/ignite3:latest` image, while `./gradlew dockerTag` builds and additional tags it with the project version.

This also allows passing additional arguments like seed nodes or node name via environment variables or command-line arguments.


PakhomovAlexander:
LGTM

PR #1135, opened 2022-09-29
IGNITE-17763 Use retry policy when connecting default channel

ptupitsyn:
Introduce `ClientOperationType.CHANNEL_CONNECT` and pass it to `RetryPolicy` when performing connection attempts.

Fixes `testRetryReadPolicyRetriesReadOperations` flakiness caused by failed handshake.

### DIFF
--- a/modules/client/src/main/java/org/apache/ignite/internal/client/ReliableChannel.java
+++ b/modules/client/src/main/java/org/apache/ignite/internal/client/ReliableChannel.java
@@ -45,6 +45,7 @@ import java.util.stream.IntStream;
 import org.apache.ignite.client.ClientOperationType;
 import org.apache.ignite.client.IgniteClientConfiguration;
 import org.apache.ignite.client.IgniteClientConnectionException;
+import org.apache.ignite.client.RetryLimitPolicy;
 import org.apache.ignite.client.RetryPolicy;
 import org.apache.ignite.client.RetryPolicyContext;
 import org.apache.ignite.internal.client.io.ClientConnectionMultiplexer;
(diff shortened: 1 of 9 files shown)


isapego (approved):

ptupitsyn:
@isapego I've added `ClientOperationType.CHANNEL_CONNECT` so that `RetryPolicy` implementation controls reconnect attempts (instead of hardcoded `DFLT_RETRY_LIMIT`). Please have another look.

modules/client/src/main/java/org/apache/ignite/internal/client/ReliableChannel.java
```diff
@@ -583,8 +588,14 @@ private ClientChannel getDefaultChannel() {
...
 
+        return shouldRetry(opType, iteration, exception);
+    }
+
+    /** Determines whether specified operation should be retried. */
+    private boolean shouldRetry(ClientOperationType opType, int iteration, IgniteClientConnectionException exception) {
         if (opType == null) {
-            return true; // System operation.
+            // System operation.
+            return iteration < RetryLimitPolicy.DFLT_RETRY_LIMIT;
```
isapego, line 598: So does `opType == null` is still used for some operations?

isapego (approved):

modules/client/src/main/java/org/apache/ignite/internal/client/ReliableChannel.java
isapego: Let's leave it this way for now, I'm not sure exposing too much internal implementation details to user will do any good here.

PR #1139, opened 2022-09-29
IGNITE-16994 .NET: Upgrade to SDK 6.0

ptupitsyn:
.NET Core 3.1 support ends in December. Upgrade all projects to .NET 6 (current LTS).

https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core


modules/platforms/dotnet/Apache.Ignite.Tests/ExceptionsTests.cs
```diff
@@ -52,13 +52,15 @@ public void TestExceptionsAreSerializableAndHaveRequiredConstructors()
                 var stream = new MemoryStream();
                 var formatter = new BinaryFormatter();
 
+#pragma warning disable SYSLIB0011 // BinaryFormatter is obsolete
```
isapego, line 55: Is it anything to be used instead?

isapego (approved):
Overall, looks good.

modules/platforms/dotnet/Apache.Ignite.Tests/ExceptionsTests.cs
ptupitsyn: Reworked this test to avoid obsolete `BinaryFormatter`.

NOTE: The future of `ISerializable` and related types is not clear, they may become obsolete too (see https://github.com/dotnet/runtime/issues/70208). But for now we follow existing guidelines and implement this interface properly.

isapego: 👍 nice

PR #1152, opened 2022-10-03
IGNITE-17801 Add support for Duration and Period types in BinaryTuple

ptupitsyn:
IEP updated: https://cwiki.apache.org/confluence/display/IGNITE/IEP-92%3A+Binary+Tuple+Format

### DIFF
--- a/modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/BinaryTupleBuilder.java
+++ b/modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/BinaryTupleBuilder.java
@@ -27,6 +27,7 @@ import java.nio.charset.CharacterCodingException;
 import java.nio.charset.CharsetEncoder;
 import java.nio.charset.CoderResult;
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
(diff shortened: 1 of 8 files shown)


modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/BinaryTupleBuilder.java
```diff
@@ -527,6 +529,78 @@ public BinaryTupleBuilder appendTimestamp(Instant value) {
...
+                    && Short.MIN_VALUE <= days && days <= Short.MAX_VALUE) {
+                putShort((short) years);
+                putShort((short) months);
+                putShort((short) days);
+            } else {
+                putInt(years);
+                putInt(months);
+                putInt(days);
+            }
+        }
```
ademakov, line 589: Isn't `1 <=  month <= 12` and `1 <= days <= 31` here? Why not to pack it in the same way as `LocalDate` ?

ptupitsyn: Any int value is allowed for all components, e.g. `Period.of(-100, 200, -300)`.

ademakov: I see, thanks.

ademakov (commented):
LGTM

PR #1155, opened 2022-10-04
IGNITE-17799 Intergrate onLeaderElected callback into CMG Manager

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-17799


modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -487,27 +467,46 @@ private CompletableFuture<Void> joinCluster(CmgRaftService service, ClusterTag c
...
                     )
                     .thenApply(service -> new CmgRaftService(service, clusterService));
-        } catch (NodeStoppingException e) {
+        } catch (Exception e) {
             return failedFuture(e);
         }
     }
 
+    private CmgRaftGroupListener createCmgRaftGroupListener() {
+        clusterStateStorage.start();
```
rpuch, line 486: It is pretty mean to hide a side effect inside a method which looks like a 'pure' method if you look at its name :) How about renaming the method? Although all the candidate names seem to be too long.

sashapolo: Renamed to `startCmgRaftGroupListener`

sashapolo: I've inlined this method

rpuch (approved):

PR #1156, opened 2022-10-04
IGNITE-17812 Fix RetryPolicy exception handling, propagate exceptions to API caller

ptupitsyn:
Fix API call hang when there is an exception in `RetryPolicy` implementation. Catch exceptions and propagate to the caller.


isapego (approved):
Looks good to me.

PR #1157, opened 2022-10-04
IGNITE-17782 `sql` with white space in repl fails

valepakh:
https://issues.apache.org/jira/browse/IGNITE-17782

Picocli treats white space as a positional argument and creates a arg group object, so we need to add a check for this.


SammyVimes (approved):
LGTM

PR #1160, opened 2022-10-04
IGNITE-17083 Skeleton classes for RAFT snapshots in partitions & new method in MvPartitionStorage.

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-17083


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -0,0 +1,117 @@
...
+     * @param snapshotUri Snapshot URI.
+     */
+    public IncomingSnapshotCopier(PartitionSnapshotStorage snapshotStorage, SnapshotUri snapshotUri) {
+        this.snapshotStorage = snapshotStorage;
+        this.snapshotUri = snapshotUri;
+    }
+
+    @Override
+    public void cancel() {
+        //TODO Implement.
```
sashapolo, line 75: Obviously this stuff has to have a ticket

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Override
+    public void join() throws InterruptedException {
+        //TODO Implement proper join.
+        threadPool.awaitTermination(1, TimeUnit.SECONDS);
```
sashapolo, line 81: How does this work? Do we call `join` after `close`?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Override
+    public void start() {
+        ClusterNode sourceNode = snapshotStorage.topologyService.getByConsistentId(snapshotUri.nodeName);
```
sashapolo, line 86: We should check that the source node can actually be resolved

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -0,0 +1,117 @@
...
+        threadPool.submit(() -> {
+            //TODO https://issues.apache.org/jira/browse/IGNITE-17262
+            // Following code is just an example of what I expect and shouldn't be considered a template.
+            CompletableFuture<NetworkMessage> metaRequestFuture = messagingService.invoke(
+                    sourceNode,
+                    MSG_FACTORY.snapshotMetaRequest().id(snapshotUri.snapshotId).build(),
+                    1000L
+            );
+
+            metaRequestFuture.whenCompleteAsync((networkMessage, throwable) -> {
```
sashapolo, line 99: Why `whenCompleteAsync`? You simply do a field assignment here

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -0,0 +1,117 @@
...
+    /** Rebalance thread-pool, used to write data into a storage. */
+    //TODO Use external pool.
+    private final ExecutorService threadPool = Executors.newSingleThreadExecutor();
+
+    /**
+     * Snapshot meta read from the leader.
+     *
+     * @see SnapshotMetaRequest
+     */
+    private SnapshotMeta snapshotMeta;
```
sashapolo, line 60: Looks like this field should be `volatile`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
```diff
@@ -0,0 +1,145 @@
...
+     * Constructor.
+     *
+     * @param messagingService Messaging service.
+     */
+    public OutgoingSnapshotsManager(MessagingService messagingService) {
+        this.messagingService = messagingService;
+    }
+
+    /**
+     * Returns a messagins service.
```
sashapolo, line 60: Sneaky little hobbitses

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
```diff
@@ -0,0 +1,145 @@
...
+    public void stop() throws Exception {
+    }
+
+    /**
+     * Registers an outgoing snapshot in the manager.
+     *
+     * @param snapshotId Snapshot id.
+     * @param outgoingSnapshot Outgoing snapshot.
+     */
+    void startOutgoingSnapshot(UUID snapshotId, OutgoingSnapshot outgoingSnapshot) {
```
sashapolo, line 81: Should it be called `registerOutgoiingSnapshot`?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
```diff
@@ -0,0 +1,145 @@
...
+        if (!(networkMessage instanceof SnapshotRequestMessage)) {
+            return;
+        }
+
+        assert correlationId != null;
+
+        OutgoingSnapshot outgoingSnapshot = outgoingSnapshots.get(((SnapshotRequestMessage) networkMessage).id());
+
+        if (outgoingSnapshot == null) {
+            if (LOG.isWarnEnabled()) {
```
sashapolo, line 105: No need for this `if` here

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
```diff
@@ -0,0 +1,145 @@
...
+                LOG.warn("Unexpected snapshot request message has been received [message={}]", networkMessage.toString());
```
sashapolo, line 106: I think `toString` will be called automatically if needed

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
```diff
@@ -0,0 +1,145 @@
...
+            }
+
+            return;
+        }
+
+        CompletableFuture<? extends NetworkMessage> responseFuture;
+
+        switch (networkMessage.messageType()) {
```
sashapolo, line 114: I would suggest to extract this into a separate method

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
```diff
@@ -0,0 +1,145 @@
...
+                break;
+
+            default:
+                return;
+        }
+
+        responseFuture.whenComplete((response, throwable) -> respond(response, throwable, sender, correlationId));
+    }
+
+    private CompletableFuture<Void> respond(
```
sashapolo, line 137: What's the point of this method? Do we need a TODO here?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
```diff
@@ -0,0 +1,145 @@
...
+import org.apache.ignite.internal.table.message.SnapshotRequestMessage;
+import org.apache.ignite.internal.table.message.SnapshotTxDataRequest;
+import org.apache.ignite.internal.table.message.TableMessageGroup;
+import org.apache.ignite.network.MessagingService;
+import org.apache.ignite.network.NetworkAddress;
+import org.apache.ignite.network.NetworkMessage;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Outgoing snapshots manager.
```
sashapolo, line 38: This is not a good javadoc

modules/table/src/main/java/org/apache/ignite/internal/table/message/SnapshotMvDataResponse.java
```diff
@@ -0,0 +1,58 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Snapshot partition data response message.
+ */
+@Transferable(TableMessageGroup.SNAPSHOT_MV_DATA_RESPONSE)
+public interface SnapshotMvDataResponse extends NetworkMessage {
+    Collection<SnapshotMvDataSingleResponse> rows();
+
+    boolean finish();
```
sashapolo, line 37: Please add a javadoc

modules/table/src/main/java/org/apache/ignite/internal/table/message/SnapshotMvDataResponse.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Single row response as a message.
+     */
+    @Transferable(TableMessageGroup.SNAPSHOT_MV_DATA_SINGLE_RESPONSE)
+    interface SnapshotMvDataSingleResponse extends NetworkMessage {
```
sashapolo, line 43: I think this class should be renamed to `ResponseEntry`, this is class is nested anyway

modules/table/src/main/java/org/apache/ignite/internal/table/message/SnapshotTxDataResponse.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.ignite.network.annotations.Marshallable;
+import org.apache.ignite.network.annotations.Transferable;
+
+/**
+ * Snapshot TX state partition data response message.
+ */
+@Transferable(TableMessageGroup.SNAPSHOT_TX_DATA_RESPONSE)
+public interface SnapshotTxDataResponse extends NetworkMessage {
+    @Marshallable
+    List<UUID> txIds();
```
sashapolo, line 33: Please add javadocs

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
```diff
@@ -419,6 +419,13 @@ public ReadResult next() {
         };
     }
 
+    @Override
+    public @Nullable RowId closestRowId(RowId lowerBound) throws StorageException {
+        Entry<RowId, VersionChain> mapEntry = map.tailMap(lowerBound).firstEntry();
```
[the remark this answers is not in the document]
ibessonov: Thank you for the hint! It's `ceilingKey` btw

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -963,11 +962,32 @@ public ReadResult next() {
...
             }
         };
     }
 
+    @Override
+    public @Nullable RowId closestRowId(RowId lowerBound) throws StorageException {
+        ByteBuffer keyBuf = prepareHeapKeyBuf(lowerBound);
+
+        try (RocksIterator it = db.newIterator(cf, scanReadOptions)) {
+            it.seek(copyOf(keyBuf.array(), ROW_PREFIX_SIZE));
```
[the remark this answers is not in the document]
ibessonov: About that... I guess I can, but documentation says that it allows direct buffers only
I'll do it, but it's a grey area for me

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -963,11 +962,32 @@ public ReadResult next() {
...
+
+            if (!it.isValid()) {
```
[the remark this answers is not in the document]
ibessonov: Do you suggest something? I thought that we'll get the exception thrown from the method if it's there

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/InitPartitionSnapshotReader.java
```diff
@@ -61,16 +60,14 @@ public Set<String> listFiles() {
 
     /** {@inheritDoc} */
     @Override
-    public @Nullable Message getFileMeta(String fileName) {
-        // No files in the snapshot.
-        return null;
+    public Message getFileMeta(String fileName) {
+        throw new UnsupportedOperationException("No files in the snapshot.");
```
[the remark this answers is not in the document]
ibessonov: We have so many exceptions with dots! I don't even know what convention to follow at this point

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionSnapshotStorage.java
```diff
@@ -33,36 +38,52 @@
  *
  * @see PartitionSnapshotStorageFactory
  */
-class PartitionSnapshotStorage implements SnapshotStorage {
+@SuppressWarnings("PublicField")
```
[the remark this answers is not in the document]
ibessonov: Forgot to do this, sorry

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/SnapshotUri.java
```diff
@@ -0,0 +1,97 @@
...
+    /** Sender node name. */
+    public final String nodeName;
+
+    /**
+     * Constructor.
+     *
+     * @param snapshotId Snapshot id.
+     * @param nodeName Sender node name.
+     */
+    public SnapshotUri(UUID snapshotId, String nodeName) {
```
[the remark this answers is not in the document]
ibessonov: There's no difference to be honest, but I'll do it private for now

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/SnapshotUri.java
```diff
@@ -0,0 +1,97 @@
...
+    /** Length of UUID's sting representation in characters. Includes 32 hexadecimals and 4 dashes. */
+    private static final int UUID_LENGTH = 36;
+
+    /**
+     * Creates a string representation of the snapshot URI.
+     *
+     * @param snapshotId Snapshot id.
+     * @param nodeName Sender node name.
+     */
+    public static String toStringUri(UUID snapshotId, String nodeName) {
```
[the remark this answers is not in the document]
ibessonov: Why creating an instance if you can create a string immediately?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
ibessonov: No, the other way around

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
ibessonov: We'll do that separately

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
ibessonov: This is not a final code, but I'll change it if you wish

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
ibessonov: Technically, it shouldn't. It's being read after `join` is called

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
ibessonov: Maybe

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
ibessonov: Why?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
ibessonov: Come on, it's not that big

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
ibessonov: The point is to implement it properly in the future. Component might be stopped, recipient may already be offline, and so on

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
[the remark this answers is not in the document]
sashapolo: Hm, you are correct, I thought floor key meant "greater or equal"

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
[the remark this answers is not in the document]
sashapolo: Looks like the documentation is lying, because the implementation clearly can handle that

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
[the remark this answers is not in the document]
sashapolo: > I thought that we'll get the exception thrown from the method if it's there

No, it won't, that's why you need to check the iterator's status every time

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/InitPartitionSnapshotReader.java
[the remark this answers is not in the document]
sashapolo: Please no dots

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/SnapshotUri.java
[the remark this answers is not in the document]
sashapolo: Because it looks better =) `Integer.toString` is a bad example, because Integer semantically equivalent to int and it only accepts a single parameter, which makes it clear to use. Nevertheless, it's your call, I won't insist

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
sashapolo: Where? Is there a ticket?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
sashapolo: Do you mean `join` that calls `awaitTermination`? Do you mean that you rely on `whenCompleteAsync` to be launched on the executors threads so that results of the lambda inside will be published after their termination?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
sashapolo: Hm, does `awaitTermination` wait for any tasks to complete even if the shoutdown request was not issued?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
sashapolo: because `Log.warn` will check the same condition internally

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
sashapolo: I stand by my suggestion)

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
sashapolo: Let's add a TODO then

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/AbstractPageMemoryMvPartitionStorage.java
```diff
@@ -684,14 +684,10 @@ private Cursor<BinaryRow> internalScan(Predicate<BinaryRow> keyFilter, UUID txId
     @Override
     public @Nullable RowId closestRowId(RowId lowerBound) throws StorageException {
         try {
-            Cursor<VersionChain> cursor = versionChainTree.find(new VersionChainKey(lowerBound), null);
-
-            if (!cursor.hasNext()) {
-                return null;
+            try (Cursor<VersionChain> cursor = versionChainTree.find(new VersionChainKey(lowerBound), null)) {
```
sashapolo, line 687: Do you need a nested `try` here? Can it be merged with the outer `try`?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
```diff
@@ -103,12 +103,23 @@ private void messageHandler(NetworkMessage networkMessage, NetworkAddress sender
...
+        if (responseFuture != null) {
+            responseFuture.whenComplete((response, throwable) -> respond(response, throwable, sender, correlationId));
+        }
+    }
+
+    private static @Nullable CompletableFuture<? extends NetworkMessage> handleSnapshotRequestMessage(
+            NetworkMessage networkMessage,
+            OutgoingSnapshot outgoingSnapshot
+    ) {
         CompletableFuture<? extends NetworkMessage> responseFuture;
```
sashapolo, line 123: This variable is now redundant

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/AbstractPageMemoryMvPartitionStorage.java
ibessonov: I'll check

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
ibessonov: true

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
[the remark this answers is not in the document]
ibessonov: BTW, I guess we can remove all "copyOf" and "copyOfRange" at this point, I'll do some day

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
ibessonov: I did the extraction

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
ibessonov: This applies to all methods, like `info` and `error`, but manual check saves on CPU. It's been shown many times

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/SnapshotUri.java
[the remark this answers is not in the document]
ibessonov: I mean, `toStringUri(x, y)` is better than `new SnapshotUri(x, y).toString()`, literally the same example

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
ibessonov: I mean that right now it's a placeholder, and "technically" (once again) I was wrong, because the implementation is sh*t, but in the future it's reasonable to expect such field without `volatile`, because `join` must give us the visibility

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
ibessonov: It's a placeholder, I should explicitly state that in the code, or simply remove it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
sashapolo: > but manual check saves on CPU. It's been shown many times

Can you please point me to where it is shown?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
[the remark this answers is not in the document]
ibessonov: I see, I have to check status

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/InitPartitionSnapshotReader.java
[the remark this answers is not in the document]
ibessonov: We'll have to fix them everywhere then

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManager.java
ibessonov: No, it's classified)
I pulled that from my memory. Usually this discussion appears when we talk about `debug` logs, but same reasoning applies to everything else

sashapolo (approved):

PR #1162, opened 2022-10-05
IGNITE-17825 ThreadPoolMetricTest.test is flaky

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-17825

### DIFF
--- a/modules/metrics/pom.xml
+++ b/modules/metrics/pom.xml
@@ -85,6 +85,13 @@
             <scope>test</scope>
             <type>test-jar</type>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.ignite</groupId>
+            <artifactId>ignite-core</artifactId>
+            <scope>test</scope>
+            <type>test-jar</type>
+        </dependency>
     </dependencies>
 
     <build>
--- a/modules/metrics/src/test/java/org/apache/ignite/internal/metrics/ThreadPoolMetricTest.java
+++ b/modules/metrics/src/test/java/org/apache/ignite/internal/metrics/ThreadPoolMetricTest.java
@@ -17,9 +17,9 @@
 
 package org.apache.ignite.internal.metrics;
 
+import static org.apache.ignite.internal.testframework.IgniteTestUtils.waitForCondition;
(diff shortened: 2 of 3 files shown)


modules/metrics/src/test/java/org/apache/ignite/internal/metrics/ThreadPoolMetricTest.java
```diff
@@ -54,8 +54,9 @@ public void test() throws ExecutionException, InterruptedException {
 
         assertEquals(0L, completedTaskCount.value());
 
-        exec.submit(() -> {}).get();
-        assertEquals(1L, completedTaskCount.value());
+        exec.submit(() -> {}).get(1, TimeUnit.SECONDS);
+
+        waitForCondition(() -> completedTaskCount.value() > 0, 10, TimeUnit.SECONDS.toMillis(1));
```
sashapolo, line 59: There should probable be an `assertTrue` here

sashapolo: Also, the code in the end should be placed in either `finally` block, or `afterEach` block

tkalkirill: Fix it. thanks

modules/metrics/pom.xml
```diff
@@ -85,6 +85,13 @@
             <scope>test</scope>
             <type>test-jar</type>
         </dependency>
+
+        <dependency>
```
ibessonov, line 89: Please reflect this change in Gradle file as well

modules/metrics/src/test/java/org/apache/ignite/internal/metrics/ThreadPoolMetricTest.java
tkalkirill: Fix it all

modules/metrics/pom.xml
tkalkirill: Fix it

sashapolo (approved):

PR #1164, opened 2022-10-05
IGNITE-17827 Classes moved from "network" to "network-api"

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-17827


SammyVimes (approved):
LGTM!

PR #1168, opened 2022-10-06
Ignite-17834 Fixed.

sanpwc:
Signed-off-by: alapin <lapin1702@gmail.com>

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
@@ -415,7 +415,17 @@ public class InternalTableImpl implements InternalTable {
                         throw (RuntimeException) e;
                     }); // Preserve failed state.
                 } else {
-                    tx0.enlistResultFuture(fut);
+                    try {
+                        tx0.enlistResultFuture(fut);
+                    } catch (TransactionException enlistResultException) {
+                        return tx0.rollbackAsync().handle((ignored, err) -> {
+                            if (err != null) {
+                                e.addSuppressed(err);
+                            }
+
+                            throw enlistResultException;
+                        });
+                    }
 
                     return implicit ? tx0.commitAsync().thenApply(ignored -> r) : completedFuture(r);
                 }
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
@@ -415,6 +415,7 @@ public class InternalTableImpl implements InternalTable {
                         throw (RuntimeException) e;
                     }); // Preserve failed state.
                 } else {
+<<<<<<< HEAD
                     try {
                         tx0.enlistResultFuture(fut);
                     } catch (TransactionException enlistResultException) {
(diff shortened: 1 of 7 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -415,7 +415,17 @@ public CompletableFuture<T> apply(T r, Throwable e) {
                         throw (RuntimeException) e;
                     }); // Preserve failed state.
                 } else {
-                    tx0.enlistResultFuture(fut);
+                    try {
+                        tx0.enlistResultFuture(fut);
+                    } catch (TransactionException enlistResultException) {
+                        return tx0.rollbackAsync().handle((ignored, err) -> {
```
vldpyatkov, line 421: Why are you rolling back the transaction if it is already finished?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TransactionImpl.java
```diff
@@ -54,10 +56,13 @@ public class TransactionImpl implements InternalTransaction {
...
     /** Enlisted replication groups: replication group id -> (primary replica node, raft term). */
-    private Map<String, IgniteBiTuple<ClusterNode, Long>> enlisted = new ConcurrentSkipListMap<>();
+    private final Map<String, IgniteBiTuple<ClusterNode, Long>> enlisted = new ConcurrentSkipListMap<>();
 
     /** Enlisted operation futures in this transaction. */
-    private volatile List<CompletableFuture<?>>  enlistedResults = new ArrayList<>();
+    private final List<CompletableFuture<?>> enlistedResults = new ArrayList<>();
+
+    /** Guard that prevents finishing or enlisting new operations into already finished or finishing transaction. */
+    private boolean txInFinishState = false;
```
vldpyatkov, line 65: Way do not you want to make txInFinishState to AtomicBoolean and get rid of synchronized in the finish method?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -415,7 +415,17 @@ public CompletableFuture<T> apply(T r, Throwable e) {
...
+                            if (err != null) {
+                                e.addSuppressed(err);
```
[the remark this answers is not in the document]
sanpwc: Fixed.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TransactionImpl.java
sanpwc: Because it's required not only to change txInFinishState atomically but also to populate enlistedResults with new value
```
    public synchronized void enlistResultFuture(CompletableFuture<?> resultFuture) {
        if (txInFinishState) {
            throw new TransactionException(TX_ALREADY_FINISHED_ERR,
                    "Failed to enlist operation into already finished transaction txId={" + id + '}');
        }

        enlistedResults.add(resultFuture);
    }
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
sanpwc: Because it's required to cleanup locks and writeIntents for operation that failed to enlistResultFuture.

vldpyatkov: So that will not happen because the transaction is already finished.

sanpwc: You are right we need an escape path in order to rollback partially finished transaction.

PR #1169, opened 2022-10-06
IGNITE-17249 DefaultMessagingService incorrect message sending order

rpuch:
https://issues.apache.org/jira/browse/IGNITE-17249


modules/network/src/main/java/org/apache/ignite/internal/network/netty/NettyUtils.java
```diff
@@ -43,7 +45,26 @@ public static <T, R, F extends Future<R>> CompletableFuture<T> toCompletableFutu
...
+     * @param <R>         Netty future result type.
+     * @param <F>         Netty future type.
+     * @return CompletableFuture.
+     */
+    public static <T, R, F extends Future<R>> CompletableFuture<T> toCompletableFuture(
+            @Schedule F nettyFuture,
+            Function<F, T> mapper,
+            Supplier<? extends CompletableFuture<T>> completableFutureFactory
+    ) {
+        var fut = completableFutureFactory.get();
```
SammyVimes, line 67: we can only use `var` for `var smth = new Smth` statements

modules/network/src/test/java/org/apache/ignite/network/DefaultMessagingServiceTest.java
```diff
@@ -0,0 +1,223 @@
...
+            this.connectionManager = connectionManager;
+            this.messagingService = messagingService;
+        }
+
+        @Override
+        public void close() throws Exception {
+            IgniteUtils.closeAll(connectionManager::stop, messagingService::stop);
+        }
+    }
+}
```
SammyVimes, line 223: Missing newline

modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
```diff
@@ -92,6 +92,9 @@ public class ConnectionManager {
     /** Recovery descriptor provider. */
     private final RecoveryDescriptorProvider descriptorProvider = new DefaultRecoveryDescriptorProvider();
 
+    /** An action that is invoked before handshake completes. Only useful for tests. */
+    private volatile Runnable beforeHandshakeComplete = () -> {};
```
SammyVimes, line 96: I think we should avoid test code in prod code (except for getters, probably). Shall we maybe leverage the factory pattern to create different handshake handler?

SammyVimes (changes requested):

modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
rpuch: `RecoveryClientHandshakeManager` constructor currently requires 5 arguments. We can pass message factory directly, but we would still need to create a factory with 4 arguments which makes the factory sooo specific. I'm not sure whether it's the best way to go, but I'll try and we'll look at the result.

rpuch: Done. Does this look botter, WDYT?

SammyVimes (approved):
LGTM

PR #1170, opened 2022-10-06
IGNITE-17832 Fix flaky ItClusterManagerTest#testNodeRestart

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-17832


modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/RocksDbClusterStateStorage.java
```diff
@@ -108,6 +108,22 @@ public void put(byte[] key, byte[] value) {
         }
     }
 
+    @Override
+    public void replaceAll(byte[] prefix, byte[] key, byte[] value) {
+        try (
+                var batch = new WriteBatch();
+                var options = new WriteOptions();
+        ) {
+            batch.deleteRange(prefix, RocksUtils.incrementArray(prefix));
```
rpuch, line 117: It seems that, if `prefix` consists of `0xFF` bytes only, `incrementArray()` will return `null` and `deleteRange()` will fail.

For instance, the prefix is {0xFF}, it contains key {0xFF, 0x5}, and we want to replace it with {0xFF, 0x10}.

Should a scan be added for such unfortunate prefixes?

modules/cluster-management/src/test/java/org/apache/ignite/internal/cluster/management/raft/AbstractClusterStateStorageTest.java
```diff
@@ -110,6 +110,50 @@ void testPutReplace() {
         assertThat(storage.get(key), is(equalTo(value2)));
     }
 
+    @Test
+    void testReplaceAll() {
```
rpuch, line 114: This test actually contains 3 test cases for 3 properties of the tested method. How about breaking the test method to 3, one per test case?

modules/cluster-management/src/testFixtures/java/org/apache/ignite/internal/cluster/management/raft/TestClusterStateStorage.java
```diff
@@ -62,62 +69,89 @@ public boolean isStarted() {
 
     @Override
     public void put(byte[] key, byte[] value) {
-        map.put(new ByteArray(key), value);
+        lock.readLock().lock();
```
rpuch, line 72: This is a modification operations, should write lock be taken?

rpuch (changes requested):

modules/cluster-management/src/testFixtures/java/org/apache/ignite/internal/cluster/management/raft/TestClusterStateStorage.java
sashapolo: ooops, this is a typo

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/RocksDbClusterStateStorage.java
sashapolo: This is currently impossible, because we always have a prefix that is does not consist of 0xFFs. I can add an assertion if you like

modules/cluster-management/src/test/java/org/apache/ignite/internal/cluster/management/raft/AbstractClusterStateStorageTest.java
sashapolo: we have a lot of similar situations in other tests. In this case they are very small, do you think we will benefit from that?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/RocksDbClusterStateStorage.java
rpuch: Yes, please, it would be great to have an assertion

rpuch (approved):

PR #1171, opened 2022-10-06
IGNITE-17777 Thin 3.0: use BinaryTuple for Compute and SQL results and arguments

ptupitsyn:
Use BinaryTuple for:
* Compute arguments and results
* SQL arguments, results, and properties

As a result, MsgPack is no longer used to encode user data and does not need to support Ignite data types. Unused code removed.

.NET, Java, JDBC clients updated.

Additionally:
* Fixed decimal serialization in .NET


modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientBinaryTupleUtils.java
```diff
@@ -49,72 +79,251 @@ public static void readAndSetColumnValue(
         }
 
         switch (clientDataType) {
-            case ClientDataType.INT8:
+            case INT8:
```
isapego, line 82: Is importing all the enum values separately required by the coding convention?

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientBinaryTupleUtils.java
```diff
@@ -49,72 +79,251 @@ public static void readAndSetColumnValue(
...
+     */
+    public static void appendObject(BinaryTupleBuilder builder, Object obj) {
+        if (obj == null) {
+            builder.appendNull();
+            builder.appendInt(0);
+            builder.appendNull();
+        } else if (obj instanceof Byte) {
+            builder.appendInt(ClientDataType.INT8);
+            builder.appendInt(0);
+            builder.appendByte((Byte) obj);
```
isapego, line 243: Are those `type`, `scale` and `value`? Maybe we should add a function with 3 args just so it's more clear what is written.

modules/platforms/dotnet/Apache.Ignite/Internal/Proto/BinaryTuple/BinaryTupleBuilder.cs
```diff
@@ -534,6 +518,126 @@ public void AppendObject(object? value, ClientDataType colType, int scale)
...
+        /// </summary>
+        /// <param name="value">Value.</param>
+        public void AppendObjectWithType(object? value)
+        {
+            switch (value)
+            {
+                case null:
+                    AppendNull();
+                    AppendNull();
+                    AppendNull();
```
isapego, line 532: The same comment as for Java. Also, there is some mismatch to Java packing for `null`

isapego (approved):

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientBinaryTupleUtils.java
ptupitsyn: Coding convention does not mandate this, as far as I understand.

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientBinaryTupleUtils.java
ptupitsyn: Good point, done.

modules/platforms/dotnet/Apache.Ignite/Internal/Proto/BinaryTuple/BinaryTupleBuilder.cs
ptupitsyn: Fixed.

isapego (approved):

PR #1177, opened 2022-10-07
IGNITE-17263 Leader to replica safe time propagation implemented

denis-chudov:


modules/replicator/src/main/java/org/apache/ignite/internal/replicator/Replica.java
```diff
@@ -59,6 +83,28 @@ public CompletableFuture<Object> processRequest(ReplicaRequest request) {
...
+
+    /**
+     * Propagates safe time on non-primary replicas, if this replica is primary.
+     */
+    public void propagateSafeTime() {
+        if (!isPrimary()) {
+            return;
+        }
 
+        raftClient.run(new SafeTimeSyncCommand());
```
vldpyatkov, line 108: I think, that the Replica can only to invoke a replica listener. It shouldn't invoke RAFT as is.

vldpyatkov: As we discussed, it will be better to invoke replica listener here (with specific SafeImeSycnReplicaRequest) and the listener invokes RAFT. Because the approach doesn't add a particular protocol reference to the replication layer.

modules/core/src/main/java/org/apache/ignite/hlc/HybridClock.java
```diff
@@ -40,13 +47,25 @@ public class HybridClock {
...
-        this.latestTime = new HybridTimestamp(Clock.systemUTC().instant().toEpochMilli(), 0);
+        this.currentTimeMillisProvider = () -> Clock.systemUTC().instant().toEpochMilli();
+
+        this.latestTime = new HybridTimestamp(currentTimeMillisProvider.get(), 0);
+    }
+
+    /**
+     * The constructor which initializes the latest time to the given provider.
+     */
+    @TestOnly
```
sanpwc, line 64: I don't like an idea of adding test related code to HybridClock, neither @TestOnly nor protected along with all suppliers.

modules/core/src/main/java/org/apache/ignite/hlc/HybridClock.java
```diff
@@ -40,13 +47,25 @@ public class HybridClock {
     }
 
     /** Latest timestamp. */
-    private volatile HybridTimestamp latestTime;
+    protected volatile HybridTimestamp latestTime;
```
sanpwc, line 50: I don't like an idea of adding test related code to HybridClock. Neither supplier nor protected or TestOnly. I'd rather extract HybridClock interface with several implementaions, e.g. common HybridClockImpl and TrackableHybridClock in test package. What do u think, will it work?

PR #1179, opened 2022-10-07
IGNITE-17855 Implement inline size calculation for B+tree

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-17855

### DIFF
--- a/modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
+++ b/modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
@@ -141,33 +141,23 @@ public class InlineUtils {
         if (indexDescriptor.columns().stream().anyMatch(c -> !c.type().spec().fixedLength())) {
             int itemSize = binaryTupleInlineSize + itemHeaderSize;
 
-            // Let's calculate how much space remains in the BplusInnerIo and the BplusLeafIo.
-            int remainingInnerNodePayloadSize = (innerNodePayloadSize(pageSize) - CHILD_LINK_SIZE) % (itemSize + CHILD_LINK_SIZE);
-            int remainingLeafNodePayloadSize = leafNodePayloadSize(pageSize) % itemSize;
+            int innerNodeItemSize =
+                    optimizeItemSize(innerNodePayloadSize(pageSize) - CHILD_LINK_SIZE, itemSize + CHILD_LINK_SIZE) - CHILD_LINK_SIZE;
 
-            if (remainingInnerNodePayloadSize > 0 && remainingLeafNodePayloadSize > 0) {
-                int innerNodeItemCount = (innerNodePayloadSize(pageSize) - CHILD_LINK_SIZE) / (itemSize + CHILD_LINK_SIZE);
-                int leafNodeItemCount = leafNodePayloadSize(pageSize) / itemSize;
+            int leafNodeItemSize = optimizeItemSize(leafNodePayloadSize(pageSize), itemSize);
 
-                // Let's calculate how many additional bytes can be added for each item of the BplusInnerIo and the BplusLeafIo.
-                int additionalInnerNodeItemBytes = remainingInnerNodePayloadSize / innerNodeItemCount;
-                int additionalLeafNodeItemBytes = remainingLeafNodePayloadSize / leafNodeItemCount;
+            int optimizedItemSize = Math.min(innerNodeItemSize, leafNodeItemSize);
 
-                if (additionalInnerNodeItemBytes > 0 && additionalLeafNodeItemBytes > 0) {
-                    itemSize += Math.min(additionalInnerNodeItemBytes, additionalLeafNodeItemBytes);
+            assert leafNodePayloadSize(pageSize) / itemSize == leafNodePayloadSize(pageSize) / optimizedItemSize;
 
-                    assert leafNodePayloadSize(pageSize) / itemSize == leafNodeItemCount;
-
-                    binaryTupleInlineSize = itemSize - itemHeaderSize;
-                }
-            }
+            binaryTupleInlineSize = optimizedItemSize - itemHeaderSize;
         }
 
         return Math.min(binaryTupleInlineSize, MAX_BINARY_TUPLE_INLINE_SIZE);
     }
 
     /**
-     * Returns number of bytes that can be used to store items and references to child nodes in an inner node.
+     * Returns number of bytes that can be used to store items and links to child nodes in an inner node.
      *
      * @param pageSize Page size in bytes.
      */
@@ -182,5 +172,26 @@ public class InlineUtils {
      */
     static int leafNodePayloadSize(int pageSize) {
         return pageSize - BplusLeafIo.HEADER_SIZE;
+    }
+
+    /**
+     * Optimizes the item size for a {@link BplusInnerIo} or {@link BplusLeafIo} if there is free space for each item.
+     *
+     * <p>We try to use the available memory on the page as much as possible.
+     *
+     * @param nodePayloadSize Payload size of a {@link BplusInnerIo} or {@link BplusLeafIo}, in bytes.
+     * @param itemSize Size of the item in {@link BplusInnerIo} or {@link BplusLeafIo}, in bytes.
+     * @return Size in bytes.
+     */
+    static int optimizeItemSize(int nodePayloadSize, int itemSize) {
+        // Let's calculate how much space remains in the BplusInnerIo or the BplusLeafIo.
+        int remainingNodePayloadSize = nodePayloadSize % itemSize;
+
+        int nodeItemCount = nodePayloadSize / itemSize;
+
+        // Let's calculate how many additional bytes can be added for each item of the BplusInnerIo and the BplusLeafIo.
+        int additionalNodeItemBytes = remainingNodePayloadSize / nodeItemCount;
+
+        return itemSize + additionalNodeItemBytes;
     }
 }
--- a/modules/storage-page-memory/src/test/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtilsTest.java
+++ b/modules/storage-page-memory/src/test/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtilsTest.java
@@ -25,6 +25,7 @@ import static org.apache.ignite.internal.storage.pagememory.index.InlineUtils.bi
 import static org.apache.ignite.internal.storage.pagememory.index.InlineUtils.inlineSize;
 import static org.apache.ignite.internal.storage.pagememory.index.InlineUtils.innerNodePayloadSize;
 import static org.apache.ignite.internal.storage.pagememory.index.InlineUtils.leafNodePayloadSize;
+import static org.apache.ignite.internal.storage.pagememory.index.InlineUtils.optimizeItemSize;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.empty;
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -268,6 +269,13 @@ public class InlineUtilsTest {
                 BinaryTupleCommon.HEADER_SIZE + 1 + 5 + NativeTypes.INT64.sizeInBytes() + 3 * NativeTypes.UUID.sizeInBytes() + 32 + 6,
                 binaryTupleInlineSize(pageSize, itemHeaderSize, indexDescriptor)
         );
+    }
+
+    @Test
+    void testOptimizeItemSize() {
+        assertEquals(100, optimizeItemSize(1000, 100));
+
+        assertEquals(333, optimizeItemSize(1000, 330));
     }
 
     private static IndexDescriptor testIndexDescriptor(ColumnDescriptor... columnDescriptors) {


modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/InlineUtils.java
```diff
@@ -0,0 +1,106 @@
...
+            return nativeType.sizeInBytes();
+        }
+
+        // Variable length columns.
+
+        switch (spec) {
+            case STRING: {
+                int length = ((VarlenNativeType) nativeType).length();
+
+                return length == Integer.MAX_VALUE ? UNDEFINED_VARLEN_INLINE_SIZE : length * 2;
```
AMashenkov, line 67:
```suggestion
                return length == Integer.MAX_VALUE ? UNDEFINED_VARLEN_INLINE_SIZE : length;
```
Do you expect most of users will use 2 bytes chars? 
Why should an index over columns of ASCII chars be twice large by default?

tkalkirill: I proceed from the fact that in Java characters is `char` (2 bytes).
We can assume, as in Java, that in most cases the characters will be ascii (1 byte).
We're just doing an approximate inline count, I don't know how often we'll see 4 byte characters.

Later we can add an encoding (utf-8, 16, 32) to the configuration or ddl.

wdyt?

AMashenkov: ok, also I'd limit default value for var-length types (e.g. with 64-256 bytes).

Usually, the upper limit for var-length columns is usually calculated with some reserve. 
So, the actual values size is less the limit (often noticeably).
Also, large inline-size means less data will fit to the index page and may have negative impact on performance.

In case of relatively large values with similar prefixes user always can increase inline-size manually, 
but imho, in most of cases, this limit will reduce index size with no performance penalty.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,147 @@
...
+
+        // Variable length columns.
+
+        switch (spec) {
+            case STRING:
+            case BYTES:
+                return Math.min(maxVarlenInlineSize, ((VarlenNativeType) nativeType).length());
+
+            case DECIMAL:
+                // It can be converted to float for greater or lesser comparisons.
```
ibessonov, line 64: What do you mean by this comment?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,147 @@
...
+                return Float.BYTES;
+
+            case NUMBER:
+                // It can be converted to int for greater or lesser comparisons.
```
ibessonov, line 68: And by this. I don't get it

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,147 @@
...
+    public static final int MAX_BINARY_TUPLE_INLINE_SIZE = 2 * KiB;
+
+    /**
+     * Calculates inline size for column.
+     *
+     * @param nativeType Column type.
+     * @param maxVarlenInlineSize Maximum inline size of variable length column in bytes.
+     * @return Inline size in bytes.
+     */
+    static int inlineSize(NativeType nativeType, int maxVarlenInlineSize) {
```
ibessonov, line 49: Why is `maxVarlenInlineSize` a parameter?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,147 @@
...
+     * @param indexDescriptor Index descriptor.
+     * @param maxVarlenInlineSize Maximum inline size of variable length column in bytes.
+     * @return Inline size in bytes, not more than the {@link #MAX_BINARY_TUPLE_INLINE_SIZE}.
+     */
+    static int binaryTupleInlineSize(IndexDescriptor indexDescriptor, int maxVarlenInlineSize) {
+        List<? extends ColumnDescriptor> columns = indexDescriptor.columns();
+
+        boolean hasNullColumns = columns.stream().anyMatch(ColumnDescriptor::nullable);
+
+        // 2-byte size class covers most cases.
```
ibessonov, line 88: What is "class" here? Looks more like offset size

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,147 @@
...
+        int sizeClass = 2;
+
+        if (columns.stream().allMatch(c -> c.type().spec().fixedLength())) {
+            int totalFixedColumnSize = columns.stream().mapToInt(c -> c.type().sizeInBytes()).sum();
+
+            sizeClass = Math.min(sizeClass, BinaryTupleCommon.flagsToEntrySize(BinaryTupleCommon.valueSizeToFlags(totalFixedColumnSize)));
```
ibessonov, line 94: This line is so messy. We can use static imports or extract some variables, what do you think?

ibessonov: Also, it's a double conversion, not very intuitive. Can we have "valueSizeToEntrySize"?
BTW, entry size if better than "class"

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,147 @@
...
+     * @param maxVarlenInlineSize Maximum inline size of variable length column in bytes.
+     * @return Inline size in bytes, not more than the {@link #MAX_BINARY_TUPLE_INLINE_SIZE}.
+     */
+    public static int binaryTupleBplusIoInlineSize(
+            int pageSize,
+            int indexBplusIoHeaderSize,
+            IndexDescriptor indexDescriptor,
+            int maxVarlenInlineSize
+    ) {
+        int innerNodePayloadSize = pageSize - BplusInnerIo.HEADER_SIZE - PARTITIONLESS_LINK_SIZE_BYTES;
```
ibessonov, line 124: I think we should introduce a constant like `CHILD_LINK_SIZE` in BplusInnerIo, what do you think? Having explicit mentions of partitionless links is too mouthful

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,147 @@
...
+        int leafNodePayloadSize = pageSize - BplusLeafIo.HEADER_SIZE;
+
+        // So that at least 2 items fit on the BplusInnerIo.
+        int maxInnerNodeItemSize = (innerNodePayloadSize / 2) - PARTITIONLESS_LINK_SIZE_BYTES;
+
+        int itemSize = Math.min(maxInnerNodeItemSize, binaryTupleInlineSize(indexDescriptor, maxVarlenInlineSize) + indexBplusIoHeaderSize);
+
+        // If there is a variable length column.
+        if (indexDescriptor.columns().stream().anyMatch(not(c -> c.type().spec().fixedLength()))) {
```
ibessonov, line 133: Or you could write `!foo.allMatch(c -> c...fixedLength())`, would be easier to understand

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,147 @@
...
+
+        return Math.min(inlineSize, MAX_BINARY_TUPLE_INLINE_SIZE);
+    }
+
+    /**
+     * Calculates the inline size for the {@link BinaryTuple} that will be stored in the index {@link BplusInnerIo} and {@link BplusLeafIo}.
+     * Minimum number of items in a {@link BplusInnerIo} is {@code 2}, so that there is no significant performance drop.
+     *
+     * @param pageSize Page size in bytes.
+     * @param indexBplusIoHeaderSize Index {@link BplusInnerIo} and {@link BplusLeafIo} header size in bytes.
```
ibessonov, line 113: Looking at this comment, I don't understand what it means. I think the word "header" might be misleading. Please expand the comment and maybe remove the parameter.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,147 @@
...
+
+        // So that at least 2 items fit on the BplusInnerIo.
+        int maxInnerNodeItemSize = (innerNodePayloadSize / 2) - PARTITIONLESS_LINK_SIZE_BYTES;
+
+        int itemSize = Math.min(maxInnerNodeItemSize, binaryTupleInlineSize(indexDescriptor, maxVarlenInlineSize) + indexBplusIoHeaderSize);
+
+        // If there is a variable length column.
+        if (indexDescriptor.columns().stream().anyMatch(not(c -> c.type().spec().fixedLength()))) {
+            // Let's try to increase itemSize for BplusInnerIo.
+            int additionalInnerNodeItemSize = (innerNodePayloadSize % (itemSize + PARTITIONLESS_LINK_SIZE_BYTES))
```
ibessonov, line 135: This code is really hard to read without comments. I understand why these calculations are here, but othe people may have struggles

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,147 @@
...
+                    / (innerNodePayloadSize / (itemSize + PARTITIONLESS_LINK_SIZE_BYTES));
+
+            // Let's check if we can increase itemSize for BplusInnerIo so that the number of items for BplusLeafIo does not decrease.
+            if (additionalInnerNodeItemSize > 0
```
ibessonov, line 139: This doesn't match what we discussed. I proposed choosing minimal value from two calculated deltas. But you either use a delta from inner page or a 0. Please fix it. Please also try to simplify this code by introducing some variables, like "innerPageMaxItems" or whatever.
By the way, why do you call them nodes here? Usually we refer to them as pages

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.ignite.internal.schema.NativeType;
+import org.apache.ignite.internal.schema.NativeTypeSpec;
+import org.apache.ignite.internal.schema.VarlenNativeType;
+import org.apache.ignite.internal.storage.index.IndexDescriptor;
+import org.apache.ignite.internal.storage.index.IndexDescriptor.ColumnDescriptor;
+
+/**
+ * Helper class for index inlining.
+ */
+public class InlineUtils {
```
sashapolo, line 38: Most methods in this class a package-private, while this class is public.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
sashapolo: Agree, I've had hard time understanding this

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
tkalkirill: Remove it.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
tkalkirill: Remove it.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
tkalkirill: Remove it.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
tkalkirill: Fix it.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
tkalkirill: Fix it.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
tkalkirill: Add it

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
tkalkirill: Fix it

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
tkalkirill: Fix it

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
tkalkirill: Fix it

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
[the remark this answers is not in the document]
tkalkirill: fix it

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,147 @@
...
+    static int binaryTupleInlineSize(IndexDescriptor indexDescriptor, int maxVarlenInlineSize) {
+        List<? extends ColumnDescriptor> columns = indexDescriptor.columns();
+
+        boolean hasNullColumns = columns.stream().anyMatch(ColumnDescriptor::nullable);
+
+        // 2-byte size class covers most cases.
+        int sizeClass = 2;
+
+        if (columns.stream().allMatch(c -> c.type().spec().fixedLength())) {
+            int totalFixedColumnSize = columns.stream().mapToInt(c -> c.type().sizeInBytes()).sum();
```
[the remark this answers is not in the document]
tkalkirill: fix it

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.ignite.internal.pagememory.tree.io.BplusLeafIo;
+import org.apache.ignite.internal.schema.BinaryTuple;
+import org.apache.ignite.internal.schema.NativeType;
+import org.apache.ignite.internal.schema.NativeTypeSpec;
+import org.apache.ignite.internal.schema.VarlenNativeType;
+import org.apache.ignite.internal.storage.index.IndexDescriptor;
+import org.apache.ignite.internal.storage.index.IndexDescriptor.ColumnDescriptor;
+
+/**
+ * Helper class for index inlining.
```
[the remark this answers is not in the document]
tkalkirill: This class will be used in VolatilePageMemoryTableStorage and PersistentPageMemoryTableStorage.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
tkalkirill: This class will be used in VolatilePageMemoryTableStorage and PersistentPageMemoryTableStorage.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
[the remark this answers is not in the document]
tkalkirill: Will it fit?

`Index inlining is an optimization that allows index rows (or prefix) to be compared in a B+tree without loading them from a FreeList.`

modules/storage-page-memory/src/test/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtilsTest.java
```diff
@@ -0,0 +1,290 @@
...
+        assertEquals(3, inlineSize(nativeType = NativeTypes.DATE));
+        nativeTypeSpecs.remove(nativeType.spec());
+
+        assertEquals(4, inlineSize(nativeType = NativeTypes.time()));
+        nativeTypeSpecs.remove(nativeType.spec());
+
+        assertEquals(9, inlineSize(nativeType = NativeTypes.datetime()));
+        nativeTypeSpecs.remove(nativeType.spec());
+
+        assertEquals(12, inlineSize(nativeType = NativeTypes.timestamp()));
```
ibessonov, line 91: 12 bytes per timestamp? Jesus! Why is it so big? Just asking

modules/storage-page-memory/src/test/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtilsTest.java
```diff
@@ -0,0 +1,290 @@
...
+        assertEquals(BIG_NUMBER_INLINE_SIZE, inlineSize(nativeType = NativeTypes.decimalOf(1, 1)));
+        nativeTypeSpecs.remove(nativeType.spec());
+
+        assertEquals(BIG_NUMBER_INLINE_SIZE, inlineSize(nativeType = NativeTypes.decimalOf(100, 1)));
+        nativeTypeSpecs.remove(nativeType.spec());
+
+        assertEquals(7, inlineSize(nativeType = NativeTypes.stringOf(7)));
+        nativeTypeSpecs.remove(nativeType.spec());
+
+        assertEquals(MAX_VARLEN_INLINE_SIZE, inlineSize(nativeType = NativeTypes.stringOf(Integer.MAX_VALUE)));
```
ibessonov, line 105: What about the string with length of 256?

modules/storage-page-memory/src/test/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtilsTest.java
```diff
@@ -0,0 +1,290 @@
...
+        nativeTypeSpecs.remove(nativeType.spec());
+
+        assertEquals(9, inlineSize(nativeType = NativeTypes.blobOf(9)));
+        nativeTypeSpecs.remove(nativeType.spec());
+
+        assertEquals(MAX_VARLEN_INLINE_SIZE, inlineSize(nativeType = NativeTypes.blobOf(Integer.MAX_VALUE)));
```
ibessonov, line 111: Same for blobs

modules/storage-page-memory/src/test/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtilsTest.java
```diff
@@ -0,0 +1,290 @@
...
+
+        // Let's check that all types have been checked.
+        assertThat(nativeTypeSpecs, empty());
+    }
+
+    @Test
+    void testBinaryTupleInlineSize() {
+        IndexDescriptor indexDescriptor = testIndexDescriptor();
+
+        assertEquals(BinaryTupleCommon.HEADER_SIZE, binaryTupleInlineSize(indexDescriptor));
```
ibessonov, line 128: This is funny, you know. I'd expect an exception, who would want to make a tuple with 0 columns.
On the other hand, can this be useful somewhere? Just asking

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
tkalkirill: Yep

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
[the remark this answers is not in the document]
tkalkirill: Add it.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/io/BplusInnerIo.java
```diff
@@ -44,6 +44,9 @@
     /** Offset of the right page ID of the item. */
     private final int shiftRight = SHIFT_LINK + itemSize;
 
+    /** Number of bytes a child link takes in storage. */
+    public static final int CHILD_LINK_SIZE = PARTITIONLESS_LINK_SIZE_BYTES;
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,181 @@
...
+    static final int MAX_VARLEN_INLINE_SIZE = 64;
+
+    /** Heuristic maximum size of an entry in the {@link BinaryTuple} offset table, in bytes. */
+    static final int MAX_BINARY_TUPLE_OFFSET_TABLE_ENTRY_SIZE = 2;
+
+    /**
+     * Minimum number of items in a {@link BplusInnerIo} so that the search in the B+tree does not lose much in performance due to its rapid
+     * growth.
+     */
+    static final int MIN_INNER_NODE_ITEM_COUNT = 2;
```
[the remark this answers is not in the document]
tkalkirill: Ok let there be a page

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,181 @@
...
+        // Variable length columns.
+
+        switch (spec) {
+            case STRING:
+            case BYTES:
+                return Math.min(MAX_VARLEN_INLINE_SIZE, ((VarlenNativeType) nativeType).length());
+
+            case DECIMAL:
+            case NUMBER:
+                return 4;
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,181 @@
...
+     *
+     * @param indexDescriptor Index descriptor.
+     * @return Inline size in bytes, no more than {@link #MAX_BINARY_TUPLE_INLINE_SIZE}.
+     */
+    static int binaryTupleInlineSize(IndexDescriptor indexDescriptor) {
+        List<? extends ColumnDescriptor> columns = indexDescriptor.columns();
+
+        boolean hasNullColumns = columns.stream().anyMatch(ColumnDescriptor::nullable);
+
+        // First, let's calculate the inline size for all columns.
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,181 @@
...
+     * @return Inline size in bytes, no more than {@link #MAX_BINARY_TUPLE_INLINE_SIZE}.
+     */
+    public static int binaryTupleInlineSize(int pageSize, int itemHeaderSize, IndexDescriptor indexDescriptor) {
+        int maxInnerNodeItemSize = (innerNodePayloadSize(pageSize) / MIN_INNER_NODE_ITEM_COUNT) - BplusInnerIo.CHILD_LINK_SIZE;
+
+        int binaryTupleInlineSize = Math.min(maxInnerNodeItemSize - itemHeaderSize, binaryTupleInlineSize(indexDescriptor));
+
+        if (binaryTupleInlineSize >= MAX_BINARY_TUPLE_INLINE_SIZE) {
+            return binaryTupleInlineSize;
+        }
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,181 @@
...
+        // In this case, the innerNode will fit 7 (with child links) items and 7 items in the leafNode, while 52 bytes will remain free for
+        // the innerNode items, and 100 bytes for the leafNode items, which we could use. For a innerNode, we can add (52 / 7) = 7 bytes
+        // for each item (with link), and for a leafNode, (100 / 7) = 14 bytes for each item, so we can safely use the 7 extra bytes for the
+        // innerNode and leafNode per item.
+
+        if (indexDescriptor.columns().stream().anyMatch(c -> !c.type().spec().fixedLength())) {
+            int itemSize = binaryTupleInlineSize + itemHeaderSize;
+
+            // Let's calculate whether to remain a place in the BplusInnerIo and the BplusLeafIo.
+            int remainingInnerNodePayloadSize = innerNodePayloadSize(pageSize) % (itemSize + BplusInnerIo.CHILD_LINK_SIZE);
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,181 @@
...
+                int additionalInnerNodeItemBytes = remainingInnerNodePayloadSize / innerNodeItemCount;
+                int additionalLeafNodeItemBytes = remainingLeafNodePayloadSize / leafNodeItemCount;
+
+                if (additionalInnerNodeItemBytes > 0 && additionalLeafNodeItemBytes > 0) {
+                    int minAdditionalItemBytes = Math.min(additionalInnerNodeItemBytes, additionalLeafNodeItemBytes);
+
+                    // We need to make sure that by increasing the itemSize we do not reduce the number of items in the BplusLeafIo, this
+                    // may affect its density (holes in one item may appear).
+
+                    if (leafNodePayloadSize(pageSize) / (itemSize + minAdditionalItemBytes) == leafNodeItemCount) {
```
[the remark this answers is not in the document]
tkalkirill: Okay, let there be an assertion

modules/storage-page-memory/src/test/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtilsTest.java
tkalkirill: Added check

modules/storage-page-memory/src/test/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtilsTest.java
tkalkirill: Added check

modules/storage-page-memory/src/test/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtilsTest.java
tkalkirill: it's seconds (long) and nanos(int)

modules/storage-page-memory/src/test/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtilsTest.java
tkalkirill: I don't know yet, I can add an assertion

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,181 @@
...
+                    }
+                }
+            }
+        }
+
+        return Math.min(binaryTupleInlineSize, MAX_BINARY_TUPLE_INLINE_SIZE);
+    }
+
+    /**
+     * Returns number of bytes that can be used to store items and references to child nodes.
```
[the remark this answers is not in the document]
tkalkirill: Fix it

ibessonov (approved):
You have my approval, let's wait for other guys now

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * Helper class for index inlining.
+ *
+ * <p>Index inlining is an optimization that allows index rows (or prefix) to be compared in a {@link BplusTree} without loading them from a
+ * {@link FreeList}.
+ */
+public class InlineUtils {
+    /** Maximum inline size for a {@link BinaryTuple}, in bytes. */
+    public static final int MAX_BINARY_TUPLE_INLINE_SIZE = 2 * KiB;
```
AMashenkov, line 47:
```suggestion
    public static final int MAX_INLINE_SIZE = 2 * KiB;
```

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /** Heuristic maximum inline size of a variable length column in bytes. */
+    static final int MAX_VARLEN_INLINE_SIZE = 64;
```
AMashenkov, line 50:
```suggestion
    static final int DEFAULT_VARLEN_INLINE_SIZE = 64;
```

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,186 @@
...
+        // In this case, the innerNode will fit 7 (with child links) items and 7 items in the leafNode, while 52 bytes will remain free for
+        // the innerNode items, and 100 bytes for the leafNode items, which we could use. For a innerNode, we can add (52 / 7) = 7 bytes
+        // for each item (with link), and for a leafNode, (100 / 7) = 14 bytes for each item, so we can safely use the 7 extra bytes for the
+        // innerNode and leafNode per item.
+
+        if (indexDescriptor.columns().stream().anyMatch(c -> !c.type().spec().fixedLength())) {
+            int itemSize = binaryTupleInlineSize + itemHeaderSize;
+
+            // Let's calculate whether to remain a place in the BplusInnerIo and the BplusLeafIo.
+            int remainingInnerNodePayloadSize = (innerNodePayloadSize(pageSize) - CHILD_LINK_SIZE) % (itemSize + CHILD_LINK_SIZE);
```
AMashenkov, line 145: Let's simplify these calculations.
You already have a desired item size 'itemSize' 
`(itemSize = binaryTupleInlineSize + itemHeaderSize)`.

Then you can get amount of items that fit in the page, like 'innerNodeItemCount' below.
`int innerNodeItemCount = (innerNodePayloadSize(pageSize) - CHILD_LINK_SIZE) / (itemSize + CHILD_LINK_SIZE);`

Then you can just get available space for the every single item.
`int innerNodeItemMaxSize = (innerNodePayloadSize(pageSize) - CHILD_LINK_SIZE) / innerNodeItemCount;`

Values <='innerNodeItemMaxSize' can be safely used instead of 'itemSize' parameter.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,186 @@
...
+     * Calculates inline size for column.
+     *
+     * @param nativeType Column type.
+     * @return Inline size in bytes.
+     */
+    static int inlineSize(NativeType nativeType) {
+        NativeTypeSpec spec = nativeType.spec();
+
+        if (spec.fixedLength()) {
+            return nativeType.sizeInBytes();
```
rpuch, line 74: Isn't it possible to define a fixed-length type with huge length? Currently, we have bitmask, can we specify that it has length of 1000 bytes? Anyway, should we add a check that the size of a fixed-length type is not too big for inlining?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,186 @@
...
+     * Returns number of bytes that can be used to store items and references to child nodes.
+     *
+     * @param pageSize Page size in bytes.
+     */
+    static int innerNodePayloadSize(int pageSize) {
+        return pageSize - BplusInnerIo.HEADER_SIZE;
+    }
+
+    /**
+     * Returns number of bytes that can be used to store items.
```
rpuch, line 179:
```suggestion
     * Returns number of bytes that can be used to store items in a leaf node.
```

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,186 @@
...
+                    binaryTupleInlineSize = itemSize - itemHeaderSize;
+                }
+            }
+        }
+
+        return Math.min(binaryTupleInlineSize, MAX_BINARY_TUPLE_INLINE_SIZE);
+    }
+
+    /**
+     * Returns number of bytes that can be used to store items and references to child nodes.
```
rpuch, line 170:
```suggestion
     * Returns number of bytes that can be used to store items and references to child nodes in an inner node.
```

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
```diff
@@ -0,0 +1,186 @@
...
+        // the inline size for the BinaryTuple. Example: pageSize = 1024, binaryTupleInlineSize = 124, BplusLeafIo.HEADER_SIZE = 56.
+        // In this case, the innerNode will fit 7 (with child links) items and 7 items in the leafNode, while 52 bytes will remain free for
+        // the innerNode items, and 100 bytes for the leafNode items, which we could use. For a innerNode, we can add (52 / 7) = 7 bytes
+        // for each item (with link), and for a leafNode, (100 / 7) = 14 bytes for each item, so we can safely use the 7 extra bytes for the
+        // innerNode and leafNode per item.
+
+        if (indexDescriptor.columns().stream().anyMatch(c -> !c.type().spec().fixedLength())) {
+            int itemSize = binaryTupleInlineSize + itemHeaderSize;
+
+            // Let's calculate whether to remain a place in the BplusInnerIo and the BplusLeafIo.
```
rpuch, line 144:
```suggestion
            // Let's calculate how much space remains in the BplusInnerIo and the BplusLeafIo.
```

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
tkalkirill: I think the proposed name can be a bit confusing.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexDescriptor.java
```diff
@@ -0,0 +1,57 @@
...
+     * Index column descriptor.
+     */
+    interface ColumnDescriptor {
+        /**
+         * Returns the name of an index column.
+         */
+        String name();
+
+        /**
+         * Returns a column descriptor.
```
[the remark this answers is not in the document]
tkalkirill: Fix it.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/InlineUtils.java
tkalkirill: In theory, the user can do this, but we are limited to the `MAX_BINARY_TUPLE_INLINE_SIZE = 2 * KiB` anyway. 
I can add an assertion that there was no overflow after summing the inline for all columns.
Wdyt?

rpuch: No, I think, `Max.min()` is enough

rpuch (approved):

AMashenkov (approved):

PR #1181, opened 2022-10-09
IGNITE-17856

sanpwc:


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1051,4 +1062,42 @@ private CompletableFuture<Void> ensureReplicaIsPrimary(ReplicaRequest request) {
             return CompletableFuture.completedFuture(null);
         }
     }
+
+    /**
+     * Resolves read result to an appropriate binary row. Following rules are used for read result resolution:
```
ibessonov, line 1067: "corresponding" maybe?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1051,4 +1062,42 @@ private CompletableFuture<Void> ensureReplicaIsPrimary(ReplicaRequest request) {
...
+     *
+     * @param readResult Read result to resolve.
+     * @param txId Nullable transaction id, should be provided if resolution is performed within the context of RW transaction.
+     * @return Resolved binary row.
+     */
+    private BinaryRow resolveReadResult(ReadResult readResult, @Nullable UUID txId) {
+        if (readResult == null) {
+            return null;
+        } else {
+            if (txId != null) {
```
ibessonov, line 1084: Would it be better to join these into a single `else if` clause?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1051,4 +1062,42 @@ private CompletableFuture<Void> ensureReplicaIsPrimary(ReplicaRequest request) {
...
         }
     }
+
+    /**
+     * Resolves read result to an appropriate binary row. Following rules are used for read result resolution:
+     * <ol>
+     *     <li>If txId is not null (RW request), assert that retrieved tx id matches proposed one or that retrieved tx id is null
+     *     and return binary row. Currently it's only possible to retrieve write intents if they belong to the same transaction,
+     *     locks prevent reading write intents created by others.</li>
+     *     <li>If txId is not null (RO request), perform write intent resolution if given readResult is a write intent iselft
```
ibessonov, line 1072:
```suggestion
     *     <li>If txId is not null (RO request), perform write intent resolution if given readResult is a write intent itself
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: We will populate 
```
else {
                // RO request.
                // TODO: IGNITE-17637 Implement a commit partition path write intent resolution logic
                return readResult.binaryRow();
            }
```
soon with writeIntent resolution code https://issues.apache.org/jira/browse/IGNITE-17637

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: Fixed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: Agreed, fixed.

PR #1183, opened 2022-10-10
IGNITE-17839

ascherbakoff:


modules/api/src/main/java/org/apache/ignite/tx/IgniteTransactions.java
```diff
@@ -84,6 +84,8 @@ public interface IgniteTransactions {
     /**
      * Executes a closure within a transaction.
      *
+     * <p>Please make sure all asynchronous operations are enlisted into the transaction before returning from the callback.
```
sanpwc, line 87: I'm sure that user won't understand this. I don't even believe that user knows what enlist means.
I'd rather enforce user with javadoc to use only sync operations or anync ones with joins or gets.

modules/api/src/main/java/org/apache/ignite/tx/IgniteTransactions.java
```diff
@@ -130,4 +132,22 @@ default <T> T runInTransaction(Function<Transaction, T> clo) throws TransactionE
...
+     * is completed.
+     *
+     * <p>If the asynchronous chain resulted in no exception, the commitAsync will be automatically called.
+     *
+     * @param clo The closure.
+     * @param <T> Closure result type.
+     * @return The result.
+     */
+    default <T> CompletableFuture<T> runInTransactionAsync(Function<Transaction, CompletableFuture<T>> clo) {
+        // Rollback is expected to be called by the failure handling code
```
sanpwc, line 149: What do you mean?
It seems as inconsistency to me. As far as I understand you expect user to have op.handle(tx.rollback) within closure despite the fact that sync version of runInTransaction will rollback the tx by her own.
```
    default <T> T runInTransaction(Function<Transaction, T> clo) throws TransactionException {
        Transaction tx = begin();

        try {
            T ret = clo.apply(tx);

            System.out.println("!!!: Before commit");
            tx.commit();

            return ret;
        } catch (Throwable t) {
            try {
                tx.rollback(); // Try rolling back on user exception.
            } catch (Exception e) {
                t.addSuppressed(e);
            }

            throw t;
        }
    }
```

sanpwc (changes requested):

modules/api/src/main/java/org/apache/ignite/tx/IgniteTransactions.java
ascherbakoff: **I'm sure that user won't understand this** - why do you think a user is stupid ? This sentence is rather clear. I can rephrase it, but the current approach seems OK. You can provide your own variant to evaluate.

**I'd rather enforce user with javadoc to use only sync operations or anync ones with joins or gets** - this is not that we want. The API allows mixing sync and async ops in a free way. It is the intentional design choice and was promised to a user. Let's not break this statement.

modules/api/src/main/java/org/apache/ignite/tx/IgniteTransactions.java
ascherbakoff: **As far as I understand you expect user to have op.handle(tx.rollback)** - no I don't expect this. 
I expect that a transaction will be already rolled back by a user or automatically by txn engine (in case of error) by the async chain end, so there is no need to call it again.

modules/api/src/main/java/org/apache/ignite/tx/IgniteTransactions.java
sanpwc: > why do you think a user is stupid ? 
I don't think that the user is stupid. However proposed javadoc isn't explicit enough.
- What does enlist means?
- How can I, as a user, be sure that all operations are enlisted?
- How should I, as a user, properly write opAsync.handle clauses? Is it possible?
- etc.
> this is not that we want. The API allows mixing sync and async ops in a free way. It is the intentional design choice and was promised to a user. Let's not break this statement.
Let me rephrase 
I'd rather enforce user with javadoc to use **either** sync operations **or** anync ones with joins or gets **or the combinations of aforementioned** Let's also advice user to use runInTransactionAsync for asyncOps chaining, providing several examples or proper thenCompose, etc and illegal thenApply, etc usecases.
Generally speaking, it's a public non trivial async api, so that, being as explicit with javadoc as possible, deffinitely worth a short.

modules/api/src/main/java/org/apache/ignite/tx/IgniteTransactions.java
sanpwc: It's still possible that an exception will be thrown not from async operations, please consider following scenario.
```
    @Test
    public void testTxClosureKeyValueView2() throws TransactionException {
        var f = igniteTransactions.runInTransactionAsync(tx -> {
            KeyValueView<Tuple, Tuple> view = accounts.keyValueView();

            CompletableFuture<Tuple> read1 = view.getAsync(tx, makeKey(1));

            System.out.println(10 / 0);

            return read1;
        });

        f.join();
    }
```

ascherbakoff: Catch block and test added, javadoc is improved

sanpwc (approved):

PR #1185, opened 2022-10-10
IGNITE-17829 Document new memory storage structure

IgGusev:


docs/_docs/storage/persistent.adoc
```diff
@@ -0,0 +1,90 @@
...
+// limitations under the License.
+
+= Persistent Storage
+
+//:javaFile: {javaCodeDir}/IgnitePersistence.java
+
+== Overview
+
+Ignite Persistence is designed to provide a quick and responsive persistent storage.
+When it is enabled for the link:config/data-region[data region], Ignite stores all the data on disk, and loads as much data as it can into RAM for processing.
```
ibessonov, line 23: >When it is enabled for the link:config/data-region[data region]

Right now this sentence is incorrect, in Ignite 2 we had "persistent" flag in region, now we don't have it. Region is an attribute of the engine, it's persistent if the engine is persistent.

docs/_docs/storage/persistent.adoc
```diff
@@ -0,0 +1,90 @@
...
+* Usage of OS swap
+
+When persistence is enabled, Apache Ignite stores each partition in a separate file on disk. In addition to data partitions, Apache Ignite stores indexes and metadata.
+
+//image::images/persistent_store_structure.png[]
+
+== Enabling Persistent Storage
+
+Native Persistence is configured per link:config/data-region[data region].
+To enable native persistent storage, set the data region to `aipersist`.
```
ibessonov, line 38: This is not true again. Each engine has independent collection of data regions, thus you can't say that we assign an engine to the region, it's the opposite

docs/_docs/storage/rocksdb.adoc
```diff
@@ -0,0 +1,52 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+= RocksDB Storage
+
+RocksDB is a persistent storage engine based on LSM tree. It is best used in environments with a large number of write requests.
+
+When it is enabled for the link:config/data-region[data region], Ignite stores all the data on disk, and loads as much data as it can into RAM for processing.
```
ibessonov, line 20: Same here, but I'd suggest not writing too much about regions in rocksdb, we'll remove them after the beta release

docs/_docs/storage/volatile.adoc
```diff
@@ -0,0 +1,50 @@
...
+When it is enabled for the link:config/data-region[data region], Ignite stores all data in the data region in RAM. Data will be lost on cluster shutdown, so make sure to have a separate data region for persistent storage.
+
+== Configuration Parameters
+
+[cols="1,1,3",opts="header", stripes=none]
+|===
+|name|| The name of the data region.
+|initSize|`256 * 1024 * 1024`| Sets the initial space allocated to the data region.
+|maxSize|`256 * 1024 * 1024`| Sets the maximum space that can be allocated to the data region.
+|evictionMode|`DISABLED`| Sets the eviction algorithm to use.
```
ibessonov, line 32: Should we describe all available alternatives for such properties?

docs/_docs/storage/persistent.adoc
agura: Moreover, a data region is an attribute of some specific storage. Some storage can have data region, another can't. So all properties of persistence are storage specific.

PR #1189, opened 2022-10-11
IGNITE-17874 Added List and ByteBuffer support to message serialization

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-17874

### DIFF
--- a/modules/network-annotation-processor/src/main/java/org/apache/ignite/internal/network/processor/serialization/MessageCollectionItemTypeConverter.java
+++ b/modules/network-annotation-processor/src/main/java/org/apache/ignite/internal/network/processor/serialization/MessageCollectionItemTypeConverter.java
@@ -17,6 +17,7 @@
 
 package org.apache.ignite.internal.network.processor.serialization;
 
+import java.nio.ByteBuffer;
 import java.util.BitSet;
 import java.util.UUID;
 import javax.annotation.processing.ProcessingEnvironment;
@@ -125,6 +126,8 @@ class MessageCollectionItemTypeConverter {
             return MessageCollectionItemType.MSG;
         } else if (typeUtils.isSameType(parameterType, BitSet.class)) {
             return MessageCollectionItemType.BIT_SET;
+        } else if (typeUtils.isSameType(parameterType, ByteBuffer.class)) {
+            return MessageCollectionItemType.BYTE_BUFFER;
         } else {
             throw new ProcessingException("Unsupported MessageCollectionItemType: " + parameterType);
         }
--- a/modules/network/src/main/java/org/apache/ignite/internal/network/direct/DirectMessageReader.java
+++ b/modules/network/src/main/java/org/apache/ignite/internal/network/direct/DirectMessageReader.java
@@ -20,6 +20,7 @@ package org.apache.ignite.internal.network.direct;
 import java.nio.ByteBuffer;
 import java.util.BitSet;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import org.apache.ignite.internal.network.direct.state.DirectMessageState;
@@ -310,6 +311,17 @@ public class DirectMessageReader implements MessageReader {
         return val;
     }
 
+    @Override
+    public ByteBuffer readByteBuffer(String name) {
+        DirectByteBufferStream stream = state.item().stream;
+
+        ByteBuffer val = stream.readByteBuffer();
+
+        lastRead = stream.lastFinished();
+
+        return val;
+    }
+
     /** {@inheritDoc} */
     @Override
     public UUID readUuid(String name) {
@@ -369,6 +381,20 @@ public class DirectMessageReader implements MessageReader {
         lastRead = stream.lastFinished();
 
         return col;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public <C extends List<?>> C readList(String name, MessageCollectionItemType itemType) {
+        DirectByteBufferStream stream = state.item().stream;
+
+        Collection<?> col = stream.readCollection(itemType, this);
+
+        lastRead = stream.lastFinished();
+
+        assert col instanceof List;
+
+        return (C) col;
     }
 
     /** {@inheritDoc} */
--- a/modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
+++ b/modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
@@ -37,6 +37,7 @@ import static org.apache.ignite.internal.util.GridUnsafe.SHORT_ARR_OFF;
 
 import java.lang.reflect.Array;
 import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import java.util.ArrayList;
 import java.util.BitSet;
 import java.util.Collection;
@@ -141,6 +142,14 @@ public class DirectByteBufferStreamImplV1 implements DirectByteBufferStream {
 
     private long uuidLocId;
 
+    private int byteBufferState;
+
+    private byte byteBufferFlag;
+
+    private int byteBufferPosition;
+
+    private int byteBufferLimit;
+
     protected boolean lastFinished;
 
     /** byte-array representation of string. */
@@ -197,20 +206,28 @@ public class DirectByteBufferStreamImplV1 implements DirectByteBufferStream {
     /** {@inheritDoc} */
     @Override
     public void writeShort(short val) {
-        lastFinished = buf.remaining() >= 2;
+        lastFinished = buf.remaining() >= 3;
 
         if (lastFinished) {
-            int pos = buf.position();
-
-            long off = baseOff + pos;
-
-            if (IS_BIG_ENDIAN) {
-                GridUnsafe.putShortLittleEndian(heapArr, off, val);
+            if (val == Short.MAX_VALUE) {
+                val = Short.MIN_VALUE;
             } else {
-                GridUnsafe.putShort(heapArr, off, val);
+                val++;
             }
 
-            buf.position(pos + 2);
+            int pos = buf.position();
+
+            while ((val & 0xFF80) != 0) {
+                byte b = (byte) (val | 0x80);
+
+                GridUnsafe.putByte(heapArr, baseOff + pos++, b);
+
+                val >>>= 7;
+            }
+
+            GridUnsafe.putByte(heapArr, baseOff + pos++, (byte) val);
+
+            buf.position(pos);
         }
     }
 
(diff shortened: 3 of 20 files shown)


modules/network-annotation-processor/src/main/java/org/apache/ignite/internal/network/processor/serialization/MessageCollectionItemTypeConverter.java
```diff
@@ -125,6 +126,8 @@ private MessageCollectionItemType fromDeclaredType(DeclaredType parameterType) {
             return MessageCollectionItemType.MSG;
         } else if (typeUtils.isSameType(parameterType, BitSet.class)) {
             return MessageCollectionItemType.BIT_SET;
+        } else if (typeUtils.isSameType(parameterType, ByteBuffer.class)) {
```
SammyVimes, line 129: Shouldn't it be `isSubType`?

modules/network/src/main/java/org/apache/ignite/internal/network/direct/DirectMarshallingUtils.java
```diff
@@ -31,20 +31,31 @@ public class DirectMarshallingUtils {
      * @return Direct message type.
      */
     public static short getShort(ByteBuffer buffer) {
-        byte b0 = buffer.get();
-        byte b1 = buffer.get();
+        int primShift = 0;
+        int prim = 0;
```
SammyVimes, line 35: What is a prim?

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
```diff
@@ -496,6 +513,85 @@ public void writeBitSet(BitSet val) {
         writeLongArray(val != null ? val.toLongArray() : null);
     }
 
+    /** {@inheritDoc} */
+    @Override
+    public void writeByteBuffer(ByteBuffer val) {
+        switch (byteBufferState) {
+            case 0:
+                byte flag = 0;
```
SammyVimes, line 521: I guess you wanted to use `byteBufferFlag` here? Or I don't get it

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
```diff
@@ -496,6 +513,85 @@ public void writeBitSet(BitSet val) {
...
+
+                if (val != null) {
+                    flag |= 1;
+
+                    if (val.order() == ByteOrder.BIG_ENDIAN) {
+                        flag |= 2;
+                    }
+
+                    if (val.position() == 0) {
+                        flag |= 4;
```
SammyVimes, line 531: I'd personally prefer it inversed. Setting the flag if position is not zero

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
```diff
@@ -496,6 +513,85 @@ public void writeBitSet(BitSet val) {
...
+                if (!lastFinished || val == null) {
+                    return;
+                }
+
+                byteBufferState++;
+
+                //noinspection fallthrough
+            case 2:
+                if ((byteBufferFlag & 8) != 0) {
+                    writeInt(val.limit());
```
SammyVimes, line 562: So we write `limit` if it is equal to the capacity? Why?

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
```diff
@@ -1049,6 +1166,80 @@ public BitSet readBitSet() {
...
+                if (!lastFinished) {
+                    return null;
+                }
+
+                byteBufferState++;
+
+                //noinspection fallthrough
+            case 2:
+                if ((byteBufferFlag & 8) != 0) {
+                    byteBufferLimit = -1;
```
SammyVimes, line 1202: This seems to be correct, but in `write` method we write `limit` in case if flag is set

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
```diff
@@ -496,6 +513,85 @@ public void writeBitSet(BitSet val) {
...
+
+                    if (val.order() == ByteOrder.BIG_ENDIAN) {
+                        flag |= 2;
+                    }
+
+                    if (val.position() == 0) {
+                        flag |= 4;
+                    }
+
+                    if (val.limit() == val.capacity()) {
```
SammyVimes, line 534: I'd prefer != here. So if limit is set, then set the flag

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
```diff
@@ -1049,6 +1166,80 @@ public BitSet readBitSet() {
...
+                }
+
+                byteBufferState = 0;
+                break;
+
+            default:
+                throw new IllegalArgumentException("Unknown byteBufferState: " + uuidState);
+        }
+
+        ByteBuffer val = ByteBuffer.wrap(bytes).position(byteBufferPosition);
```
SammyVimes, line 1228: It's always a heap buffer, seems odd

modules/network/src/main/java/org/apache/ignite/internal/network/direct/DirectMessageReader.java
```diff
@@ -371,6 +383,20 @@ public <C extends Collection<?>> C readCollection(String name, MessageCollection
...
+    /** {@inheritDoc} */
+    @Override
+    public <C extends List<?>> C readList(String name, MessageCollectionItemType itemType) {
+        DirectByteBufferStream stream = state.item().stream;
+
+        Collection<?> col = stream.readCollection(itemType, this);
+
+        lastRead = stream.lastFinished();
+
+        assert col instanceof List;
```
SammyVimes, line 395: Hm, so all our collections are always lists?

SammyVimes (changes requested):
Needs tests, byte buffer writing seems to be broken

modules/network-annotation-processor/src/main/java/org/apache/ignite/internal/network/processor/serialization/MessageCollectionItemTypeConverter.java
ibessonov: Who would use a subtype of byte buffer?

modules/network/src/main/java/org/apache/ignite/internal/network/direct/DirectMarshallingUtils.java
ibessonov: I'm removing this code, too many problems with it, we'll do it later

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
ibessonov: Ok, why not

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
ibessonov: Bug...

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
ibessonov: Sure

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
ibessonov: This is intended, we'll always create heap buffers to avoid problems. Or should we change it? "allocateDirect" should be avoided IMO

modules/network/src/main/java/org/apache/ignite/internal/network/direct/DirectMessageReader.java
ibessonov: You told me that, when we read them, we always create a list

modules/network/src/test/java/org/apache/ignite/internal/network/messages/AllTypesMessage.java
```diff
@@ -78,4 +80,8 @@ public interface AllTypesMessage extends NetworkMessage, Serializable {
     Collection<NetworkMessage> netMsgCollW();
 
     Map<String, NetworkMessage> newMsgMapX();
+
+    List<NetworkMessage> netMsgListY();
```
sashapolo, line 84: Not related to this PR, but theres' a typo above: `newMsgMapX`

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
```diff
@@ -1363,8 +1541,7 @@ public <C extends Collection<?>> C readCollection(MessageCollectionItemType item
      * @return Whether array was fully written.
      */
     boolean writeArray(Object arr, long off, int len, int bytes) {
-        assert arr != null;
-        assert arr.getClass().isArray() && arr.getClass().getComponentType().isPrimitive();
+        assert arr == null || arr.getClass().isArray() && arr.getClass().getComponentType().isPrimitive();
```
sashapolo, line 1544: I don't think this change was necessary

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
```diff
@@ -1049,6 +1153,80 @@ public BitSet readBitSet() {
...
+
+                //noinspection fallthrough
+            case 1:
+                if ((byteBufferFlag & BYTE_BYFFER_HAS_POSITION_FLAG) == 0) {
+                    byteBufferPosition = 0;
+                } else {
+                    byteBufferPosition = readInt();
+                }
+
+                if (!lastFinished) {
```
sashapolo, line 1180: maybe this `if` should be moved to closer to the `readInt` call

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
```diff
@@ -65,6 +66,18 @@ public class DirectByteBufferStreamImplV1 implements DirectByteBufferStream {
     /** Poison object. */
     private static final Object NULL = new Object();
 
+    /** Flag that indicates that byte buffer is not null. */
+    private static final byte BYTE_BYFFER_NOT_NULL_FLAG = 1;
```
sashapolo, line 70: There's a typo in all these constants, its `BYTE_BUFFER`, not `BYTE_BYFFER`

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
```diff
@@ -496,6 +517,89 @@ public void writeBitSet(BitSet val) {
...
+                assert !val.isReadOnly();
+
+                int length = val.capacity();
+
+                if (val.isDirect()) {
+                    long address = GridUnsafe.bufferAddress(val);
+
+                    lastFinished = writeArray(null, address, length, length);
+                } else {
+                    byte[] array = val.array();
```
sashapolo, line 586: hm, didn't we agree that we should transfer only (limit - position) bytes?

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
ibessonov: It was, I pass "null" here for direct buffers

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
```diff
@@ -496,6 +517,89 @@ public void writeBitSet(BitSet val) {
...
+                //noinspection fallthrough
+            case 3:
+                assert !val.isReadOnly();
+
+                int length = val.capacity();
+
+                if (val.isDirect()) {
+                    long address = GridUnsafe.bufferAddress(val);
+
+                    lastFinished = writeArray(null, address, length, length);
```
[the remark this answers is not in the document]
ibessonov: This way I reused code from writeArray, otherwise I'd have to copy it, basically

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
ibessonov: Ok

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
sashapolo: Ok, didn't notice that, thanks for the explanation

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
ibessonov: Would be weird, imo. What if receiver wants to set position to 0? These two numbers are just extra values

SammyVimes (approved):

sashapolo (approved):

PR #1190, opened 2022-10-11
IGNITE-16355 .NET: Support value types in the Table API

ptupitsyn:
* Remove `where T : class` constraint from all APIs.
* Use `Option<T>` as a return type for APIs which can return "no value" results, such as `Get`, `GetAndUpsert`, `GetAndDelete`.
    * It is not possible to handle nullable value and reference types in a generic way, so we have to use a common wrapper. `Option<T>` also clearly indicates which APIs always return something and which don't.
* Fix `ObjectSerializerHandler` to support value types.


isapego (approved):
Looks good to me. Left minor comment.

isapego (approved):

PR #1191, opened 2022-10-12
IGNITE-17859 Update indexes on data modifications

korlov42:


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -814,25 +844,106 @@ private CompletableFuture<RowId> takeLocksForUpsert(ByteBuffer searchKey, UUID i
...
-                    RowId rowId = rowIdByKey(indexId, searchKey);
-
-                    if (rowId == null) {
-                        return lockManager.acquire(txId, new LockKey(indexId, searchKey), LockMode.X) // Index X lock
-                                .thenCompose(exclusiveIdxLock ->
-                                        lockManager.acquire(txId, new LockKey(tableId), LockMode.IX) // IX lock on table
-                                                .thenApply(tblLock -> null));
-                    }
+    private CompletableFuture<?> takeLocksForInsert(BinaryRow tableRow, UUID txId) {
+        return takeLockOnIndexes(tableRow, txId)
```
sanpwc, line 852: Seems that it's mandatory to take locks on keys for all indexes **only** if there is at least one sorted index, because otherwise we'll lock whole table. Will clarify.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -814,25 +844,106 @@ private CompletableFuture<RowId> takeLocksForUpsert(ByteBuffer searchKey, UUID i
...
+        int idx = 0;
+
+        for (IndexStorage index : indexes) {
+            ByteBuffer indexRow = toIndexKey(index, tableRow);
+
+            UUID indexId = indexId(index);
+
+            assert indexRow != null;
+
+            locks[idx++] = lockManager.acquire(txId, new LockKey(indexId, indexRow), LockMode.X);
```
sanpwc, line 873: Not that simple.Different type of indexes will require different type of locks on different operations on different entries

_different operations_
**Insertion** in sorted non-unique index expect following locks to be taken:
- IX_short(nextKey)
- X_commit(currentKey) if nextKey previously locked in S, X or SIX mode
- IX_commit(currentKey) otherwise

**Select** 
S_commit(currentKey)

_different entries_
More precisely select in non-unique is not always about locking current key, but sometimes is about locking +INF
- If the currentKey is bigger than the upper bound, return NOT_FOUND but take a lock anyway.
- If currentKey matches the upper bound no need to take a lock on the next key.
- If the upper bound is null, take a lock on +INF.
- If nothing is found, take a lock on +INF.

_Different type of indexes_
That part is pretty straight forward, hash index on insert requires different set of locks to be taken in comparison to sorted index, same is about unique/non-unique indexes. 


In order to properly hide given difficulties, I'd favor using Alexey's proposal with wrapping IndexStorages with decorators that'll add index specific lock stuff to put, get and remove methods. That's why, I would expect
`Supplier<List<IndexStorage>> activeIndexes` to return list of such decorator instead of IndexStorage's.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -814,25 +844,106 @@ private CompletableFuture<RowId> takeLocksForUpsert(ByteBuffer searchKey, UUID i
...
+        }
+
+        return CompletableFuture.allOf(locks);
+    }
+
+    private UUID indexId(IndexStorage index) {
+        if (index instanceof HashIndexStorage) {
```
sanpwc, line 880: I would expect indexDescriptor to be a part of IndexStorage with some common descriptor part such as id.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -364,15 +295,79 @@ public void onShutdown() {
...
+        BinaryTupleSchema tupleSchema = BinaryTupleSchema.createSchema(descriptor, indexedColumns);
+
+        var converter = new BinaryConverter(descriptor, tupleSchema);
+
+        ByteBuffer buffer = converter.toTuple(tableRow);
+
+        return new IndexRowImpl(new BinaryTuple(tupleSchema, buffer), rowId);
+    }
+
+    private int @Nullable [] resolveIndexColumns(SchemaDescriptor descriptor, IndexStorage indexStorage) {
```
sanpwc, line 332: Duplication of one in PartitionReplicaListener. POC issue?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: Clarified. We need to lock keys on all indexes in any case.

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItInternalTableScanTest.java
```diff
@@ -349,18 +349,18 @@ public void testNullPointerExceptionIsThrownInCaseOfNullSubscription() {
...
+        return new RowAssembler(
+                new SchemaDescriptor(
+                        1,
+                        new Column[]{new Column("key", NativeTypes.stringOf(100), false)},
+                        new Column[]{new Column("val", NativeTypes.stringOf(100), false)}
+                ), 1, 1
+        )
+                .appendString(Objects.requireNonNull(entryKey, "entryKey"))
+                .appendString(Objects.requireNonNull(entryVal, "entryVal"))
+                .build();
```
AMashenkov, line 363: Let's move schema to upper level.
```suggestion
    private static BinaryRow prepareRow(SchemaDescriptor desc, Object... values) {
        assert values.length == desc.length();
        RowAssembler asm = new RowAssembler(desc , 1, 1);
        
        for (int i=0; i<values.length; i++){
                assert values[i] != null || desc.column(i).nullable();
                RowAssembler.writeValue(asm, desc.column(i), values[i]);
       }
       return asm.build();
```

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTablePersistenceTest.java
```diff
@@ -197,7 +197,7 @@ public void afterSnapshot(RaftGroupService service) throws Exception {
     @Override
     public BooleanSupplier snapshotCheckClosure(JraftServerImpl restarted, boolean interactedAfterSnapshot) {
         MvPartitionStorage storage = getListener(restarted, raftGroupId()).getStorage();
-        Map<ByteBuffer, RowId> primaryIndex = getListener(restarted, raftGroupId()).getPk();
+        Map<ByteBuffer, RowId> primaryIndex = null; // getListener(restarted, raftGroupId()).getPk();
```
AMashenkov, line 200: This will cause NPE in line 206 below.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
```diff
@@ -452,6 +452,19 @@ public HashIndexStorage getOrCreateHashIndex(int partitionId, UUID indexId) {
...
+
+        RocksDbMvPartitionStorage partitionStorage = getMvPartition(partitionId);
+
+        if (partitionStorage == null) {
+            throw new StorageException(String.format("Partition ID %d does not exist", partitionId));
+        }
+
+        return storages.getOrCreateStorage(partitionStorage);
+    }
+
```
AMashenkov, line 467:
```suggestion
    @Override
    public HashIndexStorage getOrCreateHashIndex(int partitionId, HashIndexDescriptor descriptor) {
        HashIndex hashIndex = hashIndices.computeIfAbsent(descriptor.id(), id -> new HashIndex(hashIndexCf, descriptor));

        RocksDbMvPartitionStorage partitionStorage = getMvPartition(partitionId);

        if (partitionStorage == null) {
            throw new StorageException(String.format("Partition ID %d does not exist", partitionId));
        }

        return hashIndex.getOrCreateStorage(partitionStorage);
    }

```

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
```diff
@@ -452,6 +452,19 @@ public HashIndexStorage getOrCreateHashIndex(int partitionId, UUID indexId) {
         return storages.getOrCreateStorage(partitionStorage);
     }
 
+    @Override
+    public HashIndexStorage getOrCreateHashIndex(int partitionId, HashIndexDescriptor descriptor) {
+        HashIndex storages = hashIndices.computeIfAbsent(descriptor.id(), id -> new HashIndex(hashIndexCf, descriptor));
```
AMashenkov, line 457:
```suggestion
        HashIndex hashIndex = hashIndices.computeIfAbsent(descriptor.id(), id -> new HashIndex(hashIndexCf, descriptor));
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1910,4 +2005,231 @@ BiFunction<List<Peer>, Long, CompletableFuture<Void>> movePartition(Supplier<Raf
     private <T extends ConfigurationProperty<?>> T directProxy(T property) {
         return getMetadataLocallyOnly ? property : ConfigurationUtil.directProxy(property);
     }
+
+    private Supplier<List<TableIndex>> activeIndexes(UUID tableId, int partId) {
```
AMashenkov, line 2009: Why don't supplier return a map indexUid->TableIndex ?

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTablePersistenceTest.java
korlov42: this test is disabled anyway. I could adjust this code, but I don't know how to debug it since it's broken

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
korlov42: for current task it looks enough to return a list only

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItInternalTableScanTest.java
korlov42: moved to constant

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
korlov42: fixed

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1946,4 +1955,28 @@ BiFunction<List<Peer>, Long, CompletableFuture<Void>> movePartition(Supplier<Raf
...
+            NamedListView<TableIndexView> listView = tableStorage.tablesConfiguration().indexes().value();
+
+            if (listView.size() == 0) {
+                return List.of();
+            }
+
+            List<IndexStorage> indexes = new ArrayList<>(listView.size());
+
+            for (int i = 0; i < listView.size(); i++) {
+                if (!listView.get(i).tableId().equals(tableId)) {
```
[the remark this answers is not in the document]
korlov42: fixed

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
korlov42: me to =(

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
korlov42: yep, it was. Now this is fixed

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -599,42 +557,80 @@ private CompletableFuture processTxCleanupAction(TxCleanupReplicaRequest request
...
+    /**
+     * Finds the row and its identifier by given pk search row.
+     *
+     * @param key A bytes representing a primary key.
+     * @param txId An identifier of the transaction regarding which we need to resolve the given row.
+     * @param action An action to perform on a resolved row.
+     * @param <T> A type of the value returned by action.
+     * @return A future object representing the result of the given action.
+     */
+    private <T> CompletableFuture<T> resolveRowByPk(
```
ibessonov, line 612: This method has the same name and comment as the previous one. But it does some locks magic. Should we rename it and update the documentation?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -380,6 +403,50 @@ private SortedIndexDescriptor convert(SortedIndexView indexView) {
         );
     }
 
+    private CompletableFuture<BinaryTuple> toIndexKey(TableIndexView indexView, BinaryRow tableRow) {
+        return tableManager.tableAsyncInternal(indexView.tableId(), false).thenApply(table -> {
+            SchemaDescriptor descriptor = table.schemaView().schema(tableRow.schemaVersion());
```
sanpwc, line 408: Is it possible that there's no schemes with given version in local registry? What will happen in that case?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -380,6 +403,50 @@ private SortedIndexDescriptor convert(SortedIndexView indexView) {
...
+
+            int[] indexedColumns = resolveIndexColumns(descriptor, indexView);
+
+            if (indexedColumns == null) {
+                return null;
+            }
+
+            BinaryTupleSchema tupleSchema = BinaryTupleSchema.createSchema(descriptor, indexedColumns);
```
sanpwc, line 416: Hmmm, seems that index is actually versioned (within a context of schema versions) entity. Do we really recalculate tupleSchema on every toIndexKey call? I mean that there's a map of version -> descriptor and indexView may me considered as an immutable attribute of an index itself (in-place index update isn't supported, am I right&?). Thus It seems that index can either be pre-popularized with tupleSchemas or evaluate and cache them on demand, one evaluation per each new version in tableRow.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/engine/MvTableStorage.java
```diff
@@ -118,6 +119,17 @@ default IndexStorage getOrCreateIndex(int partitionId, UUID indexId) {
...
+    /**
+     * Returns an already created Hash Index with the given descriptor or creates a new one if it does not exist.
+     *
+     * @param partitionId Partition ID for which this index has been configured.
+     * @param descriptor A description of the index to be created.
+     * @return Hash Index storage.
+     * @throws StorageException If the given partition does not exist, or the given index does not exist or is not configured as a
+     *         hash index.
+     */
+    HashIndexStorage getOrCreateHashIndex(int partitionId, HashIndexDescriptor descriptor);
```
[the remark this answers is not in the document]
AMashenkov: @ibessonov maybe more general method will be better
`public IndexStorage getIndexStorage(int partId, UUID indexId)`
Seems, in some cases we don't know the type of index, e.g. index scan request contains only indexId.

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTablePersistenceTest.java
sanpwc, line 200: Could you please either fix it or create separate ticket if it's not possible because given ticket is disabled.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/IndexLocker.java
```diff
@@ -0,0 +1,48 @@
...
+     * Inserts the given row to the index.
+     *
+     * @param txId An identifier of the transaction in which the row is inserted.
+     * @param tableRow A table row to insert.
+     * @param rowId An identifier of the row in the main storage.
+     */
+    CompletableFuture<?> locksForInsert(UUID txId, BinaryRow tableRow, RowId rowId);
+
+    /**
+     * Inserts the given row to the index.
```
sanpwc, line 41: Typo, removes.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -241,6 +253,9 @@ public class TableManager extends Producer<TableEvent, TableEventParameters> imp
     /** Resolver that resolves a network address to cluster node. */
     private final Function<NetworkAddress, ClusterNode> clusterNodeResolver;
 
+    private final Map<UUID, Map<UUID, IndexStorageAdapterFactory>> indexStorageAdapterFactories = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
sanpwc: Agree

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -241,6 +253,9 @@ public class TableManager extends Producer<TableEvent, TableEventParameters> imp
...
+    private final Map<UUID, Map<UUID, IndexLockerFactory>> indexLockerFactories = new ConcurrentHashMap<>();
```
sanpwc, line 257: Same as above it's either a part of index of table, depending of whether we consider indexes as a lock-aware structures.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1557,6 +1577,103 @@ public CompletableFuture<TableImpl> tableAsyncInternal(UUID id, boolean checkCon
...
     }
 
+    /**
+     * Register the index with given id in a table.
+     *
+     * @param tableId A table id to register index in.
+     * @param indexId An index id os the index to register.
+     * @param searchRowResolver Function which converts given table row to an index key.
+     */
+    public void registerHashIndex(UUID tableId, UUID indexId, Function<BinaryRow, CompletableFuture<BinaryTuple>> searchRowResolver) {
```
sanpwc, line 1587: Is it possible to consider registerHashIndex and similar as a part of tables and not tableManager contract?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1910,4 +2030,132 @@ BiFunction<List<Peer>, Long, CompletableFuture<Void>> movePartition(Supplier<Raf
     private <T extends ConfigurationProperty<?>> T directProxy(T property) {
         return getMetadataLocallyOnly ? property : ConfigurationUtil.directProxy(property);
     }
+
+    private Supplier<List<TableSchemaAwareIndexStorage>> indexStorageAdapters(UUID tableId, int partId) {
```
sanpwc, line 2034: Known stuff, but all that mess with indexes and tables looks crappy((

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -302,30 +256,14 @@ private void handleTxCleanupCommand(TxCleanupCommand cmd, long commandIndex) {
...
-
-            Set<ByteBuffer> insertedKeys = txsInsertedKeys.getOrDefault(txId, Collections.emptySet());
-
             Set<RowId> pendingRowIds = txsPendingRowIds.getOrDefault(txId, Collections.emptySet());
 
             if (cmd.commit()) {
                 pendingRowIds.forEach(rowId -> storage.commitWrite(rowId, cmd.commitTimestamp()));
             } else {
-                pendingRowIds.forEach(rowId -> storage.abortWrite(rowId));
+                pendingRowIds.forEach(storage::abortWrite);
```
sanpwc, line 264: Seems that index cleanup on rollback is missing.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -380,6 +403,50 @@ private SortedIndexDescriptor convert(SortedIndexView indexView) {
         );
     }
 
+    private CompletableFuture<BinaryTuple> toIndexKey(TableIndexView indexView, BinaryRow tableRow) {
+        return tableManager.tableAsyncInternal(indexView.tableId(), false).thenApply(table -> {
```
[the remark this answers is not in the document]
AMashenkov: This looks like a temporary method, which is going to be refactored later.
Agree, seem, SchemaManager#tableSchema resolves the next Alex concern about late schema version registration in local registry.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -856,335 +886,218 @@ private CompletableFuture<Object> processSingleEntryAction(ReadWriteSingleRowRep
...
-                                                    .thenApply(exclusiveRowLock -> rowId);
-                                        }
-
-                                        return CompletableFuture.completedFuture(null);
-                                    });
-                        } else {
-                            rowLockFut = CompletableFuture.completedFuture(null);
-                        }
+    private CompletableFuture<RowId> takeLocksForDeleteExact(BinaryRow expectedRow, RowId rowId, BinaryRow actualRow, UUID txId) {
+        return lockManager.acquire(txId, new LockKey(tableId, expectedRow.keySlice()), LockMode.X)  // Index X lock
```
sanpwc, line 1065: I'd rather expect index locks to be acquired wihtin index wrapper. Same for takeLocksForDelete, etc.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -380,6 +403,50 @@ private SortedIndexDescriptor convert(SortedIndexView indexView) {
         );
     }
 
+    private CompletableFuture<BinaryTuple> toIndexKey(TableIndexView indexView, BinaryRow tableRow) {
```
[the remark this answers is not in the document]
korlov42: Decided to rework this a bit. Now converter should be created once for each schema

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/engine/MvTableStorage.java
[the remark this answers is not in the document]
korlov42: For the current moment I need "get-or-create" semantic, because I failed to find a better place to creates index storages

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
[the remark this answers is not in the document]
korlov42: I don't think so. These collections are updated in `TableManager` and accessed in `PartitionListener` and `PartitionReplicaListener` and can't be reached from `TableImpl`. So, why do you think `TableImpl` is a better place?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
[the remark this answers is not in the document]
korlov42: yep, it is

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
korlov42: Well, I assume (mistakenly) this should be obvious from functions signature: resolve by key and timestamp => RO => just read the resolve the row; resolve by key and txId => RW => acquire all necessary lock before resolving

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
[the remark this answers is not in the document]
korlov42: rework this part a bit to get rid of table n a converters

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
korlov42: Now it will be waiting for the latest schema if the given one is not presented in the schemaRegistry

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
korlov42: The second option is implemented, thanks

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTablePersistenceTest.java
korlov42: I believe this could be fixed within the ticket under which this test is disabled

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbBinaryTupleComparator.java
```diff
@@ -65,6 +65,14 @@ public int compare(ByteBuffer a, ByteBuffer b) {
...
         ByteBuffer secondBinaryTupleBuffer = b.slice().order(ByteOrder.LITTLE_ENDIAN);
 
+        if (!firstBinaryTupleBuffer.hasRemaining()) {
+            return -1;
+        }
+
+        if (!secondBinaryTupleBuffer.hasRemaining()) {
+            return 1;
+        }
+
```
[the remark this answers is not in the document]
korlov42: looks like artefacts after conflict resolution.

fixed

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/IndexLocker.java
korlov42: fixed

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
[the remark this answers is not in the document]
ibessonov: Because indexes belong to a table, generally speaking. Having all these disjointed collections is a mess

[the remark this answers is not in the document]
korlov42: Partition is belonging to a table as well, but we have no access to a partition from table. This is the main reason why I decided to maintain these collection in TableManager.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/engine/MvTableStorage.java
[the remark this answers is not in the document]
korlov42: Reverted this part. Now pk is full-fledged index

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/PkStorage.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.ignite.internal.storage.index.IndexStorage;
+import org.apache.ignite.internal.util.Cursor;
+
+/**
+ * Facade to ease of work with pk constraint storage.
+ *
+ * <p>Encapsulates logic of conversion of ByteBuffer representing the table's key to storage key
+ */
+public class PkStorage {
+    private static final BinaryTupleSchema PK_KEY_SCHEMA = BinaryTupleSchema.create(new Element[]{
```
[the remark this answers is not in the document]
korlov42: Removed this. Now pk is full-fledged index

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
korlov42: fixed

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
korlov42: It has to be a part of a table, but we don't have such a structure right now. Both `TableImpl` and `InternalTableImpl` have no access to the partitions the table consists of.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
korlov42: Yes, this patch is already to big, so I decided to slit clean up part to https://issues.apache.org/jira/browse/IGNITE-17974, if you don't mind

modules/core/src/main/java/org/apache/ignite/internal/util/Lazy.java
```diff
@@ -0,0 +1,58 @@
...
+    public Lazy(Supplier<T> supplier) {
+        this.supplier = supplier;
+    }
+
+    /** Returns the value. */
+    @SuppressWarnings("unchecked")
+    public T get() {
+        if (supplier != EMPTY) {
+            synchronized (this) {
+                if (val == null) {
```
sanpwc, line 49: So val is supposed to be not null? For generic purposes I'd favor the solution with @Nulable val.

modules/core/src/main/java/org/apache/ignite/internal/util/Lazy.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.function.Supplier;
+
+/**
+ * Value which will be initialized at the moment of very first access.
+ *
+ * @param <T> Type of the value.
+ */
+public class Lazy<T> {
+    private static final Supplier<?> EMPTY = () -> {
```
sanpwc, line 28: Why do we need empty supplier?

modules/core/src/main/java/org/apache/ignite/internal/util/Lazy.java
```diff
@@ -0,0 +1,58 @@
...
+        throw new IllegalStateException("Should not be called");
+    };
+
+    private volatile Supplier<T> supplier;
```
sanpwc, line 32: Seems that final should be enough.

modules/core/src/main/java/org/apache/ignite/internal/util/Lazy.java
sanpwc, line 49: So, we do assume that null is an invalid value for val? Why?

modules/core/src/main/java/org/apache/ignite/internal/util/Lazy.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public Lazy(Supplier<T> supplier) {
+        this.supplier = supplier;
+    }
+
+    /** Returns the value. */
+    @SuppressWarnings("unchecked")
+    public T get() {
+        if (supplier != EMPTY) {
+            synchronized (this) {
```
sanpwc, line 48: Why not to check val for reediness outside of synchronized block? meaning that val in that case should be volatile.

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTablePersistenceTest.java
sanpwc: Well, let's try.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sanpwc: Mmm, at first sight it is still possible.
All
```
    private final Map<UUID, Map<UUID, IndexStorageAdapterFactory>> indexStorageAdapterFactories = new ConcurrentHashMap<>();
    private final Map<UUID, Map<UUID, IndexLockerFactory>> indexLockerFactories = new ConcurrentHashMap<>();
```
and 
```
registerHashIndex
unregisterIndex
```
goes inside table

Thus 
`IndexManager#createIndexLocally` will call `tableManager.table(tableId).registerHashIndex` instead of `tableManager.registerHashIndex`

registerHashIndex inside TableManager will look similar to
```
    public void registerHashIndex(UUID indexId, boolean unique, Function<BinaryRow, BinaryTuple> searchRowResolver) {
        indexLockerFactories.computeIfAbsent(
                tableId(),
                key -> new ConcurrentHashMap<>()
        ).put(
                indexId,
                partitionId -> new HashIndexLocker(
                        indexId,
                        unique,
                        lockMgr,
                        searchRowResolver
                )
        );
        indexStorageAdapterFactories.put(
                indexId,
                partitionId -> new TableSchemaAwareIndexStorage(
                        indexId,
                        internalTable().storage().getOrCreateHashIndex(partitionId, indexId),
                        searchRowResolver
                )
        );
    }
```
Besides populating talbe with
```
    private final Map<UUID, IndexLockerFactory> indexLockerFactories = new ConcurrentHashMap<>();

    private final Map<UUID, IndexStorageAdapterFactory> indexStorageAdapterFactories = new ConcurrentHashMap<>();
```
we will also propagate lockManager to a table, not a big deal.

`indexStorageAdapters` also goes to table and in order to populate partion with list of storage adapters we call
` new Lazy<>(() -> internalTbl.indexStorageAdapters(partId).get().get(table.pkId()))` in PartitoinReplicaListener construcor.

I dind't check all that in code, however, seems that it should work.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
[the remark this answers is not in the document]
sanpwc: Still think it's possible, listed the reasons below.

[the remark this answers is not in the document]
ibessonov: Partitions should also be in table, I really hope we'll refactor this behemoth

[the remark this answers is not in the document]
korlov42: fixed

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
korlov42: moved to a table

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
korlov42: done, thanks

modules/core/src/main/java/org/apache/ignite/internal/util/Lazy.java
korlov42: To replace the one provided in construct. The supplier may prevent the object be collected by garbage collector

modules/core/src/main/java/org/apache/ignite/internal/util/Lazy.java
korlov42: no enough, if want to help GC to collect captured objects

modules/core/src/main/java/org/apache/ignite/internal/util/Lazy.java
ibessonov, line 48: Do we synchronize it every time, not only the first time? I don't like it, please implement a proper double-checked locking

korlov42: it was and is the proper double-check locking :)

korlov42: good catch, thanks!

korlov42: done

modules/core/src/main/java/org/apache/ignite/internal/util/Lazy.java
korlov42: rework this to accept nulls

PR #1195, opened 2022-10-12
IGNITE-17816:Sort out and merge Calcite tickets to Ignite 3.0 (step 7)

ygerzhedovich:
IGNITE-16443 Fix nulls processing in search rows - Fixes #9792.
Signed-off-by: Aleksey Plekhanov <plehanov.alex@gmail.com>


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -92,6 +92,11 @@
 public class SqlQueryProcessor implements QueryProcessor {
     private static final IgniteLogger LOG = Loggers.forClass(SqlQueryProcessor.class);
 
+    /**
+     * Default planner timeout, in ms.
+     */
+    private static final long PLANNER_TIMEOUT = 15000L;
```
AMashenkov, line 98:
```suggestion
    /** Default planner timeout, in ms. */
    private static final long PLANNER_TIMEOUT = 15000L;
```
Do we plan to make it configurable?

ygerzhedovich: yes, we have a separate ticket for that purpose

AMashenkov (approved):

PR #1198, opened 2022-10-12
IGNITE-17881 OrderingFuture should notify dependents asynchronously

rpuch:
https://issues.apache.org/jira/browse/IGNITE-17881

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/future/OrderingFuture.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/future/OrderingFuture.java
@@ -176,7 +176,7 @@ public class OrderingFuture<T> {
         }
     }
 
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/internal/future/OrderingFuture.java
```diff
@@ -140,20 +147,56 @@ private void completeInternal(@Nullable T result, @Nullable Throwable ex) {
...
+            waitForCompletionValuesVisibility();
+
             return;
         }
 
-        State<T> prevState;
-        ListNode<T> lastNotifiedNode = null;
+        assert state.phase == Phase.INCOMPLETE;
+
+        switchToNotyfyingStage(result, ex);
```
sashapolo, line 160:
```suggestion
        switchToNotifyingStage(result, ex);
```

sashapolo (approved):

PR #1199, opened 2022-10-12
IGNITE-17486 Fix barriers for Gradle verifications on CI

Pochatkin:
Changes required for CI Gradle verification:
* Fix scripts.
* Update .NET tests to run nodes with Gradle instead of Maven (via PlatformTestNodeRunner class).

https://issues.apache.org/jira/browse/IGNITE-17714
https://issues.apache.org/jira/browse/IGNITE-17486

### DIFF
--- a/modules/platforms/dotnet/Apache.Ignite.Tests/JavaServer.cs
+++ b/modules/platforms/dotnet/Apache.Ignite.Tests/JavaServer.cs
@@ -33,7 +33,9 @@ namespace Apache.Ignite.Tests
     {
         private const int DefaultClientPort = 10942;
 
-        private const int ConnectTimeoutSeconds = 20;
+        private const int ConnectTimeoutSeconds = 120;
+
+        private const string GradleCommandExec = ":ignite-runner:runnerPlatformTest";
 
         /** Maven command to execute the main class. */
         private const string MavenCommandExec = "exec:java@platform-test-node-runner";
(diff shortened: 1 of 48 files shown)


modules/platforms/dotnet/Apache.Ignite.Tests/JavaServer.cs
```diff
@@ -33,17 +33,25 @@ public sealed class JavaServer : IDisposable
     {
         private const int DefaultClientPort = 10942;
 
-        private const int ConnectTimeoutSeconds = 20;
+        private const int ConnectTimeoutSeconds = 120;
```
ptupitsyn, line 36: Is it necessary to increase the timeout so drastically? Does Gradle introduce some kind of a delay?

modules/platforms/dotnet/Apache.Ignite.Tests/JavaServer.cs
```diff
@@ -33,17 +33,25 @@ public sealed class JavaServer : IDisposable
...
+
+        private const string GradleCommandExec = ":ignite-runner:runnerPlatformTest";
```
ptupitsyn, line 38: Please remove unused Maven-related code below.

ptupitsyn (changes requested):

modules/platforms/dotnet/Apache.Ignite.Tests/JavaServer.cs
Pochatkin: Unlit we implement shared build cache for Gradle in CI some steps can produce additional build steps and this can increase start time. So, this is temporary change, until we do some devops magic.

ptupitsyn: Actually, `EnsureBuild` method was added for this reason. Any building, cache warmup, etc should go there. It is not covered by the timeout and can take as long as needed.

However, node startup itself should have a reasonable timeout, because we want to detect issues with cluster connections.

So it would be great to reduce the timeout to the original value and do all the required heavy lifting in `EnsureBuild`.

PR #1202, opened 2022-10-13
IGNITE-17896 Add storage error group.

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-17896


modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -301,4 +301,13 @@ public static class Replicator {
         /** The error happens when the replica is not the current primary replica. */
         public static final int REPLICA_MISS_ERR = REPLICATOR_ERR_GROUP.registerErrorCode(6);
     }
+
+    /** Storage error group. */
+    public static class Storage {
+        /** Replicator error group. */
```
sanpwc, line 307: /** Storage error group. */

sanpwc (approved):

PR #1203, opened 2022-10-13
IGNITE-17849 Remove filter from partition fullscan

rpuch:
https://issues.apache.org/jira/browse/IGNITE-17849

### DIFF
--- a/modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
+++ b/modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
@@ -175,7 +175,7 @@ public interface MvPartitionStorage extends AutoCloseable {
      * @param timestamp Timestamp. Can't be {@code null}.
      * @return Cursor.
      * @throws TxIdMismatchException If there's another pending update associated with different transaction id.
-     * @throws StorageException      If failed to read data from the storage.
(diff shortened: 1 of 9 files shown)


modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
```diff
@@ -173,13 +172,12 @@ public interface MvPartitionStorage extends AutoCloseable {
     /**
      * Scans the partition and returns a cursor of values at the given timestamp.
      *
-     * @param keyFilter Key filter. Binary rows passed to the filter may or may not have a value, filter should only check keys.
      * @param timestamp Timestamp. Can't be {@code null}.
      * @return Cursor.
      * @throws TxIdMismatchException If there's another pending update associated with different transaction id.
-     * @throws StorageException If failed to read data from the storage.
+     * @throws StorageException      If failed to read data from the storage.
```
ibessonov, line 178: As far as I know, we usually don't do these paddings in javadocs.

ibessonov (approved):
Nice!

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
rpuch: Removed the spaces

PR #1205, opened 2022-10-13
IGNITE-17748 Enrich InternalTable.scan API in order to support index scans.

AMashenkov:


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -150,6 +163,11 @@ public class PartitionReplicaListener implements ReplicaListener {
     /** Placement Driver. */
     private final PlacementDriver placementDriver;
 
+    /** Runs async scan tasks for effective tail recursion execution (avoid deep recursive calls). */
+    private final Executor scanRequestExecutor = Executors.newSingleThreadExecutor(
+            new NamedThreadFactory("scan-query-executor-", Loggers.forClass(PartitionReplicaListener.class))
```
korlov42, line 168: we need to use `org.apache.ignite.internal.thread.NamedThreadFactory#threadPrefix` for thread prefix. Also executor have to be stopped at the stop of the table

AMashenkov:
All comments resolved in https://github.com/apache/ignite-3/pull/1314

PR #1212, opened 2022-10-14
IGNITE-16226 .NET: Add KeyValueBinaryView

ptupitsyn:
Added `IKeyValueView<IIgniteTuple, IIgniteTuple> KeyValueBinaryView` to `ITable`.

KV view is just a different representation of the `RecordView`, where every record is split into two parts.
To reuse `RecordView` code, we wrap key and value into `KvPair` at the API entry point, and unwrap back only at the serializer level.

https://issues.apache.org/jira/browse/IGNITE-16226


modules/platforms/dotnet/Apache.Ignite/Table/IKeyValueView.cs
```diff
@@ -0,0 +1,203 @@
...
+    /// <summary>
+    /// Puts multiple key-value pairs.
+    /// </summary>
+    /// <param name="transaction">The transaction or <c>null</c> to auto commit.</param>
+    /// <param name="pairs">Pairs.</param>
+    /// <returns>A <see cref="Task{TResult}"/> representing the asynchronous operation.</returns>
+    Task PutAllAsync(ITransaction? transaction, IEnumerable<KeyValuePair<TK, TV>> pairs);
+
+    /// <summary>
+    /// Gets a value associated with the given key.
```
isapego, line 86: Not just gets I believe

isapego (approved):
Only minor comment regarding documentation

modules/platforms/dotnet/Apache.Ignite/Table/IKeyValueView.cs
ptupitsyn: Good catch, fixed.

PR #1217, opened 2022-10-17
IGNITE-17806 Removed tx state on tx coordinator.

sergeyuttsel:


modules/transactions/src/test/java/org/apache/ignite/internal/tx/storage/state/TxStateStorageAbstractTest.java
```diff
@@ -111,7 +111,7 @@ public void testCas() throws Exception {
 
             UUID txId = UUID.randomUUID();
 
-            TxMeta txMeta0 = new TxMeta(TxState.PENDING, new ArrayList<>(), generateTimestamp(txId));
+            TxMeta txMeta0 = new TxMeta(null, new ArrayList<>(), generateTimestamp(txId));
```
sanpwc, line 114: There's no sense in TxMeta(null). TxMeta either null itself or has not null state internally. So that, such TxMeta(null ... steps should be removed.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -111,9 +109,7 @@ public boolean changeState(UUID txId, TxState before, TxState after) {
     @Override
     public CompletableFuture<Lock> writeLock(UUID lockId, ByteBuffer keyData, UUID txId) {
```
sanpwc, line 110: It's out of the scope of you ticket, however let's just remove writeLock and readLock methods from TxManager. They are deprecated and never used.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxState.java
```diff
@@ -21,7 +21,6 @@
  * Transaction state.
  */
 public enum TxState {
-    PENDING,
```
vldpyatkov, line 24: If we are getting rid of pending state, I think, we should to modify TxManagerImpl#changeState method in order of it doesn't check the state null before CAS. The null is the certain state here, it has to check as others.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlAsynchronousApiTest.java
```diff
@@ -330,7 +329,7 @@ public void checkTransactionsWithDml() throws Exception {
 
         var states = (Map<UUID, TxState>) IgniteTestUtils.getFieldValue(txManagerInternal, TxManagerImpl.class, "states");
 
-        states.forEach((k, v) -> assertNotSame(v, TxState.PENDING));
+        assertEquals(txManagerInternal.finished(), states.size());
```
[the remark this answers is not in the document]
sanpwc: Make sense.

modules/transactions/src/test/java/org/apache/ignite/internal/tx/storage/state/TxStateStorageAbstractTest.java
sergeyuttsel: fixed

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxState.java
sergeyuttsel: fixed

sanpwc (approved):

PR #1224, opened 2022-10-18
IGNITE-17612 Fix TableScanNode data propagation.

zstan:


zstan:
@korlov42  @AMashenkov  plz review ? TC problems are the same as for main branch .NET

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/TableScanNode.java
```diff
@@ -188,7 +188,7 @@ private void push() throws Exception {
             }
         }
```
ygerzhedovich, line 190: why not 
`inBuff.size() < inBufSize` instead of `inBuff.isEmpty()`?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/TableScanNode.java
```diff
@@ -188,8 +188,10 @@ private void push() throws Exception {
...
 
-        if (waiting == 0 || activeSubscription == null) {
-            requestNextBatch();
+        if (requested > 0) {
```
ygerzhedovich, line 191: why you don't merge it with `if` code block below?

AMashenkov (approved):

zstan:
@korlov42 @ygerzhedovich guys, plz check it ? thanks !

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/RootNode.java
```diff
@@ -231,7 +231,7 @@ public void request(int rowsCnt) {
         throw new UnsupportedOperationException();
     }
 
-    private void exchangeBuffers() {
```
ygerzhedovich, line 234: I don't see reason for the change

zstan: of course

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/TableScanNode.java
```diff
@@ -241,9 +241,7 @@ public void onNext(BinaryRow binRow) {
 
             inBuff.add(row);
 
-            if (++received == inBufSize) {
-                received = 0;
-
+            if (inBuff.size() == inBufSize) {
```
ygerzhedovich, line 244:
```suggestion
            if (inBuff.remainingCapacity() == 0) {
```

ygerzhedovich: I would suggest rewrite all such place to use remainingCapacity(), it's more clear.

zstan: In such a case we need to specify queue implementation i don`t think it`s good idea.

ygerzhedovich: it could be just `BlockingQueue` interface

zstan: we don`t use BlockingQueue features here, or i miss smth ? It can be easily changed in future.

zstan: I suppose no need for using BlockingQueue here, it`s by now only for convenience

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItIndexSpoolTest.java
```diff
@@ -57,10 +58,11 @@ protected void cleanUp() {
     /**
      * Test.
      */
+    @Disabled("https://issues.apache.org/jira/browse/IGNITE-17959")
     @ParameterizedTest(name = "tableSize=" + ARGUMENTS_PLACEHOLDER)
-    @ValueSource(ints = {1, 10, IN_BUFFER_SIZE, IN_BUFFER_SIZE + 1, 2000})
+    @ValueSource(ints = {1, 10, IN_BUFFER_SIZE * 1/**partitions count*/, IN_BUFFER_SIZE * 1 + 1, 2000})
```
ygerzhedovich, line 63: maybe introduce new constant in the class for the purpose?

zstan: Thanks for pointing, i change it, now there are 2 arguments passed int the test, rowCount and partitions

PR #1225, opened 2022-10-18
ignite-17759

vldpyatkov:


modules/replicator/src/main/java/org/apache/ignite/internal/replicator/message/ReplicaRequest.java
```diff
@@ -28,5 +29,6 @@ public interface ReplicaRequest extends NetworkMessage {
      *
      * @return Replication group id.
      */
-    String groupId();
+    @Marshallable
+    TablePartitionId groupId();
```
sanpwc, line 33: Earlier we've discussed that ReplicaRequest is a general entity that also will be used within meta storage etc. Thus it's not valid to use stuff like Table in int.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/message/TablePartitionId.java
```diff
@@ -0,0 +1,99 @@
...
+
+import java.io.Serializable;
+import java.util.UUID;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * The class stores a table id together with a partition id.
+ * It is used as a replication group identifier.
+ */
+public class TablePartitionId implements Comparable<TablePartitionId>, Serializable {
```
sanpwc, line 28: Why it's Serializable?

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestSingleNode.java
```diff
@@ -398,14 +399,13 @@ protected Int2ObjectOpenHashMap<RaftGroupService> startTable(String name, UUID t
                         raftSvc -> {
                             try {
                                 replicaManagers.get(node).startReplica(
-                                        grpId,
+                                        new TablePartitionId(tblId, partId),
```
sanpwc, line 402: I don't like an idea that in some places we use ReplicationGroupId and in some old fashioned String grpId, e.g. grpId on line 387
```
                CompletableFuture<Void> partitionReadyFuture = raftServers.get(node).prepareRaftGroup(
                        grpId,
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -681,7 +682,9 @@ private void updateAssignmentInternal(ConfigurationNotificationEvent<byte[]> ass
 
                 String grpId = partitionRaftGroupName(tblId, partId);
 
-                placementDriver.updateAssignment(grpId, nodes);
+                TablePartitionId replicaGrpId = new TablePartitionId(tblId, partId);
```
sanpwc, line 685: `String grpId = partitionRaftGroupName(tblId, partId);` should be removed.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/message/TablePartitionId.java
vldpyatkov: Because it is used in raft command (look UpdateCommand).

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TransactionImpl.java
```diff
@@ -55,11 +57,14 @@ public class TransactionImpl implements InternalTransaction {
     private final TxManager txManager;
 
     /** Enlisted replication groups: replication group id -> (primary replica node, raft term). */
-    private final Map<String, IgniteBiTuple<ClusterNode, Long>> enlisted = new ConcurrentSkipListMap<>();
+    private final Map<TablePartitionId, IgniteBiTuple<ClusterNode, Long>> enlisted = new ConcurrentSkipListMap<>();
```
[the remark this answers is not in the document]
vldpyatkov: Done.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/command/UpdateCommand.java
```diff
@@ -40,14 +44,21 @@ public class UpdateCommand extends PartitionCommand {
     /**
      * Creates a new instance of UpdateCommand with the given row to be updated. The {@code rowId} should not be {@code null}.
      *
+     * @param commitTablePartitionId Committed table partition id.
```
sanpwc, line 47: Here and everywhere let's be consistent in naming commit/committed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -160,9 +160,11 @@ private void handleUpdateCommand(UpdateCommand cmd, long commandIndex) {
             BinaryRow row = cmd.getRow();
             RowId rowId = cmd.getRowId();
             UUID txId = cmd.txId();
+            UUID commitTblId = cmd.getCommitTablePartitionId().getTableId();
+            int commitPartId = cmd.getCommitTablePartitionId().getPartId();
 
             // TODO: IGNITE-17759 Need pass appropriate commitTableId and commitPartitionId.
```
sanpwc, line 166: Please remove todos IGNITE-17759

modules/transactions/src/main/java/org/apache/ignite/internal/tx/InternalTransaction.java
```diff
@@ -52,14 +53,29 @@ public interface InternalTransaction extends Transaction {
      */
     TxState state();
 
+    /**
+     * Assign a replication group that will store the transaction state.
```
sanpwc, line 57: Enlists (Assigns _up to you_) an id of a replication group that will store the transaction state.

sanpwc (changes requested):
Please check my comments.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
vldpyatkov: I removed this method, but I think, it isn't exactly what you expected, because now I am using ReplicationGroupId#toString() instead of them.

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestSingleNode.java
vldpyatkov: It rather connected with the replication layer is dedicated from RAFT protocol implementation. They both just cannot have the same class for identification replica.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/command/UpdateCommand.java
vldpyatkov: I always use commitReplicationGroupId as a name of variable.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
vldpyatkov: Done.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -890,8 +897,10 @@ private CompletableFuture<Object> processSingleEntryAction(ReadWriteSingleRowRep
                     BinaryRow lockedRow = lockedRowId != null
                             ? resolveReadResult(mvDataStorage.read(lockedRowId, HybridTimestamp.MAX_VALUE), txId) : null;
 
-                    CompletableFuture raftFut = lockedRowId != null ? applyCmdWithExceptionHandling(new UpdateCommand(lockedRowId, txId)) :
-                            CompletableFuture.completedFuture(null);
+                    CompletableFuture raftFut =
+                            lockedRowId != null ? applyCmdWithExceptionHandling(new UpdateCommand(committedPartitionId, lockedRowId, txId))
+                                    :
```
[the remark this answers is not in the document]
vldpyatkov: Done.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/InternalTransaction.java
vldpyatkov: Rewrite the comment as: Assigns a replication group id to store the transaction state.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/message/ReplicaRequest.java
sanpwc: ReplicationGroupId that is replicator module should no nothing about the table. What's the problem in implementing proposed solution with an interface and TablePartitionId implementation.

>I think, it is enough this time, we will have added something else when will use replication for metastorage.
Why would we implement incorrect solution, within the context of plans of moving meta storage towards replicas.

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestSingleNode.java
sanpwc: Why? What if you'll move move Replication to ignite-core? Semantically we use exactly the same entities to identify groups both in replicator and raft. In case of tables it's tableId + partitionIndex. However, because of String nature of in-raft ids it's required to have conversion rules, potentially both sides and explicit grpId params here and there which looks not that good.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sanpwc: Yep, answered above.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/command/UpdateCommand.java
sanpwc: Nope, sometimes it's `TablePartitionId committedPartition` and besides of that there are

- javadocs, e.g.`commitTablePartitionId Committed` 
- method names  `TablePartitionId committedPartition();`

modules/transactions/src/main/java/org/apache/ignite/internal/tx/InternalTransaction.java
sanpwc: Well, it sound like, that is replication group **id** that stores transaction state.
I mean that you assigns replication group id, that identifies a replication group that stores a transaction state.

PR #1232, opened 2022-10-20
IGNITE-17330 Support read only transactions from sql side

zstan:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -620,4 +629,10 @@ public CompletableFuture<Boolean> notify(@NotNull IndexEventParameters parameter
                     .thenApply(v -> false);
         }
     }
+
+    /** Returns {@code true} if this is data modification operation. */
+    private static boolean dataModificationOp(SqlNode sqlNode) {
+        return sqlNode instanceof SqlInsert || sqlNode instanceof SqlMerge || sqlNode instanceof SqlDelete
```
korlov42, line 635: probably it would be better to replace this with `SqlKind.DML.contains(sqlNode.getKind())`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/FragmentDescription.java
```diff
@@ -34,6 +35,8 @@ public class FragmentDescription implements Serializable {
 
     private Long2ObjectMap<List<String>> remoteSources;
 
+    private HybridTimestamp txTs;
```
korlov42, line 38: Fragment description is definitely not the place to add transaction to

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/BaseQueryContext.java
```diff
@@ -159,7 +158,8 @@ public List<MetadataHandler<?>> handlers(Class<? extends MetadataHandler<?>> hnd
 
     private final Object[] parameters;
 
-    private final InternalTransaction tx;
+    /** Transaction representation: it can be implementation or clock representation for RO transactions. */
+    private final Object tx;
```
korlov42, line 162: I don't think It's good idea to keep different objects in a single field

korlov42 (changes requested):

zstan:
@korlov42 thanks for review, all done

zstan:
@ygerzhedovich can we merge it ? wdyt?

ygerzhedovich:
> @ygerzhedovich can we merge it ? wdyt?

@zstan , why we should merge it right now if we have a dependency on https://issues.apache.org/jira/browse/IGNITE-17260 and must use new interfaces?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -387,17 +389,20 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
...
                                             .defaultSchema(schema)
-                                            .traitDefs(Commons.LOCAL_TRAITS_SET)
+                                            .traitDefs(rwTx ? Commons.LOCAL_TRAITS_SET : Commons.DISTRIBUTED_TRAITS_SET)
                                             .build()
                             )
                             .logger(LOG)
                             .cancel(queryCancel)
                             .parameters(params)
                             .transaction(outerTx)
+                            .transactionTime(rwTx ? null : new HybridClock().now())
```
ygerzhedovich, line 405: in case of outer TX we can have different time for few sequented run queries with the same outer TX

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionContext.java
```diff
@@ -348,6 +354,11 @@ public InternalTransaction transaction() {
         return tx;
     }
 
+    /** Transaction for current context. */
```
ygerzhedovich, line 357: I suggest mention that is related to RO transaction only

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -387,17 +389,21 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
...
-                                            .traitDefs(Commons.LOCAL_TRAITS_SET)
+                                            .traitDefs(rwTx ? Commons.LOCAL_TRAITS_SET : Commons.DISTRIBUTED_TRAITS_SET)
                                             .build()
                             )
                             .logger(LOG)
                             .cancel(queryCancel)
                             .parameters(params)
                             .transaction(outerTx)
+                            // use new implementation after ignite-17260
+                            .transactionTime(/*outerTx != null ? outerTx.timeStamp() : */new HybridClock().now())
```
ygerzhedovich, line 406: you shouldn't take timestamps for internal RW transaction

zstan:
@ygerzhedovich main branch rebased and ready for review

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -387,17 +389,20 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
...
                                             .defaultSchema(schema)
-                                            .traitDefs(Commons.LOCAL_TRAITS_SET)
+                                            .traitDefs(rwTx ? Commons.LOCAL_TRAITS_SET : Commons.DISTRIBUTED_TRAITS_SET)
                                             .build()
                             )
                             .logger(LOG)
                             .cancel(queryCancel)
                             .parameters(params)
                             .transaction(outerTx)
+                            .transactionTime(!rwTx ? (outerTx != null ? outerTx.readTimestamp() : new HybridClock().now()) : null)
```
ygerzhedovich, line 405: negative `rwTx` a little bit complicated to read, let's use direct rwTx

ygerzhedovich, line 405: Will it be works in case outerTx exists and it's RW transaction?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -419,7 +425,7 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
                                 return new AsyncSqlCursorImpl<>(
                                         SqlQueryType.mapPlanTypeToSqlType(plan.type()),
                                         plan.metadata(),
-                                        implicitTx,
+                                        rwTx ? implicitTx : null,
```
ygerzhedovich, line 428: why you changed it? `implicitTx` will be `null` for RO transaction.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -468,7 +474,19 @@ private List<CompletableFuture<AsyncSqlCursor<List<Object>>>> query0(
 
         CompletableFuture<Void> start = new CompletableFuture<>();
 
+        boolean needStartTx = false;
+
+        for (SqlNode sqlNode : nodes) {
+           if (SqlKind.DML.contains(sqlNode.getKind()) || SqlKind.QUERY.contains(sqlNode.getKind())) {
```
ygerzhedovich, line 480: why you need transaction for SELECT statement?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionContext.java
```diff
@@ -221,8 +223,8 @@ public String originatingNodeId() {
     /**
      * Get local node ID.
```
ygerzhedovich, line 224: javadoc

zstan: what problem with javadoc here ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
zstan: Honestly, I don`t know how select will be processed without tx.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
zstan: If this is rw tx we will set null here, is it a question ?

ygerzhedovich: I mean the following case:
1) we have an outer RW transaction
2) user do SELECT, so rwTx = false
3) you will set transaction time as not null

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
ygerzhedovich: ok, which type of query we could have here, outside of DML and QUERY?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionContext.java
ygerzhedovich: it was node id, but now it's no so.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -387,17 +389,24 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
                     return nodes.get(0);
                 })
                 .thenCompose(sqlNode -> {
+                    final boolean rwTx = dataModificationOp(sqlNode);
+                    final HybridTimestamp txTime = outerTx != null && !outerTx.isReadOnly() ? null :
```
ygerzhedovich, line 393:
```suggestion
final HybridTimestamp txTime = outerTx != null ? outerTx.readTimestamp() : rwTx ? null : new HybridClock().now();```

Comment (ygerzhedovich) [reply modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java:393]: and use the txTime as is

File: modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java [hunk 12]
```diff
@@ -387,17 +389,24 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
...
-                                            .traitDefs(Commons.LOCAL_TRAITS_SET)
+                                            .traitDefs(rwTx || (outerTx != null && !outerTx.isReadOnly()) ? Commons.LOCAL_TRAITS_SET :
+                                                    Commons.DISTRIBUTED_TRAITS_SET)
                                             .build()
                             )
                             .logger(LOG)
                             .cancel(queryCancel)
                             .parameters(params)
                             .transaction(outerTx)
+                            .transactionTime(rwTx || (outerTx != null && !outerTx.isReadOnly()) ? null :
```
ygerzhedovich, line 408:
```suggestion
                            .transactionTime(txTime)
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
zstan: i want to filter DDL here

zstan:
@ygerzhedovich all done visa is green

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/message/QueryStartRequest.java
```diff
@@ -47,4 +49,10 @@ public interface QueryStartRequest extends ExecutionContextAwareMessage {
      */
     @Marshallable
     Object[] parameters();
+
+    /**
+     * Read only transaction time or null if this is read write transaction.
+     */
+    @Marshallable
+    @Nullable HybridTimestamp txTime();
```
korlov42, line 57: Does it make sense to introduce some sort of transferrable context?

PR #1233, opened 2022-10-20
IGNITE-17894 Implement RAFT snapshot streaming receiver

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-17894


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -861,9 +876,10 @@ private RaftGroupOptions groupOptionsForPartition(
                 raftMgr.topologyService(),
                 //TODO IGNITE-17302 Use miniumum from mv storage and tx state storage.
                 new OutgoingSnapshotsManager(raftMgr.messagingService()),
-                partitionStorage::persistedIndex,
+                new MvPartitionAccess(tableStorage, partId),
```
ibessonov, line 879: Please coordinate with @rpuch on these changes to reduce the severity of merge conflicts

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/MvPartitionAccess.java
```diff
@@ -0,0 +1,123 @@
...
+
+        mvPartition.runConsistently(() -> {
+            mvPartition.lastAppliedIndex(lastAppliedIndex);
+
+            return null;
+        });
+    }
+
+    @Override
+    public void writeVersionChain(
```
ibessonov, line 87: I don't think that we should put these methods into this class. It's enough to have storage getters here and actual logic will then be implemented somewhere in copier, for example

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/MvPartitionAccess.java
```diff
@@ -0,0 +1,123 @@
...
+            List<HybridTimestamp> timestamps,
+            @Nullable UUID txId,
+            @Nullable UUID commitTableId,
+            int commitPartitionId
+    ) throws StorageException {
+        MvPartitionStorage mvPartition = tableStorage.getMvPartition(partId);
+
+        assert mvPartition != null;
+
+        mvPartition.runConsistently(() -> {
```
ibessonov, line 99: Why have you decided to have a single closure with loop inside and not the other way around?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -17,94 +17,128 @@
...
+    private final PartitionSnapshotStorage partitionSnapshotStorage;
 
-    /** Snapshot URI. */
     private final SnapshotUri snapshotUri;
 
-    /** Rebalance thread-pool, used to write data into a storage. */
-    //TODO https://issues.apache.org/jira/browse/IGNITE-17262
-    // Use external pool.
-    private final ExecutorService threadPool = Executors.newSingleThreadExecutor();
+    private final IgniteSpinBusyLock cancelBusyLock = new IgniteSpinBusyLock();
```
ibessonov, line 65: Can it be named `busyLock` like everywhere else?
We may have a separate explicit cancellation flag

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -17,94 +17,128 @@
...
+                }
+
+                cause = snapshotCopierException.getCause();
+            } else {
+                if (!isOk()) {
+                    setError(RaftError.UNKNOWN, "Unknown error on completion the copier");
+                }
+            }
+
+            throw new IllegalStateException(cause);
```
ibessonov, line 122: We should stop using `IllegalStateException`. What does it mean?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -17,94 +17,128 @@
...
-                    1000L
-            );
+        future.cancel(true);
+    }
 
-            metaRequestFuture.whenComplete((networkMessage, throwable) -> {
-                SnapshotMetaResponse metaResponse = (SnapshotMetaResponse) networkMessage;
+    @Override
+    public void close() throws IOException {
+        cancel();
```
ibessonov, line 135: Shouldn't this be called when copier already canceled or joined? Very suspicious

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -113,8 +147,174 @@ public SnapshotReader getReader() {
         return new IncomingSnapshotReader(snapshotMeta);
     }
 
-    @Override
-    public void close() throws IOException {
-        threadPool.shutdownNow();
+    private void start0() {
```
ibessonov, line 150: Javadocs would be nice even on some private methods

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -113,8 +147,174 @@ public SnapshotReader getReader() {
...
+        if (!cancelBusyLock.enterBusy()) {
+            return;
+        }
+
+        try {
+            partitionSnapshotStorage.partition().reCreatePartition()
+                    .whenComplete((unused, throwable) -> {
```
ibessonov, line 157: Why do you use `sync` version of the method? What thread will execute this code?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -113,8 +147,174 @@ public SnapshotReader getReader() {
...
+        if (!cancelBusyLock.enterBusy()) {
+            return;
+        }
+
+        try {
+            partitionSnapshotStorage.outgoingSnapshotsManager().messagingService().invoke(
+                    snapshotSenderNode,
+                    MSG_FACTORY.snapshotMetaRequest().id(snapshotUri.snapshotId).build(),
+                    NETWORK_TIMEOUT
+            ).whenComplete((networkMessage, throwable) -> {
```
ibessonov, line 194: you can call it `response`, would be easier to understand

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -113,8 +147,174 @@ public SnapshotReader getReader() {
...
+        if (!cancelBusyLock.enterBusy()) {
+            return;
+        }
+
+        try {
+            SnapshotMeta snapshotMeta0 = snapshotMeta;
+
+            assert snapshotMeta0 != null;
+
+            partitionSnapshotStorage.partition().lastAppliedIndex(snapshotMeta0.lastIncludedIndex());
```
ibessonov, line 252: I don't think I see a part when you set applied index to -1. Why did you decided to ignore it?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -113,8 +147,174 @@ public SnapshotReader getReader() {
...
+                        entry.txId(),
+                        entry.commitTableId(),
+                        entry.commitPartitionId()
+                );
+            } catch (StorageException e) {
+                future.completeExceptionally(copierException(e, RaftError.EIO, "Error writing version chain"));
+            } catch (Throwable t) {
+                future.completeExceptionally(t);
+            } finally {
+                cancelBusyLock.leaveBusy();
```
ibessonov, line 284: Such common parts could be refactored, I think.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -113,8 +147,174 @@ public SnapshotReader getReader() {
...
+
+    /**
+     * Exception along with its status for snapshot copier.
+     *
+     * <p>The class is needed because the copier status is not thread-safe.
+     *
+     * @see #isOk()
+     * @see #setError(RaftError, String, Object...)
+     */
+    static class SnapshotCopierException extends RuntimeException {
```
ibessonov, line 297: This is weird. Not even a StorageException or anything. We should start thinking in terms of error groups and such

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopierTest.java
```diff
@@ -0,0 +1,322 @@
...
+
+    private static final RaftMessagesFactory RAFT_MSG_FACTORY = new RaftMessagesFactory();
+
+    private final ExecutorService executorService = Executors.newSingleThreadExecutor();
+
+    @InjectConfiguration
+    private TablesConfiguration tablesConfig;
+
+    @InjectConfiguration
+    private TableConfiguration tableConfig;
```
ibessonov, line 112: Can this be a part of previous field?

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopierTest.java
```diff
@@ -0,0 +1,322 @@
...
+
+            partitionStorage.lastAppliedIndex(rowIds.size());
+
+            return null;
+        });
+
+        return rowIds;
+    }
+
+    private static List<ResponseEntry> createResponseEntries(MvPartitionStorage partitionStorage, List<RowId> rowIds) {
```
ibessonov, line 249: Again, having some comments here and there would be nice

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
SammyVimes: I think this is ok, it's kind of like `assert false`. This should never happen if there is no bug, right?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -404,6 +408,15 @@ public TableManager(
                 TimeUnit.MILLISECONDS,
                 new LinkedBlockingQueue<>(),
                 NamedThreadFactory.create(nodeName, "tableManager-io", LOG));
+
+        incomingSnapshotsExecutor = new ThreadPoolExecutor(
```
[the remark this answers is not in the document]
tkalkirill: I looked at the methods in **Executors**, it seems to me that they do not quite suit me, I changed the pool configuration a little.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: Discussed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/MvPartitionAccess.java
tkalkirill: Tried to fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/MvPartitionAccess.java
tkalkirill: Tried to fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
tkalkirill: Tried to fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
tkalkirill: Tried to fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
tkalkirill: Tried to fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
tkalkirill: Tried to fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
tkalkirill: Tried to fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
tkalkirill: Tried to fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
tkalkirill: Tried to fix it.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopierTest.java
tkalkirill: Tried to fix it.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopierTest.java
tkalkirill: Tried to fix it.

SammyVimes (approved):
LGTM!

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
```diff
@@ -297,19 +292,35 @@ public void close() throws Exception {
...
-    /** {@inheritDoc} */
     @Override
     public long lastAppliedIndex() {
         return lastAppliedIndex;
     }
 
-    /** {@inheritDoc} */
+    @Override
+    public void lastAppliedIndex(long lastAppliedIndex) {
+        try (WriteBatch writeBatch = new WriteBatch()) {
```
ibessonov, line 307: I don't really get why you need to have a batch here when you only write a single key-value pair

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -17,94 +17,156 @@
...
     }
 
     @Override
-    public void cancel() {
-        //TODO https://issues.apache.org/jira/browse/IGNITE-17262
-        // Implement.
+    public void start() {
+        Executor executor = partitionSnapshotStorage.getIncomingSnapshotsExecutor();
+
+        LOG.info("Copier is started for the partition: " + partId());
```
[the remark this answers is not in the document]
tkalkirill: Corrected the format, do we need `if (infoEnabled)` if there is a check inside?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -17,94 +17,156 @@
...
-            metaRequestFuture.whenComplete((networkMessage, throwable) -> {
-                SnapshotMetaResponse metaResponse = (SnapshotMetaResponse) networkMessage;
-
-                snapshotMeta = metaResponse.meta();
-            });
-        });
+    public void cancel() {
+        canceled = true;
+
+        LOG.info("Copier is cancelled");
```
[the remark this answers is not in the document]
tkalkirill: Modified a little, I don’t know what to write there yet.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -17,94 +17,156 @@
...
+
+        CompletableFuture<?> fut = future;
+
+        if (fut != null) {
+            fut.cancel(true);
+        }
+    }
+
+    @Override
+    public void close() throws IOException {
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -113,8 +175,234 @@ public SnapshotReader getReader() {
...
+                        return completedFuture(null);
+                    }
+
+                    mvPartitionStorage.runConsistently(() -> {
+                        mvPartitionStorage.lastAppliedIndex(FULL_RABALANCING_STARTED);
+
+                        return null;
+                    });
+
+                    LOG.info("Copier prepared multi-versioned storage for the partition: " + partId());
```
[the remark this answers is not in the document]
tkalkirill: Tried to fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -113,8 +175,234 @@ public SnapshotReader getReader() {
...
+            MvPartitionStorage mvPartition = partitionSnapshotStorage.partition().mvPartitionStorage();
+
+            for (ResponseEntry entry : snapshotMvDataResponse.rows()) {
+                if (canceled) {
+                    return completedFuture(null);
+                }
+
+                // Let's write all versions for the row ID.
+                mvPartition.runConsistently(() -> {
+                    RowId rowId = new RowId(partId(), entry.rowId().getMostSignificantBits(), entry.rowId().getLeastSignificantBits());
```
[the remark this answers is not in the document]
tkalkirill: Tried to fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -113,8 +175,234 @@ public SnapshotReader getReader() {
...
+
+                    for (int i = 0; i < entry.rowVersions().size(); i++) {
+                        HybridTimestamp timestamp = i + 1 <= entry.timestamps().size() ? entry.timestamps().get(i) : null;
```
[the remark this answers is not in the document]
tkalkirill: Thanks

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
tkalkirill: Tried to fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -113,8 +175,234 @@ public SnapshotReader getReader() {
...
+        }
+
+        return partitionSnapshotStorage.outgoingSnapshotsManager().messagingService().invoke(
+                snapshotSender,
+                MSG_FACTORY.snapshotMetaRequest().id(snapshotUri.snapshotId).build(),
+                NETWORK_TIMEOUT
+        ).thenAccept(response -> {
+            snapshotMeta = ((SnapshotMetaResponse) response).meta();
+
+            LOG.info("Copier has loaded the snapshot meta for the partition: " + partId());
```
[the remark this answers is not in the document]
tkalkirill: Tried to fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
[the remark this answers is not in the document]
ibessonov: I vividly remember this topic being discussed a long time ago. The point is that the call without check creates a vararg array argument. It doesn't affect anything in most cases, but sometimes it may become a problem. I didn't invent this, this is how we used to write logs in Ignite 2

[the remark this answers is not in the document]
tkalkirill: I agree with you, but we don't have rules for 3.0 yet, maybe we should discuss it and add it to the code style rules.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -17,94 +17,156 @@
...
+
+                        if (!isOk()) {
+                            setError(RaftError.UNKNOWN, "Sender node was not found or it is offline");
+                        }
+
+                        return completedFuture(null);
+                    }
+
+                    return loadSnapshotMeta(snapshotSender)
+                            .thenCompose(unused1 -> loadSnapshotMvData(snapshotSender, executor))
```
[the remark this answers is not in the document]
tkalkirill: But this would be a useless argument, why do this? Maybe I didn't understand you.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccess.java
```diff
@@ -17,43 +17,44 @@
...
+     * Destroys and recreates the multi-versioned partition storage.
      *
-     * @param lowerBound Lower bound.
-     * @throws StorageException If failed to read data from the storage.
+     * @param executor Executor.
+     * @throws StorageException If an error has occurred during the partition destruction.
      */
-    @Nullable
-    RowId closestRowId(RowId lowerBound);
+    CompletableFuture<MvPartitionStorage> reCreateMvPartitionStorage(Executor executor) throws StorageException;
```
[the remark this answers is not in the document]
tkalkirill: Are you suggesting calling **TxStateStorage#destroy** instead of **TxStateTableStorage#destroyTxStateStorage**?
I think it's better to call **TxStateTableStorage#destroyTxStateStorage**. And don't show **TxStateTableStorage** to the outside.

Wdyt?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -17,94 +17,156 @@
...
+            fut.cancel(true);
+        }
+    }
+
+    @Override
+    public void close() throws IOException {
+        cancel();
+
+        try {
+            // Because after closing nobody calls join().
```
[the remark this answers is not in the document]
tkalkirill: Tried to fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccess.java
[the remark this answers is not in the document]
ibessonov: I forgot that destroy is called from the table and not from the partition, so we can leave it, ignore my comment then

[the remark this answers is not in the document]
tkalkirill: Glory to programming!

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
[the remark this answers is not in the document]
ibessonov: Both options looks weird, so it's up to you to decide.
Maybe you should use `thenRun`, it has no unused argument.
Another thing I noticed is that you always pass executor as argument into methods like "loadSnapshotMvData". Can't it be a field?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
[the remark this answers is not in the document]
ibessonov: Yes, we should

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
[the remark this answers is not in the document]
tkalkirill: I'll leave it as is.
`thenRun` does not return the future, but I need it so that the cancel works correctly with an interrupt.
I can even not use the field, it seems to me that it's a little clearer that we will execute the task in another thread.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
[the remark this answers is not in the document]
tkalkirill: Then, after discussion, a new approach should be applied everywhere.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
ibessonov: What do you mean by "tried to fix it"? `cancel()` call is still there. Is this really required? Please add a comment with a reason

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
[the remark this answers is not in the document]
ibessonov: You're right, I forgot about it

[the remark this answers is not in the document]
ibessonov: > I can even not use the field

Why?

ibessonov (approved):

ibessonov (approved):

PR #1234, opened 2022-10-21
IGNITE-17357 JMX metrics exporter

kgusakov:
https://issues.apache.org/jira/browse/IGNITE-17357


modules/metrics/src/main/java/org/apache/ignite/internal/metrics/exporters/jmx/JmxExporter.java
```diff
@@ -0,0 +1,165 @@
...
+
+    /**
+     * Logger.
+     */
+    private static IgniteLogger LOG = Loggers.forClass(JmxExporter.class);
+
+    /**
+     * Current registered MBeans.
+     */
+    private final List<ObjectName> mbeans = new ArrayList<>();
```
denis-chudov, line 59: looks like nothing protects us from attempts of concurrent modification of this collection, are you sure that ArrayList is suitable here?

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/exporters/jmx/JmxExporter.java
```diff
@@ -0,0 +1,165 @@
...
+            } else {
+                LOG.warn("Tried to unregister the MBean for non-registered metric set " + metricSetName);
+            }
+        } catch (MalformedObjectNameException e) {
+            LOG.error("MBean for metric set " + metricSetName + " can't be unregistered.", e);
+        }
+    }
+
+    /**
+     * Unregister MBean by its' name.
```
denis-chudov, line 154: typo?

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricManager.java
```diff
@@ -178,15 +192,19 @@ public static Map<String, MetricExporter> loadExporters() {
...
     /**
-     * Disable metric source by name. See {@link MetricRegistry#disable(String)}.
+     * Disable metric set by name. See {@link MetricRegistry#disable(String)}.
      *
-     * @param srcName Source name.
+     * @param metricSetName Metric set name.
      */
-    public void disable(final String srcName) {
-        registry.disable(srcName);
+    public void disable(final String metricSetName) {
```
denis-chudov, line 204: I don't completely understand this change. It refers to `MetricRegistry#disable(String)` but it disables metric source by its name. Also, it brings inconsistency to API, as `MetricManager#enable(String)` enables metric source, not metric set. Finally, only metric source can be enabled/disabled. Could you explain?

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/exporters/jmx/JmxExporter.java
kgusakov: This private collection is supposed to use only from synchronized methods

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/exporters/jmx/JmxExporter.java
kgusakov: fixed

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MetricManager.java
kgusakov: Renamed by mistake. Fixed, thanks.

PR #1236, opened 2022-10-21
IGNITE-17941 Add learners to Raft API

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-17941


modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
```diff
@@ -0,0 +1,338 @@
...
+    @Test
+    public void testLostLearners() throws Exception {
+        RaftNode follower = nodes.get(0);
+        List<RaftNode> learners = nodes.subList(1, nodes.size());
+
+        List<CompletableFuture<RaftGroupService>> services = nodes.stream()
+                .map(node -> startRaftGroup(node, new TestRaftGroupListener(), List.of(follower), learners))
+                .collect(toList());
+
+        // Wait for the leader to be elected.
```
ibessonov, line 253: Comment got duplicated

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    private CompletableFuture<RaftGroupService> startRaftGroup(
+            RaftNode raftNode,
+            RaftGroupListener listener,
+            List<RaftNode> peers,
+            List<RaftNode> learners
+    ) {
+        try {
+            CompletableFuture<RaftGroupService> future = raftNode.loza.prepareRaftGroup(
```
ibessonov, line 272: There should be a version of this method that takes listeners directly, not their suppliers. Or maybe I'm wrong

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
```diff
@@ -0,0 +1,338 @@
...
+                    RAFT_GROUP_NAME,
+                    peers.stream().map(RaftNode::localMember).collect(toList()),
+                    learners.stream().map(RaftNode::localMember).collect(toList()),
+                    () -> listener,
+                    () -> noopLsnr,
+                    RaftGroupOptions.defaults()
+            );
+
+            return future.thenApply(s -> {
+                s.timeout(100);
```
ibessonov, line 282: Please add a comment for this line. I believe it is here to make tests faster

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
```diff
@@ -0,0 +1,338 @@
...
+
+                return s;
+            });
+        } catch (NodeStoppingException e) {
+            throw new RuntimeException(e);
```
ibessonov, line 287: Why not `fail(...)`?

modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
```diff
@@ -305,25 +280,58 @@ public CompletableFuture<RaftGroupService> startRaftGroupService(
...
+            String grpId,
+            List<Peer> peers,
+            List<Peer> learners,
+            RaftGroupListener lsnr,
+            RaftGroupEventsListener raftGrpEvtsLsnr,
+            RaftGroupOptions groupOptions
+    ) {
+        assert !peers.isEmpty();
+
+        LOG.info("Start new raft node for group={} with initial peers={}", grpId, peers);
```
ibessonov, line 300: Should we put it into `if (infoEnabled())`? That's usually a good practice

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
```diff
@@ -0,0 +1,338 @@
...
+        public boolean onSnapshotLoad(Path path) {
+            return true;
+        }
+
+        @Override
+        public void onShutdown() {
+        }
+    }
+
+    private static class TestWriteCommand implements WriteCommand {
```
ibessonov, line 327: You know what, can you make it a network message in advance? We're porting all commands right now, so you'll save us some time

modules/raft-client/src/main/java/org/apache/ignite/raft/client/service/RaftGroupService.java
```diff
@@ -239,18 +237,6 @@ public interface RaftGroupService {
...
-     * Runs a read command on a given peer.
-     *
-     * <p>Read commands can see stale data (in the past).
-     *
-     * @param peer Peer id.
-     * @param cmd  The command.
-     * @param <R>  Execution result type.
-     * @return A future with the execution result.
-     */
-    <R> CompletableFuture<R> run(Peer peer, ReadCommand cmd);
```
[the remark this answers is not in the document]
sashapolo: This API is not used and is not intended to be used, NetworkMessages will be sent directly to the replica nodes

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
sashapolo: Why?

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
sashapolo: Yes, you are right

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
sashapolo: Java compiler does not know that `fail` always throws, so I will have to add `return null`, which is ugly

modules/raft-client/src/main/java/org/apache/ignite/raft/client/service/RaftGroupService.java
```diff
@@ -49,7 +49,7 @@ public interface RaftGroupService {
     /**
      * Returns group id.
      */
-    String groupId();
+    ReplicationGroupId groupId();
```
ibessonov, line 52: Is this OK to make this replacement if we don't fully own a RAFT code? Can you consult with @alievmirza?

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
```diff
@@ -73,19 +77,38 @@
  */
 @ExtendWith(ConfigurationExtension.class)
 public class ItLearnersTest extends IgniteAbstractTest {
-    private static final String RAFT_GROUP_NAME = "test";
+    private static final ReplicationGroupId RAFT_GROUP_ID = new ReplicationGroupId() {
+        @Override
+        public String toString() {
+            return "test";
```
ibessonov, line 83: Shouldn't there be a dedicated method for the id? `toString()` usually has a different meaning

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
```diff
@@ -73,19 +77,38 @@
...
+        }
+    };
 
     private static final List<NetworkAddress> ADDRS = List.of(
             new NetworkAddress("localhost", 5000),
             new NetworkAddress("localhost", 5001),
             new NetworkAddress("localhost", 5002)
     );
 
+    private static final TestRaftMessagesFactory msgFactory = new TestRaftMessagesFactory();
```
ibessonov, line 93: Constants should be named in upper case

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
ibessonov: I mistaken it for `start...` method. Anyway, this approach with suppliers of listeners does stink, if you ask me, I tried to partially eliminate it

modules/raft-client/src/main/java/org/apache/ignite/raft/client/service/RaftGroupService.java
ibessonov: I guess it's OK after all, I'm just paranoid about raft packages

sashapolo: What replacement are you talking about?

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
sashapolo: Again, I don't understand. This class wasn't introduced by me, I'm just playing by the rules introduced by others

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
sashapolo: Technically this is not a constant, it's a static factory)

modules/raft-client/src/main/java/org/apache/ignite/raft/client/service/RaftGroupService.java
ibessonov: String -> ReplicationGroupId

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
ibessonov: Didn't noticed that at first

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
ibessonov: You know what I mean)

modules/raft/src/integrationTest/java/org/apache/ignite/internal/raft/ItLearnersTest.java
ibessonov: I see

modules/raft-client/src/main/java/org/apache/ignite/raft/client/service/RaftGroupService.java
alievmirza: Seems that this changes was introduced by Vlad today, I'm not aware of them

ibessonov (approved):

PR #1238, opened 2022-10-23
IGNITE-17954 Make inspections and codestyle IDEA config apply by default. Create default checkstyle IDEA config.

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-17954


sashapolo (approved):
Please fix missing line breaks in all files

PR #1239, opened 2022-10-24
IGNITE-17955 Reduce memory consumption during OrderingFuture completion

rpuch:
https://issues.apache.org/jira/browse/IGNITE-17955


modules/core/src/main/java/org/apache/ignite/internal/future/OrderingFuture.java
```diff
@@ -559,4 +569,16 @@ private enum Phase {
          */
         COMPLETED
     }
+
+    private static class NotificationContext {
```
tkalkirill, line 573: Lack of documentation for the class and method

modules/core/src/main/java/org/apache/ignite/internal/future/OrderingFuture.java
```diff
@@ -529,12 +537,14 @@ public void notifyHeadToTail(T result, Throwable exception, ListNode<T> lastNoti
                 stack.addFirst(node);
             }
 
+            NotificationContext context = new NotificationContext();
```
tkalkirill, line 540: Why not just store the exception in a local variable and just pass it to `org.apache.ignite.internal.future.OrderingFuture.DependentAction#onCompletion`?

rpuch: We need to create the exception lazily because not all dependents need it, so it might happen that no dependent needs the exception. Context allows us to implement that lazy creation logic. Please note that it is much more expensive to create an exception that a context object.

tkalkirill: In a local variable, the exception will also be calculated lazily, I think a class is not needed for this.

tkalkirill (approved):

PR #1249, opened 2022-10-26
IGNITE-17786 Add --verbose option to all commands

valepakh:
https://issues.apache.org/jira/browse/IGNITE-17786


modules/cli/src/test/java/org/apache/ignite/internal/cli/commands/flow/FlowTest.java
```diff
@@ -53,6 +55,25 @@ class FlowTest {
     private StringWriter out;
     private StringWriter errOut;
 
+    @Test
+    void testVerbose() throws IOException {
```
PakhomovAlexander, line 59: this test is good and it would be nice to test CallExcecutionPipeline too. Maybe it could be just a regular command test but with --verbose flag.

PakhomovAlexander (changes requested):

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/flow/builder/FlowBuilderImpl.java
```diff
@@ -46,9 +47,10 @@
     private final Flow<I, O> flow;
     private final ExceptionHandlers exceptionHandlers;
     private final DecoratorRegistry decoratorRegistry;
+    private boolean verbose;
```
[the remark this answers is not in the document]
valepakh: It's not final because it's set by the method, not the constructor.

PakhomovAlexander (approved):

PakhomovAlexander:
LGTM

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/BaseCommand.java
```diff
@@ -29,6 +29,9 @@ public abstract class BaseCommand {
     @Option(names = {"-h", "--help"}, usageHelp = true, description = "Show this help message and exit.")
     protected boolean usageHelpRequested;
 
+    @Option(names = {"-v", "--verbose"}, description = "Show additional information about errors.")
+    protected boolean verbose;
```
SammyVimes, line 33: Is it always about errors?

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/flow/builder/FlowBuilder.java
```diff
@@ -102,6 +102,14 @@ default <OT> FlowBuilder<I, OT> map(Function<O, OT> mapper) {
      */
     FlowBuilder<I, O> exceptionHandler(ExceptionHandler<?> exceptionHandler);
 
+    /**
+     * Adds verbose output from debug log to the output.
+     *
+     * @param verbose if @{code true}, flow execution will print debug logs
+     * @return instance of builder
```
SammyVimes, line 109:
```suggestion
     * @param verbose If @{code true}, flow execution will print debug logs.
     * @return Builder instance.
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/logger/HttpLogging.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * Stops logging previously started by {@link HttpLogging#startHttpLogging(PrintWriter)}.
+     */
+    public static void stopHttpLogging() {
+        ApiClient client = Configuration.getDefaultApiClient();
+        Builder builder = client.getHttpClient().newBuilder();
+
+        builder.interceptors().clear();
```
SammyVimes, line 52: Won't it clear **all** interceptors when we only want to stop http log? 
Also, adding a test checking correct behaviour of this method would be cool

SammyVimes (changes requested):

SammyVimes (approved):
LGTM

PR #1252, opened 2022-10-26
IGNITE-17968 Fix write-intents being filtered out in case if it's a tombstone

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-17968

### DIFF
--- a/modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
+++ b/modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
@@ -1290,23 +1290,13 @@ public abstract class AbstractMvPartitionStorageTest extends BaseMvStoragesTest 
     @ParameterizedTest
     @EnumSource(ScanTimestampProvider.class)
     public void scanDoesNotSeeTombstonesWhenTombstoneIsCommitted(ScanTimestampProvider tsProvider) throws Exception {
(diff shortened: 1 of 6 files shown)


modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
```diff
@@ -1269,8 +1269,22 @@ public void scanCursorHasNextReturnsFalseEachTimeAfterExhaustion(ScanTimestampPr
 
     @ParameterizedTest
     @EnumSource(ScanTimestampProvider.class)
-    public void scanDoesNotSeeTombstonesWhenTombstoneIsNotCommitted(ScanTimestampProvider tsProvider) throws Exception {
-        testScanDoesNotSeeTombstones(tsProvider, false);
+    public void scanSeesTombstonesWhenTombstoneIsNotCommitted(ScanTimestampProvider tsProvider) throws Exception {
```
rpuch, line 1272: Let's inline `testScanDoesNotSeeTombstones()`, it looks like it is now being used only in one place

rpuch (commented):
1. Let's add clarifications to `#scan()` javadocs: when exactly tombstones might be returned by a scan
2. Does `read()` have same properties (like 'interpreting only not-yet-committed tombstones')? Do we have tests about this property? If `read()` behaves differently from `scan()` in this regard, we probably need to note this in javadocs on both methods.

rpuch (approved):

PR #1255, opened 2022-10-26
IGNITE-17967 RO writeIntent resolution tests hang up in case of multi…

vldpyatkov:
… node cluster


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -327,7 +327,7 @@ private CompletableFuture<TxMeta> getTxStateConcurrently(TxStateReplicaRequest t
      * @param isPrimary Whether the given replica is primary.
      * @return Result future.
      */
-    private CompletableFuture<Object> processReadOnlyScanRetrieveBatchAction(
+    private CompletableFuture processReadOnlyScanRetrieveBatchAction(
```
sanpwc, line 330: Why?

PR #1257, opened 2022-10-26
IGNITE-17771 Get rid of internal classes in API module

AMashenkov:


ptupitsyn (approved):

modules/api/src/main/java/org/apache/ignite/lang/IgniteException.java
```diff
@@ -289,4 +288,14 @@ public static IgniteException wrap(Throwable e) {
...
         return new IgniteException(UNKNOWN_ERR, e.getMessage(), e);
     }
+
+    /**
+     * Gets the Ignite error code if the specified throwable is an {@link IgniteException}.
+     *
+     * @param t Throwable.
+     * @return Ignite error code or UNKNOWN_ERR.
+     */
+    public static int getIgniteErrorCode(Throwable t) {
```
ygerzhedovich, line 298: good change. It will fix [IGNITE-17970](https://issues.apache.org/jira/browse/IGNITE-17970)

modules/api/src/main/java/org/apache/ignite/sql/ResultSetImpl.java
```diff
@@ -15,25 +15,21 @@
  * limitations under the License.
  */
 
-package org.apache.ignite.internal.sql;
+package org.apache.ignite.sql;
```
ygerzhedovich, line 18: The class should be moved outside of API module, but you just move it to another package in the same module

AMashenkov: I've changed class visibility to package-private. 
Otherwise, we would have few copies of the same class in project. However, the class is just a sync adapter for async result set and provides no additional logic.

modules/marshaller-common/src/main/java/org/apache/ignite/internal/util/Pair.java
```diff
@@ -1,57 +0,0 @@
...
-
-package org.apache.ignite.internal.util;
-
-/**
- * Pair of objects.
- *
- * @param <T> First object.
- * @param <V> Second object.
- */
-public class Pair<T, V> {
```
korlov42, line 26: I'd rather prefer to get rid of IgniteBiTuple

korlov42 (approved):

PR #1258, opened 2022-10-26
IGNITE-17966: Fix problem with stuck Gradle processes in .NET tests

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-17966

* After CI infra fixes we can again enable .NET tests via Gradle
* Fix gradle build


SammyVimes (approved):
LGTM

PR #1262, opened 2022-10-27
IGNITE-17983 Fix work with the last applied index in RAFT

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-17983


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -287,16 +291,15 @@ private void handleSafeTimeSyncCommand(SafeTimeSyncCommand cmd) {
     /** {@inheritDoc} */
     @Override
     public void onSnapshotSave(Path path, Consumer<Throwable> doneClo) {
-        storage.flush();
+        CompletableFuture.allOf(storage.flush(), txStateStorage.flush())
```
sanpwc, line 294: max index setting is missing
```
        long maxLastAppliedIndex = Math.max(storage.lastAppliedIndex(), txStateStorage.lastAppliedIndex());
        storage.lastAppliedIndex(maxLastAppliedIndex);
        txStateStorage.lastAppliedIndex(maxLastAppliedIndex);
```

PR #1263, opened 2022-10-27
IGNITE-17302 add the propagation of the maximal storages' lastAppliedIndex on onSnapshotSave

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-17302


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -275,7 +276,17 @@ private void handleTxCleanupCommand(TxCleanupCommand cmd, long commandIndex) {
     /** {@inheritDoc} */
     @Override
     public void onSnapshotSave(Path path, Consumer<Throwable> doneClo) {
-        storage.flush();
+        long maxLastAppliedIndex = Math.max(storage.lastAppliedIndex(), txStateStorage.lastAppliedIndex());
```
tkalkirill, line 279: Please describe why you need to choose the maximum.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -275,7 +276,17 @@ private void handleTxCleanupCommand(TxCleanupCommand cmd, long commandIndex) {
...
+
+        storage.lastAppliedIndex(maxLastAppliedIndex);
```
tkalkirill, line 281: I think that it would be more correct:
```
        storage.runConsistently(() -> {
            storage.lastAppliedIndex(maxLastAppliedIndex);
            
            return null;
        });
```

alievmirza: done

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
alievmirza: done

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/PartitionCommandListenerTest.java
```diff
@@ -206,6 +215,42 @@ public void testInsertRowsBatchedAndCheck() {
...
+
+        AtomicLong counter = new AtomicLong(0);
+
+        testCommandListener.onSnapshotSave(workDir, (throwable) -> {
+            counter.incrementAndGet();
+        });
+
+        assertEquals(1L, counter.get());
+
+        assertEquals(5L, partitionDataStorage.lastAppliedIndex());
```
tkalkirill, line 249: Here it is not checked what exactly the maximum was taken, but only for `partitionDataStorage` let's run **onSnapshotSave** one more time but with a maximum for `txStateStorage`.

modules/transactions/src/testFixtures/java/org/apache/ignite/internal/tx/storage/state/test/TestConcurrentHashMapTxStateStorage.java
```diff
@@ -57,13 +57,17 @@ public boolean compareAndSet(UUID txId, TxState txStateExpected, @NotNull TxMeta
             if (old == null && txStateExpected == null) {
                 TxMeta oldMeta = storage.putIfAbsent(txId, txMeta);
                 if (oldMeta == null) {
```
tkalkirill, line 59: For example, this check is not needed, let's do something like `TxStateRocksDbStorage#compareAndSet`

tkalkirill: Fix it

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/PartitionCommandListenerTest.java
tkalkirill: fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -108,19 +108,20 @@ public void onRead(Iterator<CommandClosure<ReadCommand>> iterator) {
...
 
-            long storageAppliedIndex = storage.lastAppliedIndex();
+            // We choose the minimum applied index, since we choose it (the minimum one) on local recovery so as not to lose the data for
+            // one of the storages.
+            long storagesAppliedIndex = Math.min(storage.lastAppliedIndex(), txStateStorage.lastAppliedIndex());
 
-            assert storageAppliedIndex < commandIndex
-                    : "Pending write command has a higher index than already processed commands [commandIndex=" + commandIndex
-                    + ", storageAppliedIndex=" + storageAppliedIndex + ']';
+            assert commandIndex > storagesAppliedIndex :
```
ibessonov, line 122: I would add both indexes to the assertion message

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -108,19 +108,20 @@ public void onRead(Iterator<CommandClosure<ReadCommand>> iterator) {
...
     public void onWrite(Iterator<CommandClosure<WriteCommand>> iterator) {
         iterator.forEachRemaining((CommandClosure<? extends WriteCommand> clo) -> {
             Command command = clo.command();
 
             long commandIndex = clo.index();
 
-            long storageAppliedIndex = storage.lastAppliedIndex();
+            // We choose the minimum applied index, since we choose it (the minimum one) on local recovery so as not to lose the data for
+            // one of the storages.
+            long storagesAppliedIndex = Math.min(storage.lastAppliedIndex(), txStateStorage.lastAppliedIndex());
```
sanpwc, line 120: That means, that we should add storage specific skip logic. Probably inside storage. Or we already have it?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -147,6 +148,7 @@ public void onWrite(Iterator<CommandClosure<WriteCommand>> iterator) {
      * Handler for the {@link UpdateCommand}.
      *
      * @param cmd Command.
+     * @param commandIndex Index of the RAFT command.
      */
     private void handleUpdateCommand(UpdateCommand cmd, long commandIndex) {
         storage.runConsistently(() -> {
```
ibessonov, line 154: I see no place where you skip already applied commands

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -284,19 +288,32 @@ private void handleSafeTimeSyncCommand(SafeTimeSyncCommand cmd) {
...
 
-    /** {@inheritDoc} */
     @Override
     public void onSnapshotSave(Path path, Consumer<Throwable> doneClo) {
-        storage.flush();
+        // We need to take the max index here because otherwise there is a scenario on a node restart when a node will install
+        // snapshot instead of starting even if all data is presented. This might happen because a raft log is truncated up
+        // to the maximal index among storages after a snapshot is done in raft,
+        // and because we take minimal applied index among storages for the local recovery. In case of a cluster restart,
+        // it is possible that cluster even can't start, if we won't take max index.
```
ibessonov, line 297: This is a very convoluted description, maybe even wrong partially, please change it. I don't think that mentioning leader is necessary here. This code is required for local recovery only and a possible scenario of false node failure when we actually have all required data

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
tkalkirill: Fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
tkalkirill: Tried to fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
tkalkirill: Tried to fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -107,19 +107,22 @@ public void onRead(Iterator<CommandClosure<ReadCommand>> iterator) {
...
+            // We choose the minimum applied index, since we choose it (the minimum one) on local recovery so as not to lose the data for
+            // one of the storages.
+            long storagesAppliedIndex = Math.min(storage.lastAppliedIndex(), txStateStorage.lastAppliedIndex());
 
-            assert storageAppliedIndex < commandIndex
-                    : "Pending write command has a higher index than already processed commands [commandIndex=" + commandIndex
-                    + ", storageAppliedIndex=" + storageAppliedIndex + ']';
+            assert commandIndex > storagesAppliedIndex :
+                    "Write command must have an index greater than that of storages [commandIndex=" + commandIndex
+                            + ", storagesAppliedIndex=" + storagesAppliedIndex
```
ibessonov, line 123: Why do you need this in the message now? It can be derived from the other two storage indexes.

tkalkirill: Fix it

ibessonov (approved):

sanpwc (approved):

PR #1267, opened 2022-10-27
IGNITE-17417: Add option '--node-name' to CLI commands

Flaugh24:
Add option '--node-name' to CLI commands. 
https://issues.apache.org/jira/browse/IGNITE-17417

Share information about REST between cluster nodes via io.scalecube.cluster.Cluster#updateMetadata. 
Add information about REST to physical cluster topology.
Add pulling information about REST to the CLI.
Add option '--node-name' to node commands in the CLI.


modules/api/src/main/java/org/apache/ignite/network/NodeMetadata.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * Deserializes {@link NodeMetadata} from {@link ByteBuffer}.
+     *
+     * @return {@link NodeMetadata}
+     */
+    public static NodeMetadata fromByteBuffer(ByteBuffer metadata) {
+        try {
+            int version = readInt(metadata, 0);
```
valepakh, line 44: Why not `metadata.getInt(0)`?

modules/api/src/test/java/org/apache/ignite/network/NodeMetadataTest.java
```diff
@@ -15,8 +15,28 @@
...
+    }
 
-package org.apache.ignite.internal.rest.configuration;
+    @Test
+    void testFromByteBufferWithWrongContent() {
+        ByteBuffer buffer = ByteBuffer.allocate(8);
+        NodeMetadata fromByteBuffer = NodeMetadata.fromByteBuffer(buffer);
+        assertNull(fromByteBuffer);
+    }
+}
```
valepakh, line 42: Please add a line break at the end of the file, it can be done in IDEA settings -> Editor -> General -> Ensure every saved file ends with a line break.

modules/cli/src/main/java/org/apache/ignite/internal/cli/call/connect/ConnectCall.java
```diff
@@ -61,7 +68,7 @@ public CallOutput<String> execute(ConnectCallInput input) {
             stateConfigProvider.get().setProperty(ConfigConstants.LAST_CONNECTED_URL, nodeUrl);
             session.setJdbcUrl(constructJdbcUrl(configuration, nodeUrl));
             session.setConnectedToNode(true);
-
```
valepakh, line 64: Please keep the line breaks intact, they are helpful for readability. There are some more missing line breaks in other files.

modules/cli/src/main/java/org/apache/ignite/internal/cli/call/connect/DisconnectCall.java
```diff
@@ -34,9 +35,12 @@
 public class DisconnectCall implements Call<EmptyCallInput, String> {
     @Inject
     private final Session session;
+    @Inject
```
valepakh, line 38: Micronaut injects the constructor arguments automatically so there's no need to annotate the fields.

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/connect/ConnectCommand.java
```diff
@@ -39,17 +45,37 @@ public class ConnectCommand extends BaseCommand implements Runnable {
     @Parameters(description = NODE_URL_DESC, descriptionKey = CLUSTER_URL_KEY, converter = UrlConverter.class)
     private URL nodeUrl;
 
+    /** Node name option. */
+    @Option(names = {NODE_NAME_OPTION_SHORT, NODE_NAME_OPTION}, description = NODE_NAME_DESC)
```
valepakh, line 49: Maybe we can combine these into the single parameter and detect whether this is a node name or a node URL?

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/DynamicCompleterFactory.java
```diff
@@ -34,16 +35,18 @@ public class DynamicCompleterFactory {
...
     private final ClusterConfigShowCall clusterConfigShowCall;
     private final NodeUrlProvider urlProvider;
+    private final NodeNameRegistry nodeNameRegistry;
 
     /** Default constructor. */
     public DynamicCompleterFactory(
             NodeConfigShowCall nodeConfigShowCall,
             ClusterConfigShowCall clusterConfigShowCall,
-            NodeUrlProvider urlProvider) {
+            NodeUrlProvider urlProvider, NodeNameRegistry nodeNameRegistry) {
```
valepakh, line 44:
```suggestion
            NodeUrlProvider urlProvider,
            NodeNameRegistry nodeNameRegistry) {
```

modules/api/src/test/java/org/apache/ignite/network/NodeMetadataTest.java
```diff
@@ -15,8 +15,28 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+import java.nio.ByteBuffer;
+import org.junit.jupiter.api.Test;
+
+class NodeMetadataTest {
+
+    @Test
+    void testToByteBufferAndFromByteBuffer() {
```
PakhomovAlexander, line 29: that is not required to write 'test' prefix

modules/api/src/test/java/org/apache/ignite/network/NodeMetadataTest.java
```diff
@@ -15,8 +15,28 @@
...
+        NodeMetadata metadata = new NodeMetadata(10000);
+        ByteBuffer buffer = metadata.toByteBuffer();
+        NodeMetadata fromByteBuffer = NodeMetadata.fromByteBuffer(buffer);
+        assertEquals(metadata, fromByteBuffer);
+    }
 
-package org.apache.ignite.internal.rest.configuration;
+    @Test
+    void testFromByteBufferWithWrongContent() {
```
PakhomovAlexander, line 37: It would be nice to add more test cases like the wrong version is set.

PakhomovAlexander: And why did you decide to return null but not to throw exception?

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public List<String> getAllNames() {
+        return new ArrayList<>(nodeNameToNodeUrl.keySet());
+    }
+
+    private void updateNodeNames(String nodeUrl) {
+        PhysicalTopologyCall topologyCall = new PhysicalTopologyCall();
+        topologyCall.execute(new UrlCallInput(nodeUrl)).body()
+                .forEach(it -> {
+                    nodeNameToNodeUrl.put(it.getName(), urlFromClusterNode(it));
```
PakhomovAlexander, line 78: `put` replaces existing values but outdated ones won't be removed.

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+                });
+    }
+
+    private static String urlFromClusterNode(ClusterNode node) {
+        Objects.requireNonNull(node, "node must not be null");
+        return node.getAddress().getHost() + ":" + node.getMetadata().getRestPort();
```
PakhomovAlexander, line 84: It might be different: the TCP node address and the REST server address.  So I think it might be useful to spread not only the port but the whole address through node metadata.

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/Session.java
```diff
@@ -65,4 +65,8 @@ public String jdbcUrl() {
     public void setJdbcUrl(String jdbcUrl) {
         this.jdbcUrl = jdbcUrl;
     }
+
+    public void onConnect() {
```
PakhomovAlexander, line 69: Empty method

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
```diff
@@ -51,12 +54,15 @@
...
 import org.apache.ignite.network.NodeFinderFactory;
+import org.apache.ignite.network.NodeMetadata;
 
 /**
  * Cluster service factory that uses ScaleCube for messaging and topology services.
  */
 public class ScaleCubeClusterServiceFactory {
-    /** Logger. */
+    /**
+     * Logger.
```
PakhomovAlexander, line 64: Unnecessary change.

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
```diff
@@ -131,7 +138,7 @@ public void onMembershipEvent(MembershipEvent event) {
                                 topologyService.onMembershipEvent(event);
                             }
                         })
-                        .config(opts -> opts.memberAlias(consistentId))
+                        .config(opts -> opts.memberAlias(consistentId).metadata(new NodeMetadata(restConfiguration.port().value())))
```
PakhomovAlexander, line 141: No, the rest configuration is taken from a file but there is a `portRange` config that could be used if the original rest.port is not available. So the REST server can be started on another port, look at `RestComponent.start()`.

PakhomovAlexander (changes requested):

Flaugh24:
https://issues.apache.org/jira/browse/IGNITE-17417

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
Flaugh24: done

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/NodeUrl.java
```diff
@@ -0,0 +1,33 @@
...
+
+/** Node URL. */
+public class NodeUrl {
+    private URL nodeUrl;
+
+    public NodeUrl(URL nodeUrl) {
+        this.nodeUrl = nodeUrl;
+    }
+
+    public String getNodeUrl() {
```
PakhomovAlexander, line 30: `stringUrl()`

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/NodeUrlMixin.java
```diff
@@ -17,23 +17,62 @@
...
     public String getNodeUrl() {
-        return nodeUrl != null ? nodeUrl.toString() : null;
+        if (options == null) {
+            return null;
+        } else {
+            if (options.nodeUrl != null) {
+                return options.nodeUrl.toString();
+            } else {
+                return nodeNameRegistry.getNodeUrl(options.nodeName)
+                        .orElseThrow(() -> new IgniteCliException("Node " + options.nodeName + "not found. Provide valid name or use URL"));
```
PakhomovAlexander, line 74: `"not found` missing space.

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/converters/NodeNameOrUrlConverter.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.ignite.internal.cli.commands.node.NodeUrl;
+import org.apache.ignite.internal.cli.core.exception.ParameterException;
+import picocli.CommandLine;
+import picocli.CommandLine.TypeConversionException;
+
+/** Converter for {@link NodeUrl}. */
+public class NodeNameOrUrlConverter implements CommandLine.ITypeConverter<NodeUrl> {
+
+    private static final Pattern URL_PATTERN = Pattern.compile("^.*[/:].*");
+    private final NodeNameRegistry nodeNameRegistry;
```
PakhomovAlexander, line 33: eof

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/DynamicCompleterFactory.java
```diff
@@ -89,4 +93,13 @@ public LazyDynamicCompleter clusterConfigCompleter(Set<String> activationPrefixe
             }
         });
     }
+
+    public LazyDynamicCompleter nodeNameCompleter(String... activationPrefixes) {
```
PakhomovAlexander, line 97: Return DynamicCompleter, not lazy

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/DynamicCompleterFactory.java
```diff
@@ -89,4 +93,13 @@ public LazyDynamicCompleter clusterConfigCompleter(Set<String> activationPrefixe
...
+        return nodeNameCompleter(Set.of(activationPrefixes));
+    }
+
+    public LazyDynamicCompleter nodeNameCompleter(Set<String> activationPrefixes) {
+        return new LazyDynamicCompleter(() -> new NodeNameDynamicCompleter(activationPrefixes, nodeNameRegistry.getAllNames()));
```
PakhomovAlexander, line 102: You should not use LazyDynamicCompleter for node name completions. There is no reason for that. Also, LazyDynamicCompleter will initialize only once and **always** provide the same set of node names.

PakhomovAlexander (changes requested):

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/DynamicCompleterFactory.java
Flaugh24: done

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/DynamicCompleterFactory.java
Flaugh24: Done

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/converters/NodeNameOrUrlConverter.java
Flaugh24: Done

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/NodeUrlMixin.java
Flaugh24: Done

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/NodeUrl.java
Flaugh24: Done

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/NodeNameTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @BeforeEach
+    void setUp() {
+        nodeNameRegistry.startPullingUpdates("http://localhost:10301");
+        // wait to pulling node names
+        Await.await(() -> !nodeNameRegistry.getAllNames().isEmpty(), Duration.ofSeconds(5));
+        Optional<String> nodeName = nodeNameRegistry.getAllNames().stream().findFirst();
+        Assertions.assertTrue(nodeName.isPresent());
+        this.nodeName = nodeName.get();
+
```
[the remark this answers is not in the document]
Flaugh24: Done

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/NodeNameTest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+/** Tests for ignite node commands with a provided node name. */
+public class NodeNameTest extends CliCommandTestNotInitializedIntegrationBase {
+
+    @Inject
+    NodeNameRegistry nodeNameRegistry;
+    private String nodeName;
```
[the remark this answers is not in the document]
Flaugh24: Done

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/Await.java
```diff
@@ -0,0 +1,49 @@
...
+
+/** Utility class to await some condition. */
+public final class Await {
+
+    private Await() {
+    }
+
+    /** Awaits a condition. */
+    public static void await(Supplier<Boolean> supplier, Duration timeout) {
+        long startTime = System.currentTimeMillis();
```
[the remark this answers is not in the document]
Flaugh24: Done

modules/api/src/main/java/org/apache/ignite/network/NodeMetadata.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public NodeMetadata() {
+    }
+
+    public NodeMetadata(String restHost, int restPort) {
+        this.restHost = restHost;
+        this.restPort = restPort;
+    }
+
+    public String getRestHost() {
```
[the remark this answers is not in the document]
Flaugh24: Done

modules/api/src/main/java/org/apache/ignite/network/NodeMetadata.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Objects;
+
+/**
+ * Contains metadata of the cluster node.
+ */
+public class NodeMetadata implements Serializable {
+    private static final long serialVersionUID = 3216463261002854096L;
+
+    private String restHost;
+    private int restPort;
```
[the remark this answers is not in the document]
Flaugh24: Dome

PakhomovAlexander:
LGTM

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    private void updateNodeNames(String nodeUrl) {
+        readWriteLock.writeLock().lock();
+        try {
+            nodeNameToNodeUrl.clear();
+            PhysicalTopologyCall topologyCall = new PhysicalTopologyCall();
+            topologyCall.execute(new UrlCallInput(nodeUrl)).body()
+                    .forEach(it -> {
+                        nodeNameToNodeUrl.put(it.getName(), urlFromClusterNode(it.getMetadata()));
```
sashapolo, line 100: This may be a bug: `urlFromClusterNode` may return `null`, while `ConcurrentHashMap` does not allow null values

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.ignite.internal.cli.call.cluster.topology.PhysicalTopologyCall;
+import org.apache.ignite.internal.cli.core.call.UrlCallInput;
+import org.apache.ignite.rest.client.model.NodeMetadata;
+
+/**
+ * Registry to get a node URL by a node name.
+ */
+@Singleton
+public class NodeNameRegistry {
+    private final Map<String, String> nodeNameToNodeUrl = new ConcurrentHashMap<>();
```
sashapolo, line 40: Looks like you don't need a `ConcurrentHashMap` here, it's always accessed under the `readWriteLock`

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
```diff
@@ -0,0 +1,113 @@
...
+        readWriteLock.readLock().lock();
+        try {
+            return new ArrayList<>(nodeNameToNodeUrl.keySet());
+        } finally {
+            readWriteLock.readLock().unlock();
+        }
+    }
+
+    private void updateNodeNames(String nodeUrl) {
+        readWriteLock.writeLock().lock();
```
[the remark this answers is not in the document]
sashapolo: Nice suggestion, totally agree. You don't even need any CAS here, simple volatile field will be sufficient

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/OptionsConstants.java
```diff
@@ -38,6 +38,9 @@ public class OptionsConstants {
     /** Node URL option description. */
     public static final String NODE_URL_DESC = "URL of ignite node";
 
+    /** Node URL or name option description. */
+    public static final String NODE_URL_OR_NAME_DESC = "URL or name of ignite node";
```
sashapolo, line 42:
```suggestion
    public static final String NODE_URL_OR_NAME_DESC = "URL or name of an Ignite node";
```

sashapolo: Please fix other descriptions accordingly

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/NodeUrl.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.cli.commands.node;
+
+import java.net.URL;
+
+/** Node URL. */
+public class NodeUrl {
+    private URL nodeUrl;
```
sashapolo, line 24:
```suggestion
    private final URL nodeUrl;
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/NodeUrl.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.cli.commands.node;
+
+import java.net.URL;
+
+/** Node URL. */
+public class NodeUrl {
```
sashapolo, line 23: What's the purpose of this class? Can we simply use `URL` instead?

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/NodeUrlMixin.java
```diff
@@ -17,23 +17,63 @@
...
-        return nodeUrl != null ? nodeUrl.toString() : null;
+        if (options == null) {
+            return null;
+        } else {
+            if (options.nodeUrl != null) {
+                return options.nodeUrl.toString();
+            } else {
+                return nodeNameRegistry.getNodeUrl(options.nodeName)
+                        .orElseThrow(() -> new IgniteCliException("Node " + options.nodeName
+                                + " not found. Provide valid name or use URL"));
```
sashapolo, line 75:
```suggestion
                                + " not found. Provide a valid name or use a URL"));
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/converters/NodeNameOrUrlConverter.java
```diff
@@ -0,0 +1,62 @@
...
+        try {
+            return new URL(str);
+        } catch (MalformedURLException e) {
+            throw new TypeConversionException("Invalid URL '" + str + "' (" + e.getMessage() + ")");
+        }
+    }
+
+    @Override
+    public NodeUrl convert(String input) throws Exception {
+        boolean isUrl = URL_PATTERN.matcher(input).find();
```
sashapolo, line 49: I think `matches()` is a more suitable call here

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/converters/NodeNameOrUrlConverter.java
```diff
@@ -0,0 +1,62 @@
...
+        if (isUrl) {
+            return new NodeUrl(stringToUrl(input));
+        } else {
+            return new NodeUrl(findNodeUrlByNodeName(input));
+        }
+    }
+
+    private URL findNodeUrlByNodeName(String name) {
+        return nodeNameRegistry.getNodeUrl(name)
+                .map(NodeNameOrUrlConverter::stringToUrl)
```
sashapolo, line 59: Why do we store URL as Strings inside the `nodeNameRegistry`? I think we should store URLs there, this way we will avoid unnecessary conversions

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/converters/NodeNameOrUrlConverter.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.regex.Pattern;
+import org.apache.ignite.internal.cli.NodeNameRegistry;
+import org.apache.ignite.internal.cli.commands.node.NodeUrl;
+import picocli.CommandLine;
+import picocli.CommandLine.TypeConversionException;
+
+/** Converter for {@link NodeUrl}. */
+public class NodeNameOrUrlConverter implements CommandLine.ITypeConverter<NodeUrl> {
+
+    private static final Pattern URL_PATTERN = Pattern.compile("^.*[/:].*");
```
sashapolo, line 31: So we distinguish a URL from a name by looking for a semicolon and slashes? Maybe we can remove this logic altogether and simply try to always parse a URL and fallback to node name lookup if it fails?

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/NodeNameDynamicCompleter.java
```diff
@@ -0,0 +1,66 @@
...
+                    .filter(it -> it.startsWith(lastWord))
+                    .collect(Collectors.toList());
+        } else {
+            return Collections.emptyList();
+        }
+    }
+
+    private static String beforeLastNotEmptyWord(int index, String[] words) {
+        for (int i = words.length - 1 - index; i >= 0; i--) {
+            if (!words[i].isEmpty()) {
```
sashapolo, line 60: Shall we use `isBlank` instead of `isEmpty`?

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/NodeNameDynamicCompleter.java
```diff
@@ -0,0 +1,66 @@
...
+            return nodeNameRegistry.getAllNames().stream()
+                    .filter(it -> it.startsWith(lastWord))
+                    .collect(Collectors.toList());
+        } else {
+            return Collections.emptyList();
+        }
+    }
+
+    private static String beforeLastNotEmptyWord(int index, String[] words) {
+        for (int i = words.length - 1 - index; i >= 0; i--) {
```
sashapolo, line 59: This method does not check if `index` is actually a valid index inside the `words` array, is that ok?

modules/cli/src/main/java/org/apache/ignite/internal/cli/util/ArrayUtils.java
```diff
@@ -0,0 +1,35 @@
...
+
+/** Utility class for arrays. */
+public final class ArrayUtils {
+
+    private ArrayUtils() {
+    }
+
+    /** Finds the last not empty word in array. */
+    public static String findLastNotEmptyWord(String[] words) {
+        for (int i = words.length - 1; i >= 0; i--) {
```
sashapolo, line 28: Same thing about boundaries checks and `isBlank`

modules/network-api/src/main/java/org/apache/ignite/network/ClusterService.java
```diff
@@ -56,5 +56,7 @@ default void stop() {
      *
      * @return {@code true} if cluster service is stopped, {@code false} otherwise.
      */
-    public boolean isStopped();
+    boolean isStopped();
+
+    void updateMetadata(NodeMetadata metadata);
```
sashapolo, line 61: Please add a javadoc (I wonder if you can even pass the CI checks without it)

modules/network/src/main/java/org/apache/ignite/network/scalecube/NodeMetadataCodec.java
```diff
@@ -0,0 +1,72 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.scalecube.cluster.metadata.MetadataCodec;
+import java.nio.ByteBuffer;
+import org.apache.ignite.network.NodeMetadata;
+
+/**
+ * Codec for serialization and deserialization of {@link NodeMetadata}.
+ */
+public class NodeMetadataCodec implements MetadataCodec {
```
sashapolo, line 29: Why do you need this class? The default `MetadataCodec` in Scalecube is `JdkMetadataCodec` which is enough for our purposes, since `NodeMetadata` is `Serializable`. I think we don't need this class and Jackson dependency as well

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
```diff
@@ -59,6 +66,11 @@ public class ScaleCubeClusterServiceFactory {
     /** Logger. */
     private static final IgniteLogger LOG = Loggers.forClass(ScaleCubeClusterServiceFactory.class);
 
+    /** Metadata codec. */
+    private static final NodeMetadataCodec METADATA_CODEC = NodeMetadataCodec.INSTANCE;
+    /** Scalecube cluster executor to update metadata.  */
+    private final Scheduler scheduler = Schedulers.fromExecutor(Executors.newSingleThreadExecutor());
```
sashapolo, line 72: Why do you need this executor?

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
```diff
@@ -190,6 +201,17 @@ public void beforeNodeStop() {
...
+
+            @Override
+            public void updateMetadata(NodeMetadata metadata) {
+                cluster.updateMetadata(metadata).subscribeOn(scheduler).subscribe();
+                MembershipEvent membershipEvent = createUpdated(cluster.member(),
+                        cluster.<NodeMetadata>metadata().map(METADATA_CODEC::serialize).orElse(null),
+                        METADATA_CODEC.serialize(metadata),
+                        System.currentTimeMillis()
+                );
+                topologyService.onMembershipEvent(membershipEvent);
```
sashapolo, line 213: Why do we need to emit an event here? I can see that `updateMetadata` calls `updateIncarnation`, which will propagate this information to the remote peers

sashapolo (changes requested):

modules/api/src/main/java/org/apache/ignite/network/NodeMetadata.java
```diff
@@ -0,0 +1,71 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonGetter;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * Contains metadata of the cluster node.
+ */
+public class NodeMetadata implements Serializable {
```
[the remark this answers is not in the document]
Flaugh24: Scalecube requires serializable whether you use java serialization or not.

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
Flaugh24: Because we should provide an executor when calling io.scalecube.cluster.ClusterImpl#updateMetadata. Without that metadata will not be updated ever.

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
Flaugh24: It's a local event. We have the same logic when a cluster starts. See org.apache.ignite.network.AbstractClusterService#start

sashapolo: why do we need a local event in this particular case?

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
sashapolo: Why? I can see that `updateMetadata` is dispatched on a ScaleCube's scheduler internally:
```
return Mono.fromRunnable(() -> metadataStore.updateMetadata(metadata))
        .then(membership.updateIncarnation())
        .subscribeOn(scheduler);
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/NodeUrl.java
Flaugh24: We use it for cases when a parameter can be a URL or name. It has a special converter. And we aren't going to use that converter in cases when a parameter is just a URL. So we have to differentiate them.

modules/cli/src/main/java/org/apache/ignite/internal/cli/util/ArrayUtils.java
Flaugh24: What's wrong with boundaries?

sashapolo: oh, sorry, comment about boundaries is not applicable here

modules/api/src/main/java/org/apache/ignite/network/ClusterNode.java
```diff
@@ -33,6 +33,23 @@ public class ClusterNode implements Serializable {
     /** Network address of this node. */
     private final NetworkAddress address;
 
+    private final NodeMetadata nodeMetadata;
```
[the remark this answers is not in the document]
Flaugh24: done

modules/api/src/main/java/org/apache/ignite/network/NodeMetadata.java
```diff
@@ -0,0 +1,71 @@
...
+    private static final long serialVersionUID = 3216463261002854096L;
+
+    private String restHost;
+
+    private int restPort;
+
+    public NodeMetadata() {
```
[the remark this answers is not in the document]
Flaugh24: removed

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/NodeNameTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+/** Tests for ignite node commands with a provided node name. */
+public class NodeNameTest extends CliCommandTestNotInitializedIntegrationBase {
+
+    @Inject
+    private NodeNameRegistry nodeNameRegistry;
```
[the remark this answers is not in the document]
Flaugh24: removed

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.ignite.rest.client.model.NodeMetadata;
+
+/**
+ * Registry to get a node URL by a node name.
+ */
+@Singleton
+public class NodeNameRegistry {
+    private final Map<String, String> nodeNameToNodeUrl = new ConcurrentHashMap<>();
+    private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
+    private volatile ScheduledExecutorService executor;
```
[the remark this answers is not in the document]
Flaugh24: removed

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * Start pulling updates from a node.
+     *
+     * @param nodeUrl Node URL.
+     */
+    public void startPullingUpdates(String nodeUrl) {
+        stopPullingUpdates();
+        synchronized (this) {
```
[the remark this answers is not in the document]
Flaugh24: done

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+    }
+
+    /**
+     * Stops pulling updates.
+     */
+    public synchronized void stopPullingUpdates() {
+        if (executor != null) {
+            executor.shutdown();
```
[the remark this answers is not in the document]
Flaugh24: done

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * Start pulling updates from a node.
+     *
+     * @param nodeUrl Node URL.
+     */
+    public void startPullingUpdates(String nodeUrl) {
+        stopPullingUpdates();
+        synchronized (this) {
+            if (executor == null) {
+                executor = Executors.newScheduledThreadPool(1);
```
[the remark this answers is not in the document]
Flaugh24: done

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
Flaugh24: fixed

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
Flaugh24: replaced with volatile HashMap

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/OptionsConstants.java
Flaugh24: done

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/NodeUrlMixin.java
Flaugh24: done

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/converters/NodeNameOrUrlConverter.java
Flaugh24: done

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/NodeNameDynamicCompleter.java
Flaugh24: done

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/NodeNameDynamicCompleter.java
Flaugh24: fixed

modules/network/src/main/java/org/apache/ignite/network/scalecube/NodeMetadataCodec.java
Flaugh24: done

modules/network/src/test/java/org/apache/ignite/network/scalecube/MetadataSerDeTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+import io.scalecube.cluster.metadata.JdkMetadataCodec;
+import java.nio.ByteBuffer;
+import org.apache.ignite.network.NodeMetadata;
+import org.junit.jupiter.api.Test;
+
+class MetadataSerDeTest {
```
sashapolo, line 28: I think this test is a bit redundant, since we are testing a 3rd party class here

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
```diff
@@ -0,0 +1,119 @@
...
+     */
+    public Optional<URL> getNodeUrl(String nodeName) {
+        return Optional.ofNullable(nodeNameToNodeUrl.get(nodeName));
+    }
+
+    /**
+     * Gets all node names.
+     */
+    public List<String> getAllNames() {
+        return new ArrayList<>(nodeNameToNodeUrl.keySet());
```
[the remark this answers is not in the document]
Flaugh24: Not sure that it's a good idea. keySet() returns a Set view of the keys contained in this map. The set is backed by the map, so changes to the map are reflected in the set, and vice-versa.

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/Await.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.cli;
+
+import java.time.Duration;
+import java.util.function.Supplier;
+
+/** Utility class to await some condition. */
+public final class Await {
```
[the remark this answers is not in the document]
Flaugh24: removed

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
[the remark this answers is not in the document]
sashapolo: This map is now immutable, so no need to worry. If you are concerned about someone modifying the returned set, then this will be handled by the `Collectors.toImmutableMap`

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
```diff
@@ -190,6 +204,17 @@ public void beforeNodeStop() {
...
```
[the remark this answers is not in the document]
Flaugh24: added updateLocalMember()

modules/cli/src/main/java/org/apache/ignite/internal/cli/Main.java
```diff
@@ -80,7 +80,7 @@ public static void main(String[] args) {
     }
 
     private static boolean isatty() {
-        return System.console() != null;
+        return true;
```
sashapolo, line 83: What is this change for?

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
```diff
@@ -59,6 +68,12 @@ public class ScaleCubeClusterServiceFactory {
     /** Logger. */
     private static final IgniteLogger LOG = Loggers.forClass(ScaleCubeClusterServiceFactory.class);
 
+    /** Metadata codec. */
+    private static final JdkMetadataCodec METADATA_CODEC = new JdkMetadataCodec();
```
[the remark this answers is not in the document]
Flaugh24: done

modules/network/src/test/java/org/apache/ignite/network/scalecube/MetadataSerDeTest.java
Flaugh24: removed

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
```diff
@@ -59,6 +68,12 @@ public class ScaleCubeClusterServiceFactory {
...
+
+    /** Scalecube cluster executor to update metadata.  */
+    private final Scheduler scheduler = scheduler("ScaleCubeClusterServiceFactory");
```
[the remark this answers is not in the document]
Flaugh24: removed

modules/network/src/main/java/org/apache/ignite/network/scalecube/NodeMetadataDeserializer.java
```diff
@@ -0,0 +1,47 @@
...
+
+package org.apache.ignite.network.scalecube;
+
+import io.scalecube.cluster.metadata.JdkMetadataCodec;
+import java.nio.ByteBuffer;
+import org.apache.ignite.network.NodeMetadata;
+import org.jetbrains.annotations.Nullable;
+
+/** Deserializer for {@link org.apache.ignite.network.NodeMetadata}. */
+public class NodeMetadataDeserializer {
```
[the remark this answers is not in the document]
Flaugh24: removed

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @Nullable
+    private static URL urlFromClusterNode(NodeMetadata metadata) {
+        if (metadata == null) {
+            return null;
+        }
+        try {
+            return new URL("http://" + metadata.getRestHost() + ":" + metadata.getRestPort());
+        } catch (Exception e) {
+            return null;
```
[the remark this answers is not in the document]
Flaugh24: logged

modules/cli/src/main/java/org/apache/ignite/internal/cli/NodeNameRegistry.java
```diff
@@ -0,0 +1,119 @@
...
+    public List<String> getAllNames() {
+        return new ArrayList<>(nodeNameToNodeUrl.keySet());
+    }
+
+    /**
+     * Stops pulling updates.
+     */
+    public synchronized void stopPullingUpdates() {
+        if (executor != null) {
+            shutdownAndAwaitTermination(executor, 3, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
Flaugh24: accidentally removed

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/NodeNameTest.java
```diff
@@ -0,0 +1,90 @@
...
+/** Tests for ignite node commands with a provided node name. */
+public class NodeNameTest extends CliCommandTestNotInitializedIntegrationBase {
+
+    private String nodeName;
+
+    @BeforeEach
+    void setUp() throws InterruptedException {
+        nodeNameRegistry.startPullingUpdates("http://localhost:10301");
+        // wait to pulling node names
+        waitForCondition(() -> !nodeNameRegistry.getAllNames().isEmpty(), Duration.ofSeconds(5).toMillis());
```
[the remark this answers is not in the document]
Flaugh24: added

modules/network/src/integrationTest/java/org/apache/ignite/network/scalecube/ItClusterServiceTest.java
```diff
@@ -61,4 +68,59 @@ void testShutdown(TestInfo testInfo) {
 
         assertThat(e.getCause(), isA(NodeStoppingException.class));
     }
+
+    @Test
+    void testUpdateMetadata(TestInfo testInfo) throws Exception {
```
sashapolo, line 73: Formatting here is a total mess, please add some empty lines

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
```diff
@@ -241,4 +253,8 @@ private static List<Address> parseAddresses(List<NetworkAddress> addresses) {
                 .map(addr -> Address.create(addr.host(), addr.port()))
                 .collect(Collectors.toList());
     }
+
+    private Scheduler scheduler(String namePrefix) {
```
sashapolo, line 257: Looks like this method is no longer needed

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
Flaugh24: done

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
```diff
@@ -143,10 +151,8 @@ public void onMembershipEvent(MembershipEvent event) {
 
                 cluster.startAwait();
 
-                // emit an artificial event as if the local member has joined the topology (ScaleCube doesn't do that)
-                var localMembershipEvent = MembershipEvent.createAdded(cluster.member(), null, System.currentTimeMillis());
-
-                topologyService.onMembershipEvent(localMembershipEvent);
+                // update info about the local member (ScaleCube doesn't emit events about local members)
+                topologyService.updateLocalMember(cluster.member(), null);
```
sashapolo, line 155: This is not a correct change. We rely on the fact that adding a local member will trigger topology listeners (`onAppeared`), I wonder if the tests will even pass

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeTopologyService.java
```diff
@@ -104,6 +115,18 @@ void onMembershipEvent(MembershipEvent event) {
         }
     }
 
+    /**
+     * Updates info about the local member.
+     *
+     * @param member the local member.
+     * @param metadata metadata of the local member.
+     */
+    void updateLocalMember(Member member, @Nullable NodeMetadata metadata) {
```
sashapolo, line 124: This is strange that we need to pass a `member` to the `updateLocalMember` method. `TopologyService` has a link to the local member already

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeTopologyService.java
```diff
@@ -160,9 +184,25 @@ public ClusterNode getByConsistentId(String consistentId) {
...
+
+    /**
+     * Deserializes the given {@link ByteBuffer} to a {@link NodeMetadata}.
+     *
+     * @param buffer ByteBuffer to deserialize.
+     * @return NodeMetadata or null if something goes wrong.
+     */
+    private static NodeMetadata deserializeMetadata(ByteBuffer buffer) {
+        try {
+            return (NodeMetadata) METADATA_CODEC.deserialize(buffer);
```
sashapolo, line 202: `buffer` can be null, this only works because you catch the NPE below. Let's add an explicit check

modules/cli/src/main/java/org/apache/ignite/internal/cli/Main.java
Flaugh24: accidently

sashapolo (approved):

PR #1271, opened 2022-10-28
IGNITE-17965: Enable remote build cache for Gradle build on CI.

Pochatkin:
* Optimization and fix for race cache modification

https://issues.apache.org/jira/browse/IGNITE-17965


SammyVimes (approved):
LGTM

PR #937, opened 2022-07-13
IGNITE-16907 Init RAFT group with applied index that corresponds to persisted storage data.

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-16907

PR #1280, opened 2022-10-31
IGNITE-17935 Finish implementation of streaming RAFT snapshot sender

rpuch:
https://issues.apache.org/jira/browse/IGNITE-17935


modules/raft/src/main/java/org/apache/ignite/internal/raft/storage/SnapshotStorageFactory.java
```diff
@@ -28,7 +29,8 @@ public interface SnapshotStorageFactory {
      *
      * @param uri Snapshot URI.
      * @param raftOptions Raft options.
+     * @param logManager Log manager.
```
tkalkirill, line 32: Which log manager?

modules/raft/src/main/java/org/apache/ignite/internal/raft/storage/impl/IgniteJraftServiceFactory.java
```diff
@@ -38,7 +39,9 @@ public class IgniteJraftServiceFactory extends DefaultJRaftServiceFactory {
     private final LogStorageFactory logStorageFactory;
 
     /** Snapshot storage factory. */
-    private volatile SnapshotStorageFactory snapshotStorageFactory = LocalSnapshotStorage::new;
+    private volatile SnapshotStorageFactory snapshotStorageFactory = (path, raftOptions, logManager) -> {
+        return new LocalSnapshotStorage(path, raftOptions);
+    };
```
tkalkirill, line 44:
```suggestion
    private volatile SnapshotStorageFactory snapshotStorageFactory = (path, raftOptions, logManager) -> new LocalSnapshotStorage(path, raftOptions);
```

modules/raft/src/main/java/org/apache/ignite/raft/jraft/JRaftServiceFactory.java
```diff
@@ -42,9 +42,10 @@ public interface JRaftServiceFactory {
      *
      * @param uri The snapshot storage uri from {@link NodeOptions#getSnapshotUri()}
      * @param raftOptions the raft options.
+     * @param logManager the log manager.
```
tkalkirill, line 45: Which log manager?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionSnapshotStorageFactory.java
```diff
@@ -77,44 +73,34 @@ public PartitionSnapshotStorageFactory(
...
         this.incomingSnapshotsExecutor = incomingSnapshotsExecutor;
 
         // We must choose the minimum applied index for local recovery so that we don't skip the raft commands for the storage with the
         // lowest applied index and thus no data loss occurs.
-        persistedRaftIndex = Math.min(
-                partition.mvPartitionStorage().persistedIndex(),
-                partition.txStatePartitionStorage().persistedIndex()
+        lastIncludedRaftIndex = Math.min(
+                partition.mvPartitionStorage().lastAppliedIndex(),
+                partition.txStatePartitionStorage().lastAppliedIndex()
```
tkalkirill, line 87: It should be the last persistent (saved) index.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionSnapshotStorageFactory.java
```diff
@@ -77,44 +73,34 @@ public PartitionSnapshotStorageFactory(
...
         );
     }
 
     @Override
-    public PartitionSnapshotStorage createSnapshotStorage(String uri, RaftOptions raftOptions) {
-        SnapshotMeta snapshotMeta = new RaftMessagesFactory().snapshotMeta()
-                .lastIncludedIndex(persistedRaftIndex)
-                // According to the code of org.apache.ignite.raft.jraft.core.NodeImpl.bootstrap, it's "dangerous" to init term with a value
-                // greater than 1. 0 value of persisted index means that the underlying storage is empty.
-                .lastIncludedTerm(persistedRaftIndex > 0 ? 1 : 0)
```
tkalkirill, line 105: I think we need to leave the definition of the term as it was done before, there is an explanation in the comment.

As I understand it, then we do not need parameter `logManager`.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
```diff
@@ -71,18 +76,23 @@ public class OutgoingSnapshot {
 
     private final ReusableLockLockup mvOperationsLockup = new ReusableLockLockup(mvOperationsLock);
 
+    /**
+     * Snapshot metadata taken on snapshot scope freezing.
+     */
```
tkalkirill, line 81:
```suggestion
    /** Snapshot metadata taken on snapshot scope freezing. */
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
```diff
@@ -71,18 +76,23 @@ public class OutgoingSnapshot {
...
+    private SnapshotMeta meta;
```
tkalkirill, line 82: I suggest renaming it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
```diff
@@ -71,18 +76,23 @@ public class OutgoingSnapshot {
...
     private final Set<RowId> rowIdsToSkip = new ConcurrentHashSet<>();
 
-    // TODO: IGNITE-17935 - manage queue size
+    // TODO: IGNITE-18018 - manage queue size
     /**
      * Rows that need to be sent out of order (relative to the order in which this snapshot sends rows).
      * Versions inside rows are in oldest-to-newest order.
      */
-    private final Queue<SnapshotMvDataResponse.ResponseEntry> outOfOrderMvData = new LinkedList<>();
+    private final Queue<SnapshotMvDataResponse.ResponseEntry> outOfOrderMvData = new ArrayDeque<>();
```
tkalkirill, line 95: I don't see where it is protected when calling method `java.util.Queue#poll`.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
```diff
@@ -133,23 +146,57 @@ public PartitionKey partitionKey() {
...
-     * Reads a snapshot meta and returns a future with the response.
+     * Returns metadata corresponding to this snapshot.
      *
-     * @param metaRequest Meta request.
+     * @return This snapshot metadata.
      */
-    SnapshotMetaResponse handleSnapshotMetaRequest(SnapshotMetaRequest metaRequest) {
-        //TODO https://issues.apache.org/jira/browse/IGNITE-17935
+    public SnapshotMeta meta() {
+        assert meta != null : "No snapshot meta yet, probably the snapshot scope was not yet frozen";
```
tkalkirill, line 176: Can't there be a race?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
```diff
@@ -133,23 +146,57 @@ public PartitionKey partitionKey() {
...
+     *
+     * @param request Meta request.
+     */
+    @Nullable
+    SnapshotMetaResponse handleSnapshotMetaRequest(SnapshotMetaRequest request) {
+        if (closed) {
+            return logAlreadyClosedAndReturnNull();
+        }
+
+        assert meta != null : "No snapshot meta yet, probably the snapshot scope was not yet frozen";
```
tkalkirill, line 192: Can't there be a race?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
```diff
@@ -133,23 +146,57 @@ public PartitionKey partitionKey() {
...
+
+        return MESSAGES_FACTORY.snapshotMetaResponse().meta(meta).build();
+    }
+
+    @Nullable
+    private <T> T logAlreadyClosedAndReturnNull() {
```
tkalkirill, line 198: Maybe **Void** ?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/PartitionsSnapshots.java
```diff
@@ -30,4 +31,19 @@ public interface PartitionsSnapshots {
...
      */
     PartitionSnapshots partitionSnapshots(PartitionKey partitionKey);
+
+    /**
+     * Removes the underlying collection for snapshots of this partition.
+     *
+     * @param partitionKey Partition key.
+     */
+    void removeSnapshots(PartitionKey partitionKey);
+
```
tkalkirill, line 41:
```suggestion
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/SnapshotAwarePartitionDataStorage.java
```diff
@@ -123,11 +127,21 @@ private void handleSnapshotInterference(RowId rowId) {
...
-        // TODO: IGNITE-17935 - terminate all snapshots of this partition considering correct locking to do it consistently
+        cleanupSnapshots();
 
         partitionStorage.close();
     }
 
+    private void cleanupSnapshots() {
+        PartitionSnapshots partitionSnapshots = getPartitionSnapshots();
+
+        try (AutoLockup ignored = partitionSnapshots.acquireReadLock()) {
```
tkalkirill, line 138: Let's change the name **ignored**.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/SnapshotMetaUtils.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        return metaBuilder.build();
+    }
+
+    private static Collection<String> peersToStrings(List<PeerId> peers) {
+        return peers.stream().map(PeerId::toString).collect(toList());
+    }
+
+    private SnapshotMetaUtils() {
```
tkalkirill, line 64: It's not needed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionSnapshotStorageFactory.java
alievmirza: See this thread with more details https://github.com/apache/ignite-3/pull/937#discussion_r927744986

modules/raft/src/main/java/org/apache/ignite/internal/raft/storage/impl/IgniteJraftServiceFactory.java
rpuch: This way the line will exceed 140 characters in length

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionSnapshotStorageFactory.java
rpuch: It this moment they are equal, aren't they? I think we discussed this with  @ibessonov and decided to avoid the usage of `persistedIndex()`, but I might be mistaken.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionSnapshotStorageFactory.java
rpuch: CC @ibessonov

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
rpuch: How? `snapshotMeta`?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
rpuch: `try (AutoLockup ignored = acquireMvLock()) {` in `handleSnapshotMvDataRequest()`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
rpuch: No, the idea is to return `null` of a needed type as this method is called from different places

modules/raft/src/main/java/org/apache/ignite/internal/raft/storage/impl/IgniteJraftServiceFactory.java
tkalkirill: You can add line breaks.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionSnapshotStorageFactory.java
tkalkirill: @ibessonov is it true?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
tkalkirill: To make it clear that this field will be set on freeze.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
tkalkirill: Please add assertion

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/SnapshotAwarePartitionDataStorage.java
tkalkirill: Maybe `readLock` ?

modules/raft/src/main/java/org/apache/ignite/internal/raft/storage/impl/IgniteJraftServiceFactory.java
rpuch: Well, I kinda did :)

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionSnapshotStorageFactory.java
rpuch: Please also note that this is a **startup** meta. We don't actually restore anything here; we just say 'ok, we did restore it' and provide a meta (index+term) which JRaft will put to its internal fields. Before the restart the node was at index+term <I, T>, and if we return <I, T>, it will be at <I, T> after restart. Seems reasonable.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/SnapshotAwarePartitionDataStorage.java
rpuch: Renamed it to `ignoredLockup` (because it is a lockup and not a lock, and because it is ignored; naming it with 'ignored' in the beginning makes IDEA stop nagging about an 'unused variable')

modules/raft/src/main/java/org/apache/ignite/internal/raft/storage/impl/IgniteJraftServiceFactory.java
tkalkirill: Not quite.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
tkalkirill: Maybe **metaAfterFreezee** ?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
tkalkirill: I think it's more correct to mark the field as `@Nullable`, indicate when it can be `null`, and use a local variable when reading the field.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
tkalkirill: Posted above.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
tkalkirill: Then we need to check internally that we called the method when the snapshot was closed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/SnapshotAwarePartitionDataStorage.java
tkalkirill: I don’t understand a bit, the variable is called “not used”, but in fact it is used, it does not make it clear, reading the code I think if it is not used then the code is not needed and would have been deleted.

rpuch: The **object** is used by try-with-resources machinery, but the **variable** is not used inside `try` block. 

Try to write something like the following in IDEA:

```
try (var v = new FileOutputStream(...)) {
    ... some code
}
```

IDEA will immediately nag that `v` is not used. I want to avoid this behavior.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
rpuch: Renamed it to `frozenMeta`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/SnapshotAwarePartitionDataStorage.java
tkalkirill: In the above code, the variable is really not used, and we can say that I can safely remove this block.
In your case, the variable is not used and the block is painful to remove.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
```diff
@@ -63,26 +64,29 @@ public class OutgoingSnapshot {
...
+
     /**
      * Lock that is used for mutual exclusion of MV snapshot reading (by this class) and threads that write MV data to the same
      * partition (currently, via {@link SnapshotAwarePartitionDataStorage}).
      */
     private final ReentrantLock mvOperationsLock = new ReentrantLock();
 
-    private final ReusableLockLockup mvOperationsLockup = new ReusableLockLockup(mvOperationsLock);
+    @Nullable
+    private volatile SnapshotMeta frozenMeta;
```
tkalkirill, line 76: Missing javadoc and explanation why may be `null`.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
```diff
@@ -133,23 +140,61 @@ public PartitionKey partitionKey() {
...
+
+        SnapshotMeta meta = frozenMeta;
+
+        assert meta != null : "No snapshot meta yet, probably the snapshot scope was not yet frozen";
+
+        return MESSAGES_FACTORY.snapshotMetaResponse().meta(meta).build();
+    }
+
+    @Nullable
+    private <T> T logAlreadyClosedAndReturnNull() {
```
tkalkirill, line 196: Missing the assertion that the method was called by `closed`.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/SnapshotMetaUtils.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Utils to build {@link SnapshotMeta} instances.
+ */
+public class SnapshotMetaUtils {
+    /**
+     * Builds a {@link SnapshotMeta} corresponding to RAFT state (term, configuration) at the given log index.
+     *
+     * @param logIndex RAFT log index.
+     * @param logManager LogManager from which to load term and configuration.
```
tkalkirill, line 39: Which LogManager?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
```diff
@@ -106,12 +110,15 @@ public class OutgoingSnapshot {
      */
     private boolean finishedTxData = false;
 
+    private boolean closed = false;
```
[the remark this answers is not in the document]
rpuch: According to @ibessonov, there will be memory barriers when handling requests, so it is not required to be `volatile`.

Also, this is out of scope of this PR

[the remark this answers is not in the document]
rpuch: No, I'm wrong it was about a different field, and this change is in the scope

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
rpuch: I don't understand what you suggest here. Please elaborate.

tkalkirill: What should the method do?

rpuch: It should log that someone is trying to get data from an already closed snapshot and then return `null`; callers handle that `null` gracefully.

tkalkirill: So let's check that this method was called after knowing that we closed the snapshot.

rpuch: This does not seem necessary. The method is always closed under `if (closed)`

sergey-chugunov-1985: @tkalkirill , in this case I side with Roman, assertion here is excessive.
Two simple arguments from me:

1. This method is merely a utility method to save a line of code in several places, it doesn't do something that may put stability of the system at risk.
2. The method is private and context of its use is completely enclosed inside a rather short class. Assertions are a special tool to be used in rare situations where several properties or states of different components are supposed to have specific values. If we decide to put assertions in all methods like this one we'll end up with code where ```assert``` is the most used keyword in every class.

What do you think?

tkalkirill: My opinion:
1. `logAlreadyClosedAndReturnNull` reads "log already closed return `null`", "log already closed" sounds to me like logging is closed, "return `null`" sounds clear.
2. Yes, the method is private, but suddenly another developer does not use it correctly, otherwise he will be able to see in the log that he made a mistake and he will have to debug less.
3. For all methods it can be redundant but not terrible.

rpuch: I renamed the method to make its name less ambiguous

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
```diff
@@ -133,39 +141,84 @@ public PartitionKey partitionKey() {
...
-     * Reads a snapshot meta and returns a future with the response.
+     * Reads the snapshot meta and returns a response. Returns {@code null} if the snapshot is already closed.
      *
-     * @param metaRequest Meta request.
+     * @param request Meta request.
      */
-    SnapshotMetaResponse handleSnapshotMetaRequest(SnapshotMetaRequest metaRequest) {
-        //TODO https://issues.apache.org/jira/browse/IGNITE-17935
+    @Nullable
+    SnapshotMetaResponse handleSnapshotMetaRequest(SnapshotMetaRequest request) {
```
tkalkirill, line 184: I see that the argument is not used, we need to either remove it or check that we requested the meta for this particular snapshot.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
```diff
@@ -133,39 +141,84 @@ public PartitionKey partitionKey() {
...
+
+        SnapshotMeta meta = frozenMeta;
+
+        assert meta != null : "No snapshot meta yet, probably the snapshot scope was not yet frozen";
+
+        return MESSAGES_FACTORY.snapshotMetaResponse().meta(meta).build();
+    }
+
+    @Nullable
+    private <T> T logThatAlreadyClosedAndReturnNull() {
```
tkalkirill, line 197: What will happen to the receiver if we return `null`? will he hang?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
tkalkirill: How often can it be that we should return `null`?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
rpuch: If we return `null`, then `OutgoingSnapshotsManager#handleSnapshotRequestMessage()` will return `null`, so we'll return no response to the receiver. I guess timeout will trigger on the receiver.

Actually, I would prefer to add an ability to return an error in all 3 response types, or even change our invocation mechanism to automatically pass exceptions from the responder to the invoker and throw them on the invoker (wrapped in something like `RemoteException`), but @ibessonov suggested to just return nothing in such cases for simplicity.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
rpuch: Added an assertion

PR #1282, opened 2022-10-31
IGNITE-17899 Properly stopping resources in TableManager.

sergeyuttsel:
https://issues.apache.org/jira/browse/IGNITE-17899


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -969,18 +971,28 @@ public void stop() {
...
+                    LOG.info("Unable to stop the replica [tableName={}, partition={}]", e, table.name(), p);
                 }
+            }
 
-                table.internalTable().storage().stop();
-                table.internalTable().txStateStorage().stop();
-                table.internalTable().close();
+            try {
+                IgniteUtils.closeAll(
+                        Stream.of(table.internalTable().storage(),
```
sanpwc, line 992: It's better to skip stream part. Just use
```
                IgniteUtils.closeAll(
                        table.internalTable().storage(),
                        table.internalTable().txStateStorage(),
                        table.internalTable()
                );
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -969,18 +971,28 @@ public void stop() {
...
-            try {
-                for (int p = 0; p < table.internalTable().partitions(); p++) {
-                    TablePartitionId replicationGroupId = new TablePartitionId(table.tableId(), p);
+            for (int p = 0; p < table.internalTable().partitions(); p++) {
+                TablePartitionId replicationGroupId = new TablePartitionId(table.tableId(), p);
 
+                try {
                     raftMgr.stopRaftGroup(replicationGroupId);
+                } catch (Exception e) {
+                    LOG.info("Unable to stop the raft group [tableName={}, partition={}]", e, table.name(), p);
```
sanpwc, line 980: I don't think that we should log an exception if NodeStoppingException is thrown. I would favor cleanUpTablesResources to be an idempotent action and it's a part of it.

sanpwc, line 980: I would favor to log only one record "Unable to stop table [name={} id={}]" with partition specific details in suppress exceptions instead of spamming to the log.
And in order to hide all that suppress related logic to some inner method let's consider adding 
```
    public static void runAll(Stream<? extends Runnable> runnables) throws Exception {
        AtomicReference<Exception> ex = new AtomicReference<>();

        runnables.filter(Objects::nonNull).forEach(runnable -> {
            try {
                runnable.run();
            } catch (Exception e) {
                if (!ex.compareAndSet(null, e)) {
                    ex.get().addSuppressed(e);
                }
            }
        });

        if (ex.get() != null) {
            throw ex.get();
        }
    } 
```
along with combining two streams for 
`raftMgr.stopRaftGroup(replicationGroupId);` and `replicaMgr.stopReplica(replicationGroupId);`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -969,18 +971,28 @@ public void stop() {
...
-                table.internalTable().txStateStorage().stop();
-                table.internalTable().close();
+            try {
+                IgniteUtils.closeAll(
+                        Stream.of(table.internalTable().storage(),
+                                table.internalTable().txStateStorage(),
+                                table.internalTable())
+                );
             } catch (Exception e) {
                 LOG.info("Unable to stop table [name={}]", e, table.name());
```
sanpwc, line 997: Let's also add table id to the log.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -969,18 +971,28 @@ public void stop() {
      */
     private void cleanUpTablesResources(Map<UUID, TableImpl> tables) {
```
sanpwc, line 972: Tests are missing.

sanpwc (changes requested):

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/engine/MvTableStorage.java
```diff
@@ -36,7 +36,7 @@
 /**
  * Table storage that contains meta, partitions and SQL indexes.
  */
-public interface MvTableStorage {
+public interface MvTableStorage extends AutoCloseable {
```
tkalkirill, line 39: Can method `close()` be defined explicitly?
This will get rid of warnings on the idea that we don't close the resource after it's been used.
@sanpwc wdyt

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sergeyuttsel: Fixed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sergeyuttsel: Fixed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sergeyuttsel: Fixed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sergeyuttsel: Fixed.

sergeyuttsel: Fixed

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/engine/MvTableStorage.java
sergeyuttsel: Fixed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -978,20 +980,56 @@ public void stop() {
...
         for (TableImpl table : tables.values()) {
-            try {
-                for (int p = 0; p < table.internalTable().partitions(); p++) {
-                    TablePartitionId replicationGroupId = new TablePartitionId(table.tableId(), p);
+            List<Runnable> stopping = new ArrayList<>();
+
+            AtomicReference<Exception> exception = new AtomicReference<>();
```
sanpwc, line 985: What's the point of introducing org.apache.ignite.internal.util.IgniteUtils#runAll if you have exact the same logic at this level?

sanpwc (changes requested):

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
```diff
@@ -449,6 +452,8 @@ public void tableManagerStopTest() throws Exception {
 
         TableManager tableManager = tblManagerFut.join();
 
+        doThrow(new NodeStoppingException()).when(rm).stopRaftGroup(any());
```
sanpwc, line 455: - And what about checking exception handling logic inside org.apache.ignite.internal.util.IgniteUtils#runAll?
- And what about checking exception handling logic during storage stop?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sergeyuttsel: I removed IgniteUtils#runAll.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
sergeyuttsel: I removed IgniteUtils#runAll.
Added new tests.

sanpwc (approved):

PR #1283, opened 2022-10-31
IGNITE-18012 Add connect command to start CLI in REPL mode

valepakh:
https://issues.apache.org/jira/browse/IGNITE-18012

### DIFF
--- a/modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/connect/ConnectReplCommand.java
+++ b/modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/connect/ConnectReplCommand.java
@@ -21,7 +21,6 @@ import static org.apache.ignite.internal.cli.commands.OptionsConstants.CLUSTER_U
 import static org.apache.ignite.internal.cli.commands.OptionsConstants.NODE_URL_OR_NAME_DESC;
 
 import jakarta.inject.Inject;
-import org.apache.ignite.internal.cli.ReplManager;
 import org.apache.ignite.internal.cli.call.connect.ConnectCall;
 import org.apache.ignite.internal.cli.call.connect.ConnectCallInput;
 import org.apache.ignite.internal.cli.commands.BaseCommand;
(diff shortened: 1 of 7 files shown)


modules/cli/src/main/java/org/apache/ignite/internal/cli/ReplFactory.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.ignite.internal.cli.core.repl.Repl;
+import org.apache.ignite.internal.cli.core.repl.SessionDefaultValueProvider;
+import org.apache.ignite.internal.cli.core.repl.executor.ReplExecutorProvider;
+import org.apache.ignite.internal.cli.core.repl.prompt.PromptProvider;
+
+/**
+ * Class which runs main REPL mode, it's used both when starting directly into the REPL mode and from the `connect` command.
+ */
+@Singleton
+public class ReplFactory {
```
PakhomovAlexander, line 36: `ReplFactory` is a little confusing here because it does not create any instance. It could enter the interactive mode. I think `Repl.startReplMode()` might be better.

modules/cli/src/main/java/org/apache/ignite/internal/cli/ReplFactory.java
```diff
@@ -0,0 +1,70 @@
...
+    private SessionDefaultValueProvider defaultValueProvider;
+
+    @Inject
+    private ConnectToClusterQuestion question;
+
+    /**
+     * Enters REPL mode.
+     */
+    public void enterRepl() {
+        replExecutorProvider.get().execute(Repl.builder()
```
PakhomovAlexander, line 53: What about entering repl more than once? Would it work? Shall we check if we are already in repl mode?

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/connect/ConnectReplCommand.java
```diff
@@ -0,0 +1,55 @@
...
+    @Parameters(description = NODE_URL_DESC, descriptionKey = CLUSTER_URL_KEY, converter = UrlConverter.class)
+    private URL nodeUrl;
+
+    @Inject
+    private ConnectCall connectCall;
+
+    /** {@inheritDoc} */
+    @Override
+    public void run() {
+        CallExecutionPipeline.builder(connectCall)
```
PakhomovAlexander, line 48: If there is the same node that we are connected to now we could display something like "already connected to <nodename>".

modules/cli/src/main/java/org/apache/ignite/internal/cli/ReplFactory.java
valepakh: `ReplExecutor.execute` is a thread blocking method so I don't think it's possible to do that, also SQL REPL mode works just fine entering REPL from REPL.

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/connect/ConnectReplCommand.java
valepakh: Probably yes, but I think it's outside the scope of this ticket.

PR #1291, opened 2022-11-01
IGNITE-18032 Do not implement AutoCloseable on classes instances of which are not meant to be used in try-with-resources

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18032


modules/core/src/main/java/org/apache/ignite/internal/ManuallyCloseable.java
```diff
@@ -0,0 +1,39 @@
...
+ * and issue a warning, so it becomes a drag to work with code that uses AutoCloseable for classes which instances
+ * are used with patterns different from the try-with-resources pattern.
+ *
+ * <p>The main reason of this interface appearance was the desire to mark 'must-be-eventually-closed' types so that
+ * we don't forget closing their instances.
+ *
+ * <p>Subinterfaces and classes implementing this interface should declare a narrower {@code throws} declaration (either
+ * having concrete subclasses of {@link Exception} or declaring no exceptions at all).
+ */
+public interface ManuallyCloseable {
```
tkalkirill, line 32: Do I understand correctly that this interface is similar to a marker one? It seems to me that there is no need for it.

rpuch: @sanpwc pointed out that `AutoCloseable` provides us with a hint that 'this particular instance needs to be eventually closed'. As we want to remove `AutoCloseable` from long-lived objects, it seems nice to have a weaker replacement. This is not too dramatic, it's like a marker or a comment, you are right, but it costs us almost nothing to declare it on a type where we already want to have `close()` method.

tkalkirill: Another argument against such an interface: why there is an **ManuallyCloseable** and no **ManuallyStartable** and others like them. They will introduce redundancy into development.

modules/core/src/main/java/org/apache/ignite/internal/close/ManuallyCloseable.java
```diff
@@ -0,0 +1,39 @@
...
+package org.apache.ignite.internal.close;
+
+/**
+ * Represents something that must be eventually closed. It is different from {@link AutoCloseable} which is for being
+ * used in try-with-resources; IDEs treat any usage of an AutoCloseable outside of a try-with-resources block as a suspicious
+ * and issue a warning, so it becomes a drag to work with code that uses AutoCloseable for classes which instances
+ * are used with patterns different from the try-with-resources pattern.
+ *
+ * <p>The main reason of this interface appearance was the desire to mark 'must-be-eventually-closed' types so that
+ * we don't forget closing their instances.
```
SammyVimes, line 27: How does it help developers not to forget closing?

modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
```diff
@@ -226,10 +227,10 @@ private void destroyRocksDb() throws RocksDBException {
 
     /** {@inheritDoc} */
     @Override
-    public void close() throws Exception {
+    public void close() {
         IgniteUtils.shutdownAndAwaitTermination(snapshotExecutor, 10, TimeUnit.SECONDS);
 
-        IgniteUtils.closeAll(db, options);
+        RocksUtils.closeAll(db, options);
```
SammyVimes, line 233: Can't it throw an exception actually?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
```diff
@@ -227,7 +227,7 @@ public void stop() throws Exception {
             IgniteUtils.closeAll(
```
SammyVimes, line 227: probably this should be called `runAll` then

modules/raft/src/test/java/org/apache/ignite/raft/jraft/storage/impl/BaseLogStorageTest.java
```diff
@@ -16,6 +16,12 @@
  */
 package org.apache.ignite.raft.jraft.storage.impl;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
SammyVimes, line 19: Are you sure that imports should be here in jraft?

modules/rocksdb-common/build.gradle
```diff
@@ -23,6 +23,10 @@ dependencies {
     implementation project(':ignite-core')
     implementation libs.jetbrains.annotations
     api libs.rocksdb.jni
+
+    testImplementation libs.hamcrest.core
+    testImplementation libs.hamcrest.core
```
SammyVimes, line 28: this line is the same as above + in pom.xml you've added 4 dependencies and here only 2 (or 3 with the duplicate)

modules/transactions/src/test/java/org/apache/ignite/internal/tx/storage/state/TxStateStorageAbstractTest.java
```diff
@@ -150,10 +167,14 @@ public void testCas() throws Exception {
 
     @Test
     public void testScan() throws Exception {
-        try (TxStateTableStorage tableStorage = createStorage()) {
-            TxStateStorage storage0 = tableStorage.getOrCreateTxStateStorage(0);
-            TxStateStorage storage1 = tableStorage.getOrCreateTxStateStorage(1);
-            TxStateStorage storage2 = tableStorage.getOrCreateTxStateStorage(2);
+        try (
+                var storageHolder0 = new AutoCloser<>(tableStorage.getOrCreateTxStateStorage(0), TxStateStorage::close);
```
SammyVimes, line 171: Funny thing, you actually write more code than there were or even would be if there was no `AutoCloser`:

```
var storageHolder0 = tableStorage.getOrCreateTxStateStorage(0); 
storageHolder0.close();
```

And that's without calling `.value()` on all this holders...

SammyVimes (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/ManuallyCloseable.java
rpuch: We don't need `ManuallyStartable` because you cannot forget about starting a component (you will see an error instantly when trying to use an unstarted component).

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/ClusterStateStorage.java
```diff
@@ -104,6 +105,12 @@ public interface ClusterStateStorage extends AutoCloseable {
      */
     void restoreSnapshot(Path snapshotPath);
 
+    /**
+     * Closes this storage.
+     */
+    @Override
+    void close();
```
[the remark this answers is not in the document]
rpuch: On `ManuallyCloseable`, it is declared to throw `Exception`. This one does not declare anything, so re-declaring it here has value.

modules/core/src/main/java/org/apache/ignite/internal/close/ManuallyCloseable.java
rpuch: Only as an 'annotation'. If you are unsure about whether you must close something, you might inspect its interfaces. If it is a `ManuallyCloseable`, then you must close it.

Actually, I'm not sure this is actually needed, this was suggested by @sanpwc
Sasha, could you please step in? Do we actually need it?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
rpuch: Why? It closes all `Closeables` given to it. Those are ot `Runnable`s.

modules/raft/src/test/java/org/apache/ignite/raft/jraft/storage/impl/BaseLogStorageTest.java
rpuch: Why not?

modules/rocksdb-common/build.gradle
rpuch: I added only 2 dependencies in build.gradle because other 2 dependencies seem to be already there. They are added by

```
apply from: "$rootDir/buildscripts/java-junit5.gradle"
```

probably. At least, the project compiles (including tests) via Gradle.

I removed the duplicate, thanks

modules/transactions/src/test/java/org/apache/ignite/internal/tx/storage/state/TxStateStorageAbstractTest.java
rpuch: On the other hand, the code with try-with-resources is safer, it closes the storage.

Do you think it's not important in tests and should be replaced with simply calling `close()` in the end?

modules/raft/src/test/java/org/apache/ignite/raft/jraft/storage/impl/BaseLogStorageTest.java
SammyVimes: I mean, we are trying to change as little as we can in jraft sources and this looks like an accidental import reformat

modules/transactions/src/test/java/org/apache/ignite/internal/tx/storage/state/TxStateStorageAbstractTest.java
SammyVimes: It's hard to tell, but I think `try-with-resources` is used for the clarity of the code, not for security (because you have to close resources either way). And with AutoCloser we have to write more code with no profit. It's all my opinion of course

rpuch: `try-with-resources` provides some safeness guarantees, you cannot forget to close a resource used with `try-with-resources`, but you can if you don't use this idiom.

On the other hand, in tests it's not that critical to leak a resource, so I removed `try-with-resources` altogether.

modules/raft/src/test/java/org/apache/ignite/raft/jraft/storage/impl/BaseLogStorageTest.java
rpuch: Oh, thanks, good catch. I had no idea JRaft code has a different order of imports.

SammyVimes (approved):
LGTM but let's wait for other reviewers

modules/core/src/main/java/org/apache/ignite/internal/close/AutoCloser.java
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * A utility that allows to use something that is not an {@link AutoCloseable} in try-with-resources blocks.
+ *
+ * @param <T> Type of the resource.
+ * @param <X> Exception that can be thrown by the closer during closing. This type parameter is needed to make sure
+ *     that {@link AutoCloser#close()} throws declaration is as narrow as possible.
+ */
+@SuppressWarnings("CheckedExceptionClass")
+public class AutoCloser<T, X extends Exception> implements AutoCloseable {
```
sanpwc, line 28: Well, to be honest all that stuff seems as an over design to me.

modules/metastorage-server/src/main/java/org/apache/ignite/internal/metastorage/server/raft/MetaStorageListener.java
```diff
@@ -418,7 +418,7 @@ public boolean onSnapshotLoad(Path path) {
     public void onShutdown() {
         try {
             storage.close();
-        } catch (Exception e) {
+        } catch (RuntimeException e) {
```
sanpwc, line 421: Well, that's actually very confusing. Almost no one will handle RuntimeException.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/RocksDbClusterStateStorage.java
```diff
@@ -209,10 +209,10 @@ public void destroy() {
     }
 
     @Override
-    public void close() throws Exception {
+    public void close() {
         IgniteUtils.shutdownAndAwaitTermination(snapshotExecutor, 10, TimeUnit.SECONDS);
 
-        IgniteUtils.closeAll(options, db);
+        RocksUtils.closeAll(options, db);
```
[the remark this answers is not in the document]
rpuch: The idea of this method is to have ability to close RocksDB references in such a way that no `throws` needs to be declared on the `closeAll()` (because `AbstractNativeReference`, which implements `AutoCloseable`, does not declare any `throws`). On the other hand, a method that accepts `AutoCloseable`s (like `IgniteUtils.closeAll()`) or `ManualCloseable` (that you imagined) would have to declare `throws Exception` which is what I tried to avoid for RocksDB resources. Hence this method.

modules/core/src/main/java/org/apache/ignite/internal/close/AutoCloser.java
rpuch: Removed the stuff

sanpwc (approved):
Looks good to me.

PR #1292, opened 2022-11-01
IGNITE-17478 CLI commands with argument groups reuse previous option values

valepakh:
https://issues.apache.org/jira/browse/IGNITE-17478

Underlying picocli issue is fixed so we need to bump the version.


modules/cli/src/test/java/org/apache/ignite/internal/cli/commands/CommandLineReuseTest.java
```diff
@@ -0,0 +1,68 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+import org.junit.jupiter.api.Test;
+import picocli.CommandLine;
+import picocli.CommandLine.ArgGroup;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
+
+class CommandLineReuseTest {
```
kgusakov, line 29: Not sure, that we should copy/implement tests for fixes of external libraries :) Anyway, if you are sure, that we should, I would like  to have a small comment with a link to the original picocli issue.

I think a version bump + commit message with our ticket number is enough for this PR.

valepakh: Agreed, removed the test.

kgusakov (approved):

PR #1305, opened 2022-11-03
IGNITE-18073 Add indexes to API for full rebalance without losing user data in MvPartitionStorage on receiver

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18073


modules/storage-api/src/main/java/org/apache/ignite/internal/storage/engine/MvTableStorage.java
```diff
@@ -164,43 +166,47 @@ default IndexStorage getOrCreateIndex(int partitionId, UUID indexId) {
...
-     * <p>This method must be called before every full rebalance of the partition storage, so that in case of errors or cancellation of the
-     * full rebalance, we can restore the partition storage from the backup.
+     * <p>This method must be called before every full rebalance, so that in case of errors or cancellation of the full rebalance, we can
+     * restore the partition storage and its index storages from the backup.
      *
      * <p>Full rebalance will be completed when one of the methods is called:
      * <ol>
      *     <li>{@link #abortRebalanceMvPartition(int)} - in case of a full rebalance cancellation or failure, so that we can
-     *     restore the partition storage from a backup;</li>
+     *     restore thepartition storage and its index storages from a backup;</li>
```
ibessonov, line 178:
```suggestion
     *     restore the partition storage and its index storages from a backup;</li>
```

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/engine/MvTableStorage.java
```diff
@@ -116,16 +116,18 @@ default IndexStorage getOrCreateIndex(int partitionId, UUID indexId) {
...
      * Destroys the index under the given name and all data in it.
      *
      * <p>This method is a no-op if the index under the given name does not exist.
      *
+     * @param partitionId Partition ID.
      * @param indexId Index ID.
+     * @throws StorageException If the given partition does not exist, or fail destroying the index.
      */
-    CompletableFuture<Void> destroyIndex(UUID indexId);
+    CompletableFuture<Void> destroyIndex(int partitionId, UUID indexId) throws StorageException;
```
ibessonov, line 130: Why did you add a partition id? What does it mean? I don't understand it.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
```diff
@@ -311,51 +313,89 @@ public void testMisconfiguredIndices() {
...
-                    clock.now()
-            );
+            partitionStorage.addWriteCommitted(rowId, binaryRow, clock.now());
 
             partitionStorage.lastAppliedIndex(100);
 
             return null;
         });
 
+        hashIndexStorage.put(indexRow);
```
ibessonov, line 340: Why are you doing it without "runConsistently"? All writes should be inside of the closure.
And index updates should be in the same closure as writes, aren't they? It's irrelevant for this particular test, but as a general rule it's fine, why violating it?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
```diff
@@ -311,51 +313,89 @@ public void testMisconfiguredIndices() {
...
         MvPartitionStorage newPartitionStorage0 = tableStorage.getMvPartition(PARTITION_ID);
+        HashIndexStorage newHashIndexStorage0 = tableStorage.getOrCreateHashIndex(PARTITION_ID, hashIndexId);
+        SortedIndexStorage newSortedIndexStorage0 = tableStorage.getOrCreateSortedIndex(PARTITION_ID, sortedIndexId);
 
         assertNotNull(newPartitionStorage0);
+        assertNotNull(newHashIndexStorage0);
+        assertNotNull(newSortedIndexStorage0);
+
         assertNotSame(partitionStorage, newPartitionStorage0);
+        assertNotSame(hashIndexStorage, newHashIndexStorage0);
```
ibessonov, line 357: Why do you assert things like these? They look completely wrong. What if we have a partition listener that has a cached value of MvPartitionStorage? Should it invalidate it and ask for another one?
Old instances must work. Please fix the test and the test implementation

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvTableStorage.java
```diff
@@ -40,50 +37,27 @@
...
+     * Storages container for the partition.
      */
-    private static class SortedIndices {
-        private final SortedIndexDescriptor descriptor;
-
-        final Map<Integer, SortedIndexStorage> storageByPartitionId = new ConcurrentHashMap<>();
-
-        SortedIndices(SortedIndexDescriptor descriptor) {
-            this.descriptor = descriptor;
+    private static class Storages {
```
ibessonov, line 51: Would it be better to store indexes in the "TestMvPartitionStorage" instance? What's the ultimate purpose of a separate class?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvTableStorage.java
```diff
@@ -93,62 +67,59 @@ public TestMvTableStorage(TableConfiguration tableCfg, TablesConfiguration table
...
-            throw new StorageException("Partition ID " + partitionId + " does not exist");
-        }
+    public TestSortedIndexStorage getOrCreateSortedIndex(int partitionId, UUID indexId) throws StorageException {
+        Storages storages = storagesByPartitionId.get(partitionId);
 
-        SortedIndices sortedIndices = sortedIndicesById.computeIfAbsent(
-                indexId,
-                id -> new SortedIndices(new SortedIndexDescriptor(id, tablesCfg.value()))
-        );
+        checkPartitionStoragesExists(storages, partitionId);
```
ibessonov, line 93: Should we lazily create a partition? What does the documentation say? I believe that some of storages have different behavior, I don't like that

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvTableStorage.java
```diff
@@ -180,33 +151,51 @@ public void destroy() throws StorageException {
...
+
+        checkPartitionStoragesExists(oldStorages, partitionId);
 
-        assert oldPartitionStorage != null : "Partition does not exist: " + partitionId;
+        if (backupStoragesByPartitionId.putIfAbsent(partitionId, oldStorages) == null) {
+            Storages newStorages = new Storages(new TestMvPartitionStorage(partitionId));
 
-        if (backupPartitions.putIfAbsent(partitionId, oldPartitionStorage) == null) {
-            partitions.put(partitionId, new TestMvPartitionStorage(partitionId));
+            oldStorages.hashIndexes.keySet().forEach(indexId -> newStorages.hashIndexes.put(
```
ibessonov, line 162: Creating indexes is not mandatory, as far as I see. Rebalance itself will create indexes that it needs

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/engine/MvTableStorage.java
tkalkirill: Fix it

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
tkalkirill: Technically it can do this, but you're right, let it be inside `runConsistently`.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
tkalkirill: Here we check that all storages are new and empty.
Why is this wrong?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvTableStorage.java
tkalkirill: It's just clearer and we change the existing code less, if you insist, I can move it to `TestMvPartitionStorage`.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvTableStorage.java
tkalkirill: Please see the documentation for the base method. =)
I did not change it and kept the logic.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvTableStorage.java
tkalkirill: I suggest discussing this in person.

tkalkirill: We decided personally that we would recreate them.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/engine/MvTableStorage.java
tkalkirill: Fix it

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
tkalkirill: Fix it

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvTableStorage.java
tkalkirill: Discussed to do it through delegates so that there are less problems when we change storages. 
Therefore, move `TestHashIndexStorage`, `TestSortedIndexStorage` to `TestMvPartitionStorage`.

tkalkirill: Fix it

tkalkirill: Tried to chage it.

PR #1313, opened 2022-11-03
IGNITE-18022 Creating an API for full rebalance without losing user data in TxStateStorage on receiver

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18022


modules/transactions/src/testFixtures/java/org/apache/ignite/internal/tx/storage/state/AbstractTxStateTableStorageTest.java
```diff
@@ -0,0 +1,170 @@
...
+        assertNotNull(newStorage0);
+
+        assertSame(storage, newStorage0);
+
+        assertEquals(0L, newStorage0.lastAppliedIndex());
+        assertEquals(0L, newStorage0.persistedIndex());
+
+        assertThat(getAll(newStorage0.scan()), empty());
+
+        tableStorage.startRebalance(PARTITION_ID).get(1, TimeUnit.SECONDS);
```
ibessonov, line 90: Can we start rebalance 2 times? Looks kinda weird, I'd prohibit that

modules/transactions/src/testFixtures/java/org/apache/ignite/internal/tx/storage/state/AbstractTxStateTableStorageTest.java
```diff
@@ -0,0 +1,170 @@
...
+        TxStateStorage newStorage0 = tableStorage.getTxStateStorage(PARTITION_ID);
+
+        assertNotNull(newStorage0);
+
+        assertSame(storage, newStorage0);
+
+        assertEquals(0L, newStorage0.lastAppliedIndex());
+        assertEquals(0L, newStorage0.persistedIndex());
+
+        assertThat(getAll(newStorage0.scan()), empty());
```
ibessonov, line 88: What happens with old opened cursors? I have a bad feeling that we didn't think through some corner cases. It's easy to see that these scans will work in all storages except for persistent page memory (to some extent), but there's a moment before the complete invalidation when we will have to forcefully close cursors. To a lesser degree it's also applied to regular reads.
Did you have any thoughts about the ways that storage "clients" will operate with them? Should we sit down and think some more? I'm not confident right now that we fully realize what we're doing and a potential impacts of it. Maybe backup should be fully readable and we only swap it at the end, maybe closing cursors or transparently reopening them (when it comes to RO transactions, it should be achievable).
All of that is mainly about MV storage. Here it's not that problematic, RocksDB allows to have some pretty neat workarounds, right? We should still check everything though.
Please think of everything I said and we'll discuss it later.

modules/transactions/src/testFixtures/java/org/apache/ignite/internal/tx/storage/state/AbstractTxStateTableStorageTest.java
```diff
@@ -0,0 +1,170 @@
...
+        assertSame(storage, tableStorage.getTxStateStorage(PARTITION_ID));
+
+        assertDoesNotThrow(() -> tableStorage.finishRebalance(PARTITION_ID).get(1, TimeUnit.SECONDS));
+
+        assertEquals(500L, storage.lastAppliedIndex());
+
+        assertThat(getAll(storage.scan()), containsInAnyOrder(new IgniteBiTuple<>(newTxId, newTxMeta)));
+    }
+
+    private static <T> List<T> getAll(Cursor<T> cursor) {
```
ibessonov, line 163: Why not just throw an exception? See no point in wrapping it into a runtime exception

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/TxStateStorageOnRebalance.java
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * @param delegate Delegate.
+     */
+    public TxStateStorageOnRebalance(TxStateStorage delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public TxMeta get(UUID txId) {
+        throw createDataWriteOnlyException();
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/TxStateStorageOnRebalance.java
```diff
@@ -0,0 +1,105 @@
...
+        delegate.destroy();
+    }
+
+    @Override
+    public void close() throws Exception {
+        delegate.close();
+    }
+
+    private IllegalStateException createDataWriteOnlyException() {
+        return new IllegalStateException(ERROR_MESSAGE);
```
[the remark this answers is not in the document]
tkalkirill: yes we can fix it later

modules/transactions/src/testFixtures/java/org/apache/ignite/internal/tx/storage/state/AbstractTxStateTableStorageTest.java
tkalkirill: Fix it

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/TxStateTableStorage.java
```diff
@@ -78,4 +75,49 @@ public interface TxStateTableStorage extends AutoCloseable {
...
+     *     <li>{@link #abortRebalance(int)} - in case of a full rebalance cancellation or failure, so that we can restore the transaction
+     *     state storage from a backup;</li>
+     *     <li>{@link #finishRebalance(int)} - in case of a successful full rebalance, to remove the backup of the transaction state
+     *     storage.</li>
+     * </ol>
+     *
+     * <p>Only modification of data in transaction state storage is allowed.
+     *
+     * @param partitionId Partition ID.
+     * @return Future, if completed without errors, then {@link #getTxStateStorage} will return a new (empty) transaction state storage.
```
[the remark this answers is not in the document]
tkalkirill: `Future, if completed without errors, then {@link #getTxStateStorage} will return write only (empty) transaction state storage.`

Fine ?

modules/transactions/src/testFixtures/java/org/apache/ignite/internal/tx/storage/state/AbstractTxStateTableStorageTest.java
tkalkirill: Sounds reasonable.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/TxStateStorageOnRebalance.java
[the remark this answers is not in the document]
tkalkirill: Changed it

modules/transactions/src/testFixtures/java/org/apache/ignite/internal/tx/storage/state/AbstractTxStateTableStorageTest.java
tkalkirill: Changed it

modules/transactions/src/testFixtures/java/org/apache/ignite/internal/tx/storage/state/AbstractTxStateTableStorageTest.java
tkalkirill: Changed it

PR #1315, opened 2022-11-04
IGNITE-17977: Add support for '--help' option to all commands

Flaugh24:
https://issues.apache.org/jira/browse/IGNITE-17977


modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/NodeReplCommand.java
```diff
@@ -17,15 +17,23 @@
...
+        subcommands = {
+                NodeConfigReplCommand.class,
+                NodeStatusReplCommand.class,
+                NodeVersionReplCommand.class,
+                NodeMetricReplCommand.class,
+                HelpCommand.class
+        },
         description = "Node operations")
-public class NodeReplCommand {
+public class NodeReplCommand extends BaseCommand {
```
PakhomovAlexander, line 38: What is this inheritance for?

Flaugh24: All commands must support '--help' option, if want to use 'command help' syntax.

ptupitsyn (approved):

PR #1318, opened 2022-11-07
IGNITE-17729 Forbidding a table and an index with the same name.

xtern:


modules/schema/src/main/java/org/apache/ignite/internal/schema/configuration/TableValidatorImpl.java
```diff
@@ -40,9 +42,19 @@ public class TableValidatorImpl implements Validator<TableValidator, NamedListVi
     /** {@inheritDoc} */
     @Override
     public void validate(TableValidator annotation, ValidationContext<NamedListView<TableView>> ctx) {
+        TablesView tablesConfig = ctx.getNewRoot(TablesConfiguration.KEY);
+
+        Set<String> idxNames = tablesConfig == null
+                ? Collections.emptySet()
+                : tablesConfig.indexes().namedListKeys().stream().map(String::toUpperCase).collect(Collectors.toSet());
```
korlov42, line 49: Why do you convert names to upper case?

xtern: Fixed, the case-insensitive comparison has been removed.

korlov42 (approved):

PR #1320, opened 2022-11-08
IGNITE-17930 Enable javadoc in gradle and fix doclint

PakhomovAlexander:
Now javadoc site can be built with gradle aggregateJavadoc task. By default, this task runs DocLint which performs some checks. There were several errors in raft module that now are fixed as well.

### DIFF
--- a/modules/raft/src/main/java/org/apache/ignite/raft/jraft/Iterator.java
+++ b/modules/raft/src/main/java/org/apache/ignite/raft/jraft/Iterator.java
@@ -33,7 +33,7 @@ public interface Iterator extends java.util.Iterator<ByteBuffer> {
     /**
      * Return a unique and monotonically increasing identifier of the current task: - Uniqueness guarantees that
      * committed tasks in different peers with the same index are always the same and kept unchanged. - Monotonicity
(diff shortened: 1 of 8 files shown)


modules/raft/src/main/java/org/apache/ignite/raft/jraft/Iterator.java
```diff
@@ -33,7 +33,7 @@ public interface Iterator extends java.util.Iterator<ByteBuffer> {
     /**
      * Return a unique and monotonically increasing identifier of the current task: - Uniqueness guarantees that
      * committed tasks in different peers with the same index are always the same and kept unchanged. - Monotonicity
-     * guarantees that for any index pair i, j (i < j), task at index |i| must be applied before task at index |j| in
+     * guarantees that for any index pair i, j (i less than j), task at index |i| must be applied before task at index |j| in
```
sashapolo, line 36: Please rollback changes in the `jraft` package and exclude it from the javadoc tool

PakhomovAlexander: Do you mean we should not generate javadoc for jraft package?

sashapolo (approved):

PR #1323, opened 2022-11-08
IGNITE-18015 Column's type mismatch error when inserting data into temporal columns with custom precision

AMashenkov:


modules/table/src/test/java/org/apache/ignite/internal/table/MutableRowTupleAdapterTest.java
```diff
@@ -543,6 +543,110 @@ public void testKeyValueSerialization() throws Exception {
...
+        Tuple tuple = Tuple.create().set("key", 1)
+                .set("time", LocalTime.of(1, 2))
+                .set("datetime", LocalDateTime.of(2022, 1, 2, 3, 4))
+                .set("timestamp", Instant.ofEpochSecond(123));
+
+        Row row = new Row(schemaDescriptor, new ByteBufferRow(marshaller.marshal(tuple).bytes()));
+
+        Tuple tuple1 = deserializeTuple(serializeTuple(TableRow.tuple(row)));
+
+        assertEquals(tuple, tuple1);
```
zstan, line 568: If someone broke such a test it will obtain noninformative like : 
`org.opentest4j.AssertionFailedError: 
Expected :org.apache.ignite.table.TupleImpl@eb5b12c
Actual   :org.apache.ignite.table.TupleImpl@e2022ac`

zstan:
overall looks good, but i fill one comment plz fix it.

korlov42 (changes requested):
It's OK to lose precision as in case of temporal types, but we definitely do not want to truncate customers data as in case of varlen types

modules/table/src/test/java/org/apache/ignite/internal/table/MutableRowTupleAdapterTest.java
```diff
@@ -543,6 +546,102 @@ public void testKeyValueSerialization() throws Exception {
...
+                .set("string", "abc")
+                .set("bytes", new byte[]{1, 2, 3});
+
+        Row row = new Row(schemaDescriptor, new ByteBufferRow(marshaller.marshal(expected).bytes()));
+
+        assertEquals(expected, deserializeTuple(serializeTuple(TableRow.tuple(row))));
+    }
+
+    @Test
+    void testDecimalPrecisionConstraint() throws Exception {
```
zstan, line 610: Exception never will be thrown from here.

zstan (approved):
1 minor comment, overall LGTM

korlov42 (approved):

PR #1326, opened 2022-11-09
IGNITE-17982 testFixturesAnnotationProcessor for ignite-network-annotation-processor does not work

valepakh:
https://issues.apache.org/jira/browse/IGNITE-17982

Since `@Transferable` and `@MessageGroup` needs to be in the same compilation unit, it would be best to move them all to testFixtures so any test or integration test can use them.


modules/raft/src/test/java/org/apache/ignite/raft/jraft/rpc/AbstractRpcTest.java
```diff
@@ -16,26 +16,25 @@
...
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
-import org.apache.ignite.network.annotations.Transferable;
 import org.apache.ignite.raft.jraft.entity.PeerId;
 import org.apache.ignite.raft.messages.TestRaftMessagesFactory;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
```
sashapolo, line 33: Please revert the import changes, this class is from JRaft, we need to change this class as little as possble

sashapolo (approved):

PR #1327, opened 2022-11-10
IGNITE-17959 Retrying in InternalTableImpl for a case when an operation throws ReplicaUnavailableException

sergeyuttsel:
https://issues.apache.org/jira/browse/IGNITE-17959


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/rel/TableScanNodeExecutionTest.java
```diff
@@ -155,7 +157,11 @@ public TestInternalTableImpl(
                     mock(MvTableStorage.class),
                     mock(TxStateTableStorage.class),
                     replicaSvc,
-                    mock(HybridClock.class)
+                    mock(HybridClock.class),
+                    new ScheduledThreadPoolExecutor(
+                            Runtime.getRuntime().availableProcessors(),
+                            new NamedThreadFactory("internal-table-scheduled-pool", log)
```
sanpwc, line 163: Well, I have to admit that proposed solution doesn't look good, despite the fact that I've pre-approved it earlier. new ScheduledThreadPoolExecutor() as a parameter looks crappy. In all that cases we need to shutdown an executors, etc.
So, let's consider other solutions, e.g. registering listener within replicaManager that will trigger reactive notification about replica appearance in the map. Will provide more details during the sync.

PR #1330, opened 2022-11-11
IGNITE-17998 Close the cursor synchronously when the session is closed.

xtern:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -693,6 +682,30 @@ private CompletableFuture<Void> close(boolean cancel) {
...
+                if (!cancel) {
+                    CompletableFuture<Void> closeFut = node.closeAsync();
+
+                    return fut.thenCompose(v -> closeFut);
+                }
+
+                node.onError(new ExecutionCancelledException());
+            }
+
+            return fut;
```
AMashenkov, line 707: Before the changes competed future was expected here. Is it ok?
line 637: return CompletableFuture.completedFuture(null);

AMashenkov (approved):

PR #1331, opened 2022-11-11
IGNITE-18122 Track last applied term and group config in storages

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18122

### DIFF
--- a/modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/PartitionMeta.java
+++ b/modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/PartitionMeta.java
@@ -48,6 +48,10 @@ public class PartitionMeta {
 
     private volatile long lastAppliedIndex;
 
+    private volatile long lastAppliedTerm;
+
+    private volatile byte @Nullable [] lastGroupConfig;
+
     private volatile long rowVersionFreeListRootPageId;
 
     private volatile long indexColumnsFreeListRootPageId;
@@ -80,6 +84,8 @@ public class PartitionMeta {
     public PartitionMeta(
             @Nullable UUID checkpointId,
             long lastAppliedIndex,
+            long lastAppliedTerm,
+            byte @Nullable [] lastGroupConfig,
             long rowVersionFreeListRootPageId,
             long indexColumnsFreeListRootPageId,
             long versionChainTreeRootPageId,
@@ -87,6 +93,8 @@ public class PartitionMeta {
             int pageCount
     ) {
         this.lastAppliedIndex = lastAppliedIndex;
+        this.lastAppliedTerm = lastAppliedTerm;
+        this.lastGroupConfig = lastGroupConfig;
         this.rowVersionFreeListRootPageId = rowVersionFreeListRootPageId;
         this.indexColumnsFreeListRootPageId = indexColumnsFreeListRootPageId;
         this.versionChainTreeRootPageId = versionChainTreeRootPageId;
@@ -107,6 +115,8 @@ public class PartitionMeta {
         this(
                 checkpointId,
                 metaIo.getLastAppliedIndex(pageAddr),
+                metaIo.getLastAppliedTerm(pageAddr),
+                metaIo.getLastGroupConfig(pageAddr),
                 metaIo.getRowVersionFreeListRootPageId(pageAddr),
                 metaIo.getIndexColumnsFreeListRootPageId(pageAddr),
                 metaIo.getVersionChainTreeRootPageId(pageAddr),
@@ -123,15 +133,43 @@ public class PartitionMeta {
     }
 
     /**
-     * Sets a last applied index value.
+     * Returns a last applied term value.
+     */
+    public long lastAppliedTerm() {
+        return lastAppliedTerm;
+    }
+
+    /**
+     * Sets last applied index and term.
      *
      * @param checkpointId Checkpoint ID.
      * @param lastAppliedIndex Last applied index value.
+     * @param lastAppliedTerm Last applied term value.
      */
-    public void lastAppliedIndex(@Nullable UUID checkpointId, long lastAppliedIndex) {
+    public void lastApplied(@Nullable UUID checkpointId, long lastAppliedIndex, long lastAppliedTerm) {
         updateSnapshot(checkpointId);
 
         this.lastAppliedIndex = lastAppliedIndex;
+        this.lastAppliedTerm = lastAppliedTerm;
+    }
+
+    /**
+     * Returns byte array representation of last group config.
+     */
+    public byte @Nullable [] lastGroupConfig() {
+        return lastGroupConfig;
+    }
+
+    /**
+     * Sets last group config.
+     *
+     * @param checkpointId Checkpoint ID.
+     * @param groupConfig Byte array representation of a group config.
+     */
+    public void lastGroupConfig(@Nullable UUID checkpointId, byte[] groupConfig) {
+        updateSnapshot(checkpointId);
+
+        this.lastGroupConfig = groupConfig;
     }
 
     /**
@@ -265,6 +303,10 @@ public class PartitionMeta {
 
         private final long lastAppliedIndex;
 
+        private final long lastAppliedTerm;
+
+        private final byte[] lastGroupConfig;
+
         private final long versionChainTreeRootPageId;
 
         private final long rowVersionFreeListRootPageId;
@@ -284,6 +326,8 @@ public class PartitionMeta {
         private PartitionMetaSnapshot(@Nullable UUID checkpointId, PartitionMeta partitionMeta) {
             this.checkpointId = checkpointId;
             lastAppliedIndex = partitionMeta.lastAppliedIndex;
+            lastAppliedTerm = partitionMeta.lastAppliedTerm;
+            lastGroupConfig = partitionMeta.lastGroupConfig;
             versionChainTreeRootPageId = partitionMeta.versionChainTreeRootPageId;
             rowVersionFreeListRootPageId = partitionMeta.rowVersionFreeListRootPageId;
             indexColumnsFreeListRootPageId = partitionMeta.indexColumnsFreeListRootPageId;
@@ -296,6 +340,20 @@ public class PartitionMeta {
          */
         public long lastAppliedIndex() {
             return lastAppliedIndex;
+        }
+
+        /**
+         * Returns a last applied term value.
+         */
+        public long lastAppliedTerm() {
+            return lastAppliedTerm;
+        }
+
+        /**
+         * Returns last group config.
+         */
+        public byte @Nullable [] lastGroupConfig() {
+            return lastGroupConfig;
         }
 
         /**
@@ -341,6 +399,7 @@ public class PartitionMeta {
          */
         void writeTo(PartitionMetaIo metaIo, long pageAddr) {
             metaIo.setLastAppliedIndex(pageAddr, lastAppliedIndex);
+            metaIo.setLastAppliedTerm(pageAddr, lastAppliedTerm);
             metaIo.setVersionChainTreeRootPageId(pageAddr, versionChainTreeRootPageId);
             metaIo.setIndexColumnsFreeListRootPageId(pageAddr, indexColumnsFreeListRootPageId);
             metaIo.setRowVersionFreeListRootPageId(pageAddr, rowVersionFreeListRootPageId);
--- a/modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/io/PartitionMetaIo.java
+++ b/modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/io/PartitionMetaIo.java
@@ -26,6 +26,7 @@ import static org.apache.ignite.internal.pagememory.util.PageUtils.putLong;
 import org.apache.ignite.internal.pagememory.io.IoVersions;
 import org.apache.ignite.internal.pagememory.io.PageIo;
 import org.apache.ignite.lang.IgniteStringBuilder;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Io for partition metadata pages.
@@ -33,7 +34,9 @@ import org.apache.ignite.lang.IgniteStringBuilder;
 public class PartitionMetaIo extends PageIo {
     private static final int LAST_APPLIED_INDEX_OFF = COMMON_HEADER_END;
 
-    private static final int ROW_VERSION_FREE_LIST_ROOT_PAGE_ID_OFF = LAST_APPLIED_INDEX_OFF + Long.BYTES;
+    private static final int LAST_APPLIED_TERM_OFF = LAST_APPLIED_INDEX_OFF + Long.BYTES;
+
+    private static final int ROW_VERSION_FREE_LIST_ROOT_PAGE_ID_OFF = LAST_APPLIED_TERM_OFF + Long.BYTES;
 
     private static final int INDEX_COLUMNS_FREE_LIST_ROOT_PAGE_ID_OFF = ROW_VERSION_FREE_LIST_ROOT_PAGE_ID_OFF + Long.BYTES;
 
@@ -64,6 +67,8 @@ public class PartitionMetaIo extends PageIo {
         super.initNewPage(pageAddr, pageId, pageSize);
 
         setLastAppliedIndex(pageAddr, 0);
+        setLastAppliedTerm(pageAddr, 0);
+        setLastGroupConfig(pageAddr, null);
         setRowVersionFreeListRootPageId(pageAddr, 0);
         setIndexColumnsFreeListRootPageId(pageAddr, 0);
         setVersionChainTreeRootPageId(pageAddr, 0);
@@ -84,12 +89,55 @@ public class PartitionMetaIo extends PageIo {
     }
 
     /**
+     * Sets a last applied term value.
+     *
+     * @param pageAddr Page address.
+     * @param lastAppliedTerm Last applied term value.
+     */
+    public void setLastAppliedTerm(long pageAddr, long lastAppliedTerm) {
+        assertPageType(pageAddr);
+
+        putLong(pageAddr, LAST_APPLIED_TERM_OFF, lastAppliedTerm);
+    }
+
+    /**
+     * Sets last group config.
+     *
+     * @param pageAddr Page address.
+     * @param lastGroupConfig Byte representation of last group config..
+     */
+    public void setLastGroupConfig(long pageAddr, byte @Nullable [] lastGroupConfig) {
+        assertPageType(pageAddr);
+
+        // TODO: IGNITE-18118 - implement
+    }
+
+    /**
      * Returns a last applied index value.
      *
      * @param pageAddr Page address.
      */
     public long getLastAppliedIndex(long pageAddr) {
         return getLong(pageAddr, LAST_APPLIED_INDEX_OFF);
+    }
+
+    /**
+     * Returns a last applied term value.
+     *
+     * @param pageAddr Page address.
+     */
+    public long getLastAppliedTerm(long pageAddr) {
+        return getLong(pageAddr, LAST_APPLIED_TERM_OFF);
+    }
+
+    /**
+     * Returns byte representation of group config.
+     *
+     * @param pageAddr Page address.
+     */
+    public byte @Nullable [] getLastGroupConfig(long pageAddr) {
+        // TODO: IGNITE-18118 - implement
+        return null;
     }
 
     /**
@@ -202,6 +250,7 @@ public class PartitionMetaIo extends PageIo {
     protected void printPage(long addr, int pageSize, IgniteStringBuilder sb) {
         sb.app("TablePartitionMeta [").nl()
                 .app("lastAppliedIndex=").app(getLastAppliedIndex(addr)).nl()
+                .app("lastAppliedTerm=").app(getLastAppliedTerm(addr)).nl()
                 .app(", rowVersionFreeListRootPageId=").appendHex(getRowVersionFreeListRootPageId(addr)).nl()
                 .app(", indexColumnsFreeListRootPageId(=").appendHex(getIndexColumnsFreeListRootPageId(addr)).nl()
                 .app(", versionChainTreeRootPageId=").appendHex(getVersionChainTreeRootPageId(addr)).nl()
--- a/modules/page-memory/src/test/java/org/apache/ignite/internal/pagememory/persistence/PartitionMetaTest.java
+++ b/modules/page-memory/src/test/java/org/apache/ignite/internal/pagememory/persistence/PartitionMetaTest.java
@@ -24,11 +24,14 @@ import static org.apache.ignite.internal.pagememory.util.PageIdUtils.pageIndex;
 import static org.apache.ignite.internal.pagememory.util.PageIdUtils.partitionId;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
 
 import java.util.UUID;
 import org.apache.ignite.internal.pagememory.persistence.PartitionMeta.PartitionMetaSnapshot;
+import org.jetbrains.annotations.Nullable;
 import org.junit.jupiter.api.Test;
 
 /**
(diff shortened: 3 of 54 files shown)


modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/FSMCallerImpl.java
```diff
@@ -497,6 +497,17 @@ private void doCommitted(final long committedIndex) {
                 final LogEntry logEntry = iterImpl.entry();
                 if (logEntry.getType() != EnumOutter.EntryType.ENTRY_TYPE_DATA) {
                     if (logEntry.getType() == EnumOutter.EntryType.ENTRY_TYPE_CONFIGURATION) {
+                        ConfigurationEntry configurationEntry = new ConfigurationEntry(
+                                logEntry.getId().copy(),
+                                new Configuration(logEntry.getPeers(), logEntry.getLearners()),
+                                null
```
sashapolo, line 503: Looks like `ConfigurationEntry` does not expect to be provided with `null` here, you should use an empty configuration instead

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/FSMCallerImpl.java
```diff
@@ -665,6 +676,27 @@ private void doSnapshotLoad(final LoadSnapshotClosure done) {
             setError(e);
             return;
         }
+
+        // Tests use such strange metas, so we have to protect... in production, these are never null.
```
sashapolo, line 680: What tests? Are they ours or from Jraft?

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/GroupConfiguration.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.List;
+import java.util.Objects;
+import org.apache.ignite.internal.tostring.IgniteToStringInclude;
+import org.apache.ignite.internal.tostring.S;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A RAFT group configuration at which it was committed.
+ */
+public class GroupConfiguration implements Serializable {
```
sashapolo, line 31: Shall we rename it to `RaftGroupConfiguration`?

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
```diff
@@ -71,20 +71,39 @@ public interface MvPartitionStorage extends ManuallyCloseable {
...
 
     /**
-     * Index of the highest write command applied to the storage. {@code 0} if index is unknown.
+     * Index of the highest write command applied to the storage. {@code 0} if the index is unknown.
      */
     long lastAppliedIndex();
 
     /**
-     * Sets the last applied index value.
+     * Term of the highest write command applied to the storage. {@code 0} if the term is unknown.
```
sashapolo, line 79: What do you mean by `highest write command`?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageTest.java
```diff
@@ -1323,6 +1325,35 @@ void committedMethodCallDoesNotInterfereWithIteratingOverScanCursor(ScanTimestam
...
+        storage.runConsistently(() -> {
+            storage.committedGroupConfiguration(configToSave);
+
+            return null;
+        });
+
+        GroupConfiguration returnedConfig = storage.committedGroupConfiguration();
+
+        assertThat(returnedConfig, is(notNullValue()));
+        assertThat(returnedConfig.peers(), is(List.of("peer1", "peer2")));
```
sashapolo, line 1351: `GroupConfiguration` has `equals` implemented, so you can compare the objects directly

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTablePersistenceTest.java
```diff
@@ -213,6 +230,28 @@ public BooleanSupplier snapshotCheckClosure(JraftServerImpl restarted, boolean i
         };
     }
 
+    private static Map<ByteBuffer, RowId> rowIdsToRows(MvPartitionStorage storage) {
```
sashapolo, line 233: I think this method should be called the opposite

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotMvDataStreamingTest.java
```diff
@@ -281,6 +280,43 @@ void sendsOutOfOrderRowsWithHighestPriority() {
         assertThat(response.rows().get(1).rowId(), is(rowId1.uuid()));
     }
 
+    @Test
+    void doesNotEnqueueMissingRows() {
+        when(mvPartitionStorage.scanVersions(rowIdOutOfOrder)).thenReturn(Cursor.fromIterator(emptyIterator()));
```
sashapolo, line 285: There is a `CursorUtils#emptyCursor` method

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManagerTest.java
```diff
@@ -59,15 +67,22 @@ void emptyOngoingSnapshotsIfNoSnapshotWasRegistered() {
...
+
+        when(partitionAccess.partitionKey()).thenReturn(partitionKey);
+        when(partitionAccess.mvPartitionStorage()).thenReturn(mvPartitionStorage);
+        when(partitionAccess.txStatePartitionStorage()).thenReturn(mock(TxStateStorage.class));
+
+        when(mvPartitionStorage.committedGroupConfiguration()).thenReturn(mock(GroupConfiguration.class));
+
+        OutgoingSnapshot snapshot = new OutgoingSnapshot(UUID.randomUUID(), partitionAccess);
 
         manager.startOutgoingSnapshot(UUID.randomUUID(), snapshot);
```
sashapolo, line 81: Do you test that `startOutgoingSnapshot` does not throw any exceptions?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
```diff
@@ -115,11 +118,13 @@ public TxStateRocksDbStorage(
         this.persistedTierReadOptions = persistedTierReadOptions;
         this.partitionId = partitionId;
         this.tableStorage = tableStorage;
-        this.lastAppliedIndexKey = ByteBuffer.allocate(Short.BYTES).order(ByteOrder.BIG_ENDIAN)
+        this.lastAppliedIndexAndTermKey = ByteBuffer.allocate(Short.BYTES).order(ByteOrder.BIG_ENDIAN)
                 .putShort((short) partitionId)
                 .array();
 
-        lastAppliedIndex = readLastAppliedIndex(readOptions);
+        byte[] indexAndTermBytes = readLastAppliedIndexAndTerm(readOptions);
```
sashapolo, line 125: That's actually interesting: why do you store index and term together here, but not in the Partition Storage?

modules/page-memory/src/test/java/org/apache/ignite/internal/pagememory/persistence/PartitionMetaTest.java
```diff
@@ -41,15 +44,45 @@ void testLastAppliedIndex() {
...
         assertEquals(500, meta.lastAppliedIndex());
     }
 
+    @Test
+    void testLastAppliedTerm() {
+        PartitionMeta meta = new PartitionMeta();
+
+        assertEquals(0, meta.lastAppliedTerm());
+
+        assertDoesNotThrow(() -> meta.lastApplied(null, 100, 10));
```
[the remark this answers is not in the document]
rpuch: When you just call a method, it's not immediately clear what you want to check here. After a second of thinking you understand: oh, they are making sure it does not explode. On the other hand, using `assertDoesNotThrow()` makes this expectation explicit. I would prefer to leave it here.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/FSMCallerImpl.java
```diff
@@ -497,6 +497,17 @@ private void doCommitted(final long committedIndex) {
...
+                        );
+                        if (logEntry.getOldPeers() != null && !logEntry.getOldPeers().isEmpty()) {
+                            configurationEntry.setOldConf(new Configuration(logEntry.getOldPeers(), logEntry.getOldLearners()));
+                        }
+
+                        this.fsm.onRawConfigurationCommitted(configurationEntry);
+
                         if (logEntry.getOldPeers() != null && !logEntry.getOldPeers().isEmpty()) {
                             // Joint stage is not supposed to be noticeable by end users.
                             this.fsm.onConfigurationCommitted(new Configuration(iterImpl.entry().getPeers()));
```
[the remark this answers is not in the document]
rpuch: They probably think that a state machine might have some interest in the current peers, but not learners. I have no idea why, to be honest.

modules/page-memory/src/test/java/org/apache/ignite/internal/pagememory/persistence/PartitionMetaTest.java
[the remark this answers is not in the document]
sashapolo: But this is not true: you actually check that `lastApplied` sets the index and the term, which is checked below using `assertEquals`

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/FSMCallerImpl.java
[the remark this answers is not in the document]
sashapolo: ok, let's hope that's not a bug

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
rpuch: It's the command that was written with the highest index

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/PartitionMeta.java
```diff
@@ -341,6 +399,7 @@ public int pageCount() {
          */
         void writeTo(PartitionMetaIo metaIo, long pageAddr) {
             metaIo.setLastAppliedIndex(pageAddr, lastAppliedIndex);
+            metaIo.setLastAppliedTerm(pageAddr, lastAppliedTerm);
```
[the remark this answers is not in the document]
rpuch: Good catch!

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/FSMCallerImpl.java
rpuch: JRaft tests. I've clarified the comment.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotMvDataStreamingTest.java
rpuch: Switched to it. BTW, why wouldn't we have that method on the `Cursor` itself? It would increase its visibility for developers.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotsManagerTest.java
rpuch: Sure, this was the intention. Thanks, added the assertion

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
rpuch: In the partition storage, we store the metadata (including index and term) in a separate column family. Here, there was just one metadata originally (last applied index), so the author employed a hack: a special key was used to store the index in the same column family that hosts the data. Now we need to store 2 pieces of metadata, so it becomes a bit ugly to increase the number of keys.

I invented the following solutions:

1. Add `meta` CF as it is done for MV data storage. This is a bit cumbersome.
2. Construct metadata keys like this: `concat(partitionId, 0)` for index, `concat(partitionId, 1)` for term and then skip them when iterating over the keys in `scan()`. I implemented this first, but it causes a penalty of reading the key to analyze its length on each  iteration, and it's a bit cumbersome as well.
3. Reuse the same (the only) metadata key to store index together with term. They are always updated together, so this looked like a natural solution, I chose this one.

modules/page-memory/src/test/java/org/apache/ignite/internal/pagememory/persistence/PartitionMetaTest.java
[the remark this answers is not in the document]
rpuch: Oops, I've confused the situation. Now I know the answer: I did like it's done because there is another test that was written before me about `index`. I wanted to avoid, ahem, stylistic differences.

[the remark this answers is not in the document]
sashapolo: Can we change the other tests instead? This looks quite strange to me

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
sashapolo: Can you state that explicitly? It's too confusing otherwise

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotMvDataStreamingTest.java
sashapolo: We used to, but you will have to make `EMPTY_CURSOR` public, which I don't like, so I decided to move it here (like `Collections.emptyList()`. Maybe we can add a method to `Cursor` which delegates to `CursorUtils` but I don't know if it's a good idea

modules/page-memory/src/test/java/org/apache/ignite/internal/pagememory/persistence/PartitionMetaTest.java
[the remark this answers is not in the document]
rpuch: Removed it in first 3 test methods

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
sashapolo: Why don't we use the same approach in the Partition Storage?

rpuch: Different original authors, I guess

rpuch: Also, for TX partition storage meta is simpler. We don't need to store configuration there, which is A) many bytes, and B) it is updated independently from index+term. So this 'squashing to same value' would not work well for MV partition meta, but it does work for TX data partition meta.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
rpuch: Done. Please take a look whether it's ok now.

sashapolo (approved):

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/FSMCallerImpl.java
```diff
@@ -665,6 +676,27 @@ private void doSnapshotLoad(final LoadSnapshotClosure done) {
             setError(e);
             return;
         }
+
+        // JRaft tests use such strange metas, so we have to protect... in production, these are never null.
```
alievmirza, line 680: I didn't get this comment

rpuch: Take a look at `FSMCallerTest:166` (`testOnSnapshotLoad()`). The test constructs a snapshot meta without peers/learners info (so `peers()` and `learners()` return `null`). Here, the `if` protects from such strange metas.

In production code, `peers()` and `learners()` on `SnapshotMeta` never return `null`.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/io/PartitionMetaIo.java
```diff
@@ -83,6 +88,30 @@ public void setLastAppliedIndex(long pageAddr, long lastAppliedIndex) {
...
+    /**
+     * Sets last group config.
+     *
+     * @param pageAddr Page address.
+     * @param lastGroupConfig Byte representation of last group config..
+     */
+    public void setLastGroupConfig(long pageAddr, byte @Nullable [] lastGroupConfig) {
+        assertPageType(pageAddr);
+
+        // TODO: IGNITE-18118 - implement
```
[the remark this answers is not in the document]
rpuch: I added a description to https://issues.apache.org/jira/browse/IGNITE-18118

This discusses a couple of alternatives and requires a discussion.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/FSMCallerImpl.java
```diff
@@ -665,6 +676,27 @@ private void doSnapshotLoad(final LoadSnapshotClosure done) {
...
+                    snapshotId.copy(),
+                    new Configuration(
+                            meta.peersList().stream().map(PeerId::parsePeer).collect(toList()),
+                            meta.learnersList().stream().map(PeerId::parsePeer).collect(toList())
+                    ),
+                    new Configuration()
+            );
+            if (meta.oldPeersList() != null && !meta.oldPeersList().isEmpty()) {
+                configurationEntry.setOldConf(new Configuration(
+                        meta.oldPeersList().stream().map(PeerId::parsePeer).collect(toList()),
```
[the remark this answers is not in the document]
rpuch: Are you asking about https://github.com/apache/ignite-3/commit/aacb6ab15651a49f08a3841760537823cb8c34cf ? If yes, then yes, this commit was merged to `main` before I branched from `main`, my branch contains it.

Am I doing something wrong? I can see in my tests that peers are now identified with strings like `itrst_t_0` instead of `127.0.1.1:3344`, but everything seems working fine.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -178,11 +180,29 @@ public class RocksDbMvPartitionStorage implements MvPartitionStorage {
     /** Key to store applied index value in meta. */
     private final byte[] lastAppliedIndexKey;
 
+    /** Key to store applied term value in meta. */
+    private final byte[] lastAppliedTermKey;
```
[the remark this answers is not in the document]
rpuch: We could, but should we? This makes a code arguably more difficult to understand.

In `TxStateRocksDbStorage` this is what I did: I store them under the same key, but there we don't use a separate meta CF, so we only have 1 key that can be used in a 'not too ugly way'.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotReader.java
```diff
@@ -66,7 +66,7 @@ public void shutdown() {
 
     @Override
     public String getPath() {
-        throw new UnsupportedOperationException("No path for the rebalance snapshot");
+        return "";
```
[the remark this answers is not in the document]
rpuch: `FSMCallerImpl:672` (in `doSnapshotLoad()`):

```
if (!this.fsm.onSnapshotLoad(reader)) {
```

Which, in turn, calls `getPath()`, the result of which is later ignored by our `PartitionListener#onSnapshotLoad()`. But if an exception is thrown, it impedes the snapshot load process.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshot.java
```diff
@@ -143,24 +141,36 @@ public PartitionKey partitionKey() {
...
+            lastAppliedTerm = partition.mvPartitionStorage().lastAppliedTerm();
+        } else {
+            lastAppliedIndex = partition.txStatePartitionStorage().lastAppliedIndex();
+            lastAppliedTerm = partition.txStatePartitionStorage().lastAppliedTerm();
+        }
+
+        return SnapshotMetaUtils.snapshotMetaAt(
+                lastAppliedIndex,
+                lastAppliedTerm,
+                requireNonNull(partition.mvPartitionStorage().committedGroupConfiguration())
```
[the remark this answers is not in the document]
rpuch: In JRaft, the first entry in log (index=1, term=1) is always a configuration. So, if at least 1 log entry is applied, `committedGroupConfiguration()` will be non-null. Also, a leader might with to install a snapshot only if it is ahead of a follower, so it has something applied, so in this code the returned config` is never null.

I replaced the check with an assertion. WDYT?

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/FSMCallerImpl.java
```diff
@@ -665,6 +676,27 @@ private void doSnapshotLoad(final LoadSnapshotClosure done) {
             setError(e);
             return;
         }
+
+        // JRaft tests use such strange metas, so we have to protect... in production, these are never null.
+        if (meta.peersList() != null && meta.learnersList() != null) {
```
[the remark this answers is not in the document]
rpuch: I clarified the comment. I made the condition the most protective just in case as here we are protecting from test code. In production, this doesn't matter, neither peers no learners can be null.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -327,15 +383,23 @@ public void onSnapshotSave(Path path, Consumer<Throwable> doneClo) {
         //      4) When we try to restore data starting from the minimal lastAppliedIndex, we come to the situation
         //         that a raft node doesn't have such data, because the truncation until the maximal lastAppliedIndex from 1) has happened.
         //      5) Node cannot finish local recovery.
-        long maxLastAppliedIndex = Math.max(storage.lastAppliedIndex(), txStateStorage.lastAppliedIndex());
+        long maxLastAppliedIndex;
+        long maxLastAppliedTerm;
```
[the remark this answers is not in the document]
rpuch: Looks like we can as term must be monothonic for committed log entries (and here we only have committed ones).

Applied this change in all 3 places.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/FSMCallerImpl.java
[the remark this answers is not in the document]
ibessonov: No, nothing's wrong here, I'm worrying about conflicts, that's it

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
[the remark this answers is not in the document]
ibessonov: Maybe I'm preemptively thinking about performance. One insertion is better than two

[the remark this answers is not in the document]
rpuch: It is an insertion happening in memory inside a batch. It seems that even if we change it to one insert it will be really difficult to see any difference, even under load (where we'll probably be capped by FS performance, not by this difference).

Anyway, I don't have a strong opinion on this. If you believe we should change this, please let me know.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotReader.java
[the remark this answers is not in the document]
ibessonov: Ok, didn't encounter this situation before. A little comment would be nice to have :)

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/FSMCallerImpl.java
rpuch: I clarified the comment a bit.

PR #1332, opened 2022-11-11
IGNITE-18125 Fix NullPointerException in InternalTableImpl#enlist()

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18125

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
@@ -1125,7 +1125,7 @@ public class InternalTableImpl implements InternalTable {
         // TODO asch a leader race is possible when enlisting different keys from the same partition.
         return fut0.handle((primaryPeerAndTerm, e) -> {
             if (e != null) {
-                throw new TransactionException(REPLICA_UNAVAILABLE_ERR, "Failed to get the primary replica.", e);
+                throw withCause(TransactionException::new, REPLICA_UNAVAILABLE_ERR, "Failed to get the primary replica.", e);
             }
             if (primaryPeerAndTerm.get1() == null) {
                 throw new TransactionException(REPLICA_UNAVAILABLE_ERR, "Failed to get the primary replica.");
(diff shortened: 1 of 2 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -1124,7 +1124,10 @@ protected CompletableFuture<IgniteBiTuple<ClusterNode, Long>> enlist(int partId,
         // TODO asch IGNITE-15091 fixme need to map to the same leaseholder.
         // TODO asch a leader race is possible when enlisting different keys from the same partition.
         return fut0.handle((primaryPeerAndTerm, e) -> {
-            if (primaryPeerAndTerm.get1() == null || e != null) {
+            if (e != null) {
+                throw new TransactionException("Failed to get the primary replica.", e);
```
sanpwc, line 1128: I would rather use withCause with REPLICA_UNAVAILABLE_ERR code here.

sanpwc (changes requested):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -1124,7 +1124,10 @@ protected CompletableFuture<IgniteBiTuple<ClusterNode, Long>> enlist(int partId,
         // TODO asch IGNITE-15091 fixme need to map to the same leaseholder.
         // TODO asch a leader race is possible when enlisting different keys from the same partition.
         return fut0.handle((primaryPeerAndTerm, e) -> {
-            if (primaryPeerAndTerm.get1() == null || e != null) {
+            if (e != null) {
```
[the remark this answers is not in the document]
rpuch: I split the block to 2 `if` blocks because the exception to be thrown is constructed a bit differently: in one case it has a cause, in the other it does not.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
rpuch: Thanks, I did not find that error code myself. Fixed

rpuch: I hope I got this right this time :)

sanpwc (approved):

PR #1337, opened 2022-11-11
IGNITE-17917 Unify ignite3db start scripts

PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-17917

Now rpm, deb, zip and docker distributions use the same base variables and functions. There is no code duplication anymore.

Also, provided some zip package fixes:
- logs written to log dir
- the place where the start script is called for does not affect the script execution


Pochatkin:
LGTM

PR #1338, opened 2022-11-11
IGNITE-18081 .NET: Basic LINQ provider

ptupitsyn:
* Port LINQ provider from Ignite 2.x, adapt to new APIs.
* Initial implementation supports basic queries on top of `IRecordView<T>`.

**Example**
```csharp
ITable? table = await Client.Tables.GetTableAsync("TBL1");
IRecordView<Poco> view = table!.GetRecordView<Poco>();

var query = view.AsQueryable()
    .Where(x => x.Key > 2)
    .OrderBy(x => x.Key)
    .Select(x => x.Val);

IResultSet<string?> resultSet = await query.ToResultSetAsync();
```

The code above generates and executes the following query:
```sql
select _T0.VAL 
from PUBLIC.TBL1 as _T0 
where (_T0.KEY > ?) 
order by (_T0.KEY) asc
```


modules/platforms/dotnet/Apache.Ignite/Sql/QueryableOptions.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Apache.Ignite.Sql;
+
+using System;
+
+/// <summary>
+/// Options for LINQ queryables.
+/// </summary>
+/// <param name="Timeout">Timeout, default <see cref="SqlStatement.DefaultTimeout"/>.</param>
+/// <param name="PageSize">Page size to use when fetching results from server, default <see cref="SqlStatement.PageSize"/>.</param>
```
isapego, line 26: `SqlStatement.DefaultPageSize`?

isapego (approved):

modules/platforms/dotnet/Apache.Ignite/Sql/QueryableOptions.cs
ptupitsyn: Fixed.

PR #1339, opened 2022-11-14
IGNITE-18093 Distribution zones

sergeyuttsel:
https://issues.apache.org/jira/browse/IGNITE-18093


modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
```diff
@@ -0,0 +1,187 @@
...
+        this.dataNodesAutoAdjustScaleUp = dataNodesAutoAdjustScaleUp;
+        this.dataNodesAutoAdjustScaleDown = dataNodesAutoAdjustScaleDown;
+    }
+
+    /**
+     * Gets the zone name.
+     *
+     * @return The zone name.
+     */
+    @Nullable
```
alievmirza, line 56: This should on the same line with method

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
```diff
@@ -0,0 +1,187 @@
...
+         * Builds the distribution zone configuration.
+         *
+         * @return Distribution zone configuration.
+         */
+        public DistributionZoneCfg build() {
+            if (dataNodesAutoAdjust != null
+                    && (dataNodesAutoAdjustScaleUp != null || dataNodesAutoAdjustScaleDown != null)
+                ) {
+                throw new IllegalArgumentException(
+                        String.format("Not compatible parameters [dataNodesAutoAdjust=%s, "
```
alievmirza, line 173: I would add here some more info, like why this parameters are incompatible

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
```diff
@@ -0,0 +1,187 @@
...
+    /** Data nodes auto adjust scale up timeout. */
+    private final Integer dataNodesAutoAdjustScaleUp;
+
+    /** Data nodes auto adjust scale down timeout. */
+    private final Integer dataNodesAutoAdjustScaleDown;
+
+    /**
+     * The constructor.
+     */
+    public DistributionZoneCfg(String name,
```
alievmirza, line 41: seems like this should be private

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
```diff
@@ -0,0 +1,187 @@
...
+            this.dataNodesAutoAdjustScaleUp = dataNodesAutoAdjustScaleUp;
+
+            return this;
+        }
+
+        /**
+         * Sets timeout in seconds between node left topology event itself and data nodes switch.
+         *
+         * @param dataNodesAutoAdjustScaleDown Timeout in seconds between node left topology event itself
+         *                                     and data nodes switch.
```
alievmirza, line 154: please check formatting

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
```diff
@@ -0,0 +1,187 @@
...
+    /** Data nodes auto adjust scale down timeout. */
+    private final Integer dataNodesAutoAdjustScaleDown;
+
+    /**
+     * The constructor.
+     */
+    public DistributionZoneCfg(String name,
+                               Integer dataNodesAutoAdjust,
+                               Integer dataNodesAutoAdjustScaleUp,
+                               Integer dataNodesAutoAdjustScaleDown) {
```
alievmirza, line 44: formatting

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
```diff
@@ -0,0 +1,187 @@
...
+ */
+
+package org.apache.ignite.internal.distributionzones;
+
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Distribution zone configuration.
+ */
+public class DistributionZoneCfg {
```
sk0x50, line 25: Minor. We should not use abbreviation in class names for sure.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
sk0x50, line 173: Minor. I don't see any benefits in using `String.format()` here. Anyway, it's up to you.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
```diff
@@ -0,0 +1,187 @@
...
+                    && (dataNodesAutoAdjustScaleUp != null || dataNodesAutoAdjustScaleDown != null)
+                ) {
+                throw new IllegalArgumentException(
+                        String.format("Not compatible parameters [dataNodesAutoAdjust=%s, "
+                                        + "dataNodesAutoAdjustScaleUp=%s, dataNodesAutoAdjustScaleDown=%s]",
+                                dataNodesAutoAdjust, dataNodesAutoAdjustScaleUp, dataNodesAutoAdjustScaleDown)
+                );
+            }
+
+            return new DistributionZoneCfg(
```
sk0x50, line 179: Looks like the `name` is mandatory parameter for creating/altering a distribution zone. Thus, it would be nice to add the corresponding check right here. What do you think? Do we have a case when the `name` is not used?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
sergeyuttsel: fixed

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
sergeyuttsel: I think it's not needed. The builder is a part of internal api that will be used by SQL module. SQL module will check query before a builder creation. So actually this exception will never be thrown.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
sergeyuttsel: fixed

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
sergeyuttsel: fixed

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
sergeyuttsel: fixed

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
sergeyuttsel: fixed

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
sergeyuttsel: fixed

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneCfg.java
sergeyuttsel: fixed

PR #1344, opened 2022-11-15
IGNITE-16975 Update jansi version to fix CLI colors

Flaugh24:
https://issues.apache.org/jira/browse/IGNITE-16975


valepakh:
Verified that underline and green colors are now working!

valepakh (approved):

PR #1345, opened 2022-11-16
IGNITE-17953 Sql - fix NPE on some malformed queries

zstan:


korlov42 (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -473,11 +473,19 @@ private List<CompletableFuture<AsyncSqlCursor<List<Object>>>> query0(
...
 
         var res = new ArrayList<CompletableFuture<AsyncSqlCursor<List<Object>>>>(nodes.size());
 
-        CompletableFuture<Void> start = new CompletableFuture<>();
+        try {
+            nodes = Commons.parse(sql, FRAMEWORK_CONFIG.getParserConfig());
+        } catch (Throwable th) {
+            start.completeExceptionally(th);
+
+            res.add(CompletableFuture.completedFuture(failedCursor(th)));
```
AMashenkov, line 487: Why don't just return `List.of(CompletableFuture.completedFuture(th))`?
or `List.of(CompletableFuture.completedFuture(failedCursor(th)))` ?

Does it make sense to return failed cursor instead of just failed future?

PR #1358, opened 2022-11-21
IGNITE-18049 Cursor#close() should not declare any exceptions in throws clause

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18049


modules/rest/openapi/openapi.yaml
```diff
@@ -6,7 +6,7 @@ info:
   license:
     name: Apache 2.0
     url: https://ignite.apache.org
-  version: 3.0.0-SNAPSHOT
+  version: 3.0.0-alpha
```
ibessonov, line 9: What's this?

modules/core/src/test/java/org/apache/ignite/internal/util/CursorTest.java
```diff
@@ -0,0 +1,76 @@
...
+        cursor.close();
+
+        verify(iterator).close();
+    }
+
+    private interface CloseableIterator extends Iterator<Integer>, AutoCloseable {
+        @Override
+        void close();
+    }
+}
```
ibessonov, line 76: Please add empty line at the end

ibessonov (approved):

modules/rest/openapi/openapi.yaml
rpuch: Oh... this is an auto-generated file, IIRC, we must commit as soon as it generates with a new version. I committed it accidentally, but it probably needs to be committed. I will ask the guys in Slack.

rpuch: Reverted thechange

PR #1359, opened 2022-11-21
IGNITE-18165 Applied short term locks to sorted indexes

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-18165

### DIFF
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
@@ -1,0 +1,247 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
```diff
@@ -0,0 +1,247 @@
...
+        clearData(getOrCreateTable());
+    }
+
+    @Test
+    public void testInsertWaitScanComplete() throws Exception {
+        TableImpl table = getOrCreateTable();
+
+        InternalTable internalTable = table.internalTable();
+
+        UUID soredIndexId = getSortedIndexId();
```
denis-chudov, line 79: `sortedIndexId`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1435,9 +1452,15 @@ private CompletableFuture<Object> processSingleEntryAction(ReadWriteSingleRowRep
                     RowId rowId0 = new RowId(partId);
 
                     return takeLocksForInsert(searchRow, rowId0, txId)
-                            .thenCompose(ignored -> applyCmdWithExceptionHandling(
-                                    updateCommand(commitPartitionId, rowId0.uuid(), searchRow.byteBuffer(), txId)))
-                            .thenApply(ignored -> true);
+                            .thenCompose(rowIdLock -> applyCmdWithExceptionHandling(
+                                    updateCommand(commitPartitionId, rowId0.uuid(), searchRow.byteBuffer(), txId))
+                                    .thenApply(ignored -> rowIdLock))
```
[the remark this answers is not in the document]
vldpyatkov: This is an auto code aligns.
I see we use the same aligns in another places. I am afraid of I can do nothing, because in opposite case I will receive a warning for our check style suite.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
vldpyatkov: Fixed.

PR #1363, opened 2022-11-22
IGNITE-18148 Add more dynamic completers

PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-18148

The architecture of dynamic completers were
refactored. Now it is more flexible and the
lifecycle of completers is defined in factories.
So, if some data for suggestions (say, node list)
should be changed then the factory will create
new DynamicCompleter with updated data inside.

The Options class is introduced. It collects all
options that are used in the CLI in one place.
This gives us two advantages: DRY and the ability
to use Options class in the completion strategies. 
For example, cluster init --cluster-name the
option that should be completed only once. This
information could be placed in Options class in
the future.

To demonstrate the advantages of refactoring I've
added two more completions to interactive mode:
connect command and cluster init.

Follow-up tickets:
https://issues.apache.org/jira/browse/IGNITE-18220 
https://issues.apache.org/jira/browse/IGNITE-18221
https://issues.apache.org/jira/browse/IGNITE-18222


modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/DynamicCompleterFilter.java
```diff
@@ -43,25 +49,31 @@ private static boolean optionTyped(String[] words) {
...
+                || VERBOSE_OPTION_SHORT.equals(candidate)
+                || VERBOSE_OPTION.equals(candidate));
     }
 
     private boolean filterClusterUrl(String[] words, String candidate) {
-        if (optionTyped(words)) {
-            return true;
-        }
+        return optionTyped(words) || !session.isConnectedToNode() || (!candidate.equals(CLUSTER_URL_OPTION) && !candidate.equals(
+                NODE_URL_OPTION));
```
valepakh, line 76: Weird line break, maybe break before last `||`?

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/DynamicCompleterRegistry.java
```diff
@@ -38,60 +40,97 @@ public class DynamicCompleterRegistry {
...
+            return !conf.hasDisableOptions()
+
+                    || (conf.hasDisableOptions()
+                    && !cursorWord.equals(lastNotEmptyWord)
+                    && !conf.disableOptions().contains(lastNotEmptyWord))
+
+                    || (conf.hasDisableOptions()
+                    && cursorWord.equals(lastNotEmptyWord)
+                    && !conf.disableOptions().contains(lastNotEmptyWord)
+                    && !conf.disableOptions().contains(preLastNotEmptyWord));
```
valepakh, line 129: This expression has the same logic as the if statements above, so maybe we can replace one with the other so they are consistent?

modules/cli/src/test/java/org/apache/ignite/internal/cli/core/repl/completer/DynamicCompleterRegistryTest.java
```diff
@@ -36,6 +40,11 @@ class DynamicCompleterRegistryTest {
 
     DynamicCompleter completer3;
 
+    /** Makes reading more easy. */
```
valepakh, line 43:
```suggestion
    /** Makes reading easier. */
```

modules/cli/src/test/java/org/apache/ignite/internal/cli/core/repl/completer/DynamicCompleterRegistryTest.java
```diff
@@ -45,28 +54,30 @@ void setUp() {
...
 
     @Test
     void findsCompletersRegisteredWithStaticCommand() {
         // Given
-        registry.register(new String[]{"command1", "subcommand1"}, completer1);
-        registry.register(new String[]{"command1", "subcommand1"}, completer2);
-        registry.register(new String[]{"command2"}, completer3);
+        registry.register(CompleterConf.forCommand("command1", "subcommand1"), (words) -> completer1);
+        registry.register(CompleterConf.forCommand("command1", "subcommand1"), (words) -> completer2);
+        registry.register(CompleterConf.forCommand("command2"), (words) -> completer3);
```
valepakh, line 77:
```suggestion
        registry.register(CompleterConf.forCommand("command1", "subcommand1"), words -> completer1);
        registry.register(CompleterConf.forCommand("command1", "subcommand1"), words -> completer2);
        registry.register(CompleterConf.forCommand("command2"), words -> completer3);
```
And so on.

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/DynamicCompleterRegistry.java
PakhomovAlexander: This expression is about `disable options` and the above is about `enable options` and the logic statement is the opposite, actually. I will rewrite this expression with if-else chain to make them look similar and it will be easier to catch the difference.

valepakh (approved):

PR #1364, opened 2022-11-22
IGNITE-18212 Fix parsing error on GROUP BY with alias

korlov42:
Enable `isGroupByAlias` and `isGroupByOrdinal` in SQL conformance.

### DIFF
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItAggregatesTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItAggregatesTest.java
@@ -64,6 +64,18 @@ public class ItAggregatesTest extends AbstractBasicIntegrationTest {
                 .returns(15d, 1L)
                 .check();
 
+        // same query, but grouping by alias
+        assertQuery("select salary as sal, count(name) from person group by sal order by salary")
+                .returns(10d, 3L)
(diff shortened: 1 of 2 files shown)


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItAggregatesTest.java
```diff
@@ -64,6 +64,18 @@ public void countOfNonNumericField() {
                 .returns(15d, 1L)
                 .check();
 
+        // same query, but grouping by alias
+        assertQuery("select salary as sal, count(name) from person group by sal order by salary")
```
ptupitsyn, line 68: Can we also add a test for computed column grouping?

```sql
select (salary / 4 + 10) as sal, count(*) from person group by sal order by sal
```

ptupitsyn (changes requested):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItAggregatesTest.java
korlov42: Why not.

Have added such tests, but with a bit different expression. Still expression, though

zstan (approved):
LGTM

PR #1368, opened 2022-11-22
IGNITE-18230 Packaging: Add MSI installer for winget package

valepakh:
* Use the setupbuilder plugin to create MSI installer for the ignite3-db.
* Use `gradlew msi` to create an installer in the `packaging\db\build\distributions\ignite3-db-3.0.0.msi`.
* Update Micronaut since the old version doesn't work when installed in the path with spaces.

https://issues.apache.org/jira/browse/IGNITE-18230


Pochatkin:
LGTM

PR #1371, opened 2022-11-23
IGNITE-18240 Store logical topology as a single KV entry instead of a number of entries

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18240

### DIFF
--- a/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/RaftStorageManager.java
+++ b/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/RaftStorageManager.java
@@ -18,20 +18,25 @@
 package org.apache.ignite.internal.cluster.management.raft;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Collections.emptySet;
+import static java.util.function.Function.identity;
 import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toMap;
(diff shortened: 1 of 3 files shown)


modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/RaftStorageManager.java
```diff
@@ -42,16 +47,20 @@ class RaftStorageManager {
...
-    private static final byte[] LOGICAL_TOPOLOGY_PREFIX = "logical_".getBytes(UTF_8);
+    /** Storage key for the logical topology. */
+    private static final byte[] LOGICAL_TOPOLOGY_KEY = "logical".getBytes(UTF_8);
 
     /** Prefix for validation tokens. */
     private static final byte[] VALIDATED_NODE_PREFIX = "validation_".getBytes(UTF_8);
 
     private final ClusterStateStorage storage;
 
+    private volatile LogicalTopology currentLogicalTopology;
```
SammyVimes, line 58: maybe pre-set it with INITIAL?

SammyVimes (changes requested):

SammyVimes (approved):
LGTM

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/RaftStorageManager.java
rpuch: We initialize it in the constructor, so this seems excessive

PR #1381, opened 2022-11-28
IGNITE-18205 Sql. Allow hash output distribution for aggregations

korlov42:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/agg/IgniteColocatedSortAggregate.java
```diff
@@ -39,15 +39,15 @@
  * IgniteSingleSortAggregate.
```
ygerzhedovich, line 39: rename

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/agg/IgniteColocatedAggregateBase.java
```diff
@@ -36,9 +38,9 @@
  * IgniteSingleAggregateBase.
```
ygerzhedovich, line 38: rename

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/TableDescriptorImpl.java
```diff
@@ -127,7 +135,7 @@ public RelDataType deleteRowType(IgniteTypeFactory factory) {
     /** {@inheritDoc} */
     @Override
     public IgniteDistribution distribution() {
-        return IgniteDistributions.random();
+        return IgniteDistributions.affinity(colocationColumns, 0, dummyAffinityIdentity);
```
ygerzhedovich, line 138: why we use 0 here?

korlov42: I've added comment explaining such choice. Also check please comment in beginning of the class

ygerzhedovich (approved):

PR #1382, opened 2022-11-28
IGNITE-18147 Disable environment lookup for micronaut

PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-18147

Micronaut used to look up the environment. It caused dns resolving and if it was slow, then the startup of the application was slow. Now this behaviour is disabled by default.


rpuch (approved):
LGTM

PR #1384, opened 2022-11-28
IGNITE-18261 Use RAFT learners to get updates of logical topology on non-CMG nodes

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18261


modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/CmgRaftService.java
```diff
@@ -206,4 +216,33 @@ private ClusterNodeMessage nodeMessage(ClusterNode node) {
...
     }
+
+    /**
+     * Issues changePeersAsync request with same peers; learners are recalculated based on the current peers (which is same as
+     * CMG nodes) and known logical topology. Any node in the logical topology that is not a CMG node constitutes a learner.
+     *
+     * @param term RAFT term in which we operate (used to avoid races when changing peers/learners).
+     * @return Future that completes when the request is processed.
+     */
+    public CompletionStage<Void> updateLearners(long term) {
```
sashapolo, line 227: I'd prefer using `CompletableFuture`

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/CmgRaftService.java
```diff
@@ -206,4 +216,33 @@ private ClusterNodeMessage nodeMessage(ClusterNode node) {
...
+        List<Peer> currentPeers = raftService.peers();
+
+        assert currentPeers != null : "Raft service is not yet initialized";
+
+        Set<String> peersConsistentIds = currentPeers.stream()
+                .map(Peer::consistentId)
+                .collect(toSet());
+
+        Set<String> consistentIds = logicalTopology.getLogicalTopology().stream()
```
sashapolo, line 236: I would suggest replacing this code with:
```
Set<Peer> newLearners = logicalTopology.getLogicalTopology().stream()
        .map(ClusterNode::name)
        .filter(name -> !peersConsistentIds.contains(name))
        .map(Peer::new)
        .collect(toSet());
```

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/CmgRaftService.java
```diff
@@ -206,4 +216,33 @@ private ClusterNodeMessage nodeMessage(ClusterNode node) {
                 .port(node.address().port())
                 .build();
     }
+
+    /**
+     * Issues changePeersAsync request with same peers; learners are recalculated based on the current peers (which is same as
```
sashapolo, line 221:
```suggestion
     * Issues {@code changePeersAsync} request with same peers; learners are recalculated based on the current peers (which is same as
```

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/CmgRaftService.java
```diff
@@ -53,9 +57,15 @@ public class CmgRaftService {
 
     private final ClusterService clusterService;
 
-    public CmgRaftService(RaftGroupService raftService, ClusterService clusterService) {
+    private final LogicalTopology logicalTopology;
+
+    /**
+     * Creates a new instance.
```
sashapolo, line 63: Is this javadoc really necessary?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -466,27 +468,51 @@ private CompletableFuture<CmgRaftService> joinCluster(CmgRaftService service, Cl
      * Starts the CMG Raft service using the provided node names as its peers.
      */
     private CompletableFuture<CmgRaftService> startCmgRaftService(Collection<String> nodeNames) {
+        String thisNodeConsistentId = clusterService.topologyService().localMember().name();
+
+        // If we are not in the CMG, we must be a lerner. List of learners will be updated by a leader accordingly,
```
sashapolo, line 473:
```suggestion
        // If we are not in the CMG, we must be a learner. List of learners will be updated by a leader accordingly,
```

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -466,27 +468,51 @@ private CompletableFuture<CmgRaftService> joinCluster(CmgRaftService service, Cl
...
                     )
-                    .thenApply(service -> new CmgRaftService(service, clusterService));
+                    .thenApply(service -> new CmgRaftService(service, clusterService, logicalTopology));
         } catch (Exception e) {
             return failedFuture(e);
         }
     }
 
+    private void onLogicalTopologyChanged(long term) {
+        CompletableFuture<CmgRaftService> serviceFuture = raftService;
```
sashapolo, line 495: Hm, should we take the `raftServiceLock` here?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/CmgRaftService.java
rpuch: Checkstyle fails the build without it

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
rpuch: Good catch. Added synchronization

sashapolo (approved):

sashapolo (approved):

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/ActionOnTerm.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.cluster.management.raft;
+
+/**
+ * Action executed with a given RAFT term.
+ */
+@FunctionalInterface
+public interface ActionOnTerm {
```
tkalkirill, line 24: It seems that either there is not enough description, or it is enough to use **java.util.function.LongConsumer**

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/CmgRaftService.java
```diff
@@ -206,4 +215,29 @@ private ClusterNodeMessage nodeMessage(ClusterNode node) {
...
+
+    /**
+     * Issues {@code changePeersAsync} request with same peers; learners are recalculated based on the current peers (which is same as
+     * CMG nodes) and known logical topology. Any node in the logical topology that is not a CMG node constitutes a learner.
+     *
+     * @param term RAFT term in which we operate (used to avoid races when changing peers/learners).
+     * @return Future that completes when the request is processed.
+     */
+    public CompletableFuture<Void> updateLearners(long term) {
+        List<Peer> currentPeers = raftService.peers();
```
tkalkirill, line 226: Is it possible to use LinkedHashSet instead of a List?

rpuch: `raftService.peers()` returns a `List`. We later pass `currentPeers` to `changePeersAsync()`, which just iterates over it, so a `List` is fine there. And in line 230, we just iterate over this collection.

Why would we want to build a `LinkedHashSet` from it?

tkalkirill: Understood, ok

PR #1388, opened 2022-11-29
IGNITE-18276 Return topology version in CMGManager.logicalTopology()

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18276


SammyVimes (approved):
LGTM

PR #1391, opened 2022-11-29
IGNITE-18117 DistributionZoneManager populated with MetaStorage listeners to distribution zone events

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-18117


modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZonesUtil.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.ignite.internal.metastorage.client.CompoundCondition;
+import org.apache.ignite.internal.util.ByteUtils;
+import org.apache.ignite.lang.ByteArray;
+
+/**
+ * Util class for Distribution Zones flow.
+ */
+public class DistributionZonesUtil {
+    /** Key prefix for zone's data nodes. */
+    private static final String DISTRIBUTION_ZONE_DATA_NODES_PREFIX = "distributionZone.dataNodes";
```
sergeyuttsel, line 33: Need to add dot at the end like other prefixes: "distributionZone.dataNodes."

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
```diff
@@ -0,0 +1,364 @@
...
+        distributionZoneManager.createZone(new DistributionZoneConfigurationParameters.Builder(ZONE_NAME).build());
+
+        assertDataNodesForZone(1, clusterNodes);
+
+        keyValueStorage.put(zonesChangeTriggerKey().bytes(), ByteUtils.longToBytes(100));
+
+        distributionZoneManager.dropZone(ZONE_NAME);
+
+        verify(keyValueStorage, timeout(1000).times(2)).invoke(any());
+    }
```
sergeyuttsel, line 337: Where is the check that meta storage key has not removed?

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
```diff
@@ -0,0 +1,364 @@
...
+import org.jetbrains.annotations.Nullable;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mock;
+
+/**
+ * Tests distribution zones configuration changes and reaction to that changes.
+ */
+public class DistributionZoneManagerConfigurationChangesTest extends IgniteAbstractTest {
```
sergeyuttsel, line 73: I think we need to check a scenario like this:
1. create(zone1)
2. assertDataNodesForZone(zone1)
3. assertZonesChangeTriggerKey(1)
4. rename(zone1 -> zone2)
5. assertDataNodesForZone(zone1)
6. assertDataNodesForZone(zone2)
7. assertZonesChangeTriggerKey(2)

alievmirza: I didn't get the idea, renaming do not change any metastorage key representation, `assertDataNodesForZone` method accepts zone Id, which doesn't change during renaming. Also renaming do not change trigger key

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
alievmirza: thanks, fixed

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
sergeyuttsel: Ok. I agree with you :)

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -197,12 +235,114 @@ public CompletableFuture<Void> dropZone(String name) {
...
+     * @param zoneId Unique id of a zone
+     * @param revision Revision of an event that has triggered this method.
+     */
+    private void updateMetaStorageOnZoneCreateOrUpdate(int zoneId, long revision) {
+        if (!busyLock.enterBusy()) {
+            throw new IgniteInternalException(NODE_STOPPING_ERR, new NodeStoppingException());
+        }
+
+        try {
+            byte[] logicalTopologyBytes;
```
sanpwc, line 283: Let's declare logicalTopologyBytes closer to a place we are using it.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -197,12 +235,114 @@ public CompletableFuture<Void> dropZone(String name) {
...
+            // so we do not react on a stale events
+            CompoundCondition triggerKeyCondition = triggerKeyCondition(revision);
+
+            Set<String> nodesConsistentIds = clusterNodes.stream().map(ClusterNode::name).collect(Collectors.toSet());
+
+            logicalTopologyBytes = ByteUtils.toBytes(nodesConsistentIds);
+
+            Update dataNodesAndTriggerKeyUpd = updateDataNodesAndTriggerKey(zoneId, revision, logicalTopologyBytes);
+
+            var iif = If.iif(triggerKeyCondition, dataNodesAndTriggerKeyUpd, ops().yield(false));
```
sanpwc, line 304: AFAIK It's not valid to use var here.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -197,12 +235,114 @@ public CompletableFuture<Void> dropZone(String name) {
...
+
+            metaStorageManager.invoke(iif).thenAccept(res -> {
+                if (res.getAsBoolean()) {
+                    LOG.info("Update zones' dataNodes value [zoneId = {}, dataNodes = {}", zoneId, nodesConsistentIds);
+                } else {
+                    LOG.info("Failed to update zones' dataNodes value [zoneId = {}]", zoneId);
```
sanpwc, line 310: That'll be confusing, we'll have lots of such messages because multiple nodes will try to update dataNodes and only one will succeed.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -197,12 +235,114 @@ public CompletableFuture<Void> dropZone(String name) {
...
+        @Override
+        public CompletableFuture<?> onDelete(ConfigurationNotificationEvent<DistributionZoneView> ctx) {
+            updateMetaStorageOnZoneDelete(ctx.oldValue().zoneId(), ctx.storageRevision());
+
+            return completedFuture(null);
+        }
+
+        @Override
+        public CompletableFuture<?> onUpdate(ConfigurationNotificationEvent<DistributionZoneView> ctx) {
+            updateMetaStorageOnZoneCreateOrUpdate(ctx.newValue().zoneId(), ctx.storageRevision());
```
sanpwc, line 264: That'll be a bit incorrect, in case of zone updates it's requited to recalculate dataNodes only within following scenarios:

- In case of filter change.
- In case of timer changes if there are active timers. (re-scheduling)
In all other cases there's no need to recalculate dataNodes, however it's required to invoke triggerRevision immediately.

sanpwc (changes requested):

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
alievmirza: As far as I can understand, if got here, that means that any of `dataNodesAutoAdjust()`, `dataNodesAutoAdjustScaleUp()`, `dataNodesAutoAdjustScaleDown()` has been changed, so there is no "other cases". Case with renaming will be handled by separate handler `onRename`

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
alievmirza: Changed to debig

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
alievmirza: fixed

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -197,12 +235,114 @@ public CompletableFuture<Void> dropZone(String name) {
     /** {@inheritDoc} */
     @Override
     public void start() {
-
```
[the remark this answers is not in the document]
alievmirza: thanks!

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -197,12 +234,145 @@ public CompletableFuture<Void> dropZone(String name) {
...
+     *
+     * @param revision Revision of an event that has triggered this method.
+     */
+    private void updateMetaStorageOnZoneUpdate(long revision) {
+        if (!busyLock.enterBusy()) {
+            throw new IgniteInternalException(NODE_STOPPING_ERR, new NodeStoppingException());
+        }
+
+        try {
+            // Update data nodes for a zone only if the revision of the event is newer than value in that trigger key,
```
sanpwc, line 328: Currently that doesn't make sense, because we never update data nodes.

sanpwc (approved):

PR #1404, opened 2022-12-02
IGNITE-18097 CLI should check if it's already connected before trying to connect

valepakh:
https://issues.apache.org/jira/browse/IGNITE-18097

When trying to connect again, we either tell the user that he's already connected to the same node or asks if he really wants to connect to the different node.

### DIFF
--- a/modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/connect/ItConnectCommandTest.java
+++ b/modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/connect/ItConnectCommandTest.java
@@ -112,18 +112,23 @@ class ItConnectCommandTest extends CliCommandTestInitializedIntegrationBase {
     void connectTwice() {
         // Given connected to cluster
         execute("connect");
+        // And output is
+        assertAll(
+                this::assertErrOutputIsEmpty,
+                () -> assertOutputIs("Connected to http://localhost:10300" + System.lineSeparator())
+        );
         // And prompt is
         String promptBefore = Ansi.OFF.string(promptProvider.getPrompt());
         assertThat(promptBefore).isEqualTo("[" + nodeName() + "]> ");
 
         // When connect again
+        resetOutput();
         execute("connect");
 
         // Then
         assertAll(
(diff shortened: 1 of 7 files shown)


Pochatkin:
LGTM

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/connect/ItConnectCommandTest.java
```diff
@@ -107,6 +107,29 @@ void disconnect() {
...
+        assertThat(promptBefore).isEqualTo("[" + nodeName() + "]> ");
+
+        // When connect again
+        execute("connect");
+
+        // Then
+        assertAll(
+                this::assertErrOutputIsEmpty,
+                () -> assertOutputIs("Connected to http://localhost:10300" + System.lineSeparator()
+                        + "You are already connected to http://localhost:10300" + System.lineSeparator())
```
PakhomovAlexander, line 126: I think it is enough just to say `You are already connected to http://localhost:10300` without the first line.

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/questions/ItConnectToClusterTest.java
```diff
@@ -132,6 +138,33 @@ void connectOnStartAndSave() throws IOException {
...
+        bindAnswers("y");
+
+        // When connect to different URL
+        execute("connect", "http://localhost:10301");
+
+        // Then
+        assertAll(
+                this::assertErrOutputIsEmpty,
+                () -> assertOutputIs("Connected to http://localhost:10300" + System.lineSeparator()
+                + "Connected to http://localhost:10301" + System.lineSeparator())
```
PakhomovAlexander, line 161: That looks weird... Say I am connected to http://localhost:10300 and want to switch to another node and type `connect http://localhost:10301` and see 

```
Connected to http://localhost:10300
Connected to http://localhost:10301
```

It looks like a bug for the user. I suggest dropping the first line.

PakhomovAlexander (approved):

PR #1405, opened 2022-12-03
IGNITE-18118 Support group configuration in persistent pagememory-based MvPartitionStorage

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18118


modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/io/AbstractDataPageIo.java
```diff
@@ -1438,6 +1438,22 @@ protected abstract void writeRowData(
...
 
+
+    /**
+     * Narrows an {@code int} down to {@code short} throwing an exception if the value cannot be exactly represented as a {@code short}.
+     *
+     * @param intValue Value to narrow down.
+     * @return Resulting short value.
+     * @throws IllegalArgumentException If the provided value does not fit the {@code short} range.
+     */
+    protected static short narrowIntToShort(int intValue) {
```
tkalkirill, line 1449: Similar check is already in `org.apache.ignite.internal.pagememory.io.AbstractDataPageIo#addRow`, we can expand the existing one if necessary.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/PartitionMeta.java
```diff
@@ -50,7 +50,7 @@ public class PartitionMeta {
 
     private volatile long lastAppliedTerm;
 
-    private volatile byte @Nullable [] lastGroupConfig;
+    private volatile long lastGroupConfigLink;
```
tkalkirill, line 53: Please indicate that this is a raft group configuration.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/io/PartitionMetaIo.java
```diff
@@ -101,15 +102,15 @@ public void setLastAppliedTerm(long pageAddr, long lastAppliedTerm) {
...
     /**
-     * Sets last group config.
+     * Sets link to blob representing last group config.
      *
      * @param pageAddr Page address.
-     * @param lastGroupConfig Byte representation of last group config..
+     * @param lastGroupConfigLink Link to blob representing last group config..
      */
-    public void setLastGroupConfig(long pageAddr, byte @Nullable [] lastGroupConfig) {
+    public void setLastGroupConfig(long pageAddr, long lastGroupConfigLink) {
```
tkalkirill, line 110: Maybe `setLastGroupConfigLink` ?
Please add into toString

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/io/PartitionMetaIo.java
```diff
@@ -36,7 +35,9 @@ public class PartitionMetaIo extends PageIo {
 
     private static final int LAST_APPLIED_TERM_OFF = LAST_APPLIED_INDEX_OFF + Long.BYTES;
 
-    private static final int ROW_VERSION_FREE_LIST_ROOT_PAGE_ID_OFF = LAST_APPLIED_TERM_OFF + Long.BYTES;
+    private static final int LAST_GROUP_CONFIG_LINK_OFF = LAST_APPLIED_TERM_OFF + Long.BYTES;
+
+    private static final int ROW_VERSION_FREE_LIST_ROOT_PAGE_ID_OFF = LAST_GROUP_CONFIG_LINK_OFF + Long.BYTES;
```
tkalkirill, line 40: `blobFreeListRootPageId` skipped?

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/PartitionMeta.java
```diff
@@ -327,7 +348,7 @@ private PartitionMetaSnapshot(@Nullable UUID checkpointId, PartitionMeta partiti
             this.checkpointId = checkpointId;
             lastAppliedIndex = partitionMeta.lastAppliedIndex;
             lastAppliedTerm = partitionMeta.lastAppliedTerm;
-            lastGroupConfig = partitionMeta.lastGroupConfig;
+            lastGroupConfigLink = partitionMeta.lastGroupConfigLink;
```
tkalkirill, line 351: `blobFreeListRootPageId` skipped?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
```diff
@@ -152,7 +152,7 @@ protected static TestValue value(BinaryRow binaryRow) {
         return new IgniteBiTuple<>(key(binaryRow), value(binaryRow));
     }
 
-    protected static List<IgniteBiTuple<TestKey, TestValue>> toList(Cursor<ReadResult> cursor) throws Exception {
+    protected static List<IgniteBiTuple<TestKey, TestValue>> drainToList(Cursor<ReadResult> cursor) throws Exception {
```
tkalkirill, line 155: Why "drain"? method does not throw `Exception`.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/PersistentPageMemoryTableStorage.java
```diff
@@ -438,6 +442,52 @@ private IndexMetaTree createIndexMetaTree(
...
+                meta.blobFreeListRootPageId(lastCheckpointId(), rootPageId);
+
+                initNew = true;
+            }
+
+            return new BlobFreeList(
+                    tableView.tableId(),
+                    partId,
+                    dataRegion.pageMemory(),
+                    null,
```
tkalkirill, line 475: Why is the `reuseList` empty?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/PersistentPageMemoryTableStorage.java
```diff
@@ -438,6 +442,52 @@ private IndexMetaTree createIndexMetaTree(
...
+    /**
+     * Returns new {@link BlobFreeList} instance for partition.
+     *
+     * @param tableView Table configuration.
+     * @param partId Partition ID.
+     * @param pageMemory Persistent page memory instance.
+     * @param meta Partition metadata.
+     * @throws StorageException If failed.
+     */
+    private BlobFreeList createBlobFreeList(
```
tkalkirill, line 454: Technically, we can use `RowVersionFreeList` without creating an additional `FreeList`.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PersistentPageMemoryMvPartitionStorage.java
```diff
@@ -62,6 +68,15 @@ public class PersistentPageMemoryMvPartitionStorage extends AbstractPageMemoryMv
     /** Checkpoint listener. */
     private final CheckpointListener checkpointListener;
 
+    /** FreeList for blobs. */
+    private final BlobFreeList blobFreeList;
+
+    /** Used to read data from pagememory pages. */
+    private final DataPageReader pageReader;
+
+    /** Lock that protects group config read/write. */
```
tkalkirill, line 77: Please indicate that this is a raft group configuration.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PersistentPageMemoryMvPartitionStorage.java
```diff
@@ -215,7 +242,25 @@ public RaftGroupConfiguration committedGroupConfiguration() {
         }
 
         try {
-            return groupConfigFromBytes(meta.lastGroupConfig());
+            groupConfigReadWriteLock.readLock().lock();
```
tkalkirill, line 245: Forgot
```
if (!closeBusyLock.enterBusy()) {
            throwStorageClosedException();
        }
```

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/io/AbstractDataPageIo.java
rpuch: If I'm not mistaken, the check in `AbstractDataPageIo#addRow()` is made with `assert` keyword, so it might absent in a production environment, hence it cannot be relied upon. I'd prefer to either leave the check that I added or change the check in `addRow()` from `assert` to throwing an exception.

WDYT?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PersistentPageMemoryMvPartitionStorage.java
rpuch: We already have busy lock taken, in line 240. Isn't this enough?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
rpuch: Removed the `throws` declaration.

I renamed the method so that we can statically import `Collectors.toList`.

There is `BlockingQueue.drainTo()` that drains the queue to a collection. `drainToList` follows the same logic.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/io/AbstractDataPageIo.java
tkalkirill: I think the assertion should be enough and check all this in tests. 
In production code, this check is not needed.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
tkalkirill: Ok

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PersistentPageMemoryMvPartitionStorage.java
tkalkirill: my fault

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/io/AbstractDataPageIo.java
rpuch: Can a user configure a really large page, like 1Mb? If yes, then free space in a page might be around 1Mb, which is way more than 64kb. If the user does not have `-ea` (because they do this in production), they will get strange behavior instead of a proper exception.

Assertions should be used only for catching programmer errors (i.e. our errors). They should not be used to protect against a possible misconfiguration. Also, tests would not help in such scenario.

ibessonov: Current limit is 32k I believe

tkalkirill: I do not think that users will use our product immediately in production without test runs on test environments with -ea.

rpuch: Removed the additional check. But this should not be forgotten later.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/BlobStorage.java
```diff
@@ -0,0 +1,341 @@
...
+            int newBytesOffset = state.bytesOffset + fragmentLength;
+
+            if (newBytesOffset < state.bytes.length) {
+                long nextPageId = blobIo.getNextPageId(pageAddr);
+
+                assert nextPageId != NO_PAGE_ID;
+
+                state.nextPageId = nextPageId;
+            } else {
+                state.nextPageId = NO_PAGE_ID;
```
tkalkirill, line 253: I think it's worth making sure that there is no nextPageId in BlobIo.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/BlobStorage.java
```diff
@@ -0,0 +1,341 @@
...
+     *
+     * @param firstPageId ID of the first page in the chain storing the blob.
+     * @param bytes New blob content.
+     * @throws IgniteInternalCheckedException If something goes wrong.
+     */
+    public void updateBlob(long firstPageId, byte[] bytes) throws IgniteInternalCheckedException {
+        doStore(firstPageId, bytes);
+    }
+
+    private long doStore(long maybeFirstPageId, byte[] bytes) throws IgniteInternalCheckedException {
```
tkalkirill, line 130: What about the fact that we always free old pages and create new ones at the beginning? seems like a simpler option.

modules/storage-page-memory/src/test/java/org/apache/ignite/internal/storage/pagememory/mv/PersistentPageMemoryMvPartitionStorageTest.java
```diff
@@ -105,7 +121,94 @@ void testReadAfterRestart() throws Exception {
...
+                List.of("old-learner1", "old-learner2")
+        );
+
+        storage.runConsistently(() -> {
+            storage.committedGroupConfiguration(originalConfigOfMoreThanOnePage);
+
+            return null;
+        });
+
+        long freeSpaceBeforeTrimming = pageMemStorage.rowVersionFreeList.freeSpace();
```
tkalkirill, line 193: unused

modules/storage-page-memory/src/test/java/org/apache/ignite/internal/storage/pagememory/mv/PersistentPageMemoryMvPartitionStorageTest.java
```diff
@@ -105,7 +121,94 @@ void testReadAfterRestart() throws Exception {
...
+        restartStorage();
+
+        RaftGroupConfiguration readConfig = storage.committedGroupConfiguration();
+
+        assertThat(readConfig, is(equalTo(originalConfig)));
+    }
+
+    @Test
+    void groupConfigPagesAreRecycled() throws Exception {
+        List<String> oneMbOfPeers = IntStream.range(0, 100_000)
```
tkalkirill, line 176: I think it's worth making sure the size of these `Strings` is greater than at least 2 pages.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/io/PartitionMetaIo.java
```diff
@@ -251,6 +253,7 @@ protected void printPage(long addr, int pageSize, IgniteStringBuilder sb) {
         sb.app("TablePartitionMeta [").nl()
                 .app("lastAppliedIndex=").app(getLastAppliedIndex(addr)).nl()
                 .app("lastAppliedTerm=").app(getLastAppliedTerm(addr)).nl()
+                .app("lastRaftGroupConfigFirstPageId=").app(getLastRaftGroupConfigFirstPageId(addr)).nl()
```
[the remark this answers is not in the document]
rpuch: Why do we need commas if we already add newlines?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/BlobStorage.java
rpuch: This looks like a suggestion and not a fact :)

I wanted to avoid unneeded reallocation.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/io/PartitionMetaIo.java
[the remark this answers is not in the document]
tkalkirill: Just a general style for pages
org.apache.ignite.internal.pagememory.tree.io.BplusIo#printPage
org.apache.ignite.internal.pagememory.tree.io.BplusMetaIo#printPage
etc

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/BlobStorage.java
tkalkirill: How often will we update the configuration?

ibessonov: I agree that we should reuse old pages, it's simpler actually. And I believe that for the most part, the amount of pages will stay the same.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PersistentPageMemoryMvPartitionStorage.java
```diff
@@ -226,22 +262,37 @@ public void committedGroupConfiguration(RaftGroupConfiguration config) {
...
+        } catch (IgniteInternalCheckedException e) {
+            throw new StorageException("Cannot save committed group configuration, groupId=" + groupId + ", partitionId=" + groupId, e);
+        } finally {
+            raftGroupConfigReadWriteLock.writeLock().unlock();
+        }
     }
 
     @Nullable
-    private static RaftGroupConfiguration groupConfigFromBytes(byte @Nullable [] bytes) {
+    private static RaftGroupConfiguration raftGroupConfigFromBytes(byte @Nullable [] bytes) {
```
ibessonov, line 287: By the way, why do we have to mention RAFT here? There's a chance of us having more replication protocol in the future, can we try and abstract ourselves a little bit? At least stop renaming old methods into a new that mention RAFT even more.

It would be a good idea to rename RaftGroupConfiguration as well. Or even better. I don't understand why it's a storages' task to convert an object into a byte array. Looks like a wrong design decision.

ibessonov (commented):
I guess my review wasn't as thorough, but I had some comments. Please take a look

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/io/PartitionMetaIo.java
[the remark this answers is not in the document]
rpuch: Is this general style justified by something? If I'm not mistaken, then using this 'commas and NLs' style would produce output like the following:

```
field1=val1
,field2=val2
,field3=val3
```

while 'just NLs' style will produce

```
field1=val1
field2=val2
field3=val3
```

To me, second result seems more readable (there is less noise caused by commas, and this noise is not needed).

Why would we prefer first result to the second one?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/BlobStorage.java
rpuch: We'll update RAFT config on each configuration change, this should happen rarely. But it won't hurt to make the mechanism a bit more generic (it doesn't cost too much in terms of complexity).

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/freelist/io/IndexColumnsDataIo.java
```diff
@@ -48,7 +48,7 @@ protected IndexColumnsDataIo(int ver) {
     protected void writeRowData(long pageAddr, int dataOff, int payloadSize, IndexColumns row, boolean newRow) {
         assertPageType(pageAddr);
 
-        putShort(pageAddr, dataOff, (short) payloadSize);
+        putShort(pageAddr, dataOff, narrowIntToShort(payloadSize));
```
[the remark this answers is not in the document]
rpuch: I already reverted the change

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/BlobStorage.java
```diff
@@ -0,0 +1,341 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Used to store a limited number of blobs (just byte arrays) per partition. Each blob is stored in a sequence
+ * of pages forming a linked list (a previous page links to the next one).
+ *
+ * <p>If a lot of blobs (comparable with the number of rows) needs to be stored in a partition, another mechanism
+ * (probably using a {@link org.apache.ignite.internal.pagememory.freelist.FreeList}) should be used.
+ */
+public class BlobStorage {
```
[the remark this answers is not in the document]
rpuch: The main reason is that I didn't know that I would benefit from extending `DataStructure`. I reused page recycling from there, what other conveniences would be useful here?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/BlobStorage.java
tkalkirill: Ok

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PersistentPageMemoryMvPartitionStorage.java
rpuch: I added a ticket https://issues.apache.org/jira/browse/IGNITE-18405 because renaming this class is out of scope of this PR.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/persistence/io/PartitionMetaIo.java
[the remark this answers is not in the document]
rpuch: I added a ticket https://issues.apache.org/jira/browse/IGNITE-18407

Here, it seems better to leave it as is (just newlines, no commas) as we know consider 'commas+newlines' a tach debt.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PersistentPageMemoryMvPartitionStorage.java
rpuch: Also added https://issues.apache.org/jira/browse/IGNITE-18408 about storing bytes, not config object

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/BlobStorage.java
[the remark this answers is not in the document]
ibessonov: Methods for allocating pages, that you pretty much copy-pasted unwillingly. I think they could be reused as well

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/PersistentPageMemoryMvPartitionStorage.java
ibessonov: Thank you!

rpuch: Renamed internal fields/constants/methods (not relating to the API)

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/io/BlobFragmentIo.java
```diff
@@ -0,0 +1,130 @@
...
+ * Then, if the page is the first in the chain, total blob length is stored as 4 bytes.
+ * Finally, bytes of a blob fragment are stored.
+ */
+public class BlobFragmentIo extends PageIo {
+    /** Page IO type. */
+    public static final short T_BLOB_FRAGMENT_IO = 13;
+
+    private static final int NEXT_PAGE_ID_OFF = PageIo.COMMON_HEADER_END;
+
+    private static final int FRAGMENT_BYTES_OFF = NEXT_PAGE_ID_OFF + Long.BYTES;
```
tkalkirill, line 44: I think it's better to use `org.apache.ignite.internal.pagememory.util.PartitionlessLinks`. 
I mean not 8 bytes but 6.

It confuses me a little that this is a constant, let's make it calculated, and in the documentation we will present the layout of the page, for example, as in `org.apache.ignite.internal.storage.pagememory.index.sorted.io.SortedIndexTreeIo`.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/io/BlobFragmentIo.java
```diff
@@ -0,0 +1,130 @@
...
+import static org.apache.ignite.internal.pagememory.util.PageUtils.putInt;
+import static org.apache.ignite.internal.pagememory.util.PageUtils.putLong;
+
+import org.apache.ignite.internal.pagememory.io.IoVersions;
+import org.apache.ignite.internal.pagememory.io.PageIo;
+import org.apache.ignite.internal.pagememory.util.PageUtils;
+import org.apache.ignite.lang.IgniteStringBuilder;
+
+/**
+ * Pages IO for blob fragments.
```
tkalkirill, line 32: I think you should add a description something like:
`Pages IO for storing a blob as a chain of its fragments.`

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/BlobStorage.java
```diff
@@ -0,0 +1,329 @@
...
+
+            assert ok : state.pageId;
+        } while (!state.stop);
+
+        freePagesStartingWith(state.firstPageToFreeId);
+
+        return firstPageId;
+    }
+
+    private long allocatePageIfNeeded(long maybePageId) throws IgniteInternalCheckedException {
```
tkalkirill, line 151: I think you can use `org.apache.ignite.internal.pagememory.datastructure.DataStructure#allocatePage(org.apache.ignite.internal.pagememory.reuse.ReuseBag)` with `null`

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/BlobStorage.java
```diff
@@ -0,0 +1,329 @@
...
+            BlobFragmentIo blobIo = (BlobFragmentIo) io;
+
+            int capacityForBytes = blobIo.getCapacityForFragmentBytes(pageSize(), state.isFirstPage());
+
+            int fragmentLength = Math.min(capacityForBytes, state.bytes.length - state.bytesOffset);
+
+            if (state.isFirstPage()) {
+                blobIo.setTotalLength(pageAddr, state.bytes.length);
+            }
+            blobIo.setFragmentBytes(pageAddr, state.isFirstPage(), state.bytes, state.bytesOffset, fragmentLength);
```
tkalkirill, line 287:
```suggestion
            
            blobIo.setFragmentBytes(pageAddr, state.isFirstPage(), state.bytes, state.bytesOffset, fragmentLength);
```

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/BlobStorage.java
```diff
@@ -0,0 +1,329 @@
...
+     * @param firstPageId ID of the first page in the chain storing the blob.
+     * @param bytes New blob content.
+     * @throws IgniteInternalCheckedException If something goes wrong.
+     */
+    public void updateBlob(long firstPageId, byte[] bytes) throws IgniteInternalCheckedException {
+        doStore(firstPageId, bytes);
+    }
+
+    private long doStore(long maybeFirstPageId, byte[] bytes) throws IgniteInternalCheckedException {
+        Objects.requireNonNull(bytes, "bytes is null");
```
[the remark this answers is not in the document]
rpuch: I'd like to check this precondition. A meaningful error message is a good thing.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/BlobStorage.java
rpuch: This is what I do

tkalkirill: Ok, `allocatePage()` I do not think that it is needed, it is used in one place.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/io/BlobFragmentIo.java
rpuch: Storing page IDs as 6 bytes will make the code more complex, but the gain is too little (2 bytes per page), so it does not seem worth the complexity.

I made the way we calculate the offset more clear and clarified the page layout even more in the javadoc.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/BlobStorage.java
rpuch: Inlined the method

tkalkirill (approved):

PR #1406, opened 2022-12-05
IGNITE-18319 Pass sender consistendId instead of ClusterNode to handler

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18319


modules/network-api/src/main/java/org/apache/ignite/network/MessagingService.java
```diff
@@ -65,6 +65,20 @@ public interface MessagingService {
      */
     CompletableFuture<Void> respond(ClusterNode recipient, NetworkMessage msg, long correlationId);
 
+    /**
+     * Sends a response to a {@link #invoke} request.
+     * Guarantees are the same as for the {@link #send(ClusterNode, NetworkMessage)}.
+     *
+     * <p>If the recipient cannot be resolved (because it has already left the physical topology), the returned future is resolved
+     * with the corresponding exception.
```
ibessonov, line 73: Can you please clarify the type of the exception? It's not clear right now, what should you handle

modules/network/src/main/java/org/apache/ignite/network/ConsistentIdNotResolvedException.java
```diff
@@ -0,0 +1,30 @@
...
+
+package org.apache.ignite.network;
+
+import org.apache.ignite.lang.IgniteException;
+
+/**
+ * Thrown when consistent ID cannot be resolved to a {@link ClusterNode} instance (i.e. when
+ * there is no node with such consistent ID in the physical topology).
+ */
+public class ConsistentIdNotResolvedException extends IgniteException {
```
ibessonov, line 26: Maybe we should create proper error group and id for new exceptions that we create, otherwise we'll never migrate anything

PR #1407, opened 2022-12-05
IGNITE-18322 Define scan contract SortedIndexStorage

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18322


modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
```diff
@@ -579,6 +585,401 @@ void testNullValues(ColumnDefinition columnDefinition) throws Exception {
...
+
+        assertEquals(0, serializer.deserializeColumns(nextRow)[0]);
+        assertEquals(rowId0, nextRow.rowId());
+
+        assertFalse(scan.hasNext());
+        assertThrows(NoSuchElementException.class, scan::next);
+
+        put(indexStorage, serializer.serializeRow(new Object[]{0}, rowId0));
+
+        assertFalse(scan.hasNext());
```
ibessonov, line 865: You know what, at each step it would be cool ta have a comment that shows the state of the index and the state of the cursor. Like `0 1 2 | 3 4` or something, otherwise it might be complicated to remember everything.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
```diff
@@ -579,6 +585,401 @@ void testNullValues(ColumnDefinition columnDefinition) throws Exception {
...
+
+        assertEquals(1, serializer.deserializeColumns(nextRow)[0]);
+        assertEquals(rowId0, nextRow.rowId());
+
+        remove(indexStorage, serializer.serializeRow(new Object[]{2}, rowId1));
+
+        assertFalse(scan.hasNext());
+        assertThrows(NoSuchElementException.class, scan::next);
+    }
+
```
ibessonov, line 982: We need more tests. What if current position is removed but the new entry is inserted into "the same" place? Or n the place strictly before, for example. Combinations of insertions and deletions should be checked, at least some of them.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
```diff
@@ -45,7 +44,9 @@
  * Test implementation of MV sorted index storage.
  */
 public class TestSortedIndexStorage implements SortedIndexStorage {
-    private final ConcurrentNavigableMap<ByteBuffer, Set<RowId>> index;
+    private static final Object NULL = new Object();
+
+    private final ConcurrentNavigableMap<ByteBuffer, NavigableMap<RowId, Object>> index;
```
ibessonov, line 49: I guess `NavigableSet` should be used instead of the map. What do you store in those values?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbSortedIndexStorage.java
```diff
@@ -167,24 +167,71 @@ private Cursor<ByteBuffer> createScanCursor(byte @Nullable [] lowerBound, byte @
...
+                    } catch (RocksDBException e) {
+                        throw new StorageException("Error refreshing an iterator", e);
+                    }
+
+                    if (key == null) {
+                        it.seek(lowerBound == null ? partitionStorage.partitionStartPrefix() : lowerBound);
+                    } else {
+                        it.seekForPrev(key);
+
+                        it.next();
```
ibessonov, line 219: Technically, iterator may not be valid in this case and calling `next` is unsafe. You should add a test where index has a single value and it's deleted while iterator is "standing" on it.
It'll most likely fail.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
```diff
@@ -208,4 +184,91 @@ private void checkClosed() {
...
+                if (indexMapEntry == null) {
+                    indexMapEntry = indexMap.firstEntry();
+                }
+
+                if (rowId == null) {
+                    if (indexMapEntry != null) {
+                        rowId = getRowId(indexMapEntry.getValue().firstEntry());
+                    }
+                } else {
+                    Entry<RowId, Object> rowIdEntry = indexMapEntry.getValue().higherEntry(rowId);
```
rpuch, line 246: Should this be named `nextRowIdEntry`?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
rpuch, line 49: It looks like this can still be `ConcurrentNavigableMap<ByteBuffer, NavigableSet<RowId>>` which is simpler to use and it causes less confusion (because something that is used as a set is called Set, not Map)

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/sorted/PageMemorySortedIndexStorage.java
```diff
@@ -272,4 +271,97 @@ public R next() {
...
+                return toIndexRowImpl(treeRow);
+            } catch (IgniteInternalCheckedException e) {
+                throw new StorageException("Error while advancing the cursor", e);
+            } finally {
+                closeBusyLock.leaveBusy();
+            }
+        }
+
+        private void advanceIfNeeded() throws IgniteInternalCheckedException {
+            if (hasNext == null) {
```
rpuch, line 341: Early return pattern could be used to decrease indentation level for the principal logic of the method

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
```diff
@@ -1572,6 +1572,34 @@ public final T findOne(L row) throws IgniteInternalCheckedException {
         return findOne(row, null, null);
     }
 
+    /**
+     * Searches for the next row from the passed from the arguments.
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
```diff
@@ -1572,6 +1572,34 @@ public final T findOne(L row) throws IgniteInternalCheckedException {
...
+     *
+     * @param row Row.
```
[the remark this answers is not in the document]
tkalkirill: Yes, we can use this name.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbSortedIndexStorage.java
```diff
@@ -167,24 +167,71 @@ private Cursor<ByteBuffer> createScanCursor(byte @Nullable [] lowerBound, byte @
...
+            @Nullable
+            private Boolean hasNext;
 
-        return new RocksIteratorAdapter<>(it) {
-            @Override
-            protected ByteBuffer decodeEntry(byte[] key, byte[] value) {
-                return ByteBuffer.wrap(key).order(ORDER);
-            }
+            @Nullable
+            private byte[] key;
```
rpuch, line 175:
```suggestion
            private byte @Nullable [] key;
```

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbSortedIndexStorage.java
```diff
@@ -167,24 +167,71 @@ private Cursor<ByteBuffer> createScanCursor(byte @Nullable [] lowerBound, byte @
...
 
             @Override
             public void close() {
-                super.close();
+                it.close();
```
rpuch, line 179: Let's include `it` into the call to `closeAll()`

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
```diff
@@ -6222,6 +6250,72 @@ public void close() {
...
+
+            int cnt = io.getCount(pageAddr);
+
+            if (cnt == 0) {
+                // Empty tree.
+                assert io.getForward(pageAddr, partId) == 0L;
+            } else {
+                assert io.isLeaf() : io;
+                assert cnt > 0 : cnt;
+                assert idx >= 0 || idx == -1 : idx;
```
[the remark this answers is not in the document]
tkalkirill: You're right, I looked at the code in neighboring classes and there was this.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
```diff
@@ -6222,6 +6250,72 @@ public void close() {
...
+                assert cnt >= idx : "cnt=" + cnt + ", idx=" + idx;
+
+                checkDestroyed();
+
+                if (idx == -1) {
+                    idx = findNextRowIdx(pageAddr, io, cnt);
+                }
+
+                if (cnt - idx > 0) {
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
```diff
@@ -579,6 +585,401 @@ void testNullValues(ColumnDefinition columnDefinition) throws Exception {
...
+            put(indexStorage, serializer.serializeRow(new Object[]{i}, new RowId(TEST_PARTITION)));
+        }
+
+        // Checking without borders.
+        assertThat(
+                scan(indexStorage, null, null, 0).stream().map(objects -> objects[0]).collect(toList()),
+                contains(0, 1, 2, 3, 4)
+        );
+
+        // Let's check without borders.
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
```diff
@@ -579,6 +585,401 @@ void testNullValues(ColumnDefinition columnDefinition) throws Exception {
...
+    void testScanContractForEmptyIndex() {
+        SortedIndexDefinition indexDefinition = SchemaBuilders.sortedIndex("TEST_IDX")
+                .addIndexColumn(ColumnType.INT32.typeSpec().name()).asc().done()
+                .build();
+
+        SortedIndexStorage indexStorage = createIndexStorage(indexDefinition);
+
+        BinaryTupleRowSerializer serializer = new BinaryTupleRowSerializer(indexStorage.indexDescriptor());
+
+        Cursor<IndexRow> scanBeforeHashNext = indexStorage.scan(null, null, 0);
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
```diff
@@ -6222,6 +6250,72 @@ public void close() {
...
+                    nextRow = getRow(io, pageAddr, idx);
+                }
+            }
+
+            return true;
+        }
+
+        int findNextRowIdx(long pageAddr, BplusIo<L> io, int cnt) throws IgniteInternalCheckedException {
+            // Compare with the first row on the page.
+            int cmp = compare(0, io, pageAddr, 0, row);
```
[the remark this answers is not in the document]
tkalkirill: We will never get into this code because the `idx` is always >=0, I deleted this code.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
```diff
@@ -579,6 +585,401 @@ void testNullValues(ColumnDefinition columnDefinition) throws Exception {
...
+
+        assertThat(
+                indexStorage.scan(
+                                serializer.serializeRowPrefix(0),
+                                serializer.serializeRowPrefix(4),
+                                (GREATER_OR_EQUAL | LESS)
+                        ).stream()
+                        .map(indexRow -> serializer.deserializeColumns(indexRow)[0])
+                        .collect(toList()),
+                contains(0, 1, 2, 3)
```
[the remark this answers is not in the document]
tkalkirill: In `org.hamcrest.Matchers#contains(E...)`, it is checked for equality by the number of elements, I looked at the documentation and tried it with my hands.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
```diff
@@ -579,6 +585,401 @@ void testNullValues(ColumnDefinition columnDefinition) throws Exception {
         }
     }
 
+    /**
+     * Checks simple scenarios for a scanning cursor.
+     */
+    @Test
+    void testScanSimple() {
```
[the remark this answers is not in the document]
tkalkirill: In general, it can be done, but since it is now clearer for me.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
```diff
@@ -579,6 +585,401 @@ void testNullValues(ColumnDefinition columnDefinition) throws Exception {
...
+        Cursor<IndexRow> scanWithoutHasNext = indexStorage.scan(null, null, 0);
+
+        put(indexStorage, serializer.serializeRow(new Object[]{0}, new RowId(TEST_PARTITION)));
+
+        assertTrue(scanBeforeHashNext.hasNext());
+        assertFalse(scanAfterHasNext.hasNext());
+
+        assertEquals(0, serializer.deserializeColumns(scanBeforeHashNext.next())[0]);
+
+        assertThrows(NoSuchElementException.class, scanAfterHasNext::next);
```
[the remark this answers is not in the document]
tkalkirill: Method `java.util.Iterator#hasNext` and `java.util.Iterator#next` should be consistent, I don't think it needs to be described in the documentation.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
```diff
@@ -579,6 +585,401 @@ void testNullValues(ColumnDefinition columnDefinition) throws Exception {
...
+
+        BinaryTupleRowSerializer serializer = new BinaryTupleRowSerializer(indexStorage.indexDescriptor());
+
+        for (int i = 0; i < 5; i++) {
+            put(indexStorage, serializer.serializeRow(new Object[]{i}, new RowId(TEST_PARTITION)));
+        }
+
+        // Checking without borders.
+        assertThat(
+                scan(indexStorage, null, null, 0).stream().map(objects -> objects[0]).collect(toList()),
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
```diff
@@ -579,6 +585,401 @@ void testNullValues(ColumnDefinition columnDefinition) throws Exception {
...
+                        ).stream()
+                        .map(indexRow -> serializer.deserializeColumns(indexRow)[0])
+                        .collect(toList()),
+                contains(0, 1, 2)
+        );
+    }
+
+    @Test
+    void testScanContractForEmptyIndex() {
+        SortedIndexDefinition indexDefinition = SchemaBuilders.sortedIndex("TEST_IDX")
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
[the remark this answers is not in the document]
tkalkirill: Divided into three methods, I think it should become clearer.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
tkalkirill: Fix it

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
tkalkirill: Fix it

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
tkalkirill: Main problem in `java.util.SortedSet#first` if the set is empty it will throw an exception.

tkalkirill: Main problem in `java.util.SortedSet#first` if the set is empty it will throw an exception.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
tkalkirill: Fix it

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/sorted/PageMemorySortedIndexStorage.java
tkalkirill: Fix it

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbSortedIndexStorage.java
tkalkirill: Fix it

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbSortedIndexStorage.java
tkalkirill: Why?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbSortedIndexStorage.java
tkalkirill: Tried to test in `org.apache.ignite.internal.storage.index.AbstractSortedIndexStorageTest#testScanContractRemoveCachedRow`

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
rpuch: Well, I changed the code to migrate it from maps to sets, and tests passed. This is probably because the set/map is never empty (if we remove the last element of it, we remove the set/map from the `index`). So the fact that `SortedSet#first()` throws for an empty set does not seem to be a problem.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbSortedIndexStorage.java
rpuch: To make sure that we close as many resources as possible even if an exception happens when closing one of them. This is probably the reason `closeAll()` exists :)

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
tkalkirill: If we look at `TestSortedIndexStorage#remove`, we can get into trouble if, after getting "RowId"s, in parallel delete first element (become empty RowIds) it and then try to take the first element.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbSortedIndexStorage.java
tkalkirill: Fix it

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
rpuch: Yes, you are right. But we can just wrap the call to `set.first()` with `try`/`catch NoSuchElementException`. This needs to be done in just 2 places, and we'll have a set represented by a Set, not by a Map. Still less confusion.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
```diff
@@ -1573,17 +1573,18 @@ public final T findOne(L row) throws IgniteInternalCheckedException {
     }
 
     /**
-     * Searches for the next row from the passed from the arguments.
+     * Searches for the lowerBound that (strictly or loosely, depending on {@code includeRow}) follows the lowerBound passed as an
```
rpuch, line 1576:
```suggestion
     * Searches for the row that (strictly or loosely, depending on {@code includeRow}) follows the lowerBound passed as an
```

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
```diff
@@ -1573,17 +1573,18 @@ public final T findOne(L row) throws IgniteInternalCheckedException {
...
+     * argument.
      *
-     * @param row Row.
-     * @param includeRow {@code True} if you include the passed row in the result.
-     * @return Next row.
+     * @param lowerBound Lower bound..
+     * @param includeRow {@code True} if you include the passed lowerBound in the result.
+     * @return Next lowerBound.
```
rpuch, line 1581:
```suggestion
     * @param includeRow {@code True} if you include the passed row in the result.
     * @return Next row.
```

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
```diff
@@ -1592,9 +1593,9 @@ public final T findOne(L row) throws IgniteInternalCheckedException {
         } catch (CorruptedDataStructureException e) {
             throw e;
         } catch (IgniteInternalCheckedException e) {
-            throw new IgniteInternalCheckedException("Runtime failure on lookup next row: " + row, e);
+            throw new IgniteInternalCheckedException("Runtime failure on lookup next lowerBound: " + lowerBound, e);
```
rpuch, line 1596:
```suggestion
            throw new IgniteInternalCheckedException("Runtime failure on lookup next row: " + lowerBound, e);
```

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
```diff
@@ -1592,9 +1593,9 @@ public final T findOne(L row) throws IgniteInternalCheckedException {
...
         } catch (RuntimeException | AssertionError e) {
-            throw corruptedTreeException("Runtime failure on lookup next row: " + row, e, grpId, g.pageId);
+            throw corruptedTreeException("Runtime failure on lookup next lowerBound: " + lowerBound, e, grpId, g.pageId);
```
rpuch, line 1598:
```suggestion
            throw corruptedTreeException("Runtime failure on lookup next row: " + lowerBound, e, grpId, g.pageId);
```

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
tkalkirill: It's not good to write code based on exceptions, it's better to leave the map.

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
tkalkirill: fix it

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
tkalkirill: fix it

modules/page-memory/src/main/java/org/apache/ignite/internal/pagememory/tree/BplusTree.java
tkalkirill: fix it

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
rpuch: It's ok to use any way to pass information. In this case, the library chose to use exceptions to this end, so we are ok to catch them. This is a little ugly, but it's a lot less ugly than using a map as a set.

If you are going to leave the map, please at least add a comment to the field to explain that the map is actually used to model a set and why it is done so.

rpuch (approved):

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
tkalkirill: You are right, I should write about it in the comments.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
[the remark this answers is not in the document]
ibessonov: Cool, I didn't realize that. Poor naming, in my opinion

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
```diff
@@ -604,148 +605,148 @@ void testScanSimple() {
...
-                                null,
-                                (GREATER | LESS)
-                        ).stream()
-                        .map(indexRow -> serializer.deserializeColumns(indexRow)[0])
-                        .collect(toList()),
+                scan(
+                        indexStorage,
+                        serializer.serializeRowPrefix(1),
+                        null,
+                        (GREATER | LESS),
```
ibessonov, line 696: Just a question, why did you add parentheses here?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbSortedIndexStorage.java
```diff
@@ -171,14 +172,15 @@ private Cursor<ByteBuffer> createScanCursor(byte @Nullable [] lowerBound, byte @
...
-            private byte[] key;
+            private byte @Nullable [] key;
 
             @Override
             public void close() {
-                it.close();
-
-                RocksUtils.closeAll(options, upperBoundSlice);
+                try {
+                    closeAll(it, () -> RocksUtils.closeAll(options, upperBoundSlice));
```
ibessonov, line 180: Wait a minute ,why not `closeAll(it, options, upperBoundSlice)`? What's the reason for creating a lambda?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
tkalkirill: For clarity

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbSortedIndexStorage.java
tkalkirill: Fix it

ibessonov (approved):

PR #1408, opened 2022-12-06
IGNITE-18233: Add a green color to the sql repl prompt

Flaugh24:
https://issues.apache.org/jira/browse/IGNITE-18233


modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/sql/SqlReplCommand.java
```diff
@@ -96,7 +99,7 @@ public void run() {
             // When passing white space to this command, picocli will treat it as a positional argument
             if (execOptions == null || (execOptions.command != null && execOptions.command.isBlank())) {
                 replExecutorProvider.get().execute(Repl.builder()
-                        .withPromptProvider(() -> "sql-cli> ")
+                        .withPromptProvider(() -> ansi(fg(Color.GREEN).mark("sql-cli > ")))
```
PakhomovAlexander, line 102: our default prompt looks like `[disconnected]>` so there is no space before `>`

PakhomovAlexander (approved):

PR #1411, opened 2022-12-06
IGNITE-17871 JDKMarshaller replaced with OptimizedMarshaller for RAFT commands.

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-17871


modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/counter/GetValueCommand.java
```diff
@@ -18,9 +18,14 @@
 package org.apache.ignite.raft.server.counter;
 
 import org.apache.ignite.internal.raft.ReadCommand;
+import org.apache.ignite.network.annotations.Transferable;
 
 /**
  * Get a value command.
  */
-public class GetValueCommand implements ReadCommand {
+@Transferable(1_000)
```
tkalkirill, line 26: We do not have a single place where all types of messages?

modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/JraftServerImpl.java
```diff
@@ -113,6 +114,9 @@ public class JraftServerImpl implements RaftServer {
     /** Request executor. */
     private ExecutorService requestExecutor;
 
+    /** Marshaller. */
```
tkalkirill, line 117: Marshaller for what?

modules/raft/src/integrationTest/java/org/apache/ignite/raft/server/counter/GetValueCommand.java
ibessonov: No, we don't, I'll do it if you're asking

modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/JraftServerImpl.java
ibessonov: I'll fix it

tkalkirill (approved):

modules/raft/src/testFixtures/java/org/apache/ignite/raft/server/counter/IncrementAndGetCommand.java
```diff
@@ -18,29 +18,22 @@
...
+@Transferable(TestMessageGroup.INCREMENT_AND_GET_COMMAND)
+public interface IncrementAndGetCommand extends WriteCommand {
     /**
      * Returns the delta.
      */
-    public long delta() {
-        return delta;
+    long delta();
+
+    static IncrementAndGetCommand incrementAndGetCommand(long delta) {
```
vldpyatkov, line 36: Why do you use a static method to create an instance, if there is a factory?
This looks like contradictory patterns to me.

modules/raft/src/testFixtures/java/org/apache/ignite/raft/server/counter/IncrementAndGetCommand.java
```diff
@@ -18,29 +18,22 @@
...
-    /**
-     * Constructor.
-     *
-     * @param delta The delta.
-     */
-    public IncrementAndGetCommand(long delta) {
-        this.delta = delta;
-    }
-
+//TODO Move to integration test directory when Maven build is not supported anymore.
```
vldpyatkov, line 28: There is "todo" without a ticket number.

ibessonov: I'll add it, thank you

modules/raft/src/testFixtures/java/org/apache/ignite/raft/server/counter/IncrementAndGetCommand.java
ibessonov: Just to be clear, these are test messages. I don't do things like that in production messages.

vldpyatkov (approved):

PR #1413, opened 2022-12-06
IGNITE-18340 Gradle build doesn't produce OpenAPI spec

valepakh:
https://issues.apache.org/jira/browse/IGNITE-18340

Pass a compiler argument to the annotation processor.


SammyVimes (approved):
LGTM

PR #1419, opened 2022-12-07
IGNITE-18239: Exclude 'compute' and 'raft' from completions on 'node …

Flaugh24:
https://issues.apache.org/jira/browse/IGNITE-18239

### DIFF
--- a/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/DynamicCompleterActivationPoint.java
+++ b/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/DynamicCompleterActivationPoint.java
@@ -18,6 +18,8 @@
 package org.apache.ignite.internal.cli.core.repl.completer;
 
 import jakarta.inject.Singleton;
+import java.util.Arrays;
+import java.util.Set;
 import org.apache.ignite.internal.cli.commands.Options;
 import org.apache.ignite.internal.cli.core.repl.completer.hocon.ClusterConfigDynamicCompleterFactory;
 import org.apache.ignite.internal.cli.core.repl.completer.hocon.NodeConfigDynamicCompleterFactory;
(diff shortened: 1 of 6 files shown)


modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/DynamicCompleterActivationPoint.java
```diff
@@ -55,10 +57,22 @@ public void activateDynamicCompleter(DynamicCompleterRegistry registry) {
...
-                        .command("node", "config", "update").build(),
+                        .command("node", "config", "show").build(),
                 nodeConfigDynamicCompleterFactory
         );
+
+        registry.register(
+                CompleterConf.builder()
+                        .command("node", "config", "update")
+                        .filter((unused, candidates) -> {
+                            Set<String> exclusions = Set.of("compute", "raft");
```
PakhomovAlexander, line 68: Could you check please the case: `compute.`, `raft.`

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/DynamicCompleterActivationPoint.java
```diff
@@ -55,10 +57,22 @@ public void activateDynamicCompleter(DynamicCompleterRegistry registry) {
...
-                        .command("node", "config", "show")
-                        .command("node", "config", "update").build(),
+                        .command("node", "config", "show").build(),
                 nodeConfigDynamicCompleterFactory
         );
+
+        registry.register(
+                CompleterConf.builder()
+                        .command("node", "config", "update")
+                        .filter((unused, candidates) -> {
```
PakhomovAlexander, line 67: This is a good idea to define a filter as a lambda right here but this approach has an issue: it is not covered with unit tests. You can define a separate class like `NodeConfigUpdateFilter` and test it with unite tests.

PakhomovAlexander (changes requested):

PakhomovAlexander (approved):

PR #1427, opened 2022-12-08
IGNITE-18356 Remove already implemented issues from todo mentioning

zstan:


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AbstractPlannerTest.java
```diff
@@ -987,6 +996,220 @@ public <RowT> ModifyRow toModifyRow(ExecutionContext<RowT> ectx, RowT row, Opera
         }
     }
 
+    /** Test purpose implementation. */
+    private static class TestInternalTableImpl implements InternalTable {
```
korlov42, line 1000: I don't like the idea to create a 200 LOC stub just to return a table name. Let's extend the IgniteTable interface with `String name()` method

korlov42 (approved):

PR #1429, opened 2022-12-09
IGNITE-17976 Correct exception is thrown on KeyValueView#get in case of lost majority

sanpwc:
https://issues.apache.org/jira/browse/IGNITE-17976


alievmirza (changes requested):
Seems like the ticket is in "work in progress" state, there are a lot of debug code, please remove it. Also, please unmute test, that was disabled using this ticket.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -1324,4 +1338,23 @@ protected CompletableFuture<ClusterNode> evaluateReadOnlyRecipientNode(int partI
             }
         });
     }
+
+    /**
+     * Wraps {@code ReplicationException} or {@code ConnectException} with {@code TransactionException}.
```
alievmirza, line 1343: Let's use {@link} here

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
```diff
@@ -813,9 +812,9 @@ public void testOneNodeRestartWithGap() {
 
         assertNotNull(table);
 
-        assertThrowsWithCause(() -> table.keyValueView().get(null, Tuple.create().set("id", 0)), TimeoutException.class);
+        assertThrows(TransactionException.class, () -> table.keyValueView().get(null, Tuple.create().set("id", 0)));
 
-        createTableWithData(ignite, TABLE_NAME_2, 1, 1);
+        createTableWithoutData(ignite, TABLE_NAME_2, 1, 1);
```
alievmirza, line 817: why do we need to change the logic of the test?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
```diff
@@ -813,9 +812,9 @@ public void testOneNodeRestartWithGap() {
 
         assertNotNull(table);
 
-        assertThrowsWithCause(() -> table.keyValueView().get(null, Tuple.create().set("id", 0)), TimeoutException.class);
+        assertThrows(TransactionException.class, () -> table.keyValueView().get(null, Tuple.create().set("id", 0)));
```
alievmirza, line 815: Should we check here the cause of an exception?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
sanpwc: AFAIK, we should only check that it's possible to create table after node stop.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
sanpwc: I don't think so, because several different causes are possible: ReplicationTimeout, TimeoutException, ReplicaUnavailable, ConnectRefuse, etc.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
sanpwc: Fixed.

alievmirza (approved):

PR #1434, opened 2022-12-12
IGNITE-18211 Sql. Adjust Affinity distribution to the new colocation rules.

xtern:
* AffinityDistribution now contains distribution zone id and table id (instead of "cache id" and "identity").
  a. DistributionFunction#satisfy now depends on zone ID (instead of identity).
  b. The table ID is used to calculate the hash of the row's collocation fields (according to current Row#colocatedHash implementation).

* ColocationHashFunction was introduced to highlight implementations of colocation hash function for different types (BinaryRow and RowT).

* AffinityService has been renamed to ColocationHashFunctionFactory because that seems to be the only thing it does.

* Currently, distribution zones do not have integration with tables, so zoneId is temporarily populated with the value of the table ID (see https://issues.apache.org/jira/browse/IGNITE-18426).


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -717,4 +724,59 @@ private CompletableFuture<Void> closeExecNode(boolean cancel) {
...
+                this.colocationFields = colocationFields;
+            }
+
+            private NativeType[] colocationColumnTypes(UUID tableId, int fieldCnt) {
+                NativeType[] colTypes = new NativeType[fieldCnt];
+                TableDescriptor tblDesc = sqlSchemaManager.tableById(tableId, -1).descriptor();
+
+                List<Integer> colocationColumns = tblDesc.colocationColumns();
+
+                assert colocationColumns.size() >= fieldCnt : "fieldsCount=" + fieldCnt + ", colocationColumns=" + colocationColumns;
```
korlov42, line 757: How it's going to work if table colocated by more column than we expected?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/trait/DistributionFunction.java
```diff
@@ -243,27 +239,35 @@ public <RowT> Destination<RowT> destination(ExecutionContext<RowT> ctx, Affinity
                 }
             }
 
-            AffinityAdapter<RowT> affinity = new AffinityAdapter<>(affSrvc.affinity(intFoo()/*CU.UNDEFINED_CACHE_ID*/), k.toIntArray(),
-                    ctx.rowHandler());
+            return new Partitioned<>(assignments, row -> {
+                int hash = 0;
+
+                for (int idx : k) {
+                    hash = 31 * hash + ctx.rowHandler().get(idx, row).hashCode();
```
korlov42, line 246: looks like need to cover DistributionFunctions with unit tests

modules/schema/src/main/java/org/apache/ignite/internal/util/ColocationHashFunction.java
```diff
@@ -15,20 +15,17 @@
...
+package org.apache.ignite.internal.util;
 
 /**
- * AffinityService interface.
- * TODO Documentation https://issues.apache.org/jira/browse/IGNITE-15859
+ * Function to calculate a hash of the colocation fields of a row.
  */
-public interface AffinityService {
+@FunctionalInterface
+public interface ColocationHashFunction<T> {
```
korlov42, line 24: Why do we need such interface in a Schema module? To be honest, I don't get the idea

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/TableDescriptor.java
```diff
@@ -109,4 +110,11 @@ default RelDataType selectForUpdateRowType(IgniteTypeFactory factory) {
      * @return Actual count of columns.
      */
     int columnsCount();
+
+    /**
+     * Returns list of colocation columns of the table.
+     *
+     * @return Actual list of colocation columns.
+     */
+    List<Integer> colocationColumns();
```
korlov42, line 119: why do we need this?

modules/schema/src/main/java/org/apache/ignite/internal/util/ColocationHashFunction.java
korlov42: Why do we need to _mark_ all implementation? Why do we need several implementations anyway? Actually we need exactly one implementation of such function. Furthermore, the function should extract partition id, not the abstract hash.

xtern: _> Why do we need several implementations anyway?_
Because currently we have different implementations for BinaryRow and RowT.

My proposal  -  remove `ColocationHashFunction` interface and keep hash calculations for `BinaryRow` and `RowT` independent (as anonymous classes). They will both use HashCalculator, maybe I was wrong and we don't need to unify them any more.

WDYT?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/trait/IgniteDistributions.java
```diff
@@ -62,44 +63,28 @@ public static IgniteDistribution broadcast() {
...
      *
      * @param key      Affinity key.
-     * @param cacheId  Affinity cache ID.
-     * @param identity Affinity identity key.
+     * @param zoneId   Affinity zone ID.
      * @return Affinity distribution.
      */
-    public static IgniteDistribution affinity(int key, int cacheId, Object identity) {
-        return hash(ImmutableIntList.of(key), DistributionFunction.affinity(cacheId, identity));
+    public static IgniteDistribution affinity(int key, UUID tableId, Object zoneId) {
```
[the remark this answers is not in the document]
xtern: Production code uses tableId (UUID) instead of zoneId and we are not sure what type it will be in the future.

But, since we already have tableId (to calculate colocation hash) we can completely remove zoneId from current affinity distribution implementation and add some extra todo.
WDYT?

Personally, I don't like this because it will require more work to do in https://issues.apache.org/jira/browse/IGNITE-18426 )

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/externalize/RelJson.java
```diff
@@ -657,10 +666,13 @@ IgniteDistribution toDistribution(Object distribution) {
...
-        if (cacheId != null) {
-            return IgniteDistributions.hash(ImmutableIntList.copyOf((List<Integer>) map.get("keys")),
-                    DistributionFunction.affinity(cacheId.intValue(), cacheId));
+        String tableIdStr = (String) map.get("tableId");
+
+        if (tableIdStr != null) {
+            UUID tableId = UUID.fromString(tableIdStr);
+            Object zoneId = map.get("zoneId");
+
+            return IgniteDistributions.hash((List<Integer>) map.get("keys"), DistributionFunction.affinity(tableId, zoneId));
```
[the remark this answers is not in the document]
xtern: Done, thanks

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/TableDescriptorImpl.java
```diff
@@ -66,13 +59,17 @@ public class TableDescriptorImpl extends NullInitializerExpressionFactory implem
...
     /**
      * Constructor.
      * TODO Documentation https://issues.apache.org/jira/browse/IGNITE-15859
      */
     public TableDescriptorImpl(
             List<ColumnDescriptor> columnDescriptors,
-            IntList colocationColumns
+            IntList colocationColumns,
+            UUID tableId,
+            Object zoneId
```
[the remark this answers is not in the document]
xtern: Done, thanks

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
xtern: Fixed, thanks

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/TableDescriptor.java
xtern: Removed from the interface, left only in the implementation.

korlov42: Why don't just use distribution.getKeys() instead?

xtern: Fixed, thanks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/trait/IgniteDistributions.java
```diff
@@ -62,44 +63,28 @@ public static IgniteDistribution broadcast() {
...
-     * @return Affinity distribution.
-     */
-    public static IgniteDistribution affinity(int key, String cacheName, Object identity) {
-        // TODO: fix cacheId
-        return affinity(key, 0, identity);
-    }
-
     /**
      * Affinity.
      * TODO Documentation https://issues.apache.org/jira/browse/IGNITE-15859
```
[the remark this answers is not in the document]
xtern: Improved as best I could)

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/trait/DistributionFunction.java
```diff
@@ -220,7 +209,7 @@ public <RowT> Destination<RowT> destination(ExecutionContext<RowT> ctx, Affinity
         }
     }
 
-    private static final class HashDistribution extends DistributionFunction {
```
korlov42, line 223: why did you remove `final` modifier?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/JoinColocationPlannerTest.java
```diff
@@ -111,15 +109,16 @@ public void joinSameTableComplexAff() throws Exception {
     }
 
     /**
-     * Re-hashing based on simple affinity is possible, so bigger table with complex affinity should be sended to the smaller one.
+     * Re-hashing based on simple affinity.
+     *
+     * <p>The table with complex affinity should be sent to the table with simple affinity.
```
korlov42, line 114: COMPLEX_TBL is twice of the size of the SIMPLE_TBL, so why do you expect the bigger table being sent to the smaller one?

korlov42 (approved):

PR #1440, opened 2022-12-13
IGNITE-17588 SQL API for C++ Client

isapego:


modules/platforms/cpp/ignite/protocol/utils.cpp
```diff
@@ -28,20 +28,73 @@
...
 namespace ignite::protocol {
 
+/**
+ * Check if int value fits in @c T.
+ *
+ * @tparam T Int type to fit value to.
+ * @param value Int value.
+ */
+template<typename T>
+inline void check_int_fits(std::int64_t value) {
```
ptupitsyn, line 38: This looks suspicious. I think we can use MsgPack type code prefixes to check if the next value in the stream is `int` or something else.

modules/platforms/cpp/tests/client-test/sql_test.cpp
```diff
@@ -0,0 +1,235 @@
...
+#include <gtest/gtest.h>
+
+#include <chrono>
+
+using namespace ignite;
+
+/**
+ * Test suite.
+ */
+class sql_test : public ignite_runner_suite {
```
ptupitsyn, line 32: Let's add more tests for various failure conditions, such as:
* Invalid SQL text - does the user get a sensible error message?
* What if I iterate the result set twice?
* Can I access result set pages after closing the cursor?

ptupitsyn (changes requested):

modules/platforms/cpp/tests/client-test/sql_test.cpp
isapego: That's a good proposal. I'll add more tests.

isapego: Added tests

ptupitsyn (approved):

PR #1442, opened 2022-12-13
IGNITE-18280 Fix dependencies in Gradle packaging tasks

valepakh:
* Fix dependencies to get rid of Gradle warnings.
* Do not assembly release artifacts (such as .NET binaries) during regular build.

This fix properly declares dependencies between packaging tasks so that Gradle can optimize them correctly and don't complain.
Also this adds a property check for adding release artifacts. The reason is that adding an artifact to the configuration automatically adds a dependency to the `assemble` task so running `gradlew build` always produces artifacts intended only for the release.

https://issues.apache.org/jira/browse/IGNITE-18280

### DIFF
--- a/packaging/build.gradle
+++ b/packaging/build.gradle
@@ -31,19 +31,17 @@ configurations {
     cliArtifacts
     cliZip
     dbZip
-    cliRelease
-    dbRelease
-    platformsRelease
+    release
 }
 
 dependencies {
-    dbArtifacts(project(':ignite-runner'))
-    cliArtifacts(project(':ignite-cli'))
-    cliZip(project(path: ':packaging-cli', configuration: 'cliZip'))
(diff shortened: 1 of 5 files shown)


packaging/build.gradle
```diff
@@ -142,6 +141,7 @@ task prepareRelease(type: Copy) {
     from configurations.platformsRelease
     dependsOn createChecksums, signAllSrcZip, signAllDistZip
     from file("$buildDir/distributions")
+    from file("$buildDir/checksums")
```
PakhomovAlexander, line 144: shouldn't it be `platformsRelease` artifact instead?

Pochatkin:
LGTM

packaging/build.gradle
valepakh: We take the files from distributions and checksums folders directly since they are not declare outputs correctly.

PR #1445, opened 2022-12-13
IGNITE-18399 Adapt WorkDirectoryExtension to Gradle

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-18399

### DIFF
--- a/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/WorkDirectoryExtension.java
+++ b/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/WorkDirectoryExtension.java
@@ -56,8 +56,8 @@ import org.junit.platform.commons.support.HierarchyTraversalMode;
  * to the module, where the tests are being run, and their paths depends on the lifecycle of the folder:
  *
  * <ol>
- *     <li>For test methods: "target/work/{@literal <name-of-the-test-class>/<name-of-the-test-method>_<current_time_millis>}"</li>
- *     <li>For test classes: "target/work/{@literal <name-of-the-test-class>/static_<current_time_millis>}"</li>
+ *     <li>For test methods: "build/work/{@literal <name-of-the-test-class>/<name-of-the-test-method>_<current_time_millis>}"</li>
+ *     <li>For test classes: "build/work/{@literal <name-of-the-test-class>/static_<current_time_millis>}"</li>
  * </ol>
  *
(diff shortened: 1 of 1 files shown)


modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/WorkDirectoryExtension.java
```diff
@@ -270,4 +270,15 @@ private static boolean isEmpty(Path dir) throws IOException {
         }
     }
 
+    private static Path getBasePath() {
+        // Maven build folder
+        Path mavenPath = Path.of("target");
```
SammyVimes, line 275: Do you think we need maven?

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/WorkDirectoryExtension.java
```diff
@@ -56,8 +56,8 @@
  * to the module, where the tests are being run, and their paths depends on the lifecycle of the folder:
  *
  * <ol>
- *     <li>For test methods: "target/work/{@literal <name-of-the-test-class>/<name-of-the-test-method>_<current_time_millis>}"</li>
- *     <li>For test classes: "target/work/{@literal <name-of-the-test-class>/static_<current_time_millis>}"</li>
+ *     <li>For test methods: "build/work/{@literal <name-of-the-test-class>/<name-of-the-test-method>_<current_time_millis>}"</li>
```
SammyVimes, line 59: Seems like it can be `target` if there's `target` folder

SammyVimes (approved):
lgtm

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/WorkDirectoryExtension.java
sashapolo: Somebody is still using Maven, I don't want to ruin the Ignite experience for them)

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/WorkDirectoryExtension.java
sashapolo: Gradle is going to be the main build system, so no need to explicitly mention Maven in the docs

PR #1449, opened 2022-12-14
IGNITE-17271 Sort out sqlogic tests

zstan:


modules/runner/src/integrationTest/sql/function/generic/test_coalesce.test
```diff
@@ -45,7 +45,7 @@ statement ok
 INSERT INTO test3 VALUES (null, 22), (12, null), (13, 22)
 
 query T
-SELECT COALESCE(CAST(a AS VARCHAR), 11) FROM test3 ORDER BY a;
+SELECT COALESCE(CAST(a AS VARCHAR), 11) as b FROM test3 ORDER BY b;
```
korlov42, line 48: I believe the proper fix is to move 11 at the end

zstan: done, also all others previous discussed issues are changed too. check plz ?

PR #1451, opened 2022-12-14
IGNITE-18413 Reorganize metastorage modules

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-18413


ibessonov (approved):
Let's wait for @agura's approval as well

modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/dsl/OperationType.java
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package org.apache.ignite.internal.metastorage.common;
+package org.apache.ignite.internal.metastorage.dsl;
```
agura, line 18: I really don't like the `dsl` name. There is still a difference between a language and an API. Really minor, but I can't be silent :)

modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
```diff
@@ -0,0 +1,125 @@
+/*
```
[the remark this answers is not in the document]
sashapolo: I agree with you, however this is how it was implemented before my changes, I simply extracted an interface. I don't want to rename this in my PR in order to reduce the amount of changes.

modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/dsl/OperationType.java
sashapolo: Sure, do you have a better name in mind?

agura (approved):

PR #1454, opened 2022-12-19
IGNITE-18418: Create ignite-jdbc module

Flaugh24:


Flaugh24:
https://issues.apache.org/jira/browse/IGNITE-18418

modules/jdbc/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+    alias(libs.plugins.shadow)
+}
+
+apply from: "$rootDir/buildscripts/java-core.gradle"
+apply from: "$rootDir/buildscripts/publishing.gradle"
+apply from: "$rootDir/buildscripts/java-junit5.gradle"
+apply from: "$rootDir/buildscripts/java-integration-test.gradle"
+
+dependencies {
+    api project(':ignite-client')
```
Pochatkin, line 28: Why api level?

modules/jdbc/src/main/java/org/apache/ignite/jdbc/IgniteJdbcDriver.java
```diff
@@ -17,7 +17,7 @@
 
 package org.apache.ignite.jdbc;
 
-import static org.apache.ignite.internal.jdbc.ConnectionPropertiesImpl.URL_PREFIX;
+import static org.apache.ignite.jdbc.internal.ConnectionPropertiesImpl.URL_PREFIX;
```
Pochatkin, line 20: This is incorrect package name. Should be org.apache.ignite.internal.jdbc

modules/jdbc/build.gradle
Flaugh24: changed to implementation

modules/jdbc/src/main/java/org/apache/ignite/jdbc/JdbcPreparedStatement.java
```diff
@@ -107,7 +106,7 @@ public int[] executeBatch() throws SQLException {
 
         closeResults();
 
-        if (CollectionUtils.nullOrEmpty(batchedArgs) || StringUtil.isNullOrEmpty(sql)) {
+        if (CollectionUtils.nullOrEmpty(batchedArgs) || (sql == null || sql.isBlank())) {
```
korlov42, line 109: Perhaps, it would be better to reuse `org.apache.ignite.internal.util.StringUtils#nullOrBlank` from core module

korlov42 (commented):
all implementations of java.sql.* interfaces (like IgniteJdbcDriver or JdbcStatement, for example) should remain in `internal` package

modules/jdbc/src/main/java/org/apache/ignite/jdbc/JdbcPreparedStatement.java
Flaugh24: done

Pochatkin:
> all implementations of java.sql.* interfaces (like IgniteJdbcDriver or JdbcStatement, for example) should remain in `internal` package

We don't want to generate Javadocs for these implemenatations?

korlov42:
> We don't want to generate Javadocs for these implemenatations?

To be honest, I don't get this question.

Public API is exposed in `api` module only. Everything outside of this package should go into `internal` package. For JDBC, the public API is exposed in `java.sql.*` package. I believe, that any implementation of public interface is our internal business.

Regarding your comment, I thought, javadoc is generated for all classes, isn't it?

Pochatkin:
> > We don't want to generate Javadocs for these implemenatations?
> 
> To be honest, I don't get this question.
> 
> Public API is exposed in `api` module only. Everything outside of this package should go into `internal` package. For JDBC, the public API is exposed in `java.sql.*` package. I believe, that any implementation of public interface is our internal business.
> 
> Regarding your comment, I thought, javadoc is generated for all classes, isn't it?

No, we have  exclude 'org/apache/ignite/internal/**' for Javadoc generator

korlov42:
As far as I can see, we do not publish such javadoc in [Ignite 2.x](https://ignite.apache.org/releases/latest/javadoc/index.html), thus this looks OK

modules/jdbc/src/main/java/org/apache/ignite/internal/ConnectionProperties.java
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package org.apache.ignite.internal.jdbc;
+package org.apache.ignite.internal;
```
korlov42, line 18: I would prefer to follow common patter `org.apache.ignite.internal.<name of the module>` which is `org.apache.ignite.internal.jdbc` for module `jdbc`

modules/jdbc/src/main/java/org/apache/ignite/jdbc/IgniteJdbcDriver.java
```diff
@@ -17,7 +17,7 @@
 
 package org.apache.ignite.jdbc;
 
-import static org.apache.ignite.internal.jdbc.ConnectionPropertiesImpl.URL_PREFIX;
+import static org.apache.ignite.internal.ConnectionPropertiesImpl.URL_PREFIX;
```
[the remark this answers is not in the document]
Flaugh24: done

korlov42 (approved):

PR #1456, opened 2022-12-19
IGNITE-18206 Sql. Prohibit using 2-phase aggregates with distinct accumulator on map phase

zstan:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rule/HashAggregateConverterRule.java
```diff
@@ -74,6 +77,21 @@ protected PhysicalNode convert(RelOptPlanner planner, RelMetadataQuery mq,
         }
     }
 
+    /**
+     * Return {@code true} if observes COUNT and DISTINCT simultaneously in aggregate.
+     *
+     * @param aggCalls Aggregates.
+     * @return {@code true} If found, {@code false} otherwise.
+     */
+    private static boolean countWithDistinctAgg(List<AggregateCall> aggCalls) {
```
korlov42, line 86: what about other aggregate functions?

zstan: /me just narrow thinking ) fixed

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AggregatePlannerTest.java
```diff
@@ -372,6 +373,63 @@ public void colocated(AggregateAlgorithm algo) throws Exception {
...
+
+    /**
+     * Check that plan does not contain distinct accumulators on map nodes.
+     *
+     * @param sql Request string.
+     * @param publicSchema Schema.
+     * @throws Exception If failed.
+     */
+    private void checkDistinctInMapAggNode(String sql, IgniteSchema publicSchema) throws Exception {
+        String[][] disabledRules = new String[][] {{""}, {"MapReduceHashAggregateConverterRule"},
```
korlov42, line 419: Why do we need an empty string here?

korlov42 (approved):

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AggregatePlannerTest.java
zstan: discussed in messenger, will be change a bit.

PR #1458, opened 2022-12-19
IGNITE-18050 Possible phantom reads during sorted index scan

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-18050

### DIFF
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
@@ -23,6 +23,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.nio.ByteBuffer;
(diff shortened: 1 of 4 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
```diff
@@ -95,39 +96,54 @@ public CompletableFuture<Void> locksForLookup(UUID txId, BinaryRow tableRow) {
      * @return A future representing a result.
```
denis-chudov, line 96: there is no description, what is the result. Same for `acquireLockNextKey`.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
```diff
@@ -427,6 +428,55 @@ public void pureTableScan(Function<InternalTransaction, CompletableFuture<Intege
...
 
+    @Test
+    public void testTwiceScanInTransaction() throws Exception {
+        TableImpl table = getOrCreateTable();
+
+        InternalTable internalTable = table.internalTable();
+
+        KeyValueView kvView = table.keyValueView();
+
+        UUID soredIndexId = getSortedIndexId();
```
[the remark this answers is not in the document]
vldpyatkov: Done.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
vldpyatkov: Added description.

PR #1461, opened 2022-12-20
IGNITE-18079 Integrate RAFT streaming snapshots

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18079


modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
```diff
@@ -0,0 +1,781 @@
...
+import org.junit.jupiter.api.TestInfo;
+import org.junit.jupiter.api.Timeout;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+@SuppressWarnings("resource")
+@ExtendWith(WorkDirectoryExtension.class)
+@Timeout(60)
+class ItTableRaftSnapshotsTest {
```
tkalkirill, line 98: Why do't use `org.apache.ignite.internal.AbstractClusterIntegrationTest` ?

There is no documentation for methods and tests and it is quite clear to me what is being tested and what we expect, please add documentation.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
```diff
@@ -0,0 +1,781 @@
...
+    void shutdownCluster() {
+        cluster.shutdown();
+    }
+
+    private Path nodeWorkDir(String nodeName) {
+        return workDir.resolve(nodeName);
+    }
+
+    @Test
+    void snapshotReadOnRestartWorksCorrectly() {
```
tkalkirill, line 125: Don't quite understand what is being tested here? more like local recovery.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
```diff
@@ -0,0 +1,781 @@
...
+            knockout.reanimateNode(nodeIndex, this);
+
+            knockedOutIndices.remove(nodeIndex);
+        }
+    }
+
+    /**
+     * A way to make a node be separated from a cluster and stop receiving updates.
+     */
+    private enum NodeKnockout {
```
tkalkirill, line 727: Why enum and not just methods and functions?

modules/cluster-management/src/integrationTest/java/org/apache/ignite/internal/cluster/management/MockNode.java
```diff
@@ -77,7 +82,13 @@ private void init(int port) throws IOException {
 
         this.clusterService = ClusterServiceTestUtils.clusterService(testInfo, port, nodeFinder);
 
-        Loza raftManager = new Loza(clusterService, null, workDir, new HybridClockImpl());
+        RaftConfiguration raftConfiguration = mock(RaftConfiguration.class);
```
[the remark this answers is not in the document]
rpuch: To the constructor of `MockNode`? Because it's not needed now. The only reason it's passed here is to avoid an NPE.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
```diff
@@ -0,0 +1,781 @@
...
+
+@SuppressWarnings("resource")
+@ExtendWith(WorkDirectoryExtension.class)
+@Timeout(60)
+class ItTableRaftSnapshotsTest {
+    private static final IgniteLogger LOG = Loggers.forClass(ItTableRaftSnapshotsTest.class);
+
+    private static final int QUERY_TIMEOUT_MS = 10_000;
+
+    private static final String DEFAULT_STORAGE_ENGINE = RocksDbStorageEngine.ENGINE_NAME;
```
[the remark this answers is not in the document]
rpuch: There is a test that is run for all storages. But it does not seem reasonable to do this for all tests, as otherwise we'll have too many tests, but this will not add value.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
rpuch: To make it usable with `@ParameterizedTest` via `@EnumSource`

modules/network/src/main/java/org/apache/ignite/network/DefaultMessagingService.java
```diff
@@ -275,7 +280,13 @@ private void sendToSelf(NetworkMessage msg, @Nullable Long correlationId) {
      */
     private void onMessage(InNetworkObject obj) {
         if (isInNetworkThread()) {
-            inboundExecutor.execute(() -> onMessage(obj));
+            inboundExecutor.execute(() -> {
+                try {
+                    onMessage(obj);
+                } catch (RuntimeException e) {
+                    LOG.warn("onMessage() failed while processing " + obj.message() + " from " + obj.consistentId(), e);
```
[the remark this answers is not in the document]
rpuch: Switched it to `{}` style, added a guard to only do this work when the corresponding logging level is enabled. Switched the level to INFO (because we need to see these messages, at least during the active development phase, this really helps in debugging faulty runs on TC).

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
rpuch: JRaft 'reads' a snapshot on start, this method tested that this 'read' happens successfully. But you are right, this is also tested elsewhere. Removed the test.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
rpuch: `AbstractClusterIntegrationTest` is more rigid, it starts and inits the cluster in a `@BeforeEach` method. Here, we need more control on when and how many nodes are started.

Added javadocs, please let me know if something else should be added.

tkalkirill: I see that your code and `AbstractClusterIntegrationTest` are very similar, and you need to combine them, for example, put `Cluster` in a separate class and use it in these classes.

modules/cluster-management/src/integrationTest/java/org/apache/ignite/internal/cluster/management/MockNode.java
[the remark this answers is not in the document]
rpuch: Currently, only 3 tests use `MockNode`. Neither of them needs to pass configuration, so only the constructor that does not accept the configuration and just mocks it is used. So if I add the second constructor (that accepts `RaftConfiguration`), this constructor will simply not be used at all. I think we should not add what we don't need yet. If we need it in the future, we'll add that constructor.

modules/configuration/src/testFixtures/java/org/apache/ignite/internal/configuration/testframework/ConfigurationExtension.java
```diff
@@ -273,7 +274,7 @@ private static Object cfgValue(
         );
 
         // RootKey must be mocked, there's no way to instantiate it using a public constructor.
-        RootKey rootKey = mock(RootKey.class);
+        RootKey rootKey = mock(RootKey.class, withSettings().lenient());
```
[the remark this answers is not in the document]
rpuch: Without `lenient`, a test (unrelated to this PR) started failing. It happened because I had to add `@ExtendWith(ConfigurationExtension)`.

The mock `RootKey` has to be lenient so that it does not fail a test in which any of the mocked method is not called.

modules/network/src/main/java/org/apache/ignite/network/DefaultMessagingService.java
```diff
@@ -402,7 +419,8 @@ public void stop() {
 
     /**
      * Installs a predicate, it will be consulted with for each message being sent; when it returns {@code true}, the
-     * message will be silently dropped (it will not be sent, the corresponding future will never complete).
+     * message will be dropped (it will not be sent; the corresponding future will time out soon for {@code invoke()} methods
```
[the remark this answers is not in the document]
rpuch: 'invoke' is different from 'send' in that it has `timeout` parameter, so, if the recipient never responds, a timeout happens sooner or later. For 'send', this is not the case: if the recipient never responds, the future just hangs forever.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
rpuch: I made `Cluster` a top-level class and prepared it for being used by `AbstractClusterIntegrationTest`, but migrating the test (and its subclasses) to use `Cluster` requires some work that is out of scope of this PR, so I created a ticket https://issues.apache.org/jira/browse/IGNITE-18465 and mentioned it in TODOs.

modules/network/src/main/java/org/apache/ignite/network/DefaultMessagingService.java
```diff
@@ -275,7 +280,19 @@ private void sendToSelf(NetworkMessage msg, @Nullable Long correlationId) {
      */
     private void onMessage(InNetworkObject obj) {
         if (isInNetworkThread()) {
-            inboundExecutor.execute(() -> onMessage(obj));
+            inboundExecutor.execute(() -> {
+                try {
+                    onMessage(obj);
+                } catch (Throwable e) {
+                    if (LOG.isInfoEnabled()) {
+                        LOG.info("onMessage() failed while processing {} from {}", e, obj.message(), obj.consistentId());
```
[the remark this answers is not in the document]
rpuch: Changed this to `error()` as this is actually an error.

modules/configuration/src/testFixtures/java/org/apache/ignite/internal/configuration/testframework/ConfigurationExtension.java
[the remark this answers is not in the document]
tkalkirill: I do not quite understand how tests can fall if methods are not called?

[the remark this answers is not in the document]
tkalkirill: I added an `@ExtendWith(ConfigurationExtension.class)` for `org.apache.ignite.internal.tx.storage.state.test.TestTxStateStorageTest` and nothing fell in my pr.

modules/raft-api/src/main/java/org/apache/ignite/internal/raft/configuration/RaftConfigurationSchema.java
```diff
@@ -20,13 +20,18 @@
...
 
 /**
  * Raft configuration schema.
  */
 @SuppressWarnings("PMD.UnusedPrivateField")
 @ConfigurationRoot(rootName = "raft", type = ConfigurationType.LOCAL)
 public class RaftConfigurationSchema {
+    /** RPC Timeout for InstallSnapshot request (in milliseconds). */
+    @Value(hasDefault = true)
+    public int rpcInstallSnapshotTimeout = 5 * 60 * 1000;
```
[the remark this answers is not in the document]
tkalkirill: Will this be enough time if the partitions are large enough, for example 50 GB?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/Cluster.java
```diff
@@ -0,0 +1,346 @@
...
+import org.apache.ignite.raft.jraft.util.concurrent.ConcurrentHashSet;
+import org.apache.ignite.sql.Session;
+import org.jetbrains.annotations.Nullable;
+import org.junit.jupiter.api.TestInfo;
+
+/**
+ * Cluster of nodes used for testing.
+ */
+@SuppressWarnings("resource")
+public class Cluster {
```
tkalkirill, line 59: I think this class should be in `testFixtures`.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/Cluster.java
```diff
@@ -0,0 +1,346 @@
...
+    private final Path workDir;
+
+    private final String nodeBootstrapConfig;
+
+    /** Cluster nodes. */
+    private final List<IgniteImpl> nodes = new CopyOnWriteArrayList<>();
+
+    private volatile boolean started = false;
+
+    /** Indices of nodes that have been knocked out. */
```
tkalkirill, line 81: For me it sounds like SQL indices, I think it can be reformulated.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/Cluster.java
```diff
@@ -0,0 +1,346 @@
...
+        String config = IgniteStringFormatter.format(nodeBootstrapConfig, BASE_PORT + nodeIndex, CONNECT_NODE_ADDR);
+
+        return IgnitionManager.start(nodeName, config, workDir.resolve(nodeName))
+                .thenApply(IgniteImpl.class::cast);
+    }
+
+    /**
+     * Returns an Ignite node (a member of the cluster) by its index.
+     */
+    public IgniteImpl node(int index) {
```
tkalkirill, line 132:
```suggestion
    public @Nullable IgniteImpl node(int index) {
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/Cluster.java
```diff
@@ -0,0 +1,346 @@
...
+        return nodes.get(index);
+    }
+
+    /**
+     * Returns a node that is not stopped and not knocked out (so it can be used to interact with the cluster).
+     */
+    public IgniteImpl entryNode() {
```
tkalkirill, line 139: Not quite the right name.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/Cluster.java
```diff
@@ -0,0 +1,346 @@
...
+            newIgniteNode = startClusterNode(index).get(10, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+
+            throw new RuntimeException(e);
+        } catch (ExecutionException | TimeoutException e) {
+            throw new RuntimeException(e);
+        }
+
+        nodes.set(index, newIgniteNode);
```
tkalkirill, line 168: If memory serves me right, then if the list is smaller than the index, then there will be an error.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/Cluster.java
```diff
@@ -0,0 +1,346 @@
...
+
+        return newIgniteNode;
+    }
+
+    /**
+     * Stops a node by index.
+     *
+     * @param index Node index in the cluster.
+     */
+    public void stopNode(int index) {
```
tkalkirill, line 178: What happens if there is no node?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/Cluster.java
```diff
@@ -0,0 +1,346 @@
...
+        nodes.set(index, null);
+    }
+
+    /**
+     * Restarts a node by index.
+     *
+     * @param index Node index in the cluster.
+     * @return New node.
+     */
+    public IgniteImpl restartNode(int index) {
```
tkalkirill, line 190: What happens if there is no node?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
```diff
@@ -0,0 +1,637 @@
...
+            + "  \"raft\": {"
+            + "    \"rpcInstallSnapshotTimeout\": 10000"
+            + "  }"
+            + "}";
+
+    /**
+     * Marker that instructs to create a table with the default storage engine. Used in tests that are indifferent
+     * to a storage engine used.
+     */
+    private static final String DEFAULT_STORAGE_ENGINE = "<default>";
```
tkalkirill, line 118: We do not need to specify the default engine, it is already applied.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
```diff
@@ -0,0 +1,637 @@
...
+    void createCluster(TestInfo testInfo) {
+        cluster = new Cluster(testInfo, workDir, NODE_BOOTSTRAP_CFG);
+    }
+
+    @AfterEach
+    void shutdownCluster() {
+        cluster.shutdown();
+    }
+
+    private void doInSession(int nodeIndex, Consumer<Session> action) {
```
tkalkirill, line 140: I think it will be useful to move methods `doInSession` and executeUpdate' to `Cluster`.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/Cluster.java
rpuch: Any suggestions?

modules/raft-api/src/main/java/org/apache/ignite/internal/raft/configuration/RaftConfigurationSchema.java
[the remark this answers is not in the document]
rpuch: 5 minutes is the JRaft default. Added a TODO to think about this default.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/Cluster.java
rpuch: `testFixtures` contains classes that might be needed by tests in other modules. `Cluster` is defined in `runner`, and no module depends on it (and probably no module *can* depend on it, `runner` is designed to be on the top of the hierarchy). So there will be no tests in other modules that will need to use `testFixtures` from `runner`, so `testFixtures` is simply not needed here.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/Cluster.java
rpuch: Renamed the field to make it more explicit

modules/runner/src/integrationTest/java/org/apache/ignite/internal/Cluster.java
rpuch: This method should never return `null`, added a check

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
rpuch: Discussed in private. Added a TODO to address a problem of forgetting a storage engine.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
rpuch: Moved the methods

modules/configuration/src/testFixtures/java/org/apache/ignite/internal/configuration/testframework/ConfigurationExtension.java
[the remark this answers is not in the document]
rpuch: Take a look at `ItRaftGroupServiceTest`: it will fail if you remove `lenient()`. https://ci.ignite.apache.org/buildConfiguration/ApacheIgnite3xGradle_Test_RunAllTests/6982075?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&pluginCoverage=true&expandCode+Inspection=true&expandBuildChangesSection=true&expandBuildTestsSection=true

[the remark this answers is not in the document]
tkalkirill: It turns out interesting, locally I can not reproduce.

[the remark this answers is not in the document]
rpuch: It reproduces easily on my machine in this branch

modules/runner/src/integrationTest/java/org/apache/ignite/internal/Cluster.java
rpuch: Renamed to `aliveNode()`; also, renamed `aliveNodes()` to `runningNodes()` to avoid confusion

modules/network/src/main/java/org/apache/ignite/network/DefaultMessagingService.java
```diff
@@ -201,6 +201,11 @@ private CompletableFuture<NetworkMessage> invoke0(ClusterNode recipient, Network
             return failedFuture(new NodeStoppingException());
         }
 
+        BiPredicate<String, NetworkMessage> dropMessage = dropMessagePredicate;
```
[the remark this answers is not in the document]
rpuch: Added TODOs to remove/move this code

modules/cluster-management/src/integrationTest/java/org/apache/ignite/internal/cluster/management/MockNode.java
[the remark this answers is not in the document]
rpuch: Discussed privately, decided to leave it as is for now

ibessonov (approved):

ibessonov (approved):

PR #1462, opened 2022-12-20
IGNITE-18022 Creating an API for full rebalance of TxStateStorage on receiver

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18022

### DIFF
--- a/modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/TxStateStorage.java
+++ b/modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/TxStateStorage.java
@@ -132,10 +132,11 @@ public interface TxStateStorage extends ManuallyCloseable {
     void destroy();
 
     /**
-     * Prepares the transaction state storage for a full rebalance: clears the storage, sets the {@link #lastAppliedIndex()} to
-     * {@link #FULL_REBALANCE_IN_PROGRESS}, and closes all cursors.
+     * Prepares the transaction state storage for a full rebalance: clears the storage, sets the {@link #lastAppliedIndex()} and
+     * {@link #lastAppliedTerm()} to {@link #FULL_REBALANCE_IN_PROGRESS}, and closes all cursors.
      *
-     * <p>After calling this method, only write methods will be available, and read methods will throw {@link IgniteInternalException}.
+     * <p>After calling this method, only write methods will be available, and read methods with {@link #lastApplied(long, long)} will
+     * throw {@link IgniteInternalException}.
      *
      * <p>This method must be called before every full rebalance of transaction state storage and ends with a call to one of the methods:
      * <ul>
@@ -152,7 +153,8 @@ public interface TxStateStorage extends ManuallyCloseable {
     CompletableFuture<Void> startFullRebalance();
 
     /**
-     * Aborts a full rebalance for transaction state storage: clears the storage, sets the {@link #lastAppliedIndex()} to {@code 0}.
+     * Aborts a full rebalance for transaction state storage: clears the storage, sets the {@link #lastAppliedIndex()} and
+     * {@link #lastAppliedTerm()} to {@code 0}.
      *
      * <p>After calling this method, methods for writing and reading will be available.
      *
--- a/modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/TxStateStorage.java
+++ b/modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/TxStateStorage.java
@@ -136,7 +136,7 @@ public interface TxStateStorage extends ManuallyCloseable {
      * {@link #lastAppliedTerm()} to {@link #FULL_REBALANCE_IN_PROGRESS}, and closes all cursors.
      *
      * <p>After calling this method, only write methods will be available, and read methods with {@link #lastApplied(long, long)} will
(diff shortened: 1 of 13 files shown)


modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/TxStateStorage.java
```diff
@@ -122,9 +125,55 @@ public interface TxStateStorage extends ManuallyCloseable {
...
+     * @throws IgniteInternalException with {@link Transactions#TX_STATE_STORAGE_ERR} error code in case when the operation has failed.
      */
     void destroy();
+
+    /**
+     * Prepares the transaction state storage for a full rebalance: clears the storage, sets the {@link #lastAppliedIndex()} and
+     * {@link #lastAppliedTerm()} to {@link #FULL_REBALANCE_IN_PROGRESS}, and closes all cursors.
+     *
+     * <p>After calling this method, only write methods will be available, and read methods with {@link #lastApplied(long, long)} will
+     * throw {@link IgniteInternalException}.
```
ibessonov, line 139: Can we have a more specific exception? Maybe a link to the corresponding error code? Let's make our exceptions better.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/TxStateStorage.java
```diff
@@ -122,9 +125,55 @@ public interface TxStateStorage extends ManuallyCloseable {
...
+     *
+     * <p>This method must be called before every full rebalance of transaction state storage and ends with a call to one of the methods:
+     * <ul>
+     *     <li>{@link #abortFullRebalance()} - in case of errors or cancellation of a full rebalance;</li>
+     *     <li>{@link #finishFullRebalance(long, long)} - in case of successful completion of a full rebalance.</li>
+     * </ul>
+     *
+     * <p>If the {@link #lastAppliedIndex()} is {@link #FULL_REBALANCE_IN_PROGRESS} after a node restart, then the storage needs to be
+     * cleared before it can be used.
```
ibessonov, line 148: Ok, how do you clean it using this interface? Maybe we shouldn't mention it like this, because cleaning happens automatically during the storage start, isn't it?

modules/transactions/src/testFixtures/java/org/apache/ignite/internal/tx/storage/state/AbstractTxStateStorageTest.java
```diff
@@ -257,53 +247,178 @@ public void scanOnlySeesDataExistingAtTheMomentOfCreation() throws Exception {
             List<UUID> txIdsReturnedByScan = cursor.stream()
                     .map(IgniteBiTuple::getKey)
                     .collect(toList());
+
             assertThat(txIdsReturnedByScan, is(List.of(existingBeforeScan)));
         }
-
-        partitionStorage.close();
```
ibessonov, line 263: This implies that partitions are closed by the table, right? That's cool

modules/transactions/src/testFixtures/java/org/apache/ignite/internal/tx/storage/state/AbstractTxStateStorageTest.java
```diff
@@ -257,53 +247,178 @@ public void scanOnlySeesDataExistingAtTheMomentOfCreation() throws Exception {
...
+        assertThat(partitionStorage.lastAppliedIndex(), is(10L));
+        assertThat(partitionStorage.lastAppliedTerm(), is(2L));
+    }
+
+    @Test
+    public void testSuccessFullRebalance() throws Exception {
+        TxStateStorage storage = tableStorage.getOrCreateTxStateStorage(0);
+
+        // Nothing will happen because the full rebalance has not started.
+        storage.finishFullRebalance(100, 500).get(1, SECONDS);
```
ibessonov, line 289: Are you sure that this one shouldn't throw an exception?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/TxStateStorage.java
tkalkirill: You're right, we can make it automatic at the start of the storage, I think it's worth mentioning.

modules/transactions/src/testFixtures/java/org/apache/ignite/internal/tx/storage/state/AbstractTxStateStorageTest.java
tkalkirill: You're right, it makes sense.

modules/transactions/src/testFixtures/java/org/apache/ignite/internal/tx/storage/state/test/TestTxStateStorage.java
```diff
@@ -100,11 +100,31 @@ public void remove(UUID txId) {
...
-        return Cursor.fromIterable(copy);
+        return new Cursor<>() {
+            @Override
+            public void close() {
+                // No-op.
+            }
+
+            @Override
+            public boolean hasNext() {
+                assert fullRebalanceFutureReference.get() == null;
```
ibessonov, line 116: Why do we even need it?

ibessonov (approved):

PR #1464, opened 2022-12-21
IGNITE-18019 GC methods in MvPartitionStorage & new tests & old tests reorganization.

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-18019

### DIFF
--- a/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
+++ b/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
@@ -31,12 +31,17 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Random;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BooleanSupplier;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
 import org.apache.ignite.internal.logger.IgniteLogger;
 import org.apache.ignite.internal.logger.Loggers;
 import org.apache.ignite.internal.thread.NamedThreadFactory;
@@ -765,5 +770,44 @@ public final class IgniteTestUtils {
     @SuppressWarnings("UnusedReturnValue")
     public static <T> T await(CompletionStage<T> stage) {
         return await(stage, TIMEOUT_SEC, TimeUnit.SECONDS);
+    }
+
+    /**
+     * Runs all actions, each in a separate thread, having a {@link CyclicBarrier} before calling {@link RunnableX#run()}.
+     *
+     * @throws InterruptedException If failed to {@link Thread#join()} a thread.
+     */
+    public static void startRace(RunnableX... actions) throws InterruptedException {
+        int length = actions.length;
+
+        CyclicBarrier barrier = new CyclicBarrier(length);
+
+        Set<Throwable> throwables = ConcurrentHashMap.newKeySet();
+
+        Thread[] threads = IntStream.range(0, length).mapToObj(i -> new Thread(() -> {
+            try {
+                barrier.await();
+
+                actions[i].run();
+            } catch (Throwable e) {
+                throwables.add(e);
+            }
+        })).toArray(Thread[]::new);
+
+        Stream.of(threads).forEach(Thread::start);
+
+        for (Thread thread : threads) {
+            thread.join();
+        }
+
+        if (!throwables.isEmpty()) {
+            AssertionError assertionError = new AssertionError("One or several threads have failed.");
+
+            for (Throwable throwable : throwables) {
+                assertionError.addSuppressed(throwable);
+            }
+
+            throw assertionError;
+        }
     }
 }
--- a/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
+++ b/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
@@ -773,11 +773,17 @@ public final class IgniteTestUtils {
     }
 
     /**
-     * Runs all actions, each in a separate thread, having a {@link CyclicBarrier} before calling {@link RunnableX#run()}.
-     *
-     * @throws InterruptedException If interrupted when trying to {@link Thread#join()} a thread.
+     * {@link #runRace(long, RunnableX...)} with default timeout of 10 seconds.
      */
-    public static void runRace(RunnableX... actions) throws InterruptedException {
+    public static void runRace(RunnableX... actions) {
+        runRace(TimeUnit.SECONDS.toMillis(10), actions);
+    }
+
+    /**
+     * Runs all actions, each in a separate thread, having a {@link CyclicBarrier} before calling {@link RunnableX#run()}.
+     * Waits for threads completion or fails with the assertion if timeout exceeded.
+     */
+    public static void runRace(long timeoutMillis, RunnableX... actions) {
         int length = actions.length;
 
         CyclicBarrier barrier = new CyclicBarrier(length);
@@ -796,8 +802,16 @@ public final class IgniteTestUtils {
 
         Stream.of(threads).forEach(Thread::start);
 
-        for (Thread thread : threads) {
-            thread.join();
+        try {
+            for (Thread thread : threads) {
+                thread.join(timeoutMillis);
+            }
+        } catch (InterruptedException e) {
+            for (Thread thread : threads) {
+                thread.interrupt();
+            }
+
+            fail("Race operations took too long.");
         }
 
         if (!throwables.isEmpty()) {
--- a/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
+++ b/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
@@ -775,9 +775,9 @@ public final class IgniteTestUtils {
     /**
      * Runs all actions, each in a separate thread, having a {@link CyclicBarrier} before calling {@link RunnableX#run()}.
      *
-     * @throws InterruptedException If failed to {@link Thread#join()} a thread.
+     * @throws InterruptedException If interrupted when trying to {@link Thread#join()} a thread.
      */
-    public static void startRace(RunnableX... actions) throws InterruptedException {
+    public static void runRace(RunnableX... actions) throws InterruptedException {
         int length = actions.length;
 
         CyclicBarrier barrier = new CyclicBarrier(length);
--- a/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
+++ b/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
@@ -782,6 +782,8 @@ public final class IgniteTestUtils {
     /**
      * Runs all actions, each in a separate thread, having a {@link CyclicBarrier} before calling {@link RunnableX#run()}.
      * Waits for threads completion or fails with the assertion if timeout exceeded.
+     *
+     * @throws AssertionError In case of timeout or if any of the runnables thrown an exception.
      */
     public static void runRace(long timeoutMillis, RunnableX... actions) {
         int length = actions.length;
--- a/modules/storage-api/src/main/java/org/apache/ignite/internal/storage/BinaryRowWithRowId.java
+++ b/modules/storage-api/src/main/java/org/apache/ignite/internal/storage/BinaryRowWithRowId.java
@@ -17,16 +17,35 @@
 
 package org.apache.ignite.internal.storage;
 
-import org.apache.ignite.internal.storage.engine.StorageEngine;
-import org.apache.ignite.internal.storage.impl.TestMvPartitionStorage;
-import org.apache.ignite.internal.storage.impl.TestStorageEngine;
+import org.apache.ignite.internal.schema.BinaryRow;
+import org.jetbrains.annotations.Nullable;
 
 /**
- * MV partition storage test implementation for {@link TestMvPartitionStorage} class.
+ * Wrapper that holds both {@link BinaryRow} and {@link RowId}.
  */
-public class TestMvPartitionStorageTest extends AbstractMvPartitionStorageTest {
-    @Override
-    protected StorageEngine createEngine() {
-        return new TestStorageEngine();
+public class BinaryRowWithRowId {
+    /** Binary row. */
+    private final @Nullable BinaryRow binaryRow;
+
+    /** Row id. */
+    private final RowId rowId;
+
+    /**
+     * Constructor.
+     *
+     * @param binaryRow Binary row.
+     * @param rowId Row id.
+     */
+    public BinaryRowWithRowId(@Nullable BinaryRow binaryRow, RowId rowId) {
+        this.binaryRow = binaryRow;
+        this.rowId = rowId;
+    }
+
+    public BinaryRow binaryRow() {
+        return binaryRow;
+    }
+
+    public RowId rowId() {
+        return rowId;
     }
 }
--- a/modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageConcurrencyTest.java
+++ b/modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageConcurrencyTest.java
@@ -22,9 +22,13 @@ import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 import org.apache.ignite.internal.hlc.HybridTimestamp;
+import org.apache.ignite.internal.schema.BinaryRow;
 import org.apache.ignite.internal.storage.impl.TestStorageEngine;
+import org.jetbrains.annotations.Nullable;
 import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
 
 /**
  * Test to check for race conditions in MV partition storage.
@@ -34,7 +38,7 @@ public abstract class AbstractMvPartitionStorageConcurrencyTest extends BaseMvPa
     private static final int REPEATS = 100;
 
     @Test
-    void testAbortAndRead() throws Exception {
+    void testAbortAndRead() {
         for (int i = 0; i < REPEATS; i++) {
             addWrite(ROW_ID, BINARY_ROW, TX_ID);
 
@@ -49,7 +53,7 @@ public abstract class AbstractMvPartitionStorageConcurrencyTest extends BaseMvPa
     }
 
     @Test
-    void testCommitAndRead() throws Exception {
+    void testCommitAndRead() {
         for (int i = 0; i < REPEATS; i++) {
             addWrite(ROW_ID, BINARY_ROW, TX_ID);
 
@@ -64,7 +68,7 @@ public abstract class AbstractMvPartitionStorageConcurrencyTest extends BaseMvPa
     }
 
     @Test
-    void testUpdateAndRead() throws Exception {
+    void testUpdateAndRead() {
         for (int i = 0; i < REPEATS; i++) {
             addWrite(ROW_ID, BINARY_ROW, TX_ID);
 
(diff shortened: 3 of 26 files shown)


modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
```diff
@@ -204,7 +215,18 @@ public void commitWrite(RowId rowId, HybridTimestamp timestamp) {
...
-            return VersionChain.forCommitted(timestamp, versionChain);
+            VersionChain committedVersionChain = VersionChain.forCommitted(timestamp, versionChain);
+
+            if (committedVersionChain.next != null) {
+                // Avoid creating tombstones for tombstones.
+                if (committedVersionChain.row == null && committedVersionChain.next.row == null) {
+                    return committedVersionChain.next;
+                }
+
+                gcQueue.add(new IgniteBiTuple<>(committedVersionChain, rowId));
```
rpuch, line 226: `ConcurrentSkipListMap#compute()` may invoke the remapping function more than once due to concurrency. Here, the remapping function has a side effect (addition to the queue). It seems that it can result in multiple insertion to the queue. `VersionChain` does not define `equals()/hashCode()`, so this will really result in a chain being stored twice in the GC queue.

rpuch: Also, access to `gcQueue` is not synchronized explicitly. `ConcurrentMap` guarantees happens-before between actions with the map and actions around these actions (before or after them in the Program Order), but does it make any guarantees about what happens inside remapping functions? Documentation on `ConcurrentSkipListMap#compute()` specifically says that it does NOT guarantee atomicity for remapping functions.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
```diff
@@ -426,6 +448,51 @@ public ReadResult next() {
         return map.ceilingKey(lowerBound);
     }
 
+    @Override
+    public @Nullable BinaryRowWithRowId pollForVacuum(HybridTimestamp lowWatermark) {
+        Iterator<IgniteBiTuple<VersionChain, RowId>> it = gcQueue.iterator();
```
rpuch, line 453: No synchronization whatsoever. How about changing `gcQueue` to `ConcurrentSkipListSet`?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageGcTest.java
```diff
@@ -0,0 +1,132 @@
...
+        // Poll the next oldest row.
+        row = pollForVacuum(lowWatermark);
+
+        assertNotNull(row);
+        assertRowMatches(row.binaryRow(), binaryRow2);
+
+        // Nothing else to poll.
+        assertNull(pollForVacuum(lowWatermark));
+    }
+}
```
rpuch, line 132: Let's add a test (or a few) that models the real threading model that will be used in production code: one thread modifies the storage, another one acts as a GC. This might help catch visibility bugs (if, for some reason, `pollForVacuum()` never sees changes made by other methods).

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/BinaryRowWithRowId.java
```diff
@@ -0,0 +1,51 @@
...
+
+package org.apache.ignite.internal.storage;
+
+import org.apache.ignite.internal.schema.BinaryRow;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Wrapper that holds both {@link BinaryRow} and {@link RowId}.
+ */
+public class BinaryRowWithRowId {
```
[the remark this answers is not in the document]
ibessonov: Ok, why not

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
```diff
@@ -766,4 +771,43 @@ public static <T> T await(CompletionStage<T> stage, long timeout, TimeUnit unit)
     public static <T> T await(CompletionStage<T> stage) {
         return await(stage, TIMEOUT_SEC, TimeUnit.SECONDS);
     }
+
+    /**
+     * Runs all actions, each in a separate thread, having a {@link CyclicBarrier} before calling {@link RunnableX#run()}.
+     *
+     * @throws InterruptedException If failed to {@link Thread#join()} a thread.
+     */
+    public static void startRace(RunnableX... actions) throws InterruptedException {
```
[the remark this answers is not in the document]
ibessonov: Good

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageConcurrencyTest.java
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+
+    @Test
+    void testUpdateAndRead() throws Exception {
+        for (int i = 0; i < REPEATS; i++) {
+            addWrite(ROW_ID, BINARY_ROW, TX_ID);
+
+            startRace(
+                    () -> addWrite(ROW_ID, BINARY_ROW, TX_ID),
```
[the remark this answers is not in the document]
ibessonov: That would be a different scenario, right? Here I specifically want to update existing write intent.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageConcurrencyTest.java
```diff
@@ -0,0 +1,165 @@
...
+            );
+
+            assertRowMatches(read(ROW_ID, clock.now()), BINARY_ROW);
+        }
+    }
+
+    @Test
+    void testRegularGcAndRead() throws Exception {
+        //TODO https://issues.apache.org/jira/browse/IGNITE-18020
+        assumeTrue(engine instanceof TestStorageEngine);
```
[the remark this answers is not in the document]
ibessonov: I don't want it to fail on TC, and overriding it with disabling would mean that these tests must be protected, not package-private, like the rest of them, I didn't want them to be different.
I can replace it with overrides if you would like it more

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageConcurrencyTest.java
```diff
@@ -0,0 +1,165 @@
...
+import org.apache.ignite.internal.hlc.HybridTimestamp;
+import org.apache.ignite.internal.storage.impl.TestStorageEngine;
+import org.junit.jupiter.api.RepeatedTest;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test to check for race conditions in MV partition storage.
+ */
+public abstract class AbstractMvPartitionStorageConcurrencyTest extends BaseMvPartitionStorageTest {
+    /** To be used in a loop. {@link RepeatedTest} has a smaller failure rate for some reasons. */
```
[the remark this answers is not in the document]
ibessonov: Yeah, probably

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageConcurrencyTest.java
```diff
@@ -0,0 +1,165 @@
...
+                    () -> scanFirstEntry(clock.now())
+            );
+
+            assertNull(read(ROW_ID, clock.now()));
+        }
+    }
+
+    @Test
+    void testCommitAndRead() throws Exception {
+        for (int i = 0; i < REPEATS; i++) {
```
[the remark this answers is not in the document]
ibessonov: I thought about it, it's not a big deal in this test.
Tests that need cleanup to work properly, do it at the end of every iteration.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageConcurrencyTest.java
```diff
@@ -0,0 +1,165 @@
...
+        assumeTrue(engine instanceof TestStorageEngine);
+
+        for (int i = 0; i < REPEATS; i++) {
+            addAndCommit(BINARY_ROW);
+
+            addAndCommit(null);
+
+            startRace(
+                    () -> pollForVacuum(HybridTimestamp.MAX_VALUE),
+                    () -> addWrite(ROW_ID, BINARY_ROW, TX_ID)
```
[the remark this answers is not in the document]
ibessonov: Not necessarily, we don't compare them anyway, because that's too wasteful

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageGcTest.java
```diff
@@ -0,0 +1,132 @@
...
+        assertNull(storage.pollForVacuum(clock.now()));
+    }
+
+    @Test
+    void testRegularPoll() {
+        HybridTimestamp firstCommitTs = addAndCommit(BINARY_ROW);
+
+        HybridTimestamp tsBetweenCommits = clock.now();
+
+        HybridTimestamp secondCommitTs = addAndCommit(BINARY_ROW);
```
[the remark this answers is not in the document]
ibessonov: Ok, I'll do it in places where we compare values

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageGcTest.java
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.ignite.internal.configuration.testframework.ConfigurationExtension;
+import org.apache.ignite.internal.hlc.HybridTimestamp;
+import org.apache.ignite.internal.schema.BinaryRow;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/**
+ * Abstract test for MV partition storage GC.
+ */
+@ExtendWith(ConfigurationExtension.class)
```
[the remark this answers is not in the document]
ibessonov: I forgot to remove it apparently, thank you!

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
ibessonov: My bad, I'll fix it

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
ibessonov: Sure, thank you!

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageGcTest.java
ibessonov: I should add more scenarios to AbstractMvPartitionStorageConcurrencyTest, is this what you mean? Yeah, I guess so, let's be paranoid, I like that

ibessonov: Scenarios that are closer to the real usages will be tested later, when we implement the component that actually calls this new method.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
ibessonov: You know what, gcQueue is defined with a custom comparator, right? I guess equality is not an issue and hashCode is never called, so writing to the queue several times should be fine.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageConcurrencyTest.java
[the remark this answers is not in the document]
rpuch: My concern is about a different row (different value), but with same rowId. This would update the existing write intent, but it will change its value. In the current test the value after change is same, so it's impossible to say whether the update was applied or not.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageConcurrencyTest.java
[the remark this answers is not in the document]
rpuch: Oh, it's 'assume', not 'assert'. My bad.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
rpuch: That is true, comparator should remove this '2 entries' problem. But there is still '2 writes to the queue' problem.

Imagine that:

1. `map.compute()` makes first attempt, adds an entry to the queue
2. GC thread calls `pollForVacuum()` and removes this entry
3. `map.compute()` makes a second attempt and adds a second entry (equal to the first one) for the second time
4. GC thread polls again, so it sees the same (actually, a different, but equal) entry for the second time

Is this possible? Is this valid?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageGcTest.java
rpuch: Sure, I'm ok with `AbstractMvPartitionStorageConcurrencyTest`

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
ibessonov: GC cannot remove this entry, we only remove "old" entries, and even they must have a "parent" in the chain. Situation that you're describing is only possible in tests, and I don't see a necessity to cover it

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageConcurrencyTest.java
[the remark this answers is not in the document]
ibessonov: Ah, ok, I see

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
rpuch: Continuing the example. We now have 2 identical entries in the GC queue. When the watermark is high enough, GC grabs first of them, nullifies the pointer to the version it shadows, but this changes nothing from the point of view of the user, because this entry is not included in the chain. So GC vacuumed something, but in reality it did not. This might cause confusing failures in tests.

Why don't we just add synchronization to make sure it behaves as predictable as possible? This is a test storage, after all. If we must have a trade-off between performance and simplicity/predictability, we should choose the latter.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvPartitionStorageTest.java
```diff
@@ -47,9 +47,9 @@ public abstract class BaseMvPartitionStorageTest extends BaseMvStoragesTest {
 
     protected static final TestKey KEY = new TestKey(10, "foo");
 
-    protected static final TestValue VALUE = new TestValue(20, "bar");
+    protected static final BinaryRow BINARY_ROW = binaryRow(KEY, new TestValue(20, "bar"));
 
-    protected static final BinaryRow BINARY_ROW = binaryRow(KEY, VALUE);
+    protected static final BinaryRow BINARY_ROW2 = binaryRow(KEY, new TestValue(20, "bar"));
```
rpuch, line 52: It looks like the values of `BINARY_ROW` and `BINARY_ROW2` are identical. They should probably be different.

rpuch (commented):
Just one comment more

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
ibessonov: > We now have 2 identical entries in the GC queue.

There can be no duplicates in the queue. Anyway, I think I found a different example where the map itself can break regardless of the queue.

> Why don't we just add synchronization to make sure it behaves as predictable as possible?

I think I will

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvPartitionStorageTest.java
ibessonov: Copy-paste, sorry

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageConcurrencyTest.java
```diff
@@ -0,0 +1,240 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
+
+/**
+ * Test to check for race conditions in MV partition storage.
+ */
+public abstract class AbstractMvPartitionStorageConcurrencyTest extends BaseMvPartitionStorageTest {
+    /** To be used in a loop. {@link RepeatedTest} has a smaller failure rate due to recreating the storage every time. */
+    private static final int REPEATS = 100;
```
tkalkirill, line 38: Maybe use **org.junit.jupiter.api.RepeatedTest**?

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
```diff
@@ -766,4 +771,57 @@ public static <T> T await(CompletionStage<T> stage, long timeout, TimeUnit unit)
...
+        } catch (InterruptedException e) {
+            for (Thread thread : threads) {
+                thread.interrupt();
+            }
+
+            fail("Race operations took too long.");
+        }
+
+        if (!throwables.isEmpty()) {
+            AssertionError assertionError = new AssertionError("One or several threads have failed.");
```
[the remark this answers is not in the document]
ibessonov: Ok

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageConcurrencyTest.java
ibessonov: I explicitly said that it has smaller fail rate, which is bad for such tests.

tkalkirill (approved):

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageConcurrencyTest.java
```diff
@@ -162,4 +217,24 @@ private void cleanup() {
...
+                return ts;
+            }
+        },
+        NON_ATOMIC {
+            @Override
+            HybridTimestamp perform(AbstractMvPartitionStorageConcurrencyTest test, @Nullable BinaryRow binaryRow) {
+                return test.addAndCommit(binaryRow);
+            }
+        };
+        abstract HybridTimestamp perform(AbstractMvPartitionStorageConcurrencyTest test, @Nullable BinaryRow binaryRow);
```
rpuch, line 238: Blank line above?

rpuch (approved):

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageConcurrencyTest.java
ibessonov: I wonder why I have a green PMD check...

tkalkirill: Are you using Gradle?

rpuch (approved):

PR #1467, opened 2022-12-22
IGNITE-18057 Fixed index scan transactional consistency.

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-18057

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
@@ -46,6 +46,7 @@ import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
+import org.apache.ignite.internal.binarytuple.BinaryTupleCommon;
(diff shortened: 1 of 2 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -697,15 +699,38 @@ private CompletableFuture<List<BinaryRow>> scanSortedIndex(
...
                 Cursor<IndexRow> cursor = (Cursor<IndexRow>) cursors.computeIfAbsent(cursorId,
                         id -> {
-                            // TODO https://issues.apache.org/jira/browse/IGNITE-18057
-                            // Fix scan cursor return item closet to lowerbound and <= lowerbound
-                            // to correctly lock range between lowerbound value and the item next to lowerbound.
                             return indexStorage.scan(
                                     lowerBound,
-                                    // We need upperBound next value for correct range lock.
-                                    upperBound,
+                                    // We have to handle upperBound on a level of replication listener, to correct to take a range lock.
```
denis-chudov, line 732:
```suggestion
                                    // We have to handle upperBound on a level of replication listener, for correctness of taking of a range lock.
```

PR #1469, opened 2022-12-22
IGNITE-18227: refactoring scan nodes and add support RO index scans.

ygerzhedovich:


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlAsynchronousApiTest.java
```diff
@@ -343,37 +346,59 @@ public void checkTransactionsWithDml() throws Exception {
...
     }
 
-    private void checkTx(Session ses, boolean readOnly, boolean commit, boolean explicit) throws Exception {
+    private void checkTx(Session ses, boolean readOnly, boolean commit, boolean explicit, Matcher<String> planMatcher) throws Exception {
         Transaction outerTx = explicit ? (readOnly ? igniteTx().readOnly().begin() : igniteTx().begin()) : null;
 
-        AsyncResultSet rs = ses.executeAsync(outerTx, "SELECT VAL0 FROM TEST ORDER BY VAL0").get();
+        String query = "SELECT VAL0 FROM TEST ORDER BY VAL0";
+
+        assertQuery(query).matches(planMatcher).check();
```
korlov42, line 399: `assertQuery(query).matches(planMatcher).check();` and `ses.executeAsync(outerTx, query).get();` return not quite consistent results. The former still uses `queryAsync` under the hood, whereas the latter uses `querySingleAsync`. I think, we need o update QueryChecker. Besides, `outerTx` should be passed to the checker in order to provided valid plan

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlAsynchronousApiTest.java
```diff
@@ -343,37 +346,59 @@ public void checkTransactionsWithDml() throws Exception {
...
+
+        Matcher<String> planMatcher = containsTableScan("PUBLIC", "TEST");
+
+        checkMixedTransactions(planMatcher);
+    }
+
+
+    /** Check correctness of rw and ro transactions for table scan. */
+    @Test
+    public void checkMixedTransactionsForTable() throws Exception {
```
korlov42, line 362: looks like the name of this test and one above should be switched

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/StorageScanNode.java
```diff
@@ -0,0 +1,303 @@
...
+import java.util.function.Predicate;
+import org.apache.ignite.internal.schema.BinaryRow;
+import org.apache.ignite.internal.sql.engine.exec.ExecutionContext;
+import org.apache.ignite.internal.sql.engine.exec.RowHandler;
+import org.apache.ignite.internal.sql.engine.schema.InternalIgniteTable;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Based abstract scan node.
```
korlov42, line 38: Could you please add more details to the javadoc? I would shed some light on what ScanNode is for, and what logic it encapsulates

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/StorageScanNode.java
```diff
@@ -0,0 +1,303 @@
...
+    protected abstract Publisher<RowT> scan();
+
+    /**
+     * Proxy publisher with singe goal convert rows from {@code BinaryRow} to {@code RowT}.
+     *
+     * @param pub {@code BinaryRow} Publisher.
+     *
+     * @return Proxy publisher with conversion from {@code BinaryRow} to {@code RowT}.
+     */
+    @NotNull
```
korlov42, line 140: as far as I know, we are not using `@NotNull`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/StorageScanNode.java
```diff
@@ -0,0 +1,303 @@
...
+    public Publisher<RowT> convertPublisher(Publisher<BinaryRow> pub) {
```
korlov42, line 141: do we really need to have this method `public`?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/TableScanNode.java
```diff
@@ -19,66 +19,44 @@
...
-    private final @Nullable BitSet requiredColumns;
-
-    private int requested;
-
-    private int waiting;
-
-    private boolean inLoop;
-
-    private Subscription activeSubscription;
+    boolean dataRequested;
```
korlov42, line 47: this may be `private`, as well as `curPartIdx`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/StorageScanNode.java
```diff
@@ -0,0 +1,303 @@
...
+            activeSubscription = null;
+        }
+    }
+
+    /**
+     *  Publisher of scan node.
+     *
+     *  @return Publisher of scan node or {@code null} in case nothing to scan.
+     */
+    protected abstract Publisher<RowT> scan();
```
korlov42, line 131:
```suggestion
    protected abstract @Nullable Publisher<RowT> scan();
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/StorageScanNode.java
```diff
@@ -0,0 +1,303 @@
...
+
+    private final Function<BinaryRow, RowT> tableRowConverter;
+
+    private int requested;
+
+    private int waiting;
+
+    private boolean inLoop;
+
+    private Subscription activeSubscription;
```
korlov42, line 58:
```suggestion
    private @Nullable Subscription activeSubscription;
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/StorageScanNode.java
korlov42, line 131:
```suggestion
    protected abstract @Nullable Publisher<RowT> scan();
```

Contract of this method is not clear. The fact you need to scan several times in order to scan the whole storage once is counterintuitive

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/IndexScanNode.java
```diff
@@ -52,13 +48,9 @@
 import org.jetbrains.annotations.Nullable;
 
 /**
- * Scan node.
- * TODO: merge with {@link TableScanNode}
+ * Index scan node.
```
korlov42, line 51: could you please add more details to the javadoc?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/TableScanNode.java
```diff
@@ -19,66 +19,44 @@
...
 import org.apache.ignite.internal.sql.engine.exec.ExecutionContext;
 import org.apache.ignite.internal.sql.engine.exec.RowHandler;
 import org.apache.ignite.internal.sql.engine.schema.InternalIgniteTable;
+import org.apache.ignite.internal.sql.engine.util.CompositePublisher;
 import org.apache.ignite.internal.table.InternalTable;
 import org.jetbrains.annotations.Nullable;
 
 /**
- * Scan node.
+ * Table scan node.
```
korlov42, line 38: the same about details

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/StorageScanNode.java
```diff
@@ -0,0 +1,303 @@
...
+     * @param filters Optional filter to filter out rows.
+     * @param rowTransformer Optional projection function.
+     */
+    public StorageScanNode(
+            ExecutionContext<RowT> ctx,
+            RowHandler.RowFactory<RowT> rowFactory,
+            InternalIgniteTable schemaTable,
+            @Nullable Predicate<RowT> filters,
+            @Nullable Function<RowT, RowT> rowTransformer,
+            @Nullable BitSet requiredColumns
```
korlov42, line 75: param description is missing

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/StorageScanNode.java
ygerzhedovich: It is temporal solution due to IGNITE-18466. After fix need to change return type to CompositePublisher and the method will be invoked just once. However you are right the require shed lights for the current situation.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlAsynchronousApiTest.java
ygerzhedovich: thanks

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlAsynchronousApiTest.java
ygerzhedovich: good catch

zstan:
looks good, the only note : instead StorageScanNode i will prefer AbstractScanNode, but i can`t debate here, it`s up to you.

zstan:
I don`t like new approach, cause :
1. In case of index scan with concrete bounds (check ItSecondaryIndexTest#testSelectWithRanges) we call StorageScanNode#convertPublisher -> Subscriber#onNext  more than 5 times, while in previous approach it raises only 5 (limit param) times and as a consequence we call StorageScanNode#convert more than 5 times. 
2. we still have test hanging: 
image

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/TableScanNode.java
```diff
@@ -90,190 +64,33 @@ public TableScanNode(
...
-        Subscription subscription = this.activeSubscription;
-        if (subscription != null) {
-            subscription.request(waiting);
-        } else if (curPartIdx < parts.length) {
-            if (context().transactionTime() != null) {
-                physTable.scan(parts[curPartIdx++], context().transactionTime(), context().localNode()).subscribe(new SubscriberImpl());
+        for (int p : parts) {
+            Publisher<BinaryRow> pub;
+            if (roTx) {
+                pub = physTable.scan(p, context().transactionTime(), context().localNode());
```
korlov42, line 85: I'm not sure it's a good idea to open cursors to all partitions at the same time

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlAsynchronousApiTest.java
```diff
@@ -343,37 +346,58 @@ public void checkTransactionsWithDml() throws Exception {
...
 
-        checkTx(ses, true, false, true);
-        checkTx(ses, true, false, false);
-        checkTx(ses, true, true, true);
-        checkTx(ses, true, true, false);
-        checkTx(ses, false, true, true);
-        checkTx(ses, false, true, false);
-        checkTx(ses, false, false, true);
-        checkTx(ses, false, false, false);
+        for (int roTx = 0; roTx < 2; roTx++) {
```
AMashenkov, line 384:
```suggestion
        for (boolean roTx : List.of(Boolean.TRUE, Boolean.FALSE)) {
```

ygerzhedovich: fixed

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/IndexScanNode.java
```diff
@@ -119,282 +91,122 @@ public IndexScanNode(
...
 
-            indexPublisher(parts, cond).subscribe(new SubscriberImpl());
+            return SubscriptionUtils.concat(conditionPublishers);
         } else {
-            waiting = NOT_WAITING;
+            return indexPublisher(parts, null);
         }
     }
 
     private Publisher<RowT> indexPublisher(int[] parts, @Nullable RangeCondition<RowT> cond) {
```
korlov42, line 128: I would suggest to use `TransformingIterator` here. It will make code much cleaner

ygerzhedovich: thanks, fixed

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/IndexScanNode.java
```diff
@@ -119,282 +91,122 @@ public IndexScanNode(
...
-                cond = rangeConditionIterator.next();
-
-                rangeConditionsProcessed = !rangeConditionIterator.hasNext();
+            int i = 0;
+            while (it.hasNext()) {
+                conditionPublishers[i++] = indexPublisher(parts, it.next());
             }
 
-            indexPublisher(parts, cond).subscribe(new SubscriberImpl());
+            return SubscriptionUtils.concat(conditionPublishers);
```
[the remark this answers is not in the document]
ygerzhedovich: thanks, fixed

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/AbstractBasicIntegrationTest.java
```diff
@@ -227,8 +227,25 @@ public void tearDown(TestInfo testInfo) throws Exception {
         tearDownBase(testInfo);
     }
 
+    /**
+     * Start execute query and check all passed to the builder asserts.
```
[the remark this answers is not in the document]
ygerzhedovich: reworked

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/SubscriptionUtils.java
```diff
@@ -49,6 +50,31 @@ public static <T> Publisher<T> concat(Publisher<? extends T>... sources) {
...
 
+    /**
+     * Create thread-safe publisher wrapper of combine multiple publishers, sources for whom are suppliers what can be used for lazy
+     * opening these publishers . Generally, start consuming a source once the previous source has terminated, building a chain.
+     *
+     * @param sources Array of publisher suppliers which should be combine.
+     * @return Publisher which will be combine all of passed as parameter to single one.
+     */
+    @SafeVarargs
+    public static <T> Publisher<T> concat(Supplier<Publisher<T>>... sources) {
```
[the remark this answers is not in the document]
ygerzhedovich: thanks, reworked

korlov42 (approved):

PR #1473, opened 2022-12-23
IGNITE-18220: Implement integration tests for completers

PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-18220

### DIFF
--- a/modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/CliCommandTestInitializedIntegrationBase.java
+++ b/modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/CliCommandTestInitializedIntegrationBase.java
@@ -19,6 +19,7 @@ package org.apache.ignite.internal.cli.commands;
 
 import static org.apache.ignite.internal.testframework.IgniteTestUtils.testNodeName;
 
+import java.util.List;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.TestInfo;
 
@@ -31,10 +32,14 @@ public class CliCommandTestInitializedIntegrationBase extends CliCommandTestNotI
     @Override
     void beforeAll(TestInfo testInfo) {
         startNodes(testInfo);
-        super.initializeCluster(metaStorageNodeName(testInfo));
+        initializeCluster(metaStorageNodeName(testInfo));
     }
 
     protected String metaStorageNodeName(TestInfo testInfo) {
         return testNodeName(testInfo, 0);
+    }
+
+    protected List<String> allNodeNames() {
+        return CLUSTER_NODE_NAMES;
     }
 }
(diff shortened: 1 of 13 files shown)


modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/CliCommandTestNotInitializedIntegrationBase.java
```diff
@@ -106,6 +107,10 @@ protected void execute(String... args) {
         exitCode = cmd.execute(args);
     }
 
+    protected CommandLine getCmd() {
```
Flaugh24, line 110: Why do you need this method? You can just make cmd protected.

PakhomovAlexander: I wouldn't like to give an access to fields event for children. But I could rename the method to `cmd()`

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/CliCommandTestInitializedIntegrationBase.java
```diff
@@ -31,10 +32,14 @@ public class CliCommandTestInitializedIntegrationBase extends CliCommandTestNotI
...
+        initializeCluster(metaStorageNodeName(testInfo));
     }
 
     protected String metaStorageNodeName(TestInfo testInfo) {
         return testNodeName(testInfo, 0);
     }
+
+    protected List<String> allNodeNames() {
+        return CLUSTER_NODE_NAMES;
+    }
```
Pochatkin, line 44: Why is it here? This static field in IgniteTestBase has protected modifier and you expand it by NON static protected method in some child class, absolutely don't understand what the reason

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/CliCommandTestNotInitializedIntegrationBase.java
```diff
@@ -77,6 +77,7 @@ public class CliCommandTestNotInitializedIntegrationBase extends IntegrationTest
     @BeforeEach
     public void setUp(TestInfo testInfo) throws Exception {
         super.setUp(testInfo);
+
```
Pochatkin, line 80: Why empty line?

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/completer/NodeUrlProvider.java
```diff
@@ -40,7 +40,7 @@ public NodeUrlProvider(Session session, ConfigManagerProvider configManagerProvi
     }
 
     /** Resolves the url for given words. */
-    public String resolveUrl(String[] words) {
+    public String resolveUrl(String... words) {
```
Pochatkin, line 43: What the point of vararg here? So, lets imagine that its required change but why you didn't change usages of this method. For example org/apache/ignite/internal/cli/core/repl/completer/hocon/ClusterConfigDynamicCompleterFactory.java:58 still use String array initialization

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/CliCommandTestInitializedIntegrationBase.java
PakhomovAlexander: Agreed to add the `static` modifier to the method signature.

Pochatkin:
LGTM

PR #1475, opened 2022-12-23
IGNITE-17467 Return RowIds in MvPartitionStorage#scan() results

rpuch:
https://issues.apache.org/jira/browse/IGNITE-17467


SammyVimes (approved):
LGTM

PR #1476, opened 2022-12-26
IGNITE-18442 Sql. JOIN with DISTINCT FROM fails.

zstan:


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItJoinTest.java
```diff
@@ -784,34 +790,66 @@ public void testNatural(JoinType joinType) {
...
+        } finally {
+            sql("DROP TABLE IF EXISTS t11");
+            sql("DROP TABLE IF EXISTS t22");
+        }
+    }
 
-        private final String[] disabledRules;
+    private static Stream<Arguments> joinTypes() {
+        List<Arguments> types = Arrays.stream(JoinType.values())
+                        .flatMap(v -> Stream.of(Arguments.of(v, false), Arguments.of(v, true))).collect(Collectors.toList());
```
ygerzhedovich, line 851: you can just return stream here without collect result to list.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/RexUtils.java
```diff
@@ -197,7 +199,7 @@ public static boolean isIdentity(List<? extends RexNode> projects, RelDataType i
 
     /** Binary comparison operations. */
     private static final Set<SqlKind> BINARY_COMPARISON =
```
ygerzhedovich, line 201: why we can't use org.apache.calcite.sql.SqlKind#BINARY_COMPARISON ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/RexUtils.java
```diff
@@ -361,46 +364,58 @@ public static List<SearchBounds> buildHashIndexConditions(
...
             RelOptCluster cluster,
             RexNode condition,
-            RelDataType rowType
+            RelDataType rowType,
+            @Nullable ImmutableBitSet requiredColumns
     ) {
         condition = RexUtil.toCnf(builder(cluster), condition);
 
-        Int2ObjectMap<List<RexCall>> fieldsToPredicates = mapPredicatesToFields(condition, cluster);
+        Map<Integer, List<RexCall>> fieldsToPredicates = mapPredicatesToFields(condition, cluster);
```
ygerzhedovich, line 377: why you replace it?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/RexUtils.java
```diff
@@ -361,46 +364,58 @@ public static List<SearchBounds> buildHashIndexConditions(
...
+
+        List<RelDataType> types = RelOptUtil.getFieldTypeList(rowType);
+
+        Mappings.TargetMapping mapping = null;
+
+        if (requiredColumns != null) {
+            mapping = Commons.inverseTrimmingMapping(types.size(), requiredColumns);
+        }
+
+        for (int fldIdx : fieldsToPredicates.keySet()) {
```
ygerzhedovich, line 393: Looks like int2ObjectEntrySet() more preferable for the case

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/RexUtils.java
```diff
@@ -361,46 +364,58 @@ public static List<SearchBounds> buildHashIndexConditions(
...
+            List<RexCall> collFldPreds = fieldsToPredicates.get(fldIdx);
 
-        for (List<RexCall> collFldPreds : fieldsToPredicates.values()) {
             if (nullOrEmpty(collFldPreds)) {
                 break;
             }
 
             for (RexCall pred : collFldPreds) {
-                if (pred.getOperator().kind != SqlKind.EQUALS) {
+                if (pred.getOperator().kind != SqlKind.EQUALS && pred.getOperator().kind != IS_NOT_DISTINCT_FROM) {
```
ygerzhedovich, line 401: let's introduce colletion for both values and add explanation why we need it.

zstan: don`t understand here, for hash search we can use only **equality** or **is not distinct** operands, plz explain your idea.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/RexUtils.java
zstan: cause planner generates correlates with IS NOT DISTINCT FROM form, check for example :
AggregatePlannerTest#expandDistinctAggregates

zstan:
@ygerzhedovich thanks for review, partially fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/RexUtils.java
zstan: pred.isA

zstan:
@ygerzhedovich all done.

zstan:
@ygerzhedovich all done

PR #1478, opened 2022-12-26
IGNITE-18024 Implementation of a full rebalance for TxStateRocksDbStorage on receiver

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18024

### DIFF
--- a/modules/rocksdb-common/src/main/java/org/apache/ignite/internal/rocksdb/BusyRocksIteratorAdapter.java
+++ b/modules/rocksdb-common/src/main/java/org/apache/ignite/internal/rocksdb/BusyRocksIteratorAdapter.java
@@ -75,7 +75,7 @@ public abstract class BusyRocksIteratorAdapter<T> extends RocksIteratorAdapter<T
     @Override
     public T next() {
         if (!busyLock.enterBusy()) {
-            handleBusyFail();
+            handleBusyFail0();
         }
 
         try {
@@ -90,7 +90,7 @@ public abstract class BusyRocksIteratorAdapter<T> extends RocksIteratorAdapter<T
     @Override
     public void close() {
         if (!busyLock.enterBusy()) {
-            handleBusyFail();
+            handleBusyFail0();
         }
 
         try {
--- a/modules/rocksdb-common/src/main/java/org/apache/ignite/internal/rocksdb/BusyRocksIteratorAdapter.java
+++ b/modules/rocksdb-common/src/main/java/org/apache/ignite/internal/rocksdb/BusyRocksIteratorAdapter.java
@@ -42,10 +42,17 @@ public abstract class BusyRocksIteratorAdapter<T> extends RocksIteratorAdapter<T
     /**
      * Handles busy lock acquiring failure. This means that db has been stopped and cursor can't proceed. Must throw an exception.
      */
-    protected abstract void handleBusy();
+    protected abstract void handleBusyFail();
 
-    private void handleBusy0() {
-        handleBusy();
+    /**
+     * Handles busy lock acquiring success.
+     */
+    protected void handeBusySuccess() {
+        // No-op.
+    }
+
+    private void handleBusyFail0() {
+        handleBusyFail();
 
         assert false : "handleBusy() should have thrown an exception.";
     }
@@ -53,10 +60,12 @@ public abstract class BusyRocksIteratorAdapter<T> extends RocksIteratorAdapter<T
     @Override
     public boolean hasNext() {
         if (!busyLock.enterBusy()) {
-            handleBusy0();
+            handleBusyFail0();
         }
 
         try {
+            handeBusySuccess();
+
             return super.hasNext();
         } finally {
             busyLock.leaveBusy();
@@ -66,10 +75,12 @@ public abstract class BusyRocksIteratorAdapter<T> extends RocksIteratorAdapter<T
     @Override
     public T next() {
         if (!busyLock.enterBusy()) {
-            handleBusy0();
+            handleBusyFail();
         }
 
         try {
+            handeBusySuccess();
+
             return super.next();
         } finally {
             busyLock.leaveBusy();
@@ -79,10 +90,12 @@ public abstract class BusyRocksIteratorAdapter<T> extends RocksIteratorAdapter<T
     @Override
     public void close() {
         if (!busyLock.enterBusy()) {
-            handleBusy0();
+            handleBusyFail();
         }
 
         try {
+            handeBusySuccess();
+
             super.close();
         } finally {
             busyLock.leaveBusy();
--- a/modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
+++ b/modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
@@ -526,41 +526,47 @@ public class TxStateRocksDbStorage implements TxStateStorage {
             throw createStorageInProgressOfRebalanceException();
         }
 
-        try {
-            if (!STATE.compareAndSet(this, StorageState.RUNNABLE, StorageState.REBALANCE)) {
-                throwExceptionIfStorageClosedOrRebalance();
-            }
-
-            busyLock.block();
-
-            try (WriteBatch writeBatch = new WriteBatch()) {
-                writeBatch.deleteRange(partitionStartPrefix(), partitionEndPrefix());
-                writeBatch.put(lastAppliedIndexAndTermKey, indexAndTermToBytes(REBALANCE_IN_PROGRESS, REBALANCE_IN_PROGRESS));
-
-                db.write(writeOptions, writeBatch);
-
-                lastAppliedIndex = REBALANCE_IN_PROGRESS;
-                lastAppliedTerm = REBALANCE_IN_PROGRESS;
-                persistedIndex = REBALANCE_IN_PROGRESS;
-
-                rebalanceFuture.complete(null);
-
-                return rebalanceFuture;
-            } catch (Exception e) {
-                throw new IgniteInternalException(
-                        TX_STATE_STORAGE_REBALANCE_ERR,
-                        IgniteStringFormatter.format("Failed to clear storage for partition {} of table {}", partitionId, getTableName()),
-                        e
-                );
-            } finally {
-                busyLock.unblock();
-            }
-        } catch (IgniteInternalException e) {
-            rebalanceFuture.completeExceptionally(e);
+        if (!STATE.compareAndSet(this, StorageState.RUNNABLE, StorageState.REBALANCE)) {
+            IgniteInternalException exception = createExceptionIfStorageClosedOrRebalance();
 
             this.rebalanceFuture = null;
 
-            throw e;
+            rebalanceFuture.completeExceptionally(exception);
+
+            throw exception;
+        }
+
+        busyLock.block();
+
+        try (WriteBatch writeBatch = new WriteBatch()) {
+            writeBatch.deleteRange(partitionStartPrefix(), partitionEndPrefix());
+            writeBatch.put(lastAppliedIndexAndTermKey, indexAndTermToBytes(REBALANCE_IN_PROGRESS, REBALANCE_IN_PROGRESS));
+
+            db.write(writeOptions, writeBatch);
+
+            lastAppliedIndex = REBALANCE_IN_PROGRESS;
+            lastAppliedTerm = REBALANCE_IN_PROGRESS;
+            persistedIndex = REBALANCE_IN_PROGRESS;
+
+            rebalanceFuture.complete(null);
+
+            return rebalanceFuture;
+        } catch (Exception e) {
+            IgniteInternalException exception = new IgniteInternalException(
+                    TX_STATE_STORAGE_REBALANCE_ERR,
+                    IgniteStringFormatter.format("Failed to clear storage for partition {} of table {}", partitionId, getTableName()),
+                    e
+            );
+
+            this.rebalanceFuture = null;
+
+            state = StorageState.RUNNABLE;
+
+            rebalanceFuture.completeExceptionally(exception);
+
+            throw exception;
+        } finally {
+            busyLock.unblock();
         }
     }
 
@@ -659,16 +665,7 @@ public class TxStateRocksDbStorage implements TxStateStorage {
     }
 
     private void throwExceptionIfStorageClosedOrRebalance() {
-        StorageState state = this.state;
-
-        switch (state) {
-            case CLOSED:
-                throw createStorageClosedException();
-            case REBALANCE:
-                throw createStorageInProgressOfRebalanceException();
-            default:
-                throw createUnexpectedStorageStateException(state);
-        }
+        throw createExceptionIfStorageClosedOrRebalance();
     }
 
     private void throwExceptionIfStorageInProgressOfRebalance() {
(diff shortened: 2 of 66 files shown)


modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
```diff
@@ -57,6 +61,24 @@
  * Tx state storage implementation based on RocksDB.
  */
 public class TxStateRocksDbStorage implements TxStateStorage {
+    private static final VarHandle STATE;
+
+    private static final VarHandle REBALANCE_FUTURE;
+
+    static {
+        try {
+            STATE = MethodHandles.lookup().findVarHandle(TxStateRocksDbStorage.class, "state", StorageState.class);
```
ibessonov, line 70: What's the problem with using atomics?
Are you sure that placing these values inside of the object is beneficial? Do you have proofs?
I'd prefer a code simplicity if the advantage is not clear.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
```diff
@@ -93,8 +115,12 @@ public class TxStateRocksDbStorage implements TxStateStorage {
     /** The value of {@link #lastAppliedIndex} persisted to the device at this moment. */
     private volatile long persistedIndex;
 
-    /** Database key for the last applied index+term. */
-    private final byte[] lastAppliedIndexAndTermKey;
+    /** Current state of the storage. */
+    private volatile StorageState state = StorageState.RUNNABLE;
+
+    @Nullable
```
ibessonov, line 121: Usually you put Nullable next to the type. Why is it different this time?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
```diff
@@ -121,30 +147,28 @@ public TxStateRocksDbStorage(
...
-        persistedIndex = lastAppliedIndex;
+        initLastApplied();
     }
 
     @Override
     @Nullable
     public TxMeta get(UUID txId) {
         if (!busyLock.enterBusy()) {
-            throwStorageStoppedException();
+            throwExceptionIfStorageClosedOrRebalance();
```
ibessonov, line 157: What's up with the rebalance in the exception if the storage is closed?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
```diff
@@ -199,21 +221,22 @@ public boolean compareAndSet(UUID txId, @Nullable TxState txStateExpected, TxMet
...
+                        result = txMetaExisting.txState() == txMeta.txState()
+                                && Objects.equals(txMetaExisting.commitTimestamp(), txMeta.commitTimestamp());
                     }
                 } else {
                     result = false;
                 }
             }
 
-            writeBatch.put(lastAppliedIndexAndTermKey, indexAndTermToBytes(commandIndex, commandTerm));
+            if (state != StorageState.REBALANCE) {
```
ibessonov, line 232: I think places like this deserve comments

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
```diff
@@ -397,12 +429,18 @@ private long readLastAppliedIndex(ReadOptions readOptions) {
...
     public void destroy() {
-        try (WriteBatch writeBatch = new WriteBatch()) {
-            close();
+        if (!close0()) {
+            return;
+        }
 
-            writeBatch.deleteRange(partitionStartPrefix(), partitionEndPrefix());
+        try {
+            try (WriteBatch writeBatch = new WriteBatch()) {
```
ibessonov, line 437: Isn't a single try enough?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
```diff
@@ -445,32 +479,210 @@ private UUID keyToTxId(byte[] bytes) {
...
+
+            lastAppliedIndex = REBALANCE_IN_PROGRESS;
+            lastAppliedTerm = REBALANCE_IN_PROGRESS;
+            persistedIndex = REBALANCE_IN_PROGRESS;
+
+            CompletableFuture<Void> rebalanceFuture = completedFuture(null);
+
+            this.rebalanceFuture = rebalanceFuture;
+
+            return rebalanceFuture;
```
ibessonov, line 507: What is happening? If this future is always completed, why do you need a field for the future at all?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
```diff
@@ -445,32 +479,210 @@ private UUID keyToTxId(byte[] bytes) {
...
+        } catch (Exception e) {
+            throw new IgniteInternalException(
+                    TX_STATE_STORAGE_REBALANCE_ERR,
+                    IgniteStringFormatter.format("Failed to clear storage for partition {} of table {}", partitionId, getTableName()),
+                    e
+            );
+        } finally {
+            busyLock.unblock();
```
ibessonov, line 515: Why do you need a busy lock in this method?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
```diff
@@ -445,32 +479,210 @@ private UUID keyToTxId(byte[] bytes) {
...
-        // TODO: IGNITE-18024 Implement
-        throw new UnsupportedOperationException();
+    private void initLastApplied() {
+        byte[] indexAndTermBytes = readLastAppliedIndexAndTerm(readOptions);
+
+        if (indexAndTermBytes != null) {
+            long lastAppliedIndex = bytesToLong(indexAndTermBytes);
+
+            if (lastAppliedIndex == REBALANCE_IN_PROGRESS) {
+                try (WriteBatch writeBatch = new WriteBatch()) {
```
ibessonov, line 595: I wonder if there's a designated "start" method in the storage.
Doing a recovery in the constructor?... That's a bad idea. Reading data in constructor is also bad idea, but that was my fault I believe.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
```diff
@@ -445,32 +479,210 @@ private UUID keyToTxId(byte[] bytes) {
...
+            } else {
+                this.lastAppliedIndex = lastAppliedIndex;
+                persistedIndex = lastAppliedIndex;
+
+                lastAppliedTerm = bytesToLong(indexAndTermBytes, Long.BYTES);
+            }
+        }
+    }
+
+    private boolean close0() {
```
ibessonov, line 620: Not the best name for the method that performs so many complex actions.

modules/rocksdb-common/src/main/java/org/apache/ignite/internal/rocksdb/BusyRocksIteratorAdapter.java
```diff
@@ -42,21 +42,30 @@ protected BusyRocksIteratorAdapter(IgniteSpinBusyLock busyLock, RocksIterator it
...
      */
-    protected abstract void handleBusy();
+    protected abstract void handleBusyFail();
 
-    private void handleBusy0() {
-        handleBusy();
+    /**
+     * Handles busy lock acquiring success.
+     */
+    protected void handeBusySuccess() {
```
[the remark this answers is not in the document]
ibessonov: Ok, I found a usage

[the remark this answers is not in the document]
tkalkirill: This method is used here: `TxStateRocksDbStorage#scan`.
In method `scan` above, it is needed to check that we are not in a storage rebalance state after getting a "busy" lock.

Alternatively, we can get rid of method `handeBusySuccess` and use own cursor in `TxStateRocksDbStorage#scan' so as not to attach a crutch. I think it's better to do so.

wdyt?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
tkalkirill: Since there can be many partitions, we can, although a little, win in memory consumption for each using `VarHandle` instead of atomics. The less memory is used, the faster the GC runs, the faster the system runs.

Yes, it looks like premature optimization, but I don't use complex constructs to make the code much more complicated.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
tkalkirill: This is problem?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
tkalkirill: I did not understand the question.
If we weren't able to take the "busy" lock, then it's either because the storage is closed or it started rebalancing.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
tkalkirill: Fix it

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
tkalkirill: Because I want to make sure that when we start a storage rebalance, all operations on it will be stopped.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
tkalkirill: Added

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
tkalkirill: In general, I can get rid of `TxStateRocksDbStorage#rebalanceFuture`, and I will.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
tkalkirill: I'll move it to the "start" method.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
tkalkirill: Tried to rename and add documentation.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
tkalkirill: I'll try to change it up a bit.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
ibessonov: Ok, I give up. Generally speaking, less code is better than more code. Benefits of the memory saved is not clear at all.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
ibessonov: Just wondering, what's up with the inconsistency?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
ibessonov: Why do we perform "busyLockSuccess" check if rebalance is covered by the busy lock itself?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
```diff
@@ -479,40 +515,52 @@ private UUID keyToTxId(byte[] bytes) {
...
+            } catch (Exception e) {
+                throw new IgniteInternalException(
+                        TX_STATE_STORAGE_REBALANCE_ERR,
+                        IgniteStringFormatter.format("Failed to clear storage for partition {} of table {}", partitionId, getTableName()),
+                        e
+                );
+            } finally {
+                busyLock.unblock();
+            }
+        } catch (IgniteInternalException e) {
```
ibessonov, line 558: Is this the exception from the line 550? Are there other sources for such type of exceptions? What do we do with the state?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbTableStorage.java
```diff
@@ -153,6 +153,8 @@ public TxStateStorage getOrCreateTxStateStorage(int partitionId) {
                 partitionId,
                 this
             );
+
+            storage.start();
```
ibessonov, line 157: Technically, we should close the instance if there's an exception in "start()", right? We may fix that in the future. This is the reason why I don't like exceptions in constructors - they may easily produce leaks.

ibessonov (approved):
Looks good to me otherwise.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
tkalkirill: The "busy" lock only protects against the start of the rebalance, but not after.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbStorage.java
tkalkirill: This is the handling of exceptions from lines 531 and 550. Yes, it is necessary to deal with standing, I will correct it a little.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/storage/state/rocksdb/TxStateRocksDbTableStorage.java
tkalkirill: Ok.

PR #1479, opened 2022-12-27
IGNITE-18360 Migrate storage to new Binary Tuple format

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-18360


modules/schema/src/main/java/org/apache/ignite/internal/schema/TableRow.java
```diff
@@ -0,0 +1,105 @@
...
+    public int schemaVersion() {
+        return Short.toUnsignedInt(buf.getShort(SCHEMA_VERSION_OFFSET));
+    }
+
+    /** Get ByteBuffer slice representing the binary tuple. */
+    public ByteBuffer tupleSlice() {
+        try {
+            return buf.position(TUPLE_OFFSET).slice().order(ORDER);
+        } finally {
+            buf.position(0); // Reset bounds.
```
korlov42, line 86: we have to restore previous position here

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1931,19 +1944,19 @@ private HybridTimestampMessage hybridTimestamp(HybridTimestamp tmstmp) {
      *
      * @param tablePartId {@link TablePartitionId} object to construct {@link UpdateCommand} object with.
      * @param rowUuid Row UUID.
-     * @param rowBuf {@link ByteBuffer} representation of {@link BinaryRow}.
+     * @param row {@link BinaryRow}.
      * @param txId Transaction ID.
      * @return Constructed {@link UpdateCommand} object.
      */
-    private UpdateCommand updateCommand(TablePartitionId tablePartId, UUID rowUuid, ByteBuffer rowBuf, UUID txId) {
+    private UpdateCommand updateCommand(TablePartitionId tablePartId, UUID rowUuid, TableRow row, UUID txId) {
```
korlov42, line 1951: why did you change method's signature?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1931,19 +1944,19 @@ private HybridTimestampMessage hybridTimestamp(HybridTimestamp tmstmp) {
      *
      * @param tablePartId {@link TablePartitionId} object to construct {@link UpdateCommand} object with.
      * @param rowUuid Row UUID.
-     * @param rowBuf {@link ByteBuffer} representation of {@link BinaryRow}.
+     * @param row {@link BinaryRow}.
```
korlov42, line 1947: something is missed here

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
```diff
@@ -102,23 +104,20 @@ static void afterAll() {
...
     }
 
-    protected static BinaryRow binaryKey(TestKey key) {
-        try {
-            return kvMarshaller.marshal(key);
-        } catch (MarshallerException e) {
-            throw new IgniteException(e);
-        }
+    protected static TableRow tableRow(TestKey key, TestValue value) {
+        return TableRowConverter.fromBinaryRow(binaryRow(key, value), kvBinaryConverter);
```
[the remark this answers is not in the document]
valepakh: I would do it after the marshalers are converted to the TableRow interface.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
valepakh: Mainly to reduce clutter at the call site, but probably it would be better if the update and updateAll singatures would be similar.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/TableRowAndRowId.java
```diff
@@ -18,31 +18,32 @@
 package org.apache.ignite.internal.storage;
 
 import org.apache.ignite.internal.schema.BinaryRow;
+import org.apache.ignite.internal.schema.TableRow;
 import org.jetbrains.annotations.Nullable;
 
 /**
  * Wrapper that holds both {@link BinaryRow} and {@link RowId}. {@link BinaryRow} is null for tombstones.
```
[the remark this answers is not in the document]
korlov42: there are still mentions

modules/schema/src/main/java/org/apache/ignite/internal/schema/TableRow.java
```diff
@@ -0,0 +1,105 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.schema;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+/** Heap byte buffer-based row. */
+public class TableRow {
```
tkalkirill, line 24: Why not use **BinaryTuple**? This change looks strange.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
```diff
@@ -151,7 +151,7 @@ public interface MvPartitionStorage extends ManuallyCloseable {
      * @throws TxIdMismatchException If there's another pending update associated with different transaction id.
      * @throws StorageException If failed to write data to the storage.
      */
-    @Nullable BinaryRow addWrite(RowId rowId, @Nullable BinaryRow row, UUID txId, UUID commitTableId, int commitPartitionId)
+    @Nullable TableRow addWrite(RowId rowId, @Nullable TableRow row, UUID txId, UUID commitTableId, int commitPartitionId)
```
tkalkirill, line 154: Why not use **BinaryTuple**? This change looks strange.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/TableRowAndRowId.java
```diff
@@ -17,32 +17,32 @@
...
 /**
- * Wrapper that holds both {@link BinaryRow} and {@link RowId}. {@link BinaryRow} is null for tombstones.
+ * Wrapper that holds both {@link TableRow} and {@link RowId}. {@link TableRow} is null for tombstones.
  */
-public class BinaryRowAndRowId {
-    /** Binary row. */
-    private final @Nullable BinaryRow binaryRow;
+public class TableRowAndRowId {
+    /** Table row. */
+    private final @Nullable TableRow tableRow;
```
tkalkirill, line 28: Why not use **BinaryTuple**? This change looks strange.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
korlov42: table can change over time, so we must store the version of the schema that was used to build the tuple so we can read its contents later

modules/schema/src/main/java/org/apache/ignite/internal/schema/TableRow.java
valepakh: BinaryTuple is an existing class which is an utility that uses a schema to read values. I don't like the name either.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
ademakov: Within a transaction at any given moment of time only one table schema is in effect. It does not change arbitrarily. It is redundant to specify schema for every inserted row. The storage engine does not have to store the schema with every row. It is sufficient to remember the transaction timestamps since which a given table schema version is in effect.

korlov42: > The storage engine does not have to store the schema version together with every row too. It is sufficient to remember the transaction timestamp since which a given table schema version is in effect.

That is not how it's working right now.

ademakov: Columnar storage does not store schema version with rows right now.

The whole effort was triggered because BinaryRow had too many assumptions that were plain wrong for columnar. Now this PR adds TableRow that basically has the same problem.

Schema version should be handled in the same fashion as rowid is handled now. That is it should be passed as an additional parameter or field or whatever. It should never be embedded into the binary data. This way we will avoid troubles for storage engines that do not need it.

modules/schema/src/main/java/org/apache/ignite/internal/schema/TableRow.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * Factory method which creates an instance using a schema version and an existing {@link BinaryTuple}.
+     *
+     * @param schemaVersion Schema version.
+     * @param binaryTuple Binary tuple.
+     * @return Created TableRow instance.
+     */
+    public static TableRow createFromTuple(int schemaVersion, BinaryTuple binaryTuple) {
```
ibessonov, line 67: What is the reason of encoding schema version directly into the binary payload? This leads to too much memory manipulation. Storages can do that more efficiently.
I suggest storing version it as an integer field, and only assemble a byte[] when asked.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
korlov42: Ignite does not have columnar storage

modules/schema/src/main/java/org/apache/ignite/internal/schema/TableRow.java
korlov42: > What is the reason of encoding schema version directly into the binary payload? 

it is a direct replacement for ByteBufferRow to facilitate migration

>  I suggest storing version as an integer field

if the storage doesn't mind to work not with abstract bytes, but with a specific row object, let's do this way

valepakh: I think we should make these migrations as simple as possible and avoid making changes not relevant to the binary format.

ibessonov: Ok, but we should fix this in the future. There must be a TODO. Current approach is not efficient

PR #1482, opened 2022-12-27
IGNITE-18435 .NET: Add support for Enum mapping

ptupitsyn:
* Map integer columns (int8, int16, int32, int64) to .NET enums.
* Includes support in `RecordView`, `KeyValueView`, and LINQ.

### DIFF
--- a/modules/platforms/dotnet/Apache.Ignite.Tests/Table/RecordViewPocoTests.cs
+++ b/modules/platforms/dotnet/Apache.Ignite.Tests/Table/RecordViewPocoTests.cs
@@ -25,6 +25,7 @@ namespace Apache.Ignite.Tests.Table
     using System.Threading.Tasks;
     using NodaTime;
     using NUnit.Framework;
+    using NUnit.Framework.Internal;
 
     /// <summary>
     /// Tests for POCO view.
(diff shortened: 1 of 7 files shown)


modules/platforms/dotnet/Apache.Ignite/Internal/Table/Serialization/BinaryTupleMethods.cs
```diff
@@ -176,34 +176,36 @@ internal static class BinaryTupleMethods
...
-        /// <param name="valueType">Type of the value to read.</param>
+        /// <param name="type">Type of the value to read.</param>
         /// <returns>Read method for the specified value type.</returns>
-        public static MethodInfo? GetReadMethodOrNull(Type valueType) => ReadMethods.GetValueOrDefault(valueType);
+        public static MethodInfo? GetReadMethodOrNull(Type type) => ReadMethods.GetValueOrDefault(Unwrap(type));
 
         private static IgniteClientException GetUnsupportedTypeException(Type valueType) =>
             new(ErrorGroups.Client.Configuration, "Unsupported type: " + valueType);
+
+        private static Type Unwrap(Type type) => type.IsEnum ? Enum.GetUnderlyingType(type) : type;
```
ashapkin, line 209: Make it internal and use in [ObjectSerializerHandler.cs]?

modules/platforms/dotnet/Apache.Ignite/Internal/Table/Serialization/ObjectSerializerHandler.cs
```diff
@@ -500,11 +500,13 @@ private static void EmitFieldRead(FieldInfo? fieldInfo, ILGenerator il, Column c
...
         {
             var columnType = column.Type.ToType();
+
             var fieldType = Nullable.GetUnderlyingType(fieldInfo.FieldType) ?? fieldInfo.FieldType;
+            fieldType = fieldType.IsEnum ? Enum.GetUnderlyingType(fieldType) : fieldType;
 
             if (fieldType != columnType)
             {
-                var message = $"Can't map field '{fieldInfo.DeclaringType?.Name}.{fieldInfo.Name}' of type '{fieldType}' " +
+                var message = $"Can't map field '{fieldInfo.DeclaringType?.Name}.{fieldInfo.Name}' of type '{fieldInfo.FieldType}' " +
```
ashapkin, line 509: Just an idea - provide a list of supported types or a link to a future documentation link?

modules/platforms/dotnet/Apache.Ignite.Tests/Table/RecordViewPocoTests.cs
```diff
@@ -721,6 +721,34 @@ public async Task TestAllColumnsPocoNullable()
...
+        [Test]
+        public async Task TestEnumColumns()
+        {
+            // Normal values.
+            await Test(new PocoEnums.PocoIntEnum(1, PocoEnums.IntEnum.Foo));
+            await Test(new PocoEnums.PocoByteEnum(1, PocoEnums.ByteEnum.Foo));
+            await Test(new PocoEnums.PocoShortEnum(1, PocoEnums.ShortEnum.Foo));
+            await Test(new PocoEnums.PocoLongEnum(1, PocoEnums.LongEnum.Foo));
+
+            // Values that are not represented in the enum (it is just a number underneath).
```
ashapkin, line 733: We can also test the default value, like
await Test(new PocoEnums.PocoIntEnum(1, default(PocoEnums.IntEnum)));

ashapkin (approved):

modules/platforms/dotnet/Apache.Ignite/Internal/Table/Serialization/BinaryTupleMethods.cs
ptupitsyn: Done.

modules/platforms/dotnet/Apache.Ignite/Internal/Table/Serialization/ObjectSerializerHandler.cs
ptupitsyn: In this case the issue is not about unsupported type, but about type mismatch.

Unlike Ignite 2.x, you can't put an arbitrary object into the table (cache). First, you create a table with SQL. Then you can map a class to that table. And if you make a mistake with field types, you'll get a message like "Can't map int field 'foo' to string column 'foo'". So you go ahead and change the field type. No need to check the docs.

modules/platforms/dotnet/Apache.Ignite.Tests/Table/RecordViewPocoTests.cs
ptupitsyn: Good point, added.

PR #1485, opened 2022-12-28
IGNITE-18470 Exception handling in DistributionZoneManager was reworked.

sergeyuttsel:
https://issues.apache.org/jira/browse/IGNITE-18470


modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -483,47 +560,206 @@ private void initMetaStorageKeysOnStart() {
...
+            } else {
+                LOG.debug("Failed to delete zones' dataNodes key [zoneId = {}]", zoneId);
+            }
+        });
+    }
+
+    /**
+     * Unwraps distribution zone exceptions from {@link ConfigurationChangeException} if it is possible.
+     */
+    private static Throwable unwrapDistributionZoneException(Throwable e) {
```
sk0x50, line 753: This method looks weird to me, especially `e.getCause().getCause()` part. I would propose something like the following:

```
    private static Throwable unwrapDistributionZoneException(Throwable e, Class<? extends Throwable>... expectedClz) {
        Throwable ret = unwrapDistributionZoneExceptionRecursively(e, expectedClz);

        return ret != null ? ret : e;
    }

    private static Throwable unwrapDistributionZoneExceptionRecursively(Throwable e, Class<? extends Throwable>... expectedClz) {
        if ((e instanceof CompletionException || e instanceof ConfigurationChangeException) && e.getCause() != null) {
            return unwrapDistributionZoneExceptionRecursively(e.getCause(), expectedClz);
        }

        for (Class<?> expected : expectedClz) {
            if (expected.isAssignableFrom(e.getClass()))
                return e;
        }

        return null;
    }
```

By the way, it allows you clearly to define expected types of exceptions:
```
    public CompletableFuture<Void> createZone(DistributionZoneConfigurationParameters distributionZoneCfg) {
        ....
            })).whenComplete((res, e) -> {
                if (e != null) {
                    fut.completeExceptionally(unwrapDistributionZoneException(e, DistributionZoneAlreadyExistsException.class, ConfigurationValidationException.class));
                } else {
                    fut.complete(null);
                }
            });
```

What do you think?

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerTest.java
```diff
@@ -156,30 +157,36 @@ public void testCreateZoneIfExists() throws Exception {
...
-                    new DistributionZoneConfigurationParameters.Builder(ZONE_NAME).dataNodesAutoAdjust(100).build()
-            ).get(5, TimeUnit.SECONDS);
+            fut.get(5, TimeUnit.SECONDS);
         } catch (Exception e0) {
             e = e0;
         }
 
         assertTrue(e != null);
-        assertTrue(e.getCause().getCause() instanceof DistributionZoneAlreadyExistsException, e.toString());
+        assertTrue(e.getCause() instanceof DistributionZoneAlreadyExistsException, e.toString());
```
sk0x50, line 173: I like the approach that always requires a message, it looks a lot friendlier.
```
    assertTrue("Expected exception was not thrown.", e != null);
    assertTrue("Unexpected type of exception (requires DistributionZoneAlreadyExistsException)", e.getCause() instanceof DistributionZoneAlreadyExistsException);
```

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerTest.java
```diff
@@ -156,30 +157,36 @@ public void testCreateZoneIfExists() throws Exception {
...
     }
 
     @Test
-    public void testDropZoneIfNotExists() {
+    public void testDropZoneIfNotExists() throws NodeStoppingException {
```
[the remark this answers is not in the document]
sergeyuttsel: Fixed

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerTest.java
sergeyuttsel: Makes sense. Fixed.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
sergeyuttsel: Yes, I think it's better.

PR #1486, opened 2022-12-29
IGNITE-18451 Node stop may hang if an index is not created yet

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18451

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
@@ -371,6 +371,14 @@ public class TableImpl implements Table {
         allOf(toWait.toArray(CompletableFuture[]::new)).join();
     }
 
+    /**
+     * Prepares this table for being closed. Currently, this means cancellation of PK future so that RAFT state machine
+     * does not need to wait for PK to be initialized, which allows to avoid a deadlock on stop.
+     */
+    public void prepareToClose() {
+        pkId.cancel(false);
+    }
+
     @FunctionalInterface
     private interface IndexLockerFactory {
         /** Creates the index decorator for given partition. */
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
@@ -372,8 +372,7 @@ public class TableImpl implements Table {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
```diff
@@ -371,6 +371,14 @@ private void awaitIndexes() {
         allOf(toWait.toArray(CompletableFuture[]::new)).join();
     }
 
+    /**
+     * Prepares this table for being closed. Currently, this means cancellation of PK future so that RAFT state machine
+     * does not need to wait for PK to be initialized, which allows to avoid a deadlock on stop.
+     */
+    public void prepareToClose() {
```
sashapolo, line 378: I would suggest to call it `beforeClose`

AMashenkov: Why do we mention RAFT here?
Why do we mention a deadlock? It just releasing resources in safe way.

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
```diff
@@ -371,6 +371,14 @@ private void awaitIndexes() {
         allOf(toWait.toArray(CompletableFuture[]::new)).join();
     }
 
+    /**
+     * Prepares this table for being closed. Currently, this means cancellation of PK future so that RAFT state machine
+     * does not need to wait for PK to be initialized, which allows to avoid a deadlock on stop.
+     */
+    public void beforeClose() {
+        pkId.cancel(false);
```
AMashenkov, line 379: Future cancellation finishes future with meaningful exception. Let's use some IgniteException exception here.

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
rpuch: Removed the specifics

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
rpuch: Fixed

sashapolo (approved):

AMashenkov (approved):

SammyVimes (approved):
LGTM

PR #1489, opened 2022-12-30
[IGNITE-17972] SQL. Incorrect error message for explicit NULL values for NON NULL column

lowka:


korlov42 (changes requested):
Calcite allows to override messages via `CalciteResource.properties`. I would prefer to stick with that approach

lowka:
Done

korlov42 (approved):

PR #1493, opened 2023-01-04
IGNITE-18456 Explain treading model in corresponding README.md file f…

vldpyatkov:
…or MetaStorageManager
https://issues.apache.org/jira/browse/IGNITE-18456

### DIFF
--- a/modules/metastorage/README.md
+++ b/modules/metastorage/README.md
@@ -1,46 +1,45 @@
 # Metastorage
 
-The module for store and access to metadata. There is module is connected with this one:
+The module for storing and accessing metadata. This module is linked to another one:
 
-- metastorage-api - the module contains classes to access to metastorage service from another components.
+- metastorage-api - The module contains classes that other components can use to access the metastorage service.
 
-The storage is distributed to prevent losing data and based of RAFT. That distributed storage persists on top of every cluster node, however
-most of such participants are learners and only few of them are voting ones. In typical case, the group consists of only several voted nodes
-of the cluster (amount of nodes have to be odd 3 or 5). The rest of nodes listens metadata update but does not participant in voting (RAFT 
-learners).
+To avoid data loss, the storage is distributed using the RAFT consensus algorithm. Every cluster node has access to such distributed
+storage, but the majority of members are learners, and only a small number are voters in terms of RAFT algorithm. Typically, only a small
+number of the cluster's voting nodes make up the raft-group (number of nodes must be odd, either 3 or 5.). The remaining nodes listen to
+metadata updates but do not vote (they called learners in terms of RAFT).
 
 ## Threading model
 
-There is only one dedicate thread to notify watches. This is a thread with prefix metastorage-watch-executor.
-The thread is created from executor in KeyValueStorage:
+There is only one thread specifically for notifying watchers. The thread is created in KeyValueStorage from the executor:
 
 ```java
 private ExecutorService watchExecutor = Executors.newSingleThreadExecutor(new NamedThreadFactory(NamedThreadFactory
         .threadPrefix(nodeName, "metastorage-watch-executor"),LOG));
 ```
 
-Also, the storage internally contains a two threads executor to create a snapshot:
+Additionally, the storage internally has two thread executors for creating a snapshot:
 
 ```java
 private ExecutorService snapshotExecutor = Executors.newFixedThreadPool(2,new NamedThreadFactory(NamedThreadFactory
         .threadPrefix(nodeName, "metastorage-snapshot-executor"),LOG));
 ```
 
-Both of the executors have a node name in their prefixes to distinguish to which node the particular thread belongs.
+To distinguish to which node the specific thread belongs, each executor has a node name in its prefix.
 
 ### Interface methods
 
-Various operations (*get()*, *getAll()*, *invoke()*) in Metastorage manager return futures. Those futures are completed when the matched
-RAFT command completes to the Metastorage group. This result appears in RAFT client executor (prefix <NODE_NAME>%Raft-Group-Client), but the
-entire replication procedure is happened tin the RAFT threads (description of that treading model available in 
-[Raft module](../raft/README.md)).
+Futures are returned by a number of Metastorage Manager operations, like get(), getAll(), invoke(), etc. Those futures are completed when
+the corresponding RAFT command is completed in the Metastorage group. Although the entire replication process took place in the RAFT
+threads, this result appears in the RAFT client executor with prefix <NODE_NAME>%Raft-Group-Client. See RAFT module for more information
+about its threading model.
 
-Although another methods are returned futures, but in most cases they are executed synchronously. The futures are conditioned by
-asynchronous Metastorage initialization, that happens because required a time to start a RAFT group. In other words, we have a chance of
-complete those features in thread of RAFT client executor (prefix <NODE_NAME>%Raft-Group-Client).
+Although some methods return futures, they are often run synchronously. Futures are dependent on asynchronous Metastorage initialization,
+since starting an RAFT group requires time. In other words, we have a chance to complete those features in the RAFT client executor thread (
+prefix NODE NAME>%Raft-Group-Client).
 
 ### Using common pool
 
-The component is used common ForkJoinPool on start (in fact, it is not necessary, because all components started in asynchronously in the
-same ForkJoinPool). The using of the common pool is dangerous, because the pool can be busy by another threads that hosted on the same JVM 
+The component uses common ForkJoinPool on start (in fact, it is not necessary, because all components starts asynchronously in the same
+ForkJoinPool). The using of the common pool is dangerous, because the pool can be busy by another threads that hosted on the same JVM 
 (TODO: IGNITE-18505 Avoid using common pool on start components).
--- a/modules/metastorage/README.md
+++ b/modules/metastorage/README.md
@@ -1,23 +1,40 @@
-<h2>Metastorage</h2>
-The module for store and access to metadata.
-The storage is distributed to prevent losing data and based of RAFT. In typical case, the group consists of only several nodes of the cluster (amount of nodes have to be odd 3 or 5). The rest of nodes listens metadata update but does not participant in voting (RAFT listeners).
+# Metastorage
 
-<h2>Threading model</h2>
+The module for store and access to metadata.
+
+The storage is distributed to prevent losing data and based of RAFT. In typical case, the group consists of only several nodes of the
+cluster (amount of nodes have to be odd 3 or 5). The rest of nodes listens metadata update but does not participant in voting (RAFT
+listeners).
+
+## Threading model
 
 There is only one dedicate thread to notify watches. This is a thread with prefix metastorage-watch-executor.
 The thread is created from executor in KeyValueStorage:
+
 ```java
-private final ExecutorService watchExecutor = Executors.newSingleThreadExecutor(new NamedThreadFactory("metastorage-watch-executor", LOG));
+private final ExecutorService watchExecutor = Executors.newSingleThreadExecutor(new NamedThreadFactory("metastorage-watch-executor",LOG));
 ```
+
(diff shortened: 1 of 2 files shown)


modules/metastorage/README.md
```diff
@@ -0,0 +1,40 @@
...
+```java
+private final ExecutorService snapshotExecutor = Executors.newFixedThreadPool(2,new NamedThreadFactory("metastorage-snapshot-executor",LOG));
+```
+
+Both of the executors have no a node name in their names. The names are required to identify threads belonged to different Ignite nodes on
+the shared java machine (TODO: IGNITE-18504 Add nodes name in prefix to Metastorage executors).
+
+### Interface methods
+
+Various operations (*get()*, *getAll()*, *invoke()*) in Metastorage manager return futures. Those futures are completed when the matched
```
sanpwc, line 29: As we've discussed let's add few more details here.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageManagerImpl.java
```diff
@@ -200,7 +200,7 @@ public void onDisappeared(ClusterNode member) {
     public void start() {
         this.metaStorageSvcFut = cmgMgr.metaStorageNodes()
                 // use default executor to avoid blocking CMG manager threads
-                .thenComposeAsync(metaStorageNodes -> {
+                .thenCompose(metaStorageNodes -> {
```
sanpwc, line 203: Is that about 
> The component is used common ForkJoinPool on start

?
There's a comment
> // use default executor to avoid blocking CMG manager threads

So, do you really think that using thenCompose instead of thenComposeAsync won't block CMG manager threads?

modules/metastorage/README.md
```diff
@@ -0,0 +1,40 @@
...
+Also, the storage internally contains a two threads executor to create a snapshot:
+
+```java
+private final ExecutorService snapshotExecutor = Executors.newFixedThreadPool(2,new NamedThreadFactory("metastorage-snapshot-executor",LOG));
+```
+
+Both of the executors have no a node name in their names. The names are required to identify threads belonged to different Ignite nodes on
+the shared java machine (TODO: IGNITE-18504 Add nodes name in prefix to Metastorage executors).
+
+### Interface methods
```
sanpwc, line 27: Initially, there were asyncs on cursors, we do expect this to be reworked with subscribers, correct? If true this subscribers will probably add some (one?) new threads. Could you please create a follow-up ticket in order to update ms thread-model after Alexander will finish his ms-cursors refactoring?

sanpwc (changes requested):

modules/metastorage/README.md
```diff
@@ -0,0 +1,40 @@
+# Metastorage
+
+The module for store and access to metadata.
+
+The storage is distributed to prevent losing data and based of RAFT. In typical case, the group consists of only several nodes of the
+cluster (amount of nodes have to be odd 3 or 5). The rest of nodes listens metadata update but does not participant in voting (RAFT
+listeners).
```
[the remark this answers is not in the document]
vldpyatkov: Done.

modules/metastorage/README.md
```diff
@@ -0,0 +1,40 @@
+# Metastorage
+
+The module for store and access to metadata.
+
+The storage is distributed to prevent losing data and based of RAFT. In typical case, the group consists of only several nodes of the
```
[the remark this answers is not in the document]
vldpyatkov: Failure tolerance has the same stance in case of one node leaves the cluster and data should be still available. In my mind, the prevention of losing data is a general characteristic for all requirements that an any distributed system has.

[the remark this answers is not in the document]
vldpyatkov: I reformulated this part.

modules/metastorage/README.md
vldpyatkov: Added link to raft module.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageManagerImpl.java
vldpyatkov: I am sure, because the code does not execute in CMG thread after all components began start in the common thread pool.

vldpyatkov: The comment about CMG thread are removed from here.

modules/metastorage/README.md
vldpyatkov: We even have no an JIRA ticket to redesign cursor API to reactive model for Metastorage. But I notified @sashapolo about the modification can be required after the API will have changed.

modules/metastorage/README.md
```diff
@@ -0,0 +1,46 @@
+# Metastorage
+
+The module for store and access to metadata. There is module is connected with this one:
```
alievmirza, line 3: The module for storing and accessing metadata. This module is linked to another one:

modules/metastorage/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+- metastorage-api - the module contains classes to access to metastorage service from another components.
```
alievmirza, line 5: The module contains classes that other components can use to access the metastorage service.

modules/metastorage/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+The storage is distributed to prevent losing data and based of RAFT. That distributed storage persists on top of every cluster node, however
+most of such participants are learners and only few of them are voting ones. In typical case, the group consists of only several voted nodes
+of the cluster (amount of nodes have to be odd 3 or 5). The rest of nodes listens metadata update but does not participant in voting (RAFT 
+learners).
+
```
alievmirza, line 11: To avoid data loss, the storage is distributed using the RAFT consensus algorithm. Every cluster node has access to such distributed storage, but the majority of members are learners, and only a small number are voters in terms of RAFT algorithm. Typically, only a small number of the cluster's voting nodes make up the raft-group (number of nodes must be odd, either 3 or 5.). The remaining nodes listen to metadata updates but do not vote (they called learners in terms of RAFT).

modules/metastorage/README.md
```diff
@@ -0,0 +1,46 @@
...
+## Threading model
+
+There is only one dedicate thread to notify watches. This is a thread with prefix metastorage-watch-executor.
+The thread is created from executor in KeyValueStorage:
+
```
alievmirza, line 16: There is only one thread specifically for notifying watchers.
The thread is created in KeyValueStorage from the executor:

modules/metastorage/README.md
```diff
@@ -0,0 +1,46 @@
...
+```java
+private ExecutorService watchExecutor = Executors.newSingleThreadExecutor(new NamedThreadFactory(NamedThreadFactory
+        .threadPrefix(nodeName, "metastorage-watch-executor"),LOG));
+```
+
+Also, the storage internally contains a two threads executor to create a snapshot:
+
```
alievmirza, line 23: Additionally, the storage internally has two thread executors for creating a snapshot:

modules/metastorage/README.md
```diff
@@ -0,0 +1,46 @@
...
+```java
+private ExecutorService snapshotExecutor = Executors.newFixedThreadPool(2,new NamedThreadFactory(NamedThreadFactory
+        .threadPrefix(nodeName, "metastorage-snapshot-executor"),LOG));
+```
+
+Both of the executors have a node name in their prefixes to distinguish to which node the particular thread belongs.
+
```
alievmirza, line 30: To distinguish to which node the specific thread belongs, each executor has a node name in its prefix.

modules/metastorage/README.md
```diff
@@ -0,0 +1,46 @@
...
+### Interface methods
+
+Various operations (*get()*, *getAll()*, *invoke()*) in Metastorage manager return futures. Those futures are completed when the matched
+RAFT command completes to the Metastorage group. This result appears in RAFT client executor (prefix <NODE_NAME>%Raft-Group-Client), but the
+entire replication procedure is happened tin the RAFT threads (description of that treading model available in 
+[Raft module](../raft/README.md)).
```
alievmirza, line 36: Futures are returned by a number of Metastorage Manager operations, like *get()*, *getAll()*, *invoke()*, etc. Those futures are completed when the corresponding RAFT command is completed in the Metastorage group. Although the entire replication process took place in the RAFT threads, this result appears in the RAFT client executor with prefix  <NODE_NAME>%Raft-Group-Client. See RAFT module for more information about its threading model.

modules/metastorage/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Although another methods are returned futures, but in most cases they are executed synchronously. The futures are conditioned by
+asynchronous Metastorage initialization, that happens because required a time to start a RAFT group. In other words, we have a chance of
+complete those features in thread of RAFT client executor (prefix <NODE_NAME>%Raft-Group-Client).
+
```
alievmirza, line 41: Although some methods return futures, they are often run synchronously. Futures are dependent on asynchronous Metastorage initialization, since starting an RAFT group requires time. In other words, we have a chance to complete those features in the RAFT client executor thread (prefix NODE NAME>%Raft-Group-Client).

modules/metastorage/README.md
```diff
@@ -0,0 +1,46 @@
...
+### Using common pool
+
+The component is used common ForkJoinPool on start (in fact, it is not necessary, because all components started in asynchronously in the
+same ForkJoinPool). The using of the common pool is dangerous, because the pool can be busy by another threads that hosted on the same JVM 
+(TODO: IGNITE-18505 Avoid using common pool on start components).
```
alievmirza, line 46: The component uses common ForkJoinPool on start (in fact, it is not necessary, because all components starts asynchronously in the
same ForkJoinPool). The using of the common pool is dangerous, because the pool can be busy by another threads that hosted on the same JVM 
(TODO: IGNITE-18505 Avoid using common pool on start components).

modules/metastorage/README.md
```diff
@@ -0,0 +1,45 @@
...
+To distinguish to which node the specific thread belongs, each executor has a node name in its prefix.
+
+### Interface methods
+
+Futures are returned by a number of Metastorage Manager operations, like get(), getAll(), invoke(), etc. Those futures are completed when
+the corresponding RAFT command is completed in the Metastorage group. Although the entire replication process took place in the RAFT
+threads, this result appears in the RAFT client executor with prefix <NODE_NAME>%Raft-Group-Client. See RAFT module for more information
+about its threading model.
+
+Although some methods return futures, they are often run synchronously. Futures are dependent on asynchronous Metastorage initialization,
```
denis-chudov, line 37: shouldn'we specify these methods? "some methods" sounds too uncertain

modules/metastorage/README.md
```diff
@@ -0,0 +1,45 @@
...
+since starting an RAFT group requires time. In other words, we have a chance to complete those features in the RAFT client executor thread (
+prefix NODE NAME>%Raft-Group-Client).
+
+### Using common pool
+
+The component uses common ForkJoinPool on start (in fact, it is not necessary, because all components starts asynchronously in the same
+ForkJoinPool). The using of the common pool is dangerous, because the pool can be busy by another threads that hosted on the same JVM
```
denis-chudov, line 44:
```suggestion
ForkJoinPool). The using of the common pool is dangerous, because the pool can be occupied by another threads that hosted on the same JVM 
```

modules/metastorage/README.md
vldpyatkov: Honestly, I do not shut what exactly methods can appear in Metastorage manager, except method that are using to access data (but they are, for example, method of stopping watch). Rather, I mean this about all method that does not connected with data access, but return a future.

vldpyatkov: I don't know exactly because this moment the Metastorae API is changing.

PR #1499, opened 2023-01-05
IGNITE-18508 Fix metastorage gradle publishing

Berkof:
Add publishing specifications into metastorage modules gradle configuration.


Pochatkin:
LGTM

PR #1501, opened 2023-01-06
IGNITE-17951 Sql. Enlist partitions in transaction before executing a query.

xtern:
What is done in the PR:

1. Added a new InternalTable#scan method that requires a transaction id and a RAFT leader term to scan a local partition.
  (it seems that this procedure can be improved by introducing a "completely local" scan).
2. The "RAFT leader term" can be different for each partition group, so during the SQL mapping we collect not only the RAFT "leader", but also the "term" for each partition.
  (see `InternalTable#leaderAssignmentsWithTerm()`)
3. Partitions are enlisted in the transaction before plan fragments are sent, right after the fragment mapping step.
  (see `ExecutionServiceImpl.DistributedQueryManager#enlistPartitions`)
4. All tests that used the old explicit transaction method have been reworked to use the new method.


modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
```diff
@@ -768,4 +789,26 @@ private static Row createKeyRow(int id) {
...
+     */
+    private InternalTransaction startTxWithEnlistedPartition(int partId) {
+        Ignite ignite = CLUSTER_NODES.get(0);
+
+        InternalTransaction tx = (InternalTransaction) ignite.transactions().begin();
+
+        InternalTable table = ((TableImpl) ignite.tables().table(TABLE_NAME)).internalTable();
+        TablePartitionId tblPartId = new TablePartitionId(table.tableId(), partId);
+        RaftGroupService raftSvc = table.partitionRaftGroupService(partId);
+        long term = raftSvc.refreshAndGetLeaderWithTerm().join().term();
```
ygerzhedovich, line 807: can we amend join() to IgniteTestUtils#await ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -401,19 +402,22 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
...
+                    UUID txId = outerTx == null || distributedSqlDisabled ? null : outerTx.id();
 
                     BaseQueryContext ctx = BaseQueryContext.builder()
                             .frameworkConfig(
                                     Frameworks.newConfigBuilder(FRAMEWORK_CONFIG)
                                             .defaultSchema(schema)
-                                            .traitDefs(rwOp || (outerTx != null && !outerTx.isReadOnly()) ? Commons.LOCAL_TRAITS_SET :
-                                                    Commons.DISTRIBUTED_TRAITS_SET)
+                                            .traitDefs(distributedSqlDisabled ? Commons.LOCAL_TRAITS_SET : Commons.DISTRIBUTED_TRAITS_SET)
+                                            .traitDefs(Commons.DISTRIBUTED_TRAITS_SET)
```
ygerzhedovich, line 413: why do you have the row?  It looks as repeat )

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -401,19 +402,22 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
...
                                             .build()
                             )
                             .logger(LOG)
                             .cancel(queryCancel)
                             .parameters(params)
                             .transaction(outerTx)
+                            .transactionId(txId)
```
ygerzhedovich, line 420: it looks as you pass three parameters instead of one transaction, which contains all of them

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
xtern: No, since `IgniteTestUtils` should be used only in tests.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
xtern: Thanks, fixed.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
xtern: oops... sorry

Fixed in both tests, thanks.

modules/index/src/main/java/org/apache/ignite/internal/index/SortedIndex.java
```diff
@@ -92,7 +99,9 @@ default Publisher<BinaryRow> scan(
      * @return A cursor from resulting rows.
      * @see SortedIndex#INCLUDE_LEFT
      * @see SortedIndex#INCLUDE_RIGHT
+     * @deprecated Use {@link #scan(int, UUID, ClusterNode, long, BinaryTuplePrefix, BinaryTuplePrefix, int, BitSet)} instead.
      */
+    @Deprecated
```
ygerzhedovich, line 104: please add mention IGNITE-17952  here, to be sure that we remove it

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
```diff
@@ -492,7 +505,8 @@ public void testTwiceScanInTransaction() throws Exception {
 
         assertEquals(ROW_IDS.size() + 1, scannedRows.size());
 
-        var publisher1 = internalTable.scan(0, tx, sortedIndexId, null, null, 0, null);
+        var publisher1 =
```
ygerzhedovich, line 508: 'var' is not allowed here.

modules/index/src/main/java/org/apache/ignite/internal/index/SortedIndex.java
xtern: Added jira issue number after `@deprecated` in javadoc.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
xtern: Fixed, thanks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -428,7 +431,8 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
                                 InternalTransaction implicitTx = implicitTxRequired ? txManager.begin() : null;
 
                                 BaseQueryContext enrichedContext =
-                                        implicitTxRequired ? ctx.toBuilder().transaction(implicitTx).build() : ctx;
+                                        implicitTxRequired ? ctx.toBuilder().transaction(implicitTx)
+                                                .transactionId(distributedSqlDisabled ? null : implicitTx.id()).build() : ctx;
```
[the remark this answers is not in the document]
xtern: As I understand, you suggest to set `transactionId` mandatory.
But such change will require to change condition order in scan nodes.
```
if (readOnlyTx) {
    pub = physTable.scan(...
} else if (context().transactionId() != null) {
    pub = physTable.scan(...
} else {
    // TODO IGNITE-17952 This block should be removed.
    // Workaround to make RW scan work from tx coordinator.
    pub = physTable.scan(partId, context().transaction());
}
```
to something like this
```
if (readOnlyTx) {
    pub = physTable.scan(...
} else if (context.transaction() != null) {
    // TODO IGNITE-17952 This block should be removed.
    // Workaround to make RW scan work from tx coordinator.
    pub = physTable.scan(partId, context().transaction());
} else {
    pub = physTable.scan(...
} 
```
otherwise, we will try to do an RW scan on the local host because DISTRIBUTED_TRAIT is currently disabled and we send all plan fragments to the local host.

Is it ok?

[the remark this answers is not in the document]
xtern: After some thought, I decided that this makes the code look cleaner and easier to make the changes needed to enable distributive SQL (in IGNITE-17952). So I decided to follow your suggestion and made `txId` mandatory, thanks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/IndexScanNode.java
```diff
@@ -85,6 +89,7 @@ public IndexScanNode(
             IgniteIndex schemaIndex,
             InternalIgniteTable schemaTable,
             int[] parts,
+            long @Nullable[] terms,
```
zstan, line 92: ups, is it good idea to mix raft internals and execution node ? probably we can some facade if it`s necessarily ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -165,6 +178,7 @@ public static <RowT> ExecutionServiceImpl<RowT> create(
     public ExecutionServiceImpl(
             ClusterNode localNode,
             MessageService msgSrvc,
+            TopologyService topSrvc,
```
ygerzhedovich, line 181: in this case, we could don't pass localNode parameter

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/IndexScanNode.java
```diff
@@ -117,19 +124,21 @@ protected Publisher<RowT> scan() {
...
     private Publisher<RowT> indexPublisher(int[] parts, @Nullable RangeCondition<RowT> cond) {
-        Iterator<Publisher<? extends RowT>> it = new TransformingIterator<>(
-                Arrays.stream(parts).iterator(),
-                part -> partitionPublisher(part, cond)
-        );
+        boolean roTx = context().transactionTime() != null;
+        List<Publisher<? extends RowT>> publishers = new ArrayList<>(parts.length);
+
+        for (int i = 0; i < parts.length; i++) {
+            publishers.add(partitionPublisher(parts[i], roTx ? -1 : terms[i], cond));
```
ygerzhedovich, line 131: the change removed the lazy opening of the cursors

ygerzhedovich: Just to use `IntStream.range(0, parts.length).iterator()` as source for array indexes

xtern: Fixed, thanks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
xtern: Removed localNode parameter from signature

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -962,6 +977,60 @@ private Publisher<BinaryRow> scan(
...
+                            .transactionId(txId)
+                            .scanId(scanId)
+                            .indexToUse(indexId)
+                            .exactKey(exactKey)
+                            .lowerBound(lowerBound)
+                            .upperBound(upperBound)
+                            .flags(flags)
+                            .columnsToInclude(columnsToInclude)
+                            .batchSize(batchSize)
+                            .timestamp(clock.now())
```
ygerzhedovich, line 1024: Why we should pass the clock here? Seems it should work just with txId.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -1010,6 +1079,30 @@ public List<String> assignments() {
...
     }
 
+    /** {@inheritDoc} */
+    // TODO: IGNITE-17256 Use a placement driver for getting a primary replica.
+    @Override
+    public List<IgniteBiTuple<String, Long>> leaderAssignmentsWithTerm() {
+        List<Entry<RaftGroupService>> entries = new ArrayList<>(partitionMap.int2ObjectEntrySet());
+        List<CompletableFuture<LeaderWithTerm>> futs = new ArrayList<>();
+
+        entries.sort(Comparator.comparingInt(Entry::getIntKey));
```
ygerzhedovich, line 1089: I see a few places in the class where we are sorting the map. Let's rewrite it to use an already sorted map - Int2ObjectSortedMap

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/rel/TableScanNodeExecutionTest.java
```diff
@@ -161,29 +157,40 @@ public TestInternalTableImpl(
...
+                    public void request(long n) {
+                        int fillAmount = Math.min(dataAmount - processedPerPart[partId], (int) n);
+
+                        processedPerPart[partId] += fillAmount;
+
+                        for (int i = 0; i < fillAmount; ++i) {
+                            s.onNext(bbRow);
+                        }
+
+                        if (processedPerPart[partId] >= dataAmount) {
```
ygerzhedovich, line 183:
```suggestion
                        if (processedPerPart[partId] == dataAmount) {
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -1010,6 +1079,30 @@ public List<String> assignments() {
...
+        for (Entry<RaftGroupService> e : entries) {
+            futs.add(e.getValue().refreshAndGetLeaderWithTerm());
+        }
+
+        List<IgniteBiTuple<String, Long>> leadersWithTerm = new ArrayList<>(entries.size());
+
+        for (CompletableFuture<LeaderWithTerm> fut : futs) {
+            LeaderWithTerm leaderWithTerm = fut.join();
+
+            leadersWithTerm.add(new IgniteBiTuple<>(leaderWithTerm.leader().consistentId(), leaderWithTerm.term()));
```
ygerzhedovich, line 1100: you can use NodeWithTerm instantly instead of BiTuple

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
```diff
@@ -194,30 +186,73 @@ public ColocationGroup colocate(ColocationGroup other) throws ColocationMappingE
...
+            }
+
+            res.add(nodeWithTerm);
+        }
+
+        return res;
+    }
+
     /**
      * Constructor.
```
ygerzhedovich, line 237: Let's fix javadoc

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
```diff
@@ -194,30 +186,73 @@ public ColocationGroup colocate(ColocationGroup other) throws ColocationMappingE
...
      * TODO Documentation https://issues.apache.org/jira/browse/IGNITE-15859
      */
     public ColocationGroup finalaze() {
```
ygerzhedovich, line 240: method name seems have incorrect naming

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
```diff
@@ -194,30 +186,73 @@ public ColocationGroup colocate(ColocationGroup other) throws ColocationMappingE
...
+                    nodes.add(first.name());
                 }
                 assignments.add(first != null ? Collections.singletonList(first) : Collections.emptyList());
             }
 
             return new ColocationGroup(sourceIds, new ArrayList<>(nodes), assignments);
         }
 
-        return forNodes0(nodeNames);
+        return nodeNames == null ? this : forNodes0(nodeNames);
```
ygerzhedovich, line 255:
```suggestion
        return mapToNodes(nodeNames);
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
```diff
@@ -194,30 +186,73 @@ public ColocationGroup colocate(ColocationGroup other) throws ColocationMappingE
...
+                intersection.add(otherNodeWithTerm);
+            }
+        }
+
+        return intersection;
+    }
+
+    private List<NodeWithTerm> filter(List<NodeWithTerm> assignment, Set<String> filter) {
+        List<NodeWithTerm> res = new ArrayList<>();
+
```
ygerzhedovich, line 220: let's do check before creating `res`.
Also please pass size for `res`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
xtern: Thanks, you are right.
It's a copy-paste from current `enlistCursorInTx`.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
xtern: AFAIK, I can't do it without changing dependencies.

`LeaderWithTerm` is located in raft-api module
`NodeWithTerm` is located in sql-engine module (it depends on table module, but not raft-api)
`InternalTableImpl` is located in table module (it depends on raft-api, but not on sql-engine)

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
xtern: `finalize()` clashes with `Object#finalize()`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
xtern: Done.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
xtern: Done, thanks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
ygerzhedovich: I mean remove `TODO` and write javadoc - let's try improve our documentation step by step,

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
```diff
@@ -230,29 +266,40 @@ public ColocationGroup mapToNodes(List<String> nodeNames) {
...
     }
+
+    /**
+     * Returns a raft group leader term.
+     *
+     * @param partId Partition ID.
+     * @return Raft group leader term.
+     */
+    public long partitionLeaderTerm(int partId) {
+        return first(assignments.get(partId)).term();
```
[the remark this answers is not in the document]
xtern: Done.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/TransferredTxAttributesHolder.java
```diff
@@ -0,0 +1,118 @@
...
+import org.apache.ignite.network.ClusterNode;
+import org.apache.ignite.tx.TransactionException;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Holder of {@link InternalTransaction#id() id} and {@link InternalTransaction#readTimestamp() readTimestamp} transaction attributes
+ * that are passed to remote nodes during sql query execution.
+ */
+public class TransferredTxAttributesHolder implements InternalTransaction {
```
ygerzhedovich, line 36: let's mention that dependent on InternalTransaction temporary and will be removed in separate ticket

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -399,22 +400,22 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
                     return nodes.get(0);
                 })
                 .thenCompose(sqlNode -> {
-                    final boolean rwOp = dataModificationOp(sqlNode);
-                    final HybridTimestamp txTime = outerTx != null ? outerTx.readTimestamp() : rwOp ? null : clock.now();
+                    boolean rwOp = dataModificationOp(sqlNode);
+                    InternalTransaction outerTx0 =
```
ygerzhedovich, line 404: it's not 'outer' transaction, let's use just `tx` or smth like this

ygerzhedovich: It looks confusing - we have two internal transactions, here and the second one below. Can we have just one?

xtern: Done, thanks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/TransferredTxAttributesHolder.java
xtern: Done, thanks.

AMashenkov:
I don't think we can operate with RAFT terminology in SQL execution layer. Let's think how to avoid using the terms in API methods.
Leader and term should be a part of some transactional context, which can be passed through execution layer and indices right down to the tables layer.

AMashenkov (changes requested):
I don't think we can operate with RAFT terminology in SQL execution layer. Let's think how to avoid using the terms in API methods.
Leader and term should be a part of some transactional context, which can be passed through execution layer and indices right down to the tables layer.

xtern:
> I don't think we can operate with RAFT terminology in SQL execution layer. Let's think how to avoid using the terms in API methods. Leader and term should be a part of some transactional context, which can be passed through execution layer and indices right down to the tables layer.

Agreed, but from my point of view, we are currently blocked by [1] and/or [2] (see TODO in javadoc for ReadWriteReplicaRequest#term).

Because when we process the ReadWriteReplicaRequest, the "term" from the request is currently checked (see PartitionReplicaListener#ensureReplicaIsPrimary). It operates with request 'term', but it seems that it should check term from enlisted partitions.

So currently we can't do a RW-scan passing only the transaction id (we also need to pass the "term" somehow).

@AMashenkov Do you have any idea how to get around this?

[1] https://issues.apache.org/jira/browse/IGNITE-17256
[2] https://issues.apache.org/jira/browse/IGNITE-15083

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/IndexScanNode.java
```diff
@@ -84,7 +87,7 @@ public IndexScanNode(
             RowHandler.RowFactory<RowT> rowFactory,
             IgniteIndex schemaIndex,
             InternalIgniteTable schemaTable,
-            int[] parts,
+            ColocationGroup mapping,
```
korlov42, line 90: probably, it would be better to pass list of pairs [partition,term]

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/MappingQueryContext.java
```diff
@@ -55,4 +59,8 @@ RelOptCluster cluster() {
     public String locNodeName() {
         return locNodeName;
     }
+
+    public boolean readOnlyOperation() {
```
korlov42, line 63: javadoc is missing

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/rel/IndexScanNodeExecutionTest.java
```diff
@@ -432,12 +435,19 @@ private void validateIndexScan(
             when(rangeIterable.size()).thenReturn(1);
         }
 
+        // Assignment for 3 partitions. Partition 0 and 2 belong to the local node.
+        List<List<NodeWithTerm>> assignments = Arrays.asList(
+                Collections.singletonList(new NodeWithTerm(ectx.localNode().name())),
```
korlov42, line 440: kinda nitpicking, but `List.of` looks more succinct

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/IndexScanNode.java
xtern: Are you suggesting introducing some kind of wrapper that will hold the [int, long] pair (maybe to prevent additional boxing) or just use IgniteBiTuple<Integer, Long> and don't bother?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/IgniteTableImpl.java
```diff
@@ -480,10 +482,12 @@ private <RowT> ModifyRow deleteTuple(RowT row, ExecutionContext<RowT> ectx) {
         return new ModifyRow(new Row(schemaDescriptor, rowAssembler.build()), Operation.DELETE_ROW);
     }
 
-    private ColocationGroup partitionedGroup() {
-        List<List<String>> assignments = table.assignments().stream()
-                .map(Collections::singletonList)
-                .collect(Collectors.toList());
+    private ColocationGroup partitionedGroup(boolean collectLeaderTerms) {
+        Stream<NodeWithTerm> assignmentsStream = collectLeaderTerms
```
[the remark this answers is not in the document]
xtern: I'm not sure, but it seems to me that refreshing the leader for each partition every time we prepare a mapping might slow down the RO query execution (see `InternalTableImpl#awaitLeaderInitialization` - we refreshing them only once).
Or do you think it's not a problem?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -163,8 +176,8 @@ public static <RowT> ExecutionServiceImpl<RowT> create(
      * Constructor. TODO Documentation https://issues.apache.org/jira/browse/IGNITE-15859
      */
     public ExecutionServiceImpl(
```
korlov42, line 178: it would be nice to fix javadoc here

modules/table/src/main/java/org/apache/ignite/internal/table/InternalTable.java
```diff
@@ -374,6 +425,13 @@ Publisher<BinaryRow> lookup(
      */
     List<String> assignments();
 
+    /**
+     * Gets a list of current table leader assignments with raft term.
+     *
+     * @return List of current leader assignments with raft term.
+     */
+    List<IgniteBiTuple<String, Long>> leaderAssignmentsWithTerm();
```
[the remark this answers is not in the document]
vldpyatkov: Also, you have to understand Placement driver is not a static collection. It will be a service with a method to get a PrimaryReplica by ReplicationGroupId.

modules/table/src/main/java/org/apache/ignite/internal/table/InternalTable.java
```diff
@@ -296,6 +297,31 @@ Publisher<BinaryRow> scan(
...
+     * @param upperBound Upper search bound.
+     * @param flags Control flags. See {@link org.apache.ignite.internal.storage.index.SortedIndexStorage} constants.
+     * @param columnsToInclude Row projection.
+     * @return {@link Publisher} that reactively notifies about partition rows.
+     */
+    Publisher<BinaryRow> scan(
+            int partId,
+            UUID txId,
+            ClusterNode leaderNode,
+            long leaderTerm,
```
vldpyatkov, line 317: Leader and term need to be wrapped to PrimaryReplica class.

xtern: Done, thanks

modules/table/src/main/java/org/apache/ignite/internal/table/InternalTable.java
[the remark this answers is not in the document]
xtern: Done, thanks. 
Created `PrimaryReplica` in `table` module.

modules/table/src/main/java/org/apache/ignite/internal/table/InternalTable.java
```diff
@@ -357,6 +385,29 @@ Publisher<BinaryRow> lookup(
...
+     * @param indexId Index id.
+     * @param key Key to search.
+     * @param columnsToInclude Row projection.
+     * @return {@link Publisher} that reactively notifies about partition rows.
+     */
+    Publisher<BinaryRow> lookup(
+            int partId,
+            UUID txId,
+            ClusterNode leaderNode,
+            long leaderTerm,
```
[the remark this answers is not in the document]
xtern: Replaced with a `PrimaryReplica` object as you suggested below, thanks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/IgniteTableImpl.java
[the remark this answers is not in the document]
xtern: Patch simplified to use single method for both cases, thanks.

modules/index/src/main/java/org/apache/ignite/internal/index/SortedIndex.java
```diff
@@ -39,24 +41,28 @@ public interface SortedIndex extends Index<SortedIndexDescriptor> {
     byte INCLUDE_RIGHT = 0b10;
 
     /**
-     * Opens a range cursor for given bounds with left bound included in result and right excluded.
+     * Opens a range cursor for given bounds. Inclusion of the bounds is defined by {@code includeBounds} mask.
```
zstan, line 44: what **includeBounds** are you talking about ? can u explain ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -425,17 +423,17 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
 
                                 boolean implicitTxRequired = outerTx == null && rwOp;
 
-                                InternalTransaction implicitTx = implicitTxRequired ? txManager.begin() : null;
+                                InternalTransaction tx = implicitTxRequired ? txManager.begin()
+                                        : outerTx != null ? outerTx : new TransferredTxAttributesHolder(UUID.randomUUID(), clock.now());
```
zstan, line 427: why you use UUID.randomUUID() here and not tx.id() ? is it correct ?

zstan: oh, seems it correct for now, but it will fail distributed operations as i see.

modules/index/src/main/java/org/apache/ignite/internal/index/SortedIndex.java
xtern: Thanks, fixed.
> can u explain ?
This is just a copy-paste from the methods below) I fixed javadoc for them too.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
xtern: No, it will not fail distributed operations, because for read-only tx only timestamp matters.
see https://github.com/apache/ignite-3/blob/main/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java#L884

The reason why I added random UUID here - because it seems to me better then passing `null` (since the real transaction must always have ID).
But may be you right, since in future this class will not implement transaction it can not have ID at all.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
```diff
@@ -194,30 +186,74 @@ public ColocationGroup colocate(ColocationGroup other) throws ColocationMappingE
         return new ColocationGroup(sourceIds, nodeNames, assignments);
     }
 
+    private List<NodeWithTerm> intersect(List<NodeWithTerm> assign0, List<NodeWithTerm> assign1, @Nullable Set<String> filter, int partId)
```
AMashenkov, line 189: 'filter' can be Predicate and non-nullable.
So, for 'nodeNames == null' you can pass predicate '() -> true'

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
```diff
@@ -194,30 +186,74 @@ public ColocationGroup colocate(ColocationGroup other) throws ColocationMappingE
...
+            throws ColocationMappingException {
+        List<NodeWithTerm> intersection = new ArrayList<>();
+
+        for (NodeWithTerm nodeWithTerm : assign0) {
+            if (filter != null && !filter.contains(nodeWithTerm.name())) {
+                continue;
+            }
+
+            for (NodeWithTerm otherNodeWithTerm : assign1) {
```
AMashenkov, line 198: This solution has O(n*m) complexity.
Maybe it make sense to convert 'assign0' to HashMap (nodeName -> nodeWithTerm),
then validate 'assign1' items, which are in the Map and filter out others.
This will be O(n) + O(m).

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
```diff
@@ -230,29 +266,30 @@ public ColocationGroup mapToNodes(List<String> nodeNames) {
...
+     * Returns list of pairs containing the partition number to scan on the given node with the corresponding raft group term.
      *
-     * @param nodeNames Cluster node consistent ID.
-     * @return List of partitions to scan on the given node.
+     * @param nodeName Cluster node consistent ID.
+     * @return List of pairs containing the partition number to scan on the given node with the corresponding raft group term.
      */
-    public int[] partitions(String nodeNames) {
-        IgniteIntList parts = new IgniteIntList(assignments.size());
+    public List<IgniteBiTuple<Integer, Long>> partitionsWithTerms(String nodeName) {
```
AMashenkov: Let avoid using BiTuple and replace it with certain class or replace the List with
e.g. with Int2LongOpenHashMap

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
xtern: Done, thanks

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
xtern: Thanks, reworked a bit using hashing by node name.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
xtern: Added `PartitionWithTerm` class.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
xtern: Created a separate ticket for this improvement https://issues.apache.org/jira/browse/IGNITE-18691

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
```diff
@@ -194,30 +194,101 @@ public ColocationGroup colocate(ColocationGroup other) throws ColocationMappingE
...
+            if (assignment0.size() > assignment1.size()) {
+                List<NodeWithTerm> tmp = assignment0;
+                assignment0 = assignment1;
+                assignment1 = tmp;
+            }
+
+            List<NodeWithTerm> assignment = new ArrayList<>();
+
+            Map<String, NodeWithTerm> hashedByNameAssignment =
+                    assignment1.stream().collect(Collectors.toMap(NodeWithTerm::name, nodeWithTerm -> nodeWithTerm));
```
AMashenkov, line 220:
```suggestion
                  assignment0.stream().filter(filter).collect(Collectors.toMap(NodeWithTerm::name, nodeWithTerm -> nodeWithTerm));
```
Add to map items from smallest collection that are accepted by filter.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
```diff
@@ -194,30 +194,101 @@ public ColocationGroup colocate(ColocationGroup other) throws ColocationMappingE
...
+
+            for (NodeWithTerm first : assignment0) {
+                if (!filter.test(first.name())) {
+                    continue;
+                }
```
AMashenkov, line 225: This check is not required if 'hashedByNameAssignment' already contains filtered items.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
xtern: Done, thanks

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
xtern: Done, thanks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
```diff
@@ -151,18 +149,17 @@ public ColocationGroup colocate(ColocationGroup other) throws ColocationMappingE
...
             assignments = firstNotNull(this.assignments, other.assignments);
 
             if (assignments != null && nodeNames != null) {
-                Set<String> filter = new HashSet<>(nodeNames);
-                List<List<String>> assignments0 = new ArrayList<>(assignments.size());
+                List<List<NodeWithTerm>> assignments0 = new ArrayList<>(assignments.size());
 
                 for (int i = 0; i < assignments.size(); i++) {
-                    List<String> assignment = Commons.intersect(filter, assignments.get(i));
+                    List<NodeWithTerm> assignment = filterByNodeNames(assignments.get(i), new HashSet<>(nodeNames));
```
AMashenkov, line 160: Let's create HashMap outside the loop.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
```diff
@@ -194,30 +194,101 @@ public ColocationGroup colocate(ColocationGroup other) throws ColocationMappingE
...
+                    + ']');
+        }
+    }
+
+    private List<NodeWithTerm> filterByNodeNames(List<NodeWithTerm> assignment, Set<String> filter) {
+        if (nullOrEmpty(assignment) || nullOrEmpty(filter)) {
+            return Collections.emptyList();
+        }
+
+        List<NodeWithTerm> res = new ArrayList<>();
```
AMashenkov, line 257:
```suggestion
        List<NodeWithTerm> res = new ArrayList<>(assignment.size());
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
```diff
@@ -230,29 +300,30 @@ public ColocationGroup mapToNodes(List<String> nodeNames) {
...
      *
-     * @param nodeNames Cluster node consistent ID.
-     * @return List of partitions to scan on the given node.
+     * @param nodeName Cluster node consistent ID.
+     * @return List of pairs containing the partition number to scan on the given node with the corresponding primary replica term.
      */
-    public int[] partitions(String nodeNames) {
-        IgniteIntList parts = new IgniteIntList(assignments.size());
+    public List<PartitionWithTerm> partitionsWithTerms(String nodeName) {
+        List<PartitionWithTerm> partsWithTerms = new ArrayList<>();
```
AMashenkov, line 313:
```suggestion
        List<PartitionWithTerm> partsWithTerms = new ArrayList<>(assignments.size());
```

xtern: I'm not sure if this is a good improvement, because it will most likely result in a constant small memory leak. In most cases, only a some part of the partitions belongs to a specified node.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
xtern: Fixed, thanks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
xtern: Done, thanks.

AMashenkov (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/ColocationGroup.java
```diff
@@ -194,30 +194,96 @@ public ColocationGroup colocate(ColocationGroup other) throws ColocationMappingE
         return new ColocationGroup(sourceIds, nodeNames, assignments);
     }
 
+    private List<NodeWithTerm> intersect(List<NodeWithTerm> assignment0, List<NodeWithTerm> assignment1, Predicate<String> filter, int p)
+            throws ColocationMappingException {
```
AMashenkov, line 198:
```suggestion
    private List<NodeWithTerm> intersect(
            List<NodeWithTerm> assignment0, 
            List<NodeWithTerm> assignment1,
            Predicate<String> filter, 
            int p
    ) throws ColocationMappingException {
```

PR #1508, opened 2023-01-10
IGNITE-18121 add  scale up scheduler

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-18121


modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -402,22 +446,35 @@ public CompletableFuture<?> onCreate(ConfigurationNotificationEvent<Distribution
...
+            updateMetaStorageOnZoneUpdate(ctx.storageRevision(), zoneId);
+
+            int oldScaleUp = ctx.oldValue().dataNodesAutoAdjustScaleUp();
 
-            //TODO: Also add here rescheduling for the existing timers https://issues.apache.org/jira/browse/IGNITE-18121
+            int newScaleUp = ctx.newValue().dataNodesAutoAdjustScaleUp();
+
+            if (newScaleUp != Integer.MAX_VALUE && oldScaleUp != newScaleUp) {
+                // It is safe to zonesTimers.get(zoneId) in term of NPE because meta storage notifications are one-threaded
+                zonesTimers.get(zoneId).rescheduleOngoingScaleUp(newScaleUp);
```
sanpwc, line 466: Why do you expect mandatory ongoing timer at this point? Let's say that that timer was already fired, should we consider it as an ongoing one, probably not.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -806,4 +866,230 @@ private static Throwable unwrapDistributionZoneExceptionRecursively(Throwable e,
...
+
+        private void rescheduleScaleUp(long delay, Runnable runnable) {
+            if (scaleUp == null) {
+                scaleUp = new Timer(runnable);
+            }
+
+            scaleUp.reschedule(delay);
+        }
+
+        private void rescheduleOngoingScaleUp(long delay) {
```
sanpwc, line 988: Provided logic with (ongoing)rescheduling is error-prone and cumbersome, please rethink and reimplement it. There are 3 possible states for timers:
1. no scheduled task. 
2. task is scheduled but not fired yet.
3. task is executing right now.

In case of 1 and 3 we should schedule new task, so it's possible for current timer to execute one right now and have one more in a pending queue.
In case of option 2 we should reschedule the task. 
Aforementioned scheduling and rescheduling should be thread-safe.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -806,4 +866,230 @@ private static Throwable unwrapDistributionZoneExceptionRecursively(Throwable e,
...
+    }
+
+    private void initTriggerKeysAndDataNodes(int zoneId) {
+        If iifScaleUp = If.iif(
+                notExists(zoneScaleUpChangeTriggerKey(zoneId)),
+                ops(put(zoneScaleUpChangeTriggerKey(zoneId), ByteUtils.longToBytes(0L))).yield(true),
+                ops().yield(false)
+        );
+
+        metaStorageManager.invoke(iifScaleUp);
```
sanpwc, line 1077: Here and in updateMetaStorage... you are ignoring invoke result. It's probably not a good idea ;)

sanpwc (changes requested):

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -806,4 +866,230 @@ private static Throwable unwrapDistributionZoneExceptionRecursively(Throwable e,
...
+                    triggerScaleUpScaleDownKeysCondition(scaleUpTriggerRevision, scaleDownTriggerRevision, zoneId),
+                    dataNodesAndTriggerKeyUpd,
+                    ops().yield(false)
+            );
+
+            return metaStorageManager.invoke(iif).thenApply(StatementResult::getAsBoolean).thenCompose(invokeResult -> {
+                if (invokeResult) {
+                    lockForZone.lock();
+                    try {
+                        zoneState.nodesToAdd().clear();
```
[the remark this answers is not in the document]
alievmirza: yeah, thats obviously an inattention from my side, algorithm in pseudocode has the logic that you've described

sanpwc (approved):

PR #1518, opened 2023-01-12
IGNITE-18542 Remove some server-side MetaStorage classes

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-18542


modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/watch/WatchAggregatorTest.java
```diff
@@ -274,8 +274,8 @@ private Entry entry(String key, String val, long revision, long updateCntr) {
         return new Entry() {
             /** {@inheritDoc} */
             @Override
-            public @NotNull ByteArray key() {
-                return new ByteArray(key);
+            public @NotNull byte[] key() {
```
rpuch, line 277: Let's remove `@NotNull`

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/DistributedConfigurationStorage.java
```diff
@@ -160,11 +163,11 @@ public void close() {
                     // Meta Storage should not return nulls as values
                     assert value != null;
 
-                    if (key.equals(MASTER_KEY)) {
+                    if (Arrays.equals(key, MASTER_KEY.bytes())) {
                         continue;
                     }
 
-                    String dataKey = key.toString().substring(DISTRIBUTED_PREFIX.length());
+                    String dataKey = new String(key, StandardCharsets.UTF_8).substring(DISTRIBUTED_PREFIX.length());
```
rpuch, line 170: `UTF_8` can be imported statically decreasing the amount of syntactic noise: it is not that important that it goes from `StandardCharsets`; the important piece of information here is that it's utf-8.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/EntryImpl.java
```diff
@@ -19,17 +19,32 @@
...
+ *     <li>value - a data which is associated with a key and represented as an array of bytes.</li>
+ *     <li>revision - a number which denotes a version of whole meta storage.
+ *     Each change (which could include multiple entries) increments the revision. </li>
+ *     <li>updateCounter - a number which increments on every update in the change under one revision.</li>
+ * </ul>
+ *
+ * <p>Instance of {@link #EntryImpl} could represent:
+ * <ul>
+ *     <li>A regular entry which stores a particular key, a value and a revision number.</li>
+ *     <li>An empty entry which denotes absence a regular entry in the meta storage for a given key.
```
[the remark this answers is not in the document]
sashapolo: These javadocs have been copy-pasted from other class, so they may have some errors)

modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/watch/WatchAggregatorTest.java
sashapolo: This class will be removed altogether in my next PR, so I don't think it's necessary

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/DistributedConfigurationStorage.java
sashapolo: sure

modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/Entry.java
```diff
@@ -37,7 +35,7 @@ public interface Entry {
      *
      * @return The key.
      */
-    @NotNull ByteArray key();
+    byte[] key();
```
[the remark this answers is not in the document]
sashapolo: `ByteArray` doesn't make much sense, I think it needs to be removed altogether. First, its just unnecessary wrapping and object creation (you can compare arrays without it). Secondly, it makes an incorrect assumption that arrays are compared using signed comparison, however we use unsigned comparison in all storages

modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/dsl/Operation.java
```diff
@@ -17,158 +17,90 @@
...
-         * Returns key.
-         *
-         * @return Key.
-         */
-        @Nullable
-        @Override
-        public byte[] key() {
-            return key;
-        }
+    public static Operation put(byte[] key, byte[] val) {
```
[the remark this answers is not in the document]
sashapolo: I forgot about them, sorry

rpuch (approved):

rpuch (approved):

PR #1524, opened 2023-01-16
IGNITE-18464 Sql. Colocated sort aggregates need to compose a plans with additional sort

zstan:
…ith additional sort


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSetOpTest.java
```diff
@@ -105,7 +105,7 @@ public void testExceptNested() {
     }
 
     @Test
-    @Disabled("https://issues.apache.org/jira/browse/IGNITE-18469")
+    //@Disabled("https://issues.apache.org/jira/browse/IGNITE-18469")
```
AMashenkov, line 108: Let's either uncomment or remove the comment

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSortAggregateTest.java
```diff
@@ -67,8 +70,118 @@ public void mapReduceAggregate() {
...
+            res = sql(appendDisabledRules("SELECT COUNT(a), COUNT(DISTINCT(a)) FROM testMe", disabledRules));
+            res0 = transformResults(res);
+            assertEquals(1, res.size());
+            assertEquals(List.of(7L, 4L), res0.get(0));
+        } finally {
+            sql("DROP TABLE testMe");
+        }
+    }
+
+    private String appendDisabledRules(String sql, String rules) {
```
AMashenkov, line 162: Maybe, it would be better to extend parent `sql` method with disabled rules instead of this one. WDYT?
```
   String sql(String sql, String... rules) {
```
or
```
    private String appendDisabledRules(String sql, List<String> rules) {
```

modules/runner/src/integrationTest/sql/aggregate/aggregates/test_aggr_string.test
```diff
@@ -80,3 +117,6 @@ SELECT MAX(s), MIN(s) FROM test_strings2;
 aa	A
 
 
+
+
+
```
AMashenkov, line 122:
```suggestion
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AggregatePlannerTest.java
```diff
@@ -389,52 +390,45 @@ public void mapAggregateWithoutDistinctAcc() throws Exception {
...
+     * Check that plan does not contain distinct accumulators on map nodes with additional expectations.
      *
      * @param sql Request string.
      * @param publicSchema Schema.
      * @throws Exception If failed.
      */
     private void checkDistinctInMapAggNode(String sql, IgniteSchema publicSchema) throws Exception {
-        List<Pair<String[], Boolean>> disabledRules = List.of(new Pair<>(new String[]{""}, false),
-                new Pair<>(new String[]{"ColocatedHashAggregateConverterRule", "ColocatedSortAggregateConverterRule"}, true),
+        log.info("++check: " + sql);
```
AMashenkov, line 417:
```suggestion
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AggregatePlannerTest.java
```diff
@@ -389,52 +390,45 @@ public void mapAggregateWithoutDistinctAcc() throws Exception {
...
+                        "MapReduceSortAggregateConverterRule"}, node -> !findNodes(node, byClass(IgniteMapAggregateBase.class)).isEmpty()),
+                new Pair<>(new String[]{"MapReduceHashAggregateConverterRule", "MapReduceSortAggregateConverterRule",
+                        "ColocatedHashAggregateConverterRule"}, node -> true),
+                new Pair<>(new String[]{"MapReduceHashAggregateConverterRule", "MapReduceSortAggregateConverterRule",
+                        "ColocatedSortAggregateConverterRule"}, node -> true)
         );
 
-        for (Pair<String[], Boolean> rules : disabledRules) {
+        for (Pair<String[], Predicate<IgniteRel>> rules : disabledRules) {
+            log.info("+++disable rules: " + Arrays.toString(rules.getFirst()));
```
AMashenkov, line 428:
```suggestion
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AggregatePlannerTest.java
```diff
@@ -389,52 +390,45 @@ public void mapAggregateWithoutDistinctAcc() throws Exception {
...
+        List<Pair<String[], Predicate<IgniteRel>>> disabledRules = List.of(
                 new Pair<>(new String[]{"ColocatedHashAggregateConverterRule", "ColocatedSortAggregateConverterRule",
-                        "MapReduceSortAggregateConverterRule"}, true),
-                new Pair<>(new String[]{"ColocatedHashAggregateConverterRule", "ColocatedSortAggregateConverterRule",
-                        "MapReduceHashAggregateConverterRule"}, true)
+                        "MapReduceSortAggregateConverterRule"}, node -> !findNodes(node, byClass(IgniteMapAggregateBase.class)).isEmpty()),
+                new Pair<>(new String[]{"MapReduceHashAggregateConverterRule", "MapReduceSortAggregateConverterRule",
+                        "ColocatedHashAggregateConverterRule"}, node -> true),
+                new Pair<>(new String[]{"MapReduceHashAggregateConverterRule", "MapReduceSortAggregateConverterRule",
+                        "ColocatedSortAggregateConverterRule"}, node -> true)
```
AMashenkov, line 424: I've just found we use same rule-sets in different tests to disable some features.
We can extract e.g. 'enum' with these rules, which forces planner to force/disable features.

Please take a look at my PR, maybe we can go the better way here and there.
 https://github.com/apache/ignite-3/pull/1536

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSortAggregateTest.java
zstan: I understand but seems that check output results with **sql()** is not comfortable as with QueryChecker (which need to be rafactored a bit, issue is linked above) let`s do it in checker ?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AggregatePlannerTest.java
zstan: yep, but in case of exception someone need to append the same logging for understanding the case. Do you still think i need to remove it ?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AggregatePlannerTest.java
zstan: yep, but in case of exception someone need to append the same logging for understanding the case. Do you still think i need to remove it ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/agg/IgniteColocatedSortAggregate.java
```diff
@@ -57,9 +57,6 @@ public IgniteColocatedSortAggregate(
     ) {
         super(cluster, traitSet, input, groupSet, groupSets, aggCalls);
 
-        assert !TraitUtils.collation(traitSet).isDefault();
-        assert !groupSet.isEmpty() && groupSets.size() == 1;
```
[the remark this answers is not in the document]
zstan: another case become possible. for example select *

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AggregatePlannerTest.java
zstan: thanks, i check your PR and have no clear ideas here, probably after your and my issues have been merged we make additional issue and introduce such an enum ?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AggregatePlannerTest.java
AMashenkov: You can add a message to the next assertTrue() call.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rule/AbstractIgniteConverterRule.java
```diff
@@ -24,6 +24,7 @@
 import org.apache.calcite.rel.convert.ConverterRule;
 import org.apache.calcite.rel.metadata.RelMetadataQuery;
 import org.apache.ignite.internal.sql.engine.rel.IgniteConvention;
+import org.checkerframework.checker.nullness.qual.Nullable;
```
korlov42, line 27: I believe it has to be `org.jetbrains.annotations`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rule/SortAggregateConverterRule.java
```diff
@@ -36,6 +36,7 @@
 import org.apache.ignite.internal.sql.engine.trait.IgniteDistributions;
 import org.apache.ignite.internal.sql.engine.trait.TraitUtils;
 import org.apache.ignite.internal.sql.engine.util.HintUtils;
+import org.checkerframework.checker.nullness.qual.Nullable;
```
korlov42, line 39: the same

AMashenkov (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/SortAggregateNode.java
```diff
@@ -76,7 +75,6 @@ public SortAggregateNode(
             Comparator<RowT> comp
     ) {
         super(ctx);
-        assert Objects.nonNull(comp);
```
korlov42, line 79: I would prefer to leave this assertion as is. 

Without this assertion, a situation in which we would have previously obtained either an assertion error or an NPE would now lead to the wrong result.

Perhaps, we could cover single group case by providing comparator returning 0 every time, WDYT?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/SortAggregateNode.java
```diff
@@ -159,6 +161,10 @@ public void end() throws Exception {
 
         waiting = -1;
 
+        if (grp == null && accFactory != null) {
```
korlov42, line 164: this will produce empty row when none is expected in query like `SELECT min(c) FROM t GROUP BY a` when table is empty. BTW, we need execution test for this case

to initialise group properly, please, take a look at `org.apache.ignite.internal.sql.engine.exec.rel.HashAggregateNode.Grouping#init`

zstan: got it!

zstan:
@korlov42 i fix all your comments, plz check ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/SortAggregateNode.java
```diff
@@ -83,6 +84,10 @@ public SortAggregateNode(
         this.rowFactory = rowFactory;
         this.grpSet = grpSet;
         this.comp = comp;
+
+        if ((type == AggregateType.REDUCE || type == AggregateType.SINGLE) && accFactory != null && grpSet.isEmpty()) {
```
korlov42, line 88: what about rewind? (<-- let's add execution test for this case as well)

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/SortAggregateNode.java
```diff
@@ -119,7 +124,11 @@ public void push(RowT row) throws Exception {
         waiting--;
 
         if (grp != null) {
-            int cmp = comp.compare(row, prevRow);
+            int cmp = 0;
+
+            if (comp != null) {
```
korlov42, line 129: comparator is never null

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSortAggregateTest.java
```diff
@@ -48,12 +53,10 @@ static void initTestData() {
 
     @Test
     public void mapReduceAggregate() {
+        String disabledRules = " /*+ DISABLE_RULE('MapReduceHashAggregateConverterRule', 'ColocatedSortAggregateConverterRule') */ ";
```
korlov42, line 56: does it make sense to disable colocated hash aggregate rule as well?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSortAggregateTest.java
```diff
@@ -67,8 +70,134 @@ public void mapReduceAggregate() {
 
     @Test
     public void correctCollationsOnMapReduceSortAgg() {
-        var cursors = sql("SELECT PK FROM TEST_ONE_COL_IDX WHERE col0 IN (SELECT col0 FROM TEST_ONE_COL_IDX)");
+        String disabledRules = " /*+ DISABLE_RULE('MapReduceHashAggregateConverterRule', 'ColocatedSortAggregateConverterRule') */ ";
```
korlov42, line 73: the same question here

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSortAggregateTest.java
```diff
@@ -67,8 +70,134 @@ public void mapReduceAggregate() {
...
+            assertEquals(1, res.size());
+            assertEquals(List.of(7L, 4L), res0.get(0));
+        } finally {
+            sql("DROP TABLE testMe");
+        }
+    }
+
+    @WithSystemProperty(key = "IMPLICIT_PK_ENABLED", value = "true")
+    @Test
+    public void checkEmptyTable() {
```
korlov42, line 164: does it make sense to verify both colocated and 2-phase aggregates?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/SortAggregateNode.java
korlov42: let's add _execution_ test for this case

zstan:
@korlov42 all done

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSortAggregateTest.java
```diff
@@ -67,8 +71,142 @@ public void mapReduceAggregate() {
...
+        var cursors = sql(
+                appendDisabledRules("SELECT PK FROM TEST_ONE_COL_IDX WHERE col0 IN (SELECT col0 FROM TEST_ONE_COL_IDX)", disabledRules));
 
         assertEquals(ROWS, cursors.size());
     }
+
+    @Test
+    @WithSystemProperty(key = "IMPLICIT_PK_ENABLED", value = "true")
+    //TODO: rewrite with QueryChecker after
+    // https://issues.apache.org/jira/browse/IGNITE-18501 (QueryChecker still use deprecated QueryProcessor#queryAsync method)
```
ygerzhedovich, line 86: The main branch already fixed and use querySingleAsync method

korlov42 (approved):

PR #1542, opened 2023-01-17
IGNITE-18554 Remove MetaStorage learners on topology events

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-18554


modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageRaftGroupEventsListener.java
```diff
@@ -0,0 +1,247 @@
...
+        // before sending the Join Ready request (i.e. before it gets added to the Logical Topology).
+        topologyService.addEventHandler(new TopologyEventHandler() {
+            @Override
+            public void onDisappeared(ClusterNode member) {
+                executeIfLeader(service -> removeLearners(service.raftGroupService(), logicalTopology.getLogicalTopology()));
+            }
+        });
+
+        logicalTopology.addEventListener(new LogicalTopologyEventListener() {
+            // Try to add a node as a Learner when it appears in the Logical Topology. Even though usually remote nodes add themselves as
```
sanpwc, line 107: Could you please elaborate the flow in a greater detail? I didn't get the case.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageRaftGroupEventsListener.java
```diff
@@ -0,0 +1,247 @@
...
+            // though it will skip the step when it adds itself as a Learner.
+            @Override
+            public void onAppeared(ClusterNode appearedNode, LogicalTopologySnapshot newTopology) {
+                executeIfLeader(service -> addLearner(service.raftGroupService(), appearedNode));
+            }
+
+            // Try to remove a Meta Storage Learner when a node leaves the Logical Topology. This approach protects from network glitches
+            // when a node leaves and re-joins the Physical Topology.
+            @Override
+            public void onDisappeared(ClusterNode disappearedNode, LogicalTopologySnapshot newTopology) {
```
sanpwc, line 119: Seems that we don't have happens before guarantees here. In other words it's possible to handle logical.onDisappeared(A) while processing physical.onAppeared(A) just because one is about in-raft CMG thread and another is network topology internals.

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItMetaStorageMultipleNodesTest.java
```diff
@@ -0,0 +1,338 @@
...
+
+        // Check that the second node has been registered as a learner.
+        assertThat(firstNode.getMetaStorageLearners(), willBe(Set.of(secondNode.name())));
+    }
+
+    /**
+     * Tests a case when a node leaves the physical topology without entering the logical topology.
+     */
+    @Test
+    void testLearnerLeavePhysicalTopology(TestInfo testInfo) throws Exception {
```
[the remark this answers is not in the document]
sashapolo: I do not invoke `cmgManager.onJoinReady()`

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageRaftGroupEventsListener.java
```diff
@@ -0,0 +1,247 @@
...
+            CompletableFuture<MetaStorageServiceImpl> metaStorageSvcFut
+    ) {
+        this.busyLock = busyLock;
+        this.topologyService = clusterService.topologyService();
+        this.logicalTopology = logicalTopology;
+        this.metaStorageSvcFut = metaStorageSvcFut;
+    }
+
+    @Override
+    public void onLeaderElected(long term) {
```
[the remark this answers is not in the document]
sashapolo: Yes, both `addLearners` and `removeLearners` are idempotent, it's basically adding or removing a `PeerId` from a `HashSet`. See `modules/raft/src/main/java/org/apache/ignite/raft/jraft/conf/Configuration.java` `addLearner` and `removeLearner` methods.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageRaftGroupEventsListener.java
sashapolo: Normal flow:
1. A node gets discovered by the CMG leader using the Physical Topology.
2. The node gets validated on the CMG leader.
3. The CMG leader sends MS node IDs to the joining node.
4. The joining node starts a local Raft server and executes `addLearners` to add itself as a learner.
5. The joining node executes `onJoinReady` and gets added to the Logical Topology.

Now imagine that after that a network partition happens for a long enough time for the node to get kicked out of the Logical Topology. As soon as that happens, it gets removed from the MS learners as well. When the network partition is fixed, the node will be added back to the Logical Topology, but it will not be added as MS learner (because the node was not restarted and the code that starts the Raft server will not be executed).

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageRaftGroupEventsListener.java
```diff
@@ -0,0 +1,247 @@
...
+        executeIfLeader(service -> removeLearners(service.raftGroupService(), logicalTopology.getLogicalTopology()));
+    }
+
+    private void registerTopologyEventListeners() {
+        // Try to remove a Meta Storage Learner if a node leaves the Physical Topology. This handles the case when a node joins, but leaves
+        // before sending the Join Ready request (i.e. before it gets added to the Logical Topology).
+        topologyService.addEventHandler(new TopologyEventHandler() {
+            @Override
+            public void onDisappeared(ClusterNode member) {
+                executeIfLeader(service -> removeLearners(service.raftGroupService(), logicalTopology.getLogicalTopology()));
```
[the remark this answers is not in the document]
sashapolo: > Are there any order guarantees between messaging and network topology events?

I don't think so, at least there's no explicit guarantee. Technically these can be separate subsystems.

> In other words is it possible that we will have a race between processing nodeJoinRequest and handling onDissapeared, meaning the race between adding and removing the learner in this particular case? 

Can you please clarify the situation?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageRaftGroupEventsListener.java
sashapolo: Again, can you please clarify the situation a little bit?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageRaftGroupEventsListener.java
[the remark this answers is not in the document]
sanpwc: I mean following:
0. There's node A up and running that is also CMG and MS Leader, meaning that both CMG and MS are build on top of single node A.
1. Node B tries to join the cluster. At some point it sends `service.addLearners(List.of(localPeer))` message to the A.
2. addLearner(B) processing hangs a bit. 
3. Node A sees onDisappeared(B) and tries to remove B from learners, despite the fact that there's no B in the learners set.
4. addLearner(B) wakes up and successfully adds B as a Learner.
Thus we have a reordering between adding and removing Learner B.

So, basically, my question is about possibility of such reordering.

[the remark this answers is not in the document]
sashapolo: yes, this can happen =(

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageRaftGroupEventsListener.java
sanpwc: Got it. I would expect that joining process should be aborted if joining node was removed from physical topology and restarted from the very beginning if needed. Otherwise we may have inconsistencies in node id, etc.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageRaftGroupEventsListener.java
sanpwc: My fault. We do noting on physical.onAppeared(A), so there's no race.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageRaftGroupEventsListener.java
sashapolo: Do you suggest that if a node leaves the logical topology for whatever reason, should re-run the joining process even if it was not restarted?

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItMetaStorageMultipleNodesTest.java
```diff
@@ -0,0 +1,338 @@
...
+        firstNode.startDroppingMessagesTo(secondNode);
+
+        assertTrue(waitForCondition(() -> firstNode.getMetaStorageLearners().join().isEmpty(), 10_000));
+
+        // Make the first node discover the second node again. The second node should be added as a Meta Storage Learner again.
+        firstNode.stopDroppingMessages();
+
+        assertTrue(waitForCondition(() -> firstNode.getMetaStorageLearners().join().equals(Set.of(secondNode.name())), 10_000));
+    }
+}
```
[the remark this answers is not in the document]
sashapolo: I'm writing such test, it's going to take a while(

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageRaftGroupEventsListener.java
sanpwc: However, seems, that we may have a race on CMG leadership transfer.
1. C joins the topology and register itself as a learner.
2. B becomes MS leader and fires executeIfLeader(service -> removeLearners(service.raftGroupService(), logicalTopology.getLogicalTopology())); However B doesn't know yet that C is part of logical topology, so it removes it.
3. Yep, you do expect that when onAppeared event will be fired you'll add C (which is wrong from my point of view), however...
4. ...A becomes MS leader, without B receiving  onAppeared(C), and A already received onAppeared(C), so no one will restore C as a learner.

sanpwc (approved):

PR #1543, opened 2023-01-18
IGNITE-18566 Fix busyLock usage in RocksDbTableStorage and related

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18566


modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -251,41 +251,36 @@ public <V> V runConsistently(WriteClosure<V> closure) throws StorageException {
         if (threadLocalWriteBatch.get() != null) {
             return closure.execute();
         } else {
-            if (!busyLock.enterBusy()) {
-                throw new StorageClosedException();
-            }
-
-            try (var writeBatch = new WriteBatchWithIndex()) {
-                threadLocalWriteBatch.set(writeBatch);
+            return busy(() -> {
```
rpuch, line 254: Aren't we concerned about allocations anymore? Each call to `runConsistently()` will allocate a `Supplier` index.

If we use it here, should we also use this pattern everywhere (at least, in all storages)?

tkalkirill: I think we won't care much about lambda allocation yet. 
For other MV storages I will do the same but in other tickets.

rpuch (approved):

PR #1545, opened 2023-01-18
IGNITE-18571 Fix busyLock usage in AbstractPageMemoryTableStorage and related

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18571

### DIFF
--- a/modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/AbstractPageMemoryTableStorage.java
+++ b/modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/AbstractPageMemoryTableStorage.java
@@ -18,6 +18,7 @@
 package org.apache.ignite.internal.storage.pagememory;
 
 import static java.util.concurrent.CompletableFuture.completedFuture;
+import static org.apache.ignite.internal.util.IgniteUtils.inBusyLock;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/AbstractPageMemoryTableStorage.java
```diff
@@ -82,16 +87,20 @@ public TablesConfiguration tablesConfiguration() {
 
     @Override
     public void start() throws StorageException {
-        TableView tableView = tableCfg.value();
-
-        mvPartitions = new AtomicReferenceArray<>(tableView.partitions());
+        inBusyLock(busyLock, () -> {
```
rpuch, line 90: Let's reference the busy lock in `inBusyLock()` method using the field directly instead of passing it via a parameter. We have only one busy lock in this object, so we can reference it directly. This will reduce the amount of syntactic noise.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/AbstractPageMemoryTableStorage.java
```diff
@@ -167,86 +180,94 @@ public CompletableFuture<Void> destroy() {
...
+            mvPartitions.set(partitionId, partition);
 
-        return partition;
+            return partition;
+        });
     }
 
     @Override
     public @Nullable AbstractPageMemoryMvPartitionStorage getMvPartition(int partitionId) {
-        assert started : "Storage has not started yet";
```
rpuch, line 187: Why don't why check whether the storage has been started anymore?

tkalkirill: I think we no longer need this check, if we have not started, then field `AbstractPageMemoryTableStorage#mvPartitions` will be `null`.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/AbstractPageMemoryTableStorage.java
tkalkirill: Discussed in person.

rpuch (approved):

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/AbstractPageMemoryTableStorage.java
tkalkirill: Discussed in person.

PR #1548, opened 2023-01-19
IGNITE-18583 Fix C++ build

isapego:
- Improved DEVNOTES.md;
- Fixed compilation errors and warnings;
- Introduce single-node mode to lower number of TC fails until causing issue is fixed;

### DIFF
--- a/modules/platforms/cpp/ignite/client/detail/node_connection.h
+++ b/modules/platforms/cpp/ignite/client/detail/node_connection.h
@@ -115,6 +115,7 @@ public:
             }
         }
 
+        m_logger->log_debug("Performing request: op=" + std::to_string(int(op)) + ", req_id=" + std::to_string(reqId));
         bool sent = m_pool->send(m_id, std::move(message));
         if (!sent) {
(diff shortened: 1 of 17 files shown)


modules/platforms/cpp/ignite/client/detail/node_connection.h
```diff
@@ -115,6 +115,7 @@ class node_connection : public std::enable_shared_from_this<node_connection> {
             }
         }
 
+        m_logger->log_debug("Performing request: op=" + std::to_string(int(op)) + ", req_id=" + std::to_string(reqId));
```
ptupitsyn, line 118: Does this affect performance even when debug logging is disabled? String concatenation is performed anyway, right?

ptupitsyn (changes requested):

modules/platforms/cpp/ignite/client/detail/node_connection.h
isapego: You are right. I'll fix it.

ptupitsyn (approved):

PR #1550, opened 2023-01-19
IGNITE-18585 Sql. Introduce cache for serialized RelNodes representation

zstan:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -83,6 +86,13 @@
  * ExecutionServiceImpl. TODO Documentation https://issues.apache.org/jira/browse/IGNITE-15859
  */
 public class ExecutionServiceImpl<RowT> implements ExecutionService, TopologyEventHandler {
+    private static final int CACHE_SIZE = 1024;
+
+    private static final ConcurrentMap<String, IgniteRel> PHYS_NODES_REPR = Caffeine.newBuilder()
```
korlov42, line 91: honestly speaking, I'm not sure this cache should be static. This can affect the tests in an undesirable way, with no real benefit

zstan: ok, maybe you right .. .i remove static from there.

korlov42 (approved):

PR #1551, opened 2023-01-19
IGNITE-18527 Inline writeIntents into data storage on primary replica side

denis-chudov:
https://issues.apache.org/jira/browse/IGNITE-18527


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2009,4 +2092,16 @@ private CompletableFuture<BinaryRow> toBinaryRow(@Nullable TableRow row) {
     private CompletableFuture<BinaryConverter> getConverterFuture(int schemaVersion) {
         return schemaFut.thenApply(schemaRegistry -> BinaryConverter.forRow(schemaRegistry.schema(schemaVersion)));
     }
+
+    /**
+     * Class that stores list of futures for updates that can block tx cleanup, and locked flag.
```
vldpyatkov, line 2097: Incorrect flag name.

PR #1552, opened 2023-01-19
Build Improvements

apachont:
### Changes
- Use relative path normalization in order to improve cache relocate-ability of Checkstyle, sql-parser and OpenApi cache keys
- Disable timestamp autogeneration on Swagger causing cache misses
- CheckStyle configuration for LICENSE.txt and SuppresionFilter files depends on relative paths to prevent cache misses
- Updated javacc from version 4.0.1 to 4.1.3

### Description
These changes improves the performance on incremental builds and also increase the cache availability when running builds from different locations.

[First build](https://scans.gradle.com/s/xgqhc2mlcnxzw) 
[Second build](https://scans.gradle.com/s/o7o5mizaxkjpg)


runningcode:
@vkulichenko is there any way we can investigate some of the build failures?

clayburn:
@vkulichenko - Just wanted to follow up on @runningcode's question. We'd love to help fix this in order to get it merged, but don't have visibility into the failures here without being able to log into TeamCity. Thanks!

PR #1556, opened 2023-01-20
IGNITE-18593 Get rid of MvPartitionStorage and TxStateStorage in PartitionAccess

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18593


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccess.java
```diff
@@ -32,27 +41,111 @@ public interface PartitionAccess {
...
+     *
+     * @return Future that will complete when the partition is recreated.
+     * @throws StorageException If an error has occurred during the partition destruction.
      */
-    MvPartitionStorage mvPartitionStorage();
+    CompletableFuture<Void> reCreateMvPartitionStorage() throws StorageException;
 
     /**
-     * Returns transaction state storage for the partition.
+     * Destroys and recreates the multi-versioned partition storage.
```
rpuch, line 52:
```suggestion
     * Destroys and recreates the TX state partition storage.
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccess.java
```diff
@@ -32,27 +41,111 @@ public interface PartitionAccess {
...
+     *
+     * @throws StorageException If an error has occurred during transaction state storage for the partition destruction.
```
rpuch, line 54:
```suggestion
     * @throws StorageException If an error has occurred during destruction of the transaction state storage for the partition.
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -186,32 +186,13 @@ public SnapshotReader getReader() {
...
-                        mvPartitionStorage.lastApplied(FULL_RABALANCING_STARTED, 0);
-
-                        return null;
-                    });
-
-                    LOG.info("Copier prepared multi-versioned storage for the partition [partId={}, tableId={}]", partId(), tableId());
-
-                    return completedFuture(null);
-                }, executor);
+        return partitionSnapshotStorage.partition().reCreateMvPartitionStorage();
```
rpuch, line 195: The previous version of the code logged when it prepared the MV storage. Here, the logging is removed. Was this made on purpose, or should we still log it?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -226,18 +207,7 @@ private CompletableFuture<?> prepareTxStatePartitionStorageForRebalance(Executor
...
 
-        return CompletableFuture.supplyAsync(() -> partitionSnapshotStorage.partition().reCreateTxStatePartitionStorage(), executor)
-                .thenCompose(txStatePartitionStorage -> {
-                    if (canceled) {
-                        return completedFuture(null);
-                    }
-
-                    txStatePartitionStorage.lastApplied(FULL_RABALANCING_STARTED, 0);
-
-                    LOG.info("Copier prepared transaction state storage for the partition [partId={}, tableId={}]", partId(), tableId());
```
rpuch, line 237: Same thing: don't we want to log this anymore?

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotCommonTest.java
```diff
@@ -76,12 +64,13 @@ void returnsKeyFromStorage() {
 
     @Test
     void sendsSnapshotMeta() {
-        when(mvPartitionStorage.lastAppliedIndex()).thenReturn(100L);
-        lenient().when(mvPartitionStorage.lastAppliedTerm()).thenReturn(3L);
-        when(txStateStorage.lastAppliedIndex()).thenReturn(100L);
-        lenient().when(txStateStorage.lastAppliedTerm()).thenReturn(3L);
+        lenient().when(partitionAccess.minLastAppliedIndex()).thenReturn(90L);
```
rpuch, line 67: These `lenient()`s seem redundant and should be dropped

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionSnapshotStorageFactoryTest.java
```diff
@@ -44,24 +39,18 @@
...
-    void configureMocks() {
-        when(partitionAccess.mvPartitionStorage()).thenReturn(mvPartitionStorage);
-        when(partitionAccess.txStatePartitionStorage()).thenReturn(txStateStorage);
-    }
-
     @Test
     void testForChoosingMinimumAppliedIndexForMeta() {
-        mvPartitionStorage.lastApplied(10L, 2L);
-        txStateStorage.lastApplied(5L, 1L);
+        lenient().when(partitionAccess.minLastAppliedIndex()).thenReturn(5L);
```
rpuch, line 47: These `lenient()`s seem redundant (as the corresponding methods are always called under this test), so they should be dropped.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccessImpl.java
```diff
@@ -57,48 +70,147 @@ public PartitionKey partitionKey() {
...
+            txStateStorage.lastApplied(lastAppliedIndex, lastAppliedTerm);
+
+            mvPartitionStorage.committedGroupConfiguration(raftGroupConfig);
+
+            return null;
+        });
+    }
+
+    @Override
+    public long minLastAppliedIndex() {
```
rpuch, line 170: Please add unit tests for all 4 min/max methods. This logic was implemented in other places before this change, and it was covered by tests (as it seems really important). So we probably need these tests here now.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccess.java
tkalkirill: Fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccess.java
tkalkirill: Fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
tkalkirill: This code will be removed soon, left TODO

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotCommonTest.java
tkalkirill: I tried to delete it, but got the following error: `Please remove unnecessary stubbings or use 'lenient' strictness. More info: javadoc for UnnecessaryStubbingException class.`

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionSnapshotStorageFactoryTest.java
tkalkirill: I tried to delete it, but got the following error: `Please remove unnecessary stubbings or use 'lenient' strictness. More info: javadoc for UnnecessaryStubbingException class.`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccessImpl.java
tkalkirill: I think there are already tests:
- `org.apache.ignite.internal.table.distributed.raft.snapshot.PartitionSnapshotStorageFactoryTest#testForChoosingMinimumAppliedIndexForMeta`
- `org.apache.ignite.internal.table.distributed.raft.snapshot.outgoing.OutgoingSnapshotReaderTest#testForChoosingMaximumAppliedIndexForMeta`

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotCommonTest.java
rpuch: Then how about removing the whole line? If Mockito says the stubbing is not used, then we probably don't need it at all, right?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccessImpl.java
rpuch: Both tests that you mention make sure that `PartitionAccess#min/max...` methods get called. But here, we need to make that that these methods actually work correctly (meaning, that they actually calculate min or max of the corresponding values). This is not tested anywhere.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/OutgoingSnapshotCommonTest.java
tkalkirill: Fix it

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionSnapshotStorageFactoryTest.java
tkalkirill: Fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccessImpl.java
tkalkirill: Understood you, I will add tests.

PR #1562, opened 2023-01-23
IGNITE-18030 Integration of the new full rebalance API with IncomingSnapshotCopier

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18030

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
@@ -104,8 +104,6 @@ public class IncomingSnapshotCopier extends SnapshotCopier {
 
         LOG.info("Copier is started for the partition [{}]", createPartitionInfo());
 
-        joinFuture = new CompletableFuture<>();
-
         rebalanceFuture = partitionSnapshotStorage.partition().startRebalance()
                 .thenCompose(unused -> {
                     ClusterNode snapshotSender = getSnapshotSender(snapshotUri.nodeName);
@@ -119,15 +117,7 @@ public class IncomingSnapshotCopier extends SnapshotCopier {
                             .thenCompose(unused1 -> loadSnapshotTxData(snapshotSender, executor));
                 });
 
-        rebalanceFuture.handle((unused, throwable) -> completesRebalance(throwable))
-                .thenCompose(Function.identity())
-                .whenComplete((unused, throwable) -> {
-                    if (throwable == null) {
-                        joinFuture.complete(null);
-                    } else {
-                        joinFuture.completeExceptionally(throwable);
-                    }
-                });
+        joinFuture = rebalanceFuture.handle((unused, throwable) -> completeRebalance(throwable)).thenCompose(Function.identity());
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -96,38 +102,37 @@ public IncomingSnapshotCopier(PartitionSnapshotStorage partitionSnapshotStorage,
...
                     }
 
                     return loadSnapshotMeta(snapshotSender)
                             .thenCompose(unused1 -> loadSnapshotMvData(snapshotSender, executor))
-                            .thenCompose(unused1 -> loadSnapshotTxData(snapshotSender, executor))
-                            .thenAcceptAsync(unused1 -> updateLastAppliedIndexFromSnapshotMetaForStorages(), executor);
+                            .thenCompose(unused1 -> loadSnapshotTxData(snapshotSender, executor));
+                });
+
+        rebalanceFuture.handle((unused, throwable) -> completesRebalance(throwable))
```
ibessonov, line 122: Why "completes" instead of "complete"?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -96,38 +102,37 @@ public IncomingSnapshotCopier(PartitionSnapshotStorage partitionSnapshotStorage,
...
+                .thenCompose(Function.identity())
+                .whenComplete((unused, throwable) -> {
+                    if (throwable == null) {
+                        joinFuture.complete(null);
```
ibessonov, line 126: Why do you need a second object? This is not explained and not obvious at all

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -157,9 +162,9 @@ public void join() throws InterruptedException {
     public void cancel() {
         canceled = true;
 
-        LOG.info("Copier is canceled for partition [partId={}, tableId={}]", partId(), tableId());
+        LOG.info("Copier is canceled for partition [{}]", createPartitionInfo());
 
-        CompletableFuture<?> fut = future;
+        CompletableFuture<?> fut = rebalanceFuture;
```
ibessonov, line 167: Does it work? Do we have a test that would show that "cancel" works when there are no messages received by the node?

tkalkirill: Discussed personally, will work.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
tkalkirill: Type. fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
tkalkirill: Fix it.

ibessonov (approved):

PR #1563, opened 2023-01-23
IGNITE-18358 IGN-TX-5 on concurrent transactional single key load

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-18358

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
+++ b/modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
@@ -266,7 +266,7 @@ public class ErrorGroups {
         public static final int TX_UNEXPECTED_STATE_ERR = TX_ERR_GROUP.registerErrorCode(3);
 
         /** Failed to acquire a lock on a key due to a conflict. */
-        public static final int ACQUIRE_LOCK_ERR = TX_ERR_GROUP.registerErrorCode(5);
+        public static final int ACQUIRE_LOCK_ERR = TX_ERR_GROUP.registerErrorCode(4);
 
         /** Failed to acquire a lock on a key within a timeout. */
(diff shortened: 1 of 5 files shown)


modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -265,9 +265,6 @@ public static class Transactions {
         /** Error of unexpected tx state on state change. */
         public static final int TX_UNEXPECTED_STATE_ERR = TX_ERR_GROUP.registerErrorCode(3);
 
-        /** Failed to release a lock on a key. */
-        public static final int RELEASE_LOCK_ERR = TX_ERR_GROUP.registerErrorCode(4);
-
         /** Failed to acquire a lock on a key due to a conflict. */
         public static final int ACQUIRE_LOCK_ERR = TX_ERR_GROUP.registerErrorCode(5);
```
denis-chudov, line 270: shouldnt we descrease the numbers of ACQUIRE_LOCK_ERR and following codes?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -20,7 +20,9 @@
 import static java.util.concurrent.CompletableFuture.completedFuture;
 import static java.util.concurrent.CompletableFuture.failedFuture;
 import static org.apache.ignite.internal.util.ExceptionUtils.withCause;
+import static org.apache.ignite.lang.ErrorGroups.Common.UNKNOWN_ERR;
```
denis-chudov, line 23: why UNKNOWN_ERR is deprecated? im not sure that it is good pattern to use deprecated value, but if it is ok, the annotation should be removed

PR #1565, opened 2023-01-23
IGNITE-18086 Add index building in IncomingSnapshotCopier

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18086

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccessImpl.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccessImpl.java
@@ -141,7 +141,7 @@ public class PartitionAccessImpl implements PartitionAccess {
     }
 
     @Override
-    public void addWrite(RowId rowId, TableRow row, UUID txId, UUID commitTableId, int commitPartitionId) {
+    public void addWrite(RowId rowId, @Nullable TableRow row, UUID txId, UUID commitTableId, int commitPartitionId) {
         MvPartitionStorage mvPartitionStorage = getMvPartitionStorage(partitionId());
 
         mvPartitionStorage.runConsistently(() -> {
@@ -154,7 +154,7 @@ public class PartitionAccessImpl implements PartitionAccess {
     }
 
     @Override
-    public void addWriteCommitted(RowId rowId, TableRow row, HybridTimestamp commitTimestamp) {
+    public void addWriteCommitted(RowId rowId, @Nullable TableRow row, HybridTimestamp commitTimestamp) {
         MvPartitionStorage mvPartitionStorage = getMvPartitionStorage(partitionId());
 
         mvPartitionStorage.runConsistently(() -> {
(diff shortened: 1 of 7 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccessImpl.java
```diff
@@ -213,4 +229,8 @@ private TxStateStorage getTxStateStorage(int partitionId) {
 
         return txStateStorage;
     }
+
+    private void addToIndexes(TableRow tableRow, RowId rowId) {
+        indexes.get().forEach(index -> index.put(tableRow, rowId));
```
rpuch, line 234: It looks like we forgot about deletions. For a deletion, on the sending side, the `ReadResult.tableRow()` will return `null`, and `ResponseEntry.rowVersions[i]' will be `null`, so `IncomingSnapshotCopier` on line 268

```
TableRow tableRow = new TableRow(entry.rowVersions().get(i).rewind());
```

should make `tableRow = null` for a null `rowVersion` buffer, so `addToIndexes()` will get a `null` `TableRow`, so it should check for null and do not add anything to an index for a null value (actually, the code in `PartitionListener` seems to do just this).

rpuch: Also, we probably need to add a scenario with a DELETE to our IT tests for this feature. Existing tests only do INSERTs.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccessImplTest.java
```diff
@@ -119,4 +128,56 @@ void testMinMaxLastAppliedTerm(@InjectConfiguration("mock.tables.foo {}") Tables
...
+        );
+
+        RowId rowId = new RowId(TEST_PARTITION_ID);
+        TableRow tableRow = mock(TableRow.class);
+        UUID txId = UUID.randomUUID();
+        UUID commitTableId = UUID.randomUUID();
+
+        partitionAccess.addWrite(rowId, tableRow, txId, commitTableId, TEST_PARTITION_ID);
+
+        verify(mvPartitionStorage, times(1)).addWrite(eq(rowId), eq(tableRow), eq(txId), eq(commitTableId), eq(TEST_PARTITION_ID));
```
rpuch, line 154: Just a comment: `times(1)` is redundant, if we omit it, it will have the same effect. But it can be used to highlight to the reader that you are expecting exactly 1 call (not more), this can be useful sometimes.

tkalkirill: Thanks for the tip =), I'll leave it as it is.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccessImpl.java
tkalkirill: You're right, I'll fix it.

rpuch (approved):

PR #1579, opened 2023-01-25
IGNITE-18446 Add busy lock to raft service

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-18446

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
@@ -49,6 +49,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
@@ -906,6 +907,30 @@ public class IgniteUtils {
         }
 
         return result;
+    }
+
+    /**
+     * Cancels the future and runs a consumer on future's result if it was completed before the cancellation.
+     * Does nothing if future is cancelled or completed exceptionally.
+     *
+     * @param future Future.
+     * @param consumer Consumer that accepts future's result.
+     * @param <T> Future's result type.
+     */
+    public static <T> void cancelOrConsume(CompletableFuture<T> future, Consumer<T> consumer) {
+        future.cancel(true);
+
+        if (future.isCancelled() || future.isCompletedExceptionally()) {
+            return;
+        }
+
+        assert future.isDone();
+
+        T res = future.join();
+
+        assert res != null;
+
+        consumer.accept(res);
     }
 
     /**
--- a/modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
+++ b/modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
@@ -1224,6 +1224,8 @@ public class RocksDbKeyValueStorage implements KeyValueStorage {
         long minWatchRevision = watchProcessor.minWatchRevision().orElse(-1);
 
         if (minWatchRevision == -1 || minWatchRevision > upperRevision) {
+            // No events to replay, we can start processing more recent events from the event queue.
+            startWatchExecutor();
             return;
         }
 
--- a/modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
+++ b/modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
@@ -58,6 +58,7 @@ import org.apache.ignite.internal.raft.service.LeaderWithTerm;
 import org.apache.ignite.internal.raft.service.RaftGroupService;
 import org.apache.ignite.internal.replicator.ReplicationGroupId;
 import org.apache.ignite.internal.tostring.S;
+import org.apache.ignite.internal.util.IgniteSpinBusyLock;
 import org.apache.ignite.lang.IgniteException;
 import org.apache.ignite.network.ClusterNode;
 import org.apache.ignite.network.ClusterService;
@@ -107,6 +108,9 @@ public class RaftGroupServiceImpl implements RaftGroupService {
 
     /** Executor for scheduling retries of {@link RaftGroupServiceImpl#sendWithRetry} invocations. */
     private final ScheduledExecutorService executor;
+
+    /** Busy lock. */
+    private final IgniteSpinBusyLock busyLock = new IgniteSpinBusyLock();
 
     /**
      * Constructor.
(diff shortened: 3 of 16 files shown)


modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -908,6 +909,32 @@ public static <T> List<T> collectStaticFields(Class<?> sourceCls, Class<? extend
...
+    /**
+     * Cancels the future and runs a consumer on future's result if it was completed before the cancellation.
+     * Does nothing if future is cancelled or completed exceptionally.
+     *
+     * @param future Future.
+     * @param consumer Consumer that accepts future's result.
+     * @param <T> Future's result type.
+     */
+    public static <T> void cancelOrConsume(CompletableFuture<T> future, Consumer<T> consumer) {
+        assert future != null;
```
sashapolo, line 921: Why do you need this assertion?

modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
```diff
@@ -679,6 +694,14 @@ private Peer randomNode(@Nullable Peer excludedPeer) {
 
         int lastPeerIndex = excludedPeer == null ? -1 : peers0.indexOf(excludedPeer);
 
+        if (peers0.size() == 1) {
+            if (lastPeerIndex != -1) {
```
sashapolo, line 698: Why do we even need to check this if the peer list has a single element?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -217,20 +217,24 @@ private Replica startReplicaInternal(
...
+                    return CompletableFuture.completedFuture(newReplica);
+                }
+
+                replicaFut.complete(newReplica);
 
                 return replicaFut;
             }
         });
 
         // replicaFut is always completed here.
```
sashapolo, line 236: This comment is no longer relevant

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -217,20 +217,24 @@ private Replica startReplicaInternal(
...
+
         replicas.compute(replicaGrpId, (replicationGroupId, replicaFut) -> {
             if (replicaFut == null) {
-                replicaFut = CompletableFuture.completedFuture(new Replica(replicaGrpId, listener));
-
-                return replicaFut;
+                return CompletableFuture.completedFuture(newReplica);
             } else {
-                replicaFut.complete(new Replica(replicaGrpId, listener));
+                if (replicaFut.isDone() && !replicaFut.isCancelled() && !replicaFut.isCompletedExceptionally()) {
```
sashapolo, line 226: When can this happen? Are you sure we can simply ignore the previous replica value?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
```diff
@@ -1224,6 +1224,8 @@ private void replayUpdates(long upperRevision) {
         long minWatchRevision = watchProcessor.minWatchRevision().orElse(-1);
 
         if (minWatchRevision == -1 || minWatchRevision > upperRevision) {
+            // No events to replay, we can start processing more recent events from the event queue.
+            startWatchExecutor();
```
sashapolo, line 1228: Can you please add a test for this in the `AbstractKeyValueStorageTest`?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
SammyVimes: ok, how about `@Contract("null -> fail")`?

modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
SammyVimes: Well it is kind of a Fast Path, as we already have `indexOf`, so we don't need to call method, right? But yes, it looks too bizarre, I removed it

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
sashapolo: why?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
SammyVimes: It can happen on rebalance, then we start a new replica while we still have an old replica. It looks like we can, because it doesn't have a state (also, note that in `stopReplica` we just remove the object from the map)

sashapolo: But how can we have an old replica in the map? Why wasn't `stopReplica` called?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -217,20 +217,24 @@ private Replica startReplicaInternal(
...
+                    return CompletableFuture.completedFuture(newReplica);
+                }
+
+                replicaFut.complete(newReplica);
```
AMashenkov, line 230: Here we can complete the future inside the synchronized block.
Should the future be completed in the current thread? or completeAsync should be used instead?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
SammyVimes: So that idea would highlight that null is not an option here

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
SammyVimes: Because the node was either in pending or in stable assignments

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
SammyVimes: I'm not sure this should be addressed in current PR, probably we should cleanup CompletableFuture usage in a single ticket (or multiple separate tickets)

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
sashapolo: It should highlight that anyway, because this parameter is not marked as `@Nullable`

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
sashapolo: Can you please describe this code path?

sashapolo (approved):

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
SammyVimes: Sure: if we are doing rebalancing and `STABLE_ASSIGNMENTS_*` key changed in metastorage, then we need to decide if replica must be stopped. We will stop it only if the node is not in stableAssignments and not in pendingAssignments. We do start replica on PENDING_ASSIGNMENTS_* and in `updateAssignmentInternal()`, so the replica may be not stopped.

PR #1581, opened 2023-01-25
IGNITE-18521 Sql. Query hangs if exception is thrown in mapping phase.

xtern:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -576,36 +572,48 @@ private void submitFragment(String initiatorNode, String fragmentString, Fragmen
...
 
-                        for (String nodeName : fragmentDesc.nodeNames()) {
-                            sendFragment(nodeName, fragment, fragmentDesc);
+                            break;
                         }
                     }
-                } catch (Throwable e) {
-                    root.thenAccept(root -> root.onError(e));
+                } catch (Throwable t) {
+                    LOG.error("An error occurred while executing the query.", t);
```
korlov42, line 614: we should not use ERROR severity here. PLease take a look at [this](https://cwiki.apache.org/confluence/display/IGNITE/Java+Logging+Guidelines#JavaLoggingGuidelines-ERROR) guidelines

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -576,36 +572,48 @@ private void submitFragment(String initiatorNode, String fragmentString, Fragmen
...
+
+                    root.completeExceptionally(t);
```
korlov42, line 616: what if `root` has been already completed at this point?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
xtern: thanks, additional logging removed

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
xtern: Fixed a bit to handle such scenario :thinking:

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImplTest.java
```diff
@@ -607,7 +633,8 @@ protected TestTable createTable(String name, int size, IgniteDistribution distr,
                 new TableDescriptorImpl(columns, distr),
                 name,
                 ColocationGroup.forNodes(nodeNames),
-                size
+                size,
+                () -> mappingException
```
AMashenkov, line 637:
```suggestion
                size
```
Let's just override the `colocationGroup` method instead, and add a short explanation on why we throws an exception from here.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestTable.java
```diff
@@ -234,6 +239,12 @@ public boolean rolledUpColumnValidInsideAgg(
     /** {@inheritDoc} */
     @Override
     public ColocationGroup colocationGroup(MappingQueryContext ctx) {
+        RuntimeException ex = mappingExceptionSupplier.get();
+
+        if (ex != null) {
+            throw ex;
+        }
+
```
AMashenkov, line 247:
```suggestion
```
This looks like a dirty hack.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImplTest.java
xtern: Thanks, I've fixed it as per your suggestion.

AMashenkov (approved):

PR #1585, opened 2023-01-26
IGNITE-18638 Make tests print thread dump when timing out

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18638

### DIFF
--- a/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/junit/DumpThreadsOnTimeout.java
+++ b/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/junit/DumpThreadsOnTimeout.java
@@ -51,6 +51,7 @@ public class DumpThreadsOnTimeout implements TestExecutionExceptionHandler, Life
 
     private static void handleThrowable(Throwable throwable, String category) throws Throwable {
         if (isJunitMethodTimeout(throwable)) {
+            // We take throwable.getMessage() because TimeoutExtension puts a description of 'what has timed out' to it.
             System.out.println(category + " " + throwable.getMessage() + ", dumping threads");
 
             dumpThreadsToStdout();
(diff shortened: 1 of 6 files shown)


modules/core/src/testFixtures/resources/META-INF/services/org.junit.jupiter.api.extension.Extension
```diff
@@ -0,0 +1 @@
+org.apache.ignite.internal.testframework.junit.DumpThreadsOnTimeout
```
sashapolo, line 1: Can we use auto-service instead?

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/junit/DumpThreadsOnTimeout.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+public class DumpThreadsOnTimeout implements TestExecutionExceptionHandler, LifecycleMethodExecutionExceptionHandler {
+    @Override
+    public void handleTestExecutionException(ExtensionContext context, Throwable throwable) throws Throwable {
+        handleThrowable(throwable, "Test method");
+    }
+
+    private static void handleThrowable(Throwable throwable, String category) throws Throwable {
+        if (isJunitMethodTimeout(throwable)) {
+            System.out.println(category + " " + throwable.getMessage() + ", dumping threads");
```
sashapolo, line 51: Can you provide an example how this message will look like?

rpuch: https://ci.ignite.apache.org/buildConfiguration/ApacheIgnite3xGradle_Test_RunAllTests/7032302?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true&expandBuildProblemsSection=true&expandCode+Inspection=true&pluginCoverage=true

It's just after the failure stack trace:

`Test method testIntLocking() timed out after 60 seconds, dumping threads`

modules/core/src/testFixtures/resources/META-INF/services/org.junit.jupiter.api.extension.Extension
rpuch: Yes we can! Switched to auto-service.

sashapolo (approved):

PR #1586, opened 2023-01-26
IGNITE-18633 Storage cleanup integration if one of them has not finished rebalancing with TableManager

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18633

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/PartitionStoragesHolder.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/PartitionStoragesHolder.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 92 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/PartitionStoragesHolder.java
```diff
@@ -0,0 +1,55 @@
...
+
+package org.apache.ignite.internal.table.distributed.storage;
+
+import org.apache.ignite.internal.storage.MvPartitionStorage;
+import org.apache.ignite.internal.tx.storage.state.TxStateStorage;
+
+/**
+ * Partition storages holder.
+ */
+public class PartitionStoragesHolder {
```
rpuch, line 26: Do we need this `Holder` suffix? `PartitionStorages` seems to adequately denote what it is: a bunch of storages.

tkalkirill: Why not, yes, I will.

rpuch (approved):

PR #1588, opened 2023-01-26
IGNITE-18369: SQL. Incorrect inferred return type for functions for NUMERIC types

lowka:


AMashenkov (approved):
Please fix conflicts.

zstan:
but standard is : 
image

zstan:
COALESCE (V1, V2) is equivalent to the following <case specification>:
CASE WHEN V1
IS NOT NULL THEN
V1 ELSE
V2 END

lowka:
@zstan 

> COALESCE (V1, V2) is equivalent to the following :
> CASE WHEN V1
> IS NOT NULL THEN
> V1 ELSE
> V2 END
Well sure it is but could you please elaborate? Do we need some additional test cases or what?

zstan:
I think this mean that this test need to be passed, wdyt ? 
assertQuery("SELECT COALESCE(?, ?)").withParams("a", 10).returns("a").check();

lowka:
> assertQuery("SELECT COALESCE(?, ?)").withParams("a", 10).returns("a").check();

This test case should fail b/c we do int -> varchar coercion was explicitly disabled.

PR #1590, opened 2023-01-26
IGNITE-18625 Sql. Code cleanup. Unmute tests.

AMashenkov:


modules/runner/src/integrationTest/java/org/apache/ignite/internal/index/ItIndexManagerTest.java
```diff
@@ -71,7 +71,7 @@ public void eventsAreFiredWhenIndexesCreatedAndDropped() {
         CompletableFuture<IndexEventParameters> indexCreatedFuture = registerListener(indexManager, IndexEvent.CREATE);
 
         indexManager.createIndexAsync("PUBLIC", "INAME", "TNAME", true, tableIndexChange ->
-                tableIndexChange.convert(HashIndexChange.class).changeColumnNames("C3", "C2")).join();
+                tableIndexChange.convert(HashIndexChange.class).changeColumnNames("C3", "C2")).get();
```
ygerzhedovich, line 74: you keep potentially infinity lock.
Let's use `await(CompletionStage stage)` instead of any 'join', `get`

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDmlTest.java
```diff
@@ -107,13 +106,7 @@ public void mergeOpChangePrimaryKey() {
...
                 + "WHEN NOT MATCHED THEN INSERT (k1, k2, a, b) VALUES (src.k1, src.k2, src.a, src.b)";
 
-        sql(sql);
-
-        assertQuery("SELECT * FROM test2 ORDER BY k1")
-                .returns(222, 222, 1, 300, null)
-                .returns(111, 333, 0, 100, "")
-                .returns(444, 444, 2, 200, null)
-                .check();
+        assertThrows(CalciteContextException.class, () -> sql(sql));
```
ygerzhedovich, line 109: Let's add the check SQL code here.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/index/ItIndexManagerTest.java
AMashenkov: The main issue here was the join() is not interruptible, but get() is.

ygerzhedovich: Agree, but with `get` we could wait for a test timeout instead of a short wait timeout in the utility method.

zstan: +1 for @AMashenkov approach - we need not guessing each test work time.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDmlTest.java
AMashenkov: We have to throw SqlException to check error code, but CalciteContextException is thrown.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/util/CompositeSubscriptionTest.java
```diff
@@ -157,28 +157,27 @@ private void doPublishSubscribe(
...
-                        assert false;
-                    }
-
-                    @Override
-                    public void onComplete() {
-                        lsnr.onComplete();
-                    }
-                })
-        ).join();
+        compPublisher.subscribe(new Subscriber<>() {
```
ygerzhedovich, line 160: please return back runAsync here, it requeres for the test

AMashenkov: I think all the guaranties are preserved.
Test thread will continue work right after subscribe method will be finished in background thread.

PR #1594, opened 2023-01-27
IGNITE-18291 Enhance C++ Binary Tuple code

ademakov:


modules/platforms/cpp/ignite/client/sql/column_metadata.h
```diff
@@ -18,7 +18,7 @@
 #pragma once
 
 #include "ignite/client/sql/column_origin.h"
-#include "ignite/client/sql/sql_column_type.h"
+#include "ignite/schema/ignite_type.h"
```
isapego, line 21: This is not good. `schema` is not a public library for client, so removal of `sql_column_type` here should be reversed.

modules/platforms/cpp/ignite/schema/types.h
```diff
@@ -25,22 +25,23 @@
...
-static constexpr IntT NO_NUM = -1;
+static constexpr number_t NOT_NUM = -1;
 
 /** Binary value for a potentially nullable column. */
-using element_view = std::optional<bytes_view>;
+using value_view = std::optional<bytes_view>;
 
 /** A set of binary values for a whole or partial row. */
-using tuple_view = std::vector<element_view>;
+using tuple_view = std::vector<value_view>;
```
isapego, line 41: By some reason, this change breaks client tests on Windows where `tuple_view` is used as a variable name.

isapego: Maybe we need to re-name variables in tests.

isapego: Update: the cause of this is leakage of `schema` headers in clients public headers due to changes in `client/sql/column_metadata.h`

isapego (approved):

modules/platforms/cpp/ignite/client/detail/compute/compute_impl.cpp
```diff
@@ -15,9 +15,12 @@
  * limitations under the License.
  */
 
-#include "ignite/client/detail/compute/compute_impl.h"
+#include "compute_impl.h"
```
isapego, line 18: A there a requirement like that? I prefer using full names.

modules/platforms/cpp/ignite/client/detail/sql/sql_impl.cpp
```diff
@@ -15,11 +15,12 @@
...
  */
 
-#include "ignite/client/detail/sql/sql_impl.h"
+#include "sql_impl.h"
+
 #include "ignite/client/detail/sql/result_set_impl.h"
 #include "ignite/client/detail/utils.h"
 
-#include "ignite/schema/binary_tuple_builder.h"
+#include <ignite/tuple/binary_tuple_builder.h>
```
isapego, line 23: Why `<>` instead of `""`?

modules/platforms/cpp/ignite/common/ignite_type.h
```diff
@@ -70,6 +74,7 @@ enum class column_type {
 
     /** Binary data. */
     BYTE_ARRAY = 15,
+    BINARY = BYTE_ARRAY,
```
isapego, line 77: Do you think we need this synonym?

modules/platforms/cpp/tests/client-test/transactions_test.cpp
```diff
@@ -64,7 +64,7 @@ TEST_F(transactions_test, empty_transaction_commit) {
     auto api = m_client.get_transactions();
 
     auto tx = api.begin();
-    tx.rollback();
+    tx.commit();
```
isapego, line 67: Nice catch

modules/platforms/cpp/ignite/client/detail/sql/sql_impl.cpp
ademakov: In this case there is no difference, but I prefer to use the same pattern for all #include directives.

If we have a public header installed e.g. in /usr/local/include/ignite and this header includes another public header, the user can compile a client app with -I/usr/local/include compiler option then it will find all the headers.

So if you include a header that starts with ignite/... I prefer to have it always in angle brackets.

But if you include a header in the same subdir you can use relative paths in quotes and omitting the common prefix.

modules/platforms/cpp/ignite/common/ignite_type.h
isapego: I kind of agree, but this is a common name across all Ignite codebase.

ademakov: Removed

PR #1597, opened 2023-01-30
IGNITE-16785 Add configuration for the timeout for removing nodes from logical topology

rpuch:
https://issues.apache.org/jira/browse/IGNITE-16785

### DIFF
--- a/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
+++ b/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
@@ -610,7 +610,7 @@ public class ClusterManagementGroupManager implements IgniteComponent {
             if (physicalTopologyNode == null || !physicalTopologyNode.id().equals(node.id())) {
                 raftService.removeFromCluster(Set.of(node));
             }
-        }, configuration.logicalTopologyRemovalDelayAfterDisappearance().value(), TimeUnit.MILLISECONDS);
+        }, configuration.failoverTimeout().value(), TimeUnit.MILLISECONDS);
     }
 
     private void sendClusterState(CmgRaftService raftService, ClusterNode node) {
--- a/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
+++ b/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
@@ -36,6 +36,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 import org.apache.ignite.internal.cluster.management.LocalStateStorage.LocalState;
+import org.apache.ignite.internal.cluster.management.configuration.ClusterManagementConfiguration;
 import org.apache.ignite.internal.cluster.management.network.CmgMessageHandlerFactory;
 import org.apache.ignite.internal.cluster.management.network.messages.CancelInitMessage;
 import org.apache.ignite.internal.cluster.management.network.messages.ClusterStateMessage;
@@ -83,9 +84,6 @@ import org.jetbrains.annotations.TestOnly;
  * for the description of the Cluster Management Group and its responsibilities.
  */
 public class ClusterManagementGroupManager implements IgniteComponent {
-    // TODO: timeout should be configurable, see https://issues.apache.org/jira/browse/IGNITE-16785
-    private static final int NETWORK_INVOKE_TIMEOUT = 500;
-
     private static final IgniteLogger LOG = Loggers.forClass(ClusterManagementGroupManager.class);
 
     /** Busy lock to stop synchronously. */
@@ -121,6 +119,10 @@ public class ClusterManagementGroupManager implements IgniteComponent {
 
     private final LogicalTopology logicalTopology;
 
+    private final ClusterManagementConfiguration configuration;
+
+    private long networkInvokeTimeoutMillis;
+
     /** Local state. */
     private final LocalStateStorage localStateStorage;
 
@@ -147,12 +149,15 @@ public class ClusterManagementGroupManager implements IgniteComponent {
             ClusterService clusterService,
             RaftManager raftManager,
             ClusterStateStorage clusterStateStorage,
-            LogicalTopology logicalTopology
+            LogicalTopology logicalTopology,
+            ClusterManagementConfiguration configuration
     ) {
         this.clusterService = clusterService;
         this.raftManager = raftManager;
         this.clusterStateStorage = clusterStateStorage;
         this.logicalTopology = logicalTopology;
+        this.configuration = configuration;
+
         this.localStateStorage = new LocalStateStorage(vault);
         this.clusterInitializer = new ClusterInitializer(clusterService);
     }
@@ -188,6 +193,8 @@ public class ClusterManagementGroupManager implements IgniteComponent {
 
     @Override
     public void start() {
+        networkInvokeTimeoutMillis = configuration.networkInvokeTimeout().value();
+
         synchronized (raftServiceLock) {
             raftService = recoverLocalState();
         }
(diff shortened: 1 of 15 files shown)


modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -188,6 +193,8 @@ public void initCluster(
 
     @Override
     public void start() {
+        networkInvokeTimeoutMillis = configuration.networkInvokeTimeout().value();
```
ibessonov, line 196: Do we want it to be dynamically-updated? If we do, then we should read it from the configuration every time

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -597,14 +604,13 @@ public void onDisappeared(ClusterNode member) {
...
     private void scheduleRemoveFromLogicalTopology(CmgRaftService raftService, ClusterNode node) {
-        // TODO: delay should be configurable, see https://issues.apache.org/jira/browse/IGNITE-16785
         scheduledExecutor.schedule(() -> {
             ClusterNode physicalTopologyNode = clusterService.topologyService().getByConsistentId(node.name());
 
             if (physicalTopologyNode == null || !physicalTopologyNode.id().equals(node.id())) {
                 raftService.removeFromCluster(Set.of(node));
             }
-        }, 0, TimeUnit.MILLISECONDS);
+        }, configuration.logicalTopologyRemovalDelayAfterDisappearance().value(), TimeUnit.MILLISECONDS);
```
ibessonov, line 613: Configuration names this long might now be user-friendly. Some day we should revisit all names. Also, should this be a top-level configuration property, or maybe nested inside of some sub-configuration?

Anyway, name is too specific, user should not know what a disappearance. In Ignite 2 we called it a "failover timeout" if I remember correctly. We can discuss other options as well

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/configuration/ClusterManagementConfigurationModule.java
```diff
@@ -0,0 +1,41 @@
...
+@AutoService(ConfigurationModule.class)
+public class ClusterManagementConfigurationModule implements ConfigurationModule {
+    @Override
+    public ConfigurationType type() {
+        return ConfigurationType.LOCAL;
+    }
+
+    @Override
+    public Collection<RootKey<?, ?>> rootKeys() {
+        return List.of(ClusterManagementConfiguration.KEY);
```
ibessonov, line 39: So, cluster management configuration is a local configuration. Is that expected, or is it a mistake?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/configuration/ClusterManagementConfigurationSchema.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.ignite.configuration.annotation.Value;
+
+/**
+ * Cluster management configuration schema.
+ */
+@ConfigurationRoot(rootName = "cluster", type = ConfigurationType.LOCAL)
+public class ClusterManagementConfigurationSchema {
+    /** Invoke timeout used by Cluster Management module (ms). */
+    @Value(hasDefault = true)
+    public long networkInvokeTimeout = 500;
```
ibessonov, line 32: Please ask `@Range(...)` with adequate values for validation

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/configuration/ClusterManagementConfigurationSchema.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Delay between a moment a node drops out from the physical topology and when it gets removed from the logical topology (ms).
+     */
+    @Value(hasDefault = true)
+    public long logicalTopologyRemovalDelayAfterDisappearance = 0;
```
ibessonov, line 38: So, we remove node immediately. There should be a TODO here I guess, this value must be changed to something better in the nearest future

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -332,7 +333,8 @@ public class IgniteImpl implements Ignite {
                 clusterSvc,
                 raftMgr,
                 clusterStateStorage,
-                logicalTopology
+                logicalTopology,
+                nodeCfgMgr.configurationRegistry().getConfiguration(ClusterManagementConfiguration.KEY)
```
ibessonov, line 337: Don't we have a variable that's equal to `nodeCfgMgr.configurationRegistry()`? Please check

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/configuration/ClusterManagementConfigurationModule.java
rpuch: Distributed configuration requires MetaStorage, which requires CMG, so it seems that CMG cannot have a distributed configuration, it has to be local

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
rpuch: We don't, but we have

```
public ConfigurationRegistry nodeConfiguration() {
    return nodeCfgMgr.configurationRegistry();
}
```

in this class. At the same time, it seems awkward to use a public method in our own constructor, this seems a bit fragile.

So I just extracted 2 local variables in the constructor (one for node config registry, another for cluster config registry).

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
rpuch: We probably want it to be updated dynamically, switched to this style

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
rpuch: For now we just have 3 properties there, all of them are timeouts, one is for invoke timeout, other 2 are about CMG logic. It seems that it's too early to try to establish any substructure, at least it would be too hazy.

I agree, the name is too long. I failed to invent anything both meaningful and short, so I just adopt your suggestion about `failoverTimeout`.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/configuration/ClusterManagementConfigurationSchema.java
rpuch: Added min restriction; max restrictions appear to be too restricting in all 3 cases (as it seems to be difficult to determine the frontier between 'reasonable' and 'too much')

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
ibessonov: Thank you! I still hope that one day we will revisit all configuration names. And I mean all of them

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/configuration/ClusterManagementConfigurationModule.java
ibessonov: Ok, so maybe this should be a local network setting, without introducing specific CMG schema. We can ask more people on their opinions

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
ibessonov: Ok, thank you! I really thought that we should have a var like this

ibessonov (approved):

PR #1598, opened 2023-01-30
IGNITE-17992 .NET: Fix colocation hash for LocalTime, LocalDateTime, Timestamp with custom precision

ptupitsyn:
The problem: when temporal type (Time, DateTime, Timestamp) precision is less than `MAX_TIME_PRECISION`, column value is adjusted (truncated) by the server on insert, resulting in a different colocation hash. For example, nanosecond component is dropped when precision is 0.

* Propagate `Column.Precision` to .NET client (already sent by the server).
* Pass precision to `BinaryTupleBuilder` methods, and apply the same `NormalizeNanos` logic that `RowAssembler` uses.


modules/schema/src/main/java/org/apache/ignite/internal/schema/row/TemporalTypesHelper.java
```diff
@@ -272,13 +272,13 @@ private static int truncateTo(int precision, int nanos) {
...
-                return (nanos / 100_000) * 100; // 100mcs precision.
+                return (nanos / 100_000) * 100; // 100μs precision.
             }
             case 5: {
-                return (nanos / 10_000) * 10; // 10mcs precision.
+                return (nanos / 10_000) * 10; // 10μs precision.
             }
             case 6: {
-                return nanos / 1_000; // 1mcs precision.
+                return nanos / 1_000; // 1μs precision.
```
isapego, line 281: Unicode in sources? Hmm, I just use `us`.

isapego (approved):

PR #1599, opened 2023-01-30
IGNITE-17958 PartitionListener should not close partition storage

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-17958

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
@@ -2128,7 +2128,7 @@ public class TableManager extends Producer<TableEvent, TableEventParameters> imp
                 new TablePartitionId(table.tableId(), partitionId)
         );
 
-        return destroyStoragesFuture == null ? completedFuture(null) : destroyStoragesFuture
+        return (destroyStoragesFuture == null ? completedFuture(null) : destroyStoragesFuture)
                 .thenComposeAsync(unused -> {
                     MvPartitionStorage mvPartitionStorage = table.internalTable().storage().getOrCreateMvPartition(partitionId);
                     TxStateStorage txStateStorage = table.internalTable().txStateStorage().getOrCreateTxStateStorage(partitionId);
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
@@ -143,7 +143,6 @@ import org.apache.ignite.internal.tx.storage.state.TxStateStorage;
 import org.apache.ignite.internal.tx.storage.state.TxStateTableStorage;
 import org.apache.ignite.internal.tx.storage.state.rocksdb.TxStateRocksDbTableStorage;
 import org.apache.ignite.internal.util.ByteUtils;
-import org.apache.ignite.internal.util.ExceptionUtils;
 import org.apache.ignite.internal.util.IgniteNameUtils;
 import org.apache.ignite.internal.util.IgniteSpinBusyLock;
 import org.apache.ignite.internal.util.IgniteUtils;
@@ -968,6 +967,14 @@ public class TableManager extends Producer<TableEvent, TableEventParameters> imp
 
         tablesToStopInCaseOfError.clear();
 
+        CompletableFuture<?>[] destroyPartitionFutures = destroyFutureByTablePartitionId.values().stream()
+                .filter(not(CompletableFuture::isDone))
+                .toArray(CompletableFuture[]::new);
+
+        allOf(destroyPartitionFutures).join();
+
+        destroyFutureByTablePartitionId.clear();
+
         shutdownAndAwaitTermination(rebalanceScheduler, 10, TimeUnit.SECONDS);
         shutdownAndAwaitTermination(ioExecutor, 10, TimeUnit.SECONDS);
         shutdownAndAwaitTermination(txStateStoragePool, 10, TimeUnit.SECONDS);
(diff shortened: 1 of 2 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1952,50 +1952,56 @@ public void onError(Throwable e) {
...
 
                     if (shouldStopLocalServices) {
+                        TableImpl table = tablesByIdVv.latest().get(tableId);
+
+                        CompletableFuture<Void> destroyMvPartitionStorageFuture
+                                = table.internalTable().storage().destroyPartition(partitionId);
+
+                        table.internalTable().txStateStorage().destroyTxStateStorage(partitionId);
+
+                        // Necessary evil.
```
sanpwc, line 1993: Are you sure that it's necessary?
Why do we need to await the destruction?
If it's only about awaiting on component stop, it's possible to await allOf storage destruction futures before components stop itself.

tkalkirill: Methods `TxStateTableStorage#destroyTxStateStorage` and `MvTableStorage#destroyPartition` close the storages and clean up the data.
Thus, if we close the storages, then in order to clear them, we will need to create them again, which may not look good.
But in fact, deleting storages is done in almost O(1) at the moment.
I can describe it in the comments so that there are fewer questions.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1952,50 +1952,56 @@ public void onError(Throwable e) {
...
                     String localMemberName = clusterService.topologyService().localMember().name();
 
                     boolean shouldStopLocalServices = Stream.concat(stableAssignments.stream(), pendingAssignments.stream())
                             .noneMatch(assignment -> assignment.consistentId().equals(localMemberName));
 
                     if (shouldStopLocalServices) {
+                        TableImpl table = tablesByIdVv.latest().get(tableId);
+
+                        CompletableFuture<Void> destroyMvPartitionStorageFuture
+                                = table.internalTable().storage().destroyPartition(partitionId);
```
rpuch, line 1989: I wonder why destruction is made under this Jira ticket: it's about closing the storage, not destruction. I thought we had another ticket about destruction of data of evicted partitions.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -416,12 +416,6 @@ public boolean onSnapshotLoad(Path path) {
 
     @Override
     public void onShutdown() {
-        // TODO: IGNITE-17958 - probably, we should not close the storage here as PartitionListener did not create the storage.
-        try {
-            storage.close();
```
rpuch, line 421: So the storage will be shut down either when the `TableStorage` is shutdown (on node stop), or when the partition is evicted from the node (in which case it will be not just shut down, but also destroyed)?

It would be great to have a test making sure that a non-evicted partition storage is closed on node stop.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: As far as I understand, in the general case, deletion of storages will occur here, since deletion: closing + clearing data, then why not delete them now.

In [IGNITE-18160](https://issues.apache.org/jira/browse/IGNITE-18160,) we also need to add a mechanism to clean up storages when nodes are restarted.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
tkalkirill: Storages will be closed when the `TableStorage` stops.
And are destroyed, on evict of partition or destruction of the table.

Found tests:
- org.apache.ignite.internal.table.distributed.TableManagerTest#endTableManagerStopTest
- org.apache.ignite.internal.table.distributed.TableManagerTest#testDropTable
- I don’t even know if testing is needed to evict a partition, it will be possible to add it during implementation [IGNITE-18160](https://issues.apache.org/jira/browse/IGNITE-18160,)

rpuch (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sanpwc: > Thus, if we close the storages, then in order to clear them, we will need to create them again, which may not look good.

My point is that in case of storage recreation we should chain it with storage.drop(). 
And besides that we should await storage drop on components stop.

tkalkirill: The problem of clearing the storages when the node is restarted will be solved [IGNITE-18160](https://issues.apache.org/jira/browse/IGNITE-18160) when the node is stopped, we do not need to wait for the data to be cleared, it can take a long time.

sanpwc: And what about storage recreation?
What I want to say is that if we want to drop the storage with some id X we can put destroyMvPartitionStorageFuture to the map using X as a key, and if we try to recreate the storage with same id we can chain storage creation with storage drop.

> we do not need to wait for the data to be cleared, it can take a long time.

That's a bit confusing. You've told that it's O(1)

> But in fact, deleting storages is done in almost O(1)

tkalkirill: Discussed personally, decided to do it through an external map with destruction futures.

We will wait for them at the re-creation of storages.
As well as at stopping nodes.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -292,6 +295,9 @@ public class TableManager extends Producer<TableEvent, TableEventParameters> imp
 
     private static final TableMessagesFactory TABLE_MESSAGES_FACTORY = new TableMessagesFactory();
 
+    /** Futures for destroying table partition storages({@link MvPartitionStorage} and {@link TxStateStorage}). */
+    private static final Map<TablePartitionId, CompletableFuture<Void>> destroyFutureByTablePartitionId = new ConcurrentHashMap<>();
```
sanpwc, line 299: Why it's static?

tkalkirill: Typo. fix it =)

sanpwc (approved):

PR #1603, opened 2023-01-30
IGNITE-18569 Fix planning with MRSortAggregate.

zstan:


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItAggregatesTest.java
```diff
@@ -21,149 +21,216 @@
...
     @BeforeAll
     static void initTestData() {
         createAndPopulateTable();
+
+        sql("CREATE TABLE test (id INT PRIMARY KEY, grp0 INT, grp1 INT, val0 INT, val1 INT) WITH replicas=2,partitions=10");
+        sql("CREATE TABLE test_one_col_idx (pk INT PRIMARY KEY, col0 INT)");
+
+        // TODO: https://issues.apache.org/jira/browse/IGNITE-17304 uncomment this
+        // sql("CREATE INDEX test_idx ON test(grp0, grp1)");
+        // sql("CREATE INDEX test_one_col_idx_idx ON test_one_col_idx(col0)");
```
AMashenkov, line 57:
```suggestion
        sql("CREATE INDEX test_idx ON test(grp0, grp1)");
        sql("CREATE INDEX test_one_col_idx_idx ON test_one_col_idx(col0)");
        
        // FIXME: https://issues.apache.org/jira/browse/IGNITE-18203
        waitForIndex("test_idx");
        waitForIndex("test_one_col_idx_idx");
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItAggregatesTest.java
```diff
@@ -185,36 +252,42 @@ public void testAnyValAggr() {
...
+            sql("DROP TABLE IF EXISTS t1");
+            sql("DROP TABLE IF EXISTS t2");
         }
-
-        sql("INSERT INTO t2 VALUES (0, 'val0', 'val0', 'val0'), (1, 'val1', 'val1', 'val1')");
-
-        String sql = "SELECT val1, count(val2) FROM t1 GROUP BY val1";
-
-        assertQuery(sql)
-                .matches(QueryChecker.matches(".*Exchange.*Colocated.*Aggregate.*"))
```
AMashenkov, line 205: I guess we checked here some pushdown-aggregate rules are applied,
but now we disable rules and never check any rules are applied.
Should we have both kind of tests?

zstan: I check blame history also in ai-2 and can`t found issue linked with this test, i revert it and seems we need not integration but execution test for such a thing.

AMashenkov (approved):

PR #1604, opened 2023-01-31
IGNITE-18630 Try to deliver a message until receiver drops out from logical topology

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18630

### DIFF
--- a/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
+++ b/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
@@ -88,6 +88,9 @@ import org.jetbrains.annotations.TestOnly;
 public class ClusterManagementGroupManager implements IgniteComponent {
     private static final IgniteLogger LOG = Loggers.forClass(ClusterManagementGroupManager.class);
 
+    /** How much time to have for sending a bye-bye in the form of 'logical topology leave' on stop. */
+    private static final long GRACEFUL_LEAVE_TIMEOUT_MILLIS = 1_000;
+
     /** Busy lock to stop synchronously. */
     private final IgniteSpinBusyLock busyLock = new IgniteSpinBusyLock();
 
@@ -384,15 +387,16 @@ public class ClusterManagementGroupManager implements IgniteComponent {
     private CompletableFuture<CmgRaftService> updateLogicalTopology(CmgRaftService service) {
         return service.logicalTopology()
                 .thenCompose(logicalTopology -> {
-                    Set<String> physicalTopologyIds = clusterService.topologyService().allMembers()
+                    Set<String> discoveredMembersIds = discoveryTopologyService.allDiscoveredMembers()
                             .stream()
                             .map(ClusterNode::id)
                             .collect(toSet());
 
                     Set<ClusterNode> nodesToRemove = logicalTopology.nodes().stream()
-                            .filter(node -> !physicalTopologyIds.contains(node.id()))
+                            .filter(node -> !discoveredMembersIds.contains(node.id()))
                             .collect(toUnmodifiableSet());
 
+                    // TODO: IGNITE-18681 - respect removal timeout.
                     return nodesToRemove.isEmpty() ? completedFuture(null) : service.removeFromCluster(nodesToRemove);
                 })
                 .thenApply(v -> service);
@@ -612,7 +616,7 @@ public class ClusterManagementGroupManager implements IgniteComponent {
 
     private void scheduleRemoveFromLogicalTopology(CmgRaftService raftService, ClusterNode node) {
         scheduledExecutor.schedule(() -> {
-            ClusterNode discoveredNode = discoveryTopologyService.discoveredNodeByConsistentId(node.name());
+            ClusterNode discoveredNode = discoveryTopologyService.discoveredMemberByConsistentId(node.name());
 
             if (discoveredNode == null || !discoveredNode.id().equals(node.id())) {
                 raftService.removeFromCluster(Set.of(node));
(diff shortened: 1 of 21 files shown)


modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeTopologyService.java
```diff
@@ -132,19 +134,45 @@ void updateLocalMetadata(@Nullable NodeMetadata metadata) {
...
      */
     private void fireAppearedEvent(ClusterNode member) {
+        // Appearing in the Discovery Topology is equivalent to appearing in the Physical topology, so we trigger
+        // PT.onAppeared() right away.
+
         for (TopologyEventHandler handler : getEventHandlers()) {
-            handler.onAppeared(member);
+            try {
+                handler.onAppeared(member);
+            } catch (Exception | AssertionError e) {
```
ibessonov, line 143: Wow, why do we specifically catch assertion errors? Why not "Throwable" then?

modules/raft/src/test/java/org/apache/ignite/raft/jraft/rpc/IgniteRpcTest.java
```diff
@@ -107,6 +110,15 @@ public IgniteRpcTest(TestInfo testInfo) {
 
         service.start();
 
+        // Remove a node from Physical Topology as soon as it gets lost by SWIM.
+        InternalTopologyService internalTopologyService = (InternalTopologyService) service.topologyService();
```
ibessonov, line 114: This is interesting, why did you have to explicitly add this to a test? Is there a way to have this listener "by default" when starting all necessary components?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/AbstractClusterIntegrationTest.java
```diff
@@ -71,7 +75,8 @@ public abstract class AbstractClusterIntegrationTest extends BaseIgniteAbstractT
             + "        gossipInterval: 10\n"
             + "      },\n"
             + "    }\n"
-            + "  }\n"
+            + "  },"
+            + "  cluster.failoverTimeout: 0\n"
```
ibessonov, line 79: Looks out of place, to be honest. Maybe it should be a part of network configuration, let's consider that a possibility

modules/runner/src/integrationTest/java/org/apache/ignite/internal/BootstrapConfigTemplateMethod.java
```diff
@@ -0,0 +1,36 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Defines name of a static method defined in a test class that will be invoked to produce a bootstrap config
+ * template to use intead of a default template.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD})
+public @interface BootstrapConfigTemplateMethod {
```
ibessonov, line 31: I wonder if this annotation and the abstract test should be in test fixtures? Just a thought, you don't need to move them

modules/runner/src/integrationTest/java/org/apache/ignite/internal/compute/ItLogicalTopologyTest.java
```diff
@@ -170,6 +175,96 @@ public void onDisappeared(ClusterNode disappearedNode, LogicalTopologySnapshot n
         assertTrue(secondIgniteDisappeared.await(10, TimeUnit.SECONDS), "Did not see second node leaving in time");
     }
 
+    @Test
+    @BootstrapConfigTemplateMethod("templateWithVeryLongDelayToRemoveFromLogicalTopology")
```
ibessonov, line 179: Is there a shorter name for "DelayToRemoveFromLogicalTopology"? I bet there is

modules/runner/src/integrationTest/java/org/apache/ignite/internal/compute/ItLogicalTopologyTest.java
```diff
@@ -170,6 +175,96 @@ public void onDisappeared(ClusterNode disappearedNode, LogicalTopologySnapshot n
...
+        IgniteImpl entryNode = node(0);
+
+        startNode(1);
+
+        entryNode.logicalTopologyService().addEventListener(listener);
+
+        // Knock the node out without allowing it to say good bye.
+        cluster.knockOutNode(1, NodeKnockout.PARTITION_NETWORK);
+
+        assertFalse(waitForCondition(() -> !events.isEmpty(), 3_000));
```
ibessonov, line 190: Can we somehow make it faster than 3 seconds?
Please add a comment why this value is chosen, it must depend on other settings, right?

SammyVimes (changes requested):

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -664,6 +679,26 @@ private void sendWithRetry(ClusterNode node, NetworkMessage msg, CompletableFutu
...
+
+    private void tryToLeaveLogicalTopologyGracefully(CompletableFuture<CmgRaftService> serviceFuture) {
+        serviceFuture.thenCompose(service -> service.removeFromCluster(Set.of(clusterService.topologyService().localMember())))
+                .orTimeout(GRACEFUL_LEAVE_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS)
+                .exceptionally(ex -> {
+                    LOG.info("Was not able to send a bye-bye in time", ex);
+
+                    return null;
+                })
+                .join();
```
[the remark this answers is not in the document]
rpuch: We probably want to give the 'bye-bye' message a chance to be delivered. If we don't wait, we could cut ourselves from RAFT services (or even network), which would prevent it to be delivered. We'll wait for at most 1 second, this does not seem too bad.

WDYT?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/AbstractClusterIntegrationTest.java
rpuch: Well, it seems to be about CMG, not about the network. You already suggested that we revisit all configuration properties, we could handle this property there as well.

rpuch:
Closed as irrelevant

PR #1605, opened 2023-01-31
IGNITE-18523 Destroy RocksDbTableStorage on destroy()

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18523

### DIFF
--- a/modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
+++ b/modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
@@ -544,37 +544,30 @@ public abstract class AbstractMvTableStorageTest extends BaseMvStoragesTest {
     }
 
     @Test
-    public void testDestroyTableStorage() throws Exception {
+    public void testDestroyTableStorage() {
         MvPartitionStorage mvPartitionStorage = tableStorage.getOrCreateMvPartition(PARTITION_ID);
         HashIndexStorage hashIndexStorage = tableStorage.getOrCreateHashIndex(PARTITION_ID, hashIdx.id());
         SortedIndexStorage sortedIndexStorage = tableStorage.getOrCreateSortedIndex(PARTITION_ID, sortedIdx.id());
 
-        RowId rowId = new RowId(PARTITION_ID);
+        List<IgniteTuple3<RowId, TableRow, HybridTimestamp>> rows = List.of(
+                new IgniteTuple3<>(new RowId(PARTITION_ID), tableRow(new TestKey(0, "0"), new TestValue(0, "0")), clock.now()),
+                new IgniteTuple3<>(new RowId(PARTITION_ID), tableRow(new TestKey(1, "1"), new TestValue(1, "1")), clock.now())
+        );
 
-        TableRow tableRow = tableRow(new TestKey(0, "0"), new TestValue(1, "1"));
+        fillStorages(mvPartitionStorage, hashIndexStorage, sortedIndexStorage, rows);
 
-        IndexRow hashIndexRow = indexRow(hashIndexStorage.indexDescriptor(), tableRow, rowId);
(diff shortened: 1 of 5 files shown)


modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
```diff
@@ -544,37 +544,30 @@ public void testStartRebalanceForClosedPartition() {
...
-        IndexRow hashIndexRow = indexRow(hashIndexStorage.indexDescriptor(), tableRow, rowId);
-        IndexRow sortedIndexRow = indexRow(sortedIndexStorage.indexDescriptor(), tableRow, rowId);
-
-        mvPartitionStorage.runConsistently(() -> {
-            mvPartitionStorage.addWriteCommitted(rowId, tableRow, clock.now());
-
-            hashIndexStorage.put(hashIndexRow);
-
-            sortedIndexStorage.put(sortedIndexRow);
+        List<IgniteTuple3<RowId, TableRow, HybridTimestamp>> rows = List.of(
```
ibessonov, line 552: Why does it have to be a list of tuples? Looks ugly.
Why a single entry wasn't enough?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/HashIndex.java
```diff
@@ -83,4 +83,12 @@ void destroy(int partitionId, WriteBatch writeBatch) throws RocksDBException {
     @Nullable RocksDbHashIndexStorage get(int partitionId) {
         return storages.get(partitionId);
     }
+
+
```
ibessonov, line 87:
```suggestion

```

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/HashIndex.java
```diff
@@ -83,4 +83,12 @@ void destroy(int partitionId, WriteBatch writeBatch) throws RocksDBException {
...
+    /**
+     * Closes all index storages.
+     */
+    void close() {
+        storages.values().forEach(RocksDbHashIndexStorage::close);
```
ibessonov, line 92: Why don't you use "closeAll"?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/SortedIndex.java
```diff
@@ -90,6 +90,8 @@ void destroy(int partitionId, WriteBatch writeBatch) throws RocksDBException {
 
     @Override
     public void close() {
+        storages.values().forEach(RocksDbSortedIndexStorage::close);
+
         indexCf.handle().close();
```
ibessonov, line 95: Why don't you use "closeAll" here as well?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/HashIndex.java
tkalkirill: Fix it

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/SortedIndex.java
tkalkirill: Fix it.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
tkalkirill: These are tests, they(`IgniteTuple3`) are already used in several tests, let's not touch them?

ibessonov (approved):

PR #1615, opened 2023-02-01
IGNITE-18160 Delete physical data after a Raft Group is stopped

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18160


modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
```diff
@@ -391,6 +435,85 @@ void testRebalanceRetryWhenCatchupFailed() throws Exception {
         assertEquals(3, getPartitionClusterNodes(2, 0).size());
     }
 
+    @Test
+    @Tag(TEST_TX_STORAGE_TAG)
```
rpuch, line 439: >Tags are used to filter which tests are executed for a given test plan

But here, they are used to signal that some customizations need to be applied in the `before()` method. This looks like an abuse. It seems that a custom annotation should be added instead.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2102,4 +2055,60 @@ private CompletableFuture<PartitionStorages> getOrCreatePartitionStorages(TableI
...
+                    // no-op
+                }
+
+                InternalTable internalTable = tablesByIdVv.latest().get(tableId).internalTable();
+
+                // Should be done fairly quickly.
+                allOf(
+                        internalTable.storage().destroyPartition(partitionId),
+                        runAsync(() -> internalTable.txStateStorage().destroyTxStateStorage(partitionId), ioExecutor)
+                ).join();
```
rpuch, line 2110: It looks like we only destroy partition data, but the raft group data (log and metastorage) remain untouched. The Jira issue mentions that we need to destroy the log as well.

I think we can move the log and RAFT metastorage destruction to another issue, but we need a TODO mentioning it.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
```diff
@@ -642,20 +815,156 @@ void stop() throws Exception {
...
+    }
+
+    private static TableDefinition createTableDefinition(String tableName) {
+        return SchemaBuilders.tableBuilder("PUBLIC", tableName).columns(
+                SchemaBuilders.column("key", ColumnType.INT64).build(),
+                SchemaBuilders.column("val", ColumnType.INT32).asNullable(true).build()
+        ).withPrimaryKey("key").build();
+    }
+
+    private void createTableForOnePartition(String tableName, int replicas, boolean testDataStorage) {
```
rpuch, line 865:
```suggestion
    private void createTableWithOnePartition(String tableName, int replicas, boolean testDataStorage) {
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
```diff
@@ -642,20 +815,156 @@ void stop() throws Exception {
...
+
+    private static @Nullable InternalTable getInternalTable(Node node, String tableName) {
+        Table table = node.tableManager.table(tableName);
+
+        assertNotNull(table, tableName);
+
+        return ((TableImpl) table).internalTable();
+    }
+
+    private static void checkInvokeDestroyPartitionStorages(Node node, String tableName, int partitionId) {
```
rpuch, line 920:
```suggestion
    private static void checkInvokeDestroyedPartitionStorages(Node node, String tableName, int partitionId) {
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
```diff
@@ -642,20 +815,156 @@ void stop() throws Exception {
...
+        InternalTable internalTable = getInternalTable(node, tableName);
+
+        verify(internalTable.storage(), times(1)).destroyPartition(eq(partitionId));
+        verify(internalTable.txStateStorage(), times(1)).destroyTxStateStorage(eq(partitionId));
```
rpuch, line 924:
```suggestion
        verify(internalTable.storage(), times(1)).destroyPartition(partitionId);
        verify(internalTable.txStateStorage(), times(1)).destroyTxStateStorage(partitionId);
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
```diff
@@ -642,20 +815,156 @@ void stop() throws Exception {
...
+    }
+
+    private static void throwExceptionOnInvokeDestroyPartitionStorages(Node node, String tableName, int partitionId) {
+        InternalTable internalTable = getInternalTable(node, tableName);
+
+        doAnswer(answer -> CompletableFuture.failedFuture(new StorageException("From test")))
+                .when(internalTable.storage())
+                .destroyPartition(eq(partitionId));
```
rpuch, line 932:
```suggestion
                .destroyPartition(partitionId);
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
```diff
@@ -642,20 +815,156 @@ void stop() throws Exception {
...
+
+        doAnswer(answer -> CompletableFuture.failedFuture(new IgniteInternalException("From test")))
+                .when(internalTable.txStateStorage())
+                .destroyTxStateStorage(eq(partitionId));
```
rpuch, line 936:
```suggestion
                .destroyTxStateStorage(partitionId);
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
tkalkirill: Fix it

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
tkalkirill: Fix it

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
tkalkirill: Fix it

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
tkalkirill: Fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: Added TODO

rpuch (approved):

PR #1616, opened 2023-02-01
IGNITE-18392 Add SQL ResultSet object mapping

ptupitsyn:
Add generic `<T> ResultSet<T> execute` API, which maps SQL result set data to user type.

Before that, we could insert and retrieve data in a strongly typed way with object mapping via `RecordView<T>` API, but SQL results could only be accessed in a weakly typed way through `Tuple`. This change fixes the asymmetry.


isapego (approved):

modules/api/src/main/java/org/apache/ignite/sql/Session.java
```diff
@@ -68,11 +69,57 @@ default ResultSet execute(@Nullable Transaction transaction, String query, @Null
...
+    /**
+     * Executes single SQL statement and maps results to objects with the provided mapper.
+     *
+     * @param transaction Transaction to execute the statement within or {@code null}.
+     * @param mapper Mapper.
+     * @param query SQL query template.
+     * @param arguments Arguments for the statement.
+     * @return SQL query results set.
+     */
+    default <T> ResultSet<T> execute(
```
korlov42, line 91: and again, let's add description for type parameter here and every where in public API.

modules/api/src/main/java/org/apache/ignite/sql/Session.java
```diff
@@ -68,11 +69,57 @@ default ResultSet execute(@Nullable Transaction transaction, String query, @Null
...
+        } catch (CompletionException e) {
+            throw IgniteException.wrap(e);
+        }
+    }
+
+    /**
+     * Executes single SQL statement and maps results to objects with the provided mapper.
+     *
+     * @param transaction Transaction to execute the statement within or {@code null}.
+     * @param mapper Mapper.
```
korlov42, line 86: I think we need to provide better description for `mapper` field. This is a public API, the user must clearly understand what to pass and how it will affect the result

modules/api/src/main/java/org/apache/ignite/sql/Session.java
```diff
@@ -98,7 +145,42 @@ default ResultSet execute(@Nullable Transaction transaction, Statement statement
...
+
+    /**
+     * Executes SQL statement in an asynchronous way and maps results to objects with the provided mapper.
+     *
+     * @param transaction Transaction to execute the statement within or {@code null}.
+     * @param mapper Mapper.
+     * @param query SQL query template.
+     * @param arguments Arguments for the statement.
+     * @return Operation future.
+     * @throws SqlException If failed.
```
korlov42, line 161: this is probably copy-pasted javadoc but anyways... async method won't throw any exception, but complete the future exceptionally

modules/api/src/main/java/org/apache/ignite/sql/ResultSet.java
```diff
@@ -32,7 +32,7 @@
  * <p>Note: one and only one of following is possible: {@link #hasRowSet()} returns {@code true}, or {@link #wasApplied()} returns
  * {@code true}, or {@link #affectedRows()} return zero or higher value.
  */
-public interface ResultSet extends Iterator<SqlRow>, AutoCloseable {
+public interface ResultSet<T> extends Iterator<T>, AutoCloseable {
```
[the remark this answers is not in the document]
ptupitsyn: Done

modules/api/src/main/java/org/apache/ignite/sql/async/AsyncResultSet.java
```diff
@@ -48,7 +47,7 @@
  *
  * @see ResultSet
  */
-public interface AsyncResultSet {
+public interface AsyncResultSet<T> {
```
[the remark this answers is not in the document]
ptupitsyn: Done

modules/api/src/main/java/org/apache/ignite/sql/Session.java
ptupitsyn: Done

modules/api/src/main/java/org/apache/ignite/sql/Session.java
ptupitsyn: Good point, added more details, please check again.

modules/api/src/main/java/org/apache/ignite/sql/Session.java
ptupitsyn: Removed

korlov42:
@ptupitsyn, I took the liberty of correcting the documentation a little bit. If you agree with proposed changes, please proceed with merging.

PR #1617, opened 2023-02-01
IGNITE-18213 Sql. Make exchange rewindable

korlov42:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/Fragment.java
```diff
@@ -64,19 +73,21 @@ public class Fragment {
...
-        this(id, root, remotes, null, null);
+        this(id, root, expectsCorrelatesFromOutside(root), remotes, null, null);
     }
 
     /**
      * Constructor.
      * TODO Documentation https://issues.apache.org/jira/browse/IGNITE-15859
      */
-    Fragment(long id, IgniteRel root, List<IgniteReceiver> remotes, @Nullable String rootSer, @Nullable FragmentMapping mapping) {
+    Fragment(long id, IgniteRel root, boolean expectsCorrelatesFromOutside, List<IgniteReceiver> remotes,
```
AMashenkov, line 83: I guess, we can pass 'expectsCorrelatesFromOutside' to constructor through FragmentProto in FragmenSplitter.go(Fragment).
see: stack.push(new FragmentProto(IdGenerator.nextId(), fragment.root()));

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/Inbox.java
```diff
@@ -475,6 +482,7 @@ enum State {
         }
 
         private final PriorityQueue<Batch<RowT>> batches = new PriorityQueue<>(IO_BATCH_CNT);
+        private final AtomicReference<SharedState> sharedStateHolder = new AtomicReference<>();
```
AMashenkov, line 485: Seems, it just used to exchange value in a single line of code without implying any atomicity requirement.
Thus it can be replace with plain field, which can be explicitly nullified in a code with the comment like "we won't send shared state more than once until next rewind".

AMashenkov (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/Fragment.java
korlov42: I reworked this part because of a races (see comment to commit [74fe27e](https://github.com/apache/ignite-3/pull/1617/commits/74fe27ec36f980bc9fe3688deff68ca5a4efbfff) for details). Your suggestion was taken into account

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/Inbox.java
korlov42: fixed

AMashenkov (approved):

PR #1619, opened 2023-02-02
IGNITE-18020 Add GC support to rocksdb storage

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-18020


gradle/libs.versions.toml
```diff
@@ -57,7 +57,7 @@ jsonpath = "2.4.0"
 classgraph = "4.8.110"
 javassist = "3.28.0-GA"
 checker = "3.10.0"
-rocksdb = "7.3.1"
+rocksdb = "7.9.2"
```
tkalkirill, line 60: Why did you decide to update the version?

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
```diff
@@ -225,8 +225,9 @@ public interface MvPartitionStorage extends ManuallyCloseable {
      * @param lowWatermark A time threshold for the row. Rows younger then the watermark value will not be removed.
      * @return A pair of table row and row id, where a timestamp of the row is less than or equal to {@code lowWatermark}.
      *      {@code null} if there's no such value.
+     * @throws StorageException If failed to poll element for vacuum.
      */
-    default @Nullable TableRowAndRowId pollForVacuum(HybridTimestamp lowWatermark) {
+    default @Nullable TableRowAndRowId pollForVacuum(HybridTimestamp lowWatermark) throws StorageException {
```
tkalkirill, line 230: `StorageException` is not checked, there is no need to write it in the signature, it is enough to mention it in the documentation.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
```diff
@@ -52,58 +52,36 @@
...
  */
 public abstract class BaseMvStoragesTest {
     /** Default reflection marshaller factory. */
-    protected static MarshallerFactory marshallerFactory;
+    protected static final MarshallerFactory marshallerFactory = new ReflectionMarshallerFactory();
 
     /** Schema descriptor for tests. */
-    protected static SchemaDescriptor schemaDescriptor;
+    protected static final SchemaDescriptor schemaDescriptor = new SchemaDescriptor(1, new Column[]{
+            new Column("intKey".toUpperCase(Locale.ROOT), NativeTypes.INT32, false),
```
tkalkirill, line 65: Why do you need **.toUpperCase(Locale.ROOT)**?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
```diff
@@ -181,7 +159,7 @@ protected static List<IgniteBiTuple<TestKey, TestValue>> drainToList(Cursor<Read
         }
     }
 
-    protected final void assertRowMatches(TableRow rowUnderQuestion, TableRow expectedRow) {
+    protected final void assertRowMatches(@Nullable TableRow rowUnderQuestion, TableRow expectedRow) {
```
tkalkirill, line 162: Why is **rowUnderQuestion** `@Nullable`? Logically, it just can't be.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/ColumnFamilyUtils.java
```diff
@@ -35,6 +35,11 @@ class ColumnFamilyUtils {
      */
     static final String PARTITION_CF_NAME = "cf-part";
 
+    /**
+     * Name of the Column Family that stores garbage collection queue.
+     */
```
tkalkirill, line 40:
```suggestion
    /** Name of the Column Family that stores garbage collection queue. */
```

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/ColumnFamilyUtils.java
```diff
@@ -49,7 +54,7 @@ class ColumnFamilyUtils {
      * Utility enum to describe a type of the column family - meta or partition.
      */
     enum ColumnFamilyType {
-        META, PARTITION, HASH_INDEX, SORTED_INDEX, UNKNOWN;
+        META, PARTITION, GC_QUEUE, HASH_INDEX, SORTED_INDEX, UNKNOWN;
```
tkalkirill, line 57: Just thoughts: why not write down the names in the enumeration, for example, `GC_QUEUE("cf-gc")`, or use the enum name as the name for the cf.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
```diff
@@ -586,7 +602,7 @@ private List<String> getExistingCfNames() {
 
             // even if the database is new (no existing Column Families), we return the names of mandatory column families, that
             // will be created automatically.
-            return existingNames.isEmpty() ? List.of(META_CF_NAME, PARTITION_CF_NAME, HASH_INDEX_CF_NAME) : existingNames;
+            return existingNames.isEmpty() ? List.of(META_CF_NAME, PARTITION_CF_NAME, GC_QUEUE_CF_NAME, HASH_INDEX_CF_NAME) : existingNames;
```
tkalkirill, line 605: Just thoughts: It would be more convenient to iterate with an enum.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
SammyVimes: First of all, read returns `nullable`, so we pass nullable to the non-null, bad.
Second, in the first line of method we assert it not to be null. So it is nullable

gradle/libs.versions.toml
SammyVimes: There is a fix for the deleteRange method. It was previously implemented in a sub-optimal way that was visible with the profiler.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
SammyVimes: Every other method of MvPartitionStorage explicitly states that it throws StorageException, so this looks consistent

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
SammyVimes: I don't need it, I moved this code from the `BeforeAll` method to the `clinit` part

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/ColumnFamilyUtils.java
SammyVimes: We can do this, but we can't use enum name as CF name, because the default one comes from the RocksDB itself

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
SammyVimes: Not really, we don't have `SORTED_INDEX` and `UNKNOWN` here

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -595,21 +619,97 @@ public void addWriteCommitted(RowId rowId, @Nullable TableRow row, HybridTimesta
...
 
             ByteBuffer keyBuf = prepareHeapKeyBuf(rowId);
-            putTimestamp(keyBuf, commitTimestamp);
+            putTimestampDesc(keyBuf, commitTimestamp);
+
+            boolean isNewValueTombstone = row == null;
 
             //TODO IGNITE-16913 Add proper way to write row bytes into array without allocations.
-            byte[] rowBytes = rowBytes(row);
+            byte[] rowBytes = row != null ? rowBytes(row) : new byte[0];
```
tkalkirill, line 627: Handle `null` - awesome.
Please use `org.apache.ignite.internal.util.ArrayUtils#BYTE_EMPTY_ARRAY`

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -570,17 +588,23 @@ public void commitWrite(RowId rowId, HybridTimestamp timestamp) throws StorageEx
...
 
                 if (valueBytes == null) {
-                    //the chain doesn't contain an uncommitted write intent
+                    // The chain doesn't contain an uncommitted write intent.
                     return null;
                 }
 
+                boolean isNewValueTombstone = valueBytes.length == VALUE_HEADER_SIZE;
+
+                boolean newAndPrevTombstones = maybeAddToGcQueue(writeBatch, rowId, timestamp, isNewValueTombstone);
```
tkalkirill, line 597: Please give a little comment on the variable, what does it mean?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -595,21 +619,97 @@ public void addWriteCommitted(RowId rowId, @Nullable TableRow row, HybridTimesta
...
             } catch (RocksDBException e) {
-                throw new StorageException("Failed to update a row in storage", e);
+                throw new StorageException("Failed to add row to the GC queue", e);
+            }
+
+            if (!newAndPrevTombstones) {
+                try {
+                    writeBatch.put(cf, copyOf(keyBuf.array(), MAX_KEY_SIZE), rowBytes);
+                } catch (RocksDBException e) {
+                    throw new StorageException("Failed to update a row in storage", e);
```
tkalkirill, line 640: Please add `RocksDbMvPartitionStorage#createStorageInfo`

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -595,21 +619,97 @@ public void addWriteCommitted(RowId rowId, @Nullable TableRow row, HybridTimesta
...
 
+            boolean newAndPrevTombstones;
             try {
-                writeBatch.put(cf, copyOf(keyBuf.array(), MAX_KEY_SIZE), rowBytes);
-
-                return null;
+                newAndPrevTombstones = maybeAddToGcQueue(writeBatch, rowId, commitTimestamp, isNewValueTombstone);
             } catch (RocksDBException e) {
-                throw new StorageException("Failed to update a row in storage", e);
+                throw new StorageException("Failed to add row to the GC queue", e);
```
tkalkirill, line 633: Please add `RocksDbMvPartitionStorage#createStorageInfo`

gradle/libs.versions.toml
tkalkirill: More like a separate problem, not related to GC.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
tkalkirill: These methods do not have an `StorageException` in their signature:
- 'org.apache.ignite.internal.storage.MvPartitionStorage#flush'
- 'org.apache.ignite.internal.storage.MvPartitionStorage#lastAppliedIndex'
- 'org.apache.ignite.internal.storage.MvPartitionStorage#lastAppliedTerm'
- 'org.apache.ignite.internal.storage.MvPartitionStorage#persistedIndex'
- 'org.apache.ignite.internal.storage.MvPartitionStorage#committedGroupConfiguration()'
- 'org.apache.ignite.internal.storage.MvPartitionStorage#committedGroupConfiguration(org.apache.ignite.internal.storage.RaftGroupConfiguration)'
- 'org.apache.ignite.internal.storage.MvPartitionStorage#close'

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
tkalkirill: You did change those lines of code) can we do better then?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
tkalkirill: Ok

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
tkalkirill: I see, thanks!

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/ColumnFamilyUtils.java
tkalkirill: Ok

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -137,6 +140,15 @@ public class RocksDbMvPartitionStorage implements MvPartitionStorage {
     /** Maximum size of the key. */
     private static final int MAX_KEY_SIZE = ROW_PREFIX_SIZE + HYBRID_TIMESTAMP_SIZE;
 
+    /** Garbage collector's queue key's timestamp offset. */
+    private static final int GC_KEY_TS_OFFSET = Short.BYTES;
+
+    /** Garbage collector's queue key's row id offset. */
```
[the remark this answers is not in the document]
SammyVimes: There is no point to do it for constants because they can change (while documentation can become stale) and IntelliJ IDEA calculates constant's value
image

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
SammyVimes: Sure

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -1015,6 +1117,149 @@ public void destroyData(WriteBatch writeBatch) throws RocksDBException {
         writeBatch.deleteRange(cf, partitionStartPrefix(), partitionEndPrefix());
     }
 
+    @Override
+    public @Nullable TableRowAndRowId pollForVacuum(HybridTimestamp lowWatermark) {
```
tkalkirill, line 1127: What about rebalancing?

If we are in the process of rebalancing, should we return rows? I think we should throw an exception, like in `RocksDbMvPartitionStorage#read`.

Since we clean up the storage at the beginning/aborting of the rebalance, I think that the queue should also be cleaned up.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -1015,6 +1117,149 @@ public void destroyData(WriteBatch writeBatch) throws RocksDBException {
...
+                    RocksIterator gcIt = db.newIterator(gc, gcOpts);
+            ) {
+                gcIt.seek(partitionStartPrefix());
+
+                if (invalid(gcIt)) {
+                    // GC queue is empty.
+                    return null;
+                }
+
+                ByteBuffer gcKey = allocateDirect(GC_KEY_SIZE);
```
tkalkirill, line 1143: Let's use the buffer too, shall we?

tkalkirill, line 1143:
```suggestion
                ByteBuffer gcKeyBuffer = allocateDirect(GC_KEY_SIZE);
```

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -1015,6 +1117,149 @@ public void destroyData(WriteBatch writeBatch) throws RocksDBException {
         writeBatch.deleteRange(cf, partitionStartPrefix(), partitionEndPrefix());
     }
 
+    @Override
+    public @Nullable TableRowAndRowId pollForVacuum(HybridTimestamp lowWatermark) {
+        return busy(() -> {
+            WriteBatchWithIndex batch = requireWriteBatch();
+
+            try (
```
tkalkirill, line 1131: Please describe what you are doing here.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -1015,6 +1117,149 @@ public void destroyData(WriteBatch writeBatch) throws RocksDBException {
...
+                HybridTimestamp timestamp = readTimestampNatural(gcKey, GC_KEY_TS_OFFSET);
+
+                if (timestamp.compareTo(lowWatermark) > 0) {
+                    // No elements to garbage collect.
+                    return null;
+                }
+
+                RowId readRowId = getRowId(gcKey, GC_KEY_ROW_ID_OFFSET);
+
+                try (
```
tkalkirill, line 1158: Please describe what you are doing here.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -1015,6 +1117,149 @@ public void destroyData(WriteBatch writeBatch) throws RocksDBException {
...
+                        }
+                    }
+
+                    if (noRowForGcElement) {
+                        gcKey.rewind();
+
+                        try {
+                            batch.delete(gc, gcKey);
+                        } catch (RocksDBException e) {
+                            throw new StorageException("Failed to collect garbage", e);
```
tkalkirill, line 1193: Please add `RocksDbMvPartitionStorage#createStorageInfo`

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -1015,6 +1117,149 @@ public void destroyData(WriteBatch writeBatch) throws RocksDBException {
...
+                        }
+
+                        return null;
+                    }
+
+                    // Check if the new element, whose insertion scheduled the GC, was a tombstone.
+                    int len = it.value(EMPTY_DIRECT_BUFFER);
```
tkalkirill, line 1200:
```suggestion
                    int len = it.value(EMPTY_DIRECT_BUFFER);
                    
```

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -1015,6 +1117,149 @@ public void destroyData(WriteBatch writeBatch) throws RocksDBException {
...
+                    dataKey.clear();
+
+                    dataKey.putShort((short) partitionId);
+                    putRowId(dataKey, readRowId);
+                    putTimestampDesc(dataKey, timestamp);
+                    dataKey.flip();
+
+                    it.seek(dataKey);
+
+                    boolean noRowForGcElement = false;
```
tkalkirill, line 1172:
```suggestion
                    boolean noRowForGcElement = false;
                    
```

gradle/libs.versions.toml
SammyVimes: Why not if we are at it?

tkalkirill: I think we should make a fix that is as close to the problem as possible.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/MvPartitionStorage.java
SammyVimes: What about:
- `V execute() throws StorageException;`
- `<V> V runConsistently(WriteClosure<V> closure) throws StorageException`
- `void lastApplied(long lastAppliedIndex, long lastAppliedTerm) throws StorageException`
- `ReadResult read(RowId rowId, HybridTimestamp timestamp) throws StorageException;`
- `@Nullable TableRow addWrite(RowId rowId, @Nullable TableRow row, UUID txId, UUID commitTableId, int commitPartitionId) throws TxIdMismatchException, StorageException` (this one has two!)
- `@Nullable TableRow abortWrite(RowId rowId) throws StorageException`
- `void commitWrite(RowId rowId, HybridTimestamp timestamp) throws StorageException`
- `void addWriteCommitted(RowId rowId, @Nullable TableRow row, HybridTimestamp commitTimestamp) throws StorageException`
- `Cursor<ReadResult> scanVersions(RowId rowId) throws StorageException`
- `PartitionTimestampCursor scan(HybridTimestamp timestamp) throws StorageException;`
- `@Nullable RowId closestRowId(RowId lowerBound) throws StorageException;`

tkalkirill: You wrote "Every other", but it's not, I suggest not to change the signature.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -1015,6 +1117,149 @@ public void destroyData(WriteBatch writeBatch) throws RocksDBException {
...
+                    return null;
+                }
+
+                ByteBuffer gcKey = allocateDirect(GC_KEY_SIZE);
+                ByteBuffer dataKey = MV_KEY_BUFFER.get();
+
+                gcKey.clear();
+                gcIt.key(gcKey);
+
+                HybridTimestamp timestamp = readTimestampNatural(gcKey, GC_KEY_TS_OFFSET);
```
tkalkirill, line 1149:
```suggestion
                HybridTimestamp keyTimestamp = readTimestampNatural(gcKey, GC_KEY_TS_OFFSET);
```

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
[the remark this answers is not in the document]
tkalkirill: Ok

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -595,21 +619,97 @@ public void addWriteCommitted(RowId rowId, @Nullable TableRow row, HybridTimesta
...
+     * @param isNewValueTombstone If new row is a tombstone.
+     * @return {@code true} if new value and previous value are both tombstones.
+     * @throws RocksDBException If failed.
+     */
+    private boolean maybeAddToGcQueue(WriteBatchWithIndex writeBatch, RowId rowId, HybridTimestamp timestamp, boolean isNewValueTombstone)
+            throws RocksDBException {
+        boolean newAndPrevTombstones = false;
+
+        // Try find previous value for the row id.
+        ByteBuffer key = allocateDirect(MAX_KEY_SIZE);
```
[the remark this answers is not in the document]
SammyVimes: Heap buffers can't be used with rocksdb. We have direct buffer (MV_BUFFER), but I'd argue that it should not be used in private methods, because they could have been already used one frame higher, thus we will have a silent but deadly bugs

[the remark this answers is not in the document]
tkalkirill: My fault, I meant that we can use `ThreadLocal` so as not to allocate the buffer each time.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
SammyVimes: What do you mean, it is used

tkalkirill: Let's use `ThreadLocal`.

SammyVimes: No way, we're already using it for another key

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -1015,6 +1123,218 @@ public void destroyData(WriteBatch writeBatch) throws RocksDBException {
...
+
+            WriteBatchWithIndex batch = requireWriteBatch();
+
+            // We retrieve the first element of the GC queue and seeking for it in the data CF.
+            // However, the element that we need to garbage collect is the next (older one) element.
+            // First we check if there's anything to garbage collect. If the element is a tombstone we remove it.
+            // If the next element is exists, that should be the element that we want to garbage collect.
+            try (
+                    var gcUpperBound = new Slice(partitionEndPrefix());
+                    ReadOptions gcOpts = new ReadOptions().setIterateUpperBound(gcUpperBound).setTotalOrderSeek(true);
```
sashapolo, line 1139: Why is `.setTotalOrderSeek(true)` needed? Please leave a comment

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -171,6 +186,9 @@ public class RocksDbMvPartitionStorage implements MvPartitionStorage {
     /** Meta column family. */
     private final ColumnFamilyHandle meta;
 
+    /** GC queue column family. */
```
sashapolo, line 189: Is there a description anywhere about what is a GC queue?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -570,17 +589,26 @@ public void commitWrite(RowId rowId, HybridTimestamp timestamp) throws StorageEx
...
+                boolean newAndPrevTombstones = maybeAddToGcQueue(writeBatch, rowId, timestamp, isNewValueTombstone);
+
                 // Delete pending write.
                 writeBatch.delete(cf, uncommittedKeyBytes);
 
-                // Add timestamp to the key, and put the value back into the storage.
-                putTimestamp(keyBuf, timestamp);
+                // We only write tombstone if the previous value for the same row id was not a tombstone.
+                // So there won't be consecutive tombstones for the same row id.
+                if (!newAndPrevTombstones) {
```
sashapolo, line 606: Why wasn't this check present in the previous implementation? Was it a bug? And when can we have two tombstones for the same key anyway?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -595,21 +623,99 @@ public void addWriteCommitted(RowId rowId, @Nullable TableRow row, HybridTimesta
...
+
+        try (
+                Slice upperBound = new Slice(partitionEndPrefix());
+                ReadOptions readOpts = new ReadOptions().setTotalOrderSeek(true).setIterateUpperBound(upperBound);
+                RocksIterator it = db.newIterator(cf, readOpts)
+        ) {
+            key.flip();
+            it.seek(key);
+
+            if (!invalid(it)) {
```
sashapolo, line 684: I would suggest to invert this condition

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -1015,6 +1123,218 @@ public void destroyData(WriteBatch writeBatch) throws RocksDBException {
...
     }
 
+    @Override
+    public @Nullable TableRowAndRowId pollForVacuum(HybridTimestamp lowWatermark) {
+        return busy(() -> {
+            throwExceptionIfStorageInProgressOfRebalance(state.get(), this::createStorageInfo);
+
+            WriteBatchWithIndex batch = requireWriteBatch();
+
+            // We retrieve the first element of the GC queue and seeking for it in the data CF.
```
sashapolo, line 1133:
```suggestion
            // We retrieve the first element of the GC queue and seek for it in the data CF.
```

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -1015,6 +1123,218 @@ public void destroyData(WriteBatch writeBatch) throws RocksDBException {
...
+            // However, the element that we need to garbage collect is the next (older one) element.
+            // First we check if there's anything to garbage collect. If the element is a tombstone we remove it.
+            // If the next element is exists, that should be the element that we want to garbage collect.
```
sashapolo, line 1136:
```suggestion
            // If the next element exists, that should be the element that we want to garbage collect.
```

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -1015,6 +1123,218 @@ public void destroyData(WriteBatch writeBatch) throws RocksDBException {
...
+
+                if (invalid(gcIt)) {
+                    // GC queue is empty.
+                    return null;
+                }
+
+                ByteBuffer gcKeyBuffer = allocateDirect(GC_KEY_SIZE);
+                gcKeyBuffer.clear();
+
+                ByteBuffer dataKeyBuffer = MV_KEY_BUFFER.get();
```
sashapolo, line 1152: Why do we clear this buffer here? What if the lowWatermark is too small and we exit early?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -1015,6 +1123,218 @@ public void destroyData(WriteBatch writeBatch) throws RocksDBException {
...
+                if (gcElementTimestamp.compareTo(lowWatermark) > 0) {
+                    // No elements to garbage collect.
+                    return null;
+                }
+
+                RowId gcElementRowId = getRowId(gcKeyBuffer, GC_KEY_ROW_ID_OFFSET);
+
+                try (
+                        var upperBound = new Slice(partitionEndPrefix());
+                        ReadOptions opts = new ReadOptions().setIterateUpperBound(upperBound).setTotalOrderSeek(true);
```
sashapolo, line 1168: Looks like these read options may be cached

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/TableRowAndRowId.java
```diff
@@ -24,7 +24,7 @@
  * Wrapper that holds both {@link TableRow} and {@link RowId}. {@link TableRow} is null for tombstones.
  */
 public class TableRowAndRowId {
-    /** Table row. */
+    /** Table row. {@code null} if tombstone. */
```
tkalkirill, line 27: Now this is not entirely true, we discussed that this is possible even not for tombstones)
I propose to describe this in the documentation for the method.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/GarbageCollector.java
```diff
@@ -0,0 +1,339 @@
...
+ *
+ * <p>Key:
+ * <pre>{@code
+ * | partId (2 bytes, BE) | timestamp (12 bytes, ASC) | rowId (16 bytes, BE) |
+ * }</pre>
+ * Value is an empty byte array.
+ *
+ * <p>For further information refer to the tech-notes/garbage-collection.md in this module.
+ */
+class GarbageCollector {
```
tkalkirill, line 59: Very noice

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/GarbageCollector.java
```diff
@@ -0,0 +1,341 @@
...
+     * @param timestamp New row's timestamp.
+     * @param isNewValueTombstone If new row is a tombstone.
+     * @return {@code true} if new value and previous value are both tombstones.
+     * @throws RocksDBException If failed.
+     */
+    boolean tryAddToGcQueue(WriteBatchWithIndex writeBatch, RowId rowId, HybridTimestamp timestamp, boolean isNewValueTombstone)
+            throws RocksDBException {
+        RocksDB db = helper.db;
+        ColumnFamilyHandle gc = helper.gc;
+        ColumnFamilyHandle cf = helper.cf;
```
tkalkirill, line 102: Let's rename to `partCf`

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/GarbageCollector.java
```diff
@@ -0,0 +1,341 @@
...
+     *
+     * @param batch Write batch.
+     * @param lowWatermark Low watermark.
+     * @return Garbage collected element.
+     * @throws RocksDBException If failed to collect the garbage.
+     */
+    @Nullable TableRowAndRowId pollForVacuum(WriteBatchWithIndex batch, HybridTimestamp lowWatermark) throws RocksDBException {
+        RocksDB db = helper.db;
+        ColumnFamilyHandle gc = helper.gc;
+        ColumnFamilyHandle cf = helper.cf;
```
tkalkirill, line 164: Let's rename to `partCf`

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/GarbageCollector.java
```diff
@@ -0,0 +1,341 @@
...
+                // No elements to garbage collect.
+                return null;
+            }
+
+            RowId gcElementRowId = helper.getRowId(gcKeyBuffer, GC_KEY_ROW_ID_OFFSET);
+
+            // Delete element from the GC queue.
+            batch.delete(gc, gcKeyBuffer);
+
+            try (
```
tkalkirill, line 199: Please also add some description.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/GarbageCollector.java
```diff
@@ -0,0 +1,341 @@
...
+     *
+     * @param it RocksDB data column family iterator.
+     * @param batch Write batch.
+     * @param dataKeyBuffer Buffer for the data column family key.
+     * @param gcElementRowId Row id of the element from the GC queue/
+     * @return {@code true} if further processing by garbage collector is needed.
+     */
+    private boolean checkHasNewerRowAndRemoveTombstone(RocksIterator it, WriteBatchWithIndex batch, ByteBuffer dataKeyBuffer,
+            RowId gcElementRowId, HybridTimestamp gcElementTimestamp) throws RocksDBException {
+        ColumnFamilyHandle cf = helper.cf;
```
tkalkirill, line 252: Let's rename to `partCf`

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/GarbageCollector.java
```diff
@@ -0,0 +1,341 @@
...
+
+        // Seek to the row id and timestamp from the GC queue.
+        // Note that it doesn't mean that the element in this iterator has matching row id or even partition id.
+        it.seek(dataKeyBuffer);
+
+        boolean hasRowForGc = true;
+
+        if (invalid(it)) {
+            // There is no row for the GC queue element.
+            hasRowForGc = false;
```
tkalkirill, line 265: Can be returned immediately. =)

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/GarbageCollector.java
```diff
@@ -0,0 +1,341 @@
...
+        } else {
+            dataKeyBuffer.clear();
+
+            it.key(dataKeyBuffer);
+
+            if (!helper.getRowId(dataKeyBuffer, ROW_ID_OFFSET).equals(gcElementRowId)) {
+                // There is no row for the GC queue element.
+                hasRowForGc = false;
```
tkalkirill, line 273: Can be returned immediately. =)

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/GarbageCollector.java
```diff
@@ -0,0 +1,341 @@
...
+
+        // Check if the new element, whose insertion scheduled the GC, was a tombstone.
+        int len = it.value(EMPTY_DIRECT_BUFFER);
+
+        if (len == 0) {
+            // This is a tombstone, we need to delete it.
+            batch.delete(cf, dataKeyBuffer);
+        }
+
+        return true;
```
tkalkirill, line 289: But what about the `hasRowForGc` ? =)

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/GarbageCollector.java
```diff
@@ -0,0 +1,341 @@
...
+        if (invalid(it)) {
+            hasRowForGc = false;
+        } else {
+            dataKeyBuffer.clear();
+
+            int keyLen = it.key(dataKeyBuffer);
+
+            if (keyLen != MAX_KEY_SIZE) {
+                // We moved to the next row id's write-intent, so there was no row to GC for the current row id.
+                hasRowForGc = false;
```
tkalkirill, line 318: Can be returned immediately. =)

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/GarbageCollector.java
```diff
@@ -0,0 +1,341 @@
...
+            } else {
+                gcRowId = helper.getRowId(dataKeyBuffer, ROW_ID_OFFSET);
+
+                if (!gcElementRowId.equals(gcRowId)) {
+                    // We moved to the next row id, so there was no row to GC for the current row id.
+                    hasRowForGc = false;
```
tkalkirill, line 324: Can be returned immediately. =)

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
```diff
@@ -181,7 +158,7 @@ protected static List<IgniteBiTuple<TestKey, TestValue>> drainToList(Cursor<Read
...
```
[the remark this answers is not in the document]
SammyVimes: Null is a valid value, because it can be passed as an argument here. Why do we perform null-check otherwise?

modules/storage-rocksdb/docs/garbage-collection.md
```diff
@@ -0,0 +1,83 @@
...
+is no such value in the storage.*
+
+| Record number | Row id | Timestamp |
+|---------------|--------|-----------|
+| 1             | Foo    | 1         |
+| 2             | Foo    | 10        |
+
+In this case, we can only remove record 1 if the low watermark is 10 or higher. If watermark is at 9, 
+then it means that there can still occur a transaction with a 9 timestamp, which means that the record number 1 
+is still needed.
```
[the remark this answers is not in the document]
SammyVimes: Nope, we need both. Because transaction with ts==9 can still occur, meaning that we need to read record 1, not `null`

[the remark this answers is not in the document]
tkalkirill: You are right!

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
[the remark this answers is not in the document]
tkalkirill: At the discretion of Alexander.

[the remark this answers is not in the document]
sashapolo: Null is not a valid value if you throw an exception when this value is passed. `@Nullable` parameter means that `null` can be accepted by a function in a meaningful way.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/GarbageCollector.java
SammyVimes: I think there is enough description in the documentation. Also now there is an exhaustive javadoc for GC's methods, so it should be pretty understandable

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/TableRowAndRowId.java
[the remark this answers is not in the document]
SammyVimes: What can it be besides tombstone?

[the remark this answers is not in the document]
tkalkirill: Then I looked at your code, you also have a check that if you delete one and the same version in parallel, then we will return zero or something like that, I think that the documentation is correct.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/GarbageCollector.java
tkalkirill: Ok

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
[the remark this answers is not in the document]
SammyVimes:
```
By convention, this annotation applied only when the value should always be checked against {@code null} because the developer could do nothing to prevent null from happening.
```
It can be accepted in a meaningful way. It means, that there's an error.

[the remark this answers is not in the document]
SammyVimes: Having `null` here is not an error in program (our test), it's an error in the tested program, isn't it?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/GarbageCollector.java
```diff
@@ -0,0 +1,315 @@
...
+     * @param writeBatch Write batch.
+     * @param rowId Row id.
+     * @param timestamp New row's timestamp.
+     * @param isNewValueTombstone If new row is a tombstone.
+     * @return {@code true} if new value and previous value are both tombstones.
+     * @throws RocksDBException If failed.
+     */
+    boolean tryAddToGcQueue(WriteBatchWithIndex writeBatch, RowId rowId, HybridTimestamp timestamp, boolean isNewValueTombstone)
+            throws RocksDBException {
+        RocksDB db = helper.db;
```
sashapolo, line 98: I think this stuff can simply be this class' fields

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/GarbageCollector.java
```diff
@@ -0,0 +1,315 @@
...
+     * If there is no row in the data column family, returns {@code false} as no further processing is required.
+     * if there is a row and this entry is a tombstone, removes tombstone.
+     *
+     * @param it RocksDB data column family iterator.
+     * @param batch Write batch.
+     * @param dataKeyBuffer Buffer for the data column family key.
+     * @param gcElementRowId Row id of the element from the GC queue/
+     * @return {@code true} if further processing by garbage collector is needed.
+     */
+    private boolean checkHasNewerRowAndRemoveTombstone(RocksIterator it, WriteBatchWithIndex batch, ByteBuffer dataKeyBuffer,
```
sashapolo, line 236: why do you need the `dataKeyBuffer` parameter if it's already a field?

modules/storage-rocksdb/docs/garbage-collection.md
```diff
@@ -0,0 +1,83 @@
...
+|--------------|-------------------------------------------|----------------|
+| 2-byte       | 12-byte (8-byte physical, 4-byte logical) | 16-byte (uuid) |
+
+The value is not stored, as we only need the key. We can make row id the value,
+because for the ascending order processing of the queue we only need the timestamp,
+however, multiple row ids can have same timestamp, so making row id a value requires storing a list of
+row ids, hence the commit in this implementation of the storage becomes more sophisticated and, probably,
+less performant.
+
+Each time a row is being committed to the storage, we perform a check whether
```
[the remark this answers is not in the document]
SammyVimes: It's obvious, come on

modules/storage-rocksdb/docs/garbage-collection.md
```diff
@@ -0,0 +1,83 @@
+# Garbage Collection in the RocksDB partition storage
+
+## Garbage Collection queue
```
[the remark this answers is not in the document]
SammyVimes: Why? We first put elements to queue and then poll the queue, not the other way around

modules/storage-rocksdb/docs/garbage-collection.md
```diff
@@ -0,0 +1,83 @@
+# Garbage Collection in the RocksDB partition storage
```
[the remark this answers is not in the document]
SammyVimes: Well, in network module doc files are in docs folder..

modules/storage-rocksdb/docs/garbage-collection.md
```diff
@@ -0,0 +1,83 @@
...
+Each time a row is being committed to the storage, we perform a check whether
+there is already a value for this row. If there is one and both it and new version are not tombstones, we put
+new commit's timestamp and row id into the GC queue. To understand why we only put new value's timestamp
+please refer to the Garbage Collection [algorithm](#garbage-collection-algorithm).  
+The queue is updated along with the data column family in a single batch and is destroyed when the storage
+is being cleared or destroyed.
+
+## Garbage Collection algorithm
+
+It's important to understand when we actually need to perform garbage collection.
```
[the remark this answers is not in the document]
SammyVimes: Well if a person doesn't know what garbage collections is, why do they need to read this documentation?

modules/storage-rocksdb/docs/garbage-collection.md
```diff
@@ -0,0 +1,83 @@
...
+| 3             | Foo    | 20        | False        |
+
+Everything said before stands for this example, however we can also remove the record number 2, because it is 
+a tombstone. So if the watermark is higher or equal to 10 and there is a transaction with timestamp higher than
+10, then we either get an empty value if timestamp is less than 20, or we get a newer version.
+
+So to sum up, the algorithm looks like this:
+1. Get an element from the GC queue, exiting if the queue is empty
+2. Add that element to the batch for removal from RocksDB
+3. Find an element in the data column family that corresponds to the element of GC queue. If a value doesn't exist, exit
```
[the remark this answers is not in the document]
SammyVimes: This will just make bloat the numbered list

modules/storage-rocksdb/docs/garbage-collection.md
```diff
@@ -0,0 +1,83 @@
...
+4. Test if it is a tombstone, if yes, add it to the batch for removal
```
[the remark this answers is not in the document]
SammyVimes: No, on step 2 we remove an element of the queue

modules/storage-rocksdb/docs/garbage-collection.md
```diff
@@ -0,0 +1,83 @@
...
+6. Add that previous version to the batch for removal
+
+You might notice that there are two cases when we can exit prematurely, apart from queue being empty.  
+We might have not found a value that triggered the addition to the GC queue and/or the value that needs to be
+garbage collected because GC can run in parallel. So if two parallel threads got the same element from the 
+queue, one of them might have already finished the GC and removed the elements.
+
+## Storage implications
+
+To save space we don't store consecutive tombstones.
```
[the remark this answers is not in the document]
SammyVimes: This is obvious, there's nowhere else we can store tombstones

modules/storage-rocksdb/docs/garbage-collection.md
```diff
@@ -0,0 +1,83 @@
...
+For example, if a user removes a certain row twice
+
+```
+storage.put(key,value);
+storage.delete(key);
+storage.delete(key);
+```
+
+There should be one row with a value and one row with a tombstone, the tombstone being
+the most recent one. This also simplifies the processing of the garbage collection queue.
```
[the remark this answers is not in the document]
SammyVimes: An absence of a value (which is the same as seeing a tombstone)

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/PartitionDataHelper.java
```diff
@@ -0,0 +1,221 @@
...
+import org.apache.ignite.internal.storage.RowId;
+import org.rocksdb.ColumnFamilyHandle;
+import org.rocksdb.ReadOptions;
+import org.rocksdb.RocksDB;
+import org.rocksdb.Slice;
+
+/** Helper for the partition data. */
+class PartitionDataHelper implements ManuallyCloseable {
+    /** Commit partition id size. */
+    static final int PARTITION_ID_SIZE = Short.BYTES;
```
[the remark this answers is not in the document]
SammyVimes: Because other constants depend on these constants and I don't want a cycle dependency between this class and a storage class

modules/storage-rocksdb/docs/garbage-collection.md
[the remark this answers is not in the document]
sashapolo: Because it better suits as the introduction on why GC is needed. GC queue is just a part of the alogrithm

modules/storage-rocksdb/docs/garbage-collection.md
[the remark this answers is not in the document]
sashapolo: you even have a link to this document in the code, which assumes that it is located in the `tech-notes` folder

modules/storage-rocksdb/docs/garbage-collection.md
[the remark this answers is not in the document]
sashapolo: Because otherwise this documents looks too abrupt

modules/storage-rocksdb/docs/garbage-collection.md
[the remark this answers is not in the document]
sashapolo: why will it see an absense of value, if there's a "foo" 1 value?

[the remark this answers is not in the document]
SammyVimes: Because it's past 2, it's 3. If we have a transaction with ts 1 we will see "foo" 1 value

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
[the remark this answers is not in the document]
sashapolo: mkay, so this is a part of the test then

sashapolo (approved):

PR #1620, opened 2023-02-02
IGNITE-18701 Incorrect binary tuple column mapping

valepakh:
https://issues.apache.org/jira/browse/IGNITE-18701


ademakov (approved):
LGTM

PR #1622, opened 2023-02-02
IGNITE-18430 Add integration test for a case when RAFT snapshot installation is cancelled in the middle

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18430


SammyVimes (approved):
LGTM

PR #1626, opened 2023-02-03
IGNITE-14691 Implement reactive Meta Storage range methods

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-14691


modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorSubscription.java
```diff
@@ -0,0 +1,185 @@
...
+    /** Cached response from the Meta Storage leader. {@code null} if no requests have been issued yet. */
+    @Nullable
+    private BatchResponse cachedResponse;
+
+    /** Index inside the cached response. */
+    private int responseIndex;
+
+    /** Amount of entries requested by the subscriber. */
+    private long demand;
+
```
tkalkirill, line 70: Shouldn't variables be thread-safe?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorSubscription.java
```diff
@@ -0,0 +1,185 @@
...
+    private final MetaStorageCommandsFactory commandsFactory;
+
+    private final ExecutorService executorService;
+
+    private final IgniteUuid cursorId;
+
+    private final Subscriber<? super Entry> subscriber;
+
+    /** Flag indicating that either the whole data range has been exhausted or the subscription has been cancelled. */
+    private boolean isDone = false;
```
tkalkirill, line 59: same

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageServiceImpl.java
```diff
@@ -82,29 +88,27 @@ public class MetaStorageServiceImpl implements MetaStorageService {
     public MetaStorageServiceImpl(RaftGroupService metaStorageRaftGrpSvc, ClusterNode localNode) {
         this.metaStorageRaftGrpSvc = metaStorageRaftGrpSvc;
         this.localNode = localNode;
+        this.executorService = Executors.newCachedThreadPool(NamedThreadFactory.create(localNode.name(), "metastorage-publisher", LOG));
```
tkalkirill, line 91: Could there be too many threads here?

modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/impl/CursorPublisherTest.java
```diff
@@ -0,0 +1,184 @@
...
+    private RaftGroupService raftService;
+
+    private CursorPublisher publisher;
+
+    @BeforeEach
+    void setUp() {
+        publisher = new CursorPublisher(
+                raftService,
+                new MetaStorageCommandsFactory(),
+                ForkJoinPool.commonPool(),
```
tkalkirill, line 71: https://avatars.mds.yandex.net/i?id=1cbe54508db85747fa40c700b3a1d6fab3115558-7086141-images-thumbs&n=13

modules/schema/src/main/java/org/apache/ignite/internal/schema/SchemaManager.java
```diff
@@ -433,21 +434,46 @@ public void stop() throws Exception {
...
+            }
+
+            @Override
+            public void onComplete() {
+                latestVersionFuture.complete(lastVer);
+            }
+        });
+
+        try {
+            return latestVersionFuture.get(10, TimeUnit.SECONDS);
```
tkalkirill, line 470: This is bad, you need to get rid of the synchronous code.

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/flow/FlowUtils.java
```diff
@@ -0,0 +1,158 @@
...
+ */
+public class FlowUtils {
+    /**
+     * Subscribes to the given publisher, collecting all values into a list.
+     */
+    public static <T> CompletableFuture<List<T>> subscribeToList(Publisher<T> publisher) {
+        var resultFuture = new CompletableFuture<List<T>>();
+
+        publisher.subscribe(new Subscriber<>() {
+            private final List<T> items = new ArrayList<>();
```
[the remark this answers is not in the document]
sashapolo: no, why?

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/flow/FlowUtils.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.concurrent.Flow.Publisher;
+import java.util.concurrent.Flow.Subscriber;
+import java.util.concurrent.Flow.Subscription;
+import java.util.concurrent.ForkJoinPool;
+import org.apache.ignite.internal.util.Cursor;
+
+/**
+ * Utility methods for extracting data from {@link Publisher}s.
+ */
+public class FlowUtils {
```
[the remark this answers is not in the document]
sashapolo: what's the point? In the test package, you can't use it by accident

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/flow/FlowUtils.java
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * Creates a Publisher that provides data form a given cursor.
+     */
+    public static <T> Publisher<T> fromCursor(Cursor<T> cursor) {
+        return subscriber -> subscriber.onSubscribe(new Subscription() {
+            private long demand;
+
+            private boolean isStarted;
+
+            private boolean isCancelled;
```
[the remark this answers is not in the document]
sashapolo: nope

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/flow/FlowUtils.java
```diff
@@ -0,0 +1,158 @@
...
+
+                if (!isStarted) {
+                    isStarted = true;
+
+                    start();
+                }
+            }
+
+            private void start() {
+                ForkJoinPool.commonPool().execute(() -> {
```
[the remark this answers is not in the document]
sashapolo: why? This is test-only class. And a separate pool requires starting and stopping it, which is not convenient

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageServiceImpl.java
sashapolo: I don't know(

modules/schema/src/main/java/org/apache/ignite/internal/schema/SchemaManager.java
sashapolo: I can't do it in this PR, that's too many changes. And this code was synchronous before

tkalkirill: Please add TODO with a ticket.

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/flow/FlowUtils.java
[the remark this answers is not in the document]
tkalkirill: Discussed in private

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/flow/FlowUtils.java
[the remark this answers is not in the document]
tkalkirill: Discussed in private

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorSubscription.java
tkalkirill: Discussed in private

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorSubscription.java
tkalkirill: Discussed in private

modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
```diff
@@ -63,32 +63,40 @@ public interface MetaStorageManager extends IgniteComponent {
...
      * @param keyPrefix Prefix of the key to retrieve the entries. Couldn't be {@code null}.
-     * @return Cursor built upon entries corresponding to the given range and revision.
+     * @return Publisher that will provide entries corresponding to the given prefix.
      * @throws OperationTimeoutException If the operation is timed out.
      * @throws CompactedException If the desired revisions are removed from the storage due to a compaction.
      * @see ByteArray
      * @see Entry
      */
-    Cursor<Entry> prefix(ByteArray keyPrefix) throws NodeStoppingException;
+    Publisher<Entry> prefix(ByteArray keyPrefix);
```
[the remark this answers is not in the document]
tkalkirill: Please describe in the javaDoc.

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/flow/FlowUtils.java
```diff
@@ -0,0 +1,158 @@
...
+            public void onComplete() {
+                resultFuture.completeExceptionally(new NoSuchElementException());
+            }
+        });
+
+        return resultFuture;
+    }
+
+    /**
+     * Creates a Publisher that provides data form a given cursor.
```
SammyVimes, line 105:
```suggestion
     * Creates a Publisher that provides data from a given cursor.
```

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/flow/FlowUtils.java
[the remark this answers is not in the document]
sashapolo: Fixed

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageServiceImpl.java
```diff
@@ -82,29 +88,27 @@ public class MetaStorageServiceImpl implements MetaStorageService {
     public MetaStorageServiceImpl(RaftGroupService metaStorageRaftGrpSvc, ClusterNode localNode) {
         this.metaStorageRaftGrpSvc = metaStorageRaftGrpSvc;
         this.localNode = localNode;
+        this.executorService = Executors.newFixedThreadPool(5, NamedThreadFactory.create(localNode.name(), "metastorage-publisher", LOG));
```
tkalkirill, line 91: Maybe the number of CPUs?

tkalkirill (approved):

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageServiceImpl.java
sashapolo: why, how is this pool related to the number of CPUs? It's IO bound

tkalkirill: Will there be any problems if the user only has 4 and 32 CPUs? Will there be starvation or underutilization of resources?

modules/schema/src/main/java/org/apache/ignite/internal/schema/SchemaManager.java
```diff
@@ -432,22 +433,48 @@ public void stop() throws Exception {
...
```
tkalkirill, line 471: Forget TODO

sashapolo: TODO is located on line 436

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageServiceImpl.java
sashapolo: No, this thread pool is not CPU bound, it's tied to the amount of requests to the Meta Storage. 5 is an arbitrary value, I will add a TODO to extract this value to configuration.

sashapolo: done

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorPublisher.java
```diff
@@ -0,0 +1,85 @@
...
+    ) {
+        this.raftService = raftService;
+        this.commandsFactory = commandsFactory;
+        this.executorService = executorService;
+        this.createCursorSupplier = createCursorSupplier;
+    }
+
+    @Override
+    public void subscribe(Subscriber<? super Entry> subscriber) {
+        IgniteUuid cursorId = UUID_GENERATOR.randomUuid();
```
sanpwc, line 70: So, you do support multiple subscriptions, right?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorPublisher.java
```diff
@@ -0,0 +1,85 @@
...
+
+        WriteCommand createCursorCommand = createCursorSupplier.apply(cursorId);
+
+        raftService.run(createCursorCommand)
```
sanpwc, line 74: Did you consider the solution that will eliminate additional rtt on cursor creation? E.g. it's possible return BatchResponse even for cursor creation and postpone cursor creation until explicit request within onSubscribe.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorPublisher.java
```diff
@@ -0,0 +1,85 @@
...
+                .whenCompleteAsync((v, e) -> {
+                    if (e == null) {
```
sanpwc, line 76: Should we also cover such code with busyLock?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorSubscription.java
```diff
@@ -0,0 +1,185 @@
...
+    private void requestNextBatch() {
+        NextBatchCommand command = commandsFactory.nextBatchCommand()
+                .cursorId(cursorId)
+                .batchSize(BATCH_SIZE)
+                .build();
+
+        raftService.<BatchResponse>run(command)
+                .whenCompleteAsync((resp, e) -> {
+                    if (e == null) {
+                        cachedResponse = resp;
```
sanpwc, line 157: Not sure that I get it. Do you expect that whenCompleteAsync will be executed in the same thread as CursorSubscription.request(n)?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorPublisher.java
sashapolo: I don't see any technical problems with that (we will simply create multiple cursors, one per subscription). Thought I don't think it's useful in any way

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorPublisher.java
sashapolo: I will rework this part soon, when we are going to migrate to ReadCommands for cursors

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorPublisher.java
sashapolo: Makes sense, I always forget about this stuff

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorSubscription.java
sashapolo: > Is it always singleTreaded executor?

No

> Do you expect that whenCompleteAsync will be executed in the same thread as CursorSubscription.request(n)

No, but I expect that there's a following HB relationships:
1. For the first ever call of `request`: request -> submission on the executor. Though now I think that this may be debatable, because we are completing the future from a Raft service thread. Looking at the code I still think that there's a HB there because of the future chains, but maybe it makes sense to make `demand` variable volatile, what do you think?
2. For all consecutive calls `request` is expected to be done in the same thread as the `processRequest`  (because it is probably called inside the `onNext` callback)

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorPublisher.java
sanpwc: Make sense.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorSubscription.java
sanpwc: My fault, I've missed the 
```
        if (cachedResponse == null) {
            requestNextBatch();
        }
```
part. So you'll never request next portion before current one is fully processed, so there will be no multiple requestNextBatch calls on different threads.

Seems that demand should be fine without being volatile.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorSubscription.java
```diff
@@ -0,0 +1,185 @@
...
+                        processRequest();
+                    } else {
+                        onError(e);
+                    }
+                }, executorService);
+    }
+
+    @Override
+    public void cancel() {
+        isDone = true;
```
sanpwc, line 169: And what about this one? Seems that it's possible and valid to cancel subscriptions within some other threads, e.g. components.stop one.

sashapolo: Can you explain how that could happen? I also expect this method to be called from the same thread (e.g. during `onNext` or `onError`).

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorSubscription.java
sashapolo: Correct, there should be no concurrent `requestNextBatch` at the same moment, though consecutive batches can be processed on different threads (and I expect that we have a HB through future chains there)

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorPublisher.java
sanpwc: Well, I believe that you should either prevent multiple subscriptions or notify all registered ones on every new entry, error, etc. Currently you subscriptions are completely independent, so that each behaves as isolated PubSub pair instead of Pub(Sub1 + Sub2 + ...).

sashapolo: ok, I'll add a protection just in case, it doesn't really make much sense to have multiple subscribers in this case

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/CursorSubscription.java
sanpwc: I mean following (pseudocode)
```
    private volatile Subscription msRangeSub;
    
        ms.range().subscribe(new Subscriber<Object>() {

        @Override
        public void onSubscribe (Subscription subscription){
            this.msRangeSub = subscription;
        }
    }
    
    stop() {
        msRangeSub.cancel();
    }
```

sanpwc: Shame on you, you are right.
`Subscription.request and Subscription.cancel MUST only be called inside of its Subscriber context.`
https://github.com/reactive-streams/reactive-streams-jvm/

sanpwc (approved):

PR #1636, opened 2023-02-06
IGNITE-18710 Make default for delay to remove from logical topology non-zero

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18710


modules/runner/src/integrationTest/java/org/apache/ignite/internal/BootstrapConfigTemplateMethod.java
```diff
@@ -0,0 +1,36 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Defines name of a static method defined in a test class that will be invoked to produce a bootstrap config
+ * template to use intead of a default template.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD})
+public @interface BootstrapConfigTemplateMethod {
```
SammyVimes, line 31: This looks a bit like overkill imo

SammyVimes (approved):
OK overall, but I think the decision with annotation is questionable

modules/runner/src/integrationTest/java/org/apache/ignite/internal/BootstrapConfigTemplateMethod.java
rpuch: I simplified the test and removed the annotation and the surrounding mechanism

rpuch:
You are right, I simplified the code

SammyVimes (approved):
LGTM!

PR #1638, opened 2023-02-06
IGNITE-18714 Thread safe usage of InternalTableImpl#partitionMap

sergeyuttsel:
https://issues.apache.org/jira/browse/IGNITE-18714


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/rel/TableScanNodeExecutionTest.java
```diff
@@ -127,7 +128,11 @@ public IgniteDistribution distribution() {
 
         @Override
         public InternalTable table() {
-            return new TestInternalTableImpl(mock(ReplicaService.class));
+            ConcurrentMap<Integer, RaftGroupService> partMap = new ConcurrentSkipListMap<>();
```
sanpwc, line 131: What about copy on write approach?

sanpwc (changes requested):

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/rel/TableScanNodeExecutionTest.java
```diff
@@ -150,7 +149,6 @@ public TestInternalTableImpl(
             super(
                     "test",
                     UUID.randomUUID(),
-                    Int2ObjectMaps.singleton(0, mock(RaftGroupService.class)),
```
sanpwc, line 153: Why you've excluded given constructor param?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -1219,7 +1217,11 @@ public void updateInternalTableRaftGroupService(int p, RaftGroupService raftGrpS
         RaftGroupService oldSrvc;
 
         synchronized (updatePartMapMux) {
```
sanpwc, line 1219: Do we need explicit synchronization here?

sanpwc (approved):

PR #1639, opened 2023-02-06
IGNITE-18578 Add SSL support for ScaleQube, Java client, JDBC client

PakhomovAlexander:
Ssl support is implemented for Scale cube networking, IgniteTcpClient and jdbc.

https://issues.apache.org/jira/browse/IGNITE-18579
https://issues.apache.org/jira/browse/IGNITE-18578


modules/client-handler/src/main/java/org/apache/ignite/client/handler/configuration/ClientConnectorSslConfigurationSchema.java
```diff
@@ -0,0 +1,35 @@
...
+package org.apache.ignite.client.handler.configuration;
+
+import org.apache.ignite.configuration.annotation.Config;
+import org.apache.ignite.configuration.annotation.ConfigValue;
+import org.apache.ignite.configuration.annotation.Value;
+import org.apache.ignite.internal.network.configuration.KeyStoreConfigurationSchema;
+
+/** Client connector ssl configuration schema. */
+@Config
+public class ClientConnectorSslConfigurationSchema {
```
ptupitsyn, line 27: Looks like there is no way to configure client auth (https://netty.io/4.1/api/io/netty/handler/ssl/ClientAuth.html), let's add this too. Should be disabled by default.

modules/client-handler/src/integrationTest/java/org/apache/ignite/client/handler/ItSslClientHandlerTest.java
```diff
@@ -0,0 +1,144 @@
...
+            unpacker.skipValue(extensionsLen);
+
+            assertArrayEquals(MAGIC, magic);
+            assertEquals(44, len);
+            assertEquals(3, major);
+            assertEquals(0, minor);
+            assertEquals(0, patch);
+            assertEquals(0, idleTimeout);
+            assertEquals("id", nodeId);
+            assertEquals("consistent-id", nodeName);
```
ptupitsyn, line 141: We expect this code to fail, I think most of it is unnecessary. Magic header should be enough to trigger disconnect due to SSL handshake error.

ptupitsyn (changes requested):

modules/client-handler/src/integrationTest/java/org/apache/ignite/client/handler/ItSslClientHandlerTest.java
PakhomovAlexander: Agreed, will get rid of most assertions.

modules/client/src/main/java/org/apache/ignite/internal/client/IgniteClientConfigurationImpl.java
```diff
@@ -157,4 +162,9 @@ public long heartbeatTimeout() {
     public @Nullable RetryPolicy retryPolicy() {
         return retryPolicy;
     }
+
+    @Override
```
ptupitsyn, line 166: Missing inheritdoc, here and below.

modules/client/src/main/java/org/apache/ignite/internal/client/SslConfigurationImpl.java
```diff
@@ -0,0 +1,59 @@
...
+    private final boolean enabled;
+
+    private final String trustStoreType;
+
+    private final String trustStorePath;
+
+    private final String trustStorePassword;
+
+    /** Main constructor. */
+    public SslConfigurationImpl(boolean enabled, String trustStoreType, String trustStorePath, String trustStorePassword) {
```
ptupitsyn, line 33: Please add `@Nullable` annotations where applicable.

modules/client/src/main/java/org/apache/ignite/internal/client/io/netty/NettyClientConnectionMultiplexer.java
```diff
@@ -75,6 +86,36 @@ public void initChannel(SocketChannel ch) {
         }
     }
 
+    private static void setupSsl(SocketChannel ch, IgniteClientConfiguration clientCfg) {
+        if (clientCfg.sslConfiguration().enabled()) {
```
ptupitsyn, line 90: We could invert condition to reduce nesting: `if !enabled return`.

modules/client/src/main/java/org/apache/ignite/internal/client/io/netty/NettyClientConnectionMultiplexer.java
```diff
@@ -75,6 +86,36 @@ public void initChannel(SocketChannel ch) {
...
+                        TrustManagerFactory.getDefaultAlgorithm()
+                );
+                trustManagerFactory.init(store);
+
+                var context = SslContextBuilder.forClient()
+                        .trustManager(trustManagerFactory)
+                        .build();
+                ch.pipeline().addFirst("ssl", context.newHandler(ch.alloc()));
+            } catch (NoSuchAlgorithmException | KeyStoreException | CertificateException | IOException e) {
+                throw new IgniteClientConnectionException(CLIENT_SSL_CONFIGURATION_ERR, "Client ssl configuration error", e);
```
ptupitsyn, line 114:
```suggestion
                throw new IgniteClientConnectionException(CLIENT_SSL_CONFIGURATION_ERR, "Client SSL configuration error: " + e.getMessage(), e);
```

modules/network/src/main/java/org/apache/ignite/internal/network/netty/NettyClient.java
```diff
@@ -48,44 +51,50 @@ public class NettyClient {
...
-    private volatile Channel channel = null;
-
     /** Message listener. */
     private final Consumer<InNetworkObject> messageListener;
 
     /** Handshake manager. */
     private final HandshakeManager handshakeManager;
 
+    /** Ssl configuration. */
+    private final SslView configuration;
```
ptupitsyn, line 64:
```suggestion
    private final SslView sslConfiguration;
```

ptupitsyn (changes requested):

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerModule.java
```diff
@@ -198,17 +201,26 @@ protected void initChannel(Channel ch) {
...
-                                        igniteTransactions,
-                                        queryProcessor,
-                                        configuration,
-                                        igniteCompute,
-                                        clusterService,
-                                        sql,
-                                        clusterId));
+                        if (configuration.ssl().enabled()) {
+                            SslContext sslContext =  SslContextProvider.forServer(
+                                    configuration.ssl().keyStore(), ClientAuth.valueOf(configuration.ssl().clientAuth().toUpperCase())
```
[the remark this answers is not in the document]
PakhomovAlexander: Configuration schema does not allow to do that. 

```java
    @OneOf({"none", "optional", "required"})
    @Value(hasDefault = true)
    public final String clientAuth = "none";

File: modules/client/src/main/java/org/apache/ignite/client/SslConfiguration.java [hunk 9]
```diff
@@ -0,0 +1,42 @@
...
+public interface SslConfiguration {
+
+    /** If set to {@code true} then the SSL connection will be used to interact with Ignite 3 node. */
+    boolean enabled();
+
+    /** Client authentication configuration. */
+    ClientAuthConfiguration clientAuth();
+
+    /** Keystore configuration. */
+    @Nullable KeystoreConfiguration keyStore();
```
ptupitsyn, line 33: This is not consistent with `SslConfigurationBuilder`, where `keyStore` and `trustStore` properties are not grouped.

Let's remove `KeystoreConfiguration`. Using it for `TrustStore` also looks weird.

modules/client/src/main/java/org/apache/ignite/internal/client/io/netty/NettyClientConnectionMultiplexer.java
```diff
@@ -75,6 +90,74 @@ public void initChannel(SocketChannel ch) {
...
+        }
+
+        try {
+            var trustStore = clientCfg.sslConfiguration().trustStore();
+            if (trustStore == null) {
+                throw new IgniteClientSslException("SSL is enabled but no trust store is provided");
+            }
+            if (trustStore.path() == null) {
+                throw new IgniteClientSslException("SSL is enabled but no path to trust store is provided");
+            }
```
ptupitsyn, line 105: None of the properties in the client-side `SslConfiguration` should be mandatory. We should be able to use SSL without providing anything (for example, you don't need a key store or a trust store to load `https://github.com`).

ptupitsyn:
Superceded by #1662 , closing.

PR #1640, opened 2023-02-06
IGNITE-18581 Bootstrap Configuration: don't save bootstrap configuration in vault

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-18581


modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/configuration/ItConfigCommandTest.java
```diff
@@ -67,6 +67,31 @@ void addConfigKeyValue() {
...
+        );
+
+        // When read the updated cluster configuration
+        execute("node", "config", "show", "--node-url", NODE_URL);
+
+        // Then
+        assertAll(
+                this::assertExitCodeIsZero,
+                this::assertErrOutputIsEmpty,
+                () -> assertOutputContains("\"netClusterNodes\" : [ \"localhost:3344\", \"localhost:3345\" ]")
```
ibessonov, line 91: Do we render it as JSON? Is this convenient?

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationManager.java
```diff
@@ -84,12 +84,12 @@ public void stop() throws Exception {
...
      * Bootstrap configuration manager with customer user cfg.
      *
-     * @param hoconStr Customer configuration in hocon format.
+     * @param hoconPath Customer configuration in hocon format.
      * @throws InterruptedException If thread is interrupted during bootstrap.
      * @throws ExecutionException   If configuration update failed for some reason.
      */
-    public void bootstrap(@Language("HOCON") String hoconStr) throws InterruptedException, ExecutionException {
-        ConfigObject hoconCfg = ConfigFactory.parseString(hoconStr).root();
+    public void bootstrap(Path hoconPath) throws InterruptedException, ExecutionException {
```
ibessonov, line 91: `configPath` maybe? Hocon is a data format. When people say JsonPath or xmlPath, they don't mean the path to the file, they mean the path inside of the file I guess.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -359,4 +359,24 @@ public static class Network {
...
+         * Node configuration error group.
+         */
+        public static final ErrorGroup NODE_CONFIGURATION_ERR_GROUP = ErrorGroup.newGroup("NODECFG", 12);
+
+        /**
+         * Config read error.
+         */
+        public static final int CONFIG_READ_ERR = NODE_CONFIGURATION_ERR_GROUP.registerErrorCode(1);
+
+        public static final int CONFIG_FILE_CREATE_ERR = NODE_CONFIGURATION_ERR_GROUP.registerErrorCode(2);
```
ibessonov, line 377: These errors lack javadocs

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -359,4 +359,24 @@ public static class Network {
...
+
+        public static final int CONFIG_WRITE_ERR = NODE_CONFIGURATION_ERR_GROUP.registerErrorCode(3);
+
```
ibessonov, line 380:
```suggestion
```

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -261,7 +261,10 @@ public class IgniteImpl implements Ignite {
      * @param serviceProviderClassLoader The class loader to be used to load provider-configuration files and provider classes, or
      *         {@code null} if the system class loader (or, failing that the bootstrap class loader) is to be used.
      */
-    IgniteImpl(String name, Path workDir, @Nullable ClassLoader serviceProviderClassLoader) {
+    IgniteImpl(String name,
+            NodeBootstrapConfiguration configuration,
+            Path workDir,
+            @Nullable ClassLoader serviceProviderClassLoader) {
```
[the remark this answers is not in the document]
ibessonov: Looks prettier this way :)

modules/runner/src/main/java/org/apache/ignite/internal/configuration/NodeBootstrapConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+@FunctionalInterface
+public interface NodeBootstrapConfiguration {
+    /**
+     * Path to node configuration file.
+     *
+     * @return Path to node configuration file in HOCON format.
+     */
+    Path config();
+
+
```
ibessonov, line 44:
```suggestion

```

modules/runner/src/main/java/org/apache/ignite/internal/configuration/NodeBootstrapConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+     */
+    static NodeBootstrapConfiguration directFile(@NotNull Path configPath) {
+        return () -> configPath;
+    }
+
+    /**
+     * Return node bootstrap configuration with content from {@param is}.
+     *
+     * @param is configuration content.
+     * @param workDir dir for configuration file location.
```
ibessonov, line 59:
```suggestion
     * @param is Configuration content.
     * @param workDir Dir for configuration file location.
```

ibessonov: Please start parameters descriptions with capital letters

modules/runner/src/main/java/org/apache/ignite/internal/configuration/NodeBootstrapConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+        return new NodeBootstrapConfiguration() {
+
+            private final AtomicReference<Path> config = new AtomicReference<>();
+
+            @Override
+            public Path config() {
+                if (config.compareAndSet(null, createEmptyConfig(workDir))) {
+                    if (is != null) {
+                        try {
+                            Files.write(config.get(), is.readAllBytes());
```
ibessonov, line 72: This file write operation is not safe.
1. Operation can be interrupted in the middle of the process
2. You don't call fsync when you're done

There must be an example of safe file writing somewhere in the code, and you can probably reuse that code

modules/runner/src/main/java/org/apache/ignite/internal/configuration/NodeBootstrapConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+     * @param workDir Configuration file location.
+     * @return Node bootstrap configuration provider to empty config.
+     */
+    static NodeBootstrapConfiguration empty(Path workDir) {
+        return () -> createEmptyConfig(workDir);
+    }
+
+    private static Path createEmptyConfig(Path workDir) {
+        try {
+            Path config = workDir.resolve("ignite-config.conf");
```
ibessonov, line 109: What's a `*.conf` by the way? Is this a common file extension for configuration files?

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -0,0 +1,224 @@
...
+import org.apache.ignite.internal.util.IgniteUtils;
+
+/**
+ * Implementation of {@link ConfigurationStorage} based on local file configuration storage.
+ */
+public class LocalFileConfigurationStorage implements ConfigurationStorage {
+    private static final IgniteLogger LOG = Loggers.forClass(LocalFileConfigurationStorage.class);
+
+    private final Path configPath;
+    private final Map<String, Serializable> lastState = new HashMap<>();
```
ibessonov, line 66: I'd rename it to `latest`, but it's up to you

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -0,0 +1,224 @@
...
+
+    private final AtomicReference<ConfigurationStorageListener> lsnrRef = new AtomicReference<>();
```
ibessonov, line 68: Why did you shorten this name?

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -0,0 +1,224 @@
...
+
+    public LocalFileConfigurationStorage(NodeBootstrapConfiguration configuration) {
+        this.configPath = configuration.config();
+        ensureConfigFile();
+    }
+
+    @Override
+    public CompletableFuture<Data> readDataOnRecovery() {
+        return readAllLatest("").thenApply(stringMap -> new Data(stringMap, 0));
+    }
```
ibessonov, line 82: Shouldn't it be empty during the start?

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -0,0 +1,224 @@
...
+
+    @Override
+    public CompletableFuture<Map<String, ? extends Serializable>> readAllLatest(String prefix) {
+        ensureConfigFile();
+        Map<String, Serializable> map = new HashMap<>();
+        if (!lastState.isEmpty()) {
+            for (Entry<String, ConfigValue> e : parseConfigOptions().entrySet()) {
```
ibessonov, line 89: There must be a huge TODO somewhere, telling that current code will break once we introduce a named list configuration to the set of node configurations.

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -0,0 +1,224 @@
...
+                String key = e.getKey();
+                if (key.startsWith(prefix)) {
+                    ConfigValue value = e.getValue();
+                    if (value.valueType() == ConfigValueType.LIST) {
+                        map.put(key, toArray(lastState.get(key).getClass(), (ConfigList) value));
```
ibessonov, line 94: There's no validation. Why don't we read values from `lastState`? It would be so much safer

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    @Override
+    public CompletableFuture<Serializable> readLatest(String key) {
+        ensureConfigFile();
+        return CompletableFuture.completedFuture((Serializable) parseConfigOptions().getValue(key).unwrapped());
+    }
+
+    @Override
+    public CompletableFuture<Boolean> write(Map<String, ? extends Serializable> newValues, long ver) {
```
ibessonov, line 111: Is this method thread-safe? I guess not?
Please explain why. If it doesn't need to be thread-safe, then please add a comment with justification

ibessonov: `ver` is completely ignored, as far as I see

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    @Override
+    public ConfigurationType type() {
+        return ConfigurationType.LOCAL;
+    }
+
+    @Override
+    public CompletableFuture<Long> lastRevision() {
+        return CompletableFuture.completedFuture(0L);
```
ibessonov, line 139: Wow, why? Class is clearly not completed.
Too bad that `org.apache.ignite.internal.configuration.storage.ConfigurationStorageTest` is not sophisticated enough, that's on us...

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    @Override
+    public void close() {
+        IgniteUtils.shutdownAndAwaitTermination(threadPool, 10, TimeUnit.SECONDS);
+
+        futureTracker.cancelInFlightFutures();
+    }
+
+    private String toString(Map<String, ? extends Serializable> values) {
```
ibessonov, line 154: Please add javadoc. It may not be clear at first glance that this method is used before saving configuration to file

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -0,0 +1,224 @@
...
+        Config other = ConfigFactory.parseMap(map);
+        Config newConfig = parseConfigOptions().withFallback(other).resolve();
+        return newConfig.isEmpty()
+                ? ""
+                : newConfig.root().render(ConfigRenderOptions.defaults().setJson(false));
+    }
+
+    private void saveValues(Map<String, ? extends Serializable> values) {
+        try {
+            Files.write(configPath, toString(values).getBytes(StandardCharsets.UTF_8));
```
ibessonov, line 171: Not safe file writing, once again. At first, you should write data in another file, and only then you have a right to corrupt/delete old file, because new one is already on disk.
On top of that, we need a "recovery" procedure that would delete tmp file or maybe rename it to proper config file if it's already missing

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+    }
+
+    private Config parseConfigOptions() {
+        return ConfigFactory.parseFile(
+                configPath.toFile(),
+                ConfigParseOptions.defaults().setAllowMissing(false));
+    }
+
+    private void ensureConfigFile() {
```
ibessonov, line 184: Name of the method doesn't reflect what it does. Please document it at least

modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/ConfigurationStorageTest.java
```diff
@@ -78,4 +78,12 @@ public void testReadAllLatest() {
 
         assertThat(latestData, willBe(anEmptyMap()));
     }
+
+    @Test
+    public void testRealConfig() {
+        Map<String, Serializable> data = Map.of("network.port", 3344,
+                "network.portRange", 5, "network.nodeFinder.netClusterNode", new String[] {"localhost:3344", "localhost:3345"});
+
+        storage.write(data, 0);
```
ibessonov, line 87: Don't we need to check something after writing data?
What's up with the test name? What do you mean by "real config"?

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/configuration/ItConfigCommandTest.java
ibessonov: Ok

modules/runner/src/main/java/org/apache/ignite/internal/configuration/NodeBootstrapConfiguration.java
ibessonov: If this code is temporary, then please add TODO with JIRA link

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
ibessonov: I only mean code simplicity. If it's always empty, why "reading" it?

modules/runner/src/main/java/org/apache/ignite/internal/configuration/NodeBootstrapConfiguration.java
```diff
@@ -0,0 +1,118 @@
...
+        return new NodeBootstrapConfiguration() {
+
+            private final AtomicReference<Path> config = new AtomicReference<>();
+
+            @Override
+            public Path configPath() {
+                if (config.compareAndSet(null, createEmptyConfig(workDir))) {
+                    if (is != null) {
+                        try {
+                            Files.write(config.get(), is.readAllBytes(), StandardOpenOption.DSYNC);
```
ibessonov, line 71: I believe you mentioned that this code is for tests only. Why not moving it to test fixtures?

ibessonov: If it's not for tests, then I don't see writing to a temp file

modules/runner/src/main/java/org/apache/ignite/internal/configuration/NodeBootstrapConfiguration.java
```diff
@@ -0,0 +1,118 @@
...
+                        } catch (IOException e) {
+                            throw new NodeConfigReadException("Failed to read config input stream.", e);
```
ibessonov, line 73: What if it was an error during the write? They're both in the same "try" section

modules/runner/src/main/java/org/apache/ignite/internal/configuration/NodeBootstrapConfiguration.java
```diff
@@ -0,0 +1,118 @@
...
+                        }
+                    }
+                }
+                return config.get();
```
ibessonov, line 77: I also see that you don't usually have empty lines between statements. What's the reason? I'd prefer Ignite code to look more unified

modules/runner/src/main/java/org/apache/ignite/internal/configuration/NodeBootstrapConfiguration.java
```diff
@@ -0,0 +1,118 @@
...
+     *
+     * @param plainConf Configuration content.
+     * @param workDir Dir for configuration file location.
+     * @return Node bootstrap configuration with lazy config file creation.
+     */
+    static NodeBootstrapConfiguration string(@Nullable @Language("HOCON") String plainConf, Path workDir) {
+        return inputStream(plainConf != null
+                ? new ByteArrayInputStream(plainConf.getBytes(StandardCharsets.UTF_8))
+                : null,
+                workDir);
```
ibessonov, line 93: Looks like an invalid padding

modules/runner/src/main/java/org/apache/ignite/internal/configuration/NodeBootstrapConfiguration.java
ibessonov: Ok. Can we have a constant for this string?

modules/runner/src/main/java/org/apache/ignite/internal/configuration/NodeBootstrapConfiguration.java
```diff
@@ -0,0 +1,118 @@
...
+    private static Path createEmptyConfig(Path workDir) {
+        try {
+            Path config = workDir.resolve("ignite-config.conf");
+            File file = config.toFile();
+            if (!file.exists()) {
+                file.createNewFile();
+            }
+            return config;
+        } catch (IOException e) {
+            throw new NodeConfigCreateException("Failed to create temp conf file.", e);
```
ibessonov, line 115: Why is it a temp file? Name suggests otherwise. And I guess no one deletes it afterwards

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -0,0 +1,241 @@
...
+     * Latest state of last applied configuration.
+     */
+    private final Map<String, Serializable> latest = new ConcurrentHashMap<>();
+
+    /**
+     *  Configuration changes listener.
+     *  */
+    private final AtomicReference<ConfigurationStorageListener> lsnrRef = new AtomicReference<>();
+
+    private final ExecutorService threadPool = Executors.newFixedThreadPool(4, new NamedThreadFactory("loc-cfg-file", LOG));
```
ibessonov, line 83: Maybe 2 threads would be enough, I see no realistic scenario where we would need 4 at the same time

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -0,0 +1,241 @@
...
+    private long lastRevision = 0L;
+
+    /**
+     * Constructor.
+     *
+     * @param configuration Node bootstrap configuration.
+     */
+    public LocalFileConfigurationStorage(NodeBootstrapConfiguration configuration) {
+        this.configPath = configuration.configPath();
+        tempConfigPath = new File(configPath.toFile().getAbsolutePath() + ".storage").toPath();
```
ibessonov, line 96: Why not `.tmp`?

ibessonov, line 96:
```suggestion
        tempConfigPath = configPath.resolveSibling(configPath.getFileName() + ".storage");
```

ibessonov: I think this should work. You don't need to use `File` when you have `Path`, it's a legacy API

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -0,0 +1,241 @@
...
+    public void close() {
+        IgniteUtils.shutdownAndAwaitTermination(threadPool, 10, TimeUnit.SECONDS);
+
+        futureTracker.cancelInFlightFutures();
+    }
+
+    private void saveValues(Map<String, ? extends Serializable> values) {
+        try {
+            String s = renderHoconString(values);
+            Files.write(tempConfigPath, s.getBytes(StandardCharsets.UTF_8), StandardOpenOption.DSYNC);
```
ibessonov, line 187: Why `DSYNC` and not `SYNC`?

modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorageTest.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Random;
+import org.apache.ignite.internal.configuration.NodeBootstrapConfiguration;
+import org.junit.jupiter.api.io.TempDir;
+
+/**
+ * Tests for the {@link LocalFileConfigurationStorage}.
+ */
+public class LocalFileConfigurationStorageTest extends ConfigurationStorageTest {
+
+    @TempDir
```
ibessonov, line 30: I wonder why you don't use `@WorkDir` (or whatever that name is, it's our own path provider)

modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorageTest.java
```diff
@@ -0,0 +1,38 @@
...
+    private Path tmpDir;
+
+    @Override
+    public ConfigurationStorage getStorage() {
+        Path resolve = tmpDir.resolve("file" + new Random().nextInt());
```
ibessonov, line 35: And why does it have to be a random name. Are you trying to prevent collisions? Looks suspicious.

PR #1641, opened 2023-02-06
IGNITE-18716 Replace withTimeout and readOnly in IgniteTransactions with TransactionOptions

ptupitsyn:
`IgniteTransactions` interface has `withTimeout` and `readOnly` methods, which return a new instance of the interface that behaves differently. This approach proved to be not very good in Ignite 2.x (`withKeepBinary` etc):

* Not clear if a given instance has some of those things enabled.
* Not possible to go back from readOnly to normal mode.
* Allocates one or more objects on heap (if we need to set multiple things, some of those objects go straight to garbage).

Introduce `TransactionOptions` class instead, and pass it to `begin`, `beginAsync`, `runInTransaction`, etc.


modules/api/src/main/java/org/apache/ignite/tx/TransactionOptions.java
```diff
@@ -0,0 +1,90 @@
...
+        if (readOnly && timeoutMillis != 0) {
+            throw illegalStateException();
+        }
+
+        this.readOnly = readOnly;
+
+        return this;
+    }
+
+    private static IllegalStateException illegalStateException() {
```
ascherbakoff, line 87: InvalidArgumentException fits better here

modules/api/src/main/java/org/apache/ignite/tx/TransactionOptions.java
```diff
@@ -0,0 +1,90 @@
...
+     *
+     * <p>Read-only transactions provide a snapshot view of data at a certain point in time.
+     * They are lock-free and perform better than normal transactions, but do not permit data modifications.
+     *
+     * @param readOnly Whether a read-only transaction should be used.
+     *
+     * @return {@code this} for chaining.
+     */
+    public TransactionOptions readOnly(boolean readOnly) {
+        if (readOnly && timeoutMillis != 0) {
```
ascherbakoff, line 78: Copy&paste code. Better to implement the method validateArguments() and call it after the property is set.

modules/api/src/main/java/org/apache/ignite/tx/TransactionOptions.java
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Sets transaction timeout, in milliseconds.
+     *
+     * @param timeoutMillis Transaction timeout, in milliseconds.
+     * @return {@code this} for chaining.
+     */
+    public TransactionOptions timeoutMillis(long timeoutMillis) {
+        if (readOnly && timeoutMillis != 0) {
+            throw illegalStateException();
```
sanpwc, line 47: Seems that I've missed that in initial discussion. Why do we consider timeout is an inappropriate attribute for readOnly transaction?

ptupitsyn: @ascherbakoff can you please comment on this?

ascherbakoff: Because it doesn't acquire any locks and can't be blocked in any way. But I recall the JDBC has query timeout, and possibly I'm wrong here. This timeout can be used to cancel misbehaving query from a client side, same as Ignite2 does.

ptupitsyn: I'm not sure if JDBC and `IgniteTransactions` interface can work together.

In any case, transaction timeouts are not implemented yet, should we postpone this decision and remove the exception?

modules/api/src/main/java/org/apache/ignite/tx/TransactionOptions.java
ptupitsyn: > call it after the property is set

At this point the instance already represents an invalid state. So it is possible to catch the exception, ignore it, and pass invalid options to the tx API.

modules/api/src/main/java/org/apache/ignite/tx/TransactionOptions.java
ptupitsyn: Fixed

modules/api/src/main/java/org/apache/ignite/tx/TransactionOptions.java
sanpwc: Yep, I'd remove the exception.

ptupitsyn: Done

modules/api/src/main/java/org/apache/ignite/tx/TransactionOptions.java
ptupitsyn: As discussed below, let's keep it simple. I've removed the exception for now.

sanpwc (approved):

ascherbakoff (approved):

PR #1648, opened 2023-02-07
IGNITE-18639 Implement distributed onLeaderElected callback within topology aware raft client

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-18639

### DIFF
--- a/modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
+++ b/modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
@@ -45,8 +45,8 @@ import org.apache.ignite.network.ClusterNode;
 import org.apache.ignite.network.ClusterService;
 import org.apache.ignite.raft.jraft.RaftMessageGroup;
 import org.apache.ignite.raft.jraft.RaftMessagesFactory;
+import org.apache.ignite.raft.jraft.rpc.CliRequests.LeaderChangeNotification;
 import org.apache.ignite.raft.jraft.rpc.CliRequests.SubscriptionLeaderChangeRequest;
-import org.apache.ignite.raft.jraft.rpc.CliRequests.SubscriptionLeaderChangeResponse;
 import org.jetbrains.annotations.Nullable;
 
 /**
@@ -105,8 +105,8 @@ public class TopologyAwareRaftGroupService implements RaftGroupService {
         this.serverEventHandler = new ServerEventHandler();
 
         cluster.messagingService().addMessageHandler(RaftMessageGroup.class, (message, senderConsistentId, correlationId) -> {
-            if (message instanceof SubscriptionLeaderChangeResponse) {
-                var msg = (SubscriptionLeaderChangeResponse) message;
+            if (message instanceof LeaderChangeNotification) {
+                var msg = (LeaderChangeNotification) message;
 
                 serverEventHandler.onLeaderElected(clusterService.topologyService().getByConsistentId(senderConsistentId), msg.term());
             }
@@ -124,8 +124,8 @@ public class TopologyAwareRaftGroupService implements RaftGroupService {
                                 .groupId(groupId())
                                 .subscribe(true)
                                 .build())
-                                .thenComposeAsync(couldLeaderChange -> {
-                                    if (couldLeaderChange) {
+                                .thenComposeAsync(subscribed -> {
+                                    if (subscribed) {
                                         return refreshAndGetLeaderWithTerm()
                                                 .thenAcceptAsync(leaderWithTerm -> {
                                                     if (leaderWithTerm.leader() != null
@@ -176,7 +176,7 @@ public class TopologyAwareRaftGroupService implements RaftGroupService {
      *
      * @param node Node.
      * @param msg  Subscribe message.
-     * @return A future that will complete when the message is sent.
+     * @return A future that completes with true when the message sent and false value when the node left the cluster.
      */
     private CompletableFuture<Boolean> sendSubscribeMessage(ClusterNode node, SubscriptionLeaderChangeRequest msg) {
         var msgSendFut = new CompletableFuture<Boolean>();
@@ -189,9 +189,10 @@ public class TopologyAwareRaftGroupService implements RaftGroupService {
     /**
      * Tries to send a subscribe message until the node leaves of the cluster.
      *
-     * @param node       Node.
-     * @param msg        Subscribe message to send.
-     * @param msgSendFut Future that will completed when the message is send or an issue prevent that.
+     * @param node Node.
+     * @param msg Subscribe message to send.
+     * @param msgSendFut Future that completes with true when the message sent and with false when the node left topology and  cannot get a
+     * cluster.
      */
     private void sendWithRetry(ClusterNode node, SubscriptionLeaderChangeRequest msg, CompletableFuture<Boolean> msgSendFut) {
         clusterService.messagingService().invoke(node, msg, raftConfiguration.responseTimeout().value()).whenCompleteAsync((unused, th) -> {
@@ -199,7 +200,7 @@ public class TopologyAwareRaftGroupService implements RaftGroupService {
                 if (recoverable(th)) {
                     logicalTopologyService.logicalTopologyOnLeader().whenCompleteAsync((logicalTopologySnapshot, topologyGetIssue) -> {
                         if (topologyGetIssue != null) {
-                            LOG.error("Actual logical topology snapshot was not get.", topologyGetIssue);
+                            LOG.error("Actual logical topology snapshot was not got.", topologyGetIssue);
 
                             msgSendFut.completeExceptionally(topologyGetIssue);
 
@@ -209,8 +210,8 @@ public class TopologyAwareRaftGroupService implements RaftGroupService {
                         if (logicalTopologySnapshot.nodes().contains(node)) {
                             sendWithRetry(node, msg, msgSendFut);
                         } else {
-                            LOG.info("Could not subscribe to leader update from a specific node, because the node had left of the cluster"
-                                    + " [node={}]", node);
+                            LOG.info("Could not subscribe to leader update from a specific node, because the node had left from the"
+                                    + " cluster [node={}]", node);
 
                             msgSendFut.complete(false);
                         }
@@ -248,6 +249,8 @@ public class TopologyAwareRaftGroupService implements RaftGroupService {
      * @param callback Callback closure.
      */
     public void subscribeLeader(Consumer<ClusterNode> callback) {
+        assert !serverEventHandler.isSubscribed() : "The node already subscribed";
+
         int peers = peers().size();
 
         var futs = new CompletableFuture[peers];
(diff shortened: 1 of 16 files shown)


modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
```diff
@@ -0,0 +1,443 @@
...
+     * @param msg        Subscribe message to send.
+     * @param msgSendFut Future that will completed when the message is send or an issue prevent that.
+     */
+    private void sendWithRetry(ClusterNode node, SubscriptionLeaderChangeRequest msg, CompletableFuture<Boolean> msgSendFut) {
+        clusterService.messagingService().invoke(node, msg, raftConfiguration.responseTimeout().value()).whenCompleteAsync((unused, th) -> {
+            if (th != null) {
+                if (recoverable(th)) {
+                    logicalTopologyService.logicalTopologyOnLeader().whenCompleteAsync((logicalTopologySnapshot, topologyGetIssue) -> {
+                        if (topologyGetIssue != null) {
+                            LOG.error("Actual logical topology snapshot was not get.", topologyGetIssue);
```
denis-chudov, line 202:
```suggestion
                            LOG.error("Actual logical topology snapshot was not got.", topologyGetIssue);
```

modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
```diff
@@ -0,0 +1,443 @@
...
+
+                            msgSendFut.completeExceptionally(topologyGetIssue);
+
+                            return;
+                        }
+
+                        if (logicalTopologySnapshot.nodes().contains(node)) {
+                            sendWithRetry(node, msg, msgSendFut);
+                        } else {
+                            LOG.info("Could not subscribe to leader update from a specific node, because the node had left of the cluster"
```
denis-chudov, line 212:
```suggestion
                            LOG.info("Could not subscribe to leader update from a specific node, because the node had left from the cluster"
```

modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
```diff
@@ -0,0 +1,443 @@
...
+     * Tries to send a subscribe message until the node leaves of the cluster.
+     *
+     * @param node       Node.
+     * @param msg        Subscribe message to send.
+     * @param msgSendFut Future that will completed when the message is send or an issue prevent that.
+     */
+    private void sendWithRetry(ClusterNode node, SubscriptionLeaderChangeRequest msg, CompletableFuture<Boolean> msgSendFut) {
+        clusterService.messagingService().invoke(node, msg, raftConfiguration.responseTimeout().value()).whenCompleteAsync((unused, th) -> {
+            if (th != null) {
+                if (recoverable(th)) {
```
denis-chudov, line 199: seems that this throwable should be added as cause exception if the future is completed with another one.

modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
```diff
@@ -0,0 +1,443 @@
...
+
+        return msgSendFut;
+    }
+
+    /**
+     * Tries to send a subscribe message until the node leaves of the cluster.
+     *
+     * @param node       Node.
+     * @param msg        Subscribe message to send.
+     * @param msgSendFut Future that will completed when the message is send or an issue prevent that.
```
denis-chudov, line 194: Am I right that that future resul is true when message is sent, false when the recipient node left topology and completes exceptionally in other cases? shouldn't we describe it in javadoc?

modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
```diff
@@ -0,0 +1,443 @@
...
+                    if (serverEventHandler.isSubscribed() && appearedNode.name().equals(peer.consistentId())) {
+                        LOG.info("New peer will be sending a leader elected notification [grpId={}, consistentId={}]", groupId(),
+                                peer.consistentId());
+
+                        sendSubscribeMessage(appearedNode, TopologyAwareRaftGroupService.this.factory.subscriptionLeaderChangeRequest()
+                                .groupId(groupId())
+                                .subscribe(true)
+                                .build())
+                                .thenComposeAsync(couldLeaderChange -> {
+                                    if (couldLeaderChange) {
```
denis-chudov, line 128: maybe `topologyChanged`?

modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
```diff
@@ -0,0 +1,443 @@
...
+
+        return t instanceof TimeoutException || t instanceof IOException;
+    }
+
+    /**
+     * Assigns a closure to call when a leader will is elected.
+     *
+     * @param callback Callback closure.
+     */
+    public void subscribeLeader(Consumer<ClusterNode> callback) {
```
denis-chudov, line 250: Every call replaces previous callback, is it correct?

vldpyatkov: Do you mean need to check that the service have not already subscribed before the method is invoked?

modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
vldpyatkov: There is no obligation the topology is changed. I'm going to change the name to more obvious.
For example: `subscribed`

modules/raft/build.gradle
```diff
@@ -76,4 +76,8 @@ dependencies {
     integrationTestImplementation project(':ignite-raft-api')
     integrationTestImplementation libs.jetbrains.annotations
     integrationTestImplementation libs.dropwizard.metrics
+
+    compileJava {
+        println "outputDir "
```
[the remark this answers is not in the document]
vldpyatkov: Done

modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/RaftServiceEventListener.java
```diff
@@ -0,0 +1,126 @@
...
+
+            Set<Consumer<Long>> nodeActions = nosesSubscriptions.get(clusterNode);
+
+            assert !CollectionUtils.nullOrEmpty(nodeActions);
+
+            Set<Consumer<Long>> grpNodeActions = new HashSet<>(actions);
+
+            grpNodeActions.retainAll(nodeActions);
+
+            assert grpNodeActions.size() == 1 : "Node is not subscribed [node=" + clusterNode + "groupId=" + groupId + ']';
```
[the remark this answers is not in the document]
vldpyatkov: Only when the action subscribes to different replication group.
A node can subscribe only one action to a single replication group.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/CliRequests.java
```diff
@@ -187,4 +189,26 @@ public interface LearnersOpResponse extends Message {
...
+        */
+        boolean subscribe();
+    }
+
+    @Transferable(value = RpcClientMessageGroup.SUBSCRIPTION_LEADER_CHANGE_RESPONSE_ACKNOWLEDGE)
+    public interface SubscriptionLeaderChangeRequestAcknowledge extends Message {
+    }
+
+    @Transferable(value = RpcClientMessageGroup.SUBSCRIPTION_LEADER_CHANGE_RESPONSE)
+    public interface SubscriptionLeaderChangeResponse extends Message {
```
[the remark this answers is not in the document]
vldpyatkov: ok

modules/raft/build.gradle
```diff
@@ -76,4 +76,8 @@ dependencies {
...
+    }
```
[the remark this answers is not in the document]
vldpyatkov: Removed.

modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/RaftServiceEventListener.java
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * RAFT server event listener.
+ */
+public class RaftServiceEventListener {
+    /** Callbacks by group ids. */
+    private ConcurrentHashMap<ReplicationGroupId, Set<Consumer<Long>>> subscriptions = new ConcurrentHashMap<>();
+
+    /** Callbacks by cluster nodes. */
+    private HashMap<ClusterNode, Set<Consumer<Long>>> nosesSubscriptions = new HashMap<>();
```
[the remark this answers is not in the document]
vldpyatkov: Many subscriptions are possible from a particular node. I will call it `nosesSubscriptions`

modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
vldpyatkov: Added assert here.
assert !serverEventHandler.isSubscribed() : "The node already subscribed";

modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    /**
+     * Sends a subscribe message to a specific node of the cluster.
+     *
+     * @param node Node.
+     * @param msg  Subscribe message.
+     * @return A future that will complete when the message is sent.
+     */
+    private CompletableFuture<Boolean> sendSubscribeMessage(ClusterNode node, SubscriptionLeaderChangeRequest msg) {
```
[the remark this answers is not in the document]
vldpyatkov: Added javadoc.

modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
vldpyatkov: Added description.

modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
vldpyatkov: I think the loop will be internally in network service layer in the future. They handle it more correctly, but currently I think the node which is catching an exception due to cannot get a topology has no cause except the particular exception.
Moreover, we do not an exception when we retry send a message and node still available in logical topology. Why do we have to write an exception into cause if we cannot get a logical topology?.

denis-chudov: maybe suppressed then?

PR #1649, opened 2023-02-07
IGNITE-18035 C++: support all native table types

isapego:
Done:
- Moved `big_decimal` and `big_integer` types from `schema` lib to `common`;
- Implemented `bit_array` type;
- Added tests for `bit_array` type;
- Added support for `big_decimal`, `big_integer`, `date`, `time`, `date_time`, `timestamp`, `bit_array` types to table and SQL APIs;
- Added all types table test;


modules/platforms/cpp/ignite/client/detail/table/table_impl.cpp
```diff
@@ -233,7 +268,7 @@ ignite_tuple read_tuple(protocol::reader &reader, const schema *sch, const ignit
         if (i < sch->key_column_count) {
             res.set(column.name, key.get(column.name));
         } else {
-            res.set(column.name, read_next_column(parser, column.type));
+            res.set(column.name, read_next_column(parser, column.type, column.scale));
```
ptupitsyn, line 271: Do we have an integration test to check that decimal scale handling in C++ client corresponds to the server? 
* Check roundtrip
* Run a Compute job to compare string representations

ptupitsyn (approved):

modules/platforms/cpp/ignite/client/detail/table/table_impl.cpp
isapego: Roundtrip is checked. I've actually found a bug this way. Compute is not yet supported in C++. I will add an SQL test - something like `SELECT CAST('12345.6789' AS DECIMAL(10, 4))`.

ptupitsyn (approved):

PR #1651, opened 2023-02-08
IGNITE-18492 SQL. Inconsistent behavior of LENGTH limit for CHAR and VARCHAR data types

zstan:
with validation and trimming (current PR):
```
Benchmark                       (clusterSize)  (fsync)  (partitionCount)  Mode  Cnt    Score    Error  Units
InsertBenchmark.sqlInsertMulti              1    false                 1  avgt   20  400.255 ± 23.553  us/op

Benchmark                  (clusterSize)  (fsync)  (partitionCount)  Mode  Cnt    Score    Error  Units
InsertBenchmark.sqlInsert              1    false                 1  avgt   20  237.118 ± 18.193  us/op
```

without:
```
Benchmark                       (clusterSize)  (fsync)  (partitionCount)  Mode  Cnt    Score    Error  Units
InsertBenchmark.sqlInsertMulti              1    false                 1  avgt   20  393.089 ± 25.825  us/op

Benchmark                  (clusterSize)  (fsync)  (partitionCount)  Mode  Cnt    Score    Error  Units
InsertBenchmark.sqlInsert              1    false                 1  avgt   20  234.271 ± 11.951  us/op
```

### DIFF
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/KeyValueModifyPlan.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/KeyValueModifyPlan.java
@@ -58,7 +58,7 @@ public class KeyValueModifyPlan implements ExplainablePlan, ExecutablePlan {
     private final IgniteKeyValueModify modifyNode;
     private final ResultSetMetadata meta;
     private final ParameterMetadata parameterMetadata;
-    private Supplier<RowSchema> schemaSupplier;
+    private RowSchema rowSchema;
 
     KeyValueModifyPlan(
             PlanId id,
@@ -140,9 +140,13 @@ public class KeyValueModifyPlan implements ExplainablePlan, ExecutablePlan {
 
                     RelDataType rowType = table().getRowType(ctx.getTypeFactory());
 
-                    if (schemaSupplier == null) {
-                        schemaSupplier = () -> rowSchemaFromRelTypes(RelOptUtil.getFieldTypeList(rowType));
-                    }
+                    Supplier<RowSchema> schemaSupplier = () -> {
+                        if (rowSchema != null) {
+                            return rowSchema;
+                        }
+                        rowSchema = rowSchemaFromRelTypes(RelOptUtil.getFieldTypeList(rowType));
+                        return rowSchema;
+                    };
 
                     row = validateCharactersOverflowAndTrimIfPossible(rowType, ctx.rowHandler(), row, schemaSupplier);
 
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/KeyValueModifyPlan.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/KeyValueModifyPlan.java
@@ -137,7 +137,7 @@ public class KeyValueModifyPlan implements ExplainablePlan, ExecutablePlan {
 
                     RelDataType rowType = table().getRowType(ctx.getTypeFactory());
 
-                    validateCharactersOverflow(rowType, row, ctx.rowHandler());
+                    validateCharactersOverflow(rowType, ctx.rowHandler(), row);
 
                     return updatableTable.insert(
                             tx, ctx, rowSupplier.get()
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/KeyValueModifyPlan.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/KeyValueModifyPlan.java
@@ -17,16 +17,12 @@
 
 package org.apache.ignite.internal.sql.engine.prepare;
 
-import static org.apache.ignite.internal.sql.engine.util.TypeUtils.rowSchemaFromRelTypes;
-import static org.apache.ignite.internal.sql.engine.util.TypeUtils.validateCharactersOverflowAndTrimIfPossible;
-
 import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 import java.util.function.Supplier;
 import org.apache.calcite.plan.RelOptUtil;
-import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.sql.SqlExplainLevel;
 import org.apache.ignite.internal.sql.engine.InternalSqlRow;
@@ -37,7 +33,6 @@ import org.apache.ignite.internal.sql.engine.exec.ExecutablePlan;
 import org.apache.ignite.internal.sql.engine.exec.ExecutableTableRegistry;
 import org.apache.ignite.internal.sql.engine.exec.ExecutionContext;
 import org.apache.ignite.internal.sql.engine.exec.UpdatableTable;
-import org.apache.ignite.internal.sql.engine.exec.row.RowSchema;
 import org.apache.ignite.internal.sql.engine.rel.IgniteKeyValueModify;
 import org.apache.ignite.internal.sql.engine.rel.IgniteRel;
 import org.apache.ignite.internal.sql.engine.schema.IgniteTable;
@@ -58,7 +53,6 @@ public class KeyValueModifyPlan implements ExplainablePlan, ExecutablePlan {
     private final IgniteKeyValueModify modifyNode;
     private final ResultSetMetadata meta;
     private final ParameterMetadata parameterMetadata;
-    private RowSchema rowSchema;
 
     KeyValueModifyPlan(
             PlanId id,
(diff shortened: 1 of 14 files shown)


zstan:
@lowka change approach, wdyt?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDataTypesTest.java
```diff
@@ -86,6 +87,23 @@ public void testUnicodeStrings() {
...
+        assertThrows(CalciteContextException.class, () -> sql("insert into limitedChar(pk, f2) values (1, 'aaaadddd')"));
+        assertThrows(CalciteContextException.class, () -> sql("insert into limitedChar values (2, 'aaaaabbbbb', 'aaaadddd')"));
+        assertThrows(CalciteContextException.class, () -> sql("insert into limitedChar values (3, '我叫什么名字', '我叫什么名字')"));
+        sql("insert into limitedChar values (4, ' aaa    ', ' bbb     ')");
+
+        List<List<Object>> res = sql("select f1, f2 from limitedChar ORDER BY pk");
+
+        List<Object> r = res.get(0);
+
+        assertEquals(" aaa ", r.get(0));
```
lowka, line 103: Insert '__aaa___'. Retrieve '_aaa_'. Do we still expect data truncation to occur?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -158,6 +163,51 @@ public void validateInsert(SqlInsert insert) {
...
+
+        for (int i = 0; i < srcFields.size(); ++i) {
+            RelDataType srcType = srcFields.get(i).getValue();
+            RelDataType dstType = dstFields.get(i).getValue();
+
+            if (SqlTypeUtil.isString(dstType) && srcType.getPrecision() != RelDataType.PRECISION_NOT_SPECIFIED) {
+                String val = srcFields.get(i).getKey();
+                String src = brace.matcher(val).replaceAll("");
+                // it`s correct here to cut probably more.
+                src = src.stripTrailing();
```
lowka, line 201: We can only replace "'" at the beginning and at the end of a literal.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDataTypesTest.java
lowka: We also need to check the case, when an input is a dynamic parameter and a target column's type has precision that is less then a value of the dynamic parameter.

lowka: > Insert '...aaa..... Retrieve '.aaa.'. Do we still expect data truncation to occur?
It seems that removing trailing blank characters is a correct behaviour.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDataTypesTest.java
```diff
@@ -86,6 +87,23 @@ public void testUnicodeStrings() {
         }
     }
 
+    /** Test correctness of char types limitation. */
+    @Test
+    public void charLimitation() {
+        sql("create table limitedChar (pk int primary key, f1 VARCHAR(5), f2 CHAR(4))");
+        assertThrows(CalciteContextException.class, () -> sql("insert into limitedChar(pk, f2) values (1, 'aaaadddd')"));
+        assertThrows(CalciteContextException.class, () -> sql("insert into limitedChar values (2, 'aaaaabbbbb', 'aaaadddd')"));
```
AMashenkov, line 95:
```suggestion
        assertThrows(CalciteContextException.class, () -> sql("insert into limitedChar(pk, f1) values (2, 'aaaaabbbbb')"));
```
Let's check CHAR and VARCHAR columns independently.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDataTypesTest.java
```diff
@@ -86,6 +87,23 @@ public void testUnicodeStrings() {
         }
     }
 
+    /** Test correctness of char types limitation. */
+    @Test
+    public void charLimitation() {
+        sql("create table limitedChar (pk int primary key, f1 VARCHAR(5), f2 CHAR(4))");
+        assertThrows(CalciteContextException.class, () -> sql("insert into limitedChar(pk, f2) values (1, 'aaaadddd')"));
```
AMashenkov, line 94:
```suggestion
        assertThrows(CalciteContextException.class, () -> sql("insert into limitedChar(pk, f2) values (1, 'aaaadddd')"));
        assertThrows(CalciteContextException.class, () -> sql("insert into limitedChar(pk, f2) values (1, '    aaaa')"));
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDataTypesTest.java
```diff
@@ -86,6 +87,23 @@ public void testUnicodeStrings() {
...
+        assertThrows(CalciteContextException.class, () -> sql("insert into limitedChar values (2, 'aaaaabbbbb', 'aaaadddd')"));
+        assertThrows(CalciteContextException.class, () -> sql("insert into limitedChar values (3, '我叫什么名字', '我叫什么名字')"));
+        sql("insert into limitedChar values (4, ' aaa    ', ' bbb     ')");
+
+        List<List<Object>> res = sql("select f1, f2 from limitedChar ORDER BY pk");
+
+        List<Object> r = res.get(0);
+
+        assertEquals(" aaa ", r.get(0));
+        assertEquals(" bbb", r.get(1));
```
AMashenkov, line 104: Let's check f2 value " bb    ", will it be trimmed to " bb " or " bb"?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDataTypesTest.java
lowka: It would be even better to have a method not only excepts some generic CalciteContextException but also checks for error message.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDataTypesTest.java
korlov42, line 94: The exception has message `Cannot assign to target field 'F2' of type VARCHAR(4) from source field ''aaaadddd'' of type VARCHAR(4)`. Something definitely is wrong here...

korlov42: this one even better: `Cannot assign to target field 'F1' of type VARCHAR(2) from source field ''a  ' :: CHAR' of type VARCHAR(2)`

korlov42 (changes requested):
Looks like this patch introduces new issues, and test coverage, obviously, is not sufficient

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/TypeUtils.java
```diff
@@ -660,4 +664,32 @@ private static TypeSpec convertToTypeSpec(RelDataType type) {
...
+
+                assert data instanceof String;
+
+                String str = (String) data;
+
+                int colPrecision = colType.getPrecision();
+
+                assert colPrecision != RelDataType.PRECISION_NOT_SPECIFIED;
+
+                if (str.stripTrailing().length() > colPrecision) {
```
korlov42, line 689: so, we use stripped string just to check length and then throw it away... this doesn't look right. We should not put to storage rows with string values whose length violates specified constraint.

```
CREATE TABLE t (id INT PRIMARY KEY, val VARCHAR(3));
INSERT INTO t VALUES (1, ' 1  ');
SELECT length(val) FROM t; -- returns 4
```

the query above returns 4, but I would expect 3 to be returned in that case

korlov42 (changes requested):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/KeyValueModifyPlan.java
```diff
@@ -130,8 +136,22 @@ public <RowT> AsyncCursor<InternalSqlRow> execute(
 
                     UpdatableTable updatableTable = execTable.updatableTable();
 
+                    RowT row = rowSupplier.get();
+
+                    RelDataType rowType = table().getRowType(ctx.getTypeFactory());
+
+                    Supplier<RowSchema> schemaSupplier = () -> {
```
zstan, line 143: probably we don`t need such optimization

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/KeyValueModifyPlan.java
```diff
@@ -130,8 +136,22 @@ public <RowT> AsyncCursor<InternalSqlRow> execute(
...
+                        if (rowSchema != null) {
+                            return rowSchema;
+                        }
+                        rowSchema = rowSchemaFromRelTypes(RelOptUtil.getFieldTypeList(rowType));
+                        return rowSchema;
+                    };
+
+                    row = validateCharactersOverflowAndTrimIfPossible(rowType, ctx.rowHandler(), row, schemaSupplier);
```
korlov42, line 151: not sure why do you do validation outside of the UpdateableTable

zstan: got it, done

korlov42 (approved):

lowka (approved):

PR #1653, opened 2023-02-08
IGNITE-18734 Add validation-related events to Logical Topology

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-18734

### DIFF
--- a/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/ValidationManager.java
+++ b/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/ValidationManager.java
@@ -17,25 +17,15 @@
 
 package org.apache.ignite.internal.cluster.management.raft;
 
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
-import org.apache.ignite.internal.close.ManuallyCloseable;
+import java.util.Collection;
+import java.util.Set;
 import org.apache.ignite.internal.cluster.management.ClusterState;
 import org.apache.ignite.internal.cluster.management.ClusterTag;
-import org.apache.ignite.internal.cluster.management.configuration.ClusterManagementConfiguration;
 import org.apache.ignite.internal.cluster.management.raft.commands.InitCmgStateCommand;
 import org.apache.ignite.internal.cluster.management.raft.commands.JoinReadyCommand;
 import org.apache.ignite.internal.cluster.management.raft.responses.ValidationErrorResponse;
 import org.apache.ignite.internal.cluster.management.topology.LogicalTopology;
-import org.apache.ignite.internal.logger.IgniteLogger;
-import org.apache.ignite.internal.logger.Loggers;
 import org.apache.ignite.internal.properties.IgniteProductVersion;
-import org.apache.ignite.internal.thread.NamedThreadFactory;
-import org.apache.ignite.internal.util.IgniteUtils;
 import org.apache.ignite.network.ClusterNode;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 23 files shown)


modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/LogicalTopologyImpl.java
```diff
@@ -146,38 +157,32 @@ public boolean isNodeInLogicalTopology(ClusterNode needle) {
     }
 
     private void fireAppeared(ClusterNode appearedNode, LogicalTopologySnapshot snapshot) {
```
rpuch, line 159: Let's rename it to `fireNodeJoined`

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/LogicalTopologyImpl.java
```diff
@@ -146,38 +157,32 @@ public boolean isNodeInLogicalTopology(ClusterNode needle) {
...
-            try {
-                listener.onAppeared(appearedNode, snapshot);
-            } catch (Throwable e) {
-                logAndRethrowIfError(e, "Failure while notifying onAppear() listener {}", listener);
-            }
-        }
+        notifyListeners(listener -> listener.onNodeJoined(appearedNode, snapshot), "onNodeJoined");
     }
 
     private void fireDisappeared(ClusterNode oldNode, LogicalTopologySnapshot snapshot) {
```
rpuch, line 163: Let's rename it too

modules/runner/src/integrationTest/java/org/apache/ignite/internal/compute/ItLogicalTopologyTest.java
```diff
@@ -115,21 +131,38 @@ void receivesLogicalTopologyEventsCausedByNodeRestart() throws Exception {
...
 
-        Event leaveEvent = events.get(0);
+        Event leftEvent = firstEvent.eventType == EventType.LEFT ? firstEvent : secondEvent;
+        Event validatedEvent = firstEvent.eventType == EventType.VALIDATED ? firstEvent : secondEvent;
 
-        assertFalse(leaveEvent.appeared);
-        assertThat(leaveEvent.node.name(), is(secondIgnite.name()));
-        assertThat(leaveEvent.topologyVersion, is(3L));
+        assertThat(leftEvent, is(notNullValue()));
+        assertThat(validatedEvent, is(notNullValue()));
```
rpuch, line 149: Non-nullability seems already have been asserted (for first and second event)

modules/runner/src/integrationTest/java/org/apache/ignite/internal/compute/ItLogicalTopologyTest.java
```diff
@@ -115,21 +131,38 @@ void receivesLogicalTopologyEventsCausedByNodeRestart() throws Exception {
 
         restartNode(1);
 
-        waitForCondition(() -> events.size() >= 2, 10_000);
+        // There's a race between a new version of the node being validated, while the old version has not yet been removed from the logical
```
rpuch, line 134: Why is there such a race? Can't we always remove previous node version before making its new version as validated?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/ValidationManager.java
```diff
@@ -146,46 +120,33 @@ ValidationResult validateNode(
...
-            LOG.info("Removing node from the list of validated nodes since no JoinReady requests have been received [node={}]", nodeId);
-
-            cleanupFutures.remove(nodeId);
+    void removeValidatedNodes(Collection<ClusterNode> nodes) {
+        Set<ClusterNode> validatedNodes = storage.getValidatedNodes();
 
-            storage.removeValidatedNode(nodeId);
-        }, configuration.incompleteJoinTimeout().value(), TimeUnit.MILLISECONDS);
+        nodes.forEach(node -> {
+            if (validatedNodes.contains(node)) {
```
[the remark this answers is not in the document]
sashapolo: I understand your concern, but how can this happen? How can two nodes have two addresses but the same ID? Is this even going to work?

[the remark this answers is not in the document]
rpuch: This reasoning implies that same node ID can only correspond to one address. This seems reasonable, but, if we switch to using just an ID, we don't need to imply anything. The code will be easier to reason about and, if, for some unknown reason (like switching to a different thing instead of SWIM?) that we cannot imagine now, the 'one address for same ID' is now longer true, we will not be affected (as we would not imply anything).

It seems that the change is not too heavy to implement it and sleep patiently :)

rpuch (approved):

PR #1655, opened 2023-02-08
IGNITE-18737 Fixed updating of partition assignments when new assignments equal to table configuration assignments and stable assignments are empty.

sergeyuttsel:
https://issues.apache.org/jira/browse/IGNITE-18737


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1863,23 +1863,36 @@ public void onUpdate(WatchEvent evt) {
...
-                                updatePendingAssignmentsKeys(
-                                        tableView.name(), replicaGrpId, dataNodes, tableView.replicas(),
-                                        evt.entryEvent().newEntry().revision(), metaStorageMgr, part
-                                ).exceptionally(e -> {
-                                    LOG.error(
-                                            "Exception on updating assignments for [table={}, partition={}]", e, tableView.name(), partId
-                                    );
+                                Set<Assignment> partAssignments = AffinityUtils.calculateAssignmentForPartition(dataNodes, part, replicas);
+
+                                // TODO IGNITE-18624 This check will not be needed when dataNodes from distribution zone will be used
```
alievmirza, line 1879: I would remove this todo, and leave this check in any cases, because, for example, we still can rewrite `dataNodes` with the same value, for example, when we start node and init data nodes from the Vauit with method `DistributionZoneManager#initDataNodesFromVaultManager`.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1863,23 +1863,36 @@ public void onUpdate(WatchEvent evt) {
...
+
                             for (int part = 0; part < tableView.partitions(); part++) {
                                 UUID tableId = ((ExtendedTableConfiguration) tableCfg).id().value();
 
                                 TablePartitionId replicaGrpId = new TablePartitionId(tableId, part);
 
-                                int partId = part;
+                                int replicas = tableView.replicas();
 
-                                updatePendingAssignmentsKeys(
```
sanpwc, line 1873: Besides updating pending assignments inside updatePendingAssignmentsKeys we also update `partition.change.trigger.revision `. Should we update it even if won't update the assignments themselves?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1863,23 +1863,36 @@ public void onUpdate(WatchEvent evt) {
...
-                                        tableView.name(), replicaGrpId, dataNodes, tableView.replicas(),
-                                        evt.entryEvent().newEntry().revision(), metaStorageMgr, part
-                                ).exceptionally(e -> {
-                                    LOG.error(
-                                            "Exception on updating assignments for [table={}, partition={}]", e, tableView.name(), partId
-                                    );
+                                Set<Assignment> partAssignments = AffinityUtils.calculateAssignmentForPartition(dataNodes, part, replicas);
```
sanpwc, line 1877: So, we will recalculate partAssignments twice, here and inside updatePendingAssignmentsKeys. Is it possible to have one calculation only?

sanpwc, line 1877: Why we check this only for assignments recalculation triggered by dataNodes change? Is it possible to face same issue in case of other updatePendingAssignmentsKeys triggers?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1863,23 +1863,36 @@ public void onUpdate(WatchEvent evt) {
...
+
+                                // TODO IGNITE-18624 This check will not be needed when dataNodes from distribution zone will be used
+                                // instead of BaselineManager.nodes.
+                                if (!partAssignments.equals(tableAssignments.get(part))) {
+                                    int partId = part;
 
-                                    return null;
-                                });
+                                    updatePendingAssignmentsKeys(
```
sanpwc, line 1884: What if we already have pending rebalance and should add calculated ones as planned ones? I mean following:
current A,B,C
pending A,B
new one that we've calculated, again A,B,C
despite the fact that current == new one we shouldn't skip it.

Please pay attention, that we have the logic similar to the one we want inside updatePendingAssignmentsKeys that checks assignments.stable and new assignments equality 
`&& partition.assignments.stable != calcPartAssighments()`

sanpwc (changes requested):

modules/table/src/test/java/org/apache/ignite/internal/utils/RebalanceUtilUpdateAssignmentsTest.java
```diff
@@ -0,0 +1,437 @@
...
+
+    private SimpleInMemoryKeyValueStorage keyValueStorage;
+
+    private ConfigurationManager clusterCfgMgr;
+
+    private ClusterService clusterService;
+
+    private MetaStorageManager metaStorageManager;
+
+    static int partNum = 2;
```
sanpwc, line 91: Related to all that static values:
- Do you really need them to be non-private?
- If not, what about converting them to constants?

modules/table/src/test/java/org/apache/ignite/internal/utils/RebalanceUtilUpdateAssignmentsTest.java
```diff
@@ -0,0 +1,437 @@
...
+    static Set<String> nodes2 = IntStream.of(5).mapToObj(i -> "nodes2_" + i).collect(toSet());
+    static Set<String> nodes3 = IntStream.of(5).mapToObj(i -> "nodes3_" + i).collect(toSet());
+    static Set<String> nodes4 = IntStream.of(5).mapToObj(i -> "nodes4_" + i).collect(toSet());
+
+    static Set<Assignment> assignments1 = calculateAssignmentForPartition(nodes1, partNum, replicas);
+    static Set<Assignment> assignments2 = calculateAssignmentForPartition(nodes2, partNum, replicas);
+    static Set<Assignment> assignments3 = calculateAssignmentForPartition(nodes3, partNum, replicas);
+    static Set<Assignment> assignments4 = calculateAssignmentForPartition(nodes4, partNum, replicas);
+
+    @BeforeEach
```
sanpwc, line 104: Why it's beforeEach and not beforeAll?

modules/table/src/test/java/org/apache/ignite/internal/utils/RebalanceUtilUpdateAssignmentsTest.java
```diff
@@ -0,0 +1,437 @@
...
+                new TestConfigurationStorage(DISTRIBUTED),
+                List.of(),
+                List.of()
+        );
+
+        clusterService = mock(ClusterService.class);
+
+        metaStorageManager = mock(MetaStorageManager.class);
+
+        clusterCfgMgr.start();
```
sanpwc, line 118: As a rule of thumb if you start something you need to stop it, and if you start the IgniteComponent you should also call beforeNodeStop before calling the stop.

sanpwc (changes requested):

modules/table/src/test/java/org/apache/ignite/internal/utils/RebalanceUtilUpdateAssignmentsTest.java
```diff
@@ -0,0 +1,437 @@
...
+        );
+    }
+
+    /**
+     * Nodes for new assignments calculating: nodes1.
+     * The table configuration assignments: assignments1.
+     * Current assignments in the metastorage: stable=assignments1, pending=assignments2, planned=null.
+     * Expected assignments in the metastorage after updating: stable=assignments1, pending=assignments2, planned=assignments1.
+     */
+    @Test
```
sanpwc, line 314: Why we only have the javadoc for test14?

modules/table/src/test/java/org/apache/ignite/internal/utils/RebalanceUtilUpdateAssignmentsTest.java
```diff
@@ -0,0 +1,437 @@
...
+            actualPlannedAssignments = ByteUtils.fromBytes(actualPlannedBytes);
+        }
+
+        LOG.info("stableAssignments " + actualStableAssignments);
+        LOG.info("pendingAssignments " + actualPendingAssignments);
+        LOG.info("plannedAssignments " + actualPlannedAssignments);
+
+        if (expectedStableAssignments != null) {
+            assertNotNull(actualStableBytes);
+            assertTrue(actualStableAssignments.containsAll(expectedStableAssignments));
```
sanpwc, line 415: what about assertEquals

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -100,32 +114,53 @@ public class RebalanceUtil {
...
         //    if empty(partition.change.trigger.revision) || partition.change.trigger.revision < event.revision:
-        //        if empty(partition.assignments.pending) && partition.assignments.stable != calcPartAssighments():
+        //        if empty(partition.assignments.pending)
+        //              && ((isNewAssignments && empty(partition.assignments.stable))
+        //                  || (partition.assignments.stable != calcPartAssighments() && !empty(partition.assignments.stable))):
         //            partition.assignments.pending = calcPartAssignments()
         //            partition.change.trigger.revision = event.revision
         //        else:
-        //            if partition.assignments.pending != calcPartAssignments
+        //            if partition.assignments.pending != calcPartAssignments && !empty(partition.assignments.pending)
```
alievmirza, line 128: Please correct pseudocode according to the new iff

alievmirza (approved):

PR #1659, opened 2023-02-09
IGNITE-18579 Support SSL for ScaleCube

PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-18579


modules/network/src/main/java/org/apache/ignite/internal/network/configuration/AbstractSslConfigurationSchema.java
```diff
@@ -0,0 +1,46 @@
...
+package org.apache.ignite.internal.network.configuration;
+
+import org.apache.ignite.configuration.annotation.AbstractConfiguration;
+import org.apache.ignite.configuration.annotation.ConfigValue;
+import org.apache.ignite.configuration.annotation.Value;
+import org.apache.ignite.configuration.validation.OneOf;
+
+/** SSL configuration schema. */
+@AbstractConfiguration
+public class AbstractSslConfigurationSchema {
```
sashapolo, line 27: why do you need this abstract schema? Are there going to be more inheritors apart from `SslConfigurationSchema`?

modules/network/src/main/java/org/apache/ignite/internal/network/configuration/AbstractSslConfigurationSchema.java
```diff
@@ -0,0 +1,46 @@
...
+    /** Enable/disable SSL. */
+    @Value(hasDefault = true)
+    public final boolean enabled = false;
+
+    /** Client authentication. */
+    @OneOf({"none", "optional", "require"})
```
sashapolo, line 33: Please update the javadoc with a description about what these options mean

modules/network/src/main/java/org/apache/ignite/internal/network/ssl/KeystoreLoader.java
```diff
@@ -0,0 +1,48 @@
...
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.cert.CertificateException;
+import org.apache.ignite.internal.network.configuration.KeyStoreView;
+
+/** Java keystore loader. */
+public final class KeystoreLoader {
+
+    private KeystoreLoader() {
+        throw new IllegalStateException("KeystoreLoader is an utility class and should not be instantiated");
```
sashapolo, line 33: I think that throwing an exception here is redundant, having a private constructor is enough

modules/network/src/main/java/org/apache/ignite/internal/network/ssl/SslContextProvider.java
```diff
@@ -0,0 +1,93 @@
...
+import javax.net.ssl.TrustManagerFactory;
+import org.apache.ignite.internal.network.configuration.SslView;
+import org.apache.ignite.lang.ErrorGroups.Common;
+import org.apache.ignite.lang.IgniteException;
+
+/** SSL context provider. */
+public final class SslContextProvider {
+
+    private SslContextProvider() {
+        throw new IllegalStateException("SslContextProvider is an utility class and should not be instantiated");
```
sashapolo, line 39: Same here about throwing an exception

modules/network/src/main/java/org/apache/ignite/internal/network/netty/NettyClient.java
```diff
@@ -48,44 +51,50 @@ public class NettyClient {
...
-    private volatile Channel channel = null;
-
     /** Message listener. */
     private final Consumer<InNetworkObject> messageListener;
 
     /** Handshake manager. */
     private final HandshakeManager handshakeManager;
 
+    /** SSL configuration. */
+    private final SslView configuration;
```
sashapolo, line 64: let's rename this field to `sslConfiguration`

modules/network/src/main/java/org/apache/ignite/internal/network/ssl/KeystoreLoader.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /** Initialize and load keystore with provided configuration. */
+    public static KeyStore load(KeyStoreView keyStoreConfiguration)
+            throws KeyStoreException, IOException, CertificateException, NoSuchAlgorithmException {
```
sashapolo, line 38: Do we really need to throw all these exceptions? Maybe we should wrap them into `IgniteException` here?

modules/network/src/main/java/org/apache/ignite/internal/network/ssl/KeystoreLoader.java
```diff
@@ -0,0 +1,48 @@
...
+
+        KeyStore ks = KeyStore.getInstance(keyStoreConfiguration.type());
+        ks.load(
+                Files.newInputStream(Path.of(keyStoreConfiguration.path())),
```
sashapolo, line 42: `Files.newInputStream` should be used with try-with-resources

modules/network/src/test/java/org/apache/ignite/network/DefaultMessagingServiceTest.java
```diff
@@ -168,6 +169,9 @@ private static void configureNetworkDefaults(
         lenient().when(networkConfigView.portRange()).thenReturn(0);
         lenient().when(networkConfigView.outbound()).thenReturn(outboundConfig);
         lenient().when(networkConfigView.inbound()).thenReturn(inboundConfig);
+        SslView ssl = mock(SslView.class);
+        lenient().when(ssl.enabled()).thenReturn(false);
```
sashapolo, line 173: I think these mocks should be rewritten using `@InjectConfiguration`

modules/network/src/test/java/org/apache/ignite/internal/network/ssl/SslContextProviderTest.java
```diff
@@ -0,0 +1,165 @@
...
+import org.apache.ignite.lang.IgniteException;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@ExtendWith(ConfigurationExtension.class)
+class SslContextProviderTest {
+
+    @InjectConfiguration
+    SslConfiguration configuration;
```
sashapolo, line 40: All these fields can be `final`

modules/network/src/test/java/org/apache/ignite/internal/network/ssl/SslContextProviderTest.java
```diff
@@ -0,0 +1,165 @@
...
+    String password;
+
+    String keyStorePkcs12Path;
+
+    String trustStoreJks12Path;
+
+    @BeforeEach
+    void setUp() {
+        password = "changeit";
+        keyStorePkcs12Path = SslContextProviderTest.class.getClassLoader().getResource("ssl/keystore.p12").getPath();
```
sashapolo, line 51: I have a question regarding these keystore and truststore files: can they be generated automatically and not placed into the repository? We could do it either manually or using a Junit extension

modules/runner/src/integrationTest/java/org/apache/ignite/internal/ssl/ItSslTest.java
```diff
@@ -0,0 +1,193 @@
...
+import org.junit.jupiter.api.TestInfo;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.TestInstance.Lifecycle;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/** SSL support integration test. */
+@ExtendWith(WorkDirectoryExtension.class)
+public class ItSslTest {
+
+    static String password;
```
sashapolo, line 43: all these fields can be `private`

modules/network/src/main/java/org/apache/ignite/internal/network/configuration/AbstractSslConfigurationSchema.java
PakhomovAlexander: We need this abstract schema because it will be inherited by `rest.ssl` and  `clientConnector.ssl`.

modules/network/src/main/java/org/apache/ignite/internal/network/ssl/KeystoreLoader.java
PakhomovAlexander: They are caught by the calling side that decides how to interpret it and wraps to IgniteException with the appropriate error code.

modules/network/src/test/java/org/apache/ignite/internal/network/ssl/SslContextProviderTest.java
PakhomovAlexander: I think they could be generated but the vast majority of users generate keystores from the command line with `openssl` and `keytool`. I think it is better to test such scenarios. 

WDYT?

modules/network/src/test/java/org/apache/ignite/internal/network/ssl/SslContextProviderTest.java
PakhomovAlexander: This is the test with the defined lifecycle: fields initialized on `setUp`. Why should we initialize them in the test's constructor instead?

modules/network/src/main/java/org/apache/ignite/internal/network/ssl/KeystoreLoader.java
sashapolo: Are they handled differently? Looks like I missed that

modules/network/src/test/java/org/apache/ignite/internal/network/ssl/SslContextProviderTest.java
sashapolo: That's a typo, sorry. I meant `private`, not `final`

modules/network/src/test/java/org/apache/ignite/internal/network/ssl/SslContextProviderTest.java
sashapolo: Why would they need to generate anything for tests? I don't understand your point

modules/network/src/test/java/org/apache/ignite/network/DefaultMessagingServiceTest.java
PakhomovAlexander: Do you mean It would be nice to rewrite all mocks here or just SslConfiguration?

modules/network/src/main/java/org/apache/ignite/internal/network/configuration/AbstractSslConfigurationSchema.java
sashapolo: got it

modules/network/src/test/java/org/apache/ignite/internal/network/ssl/SslContextProviderTest.java
PakhomovAlexander: No, they do not need it. 

But we test the scenario that is close to the real world. (The keystore and truststore were generated by me using `openssl` and `keytool`). That means they are likely the same as our users would use. 

In case we generate keystore and truststore directly from java code, I can not guarantee that they are the same as users would use.  Does it make sense?

modules/network/src/test/java/org/apache/ignite/network/DefaultMessagingServiceTest.java
sashapolo: all mocks here (I don't expect it to be a big change)

modules/network/src/test/java/org/apache/ignite/internal/network/ssl/SslContextProviderTest.java
sashapolo: I think this should be done as higher level tests, probably. Users could also use a different tool for generating all this stuff. I think it's ok to rely on some default implementations that are not written by us

modules/network/src/main/java/org/apache/ignite/internal/network/ssl/KeystoreLoader.java
PakhomovAlexander: I would say no, they are handled in the same way. 

But `KeystoreLoader` is used in `SslContextProvider` together with `TrustManagerFactory` and `KeyManagerFactory`. These classes declare almost the same set of checked exceptions in their signatures. This fact forces us to catch them in `SslContextProvider `. So, there is a place, that handles all exceptions:

```java 
        } catch (NoSuchFileException e) {
            throw new IgniteException(Common.SSL_CONFIGURATION_ERR, String.format("File %s not found", e.getMessage()), e);
        } catch (IOException | CertificateException | KeyStoreException | NoSuchAlgorithmException | UnrecoverableKeyException e) {
            throw new IgniteException(Common.SSL_CONFIGURATION_ERR, e);
        }

File: modules/network/src/test/java/org/apache/ignite/internal/network/ssl/SslContextProviderTest.java (hunk 10 above)
Comment (PakhomovAlexander) [reply modules/network/src/test/java/org/apache/ignite/internal/network/ssl/SslContextProviderTest.java]: agreed

File: modules/network/src/main/java/org/apache/ignite/internal/network/ssl/KeystoreLoader.java (hunk 6 above)
Comment (sashapolo) [reply modules/network/src/main/java/org/apache/ignite/internal/network/ssl/KeystoreLoader.java:37]: ok, looks fine either way

File: modules/network/src/test/java/org/apache/ignite/network/DefaultMessagingServiceTest.java [hunk 12]
```diff
@@ -102,10 +90,9 @@ class DefaultMessagingServiceTest {
     );
 
     @BeforeEach
-    void setUp() {
-        configureSender();
-        configureReceiver();
-
+    void setUp() throws InterruptedException, ExecutionException {
+        senderNetworkConfig.port().update(SENDER_PORT).get();
```
sashapolo, line 94: This stuff can be simplified by providing a text config, for example:
```
@InjectConfiguration("mock.port=" + RECEIVER_PORT)
private NetworkConfiguration receiverNetworkConfig;
```

sashapolo (approved):

modules/network/src/test/java/org/apache/ignite/network/DefaultMessagingServiceTest.java
PakhomovAlexander: Awesome! I tried another variant and got NPE `@InjectConfiguration("network.port=" + RECEIVER_PORT)`. Your variant works.

PR #1666, opened 2023-02-13
IGNITE-18565 Modify getOrCreateMvPartition and getMvPartition of MvTableStorage to return the future

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18565


modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/StorageOperation.java
```diff
@@ -0,0 +1,90 @@
...
+    class CreateStorageOperation implements StorageOperation {
+    }
+
+    /**
+     * Storage destruction operation.
+     */
+    class DestroyStorageOperation implements StorageOperation {
+        private final CompletableFuture<Void> destroyFuture = new CompletableFuture<>();
+
+        private final AtomicReference<CreateStorageOperation> createStorageOperationReference = new AtomicReference<>();
```
SammyVimes, line 40: Looks like this can be just AtomicBoolean

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/MvPartitionStorages.java
```diff
@@ -0,0 +1,495 @@
...
+    public CompletableFuture<Void> destroy(int partitionId, Function<T, CompletableFuture<Void>> destroyStorageFunction) {
+        DestroyStorageOperation destroyOp = (DestroyStorageOperation) operationByPartitionId.compute(partitionId, (partId, operation) -> {
+            checkStorageExists(partitionId);
+
+            throwExceptionDependingOnOperation(operation, partitionId);
+
+            return new DestroyStorageOperation();
+        });
+
+        return completedFuture(null)
```
SammyVimes, line 151: Why we need a `completedFuture` here? Can't we just use `destroyStorageFunction.apply(storageByPartitionId.getAndSet(partitionId, null))`

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/MvPartitionStorages.java
```diff
@@ -0,0 +1,495 @@
...
+     * @throws StorageRebalanceException If the storage does not exist.
+     */
+    private void checkStorageExistsForRebalance(int partitionId) {
+        if (get(partitionId) == null) {
+            throw new StorageRebalanceException(createStorageDoesNotExistErrorMessage(partitionId));
+        }
+    }
+
+    private void throwExceptionDependingOnOperation(StorageOperation operation, int partitionId) {
+        if (operation == null) {
```
SammyVimes, line 421: Maybe we can extract null check, otherwise it looks like this method always throws which is not the case

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/MvPartitionStorages.java
```diff
@@ -0,0 +1,495 @@
...
+            throw new StorageRebalanceException(createStorageInProgressOfFinishRebalanceErrorMessage(partitionId));
+        } else if (operation instanceof CleanupStorageOperation) {
+            throw new StorageException(createStorageInProgressOfCleanupErrorMessage(partitionId));
+        } else {
+            throw new StorageException(createUnknownOperationErrorMessage(partitionId, operation));
+        }
+    }
+
+    private void throwExceptionDependingOnOperationForRebalance(StorageOperation operation, int partitionId) {
+        if (operation == null) {
```
SammyVimes, line 441: Same here

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/StorageUtils.java
```diff
@@ -155,6 +155,15 @@ public static void throwStorageExceptionIfItCause(IgniteInternalCheckedException
         }
     }
 
+    /**
+     * Throws an {@link StorageException} if the storage already exists or is being created.
+     *
+     * @param storageInfo Storage information, for example in the format "table=user, partitionId=1".
+     */
+    public static void throwExceptionIfStorageExists(String storageInfo) {
+        throw new StorageException("Storage already exists or is being created: " + storageInfo);
```
SammyVimes, line 164: the method name says "ifStorageExists" but it throws unconditionally

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/TestMvTableStorageTest.java
```diff
@@ -32,7 +32,7 @@
 public class TestMvTableStorageTest extends AbstractMvTableStorageTest {
     @BeforeEach
     void setUp(
-            @InjectConfiguration("mock.tables.foo.dataStorage.name = " + TestStorageEngine.ENGINE_NAME)
+            @InjectConfiguration("mock.tables.foo{ partitions = 512, dataStorage.name = " + TestStorageEngine.ENGINE_NAME + "}")
```
SammyVimes, line 35: Why 512?

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccessImplTest.java
```diff
@@ -204,4 +208,12 @@ void testAddWriteCommitted() {
 
         verify(indexStorage, never()).put(eq(binaryRow), eq(rowId));
     }
+
+    private static MvPartitionStorage createMvPartition(MvTableStorage tableStorage, int partitionId) {
+        CompletableFuture<MvPartitionStorage> createMvPartitionFuture = tableStorage.createMvPartition(partitionId);
+
+        assertThat(createMvPartitionFuture, willCompleteSuccessfully());
+
+        return createMvPartitionFuture.join();
```
SammyVimes, line 217: Probably should be `get` with timeout

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2181,22 +2181,24 @@ private static PeersAndLearners configurationFromAssignments(Collection<Assignme
...
     private CompletableFuture<PartitionStorages> getOrCreatePartitionStorages(TableImpl table, int partitionId) {
-        return CompletableFuture
-                .supplyAsync(() -> {
-                    MvPartitionStorage mvPartitionStorage = table.internalTable().storage().getOrCreateMvPartition(partitionId);
-                    TxStateStorage txStateStorage = table.internalTable().txStateStorage().getOrCreateTxStateStorage(partitionId);
+        InternalTable internalTable = table.internalTable();
+
+        MvPartitionStorage mvPartition = internalTable.storage().getMvPartition(partitionId);
+
+        return (mvPartition != null ? completedFuture(mvPartition) : internalTable.storage().createMvPartition(partitionId))
```
SammyVimes, line 2188: Can partition be created in a parallel thread at the same time, so that getMvPartition returns null, but at the moment when we call `createMvPartition` we already called it in another thread?

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/MvPartitionStorages.java
```diff
@@ -0,0 +1,495 @@
...
+        storageByPartitionId = new AtomicReferenceArray<>(tableView.partitions());
+    }
+
+    /**
+     * Returns the multi-versioned partition storage, {@code null} if the storage does not exist (not created or destroyed).
+     *
+     * @throws IllegalArgumentException If partition ID is out of configured bounds.
+     */
+    public @Nullable T get(int partitionId) {
+        checkPartitionId(partitionId);
```
SammyVimes, line 76: shouldn't we also check it on destroy? And, probably, on other operations like create

SammyVimes (changes requested):

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/engine/MvTableStorage.java
```diff
@@ -46,14 +46,18 @@
  */
 public interface MvTableStorage extends ManuallyCloseable {
     /**
-     * Retrieves or creates a partition for the current table. Not expected to be called concurrently with the same Partition ID.
+     * Creates a partition for the current table.
+     *
+     * <p>If the partition has already been created or is in the process of being created, an {@link StorageException} will be thrown.
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/StorageOperation.java
tkalkirill: Not suitable, I need to atomically replace it with the creation operation when the destruction operation is completed if necessary.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/MvPartitionStorages.java
tkalkirill: This is a trick to make sure the operation gets rid of if it succeeds or fails.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/MvPartitionStorages.java
tkalkirill: I agree, I'll fix it.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/MvPartitionStorages.java
tkalkirill: Fix it

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/StorageUtils.java
tkalkirill: This method is not needed, I will remove it.

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/TestMvTableStorageTest.java
tkalkirill: All heirs of `org.apache.ignite.internal.storage.AbstractMvTableStorageTest` set this value, they missed it here, so some tests may fail.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccessImplTest.java
tkalkirill: It is not necessary because above we make sure that the future completed.

`assertThat(createMvPartitionFuture, willCompleteSuccessfully());`

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/MvPartitionStorages.java
tkalkirill: This is what happens, for example they call `org.apache.ignite.internal.storage.util.MvPartitionStorages#checkStorageExists` which calls this method.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: Only one thread should create storage, and receive the rest.

On the one hand, we will be able to catch those who are trying to create tables in parallel and reduce the number of code, on the other hand, we can divide it into two methods and use it correctly. Well, or store a map with storage futures and create them lazily.

wdyt?

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccessImplTest.java
SammyVimes: how about getNow() then?

tkalkirill: I do not think that this is necessary, the future is already definitely completed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
SammyVimes: I think we can leave it as it is, but we need a test for this behaviour

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/util/MvPartitionStoragesTest.java
```diff
@@ -0,0 +1,606 @@
...
+        assertThrowsWithMessage(StorageRebalanceException.class, () -> finishRebalanceMvStorage(0), "Storage does not exist");
+
+        assertThat(createMvStorage(0), willCompleteSuccessfully());
+
+        assertThrowsWithMessage(StorageRebalanceException.class, () -> finishRebalanceMvStorage(0), "Storage rebalancing did not start");
+
+        assertThat(startRebalanceMvStorage(0), willCompleteSuccessfully());
+
+        // What if there is an error during the operation?
+
```
SammyVimes, line 501: Dangling comment

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/util/MvPartitionStoragesTest.java
```diff
@@ -0,0 +1,606 @@
...
+        finishErrorDestroyMvStorageFuture.completeExceptionally(new RuntimeException("from test"));
+
+        assertThat(errorDestroyMvStorageFuture, willFailFast(RuntimeException.class));
+        assertThat(errorReCreateMvStorageFuture, willFailFast(RuntimeException.class));
+
+        assertNull(getMvStorage(0));
+    }
+
+    @Test
+    void testCreateAfterDestroy() {
```
SammyVimes, line 188: Looks like there is no testCreateAfterDestroyError

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/util/MvPartitionStoragesTest.java
```diff
@@ -0,0 +1,606 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.function.Executable;
+
+/**
+ * Class for testing {@link MvPartitionStorages}.
+ */
+@ExtendWith(ConfigurationExtension.class)
+public class MvPartitionStoragesTest {
```
SammyVimes, line 55: I think we need tests for consecutive invocation of the same operations

SammyVimes (changes requested):

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
```diff
@@ -121,7 +120,7 @@ public class RocksDbTableStorage implements MvTableStorage {
     private volatile ColumnFamily hashIndexCf;
 
     /** Partition storages. */
-    private volatile AtomicReferenceArray<RocksDbMvPartitionStorage> partitions;
+    private volatile MvPartitionStorages<RocksDbMvPartitionStorage> mvPartitionStorages;
```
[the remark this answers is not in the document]
tkalkirill: From the point of view of the product code, the answer is "yes we can", but from the point of view of the designs should do in `org.apache.ignite.internal.storage.engine.MvTableStorage#start`.

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/util/MvPartitionStoragesTest.java
tkalkirill: No, do you think it's needed? what to check in it?

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/util/MvPartitionStoragesTest.java
tkalkirill: Like separate test methods?

SammyVimes: However you like

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/util/MvPartitionStoragesTest.java
SammyVimes: It's not necessary, just it looks like other operation have operation + "Error" tests

SammyVimes (approved):
LGTM

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: When I forgot to change the code here, quite a few tests fell on me.

modules/storage-api/src/test/java/org/apache/ignite/internal/storage/util/MvPartitionStoragesTest.java
tkalkirill: Added

PR #1667, opened 2023-02-13
[IGNITE-18748] Sql. Planner optimization for JOIN. Benchmarks.

lowka:


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/benchmarks/TpchBenchmark.java
```diff
@@ -0,0 +1,111 @@
...
+@State(Scope.Benchmark)
+public class TpchBenchmark {
+
+    /**
+     * Identifiers of TPC-H queries. See {@link TpchQueries#getQuery(String)}.
+     */
+    @Param({
+            "1", "2", "3", "4", "5", "6", "7", "8", "8v", "9", "10", "11", "12", "12v",
+            "13", "14", "14v", "15", "16", "17", "18", "19", "20", "21", "22"
+    })
```
AMashenkov, line 58: Can we search for all available *.sql resources in ./tpch subdirectory? like we do in ItSqlLogic tests?

lowka: This enumeration is necessary because there is no parameter provider or anything similar to TestFactory from junit in jmh.

AMashenkov (approved):

PR #1675, opened 2023-02-15
IGNITE-18208 Sql. Restructuring of the planner tests for aggregates 

AMashenkov:


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/ColocatedSortAggregatePlannerTest.java
```diff
@@ -0,0 +1,382 @@
...
+                nodeOrAnyChild(isInstanceOf(IgniteColocatedSortAggregate.class)
+                        .and(input(isInstanceOf(IgniteExchange.class)
+                                .and(input(isIndexScan("TEST", "val0")))
+                        ))
+                ),
+                ArrayUtils.concat(disabledRules(), additionalRules));
+    }
+
+    @Override
+    protected void checkTestCase17(String sql, IgniteSchema schema, RelCollation collation) throws Exception {
```
zstan, line 244: the same as from parent implementation, and seems there are numerous the same issues, is it ok ?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AggregatePlannerTest.java
```diff
@@ -17,466 +17,371 @@
...
-import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.Arguments;
-import org.junit.jupiter.params.provider.EnumSource;
-import org.junit.jupiter.params.provider.MethodSource;
 
 /**
- * AggregatePlannerTest.
- * TODO Documentation https://issues.apache.org/jira/browse/IGNITE-15859
+ * Test that verifies plans for queries with aggregates.
+ * The test should not disable any rules and validates the best query plans.
```
zstan, line 40: Ok we not disable any rules here, but wait appropriate aggregate algo to be involved, i.e. :
`        assertPlan(sql, schema,
                nodeOrAnyChild(isInstanceOf(IgniteColocatedHashAggregate.class)
`
If cost model or calcite inner implementation will be changed - seems we will be forced to change failed tests automatically and blindly i understand that older implementation was the same, but do u think it`s still suitable?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AbstractAggregatePlannerTest.java
```diff
@@ -49,20 +44,41 @@
...
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Test;
 
 /**
- * Base class for further planner test implementations.
+ * This is a base class for aggregate-related tests aimed to verify integration of aggregate nodes
+ * into the planner.
+ *
+ * <p>Tests derived from this class can be separated onto two group.
```
ygerzhedovich, line 55:
```suggestion
 * <p>Tests derived from this class can be separated into two groups.
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AbstractAggregatePlannerTest.java
```diff
@@ -493,56 +493,26 @@ static void ensureAllCasesAreCovered() {
...
-
-    /**
-     * Rules to disable.
+     * Verifies given test case with provided predicate.
+     *
+     * <p>That is, applies predicate to the result of the query optimization with regards to the
+     * provided collection of rules which have to be disabled during optimization.
      *
-     * @return Rules.
+     * @param testCase A test case tp verify.
```
ygerzhedovich, line 501: what is `tp`

ygerzhedovich (approved):

PR #1680, opened 2023-02-15
IGNITE-18790 Add SSL support for JDBC driver

PakhomovAlexander:


modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcConnection.java
```diff
@@ -149,13 +151,14 @@ public JdbcConnection(ConnectionProperties props) throws SQLException {
...
-            client = ((TcpIgniteClient) IgniteClient
-                    .builder()
+            Builder builder = IgniteClient.builder()
                     .addresses(addrs)
                     .connectTimeout(netTimeout)
                     .reconnectThrottlingPeriod(reconnectThrottlingPeriod)
-                    .reconnectThrottlingRetries(reconnectThrottlingRetries)
-                    .build());
+                    .reconnectThrottlingRetries(reconnectThrottlingRetries);
+            setupSsl(builder);
```
AMashenkov, line 159:
```suggestion
                    .reconnectThrottlingRetries(reconnectThrottlingRetries)
                    .ssl(extractSslConfiguration(connProps)
                    .build();
```

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcConnection.java
```diff
@@ -170,6 +173,23 @@ public JdbcConnection(ConnectionProperties props) throws SQLException {
...
+                            .trustStorePath(connProps.getTrustStorePath())
+                            .trustStorePassword(connProps.getTrustStorePassword())
+                            .clientAuth(connProps.getClientAuth())
+                            .keyStoreType(connProps.getKeyStoreType())
+                            .keyStorePath(connProps.getKeyStorePath())
+                            .keyStorePassword(connProps.getKeyStorePassword())
+                            .build()
+            );
+        }
+    }
```
AMashenkov, line 191:
```suggestion
    private @Nullable SslConfiguration extractSslConfiguration(ConnectionProperties connProps) {
        if (connProps.isSslEnabled()) {
                    return SslConfiguration.builder()
                            .enabled(true)
                            .trustStoreType(connProps.getTrustStoreType())
                            .trustStorePath(connProps.getTrustStorePath())
                            .trustStorePassword(connProps.getTrustStorePassword())
                            .clientAuth(connProps.getClientAuth())
                            .keyStoreType(connProps.getKeyStoreType())
                            .keyStorePath(connProps.getKeyStorePath())
                            .keyStorePassword(connProps.getKeyStorePassword())
                            .build();
        } else {
            return null;
        }
    }
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/ssl/ItSslTest.java
```diff
@@ -104,10 +104,18 @@ void clientCouldConnectWithoutSsl() throws Exception {
                 assertThat(client.clusterNodes(), hasSize(2));
             }
         }
+
+        @Test
+        @DisplayName("Jdbc driver could establish the connection when SSL disabled")
+        void jdbcCouldConnectWithoutSsl() throws SQLException {
```
AMashenkov, line 110:
```suggestion
        void jdbcCanConnectWithoutSsl() throws SQLException {
```
Also,
Can secure jdbc connection be established with insecure server?
I guess we should expect an exception in that case, as client requirements looks more strict.
Let's add a test for this.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/ssl/ItSslTest.java
```diff
@@ -305,5 +338,23 @@ void clientCanConnectWithSslAndClientAuth() throws Exception {
...
+        @Test
+        @DisplayName("Jdbc client can connect with SSL configured")
+        void jdbcCanConnectWithSslAndClientAuth() throws SQLException {
+            var url =
+                    "jdbc:ignite:thin://127.0.0.1:10800"
+                            + "?sslEnabled=true"
+                            + "&trustStorePath=" + trustStorePath
+                            + "&trustStoreType=JKS"
+                            + "&trustStorePassword=" + password
+                            + "&clientAuth=require"
```
AMashenkov, line 351: Do we support clientAuth=optional\none ?

PakhomovAlexander: This is supported by Netty SSL handler implementation. So, basically yes, it is supported.

PakhomovAlexander: But probably, the integration test is not the best place to setup all possible configurations for the cluster.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/ssl/ItSslTest.java
```diff
@@ -143,6 +178,9 @@ class ClusterWithSsl {
                 + "    }\n"
                 + "  }\n"
                 + "}";
+        @WorkDirectory
```
AMashenkov, line 181:
```suggestion

        @WorkDirectory
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/ssl/ItSslTest.java
```diff
@@ -79,6 +75,9 @@ class ClusterWithoutSsl {
...
```
AMashenkov, line 78:
```suggestion

        @WorkDirectory
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/ssl/ItSslTest.java
AMashenkov: Ok, we check only connectivity here.

I think we can create a ticket to add more tests later, to check supported cyphers and clientAuth option, and maybe smth else.

AMashenkov (approved):

PR #1688, opened 2023-02-17
IGNITE-18835 Get rid of skipping safe time waiting on a primary node.

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-18835

### DIFF
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
@@ -32,7 +32,6 @@ import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.stream.IntStream;
 import org.apache.ignite.internal.app.IgniteImpl;
-import org.apache.ignite.internal.hlc.HybridClockImpl;
 import org.apache.ignite.internal.schema.BinaryRow;
 import org.apache.ignite.internal.schema.BinaryRowEx;
 import org.apache.ignite.internal.schema.marshaller.TupleMarshallerImpl;
@@ -175,9 +174,9 @@ public class ItRaftCommandLeftInLogUntilRestartTest extends AbstractBasicIntegra
         BinaryRowEx key = new TupleMarshallerImpl(table.schemaView()).marshal(Tuple.create().set("id", 42));
 
         if (isNode0Leader) {
-            assertNull(table.internalTable().get(key, new HybridClockImpl().now(), node1.node()).get());
+            assertNull(table.internalTable().get(key, node1.clock().now(), node1.node()).get());
         } else {
-            assertNull(table.internalTable().get(key, new HybridClockImpl().now(), node0.node()).get());
+            assertNull(table.internalTable().get(key, node1.clock().now(), node0.node()).get());
         }
 
         var tx = node0.transactions().begin();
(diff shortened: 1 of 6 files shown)


modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
```diff
@@ -175,9 +174,9 @@ private void restartClusterWithNotAppliedCommands(
         BinaryRowEx key = new TupleMarshallerImpl(table.schemaView()).marshal(Tuple.create().set("id", 42));
 
         if (isNode0Leader) {
-            assertNull(table.internalTable().get(key, new HybridClockImpl().now(), node1.node()).get());
+            assertNull(table.internalTable().get(key, node1.clock().now(), node1.node()).get());
         } else {
-            assertNull(table.internalTable().get(key, new HybridClockImpl().now(), node0.node()).get());
+            assertNull(table.internalTable().get(key, node1.clock().now(), node0.node()).get());
```
sanpwc, line 179: Should we use node0.clock().now() here or using node1 is fine?

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -944,4 +944,14 @@ public BiPredicate<String, NetworkMessage> dropMessagesPredicate() {
     public void stopDroppingMessages() {
         ((DefaultMessagingService) clusterSvc.messagingService()).stopDroppingMessages();
     }
+
+    /**
+     * Gets a hybrid clock specific of the node.
```
sanpwc, line 949: "specif of the node", sounds a bit unnatural from my point of view.
I'd rather use "Returns the node's hybrid clock." or similar.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
vldpyatkov: Here it has no difference.
The only thing that matters is that the data was applied finally.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -486,11 +487,23 @@ private CompletableFuture<BinaryRow> processReadOnlySingleEntryAction(ReadOnlySi
...
+        CompletableFuture<Void> safeReadFuture = isPrimaryInTimestamp(isPrimary, readTimestamp) ? completedFuture(null)
+                : safeTime.waitFor(request.readTimestamp());
 
         return safeReadFuture.thenCompose(unused -> resolveRowByPkForReadOnly(searchRow, readTimestamp));
     }
 
+    /**
+     * Checks that the node is primary in a passed timestamp.
+     *
+     * @param isPrimary True is the node is primary, false otherwise.
```
alievmirza, line 499:
```suggestion
     * @param isPrimary True if the node is primary, false otherwise.
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -486,11 +487,23 @@ private CompletableFuture<BinaryRow> processReadOnlySingleEntryAction(ReadOnlySi
...
+     * @param timestamp Timestamp to check.
+     * @return True if the timestamp is already come and node in primary, otherwise false.
```
alievmirza, line 501:
```suggestion
     * @return True if the timestamp is already passed in the reference system of the current node and node is primary, false otherwise.
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -486,11 +487,23 @@ private CompletableFuture<BinaryRow> processReadOnlySingleEntryAction(ReadOnlySi
...
 
-        CompletableFuture<Void> safeReadFuture = isPrimary ? completedFuture(null) : safeTime.waitFor(request.readTimestamp());
+        CompletableFuture<Void> safeReadFuture = isPrimaryInTimestamp(isPrimary, readTimestamp) ? completedFuture(null)
+                : safeTime.waitFor(request.readTimestamp());
 
         return safeReadFuture.thenCompose(unused -> resolveRowByPkForReadOnly(searchRow, readTimestamp));
     }
 
+    /**
+     * Checks that the node is primary in a passed timestamp.
```
alievmirza, line 497:
```suggestion
     * Checks that the node is primary and {@code timestamp} is already passed in the reference system of the current node.
```

PR #1690, opened 2023-02-17
IGNITE-18791 Add SSL Client authentication to REST

PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-18791

TCP Client and JDBC support this feature, now REST does it too.

### DIFF
--- a/modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
+++ b/modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
@@ -21,9 +21,11 @@ import static io.micronaut.context.env.Environment.BARE_METAL;
 
 import io.micronaut.context.ApplicationContext;
 import io.micronaut.http.server.exceptions.ServerStartupException;
+import io.micronaut.http.ssl.ClientAuthentication;
 import io.micronaut.openapi.annotation.OpenAPIInclude;
 import io.micronaut.runtime.Micronaut;
 import io.micronaut.runtime.exceptions.ApplicationStartupException;
+import io.netty.handler.ssl.ClientAuth;
 import io.swagger.v3.oas.annotations.OpenAPIDefinition;
 import io.swagger.v3.oas.annotations.info.Contact;
 import io.swagger.v3.oas.annotations.info.Info;
@@ -31,6 +33,7 @@ import io.swagger.v3.oas.annotations.info.License;
 import java.net.BindException;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import org.apache.ignite.internal.logger.IgniteLogger;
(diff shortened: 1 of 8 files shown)


modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
```diff
@@ -214,12 +217,18 @@ private int mapServerStartupException(ServerStartupException exception) {
     private Map<String, Object> properties(int port, int sslPort) {
         boolean dualProtocol = restConfiguration.dualProtocol().value();
         boolean sslEnabled = restConfiguration.ssl().enabled().value();
+        ClientAuth clientAuth = ClientAuth.valueOf(restConfiguration.ssl().clientAuth().value().toUpperCase());
```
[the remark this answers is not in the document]
PakhomovAlexander: Yes, there are safe. I've added the test for that case.

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
```diff
@@ -228,11 +242,75 @@ private Map<String, Object> properties(int port, int sslPort) {
...
+            return result;
         } else {
             return Map.of("micronaut.server.port", port);
         }
     }
 
+    private static void validateKeyStorePath(String keyStorePath) {
+        if (keyStorePath.trim().isEmpty()) {
+            throw new IgniteException(
+                    Common.UNEXPECTED_ERR,
```
ptupitsyn, line 279: Here and below - I think we should use `ErrorGroups.Common.SSL_CONFIGURATION_ERR`.

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
```diff
@@ -228,11 +242,75 @@ private Map<String, Object> properties(int port, int sslPort) {
...
+                    "Trust store path is not configured. Please check your rest.ssl.keyStore.path configuration."
+            );
+        }
+        if (!Files.exists(Path.of(keyStorePath))) {
```
ptupitsyn, line 283: Do we need those checks for existing file? Can Micronaut take care of this for us?

ptupitsyn, line 283: Here and below - please add a blank line between code blocks for readability (not sure why checkstyle does not catch this).

ptupitsyn (changes requested):

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
PakhomovAlexander: I've tried to leave this for micronaut but the error was weird and hard to understand.

ptupitsyn: Ok, got it.

ptupitsyn (approved):

PR #1692, opened 2023-02-17
IGNITE-18640 Implement placement driver best-effort single actor selector and fail-over

denis-chudov:
https://issues.apache.org/jira/browse/IGNITE-18640


modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -354,7 +369,14 @@ public class IgniteImpl implements Ignite {
...
         );
 
-        placementDriverMgr = new PlacementDriverManager(metaStorageMgr);
+        placementDriverMgr = new PlacementDriverManager(
+                MetastorageGroupId.INSTANCE,
+                clusterSvc,
+                raftConfiguration,
+                cmgMgr::metaStorageNodes,
+                logicalTopologyService,
+                raftExecutorService
```
vldpyatkov, line 378: If you extract the pool, you should drive of lifecycle of the object (I mean invoke shutdown).
But I sure, it is not necessary.

modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
```diff
@@ -418,10 +431,8 @@ private static class ServerEventHandler {
          * @param term Term.
          */
         private synchronized void onLeaderElected(ClusterNode node, long term) {
-            if (onLeaderElectedCallback != null && term > this.term) {
```
vldpyatkov, line 421: Why does the expression (term > this.term) is removed?
There is no sense to notify several times about one term.

modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
```diff
@@ -164,11 +175,12 @@ public static CompletableFuture<RaftGroupService> start(
             PeersAndLearners configuration,
             boolean getLeader,
             ScheduledExecutorService executor,
-            LogicalTopologyService logicalTopologyService
+            LogicalTopologyService logicalTopologyService,
+            boolean notifyOnSubscription
```
vldpyatkov, line 179: Maybe a better way to add this flag to subscribeLeader method.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/PlacementDriverManager.java
```diff
@@ -31,21 +46,94 @@ public class PlacementDriverManager implements IgniteComponent {
...
+                                raftConfiguration,
+                                PeersAndLearners.fromConsistentIds(placementDriverNodes),
+                                true,
+                                raftClientExecutor,
+                                logicalTopologyService,
+                                true
+                            ).thenCompose(client -> {
+                                TopologyAwareRaftGroupService topologyAwareClient = (TopologyAwareRaftGroupService) client;
 
+                                return topologyAwareClient.subscribeLeader(this::onLeaderChange).thenApply(v -> topologyAwareClient);
```
vldpyatkov, line 124: If we subscribe to the leader elected event, where will we unsubscribe?
I think it should be before stop (IgniteComponent#beforeNodeStop).

PR #1701, opened 2023-02-21
IGNITE-18810 Java client: Balance requests across connections

ptupitsyn:
When Java client is connected to more than one server node, send requests to different connections in a round-robin fashion instead of using the same default connection every time. This improves client performance and evens the load among servers.

### DIFF
--- a/modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientMessagePacker.java
+++ b/modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientMessagePacker.java
@@ -26,7 +26,6 @@ import java.nio.ByteBuffer;
 import java.util.BitSet;
 import java.util.UUID;
 import org.apache.ignite.internal.binarytuple.BinaryTupleBuilder;
-import org.msgpack.core.annotations.Nullable;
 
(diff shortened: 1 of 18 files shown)


modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientMessagePacker.java
```diff
@@ -26,6 +26,7 @@
 import java.util.BitSet;
 import java.util.UUID;
 import org.apache.ignite.internal.binarytuple.BinaryTupleBuilder;
+import org.msgpack.core.annotations.Nullable;
```
isapego, line 29: Do we use `Nullable` from msgpack?

isapego (approved):

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientMessagePacker.java
ptupitsyn: Good catch. Even though it seems to work as expected (IDEA picks up those annotations correctly), I've reverted those changes.

PR #1706, opened 2023-02-22
IGNITE-18739 Add index garbage collection

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-18739


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
```diff
@@ -252,10 +254,34 @@ private void tryRemoveFromIndexes(BinaryRow rowToRemove, RowId rowId, Cursor<Rea
...
-    private void removeFromIndex(BinaryRow row, RowId rowId) {
-        for (TableSchemaAwareIndexStorage index : indexes.get().values()) {
-            index.remove(row, rowId);
-        }
+    /**
+     * Tries removing partition's oldest stale entry and its indexes.
+     *
+     * @param lowWatermark Low watermark for the vacuum.
+     */
+    public boolean vacuum(HybridTimestamp lowWatermark) {
```
tkalkirill, line 262: Please describe the return value.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
```diff
@@ -252,10 +254,34 @@ private void tryRemoveFromIndexes(BinaryRow rowToRemove, RowId rowId, Cursor<Rea
...
+            BinaryRowAndRowId vacuumed = storage.pollForVacuum(lowWatermark);
+
+            if (vacuumed == null) {
+                // Nothing was garbage collected.
+                return false;
+            }
+
+            BinaryRow binaryRow = vacuumed.binaryRow();
+
+            assert binaryRow != null;
```
tkalkirill, line 273: Looks like there might be a mistake and we can remove the tombstone.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionDataStorage.java
```diff
@@ -165,6 +166,14 @@ public interface PartitionDataStorage extends ManuallyCloseable {
      */
     Cursor<ReadResult> scanVersions(RowId rowId) throws StorageException;
 
+    /**
+     * Tries to garbage collect the oldest stale entry of the partition.
+     *
+     * @see MvPartitionStorage#pollForVacuum(HybridTimestamp)
+     */
+    @Nullable
+    BinaryRowAndRowId pollForVacuum(HybridTimestamp lowWatermark);
```
tkalkirill, line 175:
```suggestion
    @Nullable BinaryRowAndRowId pollForVacuum(HybridTimestamp lowWatermark);
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/SnapshotAwarePartitionDataStorage.java
```diff
@@ -136,6 +137,11 @@ public Cursor<ReadResult> scanVersions(RowId rowId) throws StorageException {
         return partitionStorage.scanVersions(rowId);
     }
 
+    @Override
+    public BinaryRowAndRowId pollForVacuum(HybridTimestamp lowWatermark) {
```
tkalkirill, line 141:
```suggestion
    public @Nullable BinaryRowAndRowId pollForVacuum(HybridTimestamp lowWatermark) {
```

modules/table/src/testFixtures/java/org/apache/ignite/distributed/TestPartitionDataStorage.java
```diff
@@ -111,6 +112,11 @@ public Cursor<ReadResult> scanVersions(RowId rowId) throws StorageException {
         return partitionStorage.scanVersions(rowId);
     }
 
+    @Override
+    public BinaryRowAndRowId pollForVacuum(HybridTimestamp lowWatermark) {
```
tkalkirill, line 116:
```suggestion
    public @Nullable BinaryRowAndRowId pollForVacuum(HybridTimestamp lowWatermark) {
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
SammyVimes: Tombstones can't be returned from pollForVacuum. They're removed along with a non-tombstone value

tkalkirill: It's interesting that if you add a test to `org.apache.ignite.internal.storage.AbstractMvPartitionStorageGcTest`, then everyone except `RocksDbMvPartitionStorageGcTest` will pass it, in general this is a possible situation.    
```
    @Test
    void testTombstone() {
        addAndCommit(null);

        addAndCommit(TABLE_ROW);

        addAndCommit(TABLE_ROW2);

        BinaryRowAndRowId row = pollForVacuum(HybridTimestamp.MAX_VALUE);

        assertNull(row.binaryRow());
    }
```

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexGcTest.java
```diff
@@ -0,0 +1,340 @@
...
+import org.apache.ignite.internal.storage.index.impl.TestHashIndexStorage;
+import org.apache.ignite.internal.storage.index.impl.TestSortedIndexStorage;
+import org.apache.ignite.internal.table.distributed.replicator.TablePartitionId;
+import org.apache.ignite.internal.util.Cursor;
+import org.jetbrains.annotations.Nullable;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/** Tests indexes cleaning up on garbage collection. */
+public class IndexGcTest {
```
tkalkirill, line 61: In this and `org.apache.ignite.internal.table.distributed.IndexCleanupTest` (which, by the way, you wrote) there are a lot of similar things, they should be combined, for example, to highlight an abstract class.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexGcTest.java
```diff
@@ -0,0 +1,340 @@
...
+
+        assertEquals(3, getRowVersions(rowId).size());
+        assertThat(pkInnerStorage.allRowsIds(), contains(rowId));
+        assertThat(sortedInnerStorage.allRowsIds(), contains(rowId));
+        assertThat(hashInnerStorage.allRowsIds(), contains(rowId));
+
+        HybridTimestamp afterCommits = CLOCK.now();
+
+        assertTrue(storageUpdateHandler.vacuum(afterCommits));
+        assertTrue(storageUpdateHandler.vacuum(afterCommits));
```
tkalkirill, line 206: Let's add an intermediate check that we only removed one version.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexGcTest.java
```diff
@@ -0,0 +1,340 @@
...
+        assertFalse(storageUpdateHandler.vacuum(afterCommits));
+
+        assertEquals(1, getRowVersions(rowId).size());
+        // Older entries have different indexes, should be removed.
+        assertFalse(inIndex(tableRow1));
+        assertTrue(inIndex(tableRow2));
+    }
+
+    @Test
+    void testRemoveTombstones() {
```
tkalkirill, line 216: I think you need to test the following scenarios:
- null row row
- row null row
- row row null
- current implementation

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexGcTest.java
SammyVimes: We are already checking it in other gc tests I think + we only have three versions and vacuum returns true 2 times, so doesn't it mean that we removed only 2 versions?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
SammyVimes: That's not relevant to this ticket, but I think there's no harm in fixing it

tkalkirill: I do not agree with your reasoning, technically the situation above is possible, it must be corrected at the beginning.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexGcTest.java
tkalkirill: Well, what if your code deletes everything at once except the latest version, where do we make sure of this?

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexGcTest.java
SammyVimes: Sure, why not

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexGcTest.java
SammyVimes: Then the second pollForVacuum should have returned `false`.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
SammyVimes: I've already fixed it :)

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/SnapshotAwarePartitionDataStorage.java
```diff
@@ -131,11 +132,21 @@ public void commitWrite(RowId rowId, HybridTimestamp timestamp) throws StorageEx
...
         return partitionStorage.scanVersions(rowId);
     }
 
+    @Override
+    public @Nullable BinaryRowAndRowId pollForVacuum(HybridTimestamp lowWatermark) {
+        return partitionStorage.pollForVacuum(lowWatermark);
+    }
+
+    /**
+     * Handles the situation when a snapshot is running concurrently with write operations.
```
rpuch, line 144:
```suggestion
     * Handles the situation when snapshots are running concurrently with write operations.
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/outgoing/SnapshotAwarePartitionDataStorage.java
```diff
@@ -131,11 +132,21 @@ public void commitWrite(RowId rowId, HybridTimestamp timestamp) throws StorageEx
...
+     * In case if a row that is going to be changed was not yet sent in the current snapshot,
```
rpuch, line 145:
```suggestion
     * In case if a row that is going to be changed was not yet sent in an ongoing snapshot,
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
tkalkirill: Thanks!

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexBaseTest.java
```diff
@@ -0,0 +1,282 @@
...
+    TestMvPartitionStorage storage;
+    StorageUpdateHandler storageUpdateHandler;
+
+    @BeforeEach
+    void setUp() {
+        UUID pkIndexId = UUID.randomUUID();
+        UUID sortedIndexId = UUID.randomUUID();
+        UUID hashIndexId = UUID.randomUUID();
+
+        pkInnerStorage = new TestHashIndexStorage(null);
```
rpuch, line 106: Should we pass a real descriptor here?

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexBaseTest.java
```diff
@@ -0,0 +1,282 @@
...
+                        pkIndexId, pkStorage,
+                        sortedIndexId, sortedIndexStorage,
+                        hashIndexId, hashIndexStorage
+                )
+        );
+    }
+
+    List<ReadResult> getRowVersions(RowId rowId) {
+        try (Cursor<ReadResult> readResults = storage.scanVersions(rowId)) {
+            return readResults.stream().collect(Collectors.toList());
```
rpuch, line 149: If we import `toList` statically, the code will read more naturally

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageGcTest.java
```diff
@@ -129,4 +129,17 @@ void testManyOldVersions() {
...
+    void testTombstoneFirst() {
+        addAndCommit(null);
+
+        addAndCommit(TABLE_ROW);
+
+        addAndCommit(TABLE_ROW2);
+
+        BinaryRowAndRowId row = pollForVacuum(HybridTimestamp.MAX_VALUE);
+
+        assertNull(row.binaryRow());
```
ibessonov, line 143: Wow, did not expect that! I thought it should be TABLE_ROW.
Why did you choose such weird behavior? Very non-intuitive

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
```diff
@@ -252,10 +254,37 @@ private void tryRemoveFromIndexes(BinaryRow rowToRemove, RowId rowId, Cursor<Rea
...
+                return false;
+            }
+
+            BinaryRow binaryRow = vacuumed.binaryRow();
+
+            if (binaryRow == null) {
+                // That was a tombstone. This can only happen if the oldest version of row is a tombstone, because
+                // consecutive tombstones are removed along with non-tombstone values.
+                return true;
+            }
```
ibessonov, line 278: That's what I was talking about. This should never be a tombstone

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexBaseTest.java
```diff
@@ -0,0 +1,282 @@
...
+            SCHEMA_DESCRIPTOR.column("STRVAL").schemaIndex()
+    };
+
+    private static final BinaryTupleSchema USER_INDEX_SCHEMA = BinaryTupleSchema.createSchema(SCHEMA_DESCRIPTOR, USER_INDEX_COLS);
+
+    private static final BinaryRowConverter USER_INDEX_BINARY_TUPLE_CONVERTER = new BinaryRowConverter(TUPLE_SCHEMA, USER_INDEX_SCHEMA);
+
+    /** Key-value marshaller for tests. */
+    private static final KvMarshaller<TestKey, TestValue> KV_MARSHALLER
+            = MARSHALLER_FACTORY.create(SCHEMA_DESCRIPTOR, TestKey.class, TestValue.class);
```
ibessonov, line 88: `=` placement is inconsistent between different constants. Please make it unified

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexBaseTest.java
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    static BinaryRow defaultRow() {
+        var key = new TestKey(1, "foo");
+        var value = new TestValue(2, "bar");
+
+        return binaryRow(key, value);
+    }
+
+    boolean inIndex(BinaryRow row) {
```
rpuch, line 180: The method actually checks that the row is in all indexes. Should the method be renamed (like `inAllIndexes()`)?

Also, shouldn't we make sure that the row is either in all indices or in none of them?

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexBaseTest.java
```diff
@@ -0,0 +1,282 @@
...
+        handler.handleUpdate(
+                TX_ID,
+                rowUuid,
+                partitionId,
+                row == null ? null : row.byteBuffer(),
+                (unused) -> {}
+        );
+    }
+
+    static BinaryRow binaryRow(TestKey key, TestValue value) {
```
ibessonov, line 165: I wonder if parts of this test can be inherited from somewhere. For example, `org.apache.ignite.internal.storage.BaseMvStoragesTest` is pretty general. Why didn't you use it?

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexGcTest.java
```diff
@@ -0,0 +1,225 @@
...
+    @Test
+    void testRemoveStaleEntriesWithDifferentIndexes() {
+        UUID rowUuid = UUID.randomUUID();
+        RowId rowId = new RowId(1, rowUuid);
+
+        var key1 = new TestKey(1, "foo");
+        var value1 = new TestValue(2, "bar");
+        BinaryRow tableRow1 = binaryRow(key1, value1);
+
+        var key2 = new TestKey(1, "foo");
```
rpuch, line 68: The keys are same, why wouldn't we just use a single `key` variable to highlight the keys' 'sameness'?

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexGcTest.java
```diff
@@ -0,0 +1,225 @@
...
+
+        BinaryRow row = defaultRow();
+
+        addWrite(storageUpdateHandler, rowUuid, row);
+        commitWrite(rowId);
+
+        addWrite(storageUpdateHandler, rowUuid, row);
+        commitWrite(rowId);
+
+        assertEquals(2, getRowVersions(rowId).size());
```
rpuch, line 47: These 4 assertions do not seem to be the main assertions of this test, they just make sure that we prepare the data for a test correctly. Should we have them here, at first place? If you still prefer to keep them, we probably should highlight the 'main' assertions or hide these ones somehow (maybe with a comment? or a method like `expectRowIdAddedToAllIndexesAndHasNVersions()` - [not literally that long])

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexGcTest.java
```diff
@@ -0,0 +1,225 @@
...
+        assertEquals(3, getRowVersions(rowId).size());
+        assertThat(pkInnerStorage.allRowsIds(), contains(rowId));
+        assertThat(sortedInnerStorage.allRowsIds(), contains(rowId));
+        assertThat(hashInnerStorage.allRowsIds(), contains(rowId));
+
+        HybridTimestamp afterCommits = now();
+
+        assertTrue(storageUpdateHandler.vacuum(afterCommits));
+
+        // tableRow1 should still be index, because second write was identical to the first.
```
rpuch, line 90:
```suggestion
        // tableRow1 should still be in index, because second write was identical to the first.
```

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvPartitionStorageGcTest.java
```diff
@@ -129,4 +129,17 @@ void testManyOldVersions() {
         // Nothing else to poll.
         assertNull(pollForVacuum(lowWatermark));
     }
+
+    @Test
+    void testTombstoneFirst() {
```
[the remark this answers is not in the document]
SammyVimes: I mean, this is a GC test, vacuum is implied)

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
```diff
@@ -252,10 +254,34 @@ private void tryRemoveFromIndexes(BinaryRow rowToRemove, RowId rowId, Cursor<Rea
...
     }
 
-    private void removeFromIndex(BinaryRow row, RowId rowId) {
-        for (TableSchemaAwareIndexStorage index : indexes.get().values()) {
-            index.remove(row, rowId);
-        }
+    /**
+     * Tries removing partition's oldest stale entry and its indexes.
+     *
+     * @param lowWatermark Low watermark for the vacuum.
```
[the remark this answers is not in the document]
SammyVimes: I add a `@see` directive

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexBaseTest.java
SammyVimes: Not needed for pk

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexBaseTest.java
SammyVimes: I changed it a little bit, have a look

rpuch (approved):

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexGcTest.java
```diff
@@ -0,0 +1,192 @@
...
+        assertFalse(storageUpdateHandler.vacuum(afterCommits));
+
+        assertEquals(0, getRowVersions(rowId).size());
+        // The last entry was a tombstone, so no indexes should be left.
+        assertTrue(notInAnyIndex(row));
+    }
+
+    @Test
+    @Disabled("https://issues.apache.org/jira/browse/IGNITE-18882")
+    void testRemoveTombstonesNullRowRow() {
```
tkalkirill, line 117: Maybe remove?

tkalkirill (approved):

PR #1709, opened 2023-02-22
IGNITE-18874 Add a JUnit extension for stopping all Ignites after test suite finishes

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18874


modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/junit/StopAllIgnitesAfterTests.java
```diff
@@ -0,0 +1,93 @@
...
+    @Override
+    public void afterAll(ExtensionContext context) throws Exception {
+        // Try to stop all Ignite nodes via reflection to make sure that this extension does not break anything
+        // even in modules where IgnitionManager or IgnitionImpl are not available (so even for unit tests).
+
+        Class<?> ignitionClass = findClassByName(IGNITION_CLASS_NAME);
+        Class<?> ignitionImplClass = findClassByName(IGNITION_IMPL_CLASS_NAME);
+
+        if (isIgnitionManagerClassAvailable() && ignitionClass != null && ignitionImplClass != null) {
+            String testInstanceName = context.getTestClass().map(Class::getName).orElse("<unknown>");
```
sashapolo, line 61: Should we use `orElseThrow` instead of `orElse("<unknown>")`? When can the test class be absent?

modules/runner/src/main/java/org/apache/ignite/internal/app/IgnitionImpl.java
```diff
@@ -138,6 +140,24 @@ public void stop(String nodeName) {
...
+        List<String> nodeNames = new ArrayList<>(nodes.keySet());
+
+        if (!nodeNames.isEmpty()) {
+            LOG.info("Going to stop the following Ignite instances: " + nodeNames);
+
+            for (String nodeName : nodeNames) {
+                stop(nodeName);
+            }
+
+            LOG.info("Stopped the following Ignite instances: " + nodeNames);
```
sashapolo, line 157: Do we really need this message as well?

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/junit/StopAllIgnitesAfterTests.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private static final String IGNITION_CLASS_NAME = "org.apache.ignite.Ignition";
+
+    private static final String IGNITION_IMPL_CLASS_NAME = "org.apache.ignite.internal.app.IgnitionImpl";
+
+    private static final IgniteLogger LOG = Loggers.forClass(StopAllIgnitesAfterTests.class);
+
+    @Override
+    public void afterAll(ExtensionContext context) throws Exception {
+        // Try to stop all Ignite nodes via reflection to make sure that this extension does not break anything
```
sashapolo, line 54: I don't understand the reason for using reflection here. If this extension is only supposed to be used for Ignite instances that were created using the `Ignition` interface, then these tests should already be dependent on the `api` module. Why can't this extension use these classes directly as well? Or at least not hardcode the class names

modules/runner/src/main/java/org/apache/ignite/internal/app/IgnitionImpl.java
rpuch: Stopping of an Ingite instance might hang. It seems helpful to see the 'Everything is finished' to make sure that nothing has hung.

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/junit/StopAllIgnitesAfterTests.java
rpuch: For example, `ignite-transactions` depends on `ignite-api`, but not on `ignite-runner`, so it sees `IgnitionManager`, but not `IgnitionImpl` (and an attempt to load it will fail the test).

This is an automatically-registered extension, so it might be registered in a module that only has its tiny unit tests and does not have access to even `IgnitionManager`. It seems a nice-to-have feature for this extension to be 'omnivirous': you just don't need to think what other modules reference the module of this extension (and hence, that they pick this extension automatically).

sashapolo: > For example, ignite-transactions depends on ignite-api, but not on ignite-runner, so it sees IgnitionManager, but not IgnitionImpl (and an attempt to load it will fail the test).

Why does it need this extension then? I don't understand... Looks like the problem is that this extension is located in the `core` module and not in the `runner` module. Why is that? Is it because we need to register it in a particular order?

rpuch: For now, we only need it for `runner`, but we might have other modules that will start `Ignite`s in their tests in the future. The idea was to build the extension once and just use it when needed, so I didn't want to implement it in the `runner` module. But `core` does not seem right either.

How about creating a special module for just this extension? We could use it from `runner` tests now; if other modules like `runner` are added, we can just reference the module in their `build` scripts.

Extension ordering (between the auto-registered extensions) seemed to be important when the work was started, but now I think it's not true.

sashapolo: Again, this extension only works if you have `IgnitionManager` and `IgnitionImpl`, so I would expect this extension to be present in the same module. Why would you need it otherwise?

rpuch: Moved the extension to the `runner` module

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/internal/InternalSchemaTest.java
```diff
@@ -26,15 +26,15 @@
...
+import org.apache.ignite.internal.sql.engine.SharedClusterIntegrationTest;
 import org.apache.ignite.internal.testframework.IgniteTestUtils;
 import org.apache.ignite.sql.IgniteSql;
 import org.apache.ignite.sql.ResultSet;
 import org.apache.ignite.sql.Session;
 import org.junit.jupiter.api.Test;
 
 /** Tests for internal manipulations with schema. */
-public class InternalSchemaTest extends AbstractBasicIntegrationTest {
+public class InternalSchemaTest extends SharedClusterIntegrationTest {
```
sashapolo, line 37: Not related to this PR, but this test's name doesn't start with `It`

modules/runner/src/integrationTest/java/org/apache/ignite/internal/TestStartingIgnites.java
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.ignite.internal.testframework.BaseIgniteAbstractTest;
+import org.apache.ignite.internal.testframework.WorkDirectoryExtension;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/**
+ * A test that starts some Ignite instances (and cleans them up later if they are forgotten).
+ */
+// The order is important here.
+@ExtendWith({WorkDirectoryExtension.class, StopAllIgnitesAfterTests.class})
+public abstract class TestStartingIgnites extends BaseIgniteAbstractTest {
```
sashapolo, line 30: This class' name looks ugly to me. Can we introduce a different one? `IgniteIntegrationTest`?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/SharedClusterIntegrationTest.java
```diff
@@ -67,15 +66,13 @@
...
 
 /**
- * Abstract basic integration test.
+ * Abstract basic integration test that starts a cluster once for all the tests it runs.
  */
-@ExtendWith(WorkDirectoryExtension.class)
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
-public class AbstractBasicIntegrationTest extends BaseIgniteAbstractTest {
-    private static final IgniteLogger LOG = Loggers.forClass(AbstractBasicIntegrationTest.class);
+public abstract class SharedClusterIntegrationTest extends TestStartingIgnites {
```
sashapolo, line 74: I would prefer this class to be called `ClusterPerClassIntegrationTest`

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sqllogic/ItSqlLogicTest.java
```diff
@@ -137,11 +135,10 @@
  * @see <a href="https://www.sqlite.org/sqllogictest/doc/trunk/about.wiki">Extended format documentation.</a>
  */
 @Tag(value = "sqllogic")
-@ExtendWith({WorkDirectoryExtension.class, SystemPropertiesExtension.class})
```
sashapolo, line 140: Looks like `SystemPropertiesExtension` is still needed

rpuch: It is registered by this classes' grandfather

rpuch: After thinking a bit, I specified the extension explicitly

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/internal/InternalSchemaTest.java
sashapolo: As you wish

modules/runner/src/integrationTest/java/org/apache/ignite/internal/ClusterPerTestIntegrationTest.java
```diff
@@ -27,23 +27,19 @@
...
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.TestInfo;
 import org.junit.jupiter.api.Timeout;
-import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.platform.commons.support.ReflectionSupport;
 
 /**
- * Abstract integration test that starts and stops a cluster.
+ * Abstract integration test that starts and stops a cluster test method.
```
sashapolo, line 38: Something is not right here

sashapolo (approved):

PR #1710, opened 2023-02-22
IGNITE-18708 ItGeneratedRestClientTest.updateTheSameNodeConfiguration() is flaky

valepakh:
https://issues.apache.org/jira/browse/IGNITE-18708

The flakiness was caused by the JdbcUrlRegistryImpl not stopping after some CLI tests and spamming node configuration API from another thread so the static `Configuration.getDefaultApiClient()` was used from different threads leading to calls to different nodes than intended.


modules/cli/src/main/java/org/apache/ignite/internal/cli/core/ApiClientFactory.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Map;
+import org.apache.ignite.internal.cli.logger.CliLoggers;
+import org.apache.ignite.rest.client.invoker.ApiClient;
+
+/**
+ * This class holds the map from URLs to {@link ApiClient}.
+ */
+@Singleton
+public class ApiClientFactory {
+    private final Map<String, ApiClient> clients = new HashMap<>();
```
Flaugh24, line 31: I suppose it should be ConcurrentHashMap

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/ApiClientFactory.java
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Gets {@link ApiClient} for the base path.
+     *
+     * @param path Base path.
+     * @return created API client.
+     */
+    public ApiClient getClient(String path) {
```
Flaugh24, line 39: What about synchronization? I think it would be better to rewrite it using https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-

Flaugh24 (approved):

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/ApiClientFactory.java
```diff
@@ -0,0 +1,44 @@
...
+        ApiClient apiClient = clients.computeIfAbsent(path, s -> new ApiClient().setBasePath(s));
+        CliLoggers.addApiClient(apiClient);
```
Flaugh24, line 41: Do we really need to do that each time when we call getClient()? 
I think it could be inside computeIfAbsent()

valepakh: It could be done but I don't want to introduce side effects to it. I don't think you can be sure whether it's called in the thread-safe manner or not.

modules/cli/src/main/java/org/apache/ignite/internal/cli/logger/CliLoggers.java
```diff
@@ -59,6 +61,10 @@ public static IgniteLogger forName(String name) {
         return Loggers.forName(name, loggerFactory);
     }
 
+    public static void addApiClient(ApiClient client) {
+        httpLoggers.add(new HttpLogging(client));
```
PakhomovAlexander, line 65: `HttpLogging` does not define `equals` and `hashCode`. So, I think the method `addApiClient ` is not idempotent. I suppose we will add client loggers each time we create a client. Could you confirm that using `--verbose` flag does not duplicate the HTTP logging in this case?

PakhomovAlexander (changes requested):

modules/cli/src/main/java/org/apache/ignite/internal/cli/logger/CliLoggers.java
valepakh: I guess you're right. I can use similar map from path to logger here as in the client factory.

PakhomovAlexander: Or just define `hashcode` and `equals` that use URL as a reference.

PakhomovAlexander (approved):

PR #1711, opened 2023-02-22
[POC] IGNITE-18758 Clients 3.0: Design server-side metrics

ptupitsyn:
https://issues.apache.org/jira/browse/IGNITE-18758

### DIFF
--- a/modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerMetricSource.java
+++ b/modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerMetricSource.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.client.handler;
+
+import java.util.HashMap;
+import org.apache.ignite.internal.metrics.Metric;
+import org.apache.ignite.internal.metrics.MetricSet;
+import org.apache.ignite.internal.metrics.MetricSource;
+import org.jetbrains.annotations.Nullable;
+
+public class ClientHandlerMetricSource implements MetricSource {
+    /** Source name. */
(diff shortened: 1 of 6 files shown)


modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerMetricSource.java
```diff
@@ -0,0 +1,66 @@
...
+package org.apache.ignite.client.handler;
+
+import java.util.HashMap;
+import org.apache.ignite.internal.metrics.LongAdderMetric;
+import org.apache.ignite.internal.metrics.Metric;
+import org.apache.ignite.internal.metrics.MetricSet;
+import org.apache.ignite.internal.metrics.MetricSource;
+import org.jetbrains.annotations.Nullable;
+
+public class ClientHandlerMetricSource implements MetricSource {
```
agura, line 27: Do not implement `MetricSource` interface directly. Instead extend `AbstractMetricSource` because it has all required infrastrcuture (holder and enabled property).

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerMetricSource.java
```diff
@@ -0,0 +1,66 @@
...
+    /** Source name. */
+    private static final String SOURCE_NAME = "client-handler";
+
+    private final LongAdderMetric sessionsTotal = new LongAdderMetric("sessions.total", "Total initiated client sessions");
```
agura, line 31: By design:
>  Metric names should start with an uppercase char and use camel case.

So proper name will be `sessions.Total`

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerMetricSource.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.HashMap;
+import org.apache.ignite.internal.metrics.LongAdderMetric;
+import org.apache.ignite.internal.metrics.Metric;
+import org.apache.ignite.internal.metrics.MetricSet;
+import org.apache.ignite.internal.metrics.MetricSource;
+import org.jetbrains.annotations.Nullable;
+
+public class ClientHandlerMetricSource implements MetricSource {
+    /** Source name. */
+    private static final String SOURCE_NAME = "client-handler";
```
agura, line 29: By design:
> Metrics source name is a combination of names like Java package name separated by dots.

I didn't take into account usage of dashes. I don't think that dashes usage is a problem. But may be will be better to avoid dashes. Not a requirement.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerMetricSource.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private final LongAdderMetric sessionsTotal = new LongAdderMetric("sessions.total", "Total initiated client sessions");
+
+    private boolean enabled;
```
agura, line 33: Concurrency bug: visibility problem. Use `AbstractMetricSource` extension insted of direct implementing of `MetricSource`.

ptupitsyn:
@agura this PR is closed, it is no longer relevant. Please see #1766.

PR #1725, opened 2023-02-27
GG-36164 Added SSL/TLS

AlexLev-Tor:


docs/_docs/ssl-tls.adoc
```diff
@@ -0,0 +1,193 @@
...
+This page explains how to configure SSL/TLS encryption between the cluster nodes (server and client) and the clients that connect to your cluster.
+
+== Considerations
+
+All internal connections in the cluster context, as well as cluster's user interaction interfaces, are SSL-enabled. The communication categories are as follows: 
+
+* Between the user and the cluster (node): REST
+* Between the user and the platform clients:
+
+** CLI
```
PakhomovAlexander, line 12: I am not sure that CLI could be named as a platform client. It is a separate application that communicates with the cluster through REST.

docs/_docs/ssl-tls.adoc
```diff
@@ -0,0 +1,193 @@
...
+** Client
+
+* Between nodes: Network (Messaging, Scalecube)
+
+All SSL configurations activities are performed at the node level.
+
+Due to Micronaut limitations, Apache Ignite does nopt support direct paths to SSL certificates. Instead, it utilizes PKCS12 and JKS keystore.
```
PakhomovAlexander, line 19:
```suggestion
Due to Micronaut limitations, Apache Ignite does not support direct paths to SSL certificates. Instead, it utilizes PKCS12 and JKS keystore.
```

PakhomovAlexander, line 19: Micronaut is an internal implementation of the REST component. I don't think we should highlight this fact in the user documentation.  The Network part is not dependent on micronaut but still utilizes PKCS12 and JKS keystores.

docs/_docs/ssl-tls.adoc
```diff
@@ -0,0 +1,193 @@
...
+//Need a link to Micronaut in general; if possible, to the limitation in particular 
+
+== REST
+
+The standard implementation of SSL support for REST involves configuring a secure connection on a separate port. Apache Ignite also supports the DUAL protocol feature.
+//Need a link to a description of the DUAL protocol.
```
PakhomovAlexander, line 25: I think there is no standard specification for dual protocol. I would say it will be enough to explain that dual protocol means both HTTP and HTTPS are supported on different ports.

docs/_docs/ssl-tls.adoc
```diff
@@ -0,0 +1,193 @@
...
+
+Due to Micronaut limitations, Apache Ignite does nopt support direct paths to SSL certificates. Instead, it utilizes PKCS12 and JKS keystore.
+//Need a link to Micronaut in general; if possible, to the limitation in particular 
+
+== REST
+
+The standard implementation of SSL support for REST involves configuring a secure connection on a separate port. Apache Ignite also supports the DUAL protocol feature.
+//Need a link to a description of the DUAL protocol. 
+
+To configure SSL in Micronaut, all the necessary properties must be defined in the Micronaut notation. Map the Apache Ignite 3.x REST security configuration to the Micronaut one:
```
PakhomovAlexander, line 27: I would like to ask you not to speak about Micronaut. We configure the REST interface/component but not Micronaut.

PakhomovAlexander, line 27: there is no "Micronaut notation", we use HOCON everywhere.

docs/_docs/ssl-tls.adoc
```diff
@@ -0,0 +1,193 @@
...
+----
+SslContextBuilder builder = SslContextBuilder.forServer(certChainInput, keyInput)
+    .ciphers(getCiphers(), getCiphersFilter())
+    .sessionTimeout(serverSslConfig.getSessionTimeout())
+    .sslProvider(sslProvider);                 
+.trustManager(trustedCerts.toArray(new X509Certificate[0]))
+    .clientAuth(serverSslConfig.getClientAuth());
+----
+
+View the link:https://github.com/devsunny/netty-ssl-example/blob/master/src/main/java/com/asksunny/ssl/SecureSocketSslContextFactory.java[Netty SSL configuration].
```
PakhomovAlexander, line 74: I am not sure that this brings any peace of information for users. As well as code examples. It is just internal implementation.

docs/_docs/ssl-tls.adoc
```diff
@@ -0,0 +1,193 @@
...
+client = new OkHttpClient.Builder()
+    .sslSocketFactory(sslSocketFactory, trustManager)
+    .build();
+----
+
+You can enable SSL on the CLI side using the `cli config set` command:
+
+[source,shell]
+----
+cli.trust-store.type=
```
PakhomovAlexander, line 149: This is not the whole command that the user would execute. Could you please type here commands like:

```bash
cli config set cli.trust-store.type=<type>
cli config set cli.trust-store.path=<path>
cli config set cli.trust-store.password=<password>

File: docs/_docs/ssl-tls.adoc [hunk 7]
```diff
@@ -0,0 +1,193 @@
...
+            "path": "must not be empty",
+            "password": "may be empty"
+        }
+    }
+}
+----
+
+== SSL Client Authentication (mTLS Support)
+
+All connection types you utilize must support the client authentication feature. Configute it separately for each connection on the server side.
```
PakhomovAlexander, line 184: Actually, now it is possible to setup the TLS without the client authentication feature. It is optional.

PakhomovAlexander:
@Pochatkin could you take a look?

PR #1730, opened 2023-02-28
IGNITE-17636 Thin 3.0: Avoid BinaryTuple conversion in client handler

ptupitsyn:
* Update `ClientTableCommon.writeTuple` to obtain underlying `BinaryTuple` from `Row` and copy bytes directly to the client without conversion.
* To simplify the logic, send the entire `BinaryTuple` to the client in key-based requests (Get, GetAndUpsert, GetAndDelete, GetAndReplace). Do not extract the value part - it is not possible without rebuilding the binary tuple.
  * Fix Java, .NET, C++ clients

Benchmark does not show any significant performance difference, because other parts of the system are slow (5 ms per `get` is quite a lot).


modules/platforms/dotnet/Apache.Ignite.Benchmarks/Table/TupleGetBenchmarks.cs
```diff
@@ -23,11 +23,12 @@ namespace Apache.Ignite.Benchmarks.Table
...
 
     /// <summary>
-    /// Results on Intel Core i7-7700HQ, .NET SDK 5.0.400, Ubuntu 20.04:
-    /// | Method |     Mean |   Error |  StdDev |
-    /// |------- |---------:|--------:|--------:|
-    /// |    Get | 202.0 us | 3.71 us | 7.82 us |.
+    /// Results on i9-12900H, .NET SDK 6.0.405, Ubuntu 22.04:
+    /// | Method |     Mean |     Error |    StdDev |
+    /// |------- |---------:|----------:|----------:|
+    /// |    Get | 4.788 ms | 0.3198 ms | 0.0830 ms |.
```
isapego, line 29: Have it became slower? Or does "ms" stands for "microseconds"? Probably, "us" should be used in this case.

isapego (approved):

modules/platforms/dotnet/Apache.Ignite.Benchmarks/Table/TupleGetBenchmarks.cs
ptupitsyn: Yes, it has become slower. Previous result is from over a year ago. Units are correct, those are milliseconds.

ptupitsyn: I've updated the benchmark to compare this branch with current master. But we can't see the difference with results like those, and such a deviation.

PR #1731, opened 2023-02-28
IGNITE-18674 Introduce async logic to CLI

PakhomovAlexander:
I've introduced the core functionality for background async calls with a progress bar. The SQL completion hanging is fixed with background async initialization. 

I don't see any reason to use a UI thread with a special API.

https://issues.apache.org/jira/browse/IGNITE-18674

### DIFF
--- a/modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/sql/SqlCompleter.java
+++ b/modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/sql/SqlCompleter.java
@@ -18,7 +18,6 @@
 package org.apache.ignite.internal.cli.commands.sql;
 
 import java.util.List;
-import java.util.concurrent.CompletableFuture;
 import org.apache.ignite.internal.cli.sql.SchemaProvider;
 import org.jline.reader.Candidate;
 import org.jline.reader.Completer;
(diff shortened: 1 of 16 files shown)


modules/cli/src/main/java/org/apache/ignite/internal/cli/core/call/AsyncCallExecutionPipeline.java
```diff
@@ -0,0 +1,82 @@
...
+            return handleResult(result);
+        } catch (CompletionException e) {
+            return handleException(e.getCause());
+        } catch (Exception e) {
+            return handleException(e);
+        }
+    }
+
+    private void print(String s) {
+        output.print("\r" + s);
```
valepakh, line 79: Could you please add a comment explaining this?

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/sql/SqlCompleter.java
```diff
@@ -72,4 +73,9 @@ private static void addCandidate(String string, List<Candidate> candidates) {
         candidates.add(new Candidate(string));
         candidates.add(new Candidate(string.toLowerCase()));
     }
+
+    public void initStateAsync() {
+        // trigger schema loading in background
+        CompletableFuture.supplyAsync(schemaProvider::getSchema);
```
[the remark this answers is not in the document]
PakhomovAlexander: Probably yes

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/call/AsyncCallExecutionPipeline.java
PakhomovAlexander: Sure

valepakh:
LGTM

valepakh (approved):

PR #1732, opened 2023-03-01
IGNITE-18653: Sql. The number of dynamic parameters can be checked once.

lowka:


AMashenkov (approved):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ClusterPerClassIntegrationTest.java
```diff
@@ -411,7 +412,7 @@ protected static List<List<Object>> sql(@Nullable Transaction tx, String sql, Ob
         ));
 
         try {
-            var context = tx != null ? QueryContext.of(tx) : QueryContext.of();
+            var context =  QueryContext.create(QueryPlan.TOP_LEVEL_TYPES, tx);
```
korlov42, line 415: extra space before `QueryContext`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/sql/IgniteSqlParser.java
```diff
@@ -0,0 +1,175 @@
...
+import org.apache.calcite.sql.parser.SqlParser;
+import org.apache.calcite.sql.parser.SqlParserImplFactory;
+import org.apache.calcite.util.SourceStringReader;
+import org.apache.ignite.internal.generated.query.calcite.sql.IgniteSqlParserImpl;
+import org.apache.ignite.sql.SqlException;
+
+/**
+ * Implementation of SQL parser.
+ */
+public final class IgniteSqlParser extends IgniteSqlParserImpl {
```
korlov42, line 39: do we really want to expose extended parser? I would prefer to hide it inside of IgniteSqlParser as a private static class. The same with FACTORY. In that case we will have the only possible way to use new parser - through static methods `parse` as it meant to be used.

Secondly, what do you think about separating two different cases (single statement vs script) in parser as well? This let us avoid annoying validation of size of the returned nodeLIst every time. But we need to change API a bit: instead of IgniteSqlScriptNode extension we need to introduce ParsedResult encapsulating root of AST and count of dynamic parameters. As for now, code like below looks very confusing:

```                    
                    if (nodes.size() > 1) {
                        throw new SqlException(QUERY_INVALID_ERR, "Multiple statements aren't allowed.");
                    }

                    validateParsedStatement(context, nodes, nodes.get(0), params);

                    return nodes.get(0);
```

lowka: Just to clarify - You are talking about removing `extends IgniteSqlParserImpl`  from IgniteSqlParser? Yes, I think it would be better this way.

Regarding the single/multiple statements - I agree that it would be better if single statement/multiple statement cases were using different APIs, and I can fix that.

korlov42: > Just to clarify - You are talking about removing extends IgniteSqlParserImpl from IgniteSqlParser?

yes, I'm

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/FragmentPlan.java
```diff
@@ -45,8 +46,8 @@ public IgniteRel root() {
 
     /** {@inheritDoc} */
     @Override
-    public Type type() {
-        return Type.FRAGMENT;
+    public @Nullable Type type() {
+        return null;
```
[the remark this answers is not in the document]
lowka: Yes, it is.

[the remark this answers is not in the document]
ygerzhedovich: then let's remove it )

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/QueryPlan.java
```diff
@@ -29,6 +31,11 @@ enum Type {
         QUERY, FRAGMENT, DML, DDL, EXPLAIN
     }
 
+    /**
+     * All plan type except for {@code FRAGMENT}.
+     */
+    Set<Type> TOP_LEVEL_TYPES = EnumSet.of(Type.QUERY, Type.DDL, Type.DML, Type.EXPLAIN);
```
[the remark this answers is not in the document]
lowka: Removed QueryPlan.Type and replaced it with SqlQueryType.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/FragmentPlan.java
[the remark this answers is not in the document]
lowka: It no longer exists.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/sql/IgniteSqlParser.java
lowka: Done.

ygerzhedovich (approved):

korlov42 (approved):

PR #1734, opened 2023-03-01
IGNITE-18801 Sql. Add an SQL function that generates a random UUID.

xtern:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/IgniteMethod.java
```diff
@@ -94,7 +95,10 @@ public enum IgniteMethod {
     GREATEST2(IgniteSqlFunctions.class, "greatest2", Object.class, Object.class),
 
     /** See {@link Objects#equals(Object, Object)}. */
-    IS_NOT_DISTINCT_FROM(Objects.class, "equals", Object.class, Object.class);
+    IS_NOT_DISTINCT_FROM(Objects.class, "equals", Object.class, Object.class),
+
+    /** See {@link UUID#randomUUID()}. */
+    RAND_UUID(UUID.class, "randomUUID");
```
lowka, line 101: I think it would be better to name this function RANDOM_UUID, because abbreviations are mostly used in cases when a function is borrowed from an existing domain (such as math, finance, etc) and that name is widely adopted (so the name is defacto standard).

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItUuidTest.java
```diff
@@ -265,4 +266,22 @@ private void checkUuidComparison(Object uuid1, Object uuid2) {
...
+        assertQuery("SELECT COUNT(*) FROM t WHERE uuid_key != uuid_key2").returns(100L).check();
+    }
+
+    @Test
+    @Disabled("https://issues.apache.org/jira/browse/IGNITE-18931")
+    public void testRandomUuidComparison() {
+        assertQuery("SELECT RAND_UUID() = RAND_UUID()").returns(false).check();
+        assertQuery("SELECT RAND_UUID() != RAND_UUID()").returns(true).check();
+        assertQuery("SELECT RAND_UUID() >= RAND_UUID()").returns(true).check();
+        assertQuery("SELECT RAND_UUID() <= RAND_UUID()").returns(true).check();
```
AMashenkov: These checks will passed only if "RAND_UUID() = RAND_UUID()",
but you expect ""SELECT RAND_UUID() = RAND_UUID()" returns 'false' above.

AMashenkov: If you just want to check if RAND_UUID can participate in inequality, the try smth like this:
"SELECT COUNT(*) FROM t WHERE uuid_key > RAND_UUID or uuid_key < RAND_UUID"

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/IgniteMethod.java
xtern: I named it `RAND_UUID` because we have built-in `RAND` and `RAND_INTEGER`.
If you insist that it should be renamed, I will.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItUuidTest.java
xtern: Thanks, I just wanted to add checks that ">=" and "<=" comparisons would be correctly optimized, but the tests themselves are wrong, so I removed them.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItUuidTest.java
```diff
@@ -158,9 +159,9 @@ public void testBasicAggregates() {
     public void testUuidCaseExpression() {
         UUID other = UUID.randomUUID();
 
-        sql("INSERT INTO t VALUES (1, ?)", UUID_1);
-        sql("INSERT INTO t VALUES (2, ?)", UUID_2);
-        sql("INSERT INTO t VALUES (3, NULL)");
+        sql("INSERT INTO t (id, uuid_key) VALUES (1, ?)", UUID_1);
+        sql("INSERT INTO t (id, uuid_key) VALUES (2, ?)", UUID_2);
+        sql("INSERT INTO t (id, uuid_key) VALUES (3, NULL)");
```
ygerzhedovich, line 164: I would prefer to have a single INSERT statement for the three values, like `INSERT INTO t (id, uuid_key) VALUES (1, ?), (2,?), (3,?)`, it could catch additional problems.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItUuidTest.java
```diff
@@ -239,7 +238,7 @@ public void testUuidTypeCoercion() {
 
     @Test
     public void testTypeCoercionInDml() {
-        sql("INSERT INTO t VALUES (1, ?)",  UUID_1.toString());
+        sql("INSERT INTO t (id, uuid_key) VALUES (1, ?)",  UUID_1.toString());
```
ygerzhedovich, line 241: the following case doesn't work:
`INSERT INTO t (id, uuid_key) VALUES (10, null), (20, '" + UUID.randomUUID() + "')`

lowka: @ygerzhedovich 
It should not work because types in VALUES are do not match:
`
sql("CREATE TABLE integers(i INTEGER primary key, v2 INTEGER)");

sql("INSERT INTO integers VALUES (1, null), (4, '4')"); //1
INSERT INTO integers VALUES (2, 2), (3, '3') // 2
`
Both 1 and 2 Trigger Values passed to VALUES operator must have compatible types.

But in case of a custom data type there is a bug that prevents this error from triggering. I already have a fix and file an issue later.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItUuidTest.java
xtern: Simplified to a single INSERT statement.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/IgniteMethod.java
xtern: btw, h2 has `RAND` and `RANDOM_UUID`

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItUuidTest.java
AMashenkov: `WHERE uuid_key > RAND_UUID or uuid_key < RAND_UUID`
is the same as
`WHERE uuid_key != RAND_UUID`
but uses range checks.

xtern: Maybe I don't quite understand or something is not working right.
1. currently each call to RAND_UUID will generate different value
2. UUID is comparable

So `SELECT COUNT(*) FROM t WHERE uuid_key > RAND_UUID or uuid_key < RAND_UUID` 
should return from 0 to 100% of the number of rows.

ygerzhedovich (approved):

PR #1737, opened 2023-03-01
IGNITE-18933 Add an SSL page to AI3 documentation

AlexLev-Tor:


docs/_docs/ssl-tls.adoc
```diff
@@ -0,0 +1,181 @@
...
+        "keyStore": {
+            "type": "PKCS12",
+            "path": "must not be empty",
+            "password": "may be empty"
+        }
+    }
+}
+----
+
+== Clients and JDBC
```
PakhomovAlexander, line 41: There is no example of the tcp client and jdbc configuration. Do you plan to extend the PR or do it in another one? You can check ItSslTest#ClusterWithSsl#clientCanConnectWithSsl and ItSslTest#ClusterWithSsl#jdbcCanConnectWithSsl for an example.

docs/_docs/ssl-tls.adoc
```diff
@@ -0,0 +1,181 @@
...
+}
+----
+
+== SSL Client Authentication (mTLS Support)
+
+Optionally, the connections you utilize can support the client authentication feature. Configure it separately for each connection on the server side.
+
+Two-way authentication requires that both server and client have certificates they reciprocally trust. The client generates a private key, stores it in its keystore, and gets it signed by an entity the server's truststore trusts.
+
+To support client authentication, a connection must include the `clientAuth` property, which can have of the following values:
```
PakhomovAlexander, line 177: `clientAuth` is not enough for that.  The user also needs to define both `trustStore` and `keyStore` configurations (as well as the server)

PakhomovAlexander: ItSslTest#ClusterWithSslAndClientAuth might be useful.

docs/_docs/thin-client-comparison.csv
```diff
@@ -0,0 +1,11 @@
...
+Record Binary View,{yes},{yes},{yes}
+Key-Value Binary View,{yes},{yes},No
+Record View,{yes},{yes},No
+Key-Value View,{yes},{yes},No
+SQL API,{yes},{yes},{yes}
+Partition Awareness,{yes},{yes},No
+Transactions,{yes},{yes},{yes}
+Compute API,{yes},{yes},No
+Retry Policy,{yes},{yes},No
+Heartbeats,{yes},{yes},No
```
PakhomovAlexander, line 11: could you please add an extra end of line here?

docs/_docs/ssl-tls.adoc
AlexLev-Tor: Aleksandr, I'm a tech writer, not a developer :) Please provide the example code snippets or the exact links to these code snippets.

docs/_docs/ssl-tls.adoc
AlexLev-Tor: Aleksandr, same response: please provide the missing code snippets or links to them. I was following the Wiki page we had discussed. Whatever I found there is now here, the rest I need from a subject matter expert - in this case, you.

PR #1738, opened 2023-03-01
IGNITE-18850 Add ssl configuration validation

valepakh:
https://issues.apache.org/jira/browse/IGNITE-18850


modules/network/src/test/java/org/apache/ignite/internal/network/configuration/SslConfigurationValidatorImplTest.java
```diff
@@ -0,0 +1,116 @@
...
+    @Test
+    public void nullTrustStorePath(@TempDir Path tmpDir) throws IOException {
+        validate(createTrustStoreConfig(tmpDir, "PKCS12", null, "changeIt"),
+                "Key store path must not be blank");
+    }
+
+    @Test
+    public void emptyTrustStorePath(@TempDir Path tmpDir) throws IOException {
+        validate(createTrustStoreConfig(tmpDir, "PKCS12", "", "changeIt"),
+                "Key store path must not be blank");
```
PakhomovAlexander, line 65: I think it is better to print trust store in those cases, not key store.

PakhomovAlexander (commented):
Have you tested the patch manually or with integration tests?

valepakh:
> Have you tested the patch manually or with integration tests?

Not sure what do you mean, integration tests were run during the check.

PakhomovAlexander (approved):
LGTM

modules/network/src/main/java/org/apache/ignite/internal/network/configuration/SslConfigurationValidatorImpl.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.ignite.configuration.validation.ValidationContext;
+import org.apache.ignite.configuration.validation.ValidationIssue;
+import org.apache.ignite.configuration.validation.Validator;
+
+/**
+ * SSL configuration validator implementation.
+ */
+public class SslConfigurationValidatorImpl implements Validator<SslConfigurationValidator, SslView> {
+
+    public static final SslConfigurationValidatorImpl INSTANCE = new SslConfigurationValidatorImpl();
```
ibessonov, line 34: I never understood, why do we need a static instance. I understand that you just replicated the approach, but nonetheless, I'm curious

modules/network/src/main/java/org/apache/ignite/internal/network/configuration/SslConfigurationValidatorImpl.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public void validate(SslConfigurationValidator annotation, ValidationContext<SslView> ctx) {
+        SslView ssl = ctx.getNewValue();
+        if (ssl.enabled()) {
+            validateKeyStore(ctx, ".keyStore", "Key store", ssl.keyStore());
+
+            ClientAuth clientAuth = ClientAuth.valueOf(ssl.clientAuth().toUpperCase());
```
ibessonov, line 42: I think it'll throw an exception if enum element is not found, is that the expected behavior?

modules/network/src/main/java/org/apache/ignite/internal/network/configuration/SslConfigurationValidator.java
```diff
@@ -23,9 +23,9 @@
 import java.lang.annotation.Target;
 
 /**
- * Annotation to validate whole key store configuration.
+ * Annotation to validate SSL configuration.
  */
 @Target(ElementType.FIELD)
 @Retention(RetentionPolicy.RUNTIME)
-public @interface KeyStoreConfigurationValidator {
+public @interface SslConfigurationValidator {
```
ibessonov, line 30: I feel like the annotation is unused

modules/network/src/main/java/org/apache/ignite/internal/network/configuration/SslConfigurationValidatorImpl.java
valepakh: I added the check for this and also for the possible exception when checking the path.

modules/network/src/main/java/org/apache/ignite/internal/network/configuration/SslConfigurationValidator.java
valepakh: Nice catch, fixed!

ibessonov (approved):

PR #1739, opened 2023-03-01
IGNITE-18882 Don't store tombstones if there is no other version of row

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-18882

### DIFF
--- a/modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
+++ b/modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
@@ -273,8 +273,7 @@ public class TestMvPartitionStorage implements MvPartitionStorage {
             gcQueue.add(committedVersionChain);
         } else {
             if (committedVersionChain.row == null) {
-                // If previous version doesn't exist and current version is a tombstone,
-                // then just ignore it.
+                // If there is only one version, and it is a tombstone, then remove the chain.
                 return null;
             }
         }
--- a/modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
+++ b/modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
@@ -259,7 +259,7 @@ public class TestMvPartitionStorage implements MvPartitionStorage {
         });
     }
 
-    private VersionChain resolveCommittedVersionChain(VersionChain committedVersionChain) {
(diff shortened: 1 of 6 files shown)


modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestMvPartitionStorage.java
```diff
@@ -271,6 +271,12 @@ private VersionChain resolveCommittedVersionChain(VersionChain committedVersionC
             // Calling it from the compute is fine. Concurrent writes of the same row are impossible, and if we call the compute closure
             // several times, the same tuple will be inserted into the GC queue (timestamp and rowId don't change in this case).
             gcQueue.add(committedVersionChain);
+        } else {
+            if (committedVersionChain.row == null) {
+                // If previous version doesn't exist and current version is a tombstone,
```
tkalkirill, line 276: I would paraphrase in: 
`if only one version, and it is a tombstone, then we will remove such a chain.`

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/AddWriteCommittedInvokeClosure.java
```diff
@@ -81,6 +81,14 @@ public void call(@Nullable VersionChain oldRow) throws IgniteInternalCheckedExce
             throw new StorageException("Write intent exists: [rowId={}, {}]", oldRow.rowId(), storage.createStorageInfo());
         }
 
+        if (row == null && oldRow == null) {
+            // If previous version doesn't exist and current version is a tombstone,
```
tkalkirill, line 85: I would paraphrase in:
`If only one version, and it is a tombstone, then we will not save such a chain.`

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/CommitWriteInvokeClosure.java
```diff
@@ -80,6 +80,13 @@ public void call(@Nullable VersionChain oldRow) throws IgniteInternalCheckedExce
         RowVersion current = storage.readRowVersion(oldRow.headLink(), DONT_LOAD_VALUE);
         RowVersion next = oldRow.hasNextLink() ? storage.readRowVersion(oldRow.nextLink(), DONT_LOAD_VALUE) : null;
 
+        if (next == null && current.isTombstone()) {
+            // Previous version doesn't exist and current version is a tombstone, delete current version.
```
tkalkirill, line 84: I would paraphrase in:
`If only one version, and it is a tombstone, then we will remove such a chain.`

tkalkirill (approved):

PR #1743, opened 2023-03-02
IGNITE-18749 Sql. Planner optimization for JOIN

korlov42:
The patch introduces a query processing step at which the query is analysed, and if the size of the biggest join exceeds some threshold (hardcoded to 5; the choice made upon the results of `TpchPrepareBenchmark`), join order optimisation rules (like `JOIN_COMMUTE*`) will be excluded from the main optimisation phase.


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PlannerHelper.java
```diff
@@ -108,4 +136,49 @@ public static IgniteRel optimize(SqlNode sqlNode, IgnitePlanner planner) {
...
+     * A shuttle to estimate a biggest join to optimize.
+     *
+     * <p>There are only two rules: <ol>
+     *     <li>Each achievable leaf node contribute to join complexity, thus must be counted</li>
+     *     <li>If this shuttle reach the {@link LogicalCorrelate} node, only left shoulder will be
+     *         analysed by this shuttle. For right shoulder a new shuttle will be created, and maximum
+     *         of previously found subquery and current one will be saved</li>
+     * </ol>
+     */
+    public static class JoinSizeFinder extends RelHomogeneousShuttle {
```
ygerzhedovich, line 150: Can it be a private class?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestBuilders.java
```diff
@@ -311,6 +311,18 @@ private static class TableBuilderImpl extends AbstractTableBuilderImpl<TableBuil
         /** {@inheritDoc} */
         @Override
         public TestTable build() {
+            if (distribution == null) {
+                throw new IllegalArgumentException("distribution is not specified");
```
ygerzhedovich, line 315:
```suggestion
                throw new IllegalArgumentException("Distribution is not specified");
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestBuilders.java
```diff
@@ -311,6 +311,18 @@ private static class TableBuilderImpl extends AbstractTableBuilderImpl<TableBuil
...
+            }
+
+            if (name == null) {
+                throw new IllegalArgumentException("name is not specified");
```
ygerzhedovich, line 319:
```suggestion
                throw new IllegalArgumentException("Name is not specified");
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestBuilders.java
```diff
@@ -311,6 +311,18 @@ private static class TableBuilderImpl extends AbstractTableBuilderImpl<TableBuil
...
+            }
+
+            if (columns.isEmpty()) {
+                throw new IllegalArgumentException("table must contain at least one column");
```
ygerzhedovich, line 323:
```suggestion
                throw new IllegalArgumentException("Table must contain at least one column");
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/JoinCommutePlannerTest.java
```diff
@@ -204,4 +212,96 @@ public void testInnerCommute() throws Exception {
 
         assertTrue(costWithCommute.isLt(costWithoutCommute));
     }
+
+    /**
+     * The test verifies that queries with big join can be planned for a reasonable amount of time, thus all assertions
+     * ensure that planer return anything but null. The "reasonable amount of time" here is a timeout of the test.
+     * That is, with enabled {@link CoreRules#JOIN_COMMUTE}, optimization of join with a few dozens of tables
+     * will take eternity. Thus, if the test managed to finish before it was killed, it can be considered a success.
+     */
```
ygerzhedovich, line 221: The test verifies that queries with a considerable number of joins can be planned for a reasonable amount of time, thus all assertions ensure that the planer returns anything but null. The "reasonable amount of time" here is a timeout of the test. With enabled {@link CoreRules#JOIN_COMMUTE}, optimization of joining with a few dozens of tables will take an eternity. Thus, if the test finished before it was killed, it can be considered a success.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgnitePlanner.java
```diff
@@ -490,7 +491,7 @@ public void setDisabledRules(Set<String> disabledRuleNames) {
             List<RelOptRule> newSet = new ArrayList<>();
 
             for (RelOptRule r : rulesSet) {
-                if (!disabledRuleNames.contains(shortRuleName(r.toString()))) {
+                if (!disabledRuleNames.contains(shortRuleName(r))) {
```
ygerzhedovich, line 494: I would ask you to add real javadoc to any public methods which touched during modifications

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestBuilders.java
korlov42: fixed

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestBuilders.java
korlov42: fixed

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestBuilders.java
korlov42: fixed

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/JoinCommutePlannerTest.java
korlov42: updated

ygerzhedovich (approved):

PR #1745, opened 2023-03-02
IGNITE-18529 Deal with deleting a table/partition and rebalancing partitions

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18529


modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/MvPartitionStorages.java
```diff
@@ -442,6 +403,8 @@ private void throwExceptionDependingOnOperation(StorageOperation operation, int
             throw new StorageRebalanceException(createStorageInProgressOfFinishRebalanceErrorMessage(partitionId));
         } else if (operation instanceof CleanupStorageOperation) {
             throw new StorageException(createStorageInProgressOfCleanupErrorMessage(partitionId));
+        } else if (operation instanceof CloseStorageOperation || operation.isFinalOperation()) {
```
ibessonov, line 406: I am glad that there are individual classes for different operations.
But, I am not glad that you have a bunch of `if-else` statements instead of using polymorphism.
What's the point of having classes then?

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/StorageOperation.java
```diff
@@ -25,16 +25,55 @@
...
+
+    /**
+     * Marks the operation as final.
+     */
+    void markFinalOperation();
+
+    /**
+     * Abstract operation of the storage.
+     */
+    abstract class AbstractStorageOperation implements StorageOperation {
```
ibessonov, line 46: If all implementations must extend abstract class now, why do you need an interface? Just replace it with abstract class.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/StorageOperation.java
```diff
@@ -25,16 +25,55 @@
...
+        public void markFinalOperation() {
+            finalOperation = true;
+        }
+    }
+
     /**
      * Storage creation operation.
      */
-    class CreateStorageOperation implements StorageOperation {
+    class CreateStorageOperation extends AbstractStorageOperation {
```
ibessonov, line 70: Why are all these classes implemented as internal classes? You can move them to corresponding package.
By the way, why is it located in the "util"?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/AbstractPageMemoryTableStorage.java
```diff
@@ -103,7 +105,7 @@ public void stop() throws StorageException {
         busyLock.block();
 
         try {
-            IgniteUtils.closeAll(mvPartitionStorages.getAllForClose().stream().map(mvPartitionStorage -> mvPartitionStorage::close));
+            IgniteUtils.closeAllManually(mvPartitionStorages.getAllForCloseOrDestroy().get(10, TimeUnit.SECONDS).stream());
```
ibessonov, line 108: Please add a comment about 10 seconds, explaining why it's enough.

ibessonov: Maybe extract a future into a variable, so the code will not look as tight

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
```diff
@@ -358,19 +359,22 @@ public void stop() throws StorageException {
 
         resources.add(writeOptions);
 
-        mvPartitionStorages.getAllForClose().forEach(mvPartitionStorage -> resources.add(mvPartitionStorage::close));
+        try {
+            mvPartitionStorages
+                    .getAllForCloseOrDestroy()
+                    .get(10, TimeUnit.SECONDS)
```
ibessonov, line 365: Same here, please add a comment, why 10 seconds is enough

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/StorageOperation.java
tkalkirill: Why not, fixed it.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/util/StorageOperation.java
tkalkirill: At the moment it is used only in this package, if necessary, we will move it to another package.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/AbstractPageMemoryTableStorage.java
tkalkirill: Try to fix it.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbTableStorage.java
tkalkirill: Try to fix it.

ibessonov (approved):

PR #1748, opened 2023-03-03
IGNITE-18763 Implement parallel Watch processing

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-18763

* Watches are now processed in a parallel fashion. However, a single watch is always notified about one event at a time.
* Configuration Revision listeners are notified by every Meta Storage update (in opposite to being notified only about revisions that are associated with configuration keys). This is needed so that VersionedValue's  can complete their subscription futures faster (since they are now notified more often) and allows independent Watch Listeners to retrieve other components' state based on the received revision.
* Fixed a bug with VersionedValue history trimming.
* Watch listeners are now non-blocking. Removed blocking calls from Watch listeners in Table Manager.


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1522,20 +1522,20 @@ private CompletableFuture<Void> dropTableAsyncInternal(String name) {
...
+                                    if (tableCfg == null) {
+                                        throw new TableNotFoundException(DEFAULT_SCHEMA_NAME, name);
+                                    }
 
-                            tblChg.delete(name);
-                        }).changeIndexes(idxChg -> {
-                            List<String> indicesNames = tablesCfg.indexes().value().namedListKeys();
+                                    tblChg.delete(name);
+                                }).changeIndexes(idxChg -> {
+                                    List<String> indicesNames = tablesCfg.indexes().value().namedListKeys();
```
ibessonov, line 1534: Since you already changed this code, can I ask you to use `chg` instead of `tablesCfg` and maybe extract `chg.indexes()` into a variable if necessary?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1857,17 +1857,22 @@ private RuntimeException convertThrowable(Throwable th) {
     private WatchListener createDistributionZonesDataNodesListener() {
         return new WatchListener() {
             @Override
-            public void onUpdate(WatchEvent evt) {
+            public String id() {
+                return "dst-zones-data-nodes-watch";
```
ibessonov, line 1861: Should this be a constant?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1931,74 +1938,116 @@ public void onError(Throwable e) {
     private WatchListener createPendingAssignmentsRebalanceListener() {
         return new WatchListener() {
             @Override
-            public void onUpdate(WatchEvent evt) {
+            public String id() {
+                return "pending-assignments-watch";
```
ibessonov, line 1942: Or this?

ibessonov (commented):
Can't say I understood how the recovery upon restart works. Is there a more detailed explanation somewhere?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sashapolo: I don't know, what's the point?

modules/core/src/main/java/org/apache/ignite/internal/causality/VersionedValue.java
```diff
@@ -187,34 +189,20 @@ private CompletableFuture<T> getInternal(long causalityToken) {
         long actualToken0 = this.actualToken;
 
         if (history.floorEntry(causalityToken) == null) {
-            throw new OutdatedTokenException(causalityToken, actualToken0, historySize);
+            throw new OutdatedTokenException(causalityToken, actualToken0, maxHistorySize);
         }
```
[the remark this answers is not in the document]
sashapolo: No, it can't because there's an `initFut` variable, so this code will only be executed after history is initialized with something. I believe there's even a test for this situation

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationChanger.java
```diff
@@ -588,6 +599,10 @@ private CompletableFuture<Void> changeInternally0(ConfigurationSource src, long
                 throw new ConfigurationValidationException(validationIssues);
             }
 
+            // "allChanges" map can be empty here in case the given update matches the current state of the local configuration. We
+            // still try to write the empty update, because local configuration can be obsolete. If this is the case, then the CAS will
+            // fail and the update will be recalculated and there is a chance that the new local configuration will produce an non-empty
```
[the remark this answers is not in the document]
sashapolo: yes, you are correct, this is a typo from my side

ibessonov (approved):

denis-chudov (approved):

PR #1749, opened 2023-03-03
IGNITE-18604 Fix memory leak

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-18604


modules/rest-api/src/main/java/org/apache/ignite/internal/rest/RestFactory.java
```diff
@@ -17,8 +17,19 @@
...
  * Factory that produces all beans that is necessary for the controller class.
  */
 public interface RestFactory {
+    /**
+     * Destroy method. All resources of factory implementation must be cleaned and all field must be null-pointed.
+     *      The reason of these requirements is Micronaut design.
+     *      {@link Micronaut#start()} store shutdown hook and capture pointer
+     *      to embedded application {@link io.micronaut.http.server.netty.NettyEmbeddedServer} and as result
+     *      {@link io.micronaut.context.ApplicationContext} will be not collect by GC never.
+     *      All rest factories stored in application context and should be cleaned to prevent memory leak.
```
valepakh, line 32:
```suggestion
     * Destroy method. All resources of the factory implementation must be cleaned and all fields must be set to {@code null}.
     *      The reason of these requirements is Micronaut design.
     *      {@link Micronaut#start()} store shutdown hook and capture a pointer
     *      to the embedded application {@link io.micronaut.http.server.netty.NettyEmbeddedServer} and as a result
     *      {@link io.micronaut.context.ApplicationContext} will never be collected by the GC.
     *      All rest factories stored in the application context should be cleaned to prevent memory leak.
```

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestFactoriesDestroyer.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+@Singleton
+public class RestFactoriesDestroyer implements BeanDestroyedEventListener<RestFactory> {
+    private static final IgniteLogger LOG = Loggers.forClass(RestFactoriesDestroyer.class);
+
+    @Override
+    public void onDestroyed(BeanDestroyedEvent<RestFactory> event) {
+        RestFactory bean = event.getBean();
+        if (bean != null) {
+            LOG.info("Destroy rest factory " + bean);
```
valepakh, line 37: Maybe lower the log level to trace?

PakhomovAlexander: Debug should be ok

ibessonov (approved):
Since I took a look at the code, I mаy as well give the approval. Thank you for fixing the leak!

PR #1752, opened 2023-03-06
IGNITE-18884 .NET: Add TLS support

ptupitsyn:
**Public API Changes**

* Add `IgniteClientConfiguration.SslStreamFactory` - same API as in 2.x
  * Allow the factory to return `null` to disable SSL
* Change `IIgniteClient.GetConnections` to return a list of `IConnectionInfo` (`IClusterNode` + `ISslInfo`)

**Other Changes**

* Fix `ClientHandlerModule` to initialize `SslContext` once on startup:
  * Improves performance (no need to initialize on every client connection)
  * Fails node startup in case of SSL misconfiguration
* Use 4 nodes in `PlatformTestNodeRunner`: 2 without SSL, 1 with SSL, 1 with SSL and client authentication.


modules/platforms/dotnet/Apache.Ignite/SslStreamFactory.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// Default SSL protocols.
+    /// </summary>
+    public const SslProtocols DefaultSslProtocols = SslProtocols.None;
+
+    /// <summary>
+    /// Gets or sets the certificate file (pfx) path.
+    /// <para />
+    /// Java certificates can be converted with <c>keytool</c>:
+    /// <c>keytool -importkeystore -srckeystore thekeystore.jks -srcstoretype JKS
+    /// -destkeystore thekeystore.pfx -deststoretype PKCS12</c>.
```
isapego, line 42: Not sure this is a right place for this kind of info.

isapego (approved):

modules/platforms/dotnet/Apache.Ignite/SslStreamFactory.cs
ptupitsyn: Good point, this was carried over from Ignite 2.x, fixed.

PR #1754, opened 2023-03-06
IGNITE-18878 Sql. Introduce schema synchronization for query processing

zstan:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImpl.java
```diff
@@ -162,12 +162,14 @@ public CompletableFuture<QueryPlan> prepareAsync(SqlNode sqlNode, BaseQueryConte
                 case UNION:
                 case EXCEPT:
                 case INTERSECT:
+                    schemaWait.run();
```
korlov42, line 165: why do we need to wait for schema that is already in context?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -322,6 +328,8 @@ private void onMessage(String nodeName, QueryStartRequest msg) {
         assert nodeName != null && msg != null;
 
         DistributedQueryManager queryManager = queryManagerMap.computeIfAbsent(msg.queryId(), key -> {
+            sqlSchemaManager.waitActualSchema(msg.lastVersion());
```
korlov42, line 331: I'm afraid this is not very god idea to block inside update closure, because you are basically blocking even those queries which don't require latest schema. Moreover, it's possible to avoid thread blocking at all: we need to schedule the task after required schema will be available

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -445,6 +451,7 @@ private void sendFragment(
                     .fragmentDescription(desc)
                     .parameters(ctx.parameters())
                     .txAttributes(txAttributes)
+                    .lastVersion(lastStorageVersion.get())
```
korlov42, line 454: we need to encapsulate this version as version of the IgniteSchema

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/IgniteTableModify.java
```diff
@@ -115,10 +114,9 @@ public IgniteRel clone(RelOptCluster cluster, List<IgniteRel> inputs) {
 
     @Override
     public RelWriter explainTerms(RelWriter pw) {
+        // for correct rel obtaining from ExecutionServiceImpl#physNodesCache.
```
korlov42, line 117: To be honest, I don't understand what the problem is. Could you provide more details in the comment?

zstan: We can hold mapping on outdated rels, mute this fix and check: ItJoinTest#testIsNotDistinctFrom for example, or probably we need to invalidate this cache more carefully.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
zstan: If i correctly understand you, such approach is not working for now, or i miss you idea ? 
`        CompletableFuture<SchemaPlus> fut = sqlSchemaManager.waitActualSchema(msg.lastVersion());

        fut.thenApply((sch) -> {
            DistributedQueryManager queryManager = queryManagerMap.computeIfAbsent(msg.queryId(), key -> {
                BaseQueryContext ctx = createQueryContext(key, msg.schema(), msg.parameters());

                return new DistributedQueryManager(ctx);
            });

            return queryManager;
        }).whenComplete((mgr, e) -> mgr.submitFragment(nodeName, msg.root(), msg.fragmentDescription(), msg.txAttributes()));`

zstan: done

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/message/QueryStartRequest.java
```diff
@@ -54,4 +54,9 @@ public interface QueryStartRequest extends ExecutionContextAwareMessage {
      */
     @Marshallable
     TxAttributes txAttributes();
+
+    /**
+     * Return last schema version, just a stub, need to be removed after IGNITE-18733.
+     */
+    long lastVersion();
```
korlov42, line 61: I believe it's better to rename to `schemaVersion`. Firstly, it's not last version, but version used to prepare the query. Secondly, this field should not be removed after IGNITE-18733, because general rule remains the same: the schema should not be changed during transaction's lifespan

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -321,13 +320,22 @@ private AsyncCursor<List<Object>> executeExplain(ExplainPlan plan) {
...
 
-        DistributedQueryManager queryManager = queryManagerMap.computeIfAbsent(msg.queryId(), key -> {
-            BaseQueryContext ctx = createQueryContext(key, msg.schema(), msg.parameters());
+        CompletableFuture<?> fut = sqlSchemaManager.waitActualSchema(msg.lastVersion());
 
-            return new DistributedQueryManager(ctx);
-        });
+        fut.whenComplete((mgr, e) -> {
+            if (e != null) {
+                LOG.error("Schema operation error", e);
```
korlov42, line 327: if we failed to acquire schema of required version, why should be proceed execution of a query?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -321,13 +320,22 @@ private AsyncCursor<List<Object>> executeExplain(ExplainPlan plan) {
     private void onMessage(String nodeName, QueryStartRequest msg) {
         assert nodeName != null && msg != null;
 
-        DistributedQueryManager queryManager = queryManagerMap.computeIfAbsent(msg.queryId(), key -> {
-            BaseQueryContext ctx = createQueryContext(key, msg.schema(), msg.parameters());
+        CompletableFuture<?> fut = sqlSchemaManager.waitActualSchema(msg.lastVersion());
 
-            return new DistributedQueryManager(ctx);
-        });
+        fut.whenComplete((mgr, e) -> {
```
korlov42, line 325: Normally, there should be no need to wait. So, let's cover this case separately

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
```diff
@@ -663,6 +665,68 @@ public void emptyNodeTest() {
...
+            intRes = res1.next().intValue(0);
+
+            assertEquals(intRes, res2.next().intValue(0));
+
+            res1.close();
+
+            res2.close();
+        }
+
+        // Uncomment after IGNITE-18203
```
AMashenkov, line 718:
```suggestion
        //TODO Uncomment after IGNITE-18203
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/SqlSchemaManagerImpl.java
```diff
@@ -150,58 +151,67 @@ public SchemaPlus schema(@Nullable String schema) {
 
     /** {@inheritDoc} */
     @Override
-    @NotNull
-    public IgniteTable tableById(UUID id, int ver) {
+    public CompletableFuture<?> waitActualSchema(long ver) {
```
AMashenkov, line 154:
```suggestion
    public CompletableFuture<?> actualSchemaAsync(long ver) {
```

zstan: good point, done

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
zstan: ok let`s throw exception here.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/message/QueryStartRequest.java
zstan: done

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/BaseQueryContext.java
```diff
@@ -232,6 +235,10 @@ public long plannerTimeout() {
         return plannerTimeout;
     }
 
+    public long schemaVersion() {
+        return schemaVersion;
```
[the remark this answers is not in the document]
zstan: seems i can`t use unwrap here cause i still need to wait pk indexes creation. Or i can rewrite for waiting indexes and after call schema resolver but seems it a bit complicated and hard to understand..

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -321,13 +321,19 @@ private AsyncCursor<List<Object>> executeExplain(ExplainPlan plan) {
...
-        DistributedQueryManager queryManager = queryManagerMap.computeIfAbsent(msg.queryId(), key -> {
-            BaseQueryContext ctx = createQueryContext(key, msg.schema(), msg.parameters());
+        CompletableFuture<?> fut = sqlSchemaManager.actualSchemaAsync(msg.schemaVersion());
 
-            return new DistributedQueryManager(ctx);
-        });
+        if (fut.isDone()) {
+            submitFragment(nodeName, msg);
+        } else {
+            fut.whenComplete((mgr, ex) -> {
```
AMashenkov, line 329: Do you mean this?
```suggestion
            fut.handle((mgr, ex) -> {
```
Javadoc says
> Unlike method {@link #handle handle},
>  this method is not designed to translate completion outcomes,
>  so the supplied action should not throw an exception.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/SqlSchemaManagerImpl.java
```diff
@@ -150,58 +151,67 @@ public SchemaPlus schema(@Nullable String schema) {
...
         }
         try {
-            IgniteTable table = tablesVv.latest().get(id);
-
-            // there is a chance that someone tries to resolve table before
-            // the distributed event of that table creation has been processed
-            // by TableManager, so we need to get in sync with the TableManager
-            if (table == null || ver > table.version()) {
-                table = awaitLatestTableSchema(id);
+            if (ver == -1) {
```
AMashenkov, line 159: Let's extract a constant NOT_INITIALIZED = -1

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
AMashenkov: I think there should be a guarantee that QueryStartResponse will be sent back in case of error.

zstan:
@AMashenkov i fix all your comments

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -321,13 +320,20 @@ private AsyncCursor<List<Object>> executeExplain(ExplainPlan plan) {
...
         assert nodeName != null && msg != null;
 
-        DistributedQueryManager queryManager = queryManagerMap.computeIfAbsent(msg.queryId(), key -> {
-            BaseQueryContext ctx = createQueryContext(key, msg.schema(), msg.parameters());
+        CompletableFuture<?> fut = sqlSchemaManager.actualSchemaAsync(msg.schemaVersion());
 
-            return new DistributedQueryManager(ctx);
-        });
+        if (fut.isDone()) {
+            submitFragment(nodeName, msg);
```
AMashenkov, line 326:
```suggestion
                 {
                     if (ex != null) {
                         handleError(ex, nodeName, msg);
                         return;
                     }
                submitFragment(nodeName, msg);
            }
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/BaseQueryContext.java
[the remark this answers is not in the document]
korlov42: why don't wait for PK creation during schema resolution?

right now we have to way to acquire the version of the schema: the first one seems more intuitive but not safe, the second one requires additional call and therefore safe. This is a problem imo

korlov42 (approved):

PR #1755, opened 2023-03-06
IGNITE-18972 SSL configuration validation doesn't work

valepakh:
https://issues.apache.org/jira/browse/IGNITE-18972

The added `ItSslConfigurationValidationTest` test only tests for a single incorrect config value, other values are checked in the existing `SslConfigurationValidatorImplTest`.


modules/rest/src/main/java/org/apache/ignite/internal/rest/configuration/RestSslConfigurationValidatorImpl.java
```diff
@@ -0,0 +1,36 @@
...
+package org.apache.ignite.internal.rest.configuration;
+
+import org.apache.ignite.configuration.validation.ValidationContext;
+import org.apache.ignite.configuration.validation.Validator;
+import org.apache.ignite.internal.network.configuration.AbstractSslConfigurationValidator;
+
+/**
+ * REST SSL configuration validator implementation.
+ */
+public class RestSslConfigurationValidatorImpl implements Validator<RestSslConfigurationValidator, RestSslView> {
```
ibessonov, line 27: Why do you need two classes that perform the same validation?

modules/client-handler/src/main/java/org/apache/ignite/client/handler/configuration/ClientHandlerConfigurationModule.java
```diff
@@ -38,4 +41,9 @@ public ConfigurationType type() {
     public Collection<RootKey<?, ?>> rootKeys() {
         return Collections.singleton(ClientConnectorConfiguration.KEY);
     }
+
+    @Override
+    public Set<Validator<?, ?>> validators() {
+        return Set.of(SslConfigurationValidatorImpl.INSTANCE);
```
ibessonov, line 47: This validator is repeated in two modules, you shouldn't do that, please remove one of them

ibessonov (changes requested):

ibessonov (approved):

PakhomovAlexander (approved):

PR #1756, opened 2023-03-06
IGNITE-18900 Sql. Remove hash function creation from ExecutionContext

korlov42:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionContext.java
```diff
@@ -221,28 +213,6 @@ public ClusterNode localNode() {
...
-     *     But row for delete may contain the primary fields only, thus we need to project colocation fields
-     *     on this trimmed row.
-     * @return A hash function.
-     */
-    // this is more like workaround to limit scope of the refactoring,
-    // but it definitely need to be fixed
-    // TODO: https://issues.apache.org/jira/browse/IGNITE-18900
-    public RowHashFunction<RowT> hashFunction(UUID tableId, int[] fields) {
-        return hashFunctionCache.computeIfAbsent(
-                new HashFunctionCacheKey(tableId, fields),
```
zstan, line 241: HashFunctionCacheKey seems unused for now and can be removed.

zstan (approved):
plz remove unused class, all other looks good to me.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionContext.java
korlov42: Good catch!

PR #1759, opened 2023-03-06
IGNITE-18608 Support SSL for REST in CLI

Flaugh24:
https://issues.apache.org/jira/browse/IGNITE-18608


modules/cli/src/main/java/org/apache/ignite/internal/cli/core/exception/handler/SslExceptionHandler.java
```diff
@@ -0,0 +1,35 @@
...
+package org.apache.ignite.internal.cli.core.exception.handler;
+
+import javax.net.ssl.SSLException;
+import org.apache.ignite.internal.cli.core.exception.ExceptionHandler;
+import org.apache.ignite.internal.cli.core.exception.ExceptionWriter;
+
+public class SslExceptionHandler implements ExceptionHandler<SSLException> {
+    @Override
+    public int handle(ExceptionWriter err, SSLException e) {
+        err.write("Could not connect to Ignite server. Check SSL configuration.");
```
PakhomovAlexander, line 27: Could you please use the same API that other ExceptionHandlers use? MessageUiCompontes and others.

Also, here the details of what is wrong might be useful.

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/rest/ApiClientBuilder.java
```diff
@@ -0,0 +1,138 @@
...
+        this.keyStorePassword = keyStorePassword;
+        return this;
+    }
+
+    ApiClientBuilder setTrustStorePath(String trustStorePath) {
+        this.trustStorePath = trustStorePath;
+        return this;
+    }
+
+    ApiClientBuilder setTrustStorePassword(String trustStorePassword) {
```
PakhomovAlexander, line 74: Could you get rid of `set` prefix?

PakhomovAlexander (changes requested):
I think some tests would be useful

PakhomovAlexander (approved):

PR #1761, opened 2023-03-07
IGNITE-18933 Add SSL/TLS config docs

IgGusev:


docs/_docs/ssl-tls.adoc
```diff
@@ -0,0 +1,190 @@
...
+   .keyManager(getKeyManagerFactory())
+   .trustManager(getTrustManagerFactory());
+}
+----
+
+=== Server-side Configuration
+
+[source,java]
+----
+SslContextBuilder builder = SslContextBuilder.forServer(certChainInput, keyInput)
```
PakhomovAlexander, line 60: This java code (and above) are just examples of how the SSL might be configured in Netty. I don't think these implementation details are important for end users. What they are really interested in is the configuration part. 

I would suggest dropping those parts from the document. WDYT?

docs/_docs/ssl-tls.adoc
```diff
@@ -0,0 +1,190 @@
...
+    .clientAuth(serverSslConfig.getClientAuth());
+----
+
+Introduce the client configuration on the Apache Ignite 3.x server side and map it to Netty Security Context:
+
+[source,json]
+----
+"clientConnector": {
+    "ssl": {
+        "enabled": false,
```
PakhomovAlexander, line 74:
```suggestion
        "enabled": true,
```

docs/_docs/ssl-tls.adoc
```diff
@@ -0,0 +1,190 @@
...
+        "clientAuth": "none",
+        "keyStore": {
+            "type": "PKCS12",
+            "path": "must not be empty",
+            "password": "may be empty"
+        },
+        "trustStore": {
```
PakhomovAlexander, line 81: We configure `trustStore` if `clientAuth` is set to `require` or `optional`. In the example, we set `"clientAuth": "none"` which means we do not want mTLS and we do not have to configure `trustStore` part. 

I would suggest changing  `"clientAuth": "none"`  to  `"clientAuth": "require"`  OR remove `trustStore` part. The choice is up to you.

docs/_docs/ssl-tls.adoc
```diff
@@ -0,0 +1,190 @@
...
+
+== Network Configuration
+
+The node network is based on the Netty framework. The configuration is the same as described for the Apache Ignite Client part except for the part that addresses the Apache Ignite 3.x configuration:
+
+[source,json]
+----
+"network": {
+    "ssl": {
+        "enabled": false,
```
PakhomovAlexander, line 151:
```suggestion
        "enabled": true,
```

docs/_docs/ssl-tls.adoc
```diff
@@ -0,0 +1,190 @@
...
+
+To support client authentication, a connection must include the `clientAuth`, `trustStore` and `keyStore` properties. Here is an example of a possible client configuration:
+
+[source,json]
+----
+clientConnector.ssl: {
+  enabled: true,
+  clientAuth: "require",
+  keyStore: {
+    path: "keyStorePath",
```
PakhomovAlexander, line 181: To be consistent with the rest of the documentation I would suggest using "must not be empty" or visa-versa.

PakhomovAlexander (changes requested):
I can see the link to the JDBC configuration but there is no single word about the TCP client. Here is an example of how it might be configured: 

```java
                    SslConfiguration.builder()
                            .enabled(true)
                            .trustStoreType("JKS")
                            .trustStorePath(trustStorePath)
                            .trustStorePassword(password)
                            .clientAuth(REQUIRE)
                            .keyStorePath(keyStorePath)
                            .keyStorePassword(password)
                            .build();

            try (IgniteClient client = IgniteClient.builder()
                    .addresses("localhost:10800")
                    .ssl(sslConfiguration)
                    .build()
            ) {

PakhomovAlexander (approved):

PR #1765, opened 2023-03-08
Implement primary replica side leaseGrant handler

denis-chudov:
https://issues.apache.org/jira/browse/IGNITE-18744


modules/replicator/src/main/java/org/apache/ignite/internal/replicator/Replica.java
```diff
@@ -69,4 +121,96 @@ public CompletableFuture<?> processRequest(ReplicaRequest request) {
...
+     */
+    public CompletableFuture<LeaseGrantedMessageResponse> processLeaseGrantedMessage(LeaseGrantedMessage msg) {
+        return leaderFuture().thenCompose(leader -> {
+            if (hasAcceptedLease(msg.leaseStartTime(), msg.leaseExpirationTime())) {
+                return acceptLease(msg.leaseStartTime(), msg.leaseExpirationTime());
+            } else if (msg.force()) {
+                if (!leader.equals(localNodeSupplier.get())) {
+                    // Replica must wait till safe time reaches the lease start timestamp to make sure that all updates made on the
+                    // group leader are received.
+                    return safeTime.waitFor(msg.leaseStartTime()).thenCompose(v -> {
```
vldpyatkov, line 166: I remember we consider a safe time as a time on leader. But in the new reality it is a time in primary replica, possibly.
If the safe time is replicated from the primary replica, we have no opportunity to use it for primary replica choosing.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/Replica.java
```diff
@@ -17,33 +17,85 @@
...
+            ReplicaListener listener,
+            PendingComparableValuesTracker<HybridTimestamp> safeTime,
+            TopologyAwareRaftGroupService raftClient,
+            Supplier<ClusterNode> localNodeSupplier
     ) {
         this.replicaGrpId = replicaGrpId;
         this.listener = listener;
+        this.safeTime = safeTime;
+        this.raftClient = raftClient;
+        this.localNodeSupplier = localNodeSupplier;
```
vldpyatkov, line 96: Is it really so complicated to use a supplier for the local node?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -106,101 +119,140 @@ public ReplicaManager(
         this.clusterNetSvc = clusterNetSvc;
         this.clock = clock;
         this.messageGroupsToHandle = messageGroupsToHandle;
-        this.handler = (message, senderConsistentId, correlationId) -> {
-            if (!busyLock.enterBusy()) {
-                throw new IgniteException(new NodeStoppingException());
+        this.localNodeSupplier = () -> clusterNetSvc.topologyService().localMember();
+        this.handler = this::onReplicaMessageReceived;
+        this.placementDriverMessageHandler = this::onPlacementDriverMessageReceived;
```
vldpyatkov, line 124: Why do you save the link if it is just a reference to the local method?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -106,101 +119,140 @@ public ReplicaManager(
...
-                });
-            } finally {
-                busyLock.leaveBusy();
+    private void onPlacementDriverMessageReceived(NetworkMessage msg0, String senderConsistentId, @Nullable Long correlationId) {
+        if (!busyLock.enterBusy()) {
+            throw new IgniteException(new NodeStoppingException());
+        }
+
+        try {
+            if (!(msg0 instanceof PlacementDriverReplicaMessage)) {
```
vldpyatkov, line 208: Which other types of messages do you wanting here? I doubt the other type is possible.
It this case, assert is more preferable.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -871,7 +878,7 @@ private void updateAssignmentInternal(ConfigurationNotificationEvent<byte[]> ass
                         .thenCompose(v -> storageUpdateHandlerFut)
                         .thenComposeAsync(v -> {
                             try {
-                                return raftMgr.startRaftGroupService(replicaGrpId, newConfiguration);
+                                return raftMgr.startRaftGroupService(replicaGrpId, newConfiguration, raftGroupServiceFactory);
```
vldpyatkov, line 881: The same as before, too complicated to hold a factory only to pass it to method start.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/Replica.java
```diff
@@ -69,4 +121,96 @@ public CompletableFuture<?> processRequest(ReplicaRequest request) {
...
+     *
+     * @param msg Message to process.
+     * @return Future that contains a result.
+     */
+    public CompletableFuture<LeaseGrantedMessageResponse> processLeaseGrantedMessage(LeaseGrantedMessage msg) {
+        return leaderFuture().thenCompose(leader -> {
+            if (hasAcceptedLease(msg.leaseStartTime(), msg.leaseExpirationTime())) {
+                return acceptLease(msg.leaseStartTime(), msg.leaseExpirationTime());
+            } else if (msg.force()) {
+                if (!leader.equals(localNodeSupplier.get())) {
```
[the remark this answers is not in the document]
denis-chudov: this is unlikely situation but it doesnt break anything, as the leader can be not colocated with the primary

PR #1780, opened 2023-03-10
IGNITE-16948 Multichannel message transfer

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-16948


modules/network-api/src/main/java/org/apache/ignite/network/ChannelType.java
```diff
@@ -0,0 +1,43 @@
...
+package org.apache.ignite.network;
+
+/**
+ * Channel type with unique identifier.
+ */
+public enum ChannelType {
+    DEFAULT((short) 0),
+    DEPLOYMENT_UNITS((short) 1),
+
+    TEST(Short.MAX_VALUE);
```
ibessonov, line 27: This is what I call an abstraction leak. Please invent something else, we can't make API dependent on the implementation

modules/network-api/src/main/java/org/apache/ignite/network/ChannelType.java
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.network;
+
+/**
+ * Channel type with unique identifier.
+ */
+public enum ChannelType {
```
ibessonov, line 23: There's not enough documentation here

modules/network-api/src/main/java/org/apache/ignite/network/MessagingService.java
```diff
@@ -52,7 +52,11 @@ public interface MessagingService {
      * @param msg       Message which should be delivered.
      * @return Future of the send operation.
      */
-    CompletableFuture<Void> send(ClusterNode recipient, NetworkMessage msg);
+    default CompletableFuture<Void> send(ClusterNode recipient, NetworkMessage msg) {
+        return send(recipient, ChannelType.DEFAULT, msg);
+    }
+
+    CompletableFuture<Void> send(ClusterNode recipient, ChannelType type, NetworkMessage msg);
```
ibessonov, line 59: Public methods in api modules should be properly documented

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
```diff
@@ -296,6 +300,61 @@ public void testStopTwice() throws Exception {
...
+
+
+        TestMessage bigMessage = messageFactory.testMessage().msg(bigText).build();
+        TestMessage msg = messageFactory.testMessage().msg("test").build();
+
+        CompletableFuture<Void> send1 = sender.send(new OutNetworkObject(bigMessage, Collections.emptyList()));
+        CompletableFuture<Void> send2 = sender.send(new OutNetworkObject(msg, Collections.emptyList()));
+
+        assertThat(send2, willCompleteSuccessfully());
+        assertThat(send1.isDone(), is(true));
```
ibessonov, line 324: It's interesting, why don't you use `assertTrue`? For code consistency?
I'm ok with current assertion btw, just find it a bit mouthful

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
```diff
@@ -296,6 +300,61 @@ public void testStopTwice() throws Exception {
...
+
+
+        TestMessage bigMessage = messageFactory.testMessage().msg(bigText).build();
+        TestMessage msg = messageFactory.testMessage().msg("test").build();
+
+        CompletableFuture<Void> send1 = sender1.send(new OutNetworkObject(bigMessage, Collections.emptyList()));
+        CompletableFuture<Void> send2 = sender2.send(new OutNetworkObject(msg, Collections.emptyList()));
+
+        assertThat(send2, willCompleteSuccessfully());
+        assertThat(send1.isDone(), is(false));
```
ibessonov, line 352: No, this test will fail sooner or later, can we have something more predictable?
Making the test, that's based purely on the time that it takes to send certain amount of bytes, is not robust enough.

ibessonov (changes requested):

modules/code-deployment/src/main/java/org/apache/ignite/internal/deployunit/DeploymentManagerImpl.java
```diff
@@ -226,7 +227,8 @@ public CompletableFuture<Void> undeployAsync(String id, Version version) {
                 }).thenApply(logicalTopologySnapshot -> {
                     for (ClusterNode node : logicalTopologySnapshot.nodes()) {
                         clusterService.messagingService()
-                                .invoke(node, UndeployUnitRequestImpl.builder()
+                                .invoke(node, DEPLOYMENT_UNITS,
```
PakhomovAlexander, line 230: I think `DEPLOYMENT_UNITS_CHANNEL` is a better naming here. Reading the code now it is not clear that u send the message through the special channel.

modules/network-api/src/main/java/org/apache/ignite/network/ChannelType.java
```diff
@@ -0,0 +1,43 @@
...
+    DEFAULT((short) 0),
+    DEPLOYMENT_UNITS((short) 1),
```
PakhomovAlexander, line 25: add empty line, please.

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
PakhomovAlexander, line 324: I understood what you are testing here but it's not obvious. Could you please change the naming like: `firstSend`, `secondSend`. Because order matters here. 

Also, `bigMessage` is ok but why there is no `smallMassage`?

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
```diff
@@ -296,6 +300,61 @@ public void testStopTwice() throws Exception {
...
+
+        server1.close();
+        server2.close();
```
PakhomovAlexander, line 355: Could you close the servers in AfterEach block? In case the assert is wrong they will hang.

modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectorKey.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.ignite.network.ChannelType;
+
+/**
+ * Composite key object for connectors.
+ *
+ * @param <T> identifier type.
+ */
+public class ConnectorKey<T> {
+    private final T id;
+    private final ChannelType type;
```
PakhomovAlexander, line 29: empty line

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
```diff
@@ -87,6 +94,11 @@ final void tearDown() throws Exception {
         IgniteUtils.closeAll(startedManagers);
     }
 
+    @BeforeAll
+    static void registerChannel() {
+        testChannel = ChannelInfo.generate("ItConnectionManagerTest");
```
ibessonov, line 99: I liked the name `ChannelType` more. Right now I can't understand, what type of "info" it holds

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
```diff
@@ -302,57 +311,56 @@ public void testStopTwice() throws Exception {
...
+            NettySender sender2 = server1.openChannelTo(server2, testChannel).get(3, TimeUnit.SECONDS);
 
-        TestMessage bigMessage = messageFactory.testMessage().msg(bigText).build();
-        TestMessage msg = messageFactory.testMessage().msg("test").build();
+            TestMessage bigMessage = messageFactory.testMessage().msg(bigText).map(map).build();
 
-        CompletableFuture<Void> send1 = sender1.send(new OutNetworkObject(bigMessage, Collections.emptyList()));
-        CompletableFuture<Void> send2 = sender2.send(new OutNetworkObject(msg, Collections.emptyList()));
+            CompletableFuture<Void> send1 = sender1.send(new OutNetworkObject(bigMessage, Collections.emptyList()));
+            CompletableFuture.runAsync(() -> {
```
ibessonov, line 346: You should wait for this future completion before moving to the next test

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
```diff
@@ -302,57 +311,56 @@ public void testStopTwice() throws Exception {
...
     }
 
     @Test
     public void testTwoConnectionTypes() throws Exception {
-        int size = 1000000000;
-        char[] chars = new char[size];
-        Arrays.fill(chars, 'a');
-        String bigText = new String(chars);
-
+        String bigText = IgniteTestUtils.randomString(new Random(), 100000000);
```
ibessonov, line 333: I don't really like that we have a 100Mb message in test. Can you please document the test scenario, and justify why it has to be this big?

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
```diff
@@ -302,57 +311,56 @@ public void testStopTwice() throws Exception {
...
+                for (int i = 0; i < 100; i++) {
+                    sender2.send(new OutNetworkObject(messageFactory.emptyMessage().build(), Collections.emptyList()));
+                }
+            });
 
-        assertThat(send2, willCompleteSuccessfully());
-        assertThat(send1.isDone(), is(false));
+            AtomicBoolean atLeastOneSmallWas = new AtomicBoolean(false);
+            server2.connectionManager.addListener(inNetworkObject -> {
+                System.out.println(inNetworkObject.message().groupType());
```
ibessonov, line 354: I think you forgot to delete this

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
```diff
@@ -302,57 +311,56 @@ public void testStopTwice() throws Exception {
...
+            CompletableFuture.runAsync(() -> {
+                for (int i = 0; i < 100; i++) {
+                    sender2.send(new OutNetworkObject(messageFactory.emptyMessage().build(), Collections.emptyList()));
+                }
+            });
 
-        assertThat(send2, willCompleteSuccessfully());
-        assertThat(send1.isDone(), is(false));
+            AtomicBoolean atLeastOneSmallWas = new AtomicBoolean(false);
+            server2.connectionManager.addListener(inNetworkObject -> {
```
ibessonov, line 353: There must be a method to listen for specific message group

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
```diff
@@ -302,57 +311,56 @@ public void testStopTwice() throws Exception {
...
+                if (inNetworkObject.message().groupType() == EmptyMessageImpl.GROUP_TYPE
+                        && inNetworkObject.message().messageType() == EmptyMessageImpl.TYPE) {
+                    atLeastOneSmallWas.set(true);
+                }
+            });
 
-        server1.close();
-        server2.close();
+            assertThat(send1, willCompleteSuccessfully());
+            assertTrue(atLeastOneSmallWas.get());
```
ibessonov, line 362: I don't understand these assertions. We check that by the time we received big message, at least one small message is received as well?
I think you could also add an assertion that be the time one of the small messages is received, big message is still sending.
Anyway, all of this still feels luck-based, was it really the only way to write the test? If there are no other options, I guess I'll approve it, but generally - I don't like this approach.
Another question - why have you decided to put these tests into "integrationTest" directory? Aren't these supposed to be unit tests?

modules/code-deployment/src/main/java/org/apache/ignite/internal/deployunit/DeploymentManagerImpl.java
```diff
@@ -84,6 +84,8 @@ public class DeploymentManagerImpl implements IgniteDeployment, IgniteComponent
 
     private static final String TMP_SUFFIX = ".tmp";
 
+    private static final ChannelType DEPLOYMENT_CHANNEL = ChannelType.register((short) 1, "DeploymentUnits");
```
ibessonov, line 87: It may even be public and be located in the interface, not implementation, but since no one uses it right now, except for this class, let's leave it here

PR #1783, opened 2023-03-13
IGNITE-19013 Fix ItSslTest on Windows

ademakov:
https://issues.apache.org/jira/browse/IGNITE-19013

### DIFF
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/ssl/ItSslTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/ssl/ItSslTest.java
@@ -63,6 +63,10 @@ public class ItSslTest extends IgniteIntegrationTest {
         }
     }
 
+    static String escapeForWindows(String path) {
+        return path.replace("\\", "\\\\");
+    }
+
     @Nested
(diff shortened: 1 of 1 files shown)


modules/runner/src/integrationTest/java/org/apache/ignite/internal/ssl/ItSslTest.java
```diff
@@ -156,11 +161,11 @@ class ClusterWithSsl {
                 + "      enabled: true,\n"
                 + "      trustStore: {\n"
                 + "        password: \"" + password + "\","
-                + "        path: \"" + trustStorePath + "\""
+                + "        path: \"" + trustStorePath.replace("\\", "\\\\") + "\""
```
PakhomovAlexander, line 164: Could you please move `.replace("\\", "\\\\")` to the self-describing method like `escape(trustStorePath)` or something like this?

ademakov: Moved to `escapeForWindows`.

ademakov:
The fix is already done within another PR:   https://github.com/apache/ignite-3/pull/1779

PR #1786, opened 2023-03-13
IGNITE-18809 Java client: Reconnect secondary endpoints in background

ptupitsyn:
Add `reconnectInterval` property to client configuration, which controls periodic background connect/repair of all known endpoints.

### DIFF
--- a/modules/client/src/main/java/org/apache/ignite/client/IgniteClientConfiguration.java
+++ b/modules/client/src/main/java/org/apache/ignite/client/IgniteClientConfiguration.java
@@ -101,6 +101,7 @@ public interface IgniteClientConfiguration {
 
     /**
      * Gets the background reconnect interval, in milliseconds. Set to {@code 0} to disable background reconnect.
+     * Default is {@link #DFLT_RECONNECT_INTERVAL}.
      *
      * <p>Ignite balances requests across all healthy connections (when multiple endpoints are configured).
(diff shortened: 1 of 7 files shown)


modules/client/src/main/java/org/apache/ignite/client/IgniteClientConfiguration.java
```diff
@@ -93,10 +99,22 @@ public interface IgniteClientConfiguration {
...
+    /**
+     * Gets the background reconnect interval, in milliseconds. Set to {@code 0} to disable background reconnect.
+     *
+     * <p>Ignite balances requests across all healthy connections (when multiple endpoints are configured).
+     * Ignite also repairs connections on demand (when a request is made).
+     * However, "secondary" connections can be lost (due to network issues, or node restarts). This property controls how ofter Ignite
+     * client will check all configured endpoints and try to reconnect them in case of failure.
+     *
+     * @return Background reconnect interval, in milliseconds.
+     */
```
isapego, line 111: Should we mention default value here?

isapego (approved):

modules/client/src/main/java/org/apache/ignite/client/IgniteClientConfiguration.java
ptupitsyn: Fixed.

PR #1796, opened 2023-03-15
IGNITE-19014 Topology aware Raft client creates its own message handler which is not aware of Raft group

denis-chudov:
https://issues.apache.org/jira/browse/IGNITE-19014


modules/placement-driver/src/main/java/org/apache/ignite/internal/raft/client/TopologyAwareRaftGroupService.java
```diff
@@ -72,6 +71,10 @@ public class TopologyAwareRaftGroupService implements RaftGroupService {
     /** Leader election handler. */
     private final ServerEventHandler serverEventHandler;
 
+    private final BiConsumer<ClusterNode, Long> leaderElectionListener;
```
vldpyatkov, line 74: This property is excess.

PR #1797, opened 2023-03-15
IGNITE-19042 Initialization of distribution zone meta storage keys on node start.

sergeyuttsel:
https://issues.apache.org/jira/browse/IGNITE-19042


modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -873,25 +873,25 @@ private void initDataNodesFromVaultManager() {
...
-                                saveDataNodesAndUpdateTriggerKeysInMetaStorage(
-                                        DEFAULT_ZONE_ID,
-                                        appliedRevision,
-                                        logicalTopology
-                                );
+                            // init keys and data nodes for default zone
+                            saveDataNodesAndUpdateTriggerKeysInMetaStorage(
+                                    DEFAULT_ZONE_ID,
+                                    appliedRevision,
+                                    logicalTopology
```
sanpwc, line 882: Are we going to call saveDataNodesAndUpdateTriggerKeysInMetaStorage if vaultEntry == null or  vaultEntry.value() == null? If true
a. Why?
b. What we will use as logicalTopology?

sanpwc: I don't agree. What's the point in writing empty dataNodes and fake appliedRevision? That should be handled within saveDataNodesToMetaStorageOn...

sergeyuttsel: It is easier to initialize it once than to do null checks in saveDataNodesToMetaStorageOnScaleUp and saveDataNodesToMetaStorageOnScaleDown and may be later in other places

sanpwc: Well, the easiest way will be to catch NPE and ignore it. What I wan't to say here is that easier doesn't mean better.

sergeyuttsel: Ok. Fixed.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1093,11 +1094,35 @@ CompletableFuture<Void> saveDataNodesToMetaStorageOnScaleUp(int zoneId, long rev
                     return completedFuture(null);
                 }
 
-                Map<String, Integer> dataNodesFromMetaStorage = fromBytes(values.get(zoneDataNodesKey(zoneId)).value());
+                Entry dataNodesValue = values.get(zoneDataNodesKey(zoneId));
```
sanpwc, line 1097: What we expect to retrieve from the meta storage if there are no corresponding keys?
We will get map with three elements, right? 
The values will be null? or empty? See Entry#empty?
Is it possible to have null/empty dataNodesValue and non null/empty zoneScaleUpChangeTriggerValue?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1093,11 +1094,35 @@ CompletableFuture<Void> saveDataNodesToMetaStorageOnScaleUp(int zoneId, long rev
...
 
-                long scaleUpTriggerRevision = bytesToLong(values.get(zoneScaleUpChangeTriggerKey(zoneId)).value());
+                Entry zoneScaleUpChangeTriggerValue = values.get(zoneScaleUpChangeTriggerKey(zoneId));
 
-                long scaleDownTriggerRevision = bytesToLong(values.get(zoneScaleDownChangeTriggerKey(zoneId)).value());
+                Entry zoneScaleDownChangeTriggerValue = values.get(zoneScaleDownChangeTriggerKey(zoneId));
+
+                Map<String, Integer> dataNodesFromMetaStorage;
```
sanpwc, line 1103: It seems that we have very similar or even exact the same code both in scaleUp/Down, am I right?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZonesUtil.java
```diff
@@ -160,10 +161,23 @@ static CompoundCondition triggerKeyConditionForZonesChanges(long revision, int z
      * @return Update condition.
      */
     static CompoundCondition triggerScaleUpScaleDownKeysCondition(long scaleUpTriggerRevision, long scaleDownTriggerRevision,  int zoneId) {
-        return and(
-                value(zoneScaleUpChangeTriggerKey(zoneId)).eq(ByteUtils.longToBytes(scaleUpTriggerRevision)),
-                value(zoneScaleDownChangeTriggerKey(zoneId)).eq(ByteUtils.longToBytes(scaleDownTriggerRevision))
-        );
+        SimpleCondition scaleUpCondition;
+
+        if (scaleUpTriggerRevision != 0) {
```
sanpwc, line 166: Why do you need this? I believe that we can explicitly generate notExists condition if there are no corresponding keys is ms. Meaning, that we don't need 0 and/or empty map as a signal for emptiness.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1447,4 +1448,20 @@ private static class Augmentation {
             this.addition = addition;
         }
     }
+
+    private static Map<String, Integer> extractDataNodes(Entry dataNodesEntry) {
+        if (!dataNodesEntry.empty()) {
```
sanpwc, line 1453: So, dataNodesEntry won't be null, is that correct?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1447,4 +1448,20 @@ private static class Augmentation {
...
+        } else {
+            return emptyMap();
+        }
+    }
+
+    private static long extractChangeTriggerRevision(Entry revisionEntry) {
+        if (!revisionEntry.empty()) {
+            return bytesToLong(revisionEntry.value());
+        } else {
+            return 0;
```
sanpwc, line 1464: It's not clear, and thus a bit unreliable that zero here is the same zero you use in DistributionZoneUtils
`if (scaleUpTriggerRevision != 0)` Let's use constant, e.g. INITIAL_TRIGGER_REVISION_VALUE or similar.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1447,4 +1448,20 @@ private static class Augmentation {
...
+
+    private static Map<String, Integer> extractDataNodes(Entry dataNodesEntry) {
+        if (!dataNodesEntry.empty()) {
+            return fromBytes(dataNodesEntry.value());
+        } else {
+            return emptyMap();
+        }
+    }
+
+    private static long extractChangeTriggerRevision(Entry revisionEntry) {
```
sanpwc, line 1460: Let's move both extract methods to DistributionZonesUtil, WDYT?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZonesUtil.java
sanpwc: Discussed in person.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
sergeyuttsel: Reduce a duplicated code.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
sergeyuttsel: Fixed. We can use Entry#empty.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
sergeyuttsel: Yes.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZonesUtil.java
```diff
@@ -248,7 +271,7 @@ static Update updateLogicalTopologyAndVersion(Set<String> logicalTopology, long
      * @return Returns a set of data nodes retrieved from data nodes map, which value is more than 0.
      */
     public static Set<String> dataNodes(Map<String, Integer> dataNodesMap) {
-        return dataNodesMap.entrySet().stream().filter(e -> e.getValue() > 0).map(Entry::getKey).collect(Collectors.toSet());
+        return dataNodesMap.entrySet().stream().filter(e -> e.getValue() > 0).map(Map.Entry::getKey).collect(Collectors.toSet());
```
sanpwc, line 274: Let's restore the import java.util.Map.Entry;

sergeyuttsel: Discussed in person.

PR #1799, opened 2023-03-15
IGNITE-18860 Configuration of replicas/partitions number moved from table configuration to zone configuration

kgusakov:


sanpwc (changes requested):

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZonesUtil.java
```diff
@@ -317,4 +321,29 @@ static long extractChangeTriggerRevision(Entry revisionEntry) {
...
+     * @param dstZnsCfg Distribution zones config.
+     * @param zoneId Id of zone.
+     * @return Zone configuration with appropriate zone id.
+     */
+    public static DistributionZoneConfiguration getZoneById(DistributionZonesConfiguration dstZnsCfg, int zoneId) {
+        if (zoneId == DEFAULT_ZONE_ID) {
+            return dstZnsCfg.defaultDistributionZone();
+        }
+
+        for (String name : dstZnsCfg.distributionZones().value().namedListKeys()) {
```
sanpwc, line 337: Seems that we may use
```
ConfigurationUtil#getByInternalId(NamedConfigurationTree<T,?,?>, UUID)
ConfigurationUtil#getByInternalId(NamedListView<N>, UUID)
```

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/configuration/DistributionZoneConfigurationSchema.java
```diff
@@ -41,6 +43,16 @@ public class DistributionZoneConfigurationSchema {
     @Value(hasDefault = true)
     public int zoneId = DEFAULT_ZONE_ID;
 
+    /** Zone partitions. */
+    @Range(min = 0, max = 65_000)
```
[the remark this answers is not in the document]
kgusakov: I don't know, that was the default in table configuration... but I think it was a mistake, fixed

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZonesUtil.java
kgusakov: Sorry, maybe sounds stupid, but I can't find these methods)

PR #1800, opened 2023-03-16
IGNITE-18539 Implement build procedure for new indexes

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-18539


modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -91,33 +112,58 @@ public class IndexManager extends Producer<IndexEvent, IndexEventParameters> imp
...
+
+        int cpus = Runtime.getRuntime().availableProcessors();
+
+        buildIndexExecutor = new ThreadPoolExecutor(
+                cpus,
+                cpus,
+                30,
+                TimeUnit.SECONDS,
+                new LinkedBlockingQueue<>(),
+                new NamedThreadFactory(threadPrefix(nodeName, "build-index"), LOG)
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -534,32 +584,159 @@ public BinaryTuple convert(BinaryRow binaryRow) {
...
             return failedFuture(new IllegalStateException("Should not be called"));
         }
     }
+
+    /**
+     * Initializes the build of the index.
+     */
+    private void initIndexBuildIfNeeded(TableIndexView tableIndexView, TableImpl table) {
+        for (int partitionId = 0; partitionId < table.internalTable().partitions(); partitionId++) {
+            buildIndexExecutor.submit(new BuildIndexTask(table, tableIndexView, partitionId, true));
```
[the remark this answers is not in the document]
tkalkirill: In general, I agree with you, but by the time we get to building an index for a specific partition, a change in assignments may occur, and it is better to check this in the task itself.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -534,32 +584,159 @@ public BinaryTuple convert(BinaryRow binaryRow) {
...
+    }
+
+    /**
+     * Task of building a table index for a partition.
+     *
+     * <p>Only the leader of the raft group will manage the building of the index. Leader sends batches of row IDs via
+     * {@link BuildIndexCommand}, the next batch will only be send after the previous batch has been processed.
+     *
+     * <p>Index building itself occurs locally on each node of the raft group (majority) when processing {@link BuildIndexCommand}. This
+     * ensures that the index build in the raft group is consistent and that the index build is restored after restarting the raft group
```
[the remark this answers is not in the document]
tkalkirill: Add it.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -534,32 +584,159 @@ public BinaryTuple convert(BinaryRow binaryRow) {
...
+        }
+
+        private String createCommonTableIndexInfo() {
+            return "table=" + table.name() + ", tableId=" + table.tableId()
+                    + ", partitionId=" + partitionId
+                    + ", index=" + tableIndexView.name() + ", indexId=" + tableIndexView.id();
+        }
+    }
+
+    private String localNodeConsistentId() {
```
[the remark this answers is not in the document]
tkalkirill: Ok.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -534,32 +584,159 @@ public BinaryTuple convert(BinaryRow binaryRow) {
...
+                if (firstBatch) {
+                    LOG.info("Start building the index: [{}]", createCommonTableIndexInfo());
+                }
+
+                List<RowId> batchRowIds = createBatchRowIds(lastBuildRowId, BUILD_INDEX_ROW_ID_BATCH_SIZE);
+
+                boolean finish = batchRowIds.size() < BUILD_INDEX_ROW_ID_BATCH_SIZE;
+
+                raftGroupService.run(createBuildIndexCommand(batchRowIds, finish))
+                        .thenAccept(unused -> {
```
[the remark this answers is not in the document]
tkalkirill: Fix it.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/MockedStructuresTest.java
```diff
@@ -240,7 +240,7 @@ void before() throws Exception {
 
         tblManager = mockManagers();
 
-        idxManager = new IndexManager(tblsCfg, schemaManager, tblManager);
+        idxManager = new IndexManager("test", tblsCfg, schemaManager, tblManager, cs);
```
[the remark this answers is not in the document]
tkalkirill: Fix it.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMetaStorage.java
```diff
@@ -103,13 +112,85 @@ void putPartitionId(int partitionId) {
...
-        return ByteBuffer.allocate(PARTITION_ID_PREFIX.length + Short.BYTES)
-                .order(ByteOrder.BIG_ENDIAN)
+        return ByteBuffer.allocate(PARTITION_ID_PREFIX.length + PARTITION_ID_SIZE)
+                .order(KEY_BYTE_ORDER)
                 .put(PARTITION_ID_PREFIX)
                 .putShort((short) partitionId)
                 .array();
     }
+
+    static byte[] indexMetaKey(int partitionId, UUID indexId) {
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMetaStorage.java
```diff
@@ -103,13 +112,85 @@ void putPartitionId(int partitionId) {
...
+        ByteBuffer buffer = ByteBuffer.allocate(INDEX_META_KEY_SIZE).order(KEY_BYTE_ORDER);
+
+        buffer.put(INDEX_META_KEY_PREFIX).putShort((short) partitionId);
+
+        putUuid(buffer, indexId);
+
+        return buffer.array();
+    }
+
+    static byte[] indexLastBuildRowId(@Nullable RowId rowId) {
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -91,33 +112,58 @@ public class IndexManager extends Producer<IndexEvent, IndexEventParameters> imp
...
+            ClusterService clusterService
+    ) {
+        this.tablesConfig = Objects.requireNonNull(tablesConfig, "tablesCfg");
         this.schemaManager = Objects.requireNonNull(schemaManager, "schemaManager");
         this.tableManager = tableManager;
+        this.clusterService = clusterService;
+
+        int cpus = Runtime.getRuntime().availableProcessors();
+
+        buildIndexExecutor = new ThreadPoolExecutor(
```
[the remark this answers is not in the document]
tkalkirill: Because `maximumPoolSize == Integer.MAX_VALUE`

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexStorage.java
```diff
@@ -54,4 +55,22 @@ public interface IndexStorage {
      * @throws StorageException If failed to remove data.
      */
     void remove(IndexRow row) throws StorageException;
+
+    /**
+     * Returns last row ID for which the index was built, {@code null} means that the index was built.
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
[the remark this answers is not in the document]
tkalkirill: Since the leader of the raft group is involved in the process of building the index, if we are not the leader at the beginning of the task, then we either left the assignments, or we will receive a command to write from the leader about building the index.

The case of changing the leader of the raft group will be done in a separate ticket.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
```diff
@@ -729,6 +729,82 @@ void testDestroyStartedRebalance() {
         assertThat(tableStorage.destroyPartition(PARTITION_ID), willCompleteSuccessfully());
     }
 
+    @Test
+    void testIndexLastBuildRowId() {
+        MvPartitionStorage mvPartitionStorage = getOrCreateMvPartition(PARTITION_ID);
+
+        IndexStorage sortedIndexStorage = tableStorage.getOrCreateIndex(PARTITION_ID, sortedIdx.id());
```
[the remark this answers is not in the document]
tkalkirill: I prefer to leave it as is.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionDataStorage.java
```diff
@@ -166,6 +167,17 @@ public interface PartitionDataStorage extends ManuallyCloseable {
      */
     Cursor<ReadResult> scanVersions(RowId rowId) throws StorageException;
 
+    /**
+     * Scans the partition and returns a cursor of values at the given timestamp. This cursor filters out committed tombstones, but not
+     * tombstones in the write-intent state.
+     *
+     * @param timestamp Timestamp. Can't be {@code null}.
+     * @return Cursor.
+     * @throws TxIdMismatchException If there's another pending update associated with different transaction id.
```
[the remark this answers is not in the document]
tkalkirill: Copy-paste error, removed it.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -534,32 +584,159 @@ public BinaryTuple convert(BinaryRow binaryRow) {
...
+
+        private List<RowId> createBatchRowIds(RowId lastBuildRowId, int batchSize) {
+            MvPartitionStorage mvPartition = table.internalTable().storage().getMvPartition(partitionId);
+
+            assert mvPartition != null : createCommonTableIndexInfo();
+
+            List<RowId> batch = new ArrayList<>(batchSize);
+
+            for (int i = 0; i < batchSize; i++) {
+                lastBuildRowId = mvPartition.closestRowId(lastBuildRowId.increment());
```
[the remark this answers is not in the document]
tkalkirill: Fix it.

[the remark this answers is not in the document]
tkalkirill: Fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/command/BuildIndexCommand.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.ignite.internal.raft.WriteCommand;
+import org.apache.ignite.internal.table.distributed.TableMessageGroup;
+import org.apache.ignite.network.annotations.Transferable;
+
+/**
+ * State machine command to build a table index.
+ */
+@Transferable(TableMessageGroup.Commands.BUILD_INDEX)
+public interface BuildIndexCommand extends WriteCommand {
+    TablePartitionIdMessage tablePartitionId();
```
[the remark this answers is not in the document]
tkalkirill: try to fix it.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -534,32 +584,159 @@ public BinaryTuple convert(BinaryRow binaryRow) {
...
+     * Task of building a table index for a partition.
+     *
+     * <p>Only the leader of the raft group will manage the building of the index. Leader sends batches of row IDs via
+     * {@link BuildIndexCommand}, the next batch will only be send after the previous batch has been processed.
+     *
+     * <p>Index building itself occurs locally on each node of the raft group (majority) when processing {@link BuildIndexCommand}. This
+     * ensures that the index build in the raft group is consistent and that the index build is restored after restarting the raft group
+     * (not from the beginning).
+     */
+    private class BuildIndexTask implements Runnable {
```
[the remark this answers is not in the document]
tkalkirill: Do it.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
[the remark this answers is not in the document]
sashapolo: okay

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -437,4 +439,33 @@ public void onShutdown() {
...
+            storage.lastApplied(commandIndex, commandTerm);
+
+            storageUpdateHandler.buildIndex(cmd.indexId(), cmd.rowIds(), cmd.finish());
+
+            return null;
+        });
+
+        if (cmd.finish()) {
+            LOG.info(
+                    "Finish building the index: [tableId={}, partitionId={}, indexId={}]",
```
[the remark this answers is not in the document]
sashapolo: But shouldn't we at least print both the "initiating index build process on leader" and "start building index locally" messages? Looks confusing otherwise

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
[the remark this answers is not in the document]
sashapolo: Do we have a test for this situation? I mean, when we start building an index and in the middle of this process a leader changes? I have a feeling that something will still break

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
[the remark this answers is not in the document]
sashapolo: okay

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/hash/PageMemoryHashIndexStorage.java
```diff
@@ -329,4 +340,34 @@ public void startCleanup() {
...
         state.compareAndSet(StorageState.CLEANUP, StorageState.RUNNABLE);
     }
+
+    @Override
+    public @Nullable RowId getLastBuildRowId() {
+        return busy(() -> {
+            throwExceptionIfStorageInProgressOfRebalance(state.get(), this::createStorageInfo);
+
+            try {
+                UUID lastBuildRowIdUuid = indexMetaTree.findOne(new IndexMetaKey(indexDescriptor().id())).lastBuildRowIdUuid();
```
ibessonov, line 350: I don't like that we always read it from the tree, access to cached version would be so much faster. The only time we really must read it is the start

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/meta/UpdateLastBuildRowIdInvokeClosure.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public UpdateLastBuildRowIdInvokeClosure(@Nullable RowId newRowId) {
+        this.newRowId = newRowId;
+    }
+
+    @Override
+    public void call(@Nullable IndexMeta oldRow) throws IgniteInternalCheckedException {
+        assert oldRow != null;
+
+        newRow = new IndexMeta(oldRow.indexId(), oldRow.metaPageId(), newRowId == null ? null : newRowId.uuid());
```
ibessonov, line 47: This is not the best place to write such comment, probably, but can new value be smaller then the old one? On leader reelection, for example?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/sorted/PageMemorySortedIndexStorage.java
```diff
@@ -497,4 +511,34 @@ public void startCleanup() {
...
         state.compareAndSet(StorageState.CLEANUP, StorageState.RUNNABLE);
     }
+
+    @Override
+    public @Nullable RowId getLastBuildRowId() {
+        return busy(() -> {
+            throwExceptionIfStorageInProgressOfRebalance(state.get(), this::createStorageInfo);
+
+            try {
+                UUID lastBuildRowIdUuid = indexMetaTree.findOne(new IndexMetaKey(indexDescriptor().id())).lastBuildRowIdUuid();
```
ibessonov, line 521: Looks copy-pasted, maybe we should have a common super-class for sorted and hash indexes?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/HashIndex.java
```diff
@@ -39,9 +39,12 @@ class HashIndex {
 
     private final ConcurrentMap<Integer, RocksDbHashIndexStorage> storages = new ConcurrentHashMap<>();
 
-    HashIndex(ColumnFamily indexCf, HashIndexDescriptor descriptor) {
+    private final RocksDbMetaStorage metaStorage;
+
+    HashIndex(ColumnFamily indexCf, HashIndexDescriptor descriptor, RocksDbMetaStorage metaStorage) {
         this.indexCf = indexCf;
         this.descriptor = descriptor;
+        this.metaStorage = metaStorage;
```
ibessonov, line 47: I would prefer different name, this one looks too similar to the name of other, completely unrelated, component

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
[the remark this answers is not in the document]
tkalkirill: Building indexes on followers happens when we apply the write command, do you think we need to keep track of the first write command so we can print it to the log? I don't think it's worth it just yet.
Will do later if needed.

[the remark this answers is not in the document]
sashapolo: okay

sashapolo (approved):

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/PartitionDataHelper.java
```diff
@@ -138,14 +142,24 @@ void putRowId(ByteBuffer keyBuffer, RowId rowId) {
         assert rowId.partitionId() == partitionId : rowId;
         assert keyBuffer.order() == KEY_BYTE_ORDER;
 
-        keyBuffer.putLong(normalize(rowId.mostSignificantBits()));
-        keyBuffer.putLong(normalize(rowId.leastSignificantBits()));
+        putUuid(keyBuffer, rowId.uuid());
+    }
+
+    static void putUuid(ByteBuffer buffer, UUID uuid) {
```
ibessonov, line 148: Maybe a "rowIdUuid", because there's also "normalize", which may or may not be necessary for general-purpose UUIDs

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMetaStorage.java
```diff
@@ -103,13 +112,85 @@ void putPartitionId(int partitionId) {
...
+    /**
+     * Puts last row ID for which the index was built, {@code null} means index building is finished.
+     *
+     * @param partitionId Partition ID.
+     * @param indexId Index ID.
+     * @param rowId Row ID.
+     */
+    public void putIndexLastBuildRowId(int partitionId, UUID indexId, @Nullable RowId rowId) {
+        try {
+            metaColumnFamily.put(indexMetaKey(partitionId, indexId), indexLastBuildRowId(rowId));
```
ibessonov, line 124: Batch is encapsulated somewhere inside, right?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbHashIndexStorage.java
```diff
@@ -369,4 +373,26 @@ public void finishCleanup() {
             busyLock.unblock();
         }
     }
+
+    @Override
+    public @Nullable RowId getLastBuildRowId() {
+        return busy(() -> {
+            throwExceptionIfStorageInProgressOfRebalance(state.get(), this::createStorageInfo);
```
ibessonov, line 380: Same comment, it's much more convenient to read local field, rather than the entry from the storage

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
```diff
@@ -416,4 +416,39 @@ public void addToIndexes(@Nullable BinaryRow binaryRow, RowId rowId) {
...
+                    ReadResult next = cursor.next();
+
+                    if (!next.isEmpty()) {
+                        index.put(next.binaryRow(), lastRowId);
+                    }
+                }
+            }
+        }
+
+        assert lastRowId != null || finish : "indexId=" + indexId + ", partitionId=" + partitionId;
```
ibessonov, line 450: Oh, so you explicitly assert that last batch can be empty. Why?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1163,7 +1164,7 @@ private CompletableFuture<?> createTableLocally(long causalityToken, String name
                 partitions, clusterNodeResolver, txManager, tableStorage, txStateStorage, replicaSvc, clock);
 
         // TODO: IGNITE-16288 directIndexIds should use async configuration API
-        var table = new TableImpl(internalTable, lockMgr, () -> CompletableFuture.supplyAsync(() -> directIndexIds()));
+        var table = new TableImpl(internalTable, lockMgr, () -> supplyAsync(this::directIndexIds, ioExecutor));
```
ibessonov, line 1167: Nice, thank you!

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableMessageGroup.java
```diff
@@ -154,5 +155,8 @@ interface Commands {
 
         /** Message type for {@link TablePartitionIdMessage}. */
         short TABLE_PARTITION_ID = 61;
+
+        /** Message type for {@link BuildIndexCommand}. */
+        short BUILD_INDEX = 62;
```
ibessonov, line 160: I guess you didn't notice the pattern. Here commands are 40 and higher. 60+ is for random stuff, like TablePartitionId pojos. Please make it 44 and move it higher.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/command/BuildIndexCommand.java
```diff
@@ -0,0 +1,50 @@
...
+     */
+    UUID indexId();
+
+    /**
+     * Returns row IDs for which to build indexes.
+     */
+    List<UUID> rowIds();
+
+    /**
+     * Returns {@code true} if index building for the partition has completed.
```
ibessonov, line 47: I'd rephrase that, not "completed", but that this batch is the last

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -437,4 +439,33 @@ public void onShutdown() {
...
+
+    /**
+     * Handler for the {@link BuildIndexCommand}.
+     *
+     * @param cmd Command.
+     * @param commandIndex RAFT index of the command.
+     * @param commandTerm RAFT term of the command.
+     */
+    void handleBuildIndexCommand(BuildIndexCommand cmd, long commandIndex, long commandTerm) {
+        // Skips the write command because the storage has already executed it.
```
ibessonov, line 451: I wonder, why do we have this check in every single method, instead of making it once in "onWrite". Any ideas?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+import org.apache.ignite.network.ClusterService;
+
+/**
+ * Class for managing the index building process.
+ */
+class IndexBuilder {
+    private static final IgniteLogger LOG = Loggers.forClass(IndexBuilder.class);
+
+    /** Batch size of row IDs to build the index. */
+    private static final int BUILD_INDEX_ROW_ID_BATCH_SIZE = 100;
```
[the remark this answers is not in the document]
tkalkirill: Do you think this configuration is necessary? 
It seems that this is internal logic that does not require configuration from the outside.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+                TimeUnit.SECONDS,
+                new LinkedBlockingQueue<>(),
+                NamedThreadFactory.create(nodeName, "build-index", LOG)
+        );
+    }
+
+    /**
+     * Stops the index builder.
+     */
+    void stop() {
```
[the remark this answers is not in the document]
tkalkirill: Yes, but the busyLock is external (**IndexManager**), you can make an internal busyLock, wdyt?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+        @Override
+        public void run() {
+            if (!busyLock.enterBusy()) {
+                return;
+            }
+
+            try {
+                InternalTable internalTable = table.internalTable();
+
+                RaftGroupService raftGroupService = internalTable.partitionRaftGroupService(partitionId);
```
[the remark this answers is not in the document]
tkalkirill: `org.apache.ignite.internal.table.distributed.storage.InternalTableImpl#partitionRaftGroupService` throws an `IgniteInternalException` if not found (**null**).

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+
+                if (!isLocalNodeLeader(raftGroupService)) {
+                    // TODO: IGNITE-19053 Must handle the change of leader
+                    return;
+                }
+
+                RowId lastBuildRowId = internalTable.storage().getOrCreateIndex(partitionId, tableIndexView.id()).getLastBuildRowId();
```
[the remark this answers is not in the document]
tkalkirill: There is currently no method for getting an index without creating it.
I can create a ticket for it, wdyt?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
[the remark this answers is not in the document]
tkalkirill: Add it.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+
+                if (lastBuildRowId == null) {
+                    // Index has already been built.
+                    return;
+                }
+
+                if (firstBatch) {
```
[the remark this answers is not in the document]
tkalkirill: The first one, both on a stable topology and after a restart.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+                RaftGroupService raftGroupService = internalTable.partitionRaftGroupService(partitionId);
+
+                if (!isLocalNodeLeader(raftGroupService)) {
+                    // TODO: IGNITE-19053 Must handle the change of leader
+                    return;
+                }
+
+                RowId lastBuildRowId = internalTable.storage().getOrCreateIndex(partitionId, tableIndexView.id()).getLastBuildRowId();
+
+                if (lastBuildRowId == null) {
```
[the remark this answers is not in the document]
tkalkirill: `org.apache.ignite.internal.storage.RowId#lowestRowId`

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+                    return;
+                }
+
+                if (firstBatch) {
+                    LOG.info("Start building the index: [{}]", createCommonTableIndexInfo());
+                }
+
+                List<RowId> batchRowIds = createBatchRowIds(lastBuildRowId, BUILD_INDEX_ROW_ID_BATCH_SIZE);
+
+                boolean finish = batchRowIds.size() < BUILD_INDEX_ROW_ID_BATCH_SIZE;
```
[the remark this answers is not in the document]
tkalkirill: But this will make it possible to coordinate the indices on all members of the raft group, is that bad? 
no additional steps are needed to handle the situation of creating indexes for an empty partition on all members of raft group.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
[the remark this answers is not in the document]
tkalkirill: I do not quite understand why I should rely on the last replicated record if we have already built an index for it?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -91,33 +92,44 @@ public class IndexManager extends Producer<IndexEvent, IndexEventParameters> imp
...
-    public IndexManager(TablesConfiguration tablesCfg, SchemaManager schemaManager, TableManager tableManager) {
-        this.tablesCfg = Objects.requireNonNull(tablesCfg, "tablesCfg");
+    public IndexManager(
+            String nodeName,
+            TablesConfiguration tablesConfig,
+            SchemaManager schemaManager,
+            TableManager tableManager,
+            ClusterService clusterService
+    ) {
+        this.tablesConfig = Objects.requireNonNull(tablesConfig, "tablesCfg");
```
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItBuildIndexTest.java
```diff
@@ -0,0 +1,144 @@
...
+                Set<String> allPeers = Stream.concat(
+                        Stream.of(raftGroupService.leader()),
+                        raftGroupService.peers().stream()
+                ).map(Peer::consistentId).collect(toSet());
+
+                if (!allPeers.contains(clusterNode.name())) {
+                    continue;
+                }
+
+                IndexStorage index = internalTable.storage().getOrCreateIndex(partitionId, indexId);
```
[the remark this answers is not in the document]
tkalkirill: No.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexStorage.java
```diff
@@ -54,4 +55,22 @@ public interface IndexStorage {
...
+    /**
+     * Sets last row ID for which the index was built, {@code null} means index building is finished.
+     *
+     * @apiNote This method <b>must</b> always be called inside the corresponding partition's
+     *     {@link org.apache.ignite.internal.storage.MvPartitionStorage#runConsistently} closure.
+     *
+     * @param rowId Row ID.
+     * @throws StorageException If failed to set the last row ID.
+     */
+    void setLastBuildRowId(@Nullable RowId rowId);
```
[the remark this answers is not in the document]
tkalkirill: This is an unchecked exception, there is no need for this.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,220 @@
...
+
+        private boolean isLocalNodeLeader(RaftGroupService raftGroupService) {
+            Peer leader = raftGroupService.leader();
+
+            assert leader != null : "tableId=" + table.tableId() + ", partitionId=" + partitionId;
+
+            return localNodeConsistentId().equals(leader.consistentId());
+        }
+
+        private List<RowId> createBatchRowIds(RowId lastBuildRowId, int batchSize) {
```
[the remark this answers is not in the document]
tkalkirill: Fix it.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexStorage.java
```diff
@@ -54,4 +55,22 @@ public interface IndexStorage {
      * @throws StorageException If failed to remove data.
      */
     void remove(IndexRow row) throws StorageException;
+
+    /**
+     * Returns the last row ID that has been processed by an ongoing index build process or {@code null} if the process has finished.
```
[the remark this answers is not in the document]
tkalkirill: Fix it.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/hash/PageMemoryHashIndexStorage.java
tkalkirill: I do not think that the operation of building an index will often occur, and does it make sense to keep an empty value in memory? If you insist, I can do caching.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/meta/UpdateLastBuildRowIdInvokeClosure.java
tkalkirill: Not quite sure what comment you mean.
Maybe and maybe smaller, is that a problem?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/HashIndex.java
tkalkirill: I don't know what name to give, I tried "indexMetaStorage".

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbHashIndexStorage.java
tkalkirill: I do not think that the operation of building an index will often occur, and does it make sense to keep an empty value in memory? If you insist, I can do caching.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMetaStorage.java
tkalkirill: Forgot about that, sorry.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
tkalkirill: So that the construction of the index in the raft group is consistent and does not require additional actions.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/command/BuildIndexCommand.java
tkalkirill: fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
tkalkirill: No ideas.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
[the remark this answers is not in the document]
ibessonov: Ok. Then why it's a 100? Seems arbitrary. What if it affects other queries too much?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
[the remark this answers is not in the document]
ibessonov: Ok, and you log an error in such case? If so, then that's wrong. I saw you starting this task for all partitions, even if it's not an affinity node.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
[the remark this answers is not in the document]
ibessonov: I think that it would be nice to get rid of most "getOrCreate*" methods in separate tickets, yes

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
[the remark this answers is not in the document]
tkalkirill: This is true, but I don’t know yet whether it will be necessary for the user to configure it, if you insist, I’ll add it.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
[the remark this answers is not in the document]
tkalkirill: I log any exceptions.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexStorage.java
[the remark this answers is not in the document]
ibessonov: But you document it, why not making it explicit in the signature? Wouldn't hurt anyone, you know. Maybe the opposite, it'll make code better

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/hash/PageMemoryHashIndexStorage.java
ibessonov: I insist

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/meta/UpdateLastBuildRowIdInvokeClosure.java
ibessonov: I meant my own comment that I was writing

ibessonov: Yes, this is a problem, if you suddenly reduce the value. Makes things kinda unpredictable, don't you think so?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/sorted/PageMemorySortedIndexStorage.java
ibessonov: Ok

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/HashIndex.java
ibessonov: `indexMetaStorage` is a lot better

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
ibessonov: Thank you

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
ibessonov: Maybe you can move it outside :wink:

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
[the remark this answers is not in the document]
tkalkirill: https://issues.apache.org/jira/browse/IGNITE-19112

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexStorage.java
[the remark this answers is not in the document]
tkalkirill: For example, exceptions are specified in `java.util.Map#containsKey` but not in the signature, I don’t think that this makes the code better, do you insist?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
tkalkirill: I'm a little worried that I might break something, perhaps I'll fix it in a separate ticket along with other refactorings.
https://issues.apache.org/jira/browse/IGNITE-19113

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/hash/PageMemoryHashIndexStorage.java
tkalkirill: Try to fix it.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/index/RocksDbHashIndexStorage.java
tkalkirill: Try to fix it.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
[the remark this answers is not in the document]
tkalkirill: Firstly, I did not understand your question, so I could not answer it, so I decided to clarify what you wanted.
Now I understand your question better.

Index building commands are executed sequentially, so I think that there will not be empty (well, or 0).

I understand your example and should change this code.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/hash/PageMemoryHashIndexStorage.java
```diff
@@ -105,6 +108,8 @@ public PageMemoryHashIndexStorage(
         lowestRowId = RowId.lowestRowId(partitionId);
 
         highestRowId = RowId.highestRowId(partitionId);
+
+        lastBuiltIndexRow = readLastBuiltRowId();
```
ibessonov, line 112: `IndexMeta` could have been passed into the constructor, because you already read it. This way you wouldn't need to go to the tree second time.

tkalkirill: But it could be `null` if the index is new, and when we create the **Hash/SortedIndexTree**, we save the new `IndexMeta` to the `IndexMetaTree`.

ibessonov: Where is the code that creates the meta? Is it executed before the constructor? If the answer is "yes", then you do have the instance of the meta. If no, then where do you get the instance of the tree?

tkalkirill: Fix it

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/hash/PageMemoryHashIndexStorage.java
```diff
@@ -91,12 +91,15 @@ public class PageMemoryHashIndexStorage implements HashIndexStorage {
...
+     * @param lastBuiltIndexRowUuid Last row ID uuid that has been processed by an ongoing index build process, {@code null} if the process
+     *      has finished.
      */
     public PageMemoryHashIndexStorage(
             HashIndexDescriptor descriptor,
             IndexColumnsFreeList freeList,
             HashIndexTree hashIndexTree,
-            IndexMetaTree indexMetaTree
+            IndexMetaTree indexMetaTree,
+            @Nullable UUID lastBuiltIndexRowUuid
```
ibessonov, line 102: I recommended passing meta object, not the id itself, can you please change that?
Meta is an object, that has everything grouped, and passing such object into a constructor is a good thing. Passing some arbitrary uuid and, potentially, the rest of meta fields, looks wrong

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexStorage.java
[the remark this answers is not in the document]
tkalkirill: Fix it

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/RowId.java
```diff
@@ -163,4 +163,11 @@ public int compareTo(RowId rowId) {
     public String toString() {
         return "RowId [partitionId=" + partitionId() + ", uuid=" + uuid + ']';
     }
+
+    /**
+     * Returns {@code true} if the lowest row ID.
+     */
+    public boolean isLowest() {
```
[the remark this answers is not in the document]
tkalkirill: Remove it.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/hash/PageMemoryHashIndexStorage.java
tkalkirill: Fix it.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+    /** Index building executor. */
+    private final ExecutorService buildIndexExecutor;
+
+    IndexBuilder(String nodeName, IgniteSpinBusyLock busyLock, ClusterService clusterService) {
+        this.busyLock = busyLock;
+        this.clusterService = clusterService;
+
+        int cpus = Runtime.getRuntime().availableProcessors();
+
+        buildIndexExecutor = new ThreadPoolExecutor(
```
ibessonov, line 72: Can you use some method of `java.util.concurrent.Executors`?
What's "30 seconds" here?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+    private final ClusterService clusterService;
+
+    /** Index building executor. */
+    private final ExecutorService buildIndexExecutor;
+
+    IndexBuilder(String nodeName, IgniteSpinBusyLock busyLock, ClusterService clusterService) {
+        this.busyLock = busyLock;
+        this.clusterService = clusterService;
+
+        int cpus = Runtime.getRuntime().availableProcessors();
```
ibessonov, line 70: Do we want to configure the pool size? Probably

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+     */
+    void stop() {
+        shutdownAndAwaitTermination(buildIndexExecutor, 10, TimeUnit.SECONDS);
+    }
+
+    /**
+     * Initializes the build of the index.
+     */
+    void startIndexBuild(TableIndexView tableIndexView, TableImpl table) {
+        for (int partitionId = 0; partitionId < table.internalTable().partitions(); partitionId++) {
```
ibessonov, line 93: We should only start rebuild process for indexes that exist

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+            buildIndexExecutor.submit(new BuildIndexTask(table, tableIndexView, partitionId, null));
```
ibessonov, line 94: Please pass the real row id value instead of null.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+
+        private final TableIndexView tableIndexView;
+
+        private final int partitionId;
+
+        /**
+         * ID of the next row to build the index from the previous batch, {@code null} if it is the first row after the index was crated
+         * (both on a live node and after a restore).
+         */
+        private final @Nullable RowId nextRowIdToBuiltFromPreviousBatch;
```
ibessonov, line 119: to buil**d**
Please check your grammar when it comes to the word "build", it's the third time I correct you on it

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+        private boolean isLocalNodeLeader(RaftGroupService raftGroupService) {
+            Peer leader = raftGroupService.leader();
+
+            assert leader != null : "tableId=" + table.tableId() + ", partitionId=" + partitionId;
+
+            return localNodeConsistentId().equals(leader.consistentId());
+        }
+
+        private List<RowId> createBatchRowIds(RowId lastBuiltRowId, int batchSize) {
+            MvPartitionStorage mvPartition = table.internalTable().storage().getMvPartition(partitionId);
```
ibessonov, line 204: Why can't we store partition instance in the task, instead of retrieving it every time?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+                                boolean finish = batchRowIds.size() < BUILD_INDEX_ROW_ID_BATCH_SIZE || nextRowId == null;
+
+                                // TODO: IGNITE-19053 Must handle the change of leader
+                                return raftGroupService.run(createBuildIndexCommand(batchRowIds, finish))
+                                        .thenRun(() -> {
+                                            if (!finish) {
+                                                assert nextRowId != null : createCommonTableIndexInfo();
+
+                                                buildIndexExecutor.submit(
+                                                        new BuildIndexTask(table, tableIndexView, partitionId, nextRowId)
```
ibessonov, line 175: Can we pass `this` as the argument?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+        }
+
+        private @Nullable RowId getNextRowIdForNextBatch(List<RowId> batch) {
+            return batch.isEmpty() ? null : batch.get(batch.size() - 1).increment();
+        }
+
+        private @Nullable List<RowId> collectRowIdBatch() {
+            RowId nextRowIdToBuilt;
+
+            if (nextRowIdToBuiltFromPreviousBatch == null) {
```
ibessonov, line 255: As I already mentioned, I don't want to see this null-check, I'd prefer this value to be non-null

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItBuildIndexTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+    @AfterEach
+    void tearDown() {
+        sql("DROP TABLE IF EXISTS " + TABLE_NAME);
+    }
+
+    @ParameterizedTest
+    @MethodSource("replicas")
+    void testBuildIndexOnStableTopology(int replicas) throws Exception {
+        sql(IgniteStringFormatter.format("CREATE ZONE IF NOT EXISTS {} WITH REPLICAS={}, PARTITIONS={}",
```
ibessonov, line 65: Do we have a default zone? Maybe there's a chance of making this test simpler

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItBuildIndexTest.java
```diff
@@ -0,0 +1,143 @@
...
+        ));
+
+        sql(IgniteStringFormatter.format(
+                "INSERT INTO {} VALUES {}",
+                TABLE_NAME, toValuesString(List.of(1, 1), List.of(2, 2), List.of(3, 3), List.of(4, 4), List.of(5, 5))
+        ));
+
+        sql(IgniteStringFormatter.format("CREATE INDEX {} ON {} (i1)", INDEX_NAME, TABLE_NAME));
+
+        // FIXME: IGNITE-18733
```
ibessonov, line 81: Usually we use TODO

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItBuildIndexTest.java
```diff
@@ -0,0 +1,143 @@
...
+                .returns(1, 1)
+                .returns(2, 2)
+                .returns(3, 3)
+                .returns(4, 4)
+                .returns(5, 5)
+                .check();
+    }
+
+    private static int[] replicas() {
+        // FIXME: IGNITE-19086 Fix NullPointerException on insertAll
```
ibessonov, line 97: Same here. There are automated tools, targeted to TODO comments specifically

ibessonov: By the way, what's the reason for the NPE? Can it be fixed now?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableIndexStoragesSupplier.java
```diff
@@ -0,0 +1,38 @@
...
+
+package org.apache.ignite.internal.table.distributed;
+
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * Supplier table index storages.
+ */
+public interface TableIndexStoragesSupplier {
```
ibessonov, line 26: What's the reason of calling it a supplier? Design of this interface look peculiar, I guess it's temporary

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableIndexStoragesSupplier.java
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Returns indexes by their ID.
+     *
+     * <p>Waits for the primary key index and all other registered indexes to be created.
+     */
+    Map<UUID, TableSchemaAwareIndexStorage> get();
+
+    /**
+     * Adds a wait to create an index, if not already created, on a subsequent call to {@link #get()}.
```
ibessonov, line 35: This comment is hard to understand, please rephrase it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -422,4 +425,33 @@ public void onShutdown() {
...
+        if (commandIndex <= storage.lastAppliedIndex()) {
+            return;
+        }
+
+        storage.runConsistently(() -> {
+            storage.lastApplied(commandIndex, commandTerm);
```
ibessonov, line 443: Why is this not the last line in the closure, but the first one instead?
It should represent the end of the operation. Order of things matter in the code, it helps understanding what's going on. In the future, we may want to split the "buildIndex" loop into several "runConssitently" calls, for example. In that case, wrong operations order could lead to actual bugs.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandlerTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+        UUID indexId = UUID.randomUUID();
+
+        TableIndexStoragesSupplier indexes = mock(TableIndexStoragesSupplier.class);
+
+        when(indexes.get()).thenReturn(Map.of(indexId, indexStorage));
+
+        StorageUpdateHandler storageUpdateHandler = createStorageUpdateHandler(partitionStorage, indexes);
+
+        RowId rowId0 = new RowId(PARTITION_ID, UUID.randomUUID());
```
ibessonov, line 73: There's a constructor that doesn't require explicit UUID as a parameter. Maybe you should use it

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandlerTest.java
```diff
@@ -0,0 +1,132 @@
...
+        RowId rowId1 = new RowId(PARTITION_ID, UUID.randomUUID());
+
+        List<BinaryRow> rowVersions0 = asList(mock(BinaryRow.class), null);
+        List<BinaryRow> rowVersions1 = asList(mock(BinaryRow.class), null);
+
+        setRowVersions(partitionStorage, Map.of(rowId0.uuid(), rowVersions0, rowId1.uuid(), rowVersions1));
+
+        storageUpdateHandler.buildIndex(indexId, List.of(rowId0.uuid(), rowId1.uuid()), false);
```
ibessonov, line 81: It's interesting that you pass rowIds in potentially incorrect order. You shouldn't do it, because it's not realistic and it may spontaneously stop working in the future if we add some assertion somewhere.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/PartitionCommandListenerTest.java
```diff
@@ -462,6 +464,48 @@ public void testSafeTime() {
         applySafeTimeCommand(SafeTimeSyncCommand.class, testClock.now());
     }
 
+    @Test
+    void testBuildIndexCommand() {
+        UUID indexId = UUID.randomUUID();
+
+        doNothing().when(storageUpdateHandler).buildIndex(eq(indexId), any(List.class), anyBoolean());
+
+        List<UUID> rowUuids0 = List.of(UUID.randomUUID());
```
ibessonov, line 473: Same here, order here is random. I guess it may not be important in this particular test, right?

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/PartitionCommandListenerTest.java
```diff
@@ -462,6 +464,48 @@ public void testSafeTime() {
...
+
+        inOrder.verify(partitionDataStorage).lastApplied(10, 1);
+        inOrder.verify(storageUpdateHandler).buildIndex(indexId, rowUuids0, false);
+
+        commandListener.handleBuildIndexCommand(createBuildIndexCommand(indexId, rowUuids1, true), 20, 2);
+
+        inOrder.verify(partitionDataStorage).lastApplied(20, 2);
+        inOrder.verify(storageUpdateHandler).buildIndex(indexId, rowUuids1, true);
+
+        commandListener.handleBuildIndexCommand(createBuildIndexCommand(indexId, rowUuids2, false), 5, 1);
```
ibessonov, line 489: Why to you pass smaller values here? Shouldn't such command be ignored?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
tkalkirill: I didn't find a suitable method there.
30 seconds - how long the threads will be idle before they are removed.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
tkalkirill: I think that all (well, or almost all) thread pools will need to be configured, but now I don’t think that this needs to be done.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
tkalkirill: At this point, we do not know for sure whether the partition will be on this node or not.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
tkalkirill: At this point, we do not know for sure whether the partition will be on this node or not, so we need to make sure that we will build the index and only then use the index storage.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
tkalkirill: Fix it

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
tkalkirill: Because we can easily get it from the `TableImpl`, there is no need to store it.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
tkalkirill: We can, but why?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
tkalkirill: I wrote above why I don’t like this approach and it doesn’t fit.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItBuildIndexTest.java
tkalkirill: Yes, there is a default zone, but I will have to change it, I think this will not simplify the test.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItBuildIndexTest.java
tkalkirill: Fix it

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItBuildIndexTest.java
tkalkirill: I changed it to `TODO`, I have not yet been able to understand the reason for the NPE, the test falls even without building indexes on the main, I think it is necessary to investigate this separately.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableIndexStoragesSupplier.java
tkalkirill: I could not find the correct name, I assume that it is temporary, when catalogs appear, we will get rid of this code.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
tkalkirill: Fix it

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandlerTest.java
tkalkirill: Fix it

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/PartitionCommandListenerTest.java
tkalkirill: Add comment.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandlerTest.java
tkalkirill: I am passing in a list of `RowId` and expect that we will only execute code relative to order of that list, not the order of the `RowId`.
I don't check order `RowId`, I assume that the receiver sends the `RowId` in the correct order.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/raft/PartitionCommandListenerTest.java
tkalkirill: Right.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableIndexStoragesSupplier.java
tkalkirill: Try fix it

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
ibessonov: How come? Maybe you create these tasks too early, in this case

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
ibessonov: How does that make sense? By the time tasks starts, partition should already be available, right? How can it not be available at the time of task creation?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
ibessonov: Ok, as you wish

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
ibessonov: Because I don't understand the need to create a new object, if we already have one

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
ibessonov: I don't agree with the statement that it doesn't fit. I think you start tasks at inappropriate time, let's discuss it further

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
tkalkirill: Discussed personally, I will create TODO.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
tkalkirill: Discussed personally, I will create TODO.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
tkalkirill: What's wrong with creating fresh objects?
GC will remove old objects, will not shift to old generations.

ibessonov (approved):

PR #1802, opened 2023-03-16
IGNITE-18305 CLI command for list of enabled metrics

valepakh:
https://issues.apache.org/jira/browse/IGNITE-18305


PakhomovAlexander (approved):

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/registry/impl/MetricRegistryImpl.java
```diff
@@ -0,0 +1,102 @@
...
+     * Start pulling updates from a node.
+     *
+     * @param sessionInfo sessionInfo.
+     */
+    @Override
+    public void onConnect(SessionInfo sessionInfo) {
+        if (executor == null) {
+            executor = Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory("NodeNameRegistry", LOG));
+            executor.scheduleWithFixedDelay(() ->
+                    updateMetrics(sessionInfo.nodeUrl()), 0, 5, TimeUnit.SECONDS);
```
kgusakov, line 74: I would prefer to get the metrics directly on command call. Otherwise I can get into confused situation, when I enable some metrics and then check it by a list command -  but list command returns the old data. Also, this pulling will create the redundant parasit traffic on the node by every active cli tool.

kgusakov (approved):

PR #1809, opened 2023-03-17
IGNITE-18973 Introduce unit statuses

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-18973


modules/api/src/main/java/org/apache/ignite/deployment/DeploymentInfo.java
```diff
@@ -0,0 +1,115 @@
...
+        this.status = status;
+        this.consistentIds = Collections.unmodifiableList(consistentIds);
+    }
+
+    public DeploymentStatus status() {
+        return status;
+    }
+
+    public List<String> consistentIds() {
+        return consistentIds;
```
ptupitsyn, line 44: Returned value can be modified from outside (add/remove items), is that ok?

PakhomovAlexander (approved):

PR #1812, opened 2023-03-17
IGNITE-18943 Disable the set of endpoints for non-initialized cluster

PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-18943

### DIFF
--- a/modules/cli/src/main/java/org/apache/ignite/internal/cli/call/unit/DeployUnitCall.java
+++ b/modules/cli/src/main/java/org/apache/ignite/internal/cli/call/unit/DeployUnitCall.java
@@ -90,7 +90,7 @@ public class DeployUnitCall implements AsyncCall<DeployUnitCallInput, String> {
             ApiException apiException = (ApiException) exception;
             if (apiException.getCode() == 409) {
                 // special case when cluster is not initialized
-                if (((ApiException) exception).getResponseBody().contains("Cluster is not initialized")) {
+                if (apiException.getResponseBody().contains("Cluster is not initialized")) {
                     return DefaultCallOutput.failure(new IgniteCliApiException(exception, input.clusterUrl()));
                 }
                 return DefaultCallOutput.failure(new UnitAlreadyExistsException(input.id(), input.version()));
--- a/modules/cli/src/main/java/org/apache/ignite/internal/cli/call/unit/DeployUnitCall.java
+++ b/modules/cli/src/main/java/org/apache/ignite/internal/cli/call/unit/DeployUnitCall.java
@@ -89,6 +89,10 @@ public class DeployUnitCall implements AsyncCall<DeployUnitCallInput, String> {
         if (exception instanceof ApiException) {
             ApiException apiException = (ApiException) exception;
             if (apiException.getCode() == 409) {
+                // special case when cluster is not initialized
+                if (((ApiException) exception).getResponseBody().contains("Cluster is not initialized")) {
+                    return DefaultCallOutput.failure(new IgniteCliApiException(exception, input.clusterUrl()));
+                }
                 return DefaultCallOutput.failure(new UnitAlreadyExistsException(input.id(), input.version()));
             }
         }
--- a/modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/metric/NodeMetricDisableCommand.java
+++ b/modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/metric/NodeMetricDisableCommand.java
@@ -24,6 +24,7 @@ import org.apache.ignite.internal.cli.commands.BaseCommand;
 import org.apache.ignite.internal.cli.commands.metric.MetricSourceMixin;
 import org.apache.ignite.internal.cli.commands.node.NodeUrlProfileMixin;
 import org.apache.ignite.internal.cli.core.call.CallExecutionPipeline;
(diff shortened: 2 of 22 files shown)


modules/cli/src/main/java/org/apache/ignite/internal/cli/call/unit/DeployUnitCall.java
```diff
@@ -89,6 +89,10 @@ private static CallOutput<String> handleException(Exception exception, DeployUni
         if (exception instanceof ApiException) {
             ApiException apiException = (ApiException) exception;
             if (apiException.getCode() == 409) {
+                // special case when cluster is not initialized
+                if (((ApiException) exception).getResponseBody().contains("Cluster is not initialized")) {
```
valepakh, line 93:
```suggestion
                if (apiException.getResponseBody().contains("Cluster is not initialized")) {
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/metric/NodeMetricDisableCommand.java
```diff
@@ -48,6 +49,7 @@ public Integer call() {
                 .output(spec.commandLine().getOut())
                 .errOutput(spec.commandLine().getErr())
                 .verbose(verbose)
+                .exceptionHandler(new ClusterNotInitializedExceptionHandler("Cannot disable metrics", "cluster init"))
```
valepakh, line 52:
```suggestion
                .exceptionHandler(new ClusterNotInitializedExceptionHandler("Cannot disable metrics", "ignite cluster init"))
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/metric/NodeMetricEnableCommand.java
```diff
@@ -48,6 +49,7 @@ public Integer call() {
                 .output(spec.commandLine().getOut())
                 .errOutput(spec.commandLine().getErr())
                 .verbose(verbose)
+                .exceptionHandler(new ClusterNotInitializedExceptionHandler("Cannot enable metrics", "cluster init"))
```
valepakh, line 52:
```suggestion
                .exceptionHandler(new ClusterNotInitializedExceptionHandler("Cannot enable metrics", "ignite cluster init"))
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/metric/NodeMetricListCommand.java
```diff
@@ -47,6 +48,7 @@ public Integer call() {
                 .errOutput(spec.commandLine().getErr())
                 .decorator(new MetricListDecorator())
                 .verbose(verbose)
+                .exceptionHandler(new ClusterNotInitializedExceptionHandler("Cannot list metrics", "cluster init"))
```
valepakh, line 51:
```suggestion
                .exceptionHandler(new ClusterNotInitializedExceptionHandler("Cannot list metrics", "ignite cluster init"))
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/unit/UnitDeployCommand.java
```diff
@@ -72,6 +73,7 @@ public Integer call() throws Exception {
                 .output(spec.commandLine().getOut())
                 .errOutput(spec.commandLine().getErr())
                 .verbose(verbose)
+                .exceptionHandler(new ClusterNotInitializedExceptionHandler("Cannot deploy unit", "cluster init"))
```
valepakh, line 76: We should rewrite the UnitCommand to use `askQuestionIfNotConnected` and change this handler accordingly for non-REPL commands. Since there are a lot of them we might want to extract this to a pair of construction methods? Also since we renamed the CLI executable to `ignite3` we might want to do that now and then change the string accordingly.

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/metric/NodeMetricDisableCommand.java
PakhomovAlexander: In case this is the REPL command the `ignite` prefix does not make ane sence.

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/unit/UnitDeployCommand.java
PakhomovAlexander: Probably yes, deserves a separate ticket for that.

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/node/metric/NodeMetricDisableCommand.java
valepakh: > In case this is the REPL command the `ignite` prefix does not make ane sence.

But it's not a REPL command, there's a separate `NodeMetricDisableReplCommand`

valepakh: I would create a ticket and simply extract a factory method which will encapsulate the command string.

PR #1815, opened 2023-03-17
IGNITE-19022 ItReadOnlyTransactionTest is flaky in TC due to replica …

vldpyatkov:
…is timed out
https://issues.apache.org/jira/browse/IGNITE-19022

### DIFF
--- a/modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
+++ b/modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
@@ -311,7 +311,7 @@ public class Loza implements RaftManager {
                 FACTORY,
                 raftConfiguration,
                 membersConfiguration,
-                true,
+                false,
                 executor
         );
     }
--- a/modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
+++ b/modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
@@ -311,7 +311,7 @@ public class Loza implements RaftManager {
                 FACTORY,
                 raftConfiguration,
                 membersConfiguration,
-                false,
+                true,
                 executor
         );
     }
--- a/modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
+++ b/modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
@@ -174,7 +174,12 @@ public class Loza implements RaftManager {
             RaftGroupListener lsnr,
             RaftGroupEventsListener eventsLsnr
     ) throws NodeStoppingException {
-        return startRaftGroupNode(nodeId, configuration, lsnr, eventsLsnr, RaftGroupOptions.defaults());
+        CompletableFuture<RaftGroupService> fut = startRaftGroupNode(nodeId, configuration, lsnr, eventsLsnr, RaftGroupOptions.defaults());
+
+        // TODO: https://issues.apache.org/jira/browse/IGNITE-19047 Meta storage and cmg raft log re-application in async manner
+        raftServer.raftNodeReadyFuture(nodeId.groupId()).join();
+
+        return fut;
     }
 
     @Override
@@ -309,6 +314,17 @@ public class Loza implements RaftManager {
                 true,
                 executor
         );
+    }
+
+
+    /**
+     * Gets a future that completes when all committed updates are applied to state machine.
+     *
+     * @param groupId Raft group id.
+     * @return Future to last applied revision.
+     */
+    public CompletableFuture<Long> raftNodeReadyFuture(ReplicationGroupId groupId) {
+        return raftServer.raftNodeReadyFuture(groupId);
     }
 
     @Override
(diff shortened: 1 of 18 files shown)


modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
```diff
@@ -256,6 +261,17 @@ private CompletableFuture<RaftGroupService> startRaftGroupServiceInternal(
         );
     }
 
+
+    /**
+     * Gets a future that completes when all committed updates are applied to state machine.
```
sanpwc, line 266: Here and there. I'd add that it's only about node startup process.

modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/JraftServerImpl.java
```diff
@@ -457,6 +457,14 @@ public boolean startRaftNode(
         }
     }
 
+    @Override
+    public CompletableFuture<Long> raftNodeReadyFuture(ReplicationGroupId groupId) {
+        RaftGroupService jraftNode = nodes.entrySet().stream().filter(entry -> entry.getKey().groupId().equals(groupId))
```
sanpwc, line 462: Are you sure that there will only one node with given groupId. Seems that it should be true for AI, but not for GG because of learners. Could you please create ticket for GG if it's possible to have multiple nodes there?

modules/raft/src/main/java/org/apache/ignite/raft/jraft/RaftGroupService.java
```diff
@@ -57,7 +58,7 @@ public class RaftGroupService {
     /**
      * The raft node.
      */
-    private Node node;
+    private NodeImpl node;
```
sanpwc, line 61: Why?

modules/raft/src/main/java/org/apache/ignite/raft/jraft/RaftGroupService.java
```diff
@@ -123,6 +124,14 @@ public synchronized Node start() {
         return this.node;
     }
 
+    /**
+     * Gets a future to apply committed revisions.
```
sanpwc, line 128: Here and there, why it's 'to'?

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
```diff
@@ -1082,20 +1084,8 @@ public void onApplied( long lastAppliedLogIndex) {
...
-            } catch (InterruptedException e) {
-                LOG.error("Fail to apply committed updates.", e);
-
-                return false;
-            }
         } else {
-            if (externalAwaitStorageLatch) {
-                opts.getStorageReadyLatch().countDown();
-            }
+            logApplyComplition.complete(0L);
```
sanpwc, line 1088: Why it's 0? Despite the fact that there were no pending records in log, storage may have it's own checkpointed state with non-zero applied index. BTW, how do we use the revision, the one that is used to compete the future?

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
```diff
@@ -1116,38 +1106,53 @@ public void onApplied( long lastAppliedLogIndex) {
...
+            if (this.conf.isStable() && this.conf.getConf().size() == 1 && this.conf.getConf().contains(this.serverId)) {
+                // The group contains only this server which must be the LEADER, trigger
+                // the timer immediately.
+                electSelf();
+            }
+            else {
+                this.writeLock.unlock();
+            }
+
+            applyCommittedFuture.complete(commitIdx);
```
sanpwc, line 1143: Should we compete the future exceptionally in case of any error? Seems that otherwise we may hang forever.

sanpwc (changes requested):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -826,6 +822,10 @@ private CompletableFuture<?> updateAssignmentInternal(ConfigurationNotificationE
 
                                         try {
                                             replicaMgr.startReplica(replicaGrpId,
```
sanpwc, line 824: Formatting, let's move replicaGrpId to the next line.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2040,6 +2035,10 @@ private CompletableFuture<Void> handleChangePendingAssignmentEvent(
                             UUID tblId = tbl.tableId();
 
                             replicaMgr.startReplica(replicaGrpId,
+                                    CompletableFuture.allOf(
```
sanpwc, line 2038: Formatting, same as above.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
sanpwc: > I think, it would be useful to know a last revision that was applied before node start. Don't you?
Yep, probably.

sanpwc (approved):

PR #1820, opened 2023-03-20
IGNITE-17607 C++ Compute API

isapego:
- Added compute API to C++ client;
- Fixed logic of writing node list so it's now valid according to msgpack;
- Added tests.

### DIFF
--- a/modules/platforms/cpp/ignite/client/compute/compute.cpp
+++ b/modules/platforms/cpp/ignite/client/compute/compute.cpp
@@ -25,7 +25,6 @@ namespace ignite {
 
 template<typename T>
 typename T::value_type get_random_element(const T &cont) {
-    // TODO: Move to utils
     static std::mutex randomMutex;
     static std::random_device rd;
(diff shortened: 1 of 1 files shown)


modules/platforms/cpp/ignite/client/detail/compute/compute_impl.cpp
```diff
@@ -0,0 +1,127 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "ignite/client/detail/compute/compute_impl.h"
+#include "ignite/client/detail/utils.h"
+#include "ignite/schema/binary_tuple_builder.h"
+
+namespace ignite::detail {
```
ptupitsyn, line 22: Just curious - what does "detail" mean, is it some kind of a well-known name for internal namespaces?

ptupitsyn (approved):

PR #1823, opened 2023-03-21
IGNITE-18655: fix muted SQL tests

ygerzhedovich:


AMashenkov (approved):

modules/runner/src/integrationTest/sql/types/decimal/test_decimal_ops.test
```diff
@@ -5,9 +5,8 @@
 statement ok
 PRAGMA enable_verification
 
-skipif ignite3
-# https://issues.apache.org/jira/browse/IGNITE-18655
-statement error
+
+statement ok
 CREATE TABLE decimals_without_params(d DECIMAL);
```
xtern, line 10: may be we should somehow check precision/scale parameters, that we can something insert/select or at least add comment explaining this test case?
e.g.
```
statement ok
CREATE TABLE decimals_without_params(d DECIMAL);
INSERT INTO decimals_without_params values(1234567890.12345);

query T
SELECT * FROM decimals_without_params
----
...expected
```

Should we add a similar test with precision only defined?
```
CREATE TABLE decimals_without_params2(d DECIMAL(9));
```

ygerzhedovich: we have ItDataTypesTest which covers the cases

ygerzhedovich: @xtern tests have been added

PR #1831, opened 2023-03-22
IGNITE-18983 Sql. Jdbc. Migrate JDBC handler on new internal API

korlov42:


modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -90,62 +95,76 @@ public class JdbcQueryEventHandlerImpl implements JdbcQueryEventHandler {
...
-            );
+        // TODO: IGNITE-16960 support script execution
+        if (req.getStmtType() == JdbcStatementType.ANY_STATEMENT_TYPE) {
+            return CompletableFuture.completedFuture(new JdbcQueryExecuteResult(Response.STATUS_FAILED, "Scripts are not supported"));
         }
 
-        if (results.isEmpty()) {
+        JdbcConnectionContext connectionContext;
+        try {
+            connectionContext = resources.get(connectionId).get(JdbcConnectionContext.class);
```
AMashenkov, line 145: `resources.get(connectionId)` may return null if connection was closed concurrently/
Should we ignore request in that case?

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -345,4 +364,111 @@ private boolean validateDmlResult(ResultSetMetadata meta, boolean next) {
...
+
+                        Throwable error = resAndError.getSecond();
+
+                        if (sessionExpiredError(error)) {
+                            SessionId newSessionId = recreateSession(finalSessionId);
+
+                            return action.perform(newSessionId);
+                        }
+
+                        return CompletableFuture.failedFuture(error);
```
AMashenkov, line 419: This can be simplified with sneakyThrow.
```suggestion
                    .handle((res, error) ->
                        if (error == null) {
                            return res;
                        }

                        if (sessionExpiredError(error)) {
                            SessionId newSessionId = recreateSession(finalSessionId);

                            return action.perform(newSessionId);
                        }

                        IgniteUtils.sneakyThrow(error); // Re-throw original exception here.
```

modules/client-handler/src/test/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImplTest.java
```diff
@@ -0,0 +1,209 @@
...
+    private long acquireConnectionId() {
+        JdbcConnectResult result = await(eventHandler.connect());
+
+        assertThat(result, notNullValue());
+        assertThat(result.status(), is(STATUS_SUCCESS));
+        assertThat(result.hasResults(), is(true));
+
+        return result.connectionId();
+    }
+}
```
AMashenkov, line 209:
```suggestion
}

```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
```diff
@@ -149,7 +151,13 @@ public int defaultPageSize() {
...
     @Override
     public SessionBuilder toBuilder() {
-        return new SessionBuilderImpl(qryProc, new HashMap<>(props.toMap()))
+        Map<String, Object> propertyMap = new HashMap<>();
+
+        for (Map.Entry<Property<?>, Object> entry : props) {
+            propertyMap.put(entry.getKey().name, entry.getValue());
+        }
+
+        return new SessionBuilderImpl(qryProc, propertyMap)
```
AMashenkov, line 160:
```suggestion      
        return new SessionBuilderImpl(qryProc, props.toMap())
```
Actually, PropertiesHolder API looks ambiguous.
It is not clean if toMap() return a copy of properties, or just adapter or internal map.
I'd expect toMap() returns a copy of props, and it make sense to add asMap() that will avoid copying.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/property/PropertiesHelper.java
```diff
@@ -20,14 +20,70 @@
...
+     *
+     * <p>If conflict arises, conflicted values of secondary holder will be overridden with values
+     * of primary holder.
+     *
+     * @param primaryHolder A holder whose values will be preserved.
+     * @param secondaryHolder A holder whose values will be overridden in case of conflict.
+     * @return A holder containing properties from both holders.
+     */
+    public static PropertiesHolder merge(PropertiesHolder primaryHolder, PropertiesHolder secondaryHolder) {
+        Builder builder = builderFromHolder(secondaryHolder);
```
AMashenkov, line 56: As I see how this method is used, 'secondaryHolder' contains defaults, and 'primaryHolder' contains user props.
'builderFromHolder' validates the default values, but not the user ones.
Is it expected behaviour?

Should we validate props in builder method and then assume that property value type is correct as the only way to create PropertisHolder is using the builder.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
AMashenkov: If we clam PropertiesHolder is immutable, and can be changes only via copying with builder,
then maybe defensive-copying can be ommited or replaced with Collections.unmodifiableMap() or Map.ofEntries().

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
korlov42: suggested change won't work because `action.perform(newSessionId)` returns CompletableFuture, and we need to inline result of this future into a chain. Thus, we can't avoid `thenCompose`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionBuilderImpl.java
```diff
@@ -123,26 +125,21 @@ public SessionBuilder defaultPageSize(int pageSize) {
     /** {@inheritDoc} */
     @Override
     public SessionBuilder property(String name, @Nullable Object value) {
-        var prop = QueryProperty.byName(name);
-
-        if (prop != null) {
-            props.put(prop, value);
-        }
+        props.put(name, value);
```
ygerzhedovich, line 128: What reason to have that properties? It not use anywhere

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/property/PropertiesHolderImpl.java
```diff
@@ -38,38 +40,45 @@ class PropertiesHolderImpl implements PropertiesHolder {
...
-    @SuppressWarnings("unchecked")
     @Override
-    public <T> @Nullable T get(Property<T> prop) {
-        Object val = props.get(prop);
+    public <T> T get(Property<T> prop) {
+        T val = get0(prop);
 
         if (val == null) {
-            return null;
+            throw new RuntimeException(format("Property '{}' is not set", prop.name));
```
ygerzhedovich, line 48: seems right exception is PropertyNotFound

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionBuilderImpl.java
korlov42: they are for future use

korlov42: they are for future use

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/property/PropertiesHolderImpl.java
korlov42: good catch! Fixed

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/property/PropertyNotFound.java
```diff
@@ -15,24 +15,22 @@
...
 
-    public static Builder builder() {
-        return new Builder();
-    }
+    private static final long serialVersionUID = 4651104853616619936L;
 
-    /**
-     * Parameters builder.
-     */
-    public static class Builder extends AbstractBuilder<DropTableParams, Builder> {
```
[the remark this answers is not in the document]
korlov42: hm, this looks like duplicate...

ygerzhedovich (approved):

AMashenkov (approved):

PR #1832, opened 2023-03-22
IGNITE-19093: Create standalone MetaStorage implementation for test purposes

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-19093

### DIFF
--- a/modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
+++ b/modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
@@ -110,7 +110,6 @@ public class DistributionZoneManagerConfigurationChangesTest extends IgniteAbstr
                 .thenReturn(completedFuture(new VaultEntry(zonesLogicalTopologyKey(), toBytes(nodes))));
 
         LogicalTopologyService logicalTopologyService = mock(LogicalTopologyService.class);
-        doNothing().when(logicalTopologyService).addEventListener(any());
         when(logicalTopologyService.logicalTopologyOnLeader()).thenReturn(completedFuture(new LogicalTopologySnapshot(1, Set.of())));
 
         keyValueStorage = spy(new SimpleInMemoryKeyValueStorage("test"));
--- a/modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
+++ b/modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
@@ -39,7 +39,6 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.timeout;
 import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoInteractions;
 import static org.mockito.Mockito.when;
 
 import java.util.ArrayList;
@@ -65,7 +64,6 @@ import org.apache.ignite.lang.NodeStoppingException;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.mockito.Mockito;
 
 /**
  * Tests distribution zones configuration changes and reaction to that changes.
@@ -185,7 +183,7 @@ public class DistributionZoneManagerConfigurationChangesTest extends IgniteAbstr
     void testDataNodesNotPropagatedAfterZoneCreation() throws Exception {
         keyValueStorage.put(zonesChangeTriggerKey(1).bytes(), longToBytes(100));
 
-        Mockito.clearInvocations(keyValueStorage);
+        clearInvocations(keyValueStorage);
 
         distributionZoneManager.createZone(new DistributionZoneConfigurationParameters.Builder(ZONE_NAME).build()).get();
 
@@ -198,7 +196,7 @@ public class DistributionZoneManagerConfigurationChangesTest extends IgniteAbstr
 
     @Test
     void testZoneDeleteDoNotRemoveMetaStorageKey() throws Exception {
-        Mockito.clearInvocations(keyValueStorage);
+        clearInvocations(keyValueStorage);
 
         distributionZoneManager.createZone(new DistributionZoneConfigurationParameters.Builder(ZONE_NAME).build()).get();
 
--- a/modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
+++ b/modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
@@ -120,7 +120,7 @@ public class DistributionZoneManagerConfigurationChangesTest extends IgniteAbstr
         );
 
         // Mock logical topology for distribution zone.
-        vaultMgr.put(zonesLogicalTopologyKey(), toBytes(nodes));
+        // vaultMgr.put(zonesLogicalTopologyKey(), toBytes(nodes));
         keyValueStorage.put(zonesLogicalTopologyVersionKey().bytes(), longToBytes(1));
         keyValueStorage.put(zonesLogicalTopologyKey().bytes(), toBytes(nodes));
 
--- a/modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
+++ b/modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
@@ -33,22 +33,15 @@ import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.doNothing;
-import static org.mockito.Mockito.lenient;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.timeout;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
-import java.io.Serializable;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.atomic.AtomicLong;
-import java.util.stream.Collectors;
 import org.apache.ignite.configuration.NamedConfigurationTree;
 import org.apache.ignite.configuration.NamedListView;
 import org.apache.ignite.internal.cluster.management.topology.api.LogicalTopologyService;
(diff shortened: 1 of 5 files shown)


modules/metastorage/src/testFixtures/java/org/apache/ignite/internal/metastorage/impl/StandaloneMetaStorageManager.java
```diff
@@ -0,0 +1,135 @@
...
+    public void start() {
+        keyValueStorage.start();
+
+        MetaStorageListener metaStorageListener = new MetaStorageListener(keyValueStorage);
+
+        RaftGroupService raftGroupService = Mockito.mock(RaftGroupService.class);
+        Mockito.when(raftGroupService.run(ArgumentMatchers.any()))
+                .thenAnswer(invocation -> runCommand(invocation.getArgument(0), metaStorageListener));
+
+        var localNode = new ClusterNode(TEST_NODE_NAME, TEST_NODE_NAME, new NetworkAddress("localhost", 10000));
```
sashapolo, line 105: `localNode` should be a constructor parameter

modules/metastorage/src/testFixtures/java/org/apache/ignite/internal/metastorage/impl/StandaloneMetaStorageManager.java
```diff
@@ -0,0 +1,135 @@
...
+
+        MetaStorageServiceImpl service = new MetaStorageServiceImpl(raftGroupService, new IgniteSpinBusyLock(), localNode);
+
+        metaStorageServiceFuture().complete(service);
```
sashapolo, line 109: This looks like a hack, I would like to propose a more clean solution using mocks, this way we won't have to complete the future explicitly, but it will be completed with our mock RaftGroupService:
```
private static RaftManager mockRaftManager() {
    ArgumentCaptor<RaftGroupListener> listenerCaptor = ArgumentCaptor.forClass(RaftGroupListener.class);

    RaftManager raftManager = mock(RaftManager.class);

    RaftGroupService raftGroupService = mock(RaftGroupService.class);

    try {
        when(raftManager.startRaftGroupNode(any(), any(), listenerCaptor.capture(), any()))
                .thenReturn(completedFuture(raftGroupService));
    } catch (NodeStoppingException e) {
        throw new RuntimeException(e);
    }

    when(raftGroupService.run(any()))
            .thenAnswer(invocation -> {
                Command command = invocation.getArgument(0);

                RaftGroupListener listener = listenerCaptor.getValue();

                return runCommand(command, listener);
            });

    return raftManager;
}
```

modules/metastorage/src/testFixtures/java/org/apache/ignite/internal/metastorage/impl/StandaloneMetaStorageManager.java
```diff
@@ -0,0 +1,135 @@
...
+        return create(vaultMgr, new SimpleInMemoryKeyValueStorage(TEST_NODE_NAME));
+    }
+
+    /**
+     * Creates standalone MetaStorage manager for provided VaultManager.
+     */
+    public static StandaloneMetaStorageManager create(VaultManager vaultMgr, KeyValueStorage storage) {
+        return new StandaloneMetaStorageManager(
+                vaultMgr,
+                Mockito.mock(ClusterService.class),
```
sashapolo, line 68: Let's statically import all Mockito method, that looks cleaner

modules/metastorage/src/testFixtures/java/org/apache/ignite/internal/metastorage/impl/StandaloneMetaStorageManager.java
```diff
@@ -0,0 +1,135 @@
...
+     * The constructor.
+     *
+     * @param vaultMgr Vault manager.
+     * @param clusterService Cluster network service.
+     * @param cmgMgr Cluster management service Manager.
+     * @param logicalTopologyService Logical topology service.
+     * @param raftMgr Raft manager.
+     * @param storage Storage. This component owns this resource and will manage its lifecycle.
+     */
+    public StandaloneMetaStorageManager(VaultManager vaultMgr, ClusterService clusterService, ClusterManagementGroupManager cmgMgr,
```
sashapolo, line 88: Shouldn't this constructor be `private` since we have a factory method?

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
```diff
@@ -356,10 +211,15 @@ void testZoneDeleteDoNotRemoveMetaStorageKey() throws Exception {
         assertDataNodesForZone(1, nodes, keyValueStorage);
     }
 
-    private void mockVaultZonesLogicalTopologyKey(Set<String> nodes) {
-        byte[] newLogicalTopology = toBytes(nodes);
+    private static TablesConfiguration mockTablesConfiguration() {
```
sashapolo, line 214: I'm pretty sure this mocking is not needed at all, we can replace it with `@InjectConfiguration`

modules/metastorage/src/testFixtures/java/org/apache/ignite/internal/metastorage/impl/StandaloneMetaStorageManager.java
```diff
@@ -0,0 +1,135 @@
...
+            LogicalTopologyService logicalTopologyService, RaftManager raftMgr, KeyValueStorage storage) {
+        super(vaultMgr, clusterService, cmgMgr, logicalTopologyService, raftMgr, storage);
+
+        keyValueStorage = storage;
+    }
+
+    @Override
+    public void start() {
+        keyValueStorage.start();
```
sashapolo, line 97: You will also need to override the `stop` method in order to close the `keyValueStorage`

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
```diff
@@ -332,6 +183,8 @@ void testSeveralZoneCreationsUpdatesTriggerKey() throws Exception {
     void testDataNodesNotPropagatedAfterZoneCreation() throws Exception {
         keyValueStorage.put(zonesChangeTriggerKey(1).bytes(), longToBytes(100));
 
+        clearInvocations(keyValueStorage);
```
sashapolo, line 186: Why is this call needed?

modules/metastorage/src/testFixtures/java/org/apache/ignite/internal/metastorage/impl/StandaloneMetaStorageManager.java
AMashenkov: No. I'd better remove factory methods.

modules/metastorage/src/testFixtures/java/org/apache/ignite/internal/metastorage/impl/StandaloneMetaStorageManager.java
AMashenkov: Why? We pass fake ClusterNode just to avoid NPE. 
Standalone manager doens't imply any actual network interactions.

modules/metastorage/src/testFixtures/java/org/apache/ignite/internal/metastorage/impl/StandaloneMetaStorageManager.java
sashapolo: Let's remove the factory method then

modules/metastorage/src/testFixtures/java/org/apache/ignite/internal/metastorage/impl/StandaloneMetaStorageManager.java
sashapolo: oh, ok, let's leave it then

modules/metastorage/src/testFixtures/java/org/apache/ignite/internal/metastorage/impl/StandaloneMetaStorageManager.java
AMashenkov: I just wanted to avoid all RAFT related stuff.
Your approach is a bit complex and allow to emulate MetaStorage nodes and learner, I guess.

If you see it useful, ok, I'll fix this.

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
AMashenkov: DistributedZoneManager can write to metastorage during initialization.
This just reset Mockito counters in Spy object.

modules/metastorage/src/testFixtures/java/org/apache/ignite/internal/metastorage/impl/StandaloneMetaStorageManager.java
sashapolo: yes, it's still more straightforward than completing an internal future explicitly

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
```diff
@@ -115,178 +98,47 @@ public void setUp() {
         DistributionZonesConfiguration zonesConfiguration = clusterCfgMgr.configurationRegistry()
                 .getConfiguration(DistributionZonesConfiguration.KEY);
 
-        MetaStorageManager metaStorageManager = mock(MetaStorageManager.class);
-
-        when(metaStorageManager.appliedRevision(any())).thenReturn(completedFuture(0L));
+        vaultMgr = new VaultManager(new InMemoryVaultService());
 
         LogicalTopologyService logicalTopologyService = mock(LogicalTopologyService.class);
+        doNothing().when(logicalTopologyService).addEventListener(any());
```
sashapolo, line 104: Why do you need this stubbing? It's already a mock and will do nothing

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
sashapolo: I think it's better to improve the verification method instead. Currently we use `verify(keyValueStorage, timeout(1000).times(1)).invoke(any())`, maybe we can verify actual invocation arguments instead

modules/metastorage/src/testFixtures/java/org/apache/ignite/internal/metastorage/impl/StandaloneMetaStorageManager.java
AMashenkov: Ok. Anyway, this will significantly reduce code complexity in tests where this manager will be used.

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
AMashenkov: It doesn't seems possible, argument is a general class. `StatementResult invoke(If iif)`

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
sashapolo: yes, but you don't need to explicitly mock this method, this is the default behavior anyway. Or am I missing something?

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
sashapolo: that's a pity =( Ok then

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
sashapolo: Looks like this comment hast not been fixed, why did you resolve it?

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
```diff
@@ -115,178 +98,46 @@ public void setUp() {
...
-        clusterCfgMgr.start();
-
-        doNothing().when(logicalTopologyService).addEventListener(any());
-
-        when(logicalTopologyService.logicalTopologyOnLeader()).thenReturn(completedFuture(new LogicalTopologySnapshot(1, Set.of())));
-
-        mockVaultZonesLogicalTopologyKey(nodes);
+        // Mock logical topology for distribution zone.
+        // vaultMgr.put(zonesLogicalTopologyKey(), toBytes(nodes));
+        keyValueStorage.put(zonesLogicalTopologyVersionKey().bytes(), longToBytes(1));
```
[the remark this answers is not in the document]
AMashenkov:
```
 vaultMgr = mock(VaultManager.class);
        when(vaultMgr.get(any())).thenReturn(completedFuture(null));
        CompletableFuture<VaultEntry> fut = new CompletableFuture<>();;
        
        when(vaultMgr.get(zonesLogicalTopologyKey()))
                .thenReturn(fut.completeAsync(
                        () -> new VaultEntry(zonesLogicalTopologyKey(), toBytes(nodes)),
                        CompletableFuture.delayedExecutor(1, TimeUnit.SECONDS)));

```
Test fails if we delay future completion.
That means DistributedZoneManager make some interactions with other components, that are not tested well.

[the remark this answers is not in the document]
AMashenkov: I've revert to the previous approach. 
Looks fine.

sashapolo (approved):

PR #1848, opened 2023-03-27
IGNITE-18950 Mask passwords in the configuration

Flaugh24:


Flaugh24:
https://issues.apache.org/jira/browse/IGNITE-18950

modules/rest/src/main/java/org/apache/ignite/internal/rest/configuration/AbstractConfigurationController.java
```diff
@@ -17,15 +17,24 @@
...
 import org.apache.ignite.configuration.validation.ConfigurationValidationException;
 import org.apache.ignite.lang.IgniteException;
 
 /**
  * Base configuration controller.
  */
 public abstract class AbstractConfigurationController {
+
+    private final Set<String> keysToMask = Set.of("password");
+    private final Pattern sensitiveInfoprmationPattern = sensitiveInfoprmationPattern();
```
PakhomovAlexander, line 35:
```suggestion
    private final Pattern sensitiveInformationPattern = sensitiveInformationPattern();
```

modules/rest/src/main/java/org/apache/ignite/internal/rest/configuration/AbstractConfigurationController.java
```diff
@@ -39,7 +48,7 @@ public AbstractConfigurationController(ConfigurationPresentation<String> cfgPres
      * @return the presentation of configuration.
      */
     public String getConfiguration() {
-        return cfgPresentation.represent();
+        return maskSensitiveInformation("", cfgPresentation.represent());
```
PakhomovAlexander, line 51: Could you override the method in a way it would take only one argument `String configuration` and call `maskSensitiveInformation("", configuration)` inside?

modules/rest/src/main/java/org/apache/ignite/internal/rest/configuration/JsonMasker.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Locale;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * JSON masker.
+ */
+public class JsonMasker {
+
+    private static final String MASKING_DIGIT = "*";
```
PakhomovAlexander, line 39:
```suggestion
    private static final String MASKING_SYMBOL = "*";
```

modules/rest/src/main/java/org/apache/ignite/internal/rest/configuration/JsonMasker.java
```diff
@@ -0,0 +1,98 @@
...
+                        }
+                    }
+            );
+        }
+
+        return target;
+    }
+
+    private String maskString(String str) {
+        return MASKING_DIGIT.repeat(str.length());
```
PakhomovAlexander, line 86: No, we cannot use the password value as a reference for the length. We expose the password length here. Let's use some fixed number like 6.

modules/rest/src/test/java/org/apache/ignite/internal/rest/configuration/JsonMaskerTest.java
```diff
@@ -0,0 +1,98 @@
...
+import static java.util.Collections.emptySet;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+
+class JsonMaskerTest {
+
+    private final Set<String> keysToMask = Set.of("credentials");
+    private final JsonMasker jsonMasker = new JsonMasker();
```
PakhomovAlexander, line 29: Why JsonMasker can't set `keysToMask` in constructor? Seems like it is a part of object configuration and can be put to the constructor instead of propagating it in the method signature.

PakhomovAlexander (changes requested):

modules/rest/src/test/java/org/apache/ignite/internal/rest/configuration/JsonMaskerTest.java
Flaugh24: It can, but in that case, we have to create JsonMasker for each set of rules.

PR #1851, opened 2023-03-28
IGNITE-19142 IncomingSnapshotCopier.cancel() blocks forever if called from multiple threads

rpuch:
https://issues.apache.org/jira/browse/IGNITE-19142


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -151,6 +155,11 @@ public void join() throws InterruptedException {
 
     @Override
     public void cancel() {
+        // Cancellation from one thread must not block cancellations from other threads. hence this check.
```
SammyVimes, line 158:
```suggestion
        // Cancellation from one thread must not block cancellations from other threads, hence this check.
```

SammyVimes (approved):

SammyVimes (approved):

PR #1854, opened 2023-03-28
IGNITE-18865 Test storage recovery on rebalance abortion

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-18865

### DIFF
--- a/modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
+++ b/modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
@@ -19,6 +19,7 @@ package org.apache.ignite.internal.table.distributed;
 
 import static java.util.concurrent.CompletableFuture.completedFuture;
 import static org.apache.ignite.internal.testframework.IgniteTestUtils.await;
+import static org.apache.ignite.internal.testframework.matchers.CompletableFutureMatcher.willCompleteSuccessfully;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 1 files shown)


modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
```diff
@@ -429,8 +438,8 @@ public void tableManagerStopTest1() throws Exception {
     }
 
     /**
-     * Checks that the all RAFT nodes will be stopped when Table manager is stopping and
-     * an exception that was thrown by one of the component will not prevent stopping other components.
+     * Checks that the all RAFT nodes will be stopped when Table manager is stopping and an exception that was thrown by one of the
+     * component will not prevent stopping other components.
```
rpuch, line 442:
```suggestion
     * Checks that all RAFT nodes will be stopped when Table manager is stopping and an exception that was thrown by one of the
     * components will not prevent stopping other components.
```

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
```diff
@@ -734,21 +820,21 @@ private TableManager createTableManager(CompletableFuture<TableManager> tblManag
...
-                return Mockito.spy(super.createTxStateTableStorage(tableCfg));
+                txStateTableStorage = spy(super.createTxStateTableStorage(tableCfg));
+
+                return txStateTableStorage;
             }
         };
 
         sm.start();
 
-        if (!waitingSqlSchema) {
```
rpuch, line 748: Why don't we need this wait anymore?

rpuch: Oh, I think you are right. Then I'm ok with this removal.

rpuch (approved):

PR #1855, opened 2023-03-28
IGNITE-19085 Fix waiting for indexes on recovery when building indexes

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19085


modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
```diff
@@ -413,4 +392,24 @@ private interface IndexStorageAdapterFactory {
...
+     */
+    // TODO: IGNITE-19082 Needs to be redone/improved
+    public void addIndexesToWait(Collection<UUID> indexIds) {
+        for (UUID indexId : indexIds) {
+            indexesToWait.computeIfAbsent(indexId, uuid -> new CompletableFuture<>());
+        }
+    }
+
+    private void completeRegisterIndex(UUID indexId) {
+        CompletableFuture<?> indexToWaitFuture = indexesToWait.computeIfAbsent(indexId, uuid -> new CompletableFuture<>());
```
sashapolo, line 409: It used to be `remove` but now its `computeIfAbsent`, why?

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
```diff
@@ -413,4 +392,24 @@ private interface IndexStorageAdapterFactory {
         /** Creates the index decorator for given partition. */
         TableSchemaAwareIndexStorage create(int partitionId);
     }
+
+    /**
+     * Adds indexes to wait before inserting data into the table.
+     *
+     * @param indexIds Indexes Index IDs.
+     */
+    // TODO: IGNITE-19082 Needs to be redone/improved
```
sashapolo, line 401: Did you discuss this with @AMashenkov (the author of this ticket) and that it will indeed be done under this ticket?

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
```diff
@@ -361,33 +348,21 @@ public void registerSortedIndex(
     public void unregisterIndex(UUID indexId) {
         indexLockerFactories.remove(indexId);
         indexStorageAdapterFactories.remove(indexId);
+
+        CompletableFuture<?> indexToWaitFuture = indexesToWait.remove(indexId);
+
+        if (indexToWaitFuture != null) {
+            indexToWaitFuture.complete(null);
```
sashapolo, line 355: Do we test this behavior anywhere? Shouldn't this future be completed with an exception? How does this scenario work?

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
```diff
@@ -361,33 +348,21 @@ public void registerSortedIndex(
...
     }
 
     private void awaitIndexes() {
         // TODO: replace with actual call to ids supplier
-        List<UUID> indexIds = List.of(pkId()); // activeIndexIds.get();
-
         List<CompletableFuture<?>> toWait = new ArrayList<>();
 
-        for (UUID indexId : indexIds) {
-            if (indexLockerFactories.containsKey(indexId) && indexStorageAdapterFactories.containsKey(indexId)) {
```
sashapolo, line 373: Why did you remove this check?

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
```diff
@@ -361,33 +348,21 @@ public void registerSortedIndex(
...
+        if (indexToWaitFuture != null) {
+            indexToWaitFuture.complete(null);
+        }
     }
 
     private void awaitIndexes() {
         // TODO: replace with actual call to ids supplier
-        List<UUID> indexIds = List.of(pkId()); // activeIndexIds.get();
-
         List<CompletableFuture<?>> toWait = new ArrayList<>();
```
sashapolo, line 361: Just a nitpick, but you can use an array directly, without an intermediate `ArrayList`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -673,15 +675,18 @@ private CompletableFuture<?> updateAssignmentInternal(ConfigurationNotificationE
                 return failedFuture(e);
             }
 
+            TableImpl table = tablesById.get(tblId);
+
+            // TODO: IGNITE-19082 Need another way to wait for indexes
+            table.addIndexesToWait(collectTableIndexes(tablesCfg.value(), tblId));
```
sashapolo, line 681: Why is this code located in `onUpdateAssignments` and not in table or index creation listeners?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2345,4 +2350,20 @@ private static void handleExceptionOnCleanUpTablesResources(
             nodeStoppingEx.set(true);
         }
     }
+
+    private static Collection<UUID> collectTableIndexes(TablesView tablesConfig, UUID tableId) {
```
sashapolo, line 2354: Another little nitpick: this method does not have to be `static`, since `tablesConfig` is a field of this class and can be removed from parameters

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sashapolo: Why is this code even needed? I can't understand this from the ticket description and we already populate in index-to-wait map when registering indices

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: Fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
tkalkirill: Just an array will not work for me, since the size of a `ConcurrentHashMap` can change at any time.

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
tkalkirill: In general, it can be redone for removal.
I will do so.

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
tkalkirill: Right now I'm close with the `IndexManager`, it should finish waiting for the indexes, since it will be removed in this ticket, then its author will need to take care of this problem.

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
tkalkirill: No, right now the deletion of the index is not implemented correctly in general and not tested as it should, I will create a ticket for this.

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
tkalkirill: For me, this check does not make sense at the moment.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: Before recovery raft groups, it is important for us to know the list of indices that will be updated on recovery (in `PartitionListener`), now we are waiting only for the PK index, and we forget about the rest, which is not correct.

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
sashapolo: ok, makes sense

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
sashapolo: Please add a TODO here

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
sashapolo: Can you clarify please?

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
sashapolo: I'm just checking that the author is aware of the problem you are trying to fix and knows about this TODO

modules/table/src/main/java/org/apache/ignite/internal/table/TableImpl.java
tkalkirill: If the index is registered, then the future of waiting for the indexes will not be or  completed, so there is no need to check the maps.

sashapolo: ok, I've read the code more carefully and looks like this check doesn't make much sense indeed

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sashapolo: After an in-person discussion we decided that this code should actually be moved to the table creation listeners

tkalkirill: Discussed personally, I think that the problem has become clearer.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1106,7 +1107,10 @@ private CompletableFuture<?> createTableLocally(long causalityToken, String name
                 partitions, clusterNodeResolver, txManager, tableStorage, txStateStorage, replicaSvc, clock);
 
         // TODO: IGNITE-16288 directIndexIds should use async configuration API
```
sashapolo, line 1109: Looks like this TODO is not relevant anymore

sashapolo (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: Fix it

ibessonov (approved):

PR #1856, opened 2023-03-28
IGNITE-19018: Sql. Investigate AssertionError in ScanNode.

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-19018


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/DmlPlannerTest.java
korlov42, line 1: I see no changes to the optimisation phase, so why do we need to change planner tests?

korlov42 (approved):

PR #1871, opened 2023-03-30
IGNITE-18742 Implement initial lease granting on placement driver side

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-18742


modules/placement-driver/src/integrationTest/java/org/apache/ignite/internal/placementdriver/PlacementDriverManagerTest.java
```diff
@@ -321,6 +449,45 @@ private void checkLeaseCreated(TablePartitionId grpPartId) throws InterruptedExc
...
     }
 
+    /**
+     * Checks that the lease for the group is accepted.
+     *
+     * @param grpPartId Replication group id.
+     * @throws InterruptedException If the waiting is interrupted.
+     */
+    private void checkAccepted(TablePartitionId grpPartId) throws InterruptedException {
+        assertTrue(waitForCondition(() -> {
```
sanpwc, line 459: Hmm, just curious. Is it possible to rewrite it with `assertThat(leaseFut, willBe());` ?

modules/placement-driver/src/integrationTest/java/org/apache/ignite/internal/placementdriver/PlacementDriverManagerTest.java
```diff
@@ -288,11 +355,72 @@ public void testLeaseholderUpdate() throws Exception {
...
         }, 10_000));
     }
 
+    @Test
+    public void testLeaseAccepted() throws Exception {
+        TablePartitionId grpPart0 = createTableAssignment();
+
+        checkLeaseCreated(grpPart0);
+
+        checkAccepted(grpPart0);
```
sanpwc, line 369: Should we rename it to `checkLeaseAccepted`?

modules/placement-driver/src/integrationTest/java/org/apache/ignite/internal/placementdriver/PlacementDriverManagerTest.java
```diff
@@ -288,11 +355,72 @@ public void testLeaseholderUpdate() throws Exception {
...
+    }
+
+    @Test
+    public void testExceptionOnAcceptance() throws Exception {
+        CountDownLatch latch = new CountDownLatch(1);
+
+        leaseGrantHandler = (req, sender) -> {
+            latch.countDown();
+
+            throw new RuntimeException("test");
```
sanpwc, line 394: Why we need it?

modules/placement-driver/src/integrationTest/java/org/apache/ignite/internal/placementdriver/PlacementDriverManagerTest.java
```diff
@@ -321,6 +449,45 @@ private void checkLeaseCreated(TablePartitionId grpPartId) throws InterruptedExc
...
+
+                return lease.isAccepted();
+            }
+
+            return false;
+        }, 10_000));
+    }
+
+    /**
+     * Reads a leas from Meta storage. When you call this method the lease is expected presenting in Meta storage.
```
sanpwc, line 475: typos

- leas -> lease
- the lease is expected presenting in Meta storage. -> the lease is expected to be present in Meta storage.

sanpwc (changes requested):

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
```diff
@@ -146,18 +172,53 @@ public void deactivate() {
...
+            }
+
+            if (hasInAssignments) {
+                ClusterNode candidate = topologyTracker.nodeByConsistentId(agreement.getRedirectTo());
+
+                if (candidate != null) {
+                    return candidate;
+                }
+            }
+        }
```
denis-chudov, line 210: I strongly doubt that this should be handled here. I would mention the interface that I have suggested in https://issues.apache.org/jira/browse/IGNITE-18879 : on this stage we need only candidate for a group, without consideration of redirect proposal. Moreover, I see no reason to store intermediate state with redirect, we can only remember that we have chosen the candidate (and that candidate was saved to meta storage) and redirect proposal can be handled reactively.

Additionally (and it is not only about this part of code) the implementation seems too complicated for me and definitely lacks reactiveness. We have updater scheduler, map with leases, map with lease agreements, metastorage update listeners that can change these maps, several intermediate states, etc. - we will have pretty long explanation just to describe all of the logic that is contained here. I suggest to think about how to simplify it.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
```diff
@@ -179,24 +240,32 @@ public void run() {
 
                     Lease lease = leaseTracker.getLease(grpId);
 
+                    if (!lease.isAccepted()) {
+                        LeaseAgreement agreement = leaseConciliator.conciliated(grpId);
+
+                        if (agreement.isAccepted()) {
+                            acceptLeasInMetaStorage(grpId, lease);
+                        }
+                    }
```
denis-chudov, line 249: There should be a check that the agreement contains the same lease.

denis-chudov (changes requested):

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
```diff
@@ -130,6 +144,18 @@ public void deInit() {
      * Activates a lease updater to renew leases.
      */
     public void activate() {
+        leaseConciliator = new LeaseConciliator(clusterService);
```
sanpwc, line 147: Busy locks are missing everywhere except stop!

modules/placement-driver/src/integrationTest/java/org/apache/ignite/internal/placementdriver/PlacementDriverManagerTest.java
```diff
@@ -111,11 +128,18 @@ public class PlacementDriverManagerTest extends IgniteAbstractTest {
...
+    /** This closure handles {@link LeaseGrantedMessage} to check the placement driver manager behavior. */
+    private BiFunction<LeaseGrantedMessage, String, LeaseGrantedMessageResponse> leaseGrantHandler;
+
     @BeforeEach
     public void beforeTest(TestInfo testInfo) throws NodeStoppingException {
         this.nodeName = testNodeName(testInfo, PORT);
+        this.anotherNodeName = testNodeName(testInfo, PORT + 1);
         this.testInfo = testInfo;
 
+        assertTrue(nodeName.hashCode() < anotherNodeName.hashCode(),
```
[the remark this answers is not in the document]
vldpyatkov: In my patch, yes.
If we rework that code in the future, required to rewrite the test.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/conciliation/LeaseConciliator.java
```diff
@@ -0,0 +1,122 @@
...
+ */
+public class LeaseConciliator {
+    /** The logger. */
+    private static final IgniteLogger LOG = Loggers.forClass(LeaseConciliator.class);
+
+    /** Placement driver messages factory. */
+    private static final PlacementDriverMessagesFactory PLACEMENT_DRIVER_MESSAGES_FACTORY = new PlacementDriverMessagesFactory();
+
+    /** Leases ready to accept. */
+    private final Map<ReplicationGroupId, LeaseAgreement> leaseToConciliate;
```
denis-chudov, line 45: The purpose of lease agreement is to store the redirect proposals, right? Because Lease itself has information about whether it was accepted or not. Seems that if we had reactive logic of handling lease grant responses, the lease agreements wouldn't be needed at all. Besides, we have separate ticket to handle lease grant responses.

modules/placement-driver/src/integrationTest/java/org/apache/ignite/internal/placementdriver/PlacementDriverManagerTest.java
vldpyatkov: The test shows that our logic is persistent against any exception.

modules/placement-driver/src/integrationTest/java/org/apache/ignite/internal/placementdriver/PlacementDriverManagerTest.java
```diff
@@ -111,11 +128,18 @@ public class PlacementDriverManagerTest extends IgniteAbstractTest {
 
     private TestInfo testInfo;
 
+    /** This closure handles {@link LeaseGrantedMessage} to check the placement driver manager behavior. */
+    private BiFunction<LeaseGrantedMessage, String, LeaseGrantedMessageResponse> leaseGrantHandler;
```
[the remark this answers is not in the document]
vldpyatkov: Here is a test, I will use it for all the tests needs without thread safe violation.

modules/placement-driver/src/integrationTest/java/org/apache/ignite/internal/placementdriver/PlacementDriverManagerTest.java
vldpyatkov: Here is a waiting of the lease creation. Of course, it is possible to create an infinity loop and complete a future after the loop found the lease, but that will be another code.

modules/placement-driver/src/integrationTest/java/org/apache/ignite/internal/placementdriver/PlacementDriverManagerTest.java
```diff
@@ -321,6 +449,45 @@ private void checkLeaseCreated(TablePartitionId grpPartId) throws InterruptedExc
...
+            var leaseEntry = leaseFut.join();
+
+            if (leaseEntry != null && !leaseEntry.empty()) {
+                Lease lease = ByteUtils.fromBytes(leaseEntry.value());
+
+                return lease.isAccepted();
+            }
+
+            return false;
+        }, 10_000));
```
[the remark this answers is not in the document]
vldpyatkov: We don't deserialize the object in `checkLeaseCreated` method. But if you insist, I will rewrite `checkLeaseCreated` and that through the one method (`readLease`)

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/Lease.java
```diff
@@ -0,0 +1,143 @@
...
+import org.apache.ignite.internal.hlc.HybridTimestamp;
+import org.apache.ignite.network.ClusterNode;
+
+/**
+ * A lease representation in memory.
+ * The real lease is stored in Meta storage.
+ */
+public class Lease implements Serializable {
+    /** The object is used when nothing holds the lease. */
+    public static Lease EMPTY_LEASE = new Lease(null, null, null, true);
```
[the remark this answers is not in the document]
vldpyatkov: Because it is convenient to use.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/conciliation/LeaseConciliator.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.ignite.internal.placementdriver.leases.Lease;
+import org.apache.ignite.internal.placementdriver.message.LeaseGrantedMessageResponse;
+import org.apache.ignite.internal.placementdriver.message.PlacementDriverMessagesFactory;
+import org.apache.ignite.internal.replicator.ReplicationGroupId;
+import org.apache.ignite.network.ClusterService;
+
+/**
+ * This class conciliates a lease with leaseholder. If the lease is conciliated, it is ready available to accept.
+ */
+public class LeaseConciliator {
```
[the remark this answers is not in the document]
vldpyatkov: The class is needed, of course. Because a logic of negotiation with remote node might be more complicated than now and it doesn't relate of any other classes.
Maybe the name of the class is confused. 
Could you propose other else?

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
vldpyatkov: Without the logic, I cannot write a test. If it is now, it does not mean that it cannot be rewritten in the future.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
vldpyatkov: Here, checking a lease is not required because, because we cannot update the corresponded entry in Meta storage if it is not actual.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/conciliation/LeaseConciliator.java
vldpyatkov: This will part of another ticket.

modules/placement-driver/build.gradle
```diff
@@ -22,8 +22,9 @@ apply from: "$rootDir/buildscripts/java-integration-test.gradle"
 apply from: "$rootDir/buildscripts/java-test-fixtures.gradle"
 
 dependencies {
-    implementation project(':ignite-metastorage-api')
+    api project(':ignite-placement-driver-api')
```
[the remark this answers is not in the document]
denis-chudov: `api` is used for exporting transitive modules, i see no reason to use it here

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
vldpyatkov: Activation happens in buzy lock now.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
```diff
@@ -210,39 +237,88 @@ public void run() {
...
         /**
-         * Checks that a leaseholder candidate can take a lease on the replication group.
+         * Checks that the lease is outdated.
          *
          * @param lease Lease.
-         * @param candidate The node is a leaseholder candidate.
          * @return True when the candidate can be a leaseholder, otherwise false.
          */
-        private boolean isReplicationGroupUpdateLeaseholder(Lease lease, ClusterNode candidate) {
+        private boolean isLeaseIsOutdated(Lease lease) {
```
denis-chudov, line 318: Actually, this returns true not only if the lease is outdated, but when there is no lease at all (due to EMPTY_LEASE). Please rename this to `isLeaseIsOutdatedOrNonExistent` and explain in javadoc why the logic for both cases should be the same: we can't prolong the expired lease because we already have an interval of time when the lease was not active, so we must start ne negotiation round from the beginning; the same we do for the groups that don't have leaseholders at all.

denis-chudov (changes requested):

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/Lease.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /**
+     * Prolongs a lease until new timestamp. Only an accepted lease can be prolonged.
+     *
+     * @param to The new lease expiration timestamp.
+     * @return A new lease which will have the same properties except of expiration timestamp.
+     */
+    public Lease prolongLease(HybridTimestamp to) {
+        assert accepted : "The lease should be accepted by leaseholder before prolongation ["
```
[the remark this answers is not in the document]
vldpyatkov: Not clear. This method cannot be called with not accepted lease.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
```diff
@@ -181,22 +210,20 @@ public void run() {
 
                     HybridTimestamp now = clock.now();
 
-                    // Nothing holds the lease.
-                    if (lease == EMPTY_LEASE
-                            // The lease is near to expiration.
-                            || now.getPhysical() > (lease.getLeaseExpirationTime().getPhysical() - LEASE_PERIOD / 2)) {
+                    // The lease is expired or near to the one.
+                    if (now.getPhysical() > (lease.getExpirationTime().getPhysical() - LEASE_PERIOD / 2)) {
```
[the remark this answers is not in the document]
vldpyatkov: I don't sure that clock skew is required here. Because here we are doing a check proactive comparison, far a way from so little interval like a clock skew is.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
```diff
@@ -54,7 +58,13 @@ public class LeaseUpdater {
     private static final long UPDATE_LEASE_MS = 200L;
 
     /** Lease holding interval. */
-    private static final long LEASE_PERIOD = 10 * UPDATE_LEASE_MS;
+    public static final long LEASE_PERIOD = 10 * UPDATE_LEASE_MS;
```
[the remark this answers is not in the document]
vldpyatkov: Yes, I will change the word everywhere.

denis-chudov (changes requested):

denis-chudov (approved):

PR #1873, opened 2023-03-31
IGNITE-18875 :  Sql. Drop AbstractPlannerTest.TestTable.

yimengael:
https://issues.apache.org/jira/browse/IGNITE-18875


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/rel/IndexScanNodeExecutionTest.java
```diff
@@ -438,7 +430,7 @@ private void validateIndexScan(
                 ectx,
                 ectx.rowHandler().factory(ectx.getTypeFactory(), rowType),
                 index,
-                new TestTable(rowType, schemaDescriptor),
+                new org.apache.ignite.internal.sql.engine.framework.TestTable(rowType, schemaDescriptor),
```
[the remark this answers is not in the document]
yimengael: @zstan What are the different distribution for TestTable ? I know there is `broadcast`, what else do we have as distribution ? I would base my thinking on that in order to create decorator pattern for each type of distribution. So i would have different TestTable for each type of distribution.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/table/AbstractTestTable.java
```diff
@@ -0,0 +1,300 @@
...
+import org.apache.ignite.internal.sql.engine.trait.IgniteDistribution;
+import org.apache.ignite.internal.sql.engine.trait.TraitUtils;
+import org.apache.ignite.internal.table.InternalTable;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * AbstractTestTable.
+ * TODO Documentation
+ */
+public class AbstractTestTable implements IgniteTable {
```
zstan, line 64: What purpose of such new class ? Why you didn`t reuse org.apache.ignite.internal.sql.engine.framework.TestTable ? Also you call `new AbstractTestTable(...` but haw couls you call constructor of abstract class ?

yimengael: @zstan The problem that the class `org.apache.ignite.internal.sql.engine.framework.TestTable` is causing is that it is specifically used for some other class and not for all the class in `planner` packages. Trying to merge the class `AbstractTestTable` and TestTable will cause lots of tests to fail because of the reasons I gave above.

Short term solution : 
I found was to separate the two classes `org.apache.ignite.internal.sql.engine.framework.TestTable` and `org.apache.ignite.internal.sql.engine.table.AbstractTestTable`, so that each of them can achieve its primary goal.

Long term solution :
Identifying all the use cases of the TestTable in the code, list of the different distributions used by each type of TestTable, Using a decorator pattern to create an abstraction of the TestTable that can be used all over the test Framework. This I will not consider this as an easy.

This `org.apache.ignite.internal.sql.engine.table.AbstractTestTable` is not an abstract class, it is juts a naming to make different from the other TestTable `org.apache.ignite.internal.sql.engine.framework.TestTable`.

sk0x50:
The issue was fixed at https://github.com/apache/ignite-3/pull/2372

PR #1874, opened 2023-03-31
IGNITE-19169 Deadlock detected while calling MvPartitionStorage#pollForVacuum

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19169


modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
```diff
@@ -52,11 +52,15 @@
...
 import org.apache.ignite.lang.IgniteException;
 import org.jetbrains.annotations.Nullable;
+import org.junit.jupiter.api.RepeatedTest;
 
 /**
  * Base test for MV storages, contains pojo classes, their descriptor and a marshaller instance.
  */
 public abstract class BaseMvStoragesTest {
+    /** To be used in a loop. {@link RepeatedTest} has a smaller failure rate due to recreating the storage every time. */
+    protected static final int REPEATS = 100;
```
ibessonov, line 62: What's the reason for moving this constant here? Are you sure that every subclass should use the same value?
Maybe new class should have higher repeat value, because deadlock is harder to reproduce

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/AbstractMvStorageUpdateHandlerTest.java
```diff
@@ -0,0 +1,144 @@
...
+        partitionDataStorage = new TestPartitionDataStorage(mvPartitionStorage);
+
+        storageUpdateHandler = new StorageUpdateHandler(PARTITION_ID, partitionDataStorage, Map::of, tableConfig.dataStorage());
+    }
+
+    @AfterEach
+    void tearDown() {
+        if (storageEngine != null) {
+            storageEngine.stop();
+        }
```
ibessonov, line 91: You forgot to stop the table, and partition

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/AbstractMvStorageUpdateHandlerTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+            addWriteCommitted(partitionDataStorage, rowId0, null, clock.now());
+            addWriteCommitted(partitionDataStorage, rowId1, null, clock.now());
+
+            runRace(
+                    () -> storageUpdateHandler.vacuumBatch(HybridTimestamp.MAX_VALUE, 2),
+                    () -> storageUpdateHandler.vacuumBatch(HybridTimestamp.MAX_VALUE, 2),
+                    () -> storageUpdateHandler.vacuumBatch(HybridTimestamp.MAX_VALUE, 2),
+                    () -> storageUpdateHandler.vacuumBatch(HybridTimestamp.MAX_VALUE, 2),
+                    () -> storageUpdateHandler.vacuumBatch(HybridTimestamp.MAX_VALUE, 2)
```
ibessonov, line 128: Having 5 threads is unrealistic, we only have two in real system

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/BaseMvStoragesTest.java
tkalkirill: Fix it

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/AbstractMvStorageUpdateHandlerTest.java
tkalkirill: Fix it

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/AbstractMvStorageUpdateHandlerTest.java
tkalkirill: Fix it

ibessonov (approved):

PR #1876, opened 2023-03-31
IGNITE-19135 Benchmarks code for puts/reads

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-19135


modules/runner/src/integrationTest/java/org/apache/ignite/internal/benchmark/InsertBenchmark.java
```diff
@@ -0,0 +1,203 @@
...
+
+            statement = sql.createStatement(queryStr);
+            session = sql.createSession();
+        }
+
+        /**
+         * Closes resources.
+         */
+        @TearDown
+        public void tearDow() throws Exception {
```
kgusakov, line 126:
```suggestion
        public void tearDown() throws Exception {
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/benchmark/InsertBenchmark.java
```diff
@@ -0,0 +1,203 @@
...
+            IgniteUtils.closeAll(session/*, statement*/);
```
kgusakov, line 127: Why statement is commented?

kgusakov (changes requested):

kgusakov:
@ibessonov In general, I think we should run these benchmarks on the nightly run of main only, but not on the every change, what do you think? Benchmarks have big enough time and resource consumption and I suppose this suite will grow up.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/benchmark/InsertBenchmark.java
ibessonov: Because it throws UnsupportedOperationException, I need to add a comment

ibessonov:
@kgusakov These benchmarks don't run on TC at all, because they are not part of integration tests. They just exist in the same module. I expect people, who need to benchmark their fix, to do it locally on their own.

PR #1880, opened 2023-03-31
IGNITE-18985 Sql. Jdbc. Integration with transaction

xtern:


modules/client-common/src/main/java/org/apache/ignite/internal/jdbc/proto/event/JdbcBatchPreparedStmntRequest.java
```diff
@@ -88,9 +93,19 @@ public List<Object[]> getArgs() {
         return args;
     }
 
+    /**
+     * Get flag indicating whether auto-commit mode is enabled.
+     *
+     * @return Flag indicating whether auto-commit mode is enabled.
```
AMashenkov, line 99:
```suggestion
     * @return {@code true} if auto-commit mode is enabled, {@code false} otherwise.
```

modules/client-common/src/main/java/org/apache/ignite/internal/jdbc/proto/event/JdbcBatchExecuteRequest.java
```diff
@@ -72,9 +77,19 @@ public List<String> queries() {
         return queries;
     }
 
+    /**
+     * Get flag indicating whether auto-commit mode is enabled.
+     *
+     * @return Flag indicating whether auto-commit mode is enabled.
```
AMashenkov, line 83:
```suggestion
     * @return {@code true} if auto-commit mode is enabled, {@code false} otherwise.
```

modules/client-common/src/main/java/org/apache/ignite/internal/jdbc/proto/event/JdbcQueryExecuteRequest.java
```diff
@@ -128,9 +133,19 @@ public JdbcStatementType getStmtType() {
         return stmtType;
     }
 
+    /**
+     * Get flag indicating whether auto-commit mode is enabled.
+     *
+     * @return Flag indicating whether auto-commit mode is enabled.
```
AMashenkov, line 139:
```suggestion
     * @return {@code true} if auto-commit mode is enabled, {@code false} otherwise.
```

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcConnection.java
```diff
@@ -367,6 +374,20 @@ public void rollback(Savepoint savepoint) throws SQLException {
...
     }
 
+    /**
+     * Finish transaction.
+     *
+     * @param commit {@code True} to commit, {@code false} to rollback.
+     * @throws SQLException If failed.
+     */
+    private void finishTx(boolean commit) throws SQLException {
+        JdbcFinishTxResult res = handler().finishTxAsync(connectionId, commit).join();
```
AMashenkov, line 384: join() is uninterruptible call.

AMashenkov (approved):

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -287,6 +300,26 @@ public CompletableFuture<JdbcMetaPrimaryKeysResult> primaryKeysMetaAsync(JdbcMet
...
+    public CompletableFuture<JdbcFinishTxResult> finishTxAsync(long connectionId, boolean commit) {
+        JdbcConnectionContext connectionContext;
+
+        try {
+            connectionContext = resources.get(connectionId).get(JdbcConnectionContext.class);
+        } catch (IgniteInternalCheckedException exception) {
+            return CompletableFuture.completedFuture(new JdbcFinishTxResult(Response.STATUS_FAILED, "Connection is broken"));
+        }
+
+        return connectionContext.finishTransactionAsync(commit).handle((ignored, t) -> {
```
AMashenkov, line 314: Also, I'd split methods with flags (like connectionContext.finishTransactionAsync) into commitAsync and rollbackAsync to be. more specific.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -287,6 +300,26 @@ public CompletableFuture<JdbcMetaPrimaryKeysResult> primaryKeysMetaAsync(JdbcMet
         return meta.getPrimaryKeys(req.schemaName(), req.tableName()).thenApply(JdbcMetaPrimaryKeysResult::new);
     }
 
+    /** {@inheritDoc} */
+    @Override
+    public CompletableFuture<JdbcFinishTxResult> finishTxAsync(long connectionId, boolean commit) {
+        JdbcConnectionContext connectionContext;
+
+        try {
+            connectionContext = resources.get(connectionId).get(JdbcConnectionContext.class);
```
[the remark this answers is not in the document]
xtern: Probably yes, but I don’t quite understand from the code where we removing it now.
Currently, if the resource (ConnectionContext) is released the active transaction is rolled back.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
xtern: I tried it and I like it better the way it is now (less code/duplicate code), but if you insist I can split it.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -228,16 +240,17 @@ public CompletableFuture<JdbcBatchExecuteResult> batchPrepStatementAsync(long co
...
+    private CompletableFuture<Long> executeAndCollectUpdateCount(long connectionId, boolean autoCommit, String sql, Object[] arg) {
         JdbcConnectionContext connectionContext;
         try {
             connectionContext = resources.get(connectionId).get(JdbcConnectionContext.class);
         } catch (IgniteInternalCheckedException exception) {
             return CompletableFuture.failedFuture(new IgniteInternalException(Client.CONNECTION_ERR));
         }
 
+        Transaction tx = autoCommit ? null : connectionContext.getOrStartTransaction();
+        var context = createQueryContext(JdbcStatementType.UPDATE_STATEMENT_TYPE, tx);
```
korlov42, line 252: keyword `var` is not allowed in this context

modules/client-handler/src/test/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImplTest.java
korlov42, line 1: I believe we have to add a few new tests to this class

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcBatchSelfTest.java
```diff
@@ -636,6 +619,61 @@ public void testBatchClearPrepared() throws SQLException {
...
+    @Test
+    public void testTransactionalBatchPrepared() throws Exception {
+        conn.setAutoCommit(false);
+
+        int rowsCnt = 5;
+
+        Callable<int[]> batchOp = () -> {
+            populatePrepared(rowsCnt);
+
+            return pstmt.executeBatch();
```
korlov42, line 652: do you have test for transaction spanning several calls to `executeBatch`?

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -442,6 +481,22 @@ private SessionId recreateSession(@Nullable SessionId expectedSessionId) {
             }
         }
 
+        private Transaction getOrStartTransaction() {
+            return tx == null ? tx = igniteTransactions.begin() : tx;
```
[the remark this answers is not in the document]
xtern: My understanding is that each JDBC connection under the hood holds a client, each client holds a channel, and each channel is served by a separate worker thread. One worker thread can serve multiple channels, but each channel is serviced by the same worker thread. Therefore it seemed to me that this method cannot be called by different threads.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
xtern: Done, thanks

modules/client-handler/src/test/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImplTest.java
xtern: Added a couple of tests with mocks.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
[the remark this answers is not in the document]
xtern: Thanks, I believe I've fixed this.

Added context close flag.
Transaction creation has been moved above batch asynchronous processing.
Added javadocs that methods are not thread safe.

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcBatchSelfTest.java
xtern: This test case is now covered by integration tests.

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
```diff
@@ -0,0 +1,307 @@
...
+     * <li>Call to {@code setAutoCommit} or {@code getAutoCommit} on a closed connection throws exception.
+     * </ul>
+     *
+     * @throws SQLException If failed.
+     */
+    @Test
+    public void testAutoCommitModeChange() throws SQLException {
+        try (Connection conn = DriverManager.getConnection(URL)) {
+            assertTrue(conn.getAutoCommit());
+            conn.setAutoCommit(true);
```
korlov42, line 77: how exactly you are making sure nothing is happened when you trying to change autoCommit mode to the same state?

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
```diff
@@ -0,0 +1,307 @@
...
+
+    /**
+     * Tests {@link Connection#commit()} method behaviour.
+     *
+     * <p>Calling {@code commit} is expected to throw an exception if called in auto-commit mode or on a closed connection,
+     *
+     * @throws SQLException If failed.
+     */
+    @Test
+    public void testCommit() throws Exception {
```
korlov42, line 114: this test should be an unit test

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
```diff
@@ -0,0 +1,307 @@
...
+
+    /**
+     * Tests {@link Connection#rollback()} method behaviour.
+     *
+     * <p>Calling {@code rollback} is expected to throw an exception if called in auto-commit mode or on a closed connection,
+     *
+     * @throws SQLException If failed.
+     */
+    @Test
+    public void testRollback() throws Exception {
```
korlov42, line 137: this one should be unit as well

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
```diff
@@ -0,0 +1,307 @@
...
+            try (ResultSet rs = stmt.executeQuery("select count(*) from TEST")) {
+                rs.next();
+
+                return rs.getInt(1);
+            }
+        }
+    }
+
+    @FunctionalInterface
+    private interface BiConsumerX<T, U> {
```
korlov42, line 304: Why not introduce an interface with a meaningful name?

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
xtern: We just check that there were no exceptions.
Maybe it's redundant.

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
xtern: These tests (commit/rollback/autocommit) have been moved from ItJdbcConnectionSelfTest because you suggested moving all related to jdbc transaction tests to a separate test suite.
Personally, I don't see a big problem with these tests sitting in an integration suite (they don't take much time).
What are you suggesting to do with them - create a new Jdbc unit test with them  or leave them in ItJdbcConnectionSelfTest or move them to JdbcQueryEventHandlerImplTest?

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
```diff
@@ -0,0 +1,307 @@
...
+     *
+     * <p>It is expected that in non auto-commit mode, data updates can be rolled back and
+     * are invisible outside of an active transaction until they are committed.
+     *
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testBatchPrepared() throws Exception {
+        try (PreparedStatement pstmt = conn.prepareStatement(SQL_INSERT_PREPARED)) {
+            checkUpdate(5, 10, (off, cnt) -> {
```
korlov42, line 221: So, you've passed both rowsPerOp and actual op closure as a params to the checkUpdate to propagate them to checkTxResult to pass rowsPerOp as an input to op closure...

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
```diff
@@ -0,0 +1,307 @@
...
+     *
+     * @param ops Number of operations.
+     * @param rowsPerOp Number of rows per operation.
+     * @param batchOp Update operation.
+     * @param commit {@code True} to check transaction commit, {@code false} to check rollback.
+     * @throws SQLException If failed.
+     */
+    private void checkTxResult(int ops, int rowsPerOp, BiConsumerX<Integer, Integer> batchOp, boolean commit) throws SQLException {
+        // Creating another connection to ensure that the changes are not visible outside of the transaction.
+        try (Connection conn0 = DriverManager.getConnection(URL)) {
```
korlov42, line 264: It took some time to realise that conn0 is not the connection used to make an update... Let's rework this method to make it obvious

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
```diff
@@ -0,0 +1,307 @@
...
+     *
+     * <p>It is expected that in non auto-commit mode, data updates can be rolled back and
+     * are invisible outside of an active transaction until they are committed.
+     *
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testExecuteUpdatePrepared() throws Exception {
+        try (PreparedStatement pstmt = conn.prepareStatement(SQL_INSERT_PREPARED)) {
+            checkUpdate(2, 1, (id, ignore) -> {
```
korlov42, line 180: what about tests for autoCommit=true mode?

modules/client-handler/src/test/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImplTest.java
```diff
@@ -174,14 +182,71 @@ void sessionIdIsUsedForQueryExecution() {
...
+        verify(igniteTransactions).begin();
+        verify(tx, times(0)).rollbackAsync();
+
+        resourceRegistry.close();
+        verify(tx, times(1)).rollbackAsync();
+        verifyNoMoreInteractions(tx);
+    }
+
+    @Test
+    public void singleTxUsedForMultipleOperations() {
```
korlov42, line 213: let's add tests for commit/rollback command

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -398,6 +481,10 @@ void close() {
         }
 
         <T> CompletableFuture<T> doInSession(SessionAwareAction<T> action) {
+            if (closed) {
```
korlov42, line 484: I don't see how this helps against concurrent connection closing

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
xtern: It seems, that currently we have enough tests for auto-commit mode - ItJdbcStatementSelfTest, ItJdbcBatchSelfTest, and so on.

Do you suggest adding some checks here for this mode as well?

modules/client-handler/src/test/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImplTest.java
xtern: This test already verifies this command (JdbcQueryEventHandler#finishTxAsync).
Are you suggesting to split it into two separate tests (one for commit and one for rollback)?

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
xtern: Thanks, reworked a bit.

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
xtern: Reworked a bit :roll_eyes:

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
xtern: Thanks, fixed.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
xtern: Thank you. Slightly reworked, currently I just prevent the session from being re-created after the context is closed.

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
xtern: This check has been reworked.

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
xtern: This test has been moved to `JdbcQueryEventHandlerImplTest`

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
xtern: Test has been moved to `JdbcQueryEventHandlerImplTest`

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -514,6 +510,10 @@ <T> CompletableFuture<T> doInSession(SessionAwareAction<T> action) {
 
         private SessionId recreateSession(@Nullable SessionId expectedSessionId) {
             synchronized (mux) {
+                if (closed) {
+                    throw new IgniteInternalException(CONNECTION_ERR, "Connection is closed");
```
korlov42, line 514: it's better to cover this with tests to see what will happen

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
korlov42: I wonder how you chose a new place for this test

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
korlov42: ok, fair enough

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcTransactionTest.java
xtern: In order to add this test to ignite-jdbc, we need to mock the JdbcQueryEventHandler (connect() and finishTxAsync() methods) and therefore we will need to add the mockito dependency. And all of this just to check that the two methods are throwing an exception.
It seemed to me that this would already be a copy-paste, and among other things, in the JdbcQueryEventHandler test, I can check that extra transactions do not start.

And as I said before, I don't see any major issues with keeping these tests in the integration test suite, so I reverted this change for now.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
xtern: I agree, but creating a good test for such a scenario seems not an easy task.
Added something like a test called `sessionCreationAfterCloseRegistry` to cover this case.

korlov42 (approved):

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcConnection.java
AMashenkov: I mean the thread waiting on future.get() can be interrupted and a client application can continue work.
The future.join() ignores interrupts and the thread will hang forever. The only way is to kill JVM.
We have faced similar issues with future.join() on TC that lead to test suite hanging.

xtern: I understood that you are suggesting to replace `join()` with `get()`, and I indicated above why I am not sure that it is worth doing this specifically for commit() and rollback().
As I see our jdbc driver implementation currently does not handle thread interruptions at all (except client connect).

Are you suggesting to change all join() calls to get() there and wrap InterruptedException in SqlException?

And as I pointed above, the jdbc specification does not describe interruptionexception handling. The operation is expected to be aborted via JdbcConnection#cancel.

xtern: @AMashenkov  Thanks, finally I did replace all ``join()` to `get()`.
Since thread interruption is not described in any way by jdbc, then the user should not use it, but it can really help us in development and testing.
The only note - at the moment I do not restore the thread interruption flag when handling `InterruprionException`, please see the changes in the last two commits, is that ok?

PR #1883, opened 2023-04-02
IGNITE-19181 ConfigurationNotificationEvent API modified to include both old and new node/name values.

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-19181

### DIFF
--- a/modules/configuration-api/src/main/java/org/apache/ignite/configuration/notifications/ConfigurationNotificationEvent.java
+++ b/modules/configuration-api/src/main/java/org/apache/ignite/configuration/notifications/ConfigurationNotificationEvent.java
@@ -49,22 +49,42 @@ public interface ConfigurationNotificationEvent<VIEWT> {
     long storageRevision();
 
     /**
-     * Returns the parent (any from the root) or current configuration.
+     * Returns old value of the parent (any from the root) or current configuration.
      *
      * <p>For example, if we changed the child configuration, then we can get both the parent and the current child configuration.
      *
-     * @param configClass Configuration interface, for example {@code RootConfiguration}.
+     * @param viewClass Configuration interface, for example {@code RootView.class}.
      * @param <T> Configuration type.
      */
-    @Nullable <T extends ConfigurationProperty> T config(Class<T> configClass);
+    @Nullable <T> T oldValue(Class<T> viewClass);
 
     /**
-     * Returns the key of a named list item for the parent (any from the root) or current configuration.
+     * Returns new value of the parent (any from the root) or current configuration.
+     *
+     * <p>For example, if we changed the child configuration, then we can get both the parent and the current child configuration.
+     *
+     * @param viewClass Configuration interface, for example {@code RootView.class}.
+     * @param <T> Configuration type.
+     */
+    @Nullable <T> T newValue(Class<T> viewClass);
+
+    /**
+     * Returns old value of the key of a named list item for the parent (any from the root) or current configuration.
      *
      * <p>For example, if a column of a table has changed, then we can get the name of the table and columns for which the changes have
      * occurred.
      *
-     * @param configClass Configuration interface, for example {@code TableConfiguration}.
+     * @param viewClass Configuration interface, for example {@code TableView.class}.
      */
-    @Nullable String name(Class<? extends ConfigurationProperty> configClass);
+    @Nullable String oldName(Class<?> viewClass);
+
+    /**
+     * Returns new value of the key of a named list item for the parent (any from the root) or current configuration.
+     *
+     * <p>For example, if a column of a table has changed, then we can get the name of the table and columns for which the changes have
+     * occurred.
+     *
+     * @param viewClass Configuration interface, for example {@code TableView.class}.
+     */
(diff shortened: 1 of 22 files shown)


modules/configuration-api/src/main/java/org/apache/ignite/configuration/notifications/ConfigurationNotificationEvent.java
```diff
@@ -35,11 +35,31 @@
...
 
+    /**
+     * Returns old value of the parent (any from the root) or current configuration.
+     *
+     * <p>For example, if we changed the child configuration, then we can get both the parent and the current child configuration.
+     *
+     * @param viewClass Configuration interface, for example {@code RootView.class}.
+     * @param <T> Configuration type.
+     */
+    @Nullable <T> T oldValue(Class<T> viewClass);
```
tkalkirill, line 46: In fact, `oldValue()` is a special case of this method, do you think it's worth getting rid of `oldValue()` ?

modules/configuration-api/src/main/java/org/apache/ignite/configuration/notifications/ConfigurationNotificationEvent.java
```diff
@@ -35,11 +35,31 @@
...
 
+    /**
+     * Returns new value of the parent (any from the root) or current configuration.
+     *
+     * <p>For example, if we changed the child configuration, then we can get both the parent and the current child configuration.
+     *
+     * @param viewClass Configuration interface, for example {@code RootView.class}.
+     * @param <T> Configuration type.
+     */
+    @Nullable <T> T newValue(Class<T> viewClass);
```
tkalkirill, line 61: Same

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/notifications/ConfigurationContainer.java
```diff
@@ -17,60 +17,93 @@
...
 class ConfigurationContainer {
-    /**
-     * Configuration.
-     *
-     * <p>For {@link ConfigurationNotificationEvent#config} use {@link this#specificConfig()}.
-     */
-    final DynamicConfiguration<InnerNode, ?> config;
+    /** Parent container. */
+    @Nullable
+    public final ConfigurationContainer prev;
```
tkalkirill, line 31:
```suggestion
    public final @Nullable ConfigurationContainer prev;
```

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/notifications/ConfigurationContainer.java
```diff
@@ -17,60 +17,93 @@
...
     /**
-     * Returns the configuration for {@link ConfigurationNotificationEvent#config}.
+     * Finds a node value that corresponds to a given {@code View} class.
+     *
+     * @param clazz Class instance of the {@code *View} type.
+     * @param old Whether an old or a new value is asked.
      */
-    @Nullable ConfigurationProperty<InnerNode> specificConfig() {
-        return config.isRemovedFromNamedList() ? null : config.specificConfigTree();
+    public @Nullable <T> T find(Class<T> clazz, boolean old) {
```
tkalkirill, line 76: Maybe different methods? findOld and findNew ?

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/notifications/ConfigurationContainer.java
```diff
@@ -17,60 +17,93 @@
...
     /**
-     * Returns the configuration class.
+     * Finds a name of the value that corresponds to a given {@code View} class.
+     *
+     * @param clazz Class instance of the {@code *View} type.
+     * @param old Whether an old or a new value is asked.
      */
-    Class<?> configClass() {
-        Class<?> polymorphicInstanceConfigType = config.polymorphicInstanceConfigType();
+    public @Nullable String name(Class<?> clazz, boolean old) {
```
tkalkirill, line 92: same

tkalkirill (approved):

modules/configuration-api/src/main/java/org/apache/ignite/configuration/notifications/ConfigurationNotificationEvent.java
ibessonov: We should leave the method without parameters, because unlike `oldValue(FooView.class)`, `oldValue()` may return `String` / `Integer` / etc.

tkalkirill: Ok

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/notifications/ConfigurationContainer.java
tkalkirill: Just a suggestion, it seems a little clearer to read code like this (findOld, findNew) instead of a parameter. But it's up to you.

PR #1885, opened 2023-04-03
IGNITE-19172 Obsolete TODO comments addressed and removed.

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-19172


modules/storage-api/src/main/java/org/apache/ignite/internal/storage/DataStorageManager.java
```diff
@@ -95,7 +95,6 @@ public void stop() throws Exception {
      *      UnknownDataStorageConfigurationSchema#UNKNOWN_DATA_STORAGE} and there is only one engine, then it will be the default, otherwise
      *      there will be no default.
      */
-    // TODO: IGNITE-16835 Remove it.
```
sashapolo, line 98: Why didn't we remove it?

ibessonov: This method is still required, IGNITE-16835 is closed as `Won't Fix`

sashapolo (approved):

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/DataStorageManager.java
sashapolo: ok

PR #1888, opened 2023-04-03
IGNITE-19075 CLI should ask for SSL settings

valepakh:
https://issues.apache.org/jira/browse/IGNITE-19075


modules/cli/src/main/java/org/apache/ignite/internal/cli/call/connect/ConnectSslConfigCall.java
```diff
@@ -0,0 +1,77 @@
...
+            SslConfig config = input.getConfig();
+            ApiClientSettings settings = ApiClientSettings.builder()
+                    .basePath(input.getUrl())
+                    .keyStorePath(config.keyStorePath())
+                    .keyStorePassword(config.keyStorePassword())
+                    .trustStorePath(config.trustStorePath())
+                    .trustStorePassword(config.trustStorePassword())
+                    .build();
+            ApiClient client = ApiClientFactory.buildClient(settings);
+            new NodeConfigurationApi(client).getNodeConfiguration();
```
PakhomovAlexander, line 60: Could you move this API call to the meaningful method named `checkConnection`?

PR #1890, opened 2023-04-03
IGNITE-19197 Refactor authentication

Flaugh24:
https://issues.apache.org/jira/browse/IGNITE-19197

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
+++ b/modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
@@ -435,6 +435,6 @@ public class ErrorGroups {
         public static final ErrorGroup AUTHENTICATION_ERR_GROUP = ErrorGroup.newGroup("AUTHENTICATION", 15);
 
         /** General authentication error. */
-        public static final int UNKNOWN_AUTHENTICATION_ERR = AUTHENTICATION_ERR_GROUP.registerErrorCode(1);
+        public static final int COMMON_AUTHENTICATION_ERR = AUTHENTICATION_ERR_GROUP.registerErrorCode(1);
     }
 }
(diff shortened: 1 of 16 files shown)


modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -426,4 +426,15 @@ public static class GarbageCollector {
...
+
+    /**
+     * Authentication error group.
+     */
+    public static class Authentication {
+        /** Authentication error group. */
+        public static final ErrorGroup AUTHENTICATION_ERR_GROUP = ErrorGroup.newGroup("AUTHENTICATION", 15);
+
+        /** General authentication error. */
+        public static final int UNKNOWN_AUTHENTICATION_ERR = AUTHENTICATION_ERR_GROUP.registerErrorCode(1);
```
PakhomovAlexander, line 438: In case this is the general auth error, why the variable is called `UNKNOWN_*`? Looking at the rest of the code we can see such variables called `COMMON_*`.

modules/security/src/main/java/org/apache/ignite/internal/security/authentication/AuthenticationManagerImpl.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Implementation of {@link AuthenticationManager}.
+ */
+public class AuthenticationManagerImpl implements AuthenticationManager {
+
+    private static final IgniteLogger LOG = Loggers.forClass(AuthenticationManagerImpl.class);
+
+    private final ReadWriteLock rwLock = new ReentrantReadWriteLock();
+    private final List<Authenticator> authenticators = new ArrayList<>();
```
PakhomovAlexander, line 44: put empty lines between fields, please

PR #1891, opened 2023-04-04
IGNITE-18256 Add record_view<T> in C++

isapego:
Done:
- Introduced `convert_to_tuple` and `convert_from_tuple` function templates to provide conversion between user types and `ignite_tuple`;
- Implemented `record_view<T>` that now works with user types, for which conversion functions were specialized;
- Added tests.

### DIFF
--- a/modules/platforms/cpp/ignite/client/table/record_view.h
+++ b/modules/platforms/cpp/ignite/client/table/record_view.h
@@ -17,11 +17,12 @@
 
 #pragma once
 
-#include "ignite/client/table/ignite_tuple.h"
-#include "ignite/client/transaction/transaction.h"
+#include <ignite/client/table/ignite_tuple.h>
+#include <ignite/client/transaction/transaction.h>
+#include <ignite/client/type_mapping.h>
 
-#include "ignite/common/config.h"
-#include "ignite/common/ignite_result.h"
+#include <ignite/common/config.h>
+#include <ignite/common/ignite_result.h>
 
 #include <memory>
 #include <type_traits>
@@ -36,23 +37,8 @@ namespace detail {
 class table_impl;
 }
 
-/**
- * Record view interface provides methods to access table records.
- */
 template<typename T>
-class record_view {
-public:
-    typedef typename std::decay<T>::type value_type;
-
-    // Deleted
-    record_view(const record_view &) = delete;
-    record_view &operator=(const record_view &) = delete;
-
-    // Default
-    record_view() = default;
-    record_view(record_view &&) noexcept = default;
-    record_view &operator=(record_view &&) noexcept = default;
-};
+class record_view;
 
 /**
  * Record view interface provides methods to access table records.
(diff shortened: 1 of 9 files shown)


modules/platforms/cpp/ignite/client/table/record_view.h
```diff
@@ -475,4 +461,558 @@ class record_view<ignite_tuple> {
...
+        return sync<std::optional<value_type>>(
+            [this, tx, &record](auto callback) { get_and_replace_async(tx, record, std::move(callback)); });
+    }
+
+    /**
+     * Deletes a record with the specified key asynchronously.
+     *
+     * @param tx Optional transaction. If nullptr implicit transaction for this
+     *   single operation is used.
+     * @param key A record with key columns set..
```
ptupitsyn, line 780:
```suggestion
     * @param key A record with key columns set.
```

modules/platforms/cpp/ignite/client/table/table.h
```diff
@@ -63,14 +63,27 @@ class table {
      *
      * @return Record binary view.
      */
-    [[nodiscard]] IGNITE_API record_view<ignite_tuple> record_binary_view() const noexcept;
+    [[nodiscard]] IGNITE_API record_view<ignite_tuple> get_record_binary_view() const noexcept;
```
ptupitsyn, line 66: We have `get_` prefix here, but no prefix for `name()` above - is that right? What is the convention?

modules/platforms/cpp/ignite/client/type_mapping.h
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+#pragma once
+
+#include <ignite/client/table/ignite_tuple.h>
+
+namespace ignite {
+
+/**
+ * Function that specifies how a value of a type should be conversed to @c ignite_tuple.
```
ptupitsyn, line 25:
```suggestion
 * Function that specifies how a value of a type should be converted to @c ignite_tuple.
```

modules/platforms/cpp/ignite/client/type_mapping.h
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @tparam T Type of the value.
+ * @param value Value to convert.
+ * @return An instance of @c ignite_tuple.
+ */
+template<typename T>
+ignite_tuple convert_to_tuple(T &&value);
+
+/**
+ * Function that specifies how an @c ignite_tuple instance should be conversed to specific type.
```
ptupitsyn, line 35:
```suggestion
 * Function that specifies how an @c ignite_tuple instance should be converted to specific type.
```

modules/platforms/cpp/ignite/client/type_mapping.h
```diff
@@ -0,0 +1,65 @@
...
+template<typename T>
+std::optional<ignite_tuple> convert_to_tuple(std::optional<T> &&value) {
+    if (!value.has_value())
+        return std::nullopt;
+
+    return {convert_to_tuple<T>(*std::move(value))};
+}
+
+template<typename T>
+std::optional<T> convert_from_tuple(std::optional<ignite_tuple> &&value) {
```
ptupitsyn, line 58: Missing docs?

modules/platforms/cpp/tests/client-test/record_view_test.cpp
```diff
@@ -0,0 +1,956 @@
...
+    tuple.set("val", value.val);
+
+    return tuple;
+}
+
+template<>
+test_type convert_from_tuple(ignite_tuple&& value) {
+    test_type res;
+
+    res.key = value.get<std::int64_t>("key");
```
ptupitsyn, line 59: What happens if there is no tuple element named "key"?

ptupitsyn (approved):
Looks good to me in general.

modules/platforms/cpp/ignite/client/type_mapping.h
isapego: Addded

modules/platforms/cpp/tests/client-test/record_view_test.cpp
isapego: There is going to be a serialization error. I'll add a test for this case.

ptupitsyn (approved):

PR #1895, opened 2023-04-04
IGNITE-18962 .NET: Ensure consistent ToString format in public types

ptupitsyn:
* Implement `IgniteToStringBuilder` to encapsulate `ToString` logic. Use the same format `TypeName { Prop = Val }` that C# record types use.
* Add `TestAllPublicFacingTypesHaveConsistentToString` that checks all public types and all non-public types that implement public interfaces (= will be handled by user code) for a correct `ToString` override.
* Update all matching types.

Expected usages of `ToString` are:
* Development, prototyping (`Console.WriteLine`) and debugging (debugger shows string representation of variables, fields, etc).
* Logging and exceptions (objects may end up converted to string there).

Therefore, some implementations include internal details (like Table ID), which may help with troubleshooting.

### DIFF
--- a/modules/platforms/dotnet/Apache.Ignite/Apache.Ignite.csproj
+++ b/modules/platforms/dotnet/Apache.Ignite/Apache.Ignite.csproj
@@ -56,7 +56,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Include="..\README.md" Pack="true" PackagePath="\"/>
+    <None Include="..\README.md" Pack="true" PackagePath="\" />
   </ItemGroup>
 
 </Project>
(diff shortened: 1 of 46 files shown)


modules/platforms/dotnet/Apache.Ignite/Apache.Ignite.csproj
```diff
@@ -56,7 +56,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Include="..\README.md" Pack="true" PackagePath="\"/>
+    <None Include="..\README.md" Pack="true" PackagePath="\" />
```
isapego, line 59: Was this change automatic?

isapego (approved):

modules/platforms/dotnet/Apache.Ignite/Apache.Ignite.csproj
ptupitsyn: Yes, this line was added manually previously, but my IDE reformatted the file now. Let's leave it - keep the IDE happy :)

PR #1899, opened 2023-04-04
IGNITE-19179 Make logical topology rejoins idempotent

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-19179

### DIFF
--- a/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/LogicalTopologyImpl.java
+++ b/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/LogicalTopologyImpl.java
@@ -96,6 +96,10 @@ public class LogicalTopologyImpl implements LogicalTopology {
             // This is an update. First simulate disappearance, then appearance will be fired.
             snapshot = new LogicalTopologySnapshot(snapshot.version() + 1, mapByName.values());
 
+            if (LOG.isInfoEnabled()) {
+                LOG.info("Node removed from logical topology [node={}, topology={}]", nodeToPut, snapshot);
+            }
+
             LogicalTopologySnapshot snapshotAfterRemoval = snapshot;
             fireRemovalTask = () -> fireNodeLeft(oldNode, snapshotAfterRemoval);
         }
@@ -103,6 +107,10 @@ public class LogicalTopologyImpl implements LogicalTopology {
         mapByName.put(nodeToPut.name(), nodeToPut);
 
         snapshot = new LogicalTopologySnapshot(snapshot.version() + 1, mapByName.values());
+
+        if (LOG.isInfoEnabled()) {
+            LOG.info("Node added to logical topology [node={}, topology={}]", nodeToPut, snapshot);
+        }
 
         // Only save to storage once per call so that our writes to storage are atomic and we don't end up in a situation
         // when different CMG listener instances produce different sequences of topology snapshots.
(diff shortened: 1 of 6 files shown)


modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/LogicalTopologyImpl.java
```diff
@@ -96,6 +96,10 @@ public void putNode(LogicalNode nodeToPut) {
             // This is an update. First simulate disappearance, then appearance will be fired.
             snapshot = new LogicalTopologySnapshot(snapshot.version() + 1, mapByName.values());
 
+            if (LOG.isInfoEnabled()) {
```
ibessonov, line 99: I would like to comment the code above.
```
            if (oldNode.id().equals(nodeToPut.id())) {
                // We already have this node, nothing needs to be changed.
                return;
            }
```
How is it possible?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/api/LogicalNode.java
```diff
@@ -79,10 +78,9 @@ public Map<String, String> nodeAttributes() {
...
-    @Override
-    public String toString() {
-        // TODO: S.toString for inherited classes do not work properly https://issues.apache.org/jira/browse/IGNITE-19183
-        return S.toString(LogicalNode.class, this);
-    }
+    // TODO: S.toString for inherited classes do not work properly https://issues.apache.org/jira/browse/IGNITE-19183
+    //    @Override
+    //    public String toString() {
+    //        return S.toString(LogicalNode.class, this);
+    //    }
```
ibessonov, line 85: What was the reason for commenting this code? Maybe you forgot to uncomment it, I don't know

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/ActionRequest.java
```diff
@@ -35,7 +34,6 @@ public interface ActionRequest extends Message {
     /**
      * @return Action's command.
      */
-    @Marshallable
```
ibessonov, line 38: Thank you very much! :+1:

ibessonov (approved):

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/LogicalTopologyImpl.java
sashapolo: It is possible because a node may try to re-join (e.g. after CMG leader change)

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/api/LogicalNode.java
sashapolo: Please have a look at the ticket in the TODO above. Current implementation simply prints `LogicalNode {}` hiding all important information, so I'd like to use `toString` from `ClusterNode` instead.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/LogicalTopologyImpl.java
sashapolo: This is to make this Raft command idempotent

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/api/LogicalNode.java
ibessonov: This is so weird. In Ignite 2, we constantly used `, "super", super.toString()`, right? Why don't we do the same thing? That's a one line fix

sashapolo: I didn't know what we used in Ignite 2 =(

PR #1902, opened 2023-04-04
IGNITE-18952 Add the filter as a configuration parameter of a Distribution zone

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-18952

### DIFF
--- a/modules/distribution-zones/build.gradle
+++ b/modules/distribution-zones/build.gradle
@@ -37,6 +37,11 @@ dependencies {
     implementation libs.jetbrains.annotations
     implementation libs.fastutil.core
     implementation libs.auto.service.annotations
+    implementation(libs.jsonpath.core) {
+        //IDEA test runner doesn't apply Gradle dependency resolve strategy, this is just not implemented
+        //So, exclude asm-core transitive dependency to protect of jar-hell.
+        exclude group: 'org.ow2.asm', module: 'asm'
+    }
 
     testImplementation libs.mockito.core
     testImplementation libs.mockito.junit
--- a/modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZonesUtil.java
+++ b/modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZonesUtil.java
@@ -29,6 +29,8 @@ import static org.apache.ignite.internal.metastorage.dsl.Operations.remove;
 import static org.apache.ignite.internal.util.ByteUtils.bytesToLong;
 import static org.apache.ignite.internal.util.ByteUtils.fromBytes;
 
+import com.jayway.jsonpath.InvalidPathException;
+import com.jayway.jsonpath.JsonPath;
 import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.Map;
@@ -346,5 +348,21 @@ public class DistributionZonesUtil {
         }
 
         throw new DistributionZoneNotFoundException(zoneId);
+    }
+
+    /**
+     * Check if a passed filter is a valid {@link JsonPath} query.
+     *
+     * @param filter Filter.
+     * @return {@code true} if the passed filter is a valid filter, {@code false} otherwise.
+     */
+    public static boolean validate(String filter) {
+        try {
+            JsonPath.compile(filter);
+        } catch (InvalidPathException ignored) {
+            return false;
+        }
+
+        return true;
     }
 }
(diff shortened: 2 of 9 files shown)


modules/distribution-zones/build.gradle
```diff
@@ -37,6 +37,11 @@ dependencies {
     implementation libs.jetbrains.annotations
     implementation libs.fastutil.core
     implementation libs.auto.service.annotations
+    implementation(libs.jsonpath.core) {
+        //IDEA test runner doesn't apply Gradle dependency resolve strategy, this is just not implemented
```
kgusakov, line 41: Not sure, that I understand how the first line connected with the second, could you explain?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/configuration/DistributionZoneConfigurationSchema.java
```diff
@@ -67,4 +67,8 @@ public class DistributionZoneConfigurationSchema {
     @Range(min = 0)
     @Value(hasDefault = true)
     public int dataNodesAutoAdjustScaleDown = INFINITE_TIMER_VALUE;
+
+    @ValidFilter
+    @Value(hasDefault = true)
+    public String filter = "$..*";
```
sanpwc, line 73: What "$..*" means? Any String? Let's add a comment with an explanation.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZonesUtil.java
```diff
@@ -347,4 +349,20 @@ public static DistributionZoneConfiguration getZoneById(DistributionZonesConfigu
...
+    /**
+     * Check if a passed filter is a valid {@link JsonPath} query.
+     *
+     * @param filter Filter.
+     * @return {@code true} if the passed filter is a valid filter, {@code false} otherwise.
+     */
+    public static boolean validate(String filter) {
+        try {
+            JsonPath.compile(filter);
+        } catch (InvalidPathException ignored) {
```
sanpwc, line 362: What about validation details? It's useful to propagate validation details to the user.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/configuration/DistributionZoneConfigurationSchema.java
sanpwc, line 73: And what if user will try to change the filter from some to "anything"? Do you expect them to specify "$..*"?

modules/distribution-zones/build.gradle
alievmirza: This is a copy-paste from the other module where JsonPath was introduced, I think it's important to exclude this libs here as well

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/configuration/DistributionZoneConfigurationSchema.java
alievmirza: This is a  com.jayway.jsonpath.JsonPath expression for including all attributes of nodes.

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerTest.java
```diff
@@ -755,17 +740,118 @@ public void testTryDropDefaultZone() {
...
+        String notValidFilter = "['nodeAttributes[?(@.['region'] == 'EU')]";
+
+        assertThrowsWithCause(
+                () -> distributionZoneManager.alterZone(
+                        ZONE_NAME,
+                        new DistributionZoneConfigurationParameters.Builder(ZONE_NAME)
+                                .filter(notValidFilter).build()
+                        )
+                        .get(5, TimeUnit.SECONDS),
+                ConfigurationValidationException.class
```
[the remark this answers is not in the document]
alievmirza: done

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZonesUtil.java
alievmirza: added

sanpwc (approved):

modules/distribution-zones/build.gradle
kgusakov: I see, thanks!

PR #1905, opened 2023-04-05
IGNITE-19068 Sql. Support sql engine representation for kv api timestamp type

zstan:


zstan:
@korlov42 can u make a review plz ?

zstan:
@AMashenkov can you make a review plz ?

korlov42 (changes requested):
The patch looks good, but, apart tests on interoperation, it worth to add tests to verify behaviour of the type itself (see org.apache.ignite.internal.sql.engine.ItDataTypesTest)

zstan:
@korlov42  i append additional integration sqllogic tests, as i understand soon they become unmuted and thus it`s appropriate change, isn`t it ?

AMashenkov (approved):

PR #1915, opened 2023-04-06
IGNITE-19187 Sql. Handle StorageRebalanceException during rowsCount estimation.

xtern:
❗The test is not included in this PR for the following reasons:

- it is complex and slow.
- it is unstably reproduces current issue (~90% on local machine).
- it is difficult to maintain because it contains a lot of copy-paste code that may be changed in the nearest future.
- it uses components that currently have a LOT of issues. Take a look at TODOs in integration tests with rebalancing (ItTableRaftSnapshotsTest and ItRebalanceTest): IGNITE-18848, IGNITE-18423, IGNITE-18733, IGNITE-18481, IGNITE-18485, IGNITE-18692.

This test can be found in a separate PR: https://github.com/apache/ignite-3/pull/1919/files


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/IgniteTableImpl.java
```diff
@@ -539,20 +540,33 @@ private class StatisticsImpl implements Statistic {
...
+
+                    try {
                         size += part.rowsCount();
+                    } catch (StorageRebalanceException ignore) {
+                        noCache = true;
                     }
                 }
 
-                localRowCnt = size;
+                if (noCache) {
```
ygerzhedovich, line 565: in case the first calculation is fallen we will use zero as the rows counter, I'm not sure but maybe will be better to use incorrectly calculated value for the case

xtern: Agree, fixed a bit.

xtern: I decided to revert this change completely... currently any concurrent call can see a zero in localRowCnt (if it is calculated for the first time), but may be it's ok.

korlov42 (approved):

PR #1917, opened 2023-04-06
IGNITE-19116 Sql. Fixed UPDATE statement fails with NPE when table doesn't exist.

xtern:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -210,10 +212,12 @@ protected SqlSelect createSourceSelectForDelete(SqlDelete call) {
         final SqlNodeList selectList = new SqlNodeList(SqlParserPos.ZERO);
         final SqlValidatorTable table = getCatalogReader().getTable(((SqlIdentifier) call.getTargetTable()).names);
 
-        table.unwrap(IgniteTable.class).descriptor().deleteRowType((IgniteTypeFactory) typeFactory)
-                .getFieldNames().stream()
-                .map(name -> new SqlIdentifier(name, SqlParserPos.ZERO))
-                .forEach(selectList::add);
+        if (table != null) {
```
korlov42, line 215: it's unclear why the table may be null at this point. Besides, you're returning incorrect result here

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -169,6 +169,10 @@ protected SqlSelect createSourceSelectForUpdate(SqlUpdate call) {
         final SqlIdentifier targetTable = (SqlIdentifier) call.getTargetTable();
         final SqlValidatorTable table = getCatalogReader().getTable(targetTable.names);
 
+        if (table == null) {
+            throw newValidationError(call.getTargetTable(), RESOURCE.objectNotFound(targetTable.toString()));
```
zstan, line 173: more clear to use : RESOURCE.tableNotFound ? I see that it throws not SqlValidatorException and additional fix possibly required, wdyt ? Also - is it correct to throw not wrapped calcite exception in public api, probably we need to mention some already defined issue here ?

xtern: > I see that it throws not SqlValidatorException and additional fix possibly required,
Thanks, but I'm not fully understand your suggestion.

For example for select * from unknown calcite throws CalciteContextException caused by.SqlValidatorException.
And the same behavior here:
```
CalciteContextException: From line 1, column 8 to line 1, column 14: Object 'UNKNOWN' not found
...
Caused by: org.apache.calcite.sql.validate.SqlValidatorException: Object 'UNKNOWN' not found
```

xtern: > more clear to use : RESOURCE.tableNotFound
Thanks, but I'm not sure about it.

First, as you mentioned it's not a validator exception.
We can use _RESOURCE.tableNameNotFound_ instead, but still not sure about such a change in this task.

Because we will have inconsistent behavior for MERGE queries for example:
```
MERGE INTO non_existing_table USING existing_table
```
will produce `Table 'NON_EXISTING_TABLE' not found`, but 
```
MERGE INTO existing_table USING non_existing_table
```
will produce `Object 'NON_EXISTING_TABLE' not found`

> probably we need to mention some already defined issue here
Added TODO https://issues.apache.org/jira/browse/IGNITE-14865

korlov42 (approved):

PR #1918, opened 2023-04-06
IGNITE-19073 Add authentication to thin client protocol

Flaugh24:
* Pass credentials in client handshake extensions, authenticate when required.
* Disconnect clients when authentication configuration changes.

https://issues.apache.org/jira/browse/IGNITE-19073


modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/Extension.java
```diff
@@ -0,0 +1,57 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.client.proto;
+
+/** Extensions. */
```
ptupitsyn, line 20:
```suggestion
/**
 * Handshake extensions.
 */
```

modules/client-handler/src/integrationTest/java/org/apache/ignite/client/handler/TestServer.java
```diff
@@ -49,12 +52,15 @@ public class TestServer {
...
     private final TestSslConfig testSslConfig;
 
+    private final AuthenticationConfiguration authenticationConfiguration;
+
     private final ClientHandlerMetricSource metrics = new ClientHandlerMetricSource();
 
     private long idleTimeout = 5000;
 
-    TestServer(@Nullable TestSslConfig testSslConfig) {
+    TestServer(@Nullable TestSslConfig testSslConfig, AuthenticationConfiguration authenticationConfiguration) {
```
ptupitsyn, line 61: Can we make `authenticationConfiguration` nullable? It is not convenient to inject it in all tests where we don't actually need it.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
```diff
@@ -143,12 +152,17 @@ public class ClientInboundMessageHandler extends ChannelInboundHandlerAdapter {
     /** Context. */
     private ClientContext clientContext;
 
+    private ChannelHandlerContext channelHandlerContext;
```
ptupitsyn, line 155: All fields have javadocs in this file. `authenticationManager` added below also has it. Let's be consistent and add javadoc here too.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
```diff
@@ -608,4 +637,46 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
 
         ctx.close();
     }
+
+    @Override
+    public CompletableFuture<?> onUpdate(ConfigurationNotificationEvent<AuthenticationView> ctx) {
+        return CompletableFuture.runAsync(() -> {
+            if (clientContext != null && channelHandlerContext != null) {
+                onAuthenticationChange();
```
ptupitsyn, line 645: Inline this single-line method with a single call site?

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
```diff
@@ -608,4 +637,46 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
 
         ctx.close();
     }
+
+    @Override
+    public CompletableFuture<?> onUpdate(ConfigurationNotificationEvent<AuthenticationView> ctx) {
+        return CompletableFuture.runAsync(() -> {
```
ptupitsyn, line 643: Is there a reason to use `runAsync`? Looks like we can run the code synchroinously.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
```diff
@@ -608,4 +637,46 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
...
+        } else if (type == Integer.class) {
+            return unpacker.unpackInt();
+        } else if (type == Long.class) {
+            return unpacker.unpackLong();
+        } else if (type == Boolean.class) {
+            return unpacker.unpackBoolean();
+        } else if (type == Double.class) {
+            return unpacker.unpackDouble();
+        } else if (type == Float.class) {
+            return unpacker.unpackFloat();
```
ptupitsyn, line 677: We only use strings right now. Let's remove unused code.

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/Extension.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param key Key.
+     * @return Extension.
+     */
+    public static Extension fromKey(String key) {
+        for (Extension e : values()) {
+            if (e.key.equalsIgnoreCase(key)) {
+                return e;
+            }
+        }
+        throw new IllegalArgumentException("Unknown extension key: " + key);
```
ptupitsyn, line 47: The point of handshake extensions is to be flexible and optional. We should skip unknown extensions instead of crashing.

ptupitsyn (changes requested):

modules/client-handler/src/integrationTest/java/org/apache/ignite/client/handler/ItClientHandlerTest.java
```diff
@@ -193,4 +325,17 @@ private void writeAndFlushLoop(Socket socket) throws Exception {
...
+            change.changeEnabled(true);
+            change.changeProviders().create("basic", authenticationProviderChange -> {
+                authenticationProviderChange.convert(BasicAuthenticationProviderChange.class)
+                        .changeLogin(username)
+                        .changePassword(password)
+                        .changeName("basic");
+            });
+        }).join();
+    }
+
```
ptupitsyn, line 340:
```suggestion
```

ptupitsyn (approved):

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/Extension.java
isapego: It's probably more explicit and thus more obvious that extension can be unknown and we need to check for this result rather then checking for null, but I'm not insisting as you for sure have more expertise in Java than I.

isapego (approved):

PR #1926, opened 2023-04-10
ignite-19264 Simplify C++ Binary Tuple code

ademakov:


modules/platforms/cpp/ignite/common/uuid.h
```diff
@@ -109,7 +109,7 @@ class uuid {
         return 0;
     }
 
-public:
+private:
```
isapego, line 112: Wow, nice catch :D

isapego (approved):

PR #1927, opened 2023-04-11
IGNITE-18899 Refactor ClientTuple to wrap BinaryTuple

ptupitsyn:
Similar to `MutableRowTupleAdapter`, refactor client-related tuples:
* `ClientSqlRow`.
* `ClientTuple` on client.
* `ClientTuple` on server (new class, used instead of `TupleImpl`).

All those implementations are based on the same `MutableTupleBinaryTupleAdapter`.

As a result:
* One underlying tuple implementation; less code; simpler code in general.
* Reduced allocations, improved performance.

Additionally:
* Move iterator implementation to default implementation in `Tuple` interface.
* Move `BinaryTupleContainer` to `binarytuple` module. 
* Fix inconsistent exceptions when tuple column does not exist.
* Fix `Tuple.hashCode` for byte arrays.

```
Benchmark                                                Mode  Cnt      Score       Error   Units
ClientPutGetBenchmark.get                               thrpt    3  49418.080 ± 17639.810   ops/s (before IGNITE-18899)
ClientPutGetBenchmark.get                               thrpt    3  50305.929 ± 13924.407   ops/s (after IGNITE-18899)
ClientPutGetBenchmark.get:·gc.alloc.rate                thrpt    3    666.616 ±  1242.450  MB/sec (before IGNITE-18899)
ClientPutGetBenchmark.get:·gc.alloc.rate                thrpt    3    480.666 ±   829.508  MB/sec (after IGNITE-18899)
ClientPutGetBenchmark.get:·gc.alloc.rate.norm           thrpt    3  15572.822 ± 25042.204    B/op (before IGNITE-18899)
ClientPutGetBenchmark.get:·gc.alloc.rate.norm           thrpt    3  11046.571 ± 21170.502    B/op (after IGNITE-18899)
ClientPutGetBenchmark.get:·gc.churn.G1_Eden_Space       thrpt    3    716.282 ±   529.130  MB/sec (before IGNITE-18899)
ClientPutGetBenchmark.get:·gc.churn.G1_Eden_Space       thrpt    3    519.416 ±   555.076  MB/sec (after IGNITE-18899)
ClientPutGetBenchmark.get:·gc.churn.G1_Eden_Space.norm  thrpt    3  16757.093 ± 17212.675    B/op (before IGNITE-18899)
ClientPutGetBenchmark.get:·gc.churn.G1_Eden_Space.norm  thrpt    3  11928.927 ± 13109.114    B/op (after IGNITE-18899)
ClientPutGetBenchmark.get:·gc.churn.G1_Old_Gen          thrpt    3      0.005 ±     0.046  MB/sec (before IGNITE-18899)
ClientPutGetBenchmark.get:·gc.churn.G1_Old_Gen          thrpt    3      0.001 ±     0.001  MB/sec (after IGNITE-18899)
ClientPutGetBenchmark.get:·gc.churn.G1_Old_Gen.norm     thrpt    3      0.113 ±     1.101    B/op (before IGNITE-18899)
ClientPutGetBenchmark.get:·gc.churn.G1_Old_Gen.norm     thrpt    3      0.025 ±     0.006    B/op (after IGNITE-18899)
ClientPutGetBenchmark.get:·gc.count                     thrpt    3     40.000              counts (before IGNITE-18899)
ClientPutGetBenchmark.get:·gc.count                     thrpt    3     29.000              counts (after IGNITE-18899)
ClientPutGetBenchmark.get:·gc.time                      thrpt    3     15.000                  ms (before IGNITE-18899)
ClientPutGetBenchmark.get:·gc.time                      thrpt    3     12.000                  ms (after IGNITE-18899)
```


modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientColumnTypeConverter.java
```diff
@@ -24,12 +24,12 @@
...
 public class ClientColumnTypeConverter {
     /**
-     * Converts column type to wire code.
+     * Converts SQL column type to wire code.
      *
      * @param columnType Column type.
      * @return Wire code.
      */
-    public static int columnTypeToOrdinal(ColumnType columnType) {
+    public static int sqlColumnTypeToOrdinal(ColumnType columnType) {
```
isapego, line 32: Why not `ColumnType.ordinal()`? Is this because we want to make sure ordinals do not change in future?

modules/client/src/test/java/org/apache/ignite/client/ClientTupleTest.java
```diff
@@ -353,13 +270,36 @@ public void testTupleEqualityCompatibility() {
         assertEquals(clientTuple.hashCode(), tuple.hashCode());
     }
 
-    private static ClientTuple getBuilder() {
-        return new ClientTuple(SCHEMA);
+    private static Tuple getTuple() {
```
AMashenkov, line 273:
```suggestion
    private static Tuple createTuple() {
```

modules/client/src/test/java/org/apache/ignite/client/ClientTupleTest.java
```diff
@@ -353,13 +270,36 @@ public void testTupleEqualityCompatibility() {
...
+        var binTuple = new BinaryTupleReader(SCHEMA.columns().length, binTupleBuf);
+
+        return new ClientTuple(SCHEMA, binTuple, 0, SCHEMA.columns().length);
     }
 
-    private static Tuple getTuple() {
-        return new ClientTuple(SCHEMA)
-                .set("id", 3L)
-                .set("name", "Shirt");
+    private static ClientTuple getFullSchemaTuple() {
```
AMashenkov, line 284:
```suggestion
    private static ClientTuple createFullSchemaTuple() {
```

isapego (approved):

modules/client/src/test/java/org/apache/ignite/client/ClientTupleTest.java
ptupitsyn: Fixed.

modules/client/src/test/java/org/apache/ignite/client/ClientTupleTest.java
ptupitsyn: Fixed.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/BinaryTupleParser.java
```diff
@@ -298,15 +303,16 @@ public final float floatValue(int begin, int end) {
...
+        switch (len) {
             case 0:
                 return 0.0;
             case Float.BYTES:
                 return buffer.getFloat(begin);
             case Double.BYTES:
                 return buffer.getDouble(begin);
             default:
-                throw new BinaryTupleFormatException("Invalid length for a tuple element");
+                throw new BinaryTupleFormatException("Invalid length for a tuple element: " + len);
```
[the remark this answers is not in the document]
ptupitsyn: This exception should not happen, it indicates a bug in Ignite (when we have corrupted schema or corrupted data). So yes, stack trace should be enough to know the type, but seeing actual length may be helpful for us to debug the issue.

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientColumnTypeConverter.java
ptupitsyn: Yes, those are client protocol ordinals. But this code is not ideal, we have too many different type codes. I think we even discussed the idea of replacing all type codes with `ColumnType.ordinal` some time ago.

ptupitsyn: Ticket filed to address type code madness: https://issues.apache.org/jira/browse/IGNITE-19270

PR #1929, opened 2023-04-11
IGNITE-19152 Use schema information in LocalFileConfigurationStorage

PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-19152


modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -40,114 +53,187 @@
...
+                latest.remove(key);
+            } else {
+                latest.put(key, value);
+            }
+        });
     }
 
     private void runAsync(Runnable runnable) {
         CompletableFuture<Void> future = CompletableFuture.runAsync(runnable, threadPool);
-
```
ibessonov, line 150: Why did you remove this empty line? What's the reason?
For longest time we have specific code convention, we use empty lines throughout the code, it gives Ignite's code certain look and feel. I would say that you don't follow it enough, but that's ok.
What's not ok to me is simply removing such formatting from the code, that has nothing to do with your changes.

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -175,47 +261,66 @@ public CompletableFuture<Void> writeConfigurationRevision(long prevRevision, lon
...
-            if (value.getClass().isArray()) {
-                return Arrays.asList((Object[]) value);
-            }
-            return value;
-        }));
-        Config other = ConfigFactory.parseMap(map);
-        Config newConfig = other.withFallback(parseConfigOptions()).resolve();
+    private String renderHoconString() {
+        // Super root that'll be filled from the storage data.
+        SuperRoot rootNode = new SuperRoot(rootCreator());
```
ibessonov, line 296: I think you should call `createSuperRoot()` instead. Right now this code may not work. Have you tested it?

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -175,47 +261,66 @@ public CompletableFuture<Void> writeConfigurationRevision(long prevRevision, lon
...
+
+        fillFromPrefixMap(rootNode, toPrefixMap(latest));
+
+        addDefaults(rootNode);
```
ibessonov, line 300: Why did you add defaults here? I guess it's a copied code, right? And you didn't tests whether you write default values or not.

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -175,47 +261,66 @@ public CompletableFuture<Void> writeConfigurationRevision(long prevRevision, lon
...
+
+    private InnerNode createRootNode(RootKey<?, ?> rootKey) {
+        return cgen.instantiateNode(rootKey.schemaClass());
+    }
+
     private Config parseConfigOptions() {
-        return ConfigFactory.parseFile(
-                configPath.toFile(),
-                ConfigParseOptions.defaults().setAllowMissing(false));
+        return readHoconFromFile();
```
ibessonov, line 323: Please inline this method

modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorageTest.java
```diff
@@ -17,75 +17,401 @@
...
-        // \n instead of System.lineSeparator because Config library writes \n only
-        assertThat(contents, is("foo1=bar1\n"
-                + "foo2=bar2\n"
-                + "map {\n"
-                + "    list=[\n"
-                + "        val1,\n"
-                + "        val2\n"
-                + "    ]\n"
-                + "}\n"));
+    @DisplayName("Default values are not added enriched on read when the config file is empty")
```
ibessonov, line 108: "not added enriched" - not added or not enriched?
Why is it a display name, is that convenient? You better give the method a proper name and move this description into a javadoc, because that's how most of our tests are written

modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorageTest.java
```diff
@@ -17,75 +17,401 @@
...
+    void empty() throws IOException {
+        // Given
+        assertThat(configFileContent(), emptyString());
+
+        // When
+        var storageValues = readAllLatest();
+
+        // Then storage data only contains top level defaults
+        assertThat(storageValues.entrySet(), hasSize(1));
```
ibessonov, line 117: Can be replaced with `assertThat(storageValues, is(aMapWithSize(1)))`, it would look better.
By the way, what's a `top level defaults`? Configuration is empty, and we shouldn't store defaults in current implementation (I guess).
Also, why don't you explicitly assert the content of the map in this test? It's easy to do

modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorageTest.java
```diff
@@ -17,75 +17,401 @@
...
 
     @Test
-    void testMergeHocon() throws IOException {
-        var data = Map.of("foo1", "bar");
-        assertThat(storage.write(data, 0), willBe(true));
+    @DisplayName("Named list entities can be added")
+    void add() throws Exception {
+        // Given
+        assertThat(configFileContent(), emptyString());
+        // And
```
ibessonov, line 125: Please add empty lines here and there, this code looks too cramped

modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorageTest.java
```diff
@@ -17,75 +17,401 @@
...
+            x.changeIntVal(-1);
+        })).get();
+
+        // When
+        var storageValues = readAllLatest();
 
-        var append = Map.of("foo1", "baz", "foo2", "bar");
-        assertThat(storage.write(append, 1), willBe(true));
+        // Then
+        assertThat(storageValues, allOf(aMapWithSize(6), hasValue(-1)));
```
ibessonov, line 136: Please add a comment, what are the other 4 entries in the map. There are asserts only for 2 of them, this is not trivial

modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorageTest.java
```diff
@@ -17,75 +17,401 @@
...
+        assertThat(storageValues, allOf(aMapWithSize(6), hasValue("strVal1")));
+        // And
+        assertThat(configFileContent(), equalToCompressingWhiteSpace(
+                "top {\n"
+                        + "    namedList=[\n"
+                        + "        {\n"
+                        + "            intVal=-1\n"
+                        + "            name=name1\n"
+                        + "            strVal=strVal1\n"
+                        + "        }\n"
```
ibessonov, line 146: Can we have a TODO that would tell us to fix the rendering syntax for named lists?
Also, I would expect (in the future) the following (when there's only a single property to the list entry):
```
namedList {
    foo.attribute = a
    bar.attribute = b
}
```
instead of
```
namedList {
    foo {
        attribute = a
    }
    ...
}
```
Maybe we would require a custom rendering code. This topic must be discussed with somebody. Syntax matters, it's a user-facing part of Ignite.

modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorageTest.java
```diff
@@ -17,75 +17,401 @@
...
+
+        // When
+        var storageValues = storage.readDataOnRecovery().join().values();
+
+        // Then storage data only contains top level defaults
+        assertThat(storageValues.entrySet(), hasSize(1));
+        // And
+        assertThat(configFileContent(), equalToCompressingWhiteSpace(
+                "top {\n"
+                        + "    namedList=[]\n"
```
ibessonov, line 343: Can we avoid rendering empty lists?

modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorageTest.java
```diff
@@ -17,75 +17,401 @@
...
+    }
+
+    private Map<String, ? extends Serializable> readAllLatest() {
+        return storage.readAllLatest("").join();
+    }
+
+    // null == remove
+
+    // when read file you can see all values
+    //
```
ibessonov, line 396: What's this?

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -40,114 +53,187 @@
...
-    private final ExecutorService threadPool = Executors.newFixedThreadPool(2, new NamedThreadFactory("loc-cfg-file", LOG));
+    /** Thread pool for configuration updates notifications. */
+    private final ExecutorService threadPool = Executors.newFixedThreadPool(
+            2, new NamedThreadFactory("loc-cfg-file", LOG)
+    );
+
+    /** Thread pool for configuration updates notifications. */
+    private final ExecutorService workerThreadPool = Executors.newFixedThreadPool(
+            2, new NamedThreadFactory("cfg-file-worker", LOG)
+    );
```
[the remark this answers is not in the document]
PakhomovAlexander: Given an implementation of the configuration listener that freezes the thread, we won't stop processing the configuration updates. So, I decided to separate the notification and main executors. WDYT?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
```diff
@@ -606,7 +606,9 @@ private class Node {
                             RestConfiguration.KEY,
                             ClientConnectorConfiguration.KEY),
                     Set.of(),
-                    new LocalFileConfigurationStorage(configPath),
+                    new LocalFileConfigurationStorage(configPath, List.of(NetworkConfiguration.KEY,
```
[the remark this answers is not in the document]
PakhomovAlexander: Good point, I've just applied the same approach as we do in `ConfigurationManager`. Do you think we need to use `ConfigurationModule` there as well?

modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorageTest.java
PakhomovAlexander: The question with defaults is a good one. I think you are right and we should not store defaults.

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
PakhomovAlexander: Sure I did.

modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorageTest.java
PakhomovAlexander: https://issues.apache.org/jira/browse/IGNITE-19303

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationRegistry.java
```diff
@@ -96,8 +83,11 @@ public class ConfigurationRegistry implements IgniteComponent, ConfigurationStor
     /** Configuration change handler. */
     private final ConfigurationChanger changer;
 
-    /** Configuration generator. */
-    private final ConfigurationAsmGenerator cgen = new ConfigurationAsmGenerator();
+    /** Runtime implementations generator for node classes. */
+    private final ConfigurationTreeGenerator generator;
+
+    /** Flag that indicates if the {@link ConfigurationTreeGenerator} instance is owned by this object or not. */
+    private boolean ownConfigTreeGenerator = false;
```
ibessonov, line 90: Why does it matter?
I believe that generator can always be a constructor parameter, and another component, that instantiated it, should also close it in the end

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationTreeGenerator.java
```diff
@@ -0,0 +1,281 @@
...
+    private ConfigurationAsmGenerator generator = new ConfigurationAsmGenerator();
+
+    /**
+     * Constructor that takes a collection of root keys. Internal and polymorphic schema extensions are empty by default.
+     *
+     * @param rootKeys Root keys.
+     */
+    public ConfigurationTreeGenerator(Collection<RootKey<?, ?>> rootKeys) {
+        this(rootKeys, Set.of(), Set.of());
+    }
```
ibessonov, line 69: Is this for tests? Mark it as `@TestOnly` then

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/util/ConfigurationFlattener.java
```diff
@@ -136,6 +136,10 @@ public Void doVisitInnerNode(String key, InnerNode newNode) {
                 return null;
             }
 
+            if (oldNode == null && newNode == null) {
```
ibessonov, line 139: Can you please add a comment, explaining how it's possible?

modules/configuration/src/test/java/org/apache/ignite/internal/configuration/ConfigurationChangerTest.java
```diff
@@ -113,13 +112,13 @@ public static class ThirdConfigurationSchema {
         public String strCfg;
     }
 
-    private static ConfigurationAsmGenerator cgen = new ConfigurationAsmGenerator();
+    private static ConfigurationTreeGenerator generator = new ConfigurationTreeGenerator(List.of(KEY, DefaultsConfiguration.KEY));
```
ibessonov, line 115: Maybe this test-only constructor should use vararg, for convenience

modules/configuration/src/test/java/org/apache/ignite/internal/configuration/ConfigurationChangerTest.java
```diff
@@ -113,13 +112,13 @@ public static class ThirdConfigurationSchema {
...
 
     private final TestConfigurationStorage storage = new TestConfigurationStorage(LOCAL);
 
     @AfterAll
-    public static void afterAll() {
-        cgen = null;
+    public static void afterAll() throws Exception {
```
ibessonov, line 120: Does it have to throw an exception? This "close" method in generator simply nullifies the field, there's no way it ends with exception.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/ItSslConfigurationValidationTest.java
```diff
@@ -34,6 +35,7 @@
  * Integration test for checking SSL configuration validation.
  */
 @ExtendWith(WorkDirectoryExtension.class)
+@Disabled("https://issues.apache.org/jira/browse/IGNITE-19315")
```
ibessonov, line 38: Too bad, I expected us to preserve this function.
Do we prioritize this new feature as the next one that you'll be working on?

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -56,99 +66,146 @@
...
+        this.generator = generator;
+        this.tempConfigPath = configPath.resolveSibling(configPath.getFileName() + ".tmp");
+
         checkAndRestoreConfigFile();
     }
 
     @Override
     public CompletableFuture<Data> readDataOnRecovery() {
-        return CompletableFuture.completedFuture(new Data(Collections.emptyMap(), 0));
+        return async(() -> {
```
ibessonov, line 119: What's the deal with doing it asynchronously? It's a node start routine, we can do it in current thread I guess?
Please comment

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -56,99 +66,146 @@
...
+            try {
+                SuperRoot superRoot = generator.createSuperRoot();
+                SuperRoot copiedSuperRoot = superRoot.copy();
+
+                Config hocon = readHoconFromFile();
+                HoconConverter.hoconSource(hocon.root()).descend(copiedSuperRoot);
+
+                Map<String, Serializable> flattenedUpdatesMap = createFlattenedUpdatesMap(superRoot, copiedSuperRoot);
+                flattenedUpdatesMap.forEach((key, value) -> {
+                    if (value != null) { // filter defaults
```
ibessonov, line 130:
```suggestion
                    if (value != null) { // Filter defaults.
```

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -56,99 +66,146 @@
...
-            checkAndRestoreConfigFile();
-            Map<String, Serializable> map = latest.entrySet()
-                    .stream()
-                    .filter(entry -> entry.getKey().startsWith(prefix))
-                    .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
-            return CompletableFuture.completedFuture(map);
-        } finally {
-            lock.readLock().unlock();
-        }
+        return async(() -> {
```
ibessonov, line 150: Again, I don't get it. We read data from the field. Why do we have to do it in a separate pool? Makes no sense to me, please add a comment or revert this change.

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -56,99 +66,146 @@
...
     @Override
     public CompletableFuture<Serializable> readLatest(String key) {
-        lock.readLock().lock();
-        try {
-            checkAndRestoreConfigFile();
-            return CompletableFuture.completedFuture(latest.get(key));
-        } finally {
-            lock.readLock().unlock();
-        }
+        return async(() -> {
```
ibessonov, line 165: Same here. Literally "Map#get"

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -56,99 +66,146 @@
...
-        }
+        });
     }
 
-    private void runAsync(Runnable runnable) {
-        CompletableFuture<Void> future = CompletableFuture.runAsync(runnable, threadPool);
+    private void mergeAndSave(Map<String, ? extends Serializable> newValues) {
+        updateLatestState(newValues);
+        saveConfigFile();
+        lastRevision++;
```
ibessonov, line 198: New revision has been given to you as a parameter of "write" method. Why do you ignore it and increment the old revision instead? Please don't do that.

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -56,99 +66,146 @@
...
+    }
 
-        futureTracker.registerFuture(future);
+    private void updateLatestState(Map<String, ? extends Serializable> newValues) {
+        newValues.forEach((key, value) -> {
+            if (value == null) { // null means that we should remove this entry
```
ibessonov, line 203:
```suggestion
            if (value == null) { // Null means that we should remove this entry.
```

ibessonov: Please don't ignore coding conventions :)

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -175,56 +232,65 @@ public CompletableFuture<Void> writeConfigurationRevision(long prevRevision, lon
...
+        // Super root that'll be filled from the storage data.
+        SuperRoot rootNode = generator.createSuperRoot();
+
+        fillFromPrefixMap(rootNode, toPrefixMap(latest));
+
+        Object transformed = rootNode.accept(null, new ConverterToMapVisitor(false, true));
+
+        ConfigValue conf = ConfigImpl.fromAnyRef(
+                transformed, null
+        );
```
ibessonov, line 275: Could be a single line I guess

modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorageTest.java
ibessonov: Just in case - I don't mind custom HOCON renderer. This format is kind-of trivial, and converting tree to string would only require a few days of coding, while there are multiple advantages, like the ability to inset comments or to preserve properties order.

modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorageTest.java
```diff
@@ -17,75 +17,457 @@
...
 
-    private Path getConfigFile() {
-        return tmpDir.resolve(CONFIG_NAME);
+    /** Read configuration when inner node configured with partial content (some fields are empty). */
+    @Test
+    void innerNodeWithPartialContent() throws Exception {
+        String content = "top: { inner.boolVal: true }";
+        Files.write(getConfigFile(), content.getBytes(StandardCharsets.UTF_8));
+
+        storage.readDataOnRecovery().get();
```
ibessonov, line 419: Should we assert the size of the map?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/ItSslConfigurationValidationTest.java
PakhomovAlexander: Sure, it has "Critical" priority. Will work on it asap.

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationManager.java
```diff
@@ -40,6 +31,30 @@ public class ConfigurationManager implements IgniteComponent {
...
+     * Constructor.
+     *
+     * @param rootKeys                    Configuration root keys.
+     * @param validators                  Validators.
+     * @param storage                     Configuration storage.
+     * @param generator                   Configuration tree generator.
+     * @throws IllegalArgumentException If the configuration type of the root keys is not equal to the storage type, or if the schema or its
+     *                                  extensions are not valid.
+     */
+    public ConfigurationManager(
```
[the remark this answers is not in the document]
PakhomovAlexander: Yes, there is a reason for that. `ConfigurationManager` now gives two constructors like:

1) You can provide `ConfigurationTreeGenerator` with `LocalConfigurationStorage` (that is dependent on `ConfigurationTreeGenerator`)
2) You can provide lists of roots and extensions with `DistributedConfgurationStorage`(that is NOT dependent on `ConfigurationTreeGenerator`) 

Does it make sense?

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
PakhomovAlexander: I am not a fan of declaring the interface as an async (through CompletableFutures) but running the code in the same thread where the future is returned. What is the reason to use futures then? 

But I can agree with you in this particular case.

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
PakhomovAlexander: Sorry, did not mentioned that it is your suggestion (though it is my code).

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
PakhomovAlexander: I don't get you.  Here is the part of the `write` method that verifies the version is the same. How it can be "new" revision?

```java
            if (ver != lastRevision) {
                return CompletableFuture.completedFuture(false);
            }
```

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationManager.java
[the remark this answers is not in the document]
ibessonov: Ok, I don't mind it, but it's always weird to see multiple constructors

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationRegistry.java
ibessonov: I don't think I understand the answer. Let me rephrase myself.
Why do you need a flag "ownConfigTreeGenerator"? It can be removed, right?
Second question - why do you need a "close" method in the generator? I don't see a reason to have it, other than writing `cgen.close()` instead of `cgen = null` in some tests. Can you please explain your decision?

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
ibessonov: Reason for using the future is that there are multiple implementations. Some of them cannot physically be synchronous, otherwise they would block thread for unknown period of time

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
ibessonov: Ok, thank you, this means that I'm stupid, there are no problems here.

ibessonov (changes requested):

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationRegistry.java
PakhomovAlexander: Technically you are right, we might write `cgen = null`. But this is something hidden. How developers might know that they should nullify the link because we don't want them to hold the link and produce memory leaks? 

I decided to move this contract into the place where contracts are defined – the interface. I think a clear declaration of `close` method helps developers to understand how to work with the object.

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
```diff
@@ -147,49 +139,40 @@ private Config readHoconFromFile() {
...
     public CompletableFuture<Serializable> readLatest(String key) {
-        return async(() -> {
-            lock.readLock().lock();
-            try {
-                return latest.get(key);
-            } finally {
-                lock.readLock().unlock();
-            }
-        });
+        return CompletableFuture.completedFuture(latest.get(key));
```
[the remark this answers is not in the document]
PakhomovAlexander: Thank you for the clarification, I misunderstood the `latest` meaning. My fault.

ibessonov (approved):

PR #1931, opened 2023-04-11
IGNITE-19021 Support the directory deployment

valepakh:
https://issues.apache.org/jira/browse/IGNITE-19021


PakhomovAlexander:
@valepakh I've launched the CLI in the project's root directory and run:

`unit deploy --path ./examples --version 1.0.0 my.unit`

Then I listed the work/deployment/my.unit/1.0.0 directory and I could not see the nested content of the `exampled` directory. I think we have to support it and it would be nice to add some tests for nested directories support.

modules/rest/src/main/java/org/apache/ignite/internal/rest/deployment/DeploymentManagementController.java
```diff
@@ -123,4 +100,11 @@ public static UnitStatus fromUnitStatus(org.apache.ignite.internal.deployunit.Un
         }
         return new UnitStatus(status.id(), versionToDeploymentStatus);
     }
+
+    private static Version parseVersion(String unitVersion) {
+        if (unitVersion == null || unitVersion.isBlank()) {
```
PakhomovAlexander, line 105: Why is this logic located in `DeployManagementCongroller`? I think this is `Version.parseVersion` responsibility.

PR #1933, opened 2023-04-12
IGNITE-19176 Sql. RW transaction skips rows without value.

xtern:
The original problem was that the RW transaction skipped rows without values, while the RO transaction returned such rows.

This behavior has been corrected - now both RO and RW transactions return rows without values.

Also changed behavior of binary key-value view without changing public API.
Previously binary view ignored rows without values.
E.g. was
```
binaryView.put(tx, key, Tuple.create()) // insert empty tuple (null value produces NPE)
binaryView.contains(tx, key) -> false
binaryView.get(tx, key) -> null
```

Now
```
binaryView.put(tx, key, Tuple.create())  // insert empty tuple (null value produces NPE)
binaryView.contains(tx, key) -> true
binaryView.get(tx, key) -> empty tuple
```


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDmlTest.java
```diff
@@ -594,4 +605,89 @@ public void testInsertMultipleDefaultsWithImplicitPk() {
                 .returns(5, 2)
                 .check();
     }
+
+    @ParameterizedTest
```
ygerzhedovich, line 609: I think the test should be in another place, for example together with antoher *View tests

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDmlTest.java
ygerzhedovich, line 1: Need to provide tests for all storages

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSingleColumnTableCrossApiTest.java
```diff
@@ -0,0 +1,269 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.TestInfo;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+/**
+ * Checks basic operations on a single column table.
+ */
+public class ItSingleColumnTableCrossApiTest extends ClusterPerClassIntegrationTest {
```
ygerzhedovich, line 49: I want to say that's about all columns belong to  PRIMARY KEY, single column just a special case

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSingleColumnTableCrossApiTest.java
```diff
@@ -0,0 +1,269 @@
...
+        tearDownBase(testInfo);
+    }
+
+    @BeforeAll
+    public void beforeAll() {
+        for (String engine : ENGINES) {
+            String testZoneName = ("test_zone_" + engine).toUpperCase();
+
+            sql(String.format("create zone %s engine %s with partitions=1, replicas=3;", testZoneName, engine));
+            sql(String.format("create table %s (ID int primary key) with primary_zone='%s'", tableName(engine), testZoneName));
```
ygerzhedovich, line 75: We can have tables with the same name for different zones???? Something is wrong...

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSingleColumnTableCrossApiTest.java
xtern: Should we cover multi-column PK with tests?

AMashenkov (approved):

PR #1935, opened 2023-04-13
IGNITE-19099 Java client: Add basic authentication support

ptupitsyn:
* Add another client handshake extension `authn-type` to future-proof for different providers.
* Rename existing handshake extensions to be more generic and match `AuthenticationRequest` on server.
* Add `AnonymousRequest` to handle the case when client does not provide in handshake.
* Add `IgniteClientAuthenticator` interface on the client side, with a single `BasicAuthenticator` implementation. 
* Update `IgniteClientConfiguration` and pass credentials in the handshake.
* Move `AuthenticationException` from internal package to public in `ignite-api` module, add missing constructor.


modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
```diff
@@ -321,9 +324,26 @@ private void handshake(ChannelHandlerContext ctx, ClientMessageUnpacker unpacker
...
+    }
+
+    private static AuthenticationRequest<?, ?> createAuthenticationRequest(Map<HandshakeExtension, Object> extensions) {
+        Object authnType = extensions.get(HandshakeExtension.AUTHENTICATION_TYPE);
+
+        if (authnType == null) {
+            return new AnonymousRequest();
+        }
+
+        if (authnType.equals("basic")) {
```
isapego, line 340: Probably, need to move `"basic"` to some constant.

modules/client/src/main/java/org/apache/ignite/client/IgniteClientAuthenticator.java
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.client;
+
+/**
+ * Ignite client authenticator.
+ */
```
isapego, line 22: Lets add that it is used by a client to get credentials during authentication on a server node.

modules/client/src/main/java/org/apache/ignite/client/BasicAuthenticator.java
```diff
@@ -0,0 +1,97 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.client;
+
+/**
+ * Basic authenticator.
+ */
```
isapego, line 22: Probably, should add what it means (single user, username + password transferred as a plain text). Probably, advice using it only for testing purposes or with TLS enabled?

isapego (approved):

modules/client/src/main/java/org/apache/ignite/client/BasicAuthenticator.java
ptupitsyn: Done.

isapego (approved):

PR #1942, opened 2023-04-14
IGNITE-19193 Store sensitive CLI config separately

Flaugh24:


modules/cli/src/main/java/org/apache/ignite/internal/cli/config/CliConfigKeys.java
```diff
@@ -53,6 +57,23 @@ public String value() {
         return value;
     }
 
+    /**
+     * Returns all secret config keys.
+     */
+    public static Set<String> secretConfigKeys() {
+        return Set.of(
+                        REST_KEY_STORE_PASSWORD,
+                        REST_KEY_STORE_PATH,
```
PakhomovAlexander, line 66: I don't think that the path is sensitive information. But we could discuss it.

modules/cli/src/test/java/org/apache/ignite/internal/cli/commands/CliCommandTestBase.java
```diff
@@ -102,9 +105,10 @@ protected void assertOutputIs(String expectedOutput) {
                 .isEqualTo(expectedOutput);
     }
 
-    protected void assertOutputContains(String expectedOutput) {
+    protected void assertOutputContains(String... expectedOutput) {
+        Set<String> expectedStrings = Arrays.stream(expectedOutput).collect(Collectors.toSet());
         assertThat(sout.toString())
-                .as("Expected command output to contain: " + expectedOutput + " but was " + sout.toString())
+                .as("Expected command output to contain: " + expectedStrings + " but was " + sout.toString())
```
PakhomovAlexander, line 111: What is this change for?

PakhomovAlexander (changes requested):
There is no single test for the new feature. Could you please write some tests?

PakhomovAlexander (approved):

PR #1943, opened 2023-04-14
IGNITE-19076 Deployment unit status support

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-19076


PakhomovAlexander (commented):
There are `ItDeploymentUnitTest`  and `ItDeployUndeployCallsTest`. Could you look at them and add at least one test case for deployment unit status?

PR #1948, opened 2023-04-17
IGNITE-19262 Sql. DDL silently ignores transaction

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-19262


ptupitsyn (approved):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlAsynchronousApiTest.java
```diff
@@ -597,6 +597,19 @@ public void errors() {
             await(ars.closeAsync());
             assertThrowsWithCause(() -> await(ars.fetchNextPage()), CursorClosedException.class);
         }
+
+        // DDL is non-transactional.
```
ygerzhedovich, line 601: I would like an additional test when we do a few operations in the transaction and just after it calls DDL in the same TX. Check that transaction is not rolled back or committed

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlAsynchronousApiTest.java
```diff
@@ -597,6 +597,32 @@ public void errors() {
...
+            try {
+                assertThrowsWithCause(() -> await(ses.executeAsync(tx, "CREATE TABLE TEST2(ID INT PRIMARY KEY, VAL0 INT)")),
+                        SqlException.class,
+                        "DDL doesn't support transactions."
+                );
+            } finally {
+                tx.rollback();
+            }
+        }
+        {
```
ygerzhedovich, line 613: it looks as a separate test :)

PR #1950, opened 2023-04-18
IGNITE-19100 .NET: Add basic authentication support

ptupitsyn:


modules/platforms/dotnet/Apache.Ignite.Tests/BasicAuthenticatorTests.cs
```diff
@@ -0,0 +1,118 @@
...
+    public async Task TestAuthnOnClientNoAuthnOnServer()
+    {
+        var cfg = new IgniteClientConfiguration(GetConfig())
+        {
+            Authenticator = new BasicAuthenticator
+            {
+                Username = "u",
+                Password = "p"
+            }
+        };
```
isapego, line 51: Why not?
```suggestion
        var cfg = new IgniteClientConfiguration(GetConfig(true));
```

modules/platforms/dotnet/Apache.Ignite/IgniteClientConfiguration.cs
```diff
@@ -169,5 +170,12 @@ public IgniteClientConfiguration(IgniteClientConfiguration other)
         /// See <see cref="SslStreamFactory"/>.
         /// </summary>
         public ISslStreamFactory? SslStreamFactory { get; set; }
+
+        /// <summary>
+        /// Gets or sets the authenticator.
```
isapego, line 175: Probably should add that if it's `null` then auth is not used.

modules/platforms/dotnet/Apache.Ignite/BasicAuthenticator.cs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /// <inheritdoc />
+    public string Type => HandshakeExtensions.AuthenticationTypeBasic;
+
+    /// <inheritdoc />
+    public object Identity => Username;
+
+    /// <inheritdoc />
+    public object Secret => Password;
```
isapego, line 61: Why these are `object` if we only support strings extensions?

isapego (approved):

modules/platforms/dotnet/Apache.Ignite.Tests/BasicAuthenticatorTests.cs
ptupitsyn: Because we pass incorrect credentials intentionally - to make sure that they are ignored when authentication is disabled on the server.

modules/platforms/dotnet/Apache.Ignite/BasicAuthenticator.cs
ptupitsyn: Server-side API is not limited to strings. It can be anything in the future, for example, `byte[]` for certificates.

modules/platforms/dotnet/Apache.Ignite/IgniteClientConfiguration.cs
ptupitsyn: Added.

isapego (approved):

PR #1952, opened 2023-04-18
IGNITE-19290 Removing garbage from the partition, taking into account the safe time

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19290


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -88,25 +89,31 @@ public class PartitionListener implements RaftGroupListener {
...
+     * @param lowWatermarkSupplier Low watermark supplier, will return {@code null} if no low watermark has been assigned yet.
      */
     public PartitionListener(
             PartitionDataStorage partitionDataStorage,
             StorageUpdateHandler storageUpdateHandler,
             TxStateStorage txStateStorage,
             PendingComparableValuesTracker<HybridTimestamp> safeTime,
-            PendingComparableValuesTracker<Long> storageIndexTracker
+            PendingComparableValuesTracker<Long> storageIndexTracker,
+            Supplier<HybridTimestamp> lowWatermarkSupplier
```
ibessonov, line 109: Can we stop using suppliers? It's annoying. When seeing  `() -> null` in constructor calls, I don't understand what it is.
Why can't we have a dedicated component/interface, that would provide the access to LW?

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/AbstractMvStorageUpdateHandlerTest.java
```diff
@@ -95,7 +96,8 @@ void setUp(
                 PARTITION_ID,
                 partitionDataStorage,
                 DummyInternalTableImpl.createTableIndexStoragesSupplier(Map.of()),
-                distributionZoneConfig.dataStorage()
+                distributionZoneConfig.dataStorage(),
+                new PendingComparableValuesTracker<>(HybridTimestamp.MAX_VALUE)
```
ibessonov, line 100: MAX_VALUE?
Interesting

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexBaseTest.java
```diff
@@ -129,7 +130,8 @@ void setUp(@InjectConfiguration DataStorageConfiguration dsCfg) {
                 PARTITION_ID,
                 new TestPartitionDataStorage(storage),
                 DummyInternalTableImpl.createTableIndexStoragesSupplier(indexes),
-                dsCfg
+                dsCfg,
+                new PendingComparableValuesTracker<>(HybridTimestamp.MAX_VALUE)
```
ibessonov, line 134: Same here, it's not obvious why you've chosen the max value.
And another question second time - are you sure that adding all this stuff to storage update handler is a good idea?

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandlerTest.java
```diff
@@ -104,6 +118,56 @@ void testBuildIndex() {
...
+    }
+
+    @Test
+    void testExecuteBatchGc() {
+        PartitionDataStorage partitionStorage = createPartitionDataStorage();
+
+        StorageUpdateHandler storageUpdateHandler = createStorageUpdateHandler(partitionStorage, mock(TableIndexStoragesSupplier.class));
+
+        // Let's check that if lwm is {@code null} then nothing will happen.
+        storageUpdateHandler.executeBatchGc(null);
```
ibessonov, line 150: Not obvious. What are the circumstances, in which we would pass `null` here?
Only tests?

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandlerTest.java
```diff
@@ -104,6 +118,56 @@ void testBuildIndex() {
...
+
+        verify(partitionStorage, never()).pollForVacuum(any(HybridTimestamp.class));
+
+        // Let's check that if lvm is greater than the safe time, then nothing will happen.
+        safeTimeTracker.update(new HybridTimestamp(10, 10));
+
+        storageUpdateHandler.executeBatchGc(new HybridTimestamp(11, 1));
```
ibessonov, line 157: I guess we have 2 separate APIs for GC, for partition listener and for regular background GC. Is that correct?
If it is correct, then why?

ibessonov: I re-checked, we do have two separate APIs for these 2 components. I'm confused

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
tkalkirill: Try to fix it.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/AbstractMvStorageUpdateHandlerTest.java
tkalkirill: So that when inserting into a table, the GC always worked for us.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/IndexBaseTest.java
tkalkirill: So that when inserting into a table, the GC always worked for us.

> And another question second time - are you sure that adding all this stuff to storage update handler is a good idea?
answered above.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandlerTest.java
tkalkirill: For example, when we start the node for the first time and have not yet assigned the first low watermark.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandlerTest.java
```diff
@@ -129,4 +193,13 @@ private void setRowVersions(PartitionDataStorage partitionStorage, Map<UUID, Lis
             when(partitionStorage.scanVersions(rowId)).thenReturn(Cursor.fromIterable(readResults));
         }
     }
+
+    private static PartitionDataStorage createPartitionDataStorage() {
+        PartitionDataStorage partitionStorage = mock(PartitionDataStorage.class);
```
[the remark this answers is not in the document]
tkalkirill: Fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
```diff
@@ -41,44 +40,51 @@
...
 import org.apache.ignite.internal.table.distributed.replicator.TablePartitionId;
 import org.apache.ignite.internal.util.Cursor;
+import org.apache.ignite.internal.util.PendingComparableValuesTracker;
 import org.jetbrains.annotations.Nullable;
 
 /**
  * Handler for storage updates that can be performed on processing of primary replica requests and partition replication requests.
  */
 public class StorageUpdateHandler {
-    /** Partition id. */
```
ibessonov, line 50: Was it necessary to remove this comment? Code, where some fields have comments, and others don't, looks ugly in my opinion. Another opinion - don't change the code that's absolutely fine already

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/gc/MvGc.java
```diff
@@ -221,34 +222,62 @@ private void scheduleGcForStorage(TablePartitionId tablePartitionId) {
...
+                    return;
                 }
-            }
 
-            scheduleGcForStorage(tablePartitionId);
+                StorageUpdateHandler storageUpdateHandler = storageHandler.storageUpdateHandler;
+
+                // We can only start garbage collection when the partition safe time is reached.
+                storageUpdateHandler.getSafeTimeTracker()
+                        .waitFor(lowWatermark)
```
ibessonov, line 246: How is it possible to have bad safe time inside of the GC task? Shouldn't we only add task to the pool if low watermark is good? Otherwise we end up adding a task that only adds another task, what's the point of it?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
tkalkirill: fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/gc/MvGc.java
tkalkirill: I don't quite understand what bad safe time means.
Since the safe time is a distributed property of each partition, we have to wait for it, don't we?

The future safe time could complete in a raft thread, so it's safer to move the garbage collection task to the garbage collector's pool.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -221,7 +228,8 @@ private void handleUpdateCommand(UpdateCommand cmd, long commandIndex, long comm
                     txsPendingRowIds.computeIfAbsent(cmd.txId(), entry -> new HashSet<>()).add(rowId);
 
                     storage.lastApplied(commandIndex, commandTerm);
-                }
+                },
+                lowWatermarkSupplier.get()
```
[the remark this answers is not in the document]
ibessonov: New ticket is good. We should compromise components responsibilities due to problems with copy-paste, it should be handled differently

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandlerTest.java
ibessonov: We can use the lowest timestamp possible, like every other component in the system, such partitions or meta-storage with their safe-times. Why would you introduce null where non-null value would work just fine?

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandlerTest.java
ibessonov: We have batch method for partition listener and non-batch method for normal GC

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/gc/MvGc.java
```diff
@@ -221,34 +222,62 @@ private void scheduleGcForStorage(TablePartitionId tablePartitionId) {
...
+                                        || !storageUpdateHandler.vacuum(lowWatermark)) {
+                                    return false;
+                                }
+                            }
+
+                            return true;
+                        }, executor)
+                        .whenComplete((isLeftGarbage, throwable) -> {
+                            if (throwable != null) {
+                                if (throwable instanceof TrackerClosedException
```
ibessonov, line 260: I believe only one of these options is actually possible. Why do you have both?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/gc/MvGc.java
```diff
@@ -221,34 +222,62 @@ private void scheduleGcForStorage(TablePartitionId tablePartitionId) {
...
+                                // If the storage has been deleted or there is no garbage, then we will stop.
+                                if (!storageHandlerByPartitionId.containsKey(tablePartitionId)
+                                        || !storageUpdateHandler.vacuum(lowWatermark)) {
+                                    return false;
+                                }
+                            }
+
+                            return true;
+                        }, executor)
+                        .whenComplete((isLeftGarbage, throwable) -> {
```
ibessonov, line 258: `isLeftGarbage` - maybe it should be `isGarbageLeft`, the order of words is definitely wrong

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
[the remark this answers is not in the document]
tkalkirill: Creates https://issues.apache.org/jira/browse/IGNITE-19396

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandlerTest.java
tkalkirill: In this case, I do not see anything contradictory and bad in `null`.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandlerTest.java
tkalkirill: Added item to https://issues.apache.org/jira/browse/IGNITE-19396

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/gc/MvGc.java
tkalkirill: If you run code something like this, you'll see an error like this:
```
        CompletableFuture<?> future = new CompletableFuture<>();

        future.thenAcceptAsync(unused -> {}).whenComplete((unused, throwable) -> throwable.printStackTrace());

        future.completeExceptionally(new TrackerClosedException());
```

`java.util.concurrent.CompletionException: org.apache.ignite.internal.util.TrackerClosedException`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/gc/MvGc.java
tkalkirill: fix it

ibessonov (approved):

PR #1953, opened 2023-04-18
IGNITE-19101 Basic authentication for C++ client

isapego:
- Added `ignite_client_authenticator` class;
- Added `basic_authenticator` class;
- Added tests.


modules/platforms/cpp/ignite/client/basic_authenticator.h
```diff
@@ -0,0 +1,92 @@
...
+    [[nodiscard]] const std::string& get_identity() const override {
+        return m_identity;
+    }
+
+    /**
+     * Set identity.
+     *
+     * @param identity Identity.
+     */
+    void set_identity(std::string identity) {
```
ptupitsyn, line 62: In my opinion, `identity` and `secret` are generic terms, common for all authenticators.

Basic authenticator works with `username` and `password` and should expose those properties to the user. Future use cases may include a `CertificateAuthenticator`, for example, where only `certificate` property can be set by the user. And so on.

See https://github.com/apache/ignite-3/blob/f6e5179214485c0826bdef6b6c745e25860671fa/modules/client/src/main/java/org/apache/ignite/client/BasicAuthenticator.java, https://github.com/apache/ignite-3/blob/af3fdd9485e2caa135eca251d3007f9ca93ad992/modules/platforms/dotnet/Apache.Ignite/BasicAuthenticator.cs

ptupitsyn (changes requested):

modules/platforms/cpp/ignite/client/basic_authenticator.h
isapego: Fixed.

ptupitsyn (approved):

PR #1954, opened 2023-04-18
IGNITE-19048: Extract Subscribers and Accumulators to core module

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-19048


modules/core/src/main/java/org/apache/ignite/internal/util/subscription/ListAccumulator.java
```diff
@@ -15,44 +15,41 @@
...
-import java.util.Collections;
 import java.util.List;
 import java.util.function.Function;
-import org.apache.ignite.internal.metastorage.Entry;
 
 /**
  * Plain list accumulator.
  *
- * @param <T> Result value type.
+ * @param <R> Result value type.
```
PakhomovAlexander, line 27: could you add javadoc for `R` type?

PakhomovAlexander (approved):

PR #1959, opened 2023-04-19
IGNITE-18961 Implement lease prolongation on placement driver side

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-18961

### DIFF
--- a/modules/placement-driver/src/integrationTest/java/org/apache/ignite/internal/placementdriver/MultiActorPlacementDriverTest.java
+++ b/modules/placement-driver/src/integrationTest/java/org/apache/ignite/internal/placementdriver/MultiActorPlacementDriverTest.java
@@ -1,0 +1,504 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.placementdriver;
+
+import static java.util.concurrent.CompletableFuture.completedFuture;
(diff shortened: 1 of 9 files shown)


modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
```diff
@@ -238,7 +238,10 @@ public void run() {
 
                     // The lease is expired or close to this.
                     if (lease.getExpirationTime().getPhysical() < outdatedLeaseThreshold) {
-                        ClusterNode candidate = nextLeaseHolder(entry.getValue(), null);
+                        ClusterNode candidate = nextLeaseHolder(
+                                entry.getValue(),
+                                lease.isAccepted() ? lease.getLeaseholder().name() : null
+                        );
```
denis-chudov, line 244: Why do we still select next lease holder every time? I thought we are going to change this in this ticket, this shouldn't be done for prolongation.

Also, the condition for meta storage invoke on prolongation is not sufficient. Invoke on prolongation should be successfull only when 
keys are equal 
AND 
leaseholder is the same
AND
lease is accepted
AND 
lease start and expiration times are the same.

Shouldn't it?

Also, invoke condition on lease granting should be revisited accordingly.

modules/placement-driver/src/integrationTest/java/org/apache/ignite/internal/placementdriver/MultiActorPlacementDriverTest.java
```diff
@@ -0,0 +1,518 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/**
+ * There are tests of muti-nodes for placement driver.
+ */
+@ExtendWith(ConfigurationExtension.class)
+public class MultiActorPlacementDriverTest extends IgniteAbstractTest {
```
denis-chudov, line 87: We definitely should have tests for two conflicting placement driver active actors. For example, there should be test for prolongation scenario. Or are we going to cover this under https://issues.apache.org/jira/browse/IGNITE-19325 ?
If so, the scenarios should be mentioned there, but I would say that we could cover more scenarios under regular tickets to reduce the amount of work required for IGNITE-19325.

denis-chudov (changes requested):

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
vldpyatkov: Yes, it is true. All your conditions are covered by tests.

PR #1961, opened 2023-04-20
IGNITE-19036 Add cli config remove command

valepakh:
https://issues.apache.org/jira/browse/IGNITE-19036


modules/cli/src/main/java/org/apache/ignite/internal/cli/call/cliconfig/CliConfigRemoveCallInput.java
```diff
@@ -0,0 +1,64 @@
...
+    private CliConfigRemoveCallInput(String key, String profileName) {
+        this.key = key;
+        this.profileName = profileName;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public String getKey() {
```
PakhomovAlexander, line 37: Could you please drop `get` prefix? As far as I know, we tent not to use them.

PakhomovAlexander (approved):

PR #1968, opened 2023-04-20
IGNITE-18624  Data nodes from DistributionZoneManager instead of BaselineManager#nodes on table creation.

sergeyuttsel:
https://issues.apache.org/jira/browse/IGNITE-18624


modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -161,6 +171,17 @@ public class DistributionZoneManager implements IgniteComponent {
...
+    /**
+     * If this property is set to {@code true} then an attempt to get the configuration property directly from Meta storage will be skipped,
+     * and the local property will be returned.
+     * TODO: IGNITE-16774 This property and overall approach, access configuration directly through Meta storage,
+     * TODO: will be removed after fix of the issue.
+     */
+    private final boolean getMetadataLocallyOnly = IgniteSystemProperties.getBoolean("IGNITE_GET_METADATA_LOCALLY_ONLY");
+
+    /** Versioned store for zones id. */
+    private final IncrementalVersionedValue<Set<Integer>> zonesByIdVv;
```
sanpwc, line 183: I don't think that you need versioned value here because you only use the it's side effects and nor it's core idea.
E.g. it's possible to add cfg listener in order to await zone with needed id. I mean following:
1. Read zone cfg through direct proxy and check whether zone with given name exists. Throw an exception if it doesn't.
2. Register zones listener in order to await zone locally with given Id. 
3. Check whether we already have such zone locally. Remove listener if we do have one.
2' Within listener we should do some useful stuff and remove the listener itself. Please ask SE guys whether it's possible to unregister the listener from within listener.

```
            DistributionZoneConfiguration zoneCfg = directProxy(zonesConfiguration.distributionZones()).get(zoneName);

            // check that zone is present.
            
            zonesConfiguration.distributionZones().listenElements(new ConfigurationNamedListListener<DistributionZoneView>() {
                @Override
                public CompletableFuture<?> onCreate(ConfigurationNotificationEvent<DistributionZoneView> ctx) {

                    if (ctx.newValue().zoneId() == zoneCfg.zoneId().value()) {
                        // do something usefull
                        // unregister current listener. Please ask SE guys whether it's possible to unregister the listener from within listener.
                    }
                }
            });
```
I'm not sure whether solution above is best, however it seems better to versioned value one. We can think about other options too.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -859,7 +885,17 @@ public CompletableFuture<?> onCreate(ConfigurationNotificationEvent<Distribution
 
             saveDataNodesAndUpdateTriggerKeysInMetaStorage(zoneId, ctx.storageRevision(), logicalTopology);
 
-            return completedFuture(null);
+            return zonesByIdVv.update(ctx.storageRevision(), (zones, e) -> {
+                if (e != null) {
```
sanpwc, line 889: What about busy lock? Do we need it here?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1614,6 +1662,101 @@ CompletableFuture<Void> saveDataNodesToMetaStorageOnScaleDown(int zoneId, long r
...
+        }
+    }
+
+    /**
+     * Internal method for waiting that the zone is created locally.
+     *
+     * @param id Table id.
+     * @return Future representing pending completion of the operation.
+     */
+    private CompletableFuture<Void> waitZoneIdLocally(Integer id) {
```
sanpwc, line 1709: I'm not sure that I get how you handle null as an id. I mean that directZoneIdInternal may return null if there's no such zone in direct cfg.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1614,6 +1662,101 @@ CompletableFuture<Void> saveDataNodesToMetaStorageOnScaleDown(int zoneId, long r
...
+            }
+        } catch (NoSuchElementException e) {
+            return null;
+        }
+    }
+
+    /**
+     * Internal method for waiting that the zone is created locally.
+     *
+     * @param id Table id.
```
sanpwc, line 1706: Seems that it's not table but zone id.

sanpwc (changes requested):

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
```diff
@@ -120,21 +125,28 @@ public void setUp() throws Exception {
...
                 "test"
         );
 
         vaultMgr.start();
         clusterCfgMgr.start();
         metaStorageManager.start();
-        distributionZoneManager.start();
 
-        metaStorageManager.deployWatches();
+        deployWatchesAndUpdateMetaStorageRevision(metaStorageManager);
```
sanpwc, line 147: Why do we deploy watches before starting the distributionZoneManager? It seems incorrect, because common practice is to deploy watches after all components start up.

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerScaleUpTest.java
```diff
@@ -1366,10 +1379,10 @@ private void assertNotEqualsDataNodesForZone(int zoneId, @Nullable Set<String> c
         ));
     }
 
-    private void startDistributionZoneManager() throws NodeStoppingException {
-        distributionZoneManager.start();
+    private void startDistributionZoneManager() throws Exception {
+        deployWatchesAndUpdateMetaStorageRevision(metaStorageManager);
```
sanpwc, line 1383: And again, the order of deploying watches an starting dzm is probably incorrect.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
```diff
@@ -980,11 +983,11 @@ private void assertTablePresent(TableManager tableManager, String tableName) {
     public void testOneNodeRestartWithGap() throws InterruptedException {
         IgniteImpl ignite = startNode(0);
 
-        List<IgniteComponent> components = startPartialNode(1, null);
+        startNode(1);
```
sanpwc, line 986: Why it's full node instead of partial?

modules/table/build.gradle
```diff
@@ -41,6 +41,7 @@ dependencies {
     implementation project(':ignite-configuration-api')
     implementation project(':ignite-distribution-zones')
     implementation project(':ignite-vault')
+    implementation project(':ignite-cluster-management')
```
sanpwc, line 44: Why do we need it? For which dependency?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1343,72 +1355,132 @@ private Set<Assignment> calculateAssignments(TableConfiguration tableCfg, int pa
...
-    private CompletableFuture<Table> createTableAsyncInternal(String name, Consumer<TableChange> tableInitChange) {
+    /** See {@link #createTableAsync(String, String, Consumer)} for details. */
+    private CompletableFuture<Table> createTableAsyncInternal(
+            String name,
+            String zoneName,
+            Consumer<TableChange> tableInitChange
+    ) {
         CompletableFuture<Table> tblFut = new CompletableFuture<>();
+        tableAsyncInternal(name)
+                .thenAccept(tbl -> {
```
sanpwc, line 1377: Seems that busy locks are missing, aren't they?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1343,72 +1355,132 @@ private Set<Assignment> calculateAssignments(TableConfiguration tableCfg, int pa
...
+                    if (tbl != null) {
+                        tblFut.completeExceptionally(new TableAlreadyExistsException(DEFAULT_SCHEMA_NAME, name));
+                    } else {
+                        distributionZoneManager.zoneIdAsyncInternal(zoneName).handle((zoneId, ex) -> {
```
sanpwc, line 1381: Could you please name the tests that check table await logic?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
sergeyuttsel: Thanks. I've removed VersiondValue and used ConfigurationNamedListListener to await the zone is created locally. I discussed this approach with Ivan Bessonov and he confirmed that it is possible
to invoke NamedConfigurationTree#stopListenElements in the listener in this case.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
sergeyuttsel: Fixed. I removed this code.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
sergeyuttsel: Thanks. I've added null handling in zoneIdAsyncInternal.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
sergeyuttsel: Fixed.

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerTest.java
```diff
@@ -91,6 +91,7 @@ public void setUp() {
                 null,
                 null,
                 null,
+                x -> {},
```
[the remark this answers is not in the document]
sergeyuttsel: Fixed.

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerWatchListenerTest.java
```diff
@@ -136,20 +141,20 @@ void testStaleVaultRevisionOnZoneManagerStart() throws InterruptedException {
...
 
     @Test
-    void testDataNodesUpdatedOnZoneManagerStart() throws InterruptedException {
+    void testDataNodesUpdatedOnZoneManagerStart() throws Exception {
         Set<String> nodes = Set.of("node1", "node2");
 
         mockVaultZonesLogicalTopologyKey(nodes);
 
-        distributionZoneManager.start();
+        startDistributionZoneManager();
```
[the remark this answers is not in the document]
sergeyuttsel: I will create another ticket to fix it properly.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/MockedStructuresTest.java
```diff
@@ -162,13 +177,15 @@ public class MockedStructuresTest extends IgniteAbstractTest {
     @InjectConfiguration
     private TablesConfiguration tblsCfg;
 
-    @InjectConfiguration("mock.distributionZones.zone123{dataStorage.name = " + ENGINE_NAME + ", zoneId = 1}")
+    @InjectConfiguration("mock.distributionZones." + ZONE_NAME + "{dataStorage.name = " + ENGINE_NAME + ", zoneId = 1}")
```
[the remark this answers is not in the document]
sergeyuttsel: Fixed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sergeyuttsel: Fixed.

sergeyuttsel: Fixed

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
sanpwc: Please address given question. You may create separate ticket if needed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sanpwc: Which of the aforementioned tests do check what will we await dataNodes evaluation for immediate timers?
Meaning that there should be at least following tests:
1. The one that check that we await **topology version** within DZM.
2. The one that check that we await immediate timers but skip non-immediate ones.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
```diff
@@ -994,9 +997,9 @@ public void testOneNodeRestartWithGap() throws InterruptedException {
 
         createTableWithoutData(ignite, TABLE_NAME_2, 1, 1);
 
-        components = startPartialNode(1, null);
+        IgniteImpl ignite1 = startNode(1);
```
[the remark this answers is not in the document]
sergeyuttsel: Added TODO https://issues.apache.org/jira/browse/IGNITE-19408

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
sergeyuttsel: Added TODO https://issues.apache.org/jira/browse/IGNITE-19408

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1345,72 +1357,158 @@ private Set<Assignment> calculateAssignments(TableConfiguration tableCfg, int pa
...
+                                } else {
+                                    if (!busyLock.enterBusy()) {
+                                        throw new IgniteException(new NodeStoppingException());
+                                    }
+
+                                    try {
+                                        cmgMgr.logicalTopology()
+                                                .handle((cmgTopology, e) -> {
+                                                    if (e == null) {
+                                                        if (!busyLock.enterBusy()) {
```
sanpwc, line 1403: Here and there. Is it possible to use `IgniteUtils#inBusyLock()` instead?

sergeyuttsel: I cannot use inBusyLock because I need not only throw NodeStoppingException but also complete tblFut with NodeStoppingException.

sanpwc (approved):

PR #1969, opened 2023-04-21
IGNITE-19270 Thin 3.0: Streamline type codes

ptupitsyn:
Replace multiple sets of column types and codes used by the client protocol (`ColumnType`, `ClientDataType`, `ClientColumnTypeConverter`) with a single one - `ColumnType` from the public API.


modules/platforms/cpp/ignite/client/detail/client_data_type.h
isapego, line 1: Could be totally replaced with `ignite_type` class from `modules\platforms\cpp\ignite\common\ignite_type.h`.

isapego (changes requested):

modules/platforms/cpp/ignite/client/detail/client_data_type.h
ptupitsyn: Fixed.

ademakov (approved):

isapego (approved):

PR #1970, opened 2023-04-21
IGNITE-19107: Remove LATEST semantic from deployment API

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-19107


modules/cli/src/main/java/org/apache/ignite/internal/cli/Main.java
```diff
@@ -56,7 +56,7 @@ public static void main(String[] args) {
         try (MicronautFactory micronautFactory = new MicronautFactory(builder.start())) {
             AnsiConsole.systemInstall();
             initReplExecutor(micronautFactory);
-            if (args.length != 0 || !isatty()) { // do not enter REPL if input or output is redirected
+            if (args.length != 0 /*|| !isatty()*/) { // do not enter REPL if input or output is redirected
```
PakhomovAlexander, line 59: Could you please explain why does this change make sense?

PakhomovAlexander (approved):

PR #1978, opened 2023-04-25
IGNITE-18712 Do not allow a node excluded from Physical Topology to enter the topology again

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18712

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -121,6 +121,7 @@ subprojects {
         }
 
         maven {
+            // TODO: IGNITE-19386 - switch to io.scalecube:scalecube-cluster
             url "https://jitpack.io"
         }
     }
--- a/build.gradle
+++ b/build.gradle
@@ -119,6 +119,10 @@ subprojects {
                 releasesOnly()
             }
         }
+
+        maven {
+            url "https://jitpack.io"
+        }
     }
 
     tasks.register('printSubDependencies', DependencyReportTask)
--- a/modules/network/src/integrationTest/java/org/apache/ignite/network/scalecube/ItScaleCubeNetworkMessagingTest.java
+++ b/modules/network/src/integrationTest/java/org/apache/ignite/network/scalecube/ItScaleCubeNetworkMessagingTest.java
@@ -25,7 +25,6 @@ import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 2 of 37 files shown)


build.gradle
```diff
@@ -119,6 +119,10 @@ subprojects {
                 releasesOnly()
             }
         }
+
+        maven {
+            url "https://jitpack.io"
```
SammyVimes, line 124:
```suggestion
        maven {
            // TODO: IGNITE-19386 - switch to io.scalecube:scalecube-cluster
            url "https://jitpack.io"
```

modules/network/src/integrationTest/java/org/apache/ignite/network/scalecube/ItScaleCubeNetworkMessagingTest.java
```diff
@@ -320,6 +323,69 @@ public void testMessageGroupsHandlers(TestInfo testInfo) throws Exception {
...
+
+        testCluster.startAwait();
+
+        String outcastName = testCluster.members.get(testCluster.members.size() - 1).nodeName();
+
+        knockOutNode(outcastName);
+
+        CountDownLatch appeared = reanimateNode(outcastName);
+
+        assertFalse(appeared.await(3, TimeUnit.SECONDS), "Node reappeared");
```
SammyVimes, line 343: So this test is always 3 seconds, right? Can we maybe listen for some event like "Node failed to join because of the stale id"? I see we have a FailureHandler, that can be used for that, right?

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/VaultStateIds.java
```diff
@@ -0,0 +1,102 @@
...
+
+        Collections.addAll(result, idsArray);
+
+        return result;
+    }
+
+    @Override
+    public synchronized void markAsStale(String nodeId) {
+        if (staleIds == null) {
+            staleIds = new LinkedHashSet<>();
```
SammyVimes, line 81: Why don't we use `loadStaleIdsFromVault()` here?

And a side note, can't we do `loadStaleIdsFromVault` on start?

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
```diff
@@ -199,6 +219,25 @@ public void updateMetadata(NodeMetadata metadata) {
                 cluster.updateMetadata(metadata).subscribe();
                 topologyService.updateLocalMetadata(metadata);
             }
+
+            private String notNullClusterMemberId() {
```
SammyVimes, line 223: Looks like we don't need it anymore

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
```diff
@@ -199,6 +219,25 @@ public void updateMetadata(NodeMetadata metadata) {
...
+                    if (memberId != null) {
+                        return memberId;
+                    }
+
+                    LockSupport.parkNanos(10_000_000);
+                }
+            }
+
+            @Nullable
+            private String clusterMemberId() {
```
SammyVimes, line 236: This too, right? We can just use launchId directly

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestThreeNodesThreeReplicas.java
```diff
@@ -51,9 +51,11 @@ protected int replicas() {
     @Override
     @AfterEach
     public void after() throws Exception {
-        assertTrue(IgniteTestUtils.waitForCondition(() -> assertPartitionsSame(accounts, 0), 5_000));
-        assertTrue(IgniteTestUtils.waitForCondition(() -> assertPartitionsSame(customers, 0), 5_000));
-
-        super.after();
+        try {
+            assertTrue(IgniteTestUtils.waitForCondition(() -> assertPartitionsSame(accounts, 0), 5_000));
```
SammyVimes, line 55: `TimeUnit.SECONDS.toMillis(5)` looks a bit better

SammyVimes (changes requested):

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/VaultStateIds.java
rpuch: Oops, a bug. Fixed it.

I did not load on start to avoid bringing in the concept of starting this component, it's too tiny to complicate it that much.

modules/network/src/main/java/org/apache/ignite/internal/network/netty/HandshakeHandler.java
```diff
@@ -67,7 +67,13 @@ public void handlerAdded(ChannelHandlerContext ctx) {
     /** {@inheritDoc} */
     @Override
     public void channelActive(ChannelHandlerContext ctx) {
-        manager.onConnectionOpen();
+        try {
+            manager.onConnectionOpen();
+        } catch (RuntimeException | AssertionError e) {
+            LOG.error("Error in onConnectionOpen()", e);
```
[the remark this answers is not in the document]
SammyVimes: It's weird to catch AssertionError, maybe it should be just `Throwable` then?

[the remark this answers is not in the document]
rpuch: Fixed as you suggested

SammyVimes (approved):
LGTM

PR #1983, opened 2023-04-25
IGNITE-19362 Fix deadlock while restarting with existing raft snapshot and scheduled index rebuild

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-19362

### DIFF
--- a/modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/snapshot/SnapshotExecutorImpl.java
+++ b/modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/snapshot/SnapshotExecutorImpl.java
@@ -250,6 +250,7 @@ public class SnapshotExecutorImpl implements SnapshotExecutor {
         final FirstSnapshotLoadDone done = new FirstSnapshotLoadDone(reader);
         Requires.requireTrue(this.fsmCaller.onSnapshotLoad(done));
         try {
+            // TODO: IGNITE-19363 We want to avoid deadlock for now, but this is an ad-hoc decision.
             if (!node.getNodeId().getGroupId().contains("part")) {
                 done.waitForRun();
             } else {
--- a/modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/snapshot/SnapshotExecutorImpl.java
+++ b/modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/snapshot/SnapshotExecutorImpl.java
@@ -251,6 +251,7 @@ public class SnapshotExecutorImpl implements SnapshotExecutor {
         Requires.requireTrue(this.fsmCaller.onSnapshotLoad(done));
         try {
             // TODO: IGNITE-19363 We want to avoid deadlock for now, but this is an ad-hoc decision.
+            // We don't wait for the partition's snapshot load closure to finish here.
             if (!node.getNodeId().getGroupId().contains("part")) {
                 done.waitForRun();
             } else {
--- a/modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/snapshot/SnapshotExecutorImpl.java
+++ b/modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/snapshot/SnapshotExecutorImpl.java
@@ -255,7 +255,7 @@ public class SnapshotExecutorImpl implements SnapshotExecutor {
             if (!node.getNodeId().getGroupId().contains("part")) {
                 done.waitForRun();
             } else {
-                done.status = new Status();
+                done.status = Status.OK();
             }
         }
         catch (final InterruptedException e) {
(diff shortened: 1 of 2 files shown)


modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
```diff
@@ -878,6 +879,48 @@ public void nodeWithDataTest() throws InterruptedException {
...
+
+        createTableWithData(List.of(ignite), TABLE_NAME, 1, 20);
+
+        TableImpl table = (TableImpl) ignite.tables().table(TABLE_NAME);
+
+        InternalTableImpl internalTable = (InternalTableImpl) table.internalTable();
+
+        for (int i = 0; i < 20; i++) {
+            // Flush data on disk, so that we will have a snapshot to read on restart.
+            internalTable.storage().getMvPartition(i).flush().join();
```
sashapolo, line 897: Let's not use `join` here and use an assertion instead

sashapolo: Also, these flushes can be done in parallel, no need to wait for them sequentially

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
```diff
@@ -878,6 +879,48 @@ public void nodeWithDataTest() throws InterruptedException {
...
+        }
+
+        // Add more data, so that on restart there will be a index rebuilding operation.
+        try (Session session = ignite.sql().createSession()) {
+            for (int i = 0; i < 100; i++) {
+                session.execute(null, "INSERT INTO " + TABLE_NAME + "(id, name) VALUES (?, ?)",
+                        i + 500, VALUE_PRODUCER.apply(i + 500));
```
sashapolo, line 904: What's `+ 500` for?

modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/snapshot/SnapshotExecutorImpl.java
```diff
@@ -250,7 +250,13 @@ public boolean init(final SnapshotExecutorOptions opts) {
         final FirstSnapshotLoadDone done = new FirstSnapshotLoadDone(reader);
         Requires.requireTrue(this.fsmCaller.onSnapshotLoad(done));
         try {
-            done.waitForRun();
+            // TODO: IGNITE-19363 We want to avoid deadlock for now, but this is an ad-hoc decision.
+            // We don't wait for the partition's snapshot load closure to finish here.
+            if (!node.getNodeId().getGroupId().contains("part")) {
+                done.waitForRun();
+            } else {
+                done.status = new Status();
```
sashapolo, line 258: How does this work? Don't we have a race here with setting the task status? What will happen if snapshot loading fails?

sashapolo: Also, there's a `Status.OK()` method, which looks a little bit more descriptive

SammyVimes: This is a temporal solution, so we don't really care about failing loading the snapshot. The race doesn't matter, we won't be checking the result

sashapolo: I really hope that it is temporary, not checking the status seems dangerous

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
```diff
@@ -878,6 +879,54 @@ public void nodeWithDataTest() throws InterruptedException {
...
+
+        ignite = startNode(0);
+
+        checkTableWithData(ignite, TABLE_NAME);
+
+        table = (TableImpl) ignite.tables().table(TABLE_NAME);
+
+        // Check data that was added after flush.
+        for (int i = 0; i < 100; i++) {
+            Tuple row = table.keyValueView().get(null, Tuple.create().set("id", i));
```
sashapolo, line 924: How does this work now, if we still insert data with id equal to `i + 500`?

modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/snapshot/SnapshotExecutorImpl.java
SammyVimes: It is, I am going to start on fixing it right away

sashapolo (approved):

PR #1986, opened 2023-04-26
IGNITE-18755 Do not save the empty data nodes list to avoid the distribution zone without a data nodes. Fixed exceptions in test logs of DistributionZoneManagerScaleUpTest, DistributionZoneManagerConfigurationChangesTest, DistributionZoneManagerWatchListenerTest.

sergeyuttsel:
https://issues.apache.org/jira/browse/IGNITE-18755


modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1595,6 +1598,11 @@ CompletableFuture<Void> saveDataNodesToMetaStorageOnScaleDown(int zoneId, long r
                         .thenApply(StatementResult::getAsBoolean)
                         .thenCompose(invokeResult -> inBusyLock(busyLock, () -> {
                             if (invokeResult) {
+                                if (newDataNodes.isEmpty()) {
+                                    LOG.debug("Data nodes for a zone was not updated because"
```
sanpwc, line 1602: Should we still update trigger key?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1595,6 +1598,11 @@ CompletableFuture<Void> saveDataNodesToMetaStorageOnScaleDown(int zoneId, long r
                         .thenApply(StatementResult::getAsBoolean)
                         .thenCompose(invokeResult -> inBusyLock(busyLock, () -> {
                             if (invokeResult) {
+                                if (newDataNodes.isEmpty()) {
```
sanpwc, line 1601: How it's supposed to work?
Let's say that we've prepared new dataNodesAndTriggerKeyUpd with new revision and **same** newDataNodesBytes.
Are we going to actually update dataNodes with same value? Are we going to consider such same to same change as new rebalance trigger?

sergeyuttsel: 1. Same to same a data nodes change will not trigger a new rebalance because RebalanceUtil#updatePendingAssignmentsKeys filters the same assignments.
2. I think we can not save the same data nodes, but it will require more changes in the code. Actually a meta storage invoke with the same data nodes can not complete successfully and `saveDataNodesToMetaStorageOnScaleDown` can be repeated again but with a different set of nodes.

PR #1992, opened 2023-04-27
IGNITE-19302 Phantom reads fixed

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-19302

### DIFF
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
@@ -19,9 +19,11 @@ package org.apache.ignite.internal.table;
 
 import static org.apache.ignite.internal.index.SortedIndex.INCLUDE_LEFT;
 import static org.apache.ignite.internal.index.SortedIndex.INCLUDE_RIGHT;
+import static org.apache.ignite.internal.testframework.IgniteTestUtils.runRace;
 import static org.apache.ignite.internal.testframework.IgniteTestUtils.waitForCondition;
 import static org.apache.ignite.internal.testframework.matchers.CompletableFutureMatcher.willCompleteSuccessfully;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
@@ -38,6 +40,7 @@ import java.util.concurrent.Flow.Subscriber;
 import java.util.concurrent.Flow.Subscription;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
+import java.util.function.IntFunction;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import org.apache.ignite.Ignite;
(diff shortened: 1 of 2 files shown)


modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
```diff
@@ -178,9 +181,8 @@ public void testInsertDuringScan() throws Exception {
 
         subscription.request(1);
 
-        waitForCondition(() -> !scannedRows.isEmpty(), 10_000);
+        assertTrue(waitForCondition(() -> !scannedRows.isEmpty(), 10_000));
 
-        assertEquals(1, scannedRows.size());
```
rpuch, line 183: The change is not equivalent: earlier it checked that there is exactly 1 element, now it will allow any non-zero number of elements

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
```diff
@@ -590,14 +513,71 @@ public void testScanWithUpperBound() throws Exception {
...
+            InternalTransaction tx = startTxWithEnlistedPartition(partId);
+
+            try {
+                IgniteBiTuple<ClusterNode, Long> leaderWithTerm = tx.enlistedNodeAndTerm(new TablePartitionId(table.tableId(), partId));
+
+                PrimaryReplica recipient = new PrimaryReplica(leaderWithTerm.get1(), leaderWithTerm.get2());
+
+                Publisher<BinaryRow> publisher = internalTable.scan(partId, tx.id(), recipient, sortedIndexId, null, null, 0, null);
+
+                // Non-thread-safe collection is fine, HB is guaranteed by "Thread#join" inside of "runRace".
```
rpuch, line 545: How about just replacing the collection with a thread-safe one, so that the thread-safety question do not even arise? It's a test, performance does not seem to be critical here. Then the comment is not needed at all.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
```diff
@@ -155,15 +154,15 @@ public CompletableFuture<Lock> locksForInsert(UUID txId, BinaryRow tableRow, Row
...
-            nextKey = cursor.next().indexColumns();
-        } else { // otherwise INF
-            nextKey = POSITIVE_INF;
-        }
+        //noinspection ResultOfMethodCallIgnored
+        cursor.hasNext();
+
+        // "peek" works the same as "next" when "hasNext" is called. The only difference is that it returns null instead of throwing
+        // an exception, if nothing is found.
+        BinaryTuple nextKey = indexKey(cursor.peek());
```
rpuch, line 165: The code 'after' seems to be equivalent to the code 'before', but the code 'before' was arguably more straightforward. Are you sure this actually improves readability?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
ibessonov: I don't like using thread-safe collections when no one needs them, I consider this a bad practice.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
ibessonov: Old code was a scenario copy-pasted twice, I removed the copy-paste.
What else is equivalent? Replacement of "hasNext()" with "peek() != null" is not an equivalent change, it's a fix

rpuch (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
rpuch: My comment is not relevant anymore, please ignore it

PR #1995, opened 2023-04-27
IGNITE-18991 Move stable/planned/pending assignments from table to distribution zone root keys

kgusakov:
https://issues.apache.org/jira/browse/IGNITE-18991


modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/configuration/RebalanceValidator.java
```diff
@@ -0,0 +1,52 @@
...
+
+        int zoneId = ((DistributionZoneView) ctx.getNewOwner()).zoneId();
+
+        if (ctx.getOldValue() != null
+                && !ctx.getOldValue().equals(ctx.getNewValue())
+                && tablesView.tables().stream().filter(t -> t.zoneId() == zoneId).count() > 1) {
+            ctx.addIssue(
+                    new ValidationIssue(
+                            ctx.currentKey(),
+                            String.format("Change the number of replicas fot the zone with the > 1 tables is not supported yet")));
```
sergeyuttsel, line 49: Typo: "for" -> "for". Also `I see assert tblIds.size() == 1 : "More than one table found for zoneId=" + zoneId;` in TableManager. Is it possible to have more then one table per zone?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/configuration/IsReplicasChangeAvailable.java
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Current rebalance mechanism based on the restriction, that the target zone has at most 1 table.
+ * This restriction is temporary and must be fixed after the implementation of zone-based data distribution
+ * instead the table-based.
+ *
+ * <p>So, this annotation must be place only on the {@link DistributionZoneConfiguration#replicas()} field to provide the check
```
sanpwc, line 30: There's one more place - TableManager#createDistributionZonesDataNodesListener.

- The main point here is that given listener should be moved to distribution zone.

- Another one is that because of dataNodes trigger validator based approach might not be an option.

kgusakov:
After discussion we decided, that in the current state PR is not actual and the useful part is blocked by https://issues.apache.org/jira/browse/IGNITE-19170

PR #2000, opened 2023-04-29
IGNITE-18761 Sql. TypeOf can short-circuit only when its argument is a constant expression.

lowka:


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItFunctionsTest.java
```diff
@@ -267,16 +271,40 @@ public void testTypeOf() {
...
+        assertQuery("SELECT TYPEOF('abcd' || COALESCE('efg', ?))").withParams("2").returns("VARCHAR").check();
+
+        // An expression that produces an error
+        IgniteException failed = assertThrows(IgniteException.class, () -> assertQuery("SELECT typeof(CAST('NONE' as INTEGER))").check());
+        assertSame(NumberFormatException.class, failed.getCause().getClass(), "cause");
+        assertThat(failed.getCause().getMessage(), containsString("For input string: \"NONE\""));
+
         try {
             sql("SELECT TYPEOF()");
+            fail();
```
AMashenkov, line 284: assertThrowsWithCause(() -> sql("SELECT TYPEOF()"), SqlValidatorException.class, "Invalid number of arguments");

lowka: fixed.

PR #2001, opened 2023-04-30
IGNITE-18711 Leave logical topology gracefully when stopping Ignite node

rpuch:
https://issues.apache.org/jira/browse/IGNITE-18711

### DIFF
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/compute/ItLogicalTopologyTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/compute/ItLogicalTopologyTest.java
@@ -18,8 +18,10 @@
 package org.apache.ignite.internal.compute;
 
 import static org.apache.ignite.internal.testframework.IgniteTestUtils.testNodeName;
+import static org.apache.ignite.internal.testframework.IgniteTestUtils.waitForCondition;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.empty;
+import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.notNullValue;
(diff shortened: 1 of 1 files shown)


modules/runner/src/integrationTest/java/org/apache/ignite/internal/compute/ItLogicalTopologyTest.java
```diff
@@ -337,6 +339,27 @@ public void onNodeValidated(LogicalNode validatedNode) {
...
+    void nodeLeavesLogicalTopologyOnStop() throws Exception {
+        IgniteImpl entryNode = node(0);
+
+        IgniteImpl secondIgnite = startNode(1);
+
+        entryNode.logicalTopologyService().addEventListener(listener);
+
+        stopNode(1);
+
+        assertTrue(waitForCondition(() -> !events.isEmpty(), 10_000), "Did not see any events in time");
```
SammyVimes, line 352:
```suggestion
        assertTrue(waitForCondition(() -> !events.isEmpty(), TimeUnit.SECONDS.toMillis(10)), "Did not see any events in time");
```
Do we have to wait 10 seconds?

SammyVimes (changes requested):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/compute/ItLogicalTopologyTest.java
rpuch: Applied the change.

If the test passes, the wait will be satisfied in much less than 10 seconds. If the logic is broken, the test will take 10 seconds (not a big deal for a failing test), but we'll not see false positives due to a too short timeout.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/compute/ItLogicalTopologyTest.java
```diff
@@ -337,6 +339,27 @@ public void onNodeValidated(LogicalNode validatedNode) {
...
+    void nodeLeavesLogicalTopologyOnStop() throws Exception {
+        IgniteImpl entryNode = node(0);
+
+        IgniteImpl secondIgnite = startNode(1);
+
+        entryNode.logicalTopologyService().addEventListener(listener);
+
+        stopNode(1);
+
+        assertTrue(waitForCondition(() -> !events.isEmpty(), TimeUnit.SECONDS.toMillis(10)), "Did not see any events in time");
```
sashapolo, line 352: Why do you need to use `waitForCondition` if you can simply `poll` the event queue in order to wait for an event to appear?

SammyVimes (approved):

PR #2003, opened 2023-05-01
IGNITE-19390 Handling lease deny message in placement driver side

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-19390

### DIFF
--- a/modules/placement-driver-api/src/main/java/org/apache/ignite/internal/placementdriver/message/PlacementDriverActorMessage.java
+++ b/modules/placement-driver-api/src/main/java/org/apache/ignite/internal/placementdriver/message/PlacementDriverActorMessage.java
@@ -17,24 +17,19 @@
 
 package org.apache.ignite.internal.placementdriver.message;
 
-import static org.apache.ignite.internal.placementdriver.message.PlacementDriverMessageGroup.GROUP_NAME;
-import static org.apache.ignite.internal.placementdriver.message.PlacementDriverMessageGroup.GROUP_TYPE;
-
-import org.apache.ignite.network.annotations.MessageGroup;
+import org.apache.ignite.internal.replicator.ReplicationGroupId;
+import org.apache.ignite.network.NetworkMessage;
+import org.apache.ignite.network.annotations.Marshallable;
 
 /**
- * Message group for placement driver messages.
+ * Message which is sent from the replica side to an actor node.
  */
-@MessageGroup(groupType = GROUP_TYPE, groupName = GROUP_NAME)
-public interface PlacementDriverMessageGroup {
-    /** Placement driver message group type. */
-    short GROUP_TYPE = 11;
-
-    String GROUP_NAME = "PlacementDriverMessages";
-
-    short LEASE_GRANTED_MESSAGE = 0;
-
-    short LEASE_GRANTED_MESSAGE_RESPONSE = 1;
-
-    short STOP_LEASE_PROLONGATION = 2;
+public interface PlacementDriverActorMessage extends NetworkMessage {
+    /**
+     * Gets a replication group id.
+     *
+     * @return Replication group id.
+     */
+    @Marshallable
+    ReplicationGroupId groupId();
 }
--- a/modules/placement-driver-api/src/main/java/org/apache/ignite/internal/placementdriver/message/PlacementDriverActorMessage.java
+++ b/modules/placement-driver-api/src/main/java/org/apache/ignite/internal/placementdriver/message/PlacementDriverActorMessage.java
@@ -17,19 +17,8 @@
 
 package org.apache.ignite.internal.placementdriver.message;
 
-import org.apache.ignite.internal.replicator.ReplicationGroupId;
-import org.apache.ignite.network.NetworkMessage;
-import org.apache.ignite.network.annotations.Marshallable;
-
 /**
- * Message which is sent from the replica side to an actor node.
+ * This is a marker interface. It marks a massage type that is sent from a replica to placement driver actor.
  */
-public interface PlacementDriverActorMessage extends NetworkMessage {
-    /**
-     * Gets a replication group id.
-     *
-     * @return Replication group id.
-     */
-    @Marshallable
-    ReplicationGroupId groupId();
+public interface PlacementDriverActorMessage extends PlacementDriverMessage {
 }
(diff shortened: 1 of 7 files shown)


denis-chudov (changes requested):

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/Lease.java
```diff
@@ -95,7 +111,20 @@ public Lease acceptLease(HybridTimestamp to) {
...
+     * Denies the lease.
+     *
+     * @return Denied lease.
+     */
+    public Lease denyLease() {
+        assert accepted : "The lease is not accepted ["
+                + "leaseholder=" + leaseholder
+                + ", expirationTime=" + expirationTime + ']';
+
+        return new Lease(leaseholder, startTime, expirationTime, false, true);
```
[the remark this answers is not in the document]
vldpyatkov: Accepted lease means we cannot change their expiration timestamp. Moreover, it is worth to refers to accepted flag like a storage marker (all leases with accepted flag are available for placement driver, but not available for other nodes). And here we shouldn't to remove an active lease (not expired) from one storage, because it should be available for all until their time is not elapsed.

modules/placement-driver-api/src/main/java/org/apache/ignite/internal/placementdriver/message/PlacementDriverActorMessage.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public interface PlacementDriverActorMessage extends NetworkMessage {
+    /**
+     * Gets a replication group id.
+     *
+     * @return Replication group id.
+     */
+    @Marshallable
+    ReplicationGroupId groupId();
+}
```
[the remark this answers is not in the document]
vldpyatkov: It is done.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
```diff
@@ -395,4 +407,54 @@ private boolean isLeaseOutdated(Lease lease) {
...
+         * @param sender Sender node name.
+         * @param msg Message.
+         */
+        private void processMessageInternal(String sender, PlacementDriverActorMessage msg) {
+            ReplicationGroupId grpId = msg.groupId();
+
+            Lease lease = leaseTracker.getLease(grpId);
+
+            if (msg instanceof StopLeaseProlongationMessage) {
+                if (lease.isProlong() && sender.equals(lease.getLeaseholder().name())) {
```
denis-chudov, line 446: maybe rename this method to `isProlongable`?

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
```diff
@@ -395,4 +407,54 @@ private boolean isLeaseOutdated(Lease lease) {
...
+            if (!(msg0 instanceof PlacementDriverActorMessage)) {
+                return;
+            }
+
+            var msg = (PlacementDriverActorMessage) msg0;
+
+            if (!stateActorLock.enterBusy()) {
+                return;
+            }
+
```
denis-chudov, line 426: So we just throw this message away if this is not an active actor? Imagine following:
- active actor crashes
- PD group leader election is in progress
- some group grp0 loses majority and wants to notify PD about this
- every known PD node throws away the message
- PD elects leader fast enough, so the lease for grp0 is not outdated and the same leaseholder is chosen.

So the lease would be prolonged as if nothing happened, right?

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/Lease.java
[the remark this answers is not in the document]
denis-chudov: Well, I am wrong here about accepted flag - but still think that logging on lease denial after successful invoke might be useful

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
vldpyatkov: Renamed.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/Lease.java
```diff
@@ -81,7 +92,12 @@ public Lease prolongLease(HybridTimestamp to) {
                 + ", expirationTime=" + expirationTime
                 + ", prolongTo=" + to + ']';
 
-        return new Lease(leaseholder, startTime, to, true);
+        assert prolongable : "The lease should available to prolong ["
```
denis-chudov, line 95:
```suggestion
        assert prolongable : "The lease should be available to prolong ["
```

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
```diff
@@ -355,4 +407,54 @@ private boolean isLeaseOutdated(Lease lease) {
...
+        private void processMessageInternal(String sender, PlacementDriverActorMessage msg) {
+            ReplicationGroupId grpId = msg.groupId();
+
+            Lease lease = leaseTracker.getLease(grpId);
+
+            if (msg instanceof StopLeaseProlongationMessage) {
+                if (lease.isProlongable() && sender.equals(lease.getLeaseholder().name())) {
+                    denyLease(grpId, lease).whenComplete((res, th) -> {
+                        if (th != null) {
+                            LOG.warn("Prolongation was not denied due to exception [groupId={}]", th, grpId);
```
denis-chudov, line 449:
```suggestion
                            LOG.warn("Prolongation denial failed due to exception [groupId={}]", th, grpId);
```

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
```diff
@@ -355,4 +407,54 @@ private boolean isLeaseOutdated(Lease lease) {
...
+                        } else {
+                            LOG.info("Lease deny prolonging message was handled [groupId={}, sender={}, deny={}]", grpId, sender, res);
```
denis-chudov, line 451:
```suggestion
                            LOG.info("Lease prolongation was denied [groupId={}, sender={}, deny={}]", grpId, sender, res);
```

denis-chudov (approved):

PR #2013, opened 2023-05-02
[POC] User object serialization

Pochatkin:
POC with the concept to pass custom objects to Compute API . The transmission mechanism has the following pipeline:
1. The object is serialized into byte[] on the client/server side
2. Resulted byte[] appeded to transferred Bianry Tuple with class name.
3. On server/client side byte[] is deserialized to platform spefic object. 
   3.1 For compute job args used special method from job declaration, to define class on server side, we avoid automatic class mapping from platform class to Java class here.
   3.2 For job result used follow mechanism. Java client is using class name from message for deserialing. C# client is using generic type to define C# target class.


modules/api/src/main/java/org/apache/ignite/sql/ColumnType.java
```diff
@@ -90,7 +90,10 @@ public enum ColumnType {
     NUMBER,
 
     /** Null. */
-    NULL;
+    NULL,
+
+    /** Binary. */
+    BINARY;
```
ptupitsyn, line 96: I don't think we should add a new SQL column type for this. Let's use existing `BYTE_ARRAY`.

modules/platforms/dotnet/Apache.Ignite/Internal/Proto/BinaryTuple/BinaryTupleReader.cs
```diff
@@ -412,6 +413,17 @@ public bool IsNull(int index)
             var s => throw GetInvalidLengthException(index, 12, s.Length)
         };
 
+        /// <summary>
+        /// Deserialize object from JSON.
+        /// </summary>
+        /// <param name="index">Index.</param>
+        /// <typeparam name="T">Value type.</typeparam>
+        /// <returns>Value.</returns>
+        public T? GetFromJson<T>(int index) => Seek(index) switch
```
ptupitsyn, line 422: This logic should not be a part of `BinaryTupleReader`.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/compute/ClientComputeExecuteRequest.java
```diff
@@ -66,7 +69,14 @@ public static CompletableFuture<Void> process(
      * @param in Unpacker.
      * @return Args array.
      */
-    static Object[] unpackArgs(ClientMessageUnpacker in) {
-        return in.unpackObjectArrayFromBinaryTuple();
+    static Object[] unpackArgs(ClientMessageUnpacker in, String jobClassName) {
```
ptupitsyn, line 72: Ideally we should not unpack arguments if we are not on the target node at the moment.

ptupitsyn (changes requested):

PR #2016, opened 2023-05-03
 IGNITE-19165: Extend ConfigurationVisitor's methods with Field

Flaugh24:


modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationRegistry.java
```diff
@@ -245,12 +246,12 @@ public <T> T represent(List<String> path, ConfigurationVisitor<T> visitor) throw
         }
 
         if (node instanceof TraversableTreeNode) {
-            return ((TraversableTreeNode) node).accept(null, visitor);
+            return ((TraversableTreeNode) node).accept(null, null, visitor);
```
ibessonov, line 249: I remember that we were discussing this place. "represent" must always pass a field, otherwise user may provide a path to password configuration and you won't obscure it.
`find` must return not only the object, but also a meta-information about it. This is the only way to do it. You also should add tests for such scenario.

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/asm/InnerNodeAsmGenerator.java
```diff
@@ -163,24 +166,30 @@ class InnerNodeAsmGenerator extends AbstractAsmGenerator {
...
+    private static final Method GET_DECLARED_FIELD_MTD;
+
     /** {@code Node#convert} method name. */
     private static final String CONVERT_MTD_NAME = "convert";
 
     /** {@link ConstructableTreeNode#construct(String, ConfigurationSource, boolean)} method name. */
     private static final String CONSTRUCT_MTD_NAME = "construct";
 
+    /** {@link Field} to {@link FieldDefinition} map. */
+    private final Map<Field, FieldDefinition> dieldToDieldDefinitionMap = new HashMap<>();
```
ibessonov, line 179: "dield"? Please fix typos in the name.
Also, the fact that there is a mapping from field to field definition is self-evident from the signature.
Please expand the comment and explain the meaning behind these definitions.

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/asm/InnerNodeAsmGenerator.java
```diff
@@ -306,6 +317,14 @@ private ClassDefinition createNodeClass() {
             }
         }
 
+        MethodDefinition classInitializer = innerNodeClassDef.getClassInitializer();
+        dieldToDieldDefinitionMap.forEach((k, v) -> {
+            // get declared field
```
ibessonov, line 322:
```suggestion
            // Get declared field.
```

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/asm/InnerNodeAsmGenerator.java
```diff
@@ -306,6 +317,14 @@ private ClassDefinition createNodeClass() {
...
+            BytecodeExpression invoke = constantClass(k.getDeclaringClass())
```
ibessonov, line 323: I would rename this "invoke" to "getDeclaredFieldExp" or something similar

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/asm/InnerNodeAsmGenerator.java
```diff
@@ -514,6 +533,11 @@ private FieldDefinition addInnerNodeField(Field schemaField) {
             throw new IllegalArgumentException("Unsupported field: " + schemaField);
         }
 
+        dieldToDieldDefinitionMap.put(
+                schemaField,
+                innerNodeClassDef.declareField(EnumSet.of(PUBLIC, STATIC, FINAL), fieldName + "FieldDefinition", Field.class)
```
ibessonov, line 538: `fieldName + "FieldDefinition"` - poor naming. What do you mean by field definition here? It looks more list "schema field". I would also make it upper case, because it's a `static final` field.

Doesn't it break "internal extensions"? They may have repeating properties names, don't we have such tests? Please check

ibessonov (changes requested):

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationRegistry.java
```diff
@@ -238,20 +239,21 @@ public <V, C, T extends ConfigurationTree<V, C>> T getConfiguration(RootKey<T, V
...
             node = ConfigurationUtil.find(path, superRoot, false);
         } catch (KeyNotFoundException e) {
             throw new IllegalArgumentException(e.getMessage());
         }
 
-        if (node instanceof TraversableTreeNode) {
-            return ((TraversableTreeNode) node).accept(null, null, visitor);
+        Object value = node.value();
+        if (value instanceof TraversableTreeNode) {
+            return ((TraversableTreeNode) value).accept(null, null, visitor);
```
ibessonov, line 251: Why don't you pass the field here? I thought we need it

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/asm/InnerNodeAsmGenerator.java
```diff
@@ -175,8 +176,8 @@ class InnerNodeAsmGenerator extends AbstractAsmGenerator {
     /** {@link ConstructableTreeNode#construct(String, ConfigurationSource, boolean)} method name. */
     private static final String CONSTRUCT_MTD_NAME = "construct";
 
-    /** {@link Field} to {@link FieldDefinition} map. */
-    private final Map<Field, FieldDefinition> dieldToDieldDefinitionMap = new HashMap<>();
+    /** Mapping for each configuration schema node to it's {@link FieldDefinition}. */
+    private final Map<Field, FieldDefinition> fieldToFieldDefinitionMap = new HashMap<>();
```
ibessonov, line 180: Still not clear, but I know that this class is not documented properly, so I can't ask for too much.
_"it's FieldDefinition"_ doesn't say what this definition means, and where it is. It's like saying _"maps field to its string"_ instead of _"... to its name"_.

By the way, **"its"**, not **"it's"**

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/asm/InnerNodeAsmGenerator.java
```diff
@@ -318,11 +319,11 @@ private ClassDefinition createNodeClass() {
         }
 
         MethodDefinition classInitializer = innerNodeClassDef.getClassInitializer();
-        dieldToDieldDefinitionMap.forEach((k, v) -> {
-            // get declared field
-            BytecodeExpression invoke = constantClass(k.getDeclaringClass())
+        fieldToFieldDefinitionMap.forEach((k, v) -> {
+            // Get declared field
```
ibessonov, line 323: You forgot the dot at the end.

modules/configuration/src/test/java/org/apache/ignite/internal/configuration/util/ConfigurationUtilTest.java
```diff
@@ -204,29 +204,94 @@ public void findSuccessfully() {
                 )
         );
 
+        NodeValue<Object> root = find(List.of(), parentNode, true);
+        assertNull(root.field());
```
ibessonov, line 208: Can we assert that it's actual field that we expect? We can check its name and declaring class, for example. NotNull is not enough

modules/configuration/src/test/java/org/apache/ignite/internal/configuration/util/ConfigurationUtilTest.java
```diff
@@ -204,29 +204,94 @@ public void findSuccessfully() {
...
+        assertNotNull(elementsNameChild.field());
+        assertSame(parentChange.elements().get("name").child(), elementsNameChild.value());
+
+        NodeValue<Object> elementsNameChildStr = find(List.of("elements", "name", "child", "str"), parentNode, true);
+        assertNotNull(elementsNameChildStr.field());
+        assertSame(parentChange.elements().get("name").child().str(), elementsNameChildStr.value());
+    }
+
+    @Test
+    public void findSuccessfullyPolymorphicConfig() {
```
ibessonov, line 229: We need a test where polymorphic extensions would have a sub-configuration with the same name, that is not primitive in both cases. I bet it wouldn't work.

modules/configuration/src/test/java/org/apache/ignite/internal/configuration/util/ConfigurationUtilTest.java
```diff
@@ -204,29 +204,94 @@ public void findSuccessfully() {
...
+
+            named.createOrUpdate("name1", sub -> {
+                sub.convert(SecondPolymorphicInstanceChange.class)
+                        .changeIntVal(1)
+                        .changeLongVal(1);
+            });
+        });
+
+        NodeValue<Object> subCfgStrVal = find(List.of("polymorphicSubCfg", "strVal"), parentNode, true);
+        assertSame(String.class, subCfgStrVal.field().getGenericType());
```
ibessonov, line 256: Is there a practical reason of using "getGenericType()" instead of "getType()"?

ibessonov (approved):
Looks good to me, thank you!

PR #2021, opened 2023-05-04
IGNITE-19394 "peek" contract improved to avoid missing data on RW scans.

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-19394


modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
```diff
@@ -1506,6 +1505,49 @@ void testScanPeekReplaceRow() {
...
+
+        PeekCursor<IndexRow> scan = indexStorage.scan(null, null, 0);
+
+        RowId rowId = new RowId(TEST_PARTITION);
+
+        // index  =  [0]
+        // cursor = ^ with no cached row
+        put(indexStorage, serializer.serializeRow(new Object[]{0}, rowId));
+
+        // index  =  [0] [1]
```
tkalkirill, line 1526: I don't think `[1]` is here.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/AbstractTestIndexStorage.java
```diff
@@ -30,6 +30,8 @@
  * Test-only abstract index storage class.
  */
 abstract class AbstractTestIndexStorage implements IndexStorage {
+    protected final int partitionId;
```
sashapolo, line 33: This field is only used in `TestSortedIndexStorage` maybe it should be moved there

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
```diff
@@ -61,8 +54,21 @@ public class TestSortedIndexStorage extends AbstractTestIndexStorage implements
     public TestSortedIndexStorage(int partitionId, SortedIndexDescriptor descriptor) {
         super(partitionId);
 
+        BinaryTupleComparator binaryTupleComparator = new BinaryTupleComparator(descriptor);
+
         this.descriptor = descriptor;
-        this.index = new ConcurrentSkipListMap<>(new BinaryTupleComparator(descriptor));
+        this.index = new ConcurrentSkipListSet<>((indexRow0, indexRow1) -> {
+            int cmp = binaryTupleComparator.compare(
```
sashapolo, line 61: Can be written shorter:
```
Comparator<IndexRow> comparator = Comparator.comparing((IndexRow row) -> row.indexColumns().byteBuffer(), binaryTupleComparator)
        .thenComparing(IndexRow::rowId);
```

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
```diff
@@ -118,23 +122,34 @@ public PeekCursor<IndexRow> scan(
...
-                navigableMap = index.subMap(lowerBound.byteBuffer(), upperBound.byteBuffer());
+                navigableSet = index.subSet(
+                        prefixToIndexRow(lowerBound, lowestRowId(partitionId)),
+                        true,
+                        prefixToIndexRow(upperBound, highestRowId(partitionId)),
+                        true
+                );
             } catch (IllegalArgumentException e) {
-                navigableMap = emptyNavigableMap();
+                navigableSet = emptyNavigableSet();
```
tkalkirill, line 142: Why is exception handling happening, what can go wrong here?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
```diff
@@ -175,112 +192,54 @@ public void close() {
...
-            }
-
-            if (rowId == null) {
-                if (currentEntry != null) {
-                    rowId = getRowId(currentEntry.getValue().firstEntry());
+            if (currentRow == null) {
+                try {
+                    peekedRow = indexSet.first();
+                } catch (NoSuchElementException e) {
+                    peekedRow = null;
```
tkalkirill, line 229: Exception logic is an anti-pattern, we can avoid them if we use a `NavigableMap`.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
ibessonov: That's right, I forgot to remove it, thank you!

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
```diff
@@ -61,8 +54,21 @@ public class TestSortedIndexStorage extends AbstractTestIndexStorage implements
     public TestSortedIndexStorage(int partitionId, SortedIndexDescriptor descriptor) {
         super(partitionId);
 
+        BinaryTupleComparator binaryTupleComparator = new BinaryTupleComparator(descriptor);
+
         this.descriptor = descriptor;
-        this.index = new ConcurrentSkipListMap<>(new BinaryTupleComparator(descriptor));
+        this.index = new ConcurrentSkipListSet<>((indexRow0, indexRow1) -> {
```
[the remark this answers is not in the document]
ibessonov: I'll make the code shorter instead

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
```diff
@@ -1506,6 +1505,49 @@ void testScanPeekReplaceRow() {
...
+        // cursor = ^ with no cached row (but it remembers the last peek call)
+        remove(indexStorage, serializer.serializeRow(new Object[]{0}, rowId));
+
+        // "hasNext" and "next" must return the result of last "peek" operation. This is crucial for RW scans.
+        assertTrue(scan.hasNext());
+        assertEquals(SimpleRow.of(0, rowId), SimpleRow.of(scan.next(), firstColumn(serializer)));
+
+        // index  =
+        // cursor = ^ with no cached row
+        assertNull(scan.peek());
```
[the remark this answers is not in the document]
ibessonov: I think it's already tested in your tests. New row will be returned

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
```diff
@@ -72,31 +78,29 @@ public SortedIndexDescriptor indexDescriptor() {
...
-        return index.getOrDefault(key.byteBuffer(), emptyNavigableMap()).keySet().iterator();
+        // These must be two different instances, because "scan" call messes up headers.
+        BinaryTuplePrefix lowerBound = BinaryTuplePrefix.fromBinaryTuple(key);
+        BinaryTuplePrefix higherBound = BinaryTuplePrefix.fromBinaryTuple(key);
+
+        //noinspection resource
+        return scan(lowerBound, higherBound, GREATER_OR_EQUAL | LESS_OR_EQUAL)
+                .stream()
+                .map(IndexRow::rowId)
+                .iterator();
```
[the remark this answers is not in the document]
ibessonov: This is definitely not tested. You're right, we need to fix it and cover it with tests.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
ibessonov: This was here before. Exception may happen if upper bound is below the lower bound.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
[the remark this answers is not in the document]
tkalkirill: I suggest adding this check so as not to assume.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
ibessonov: Why is this an anti-pattern, if I use Java API the way it's intended to be used?
I think that using Map instead of Set is anti-pattern

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/AbstractTestIndexStorage.java
ibessonov: Ok

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
tkalkirill: Please leave a comment.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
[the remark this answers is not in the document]
ibessonov: It is covered in `testScanPeekForFinishedCursor`. I don't want to duplicate test scenarios for no reason

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
tkalkirill: If you use a map, then it can return `null` instead of an exception, this is better and more performant.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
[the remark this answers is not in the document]
tkalkirill: Then why a new test, can you improve the current ones?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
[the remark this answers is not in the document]
ibessonov: By the way, this property is already violated in hash indexes

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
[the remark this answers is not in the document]
ibessonov: Current test is already too long. I test specific contract in this smaller test.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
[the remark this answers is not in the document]
tkalkirill: That's no reason not to fix it.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
```diff
@@ -83,10 +79,19 @@ Iterator<RowId> getRowIdIteratorForGetByBinaryTuple(BinaryTuple key) {
         BinaryTuplePrefix higherBound = BinaryTuplePrefix.fromBinaryTuple(key);
 
         //noinspection resource
-        return scan(lowerBound, higherBound, GREATER_OR_EQUAL | LESS_OR_EQUAL)
-                .stream()
-                .map(IndexRow::rowId)
-                .iterator();
+        PeekCursor<IndexRow> peekCursor = scan(lowerBound, higherBound, GREATER_OR_EQUAL | LESS_OR_EQUAL);
+
+        return new Iterator<>() {
```
tkalkirill, line 84: Why an inner class and not a method or a separate class.
You can use for example `org.apache.ignite.internal.util.CollectionUtils#viewReadOnly(java.util.Collection<? extends T1>, java.util.function.Function<? super T1,? extends T2>)`.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
ibessonov: `subSet` documentation is explicit about it, but ok, I'll add a comment.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
ibessonov: Ok, I'll use "viewReadOnly", I forgot about its existence.

ibessonov: Wait a minute, it uses a collection, not iterator.
I used anonymous class because it's small. It's the same approach that's used in `org.apache.ignite.internal.storage.index.impl.AbstractTestIndexStorage#get`, for example, do you have issues with that code?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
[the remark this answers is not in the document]
tkalkirill: I looked at `testScanPeekForFinishedCursor`, it checks my question.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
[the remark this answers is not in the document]
ibessonov: I fixed it actually. Should I add tests here or in a separate issue?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
tkalkirill: Exactly, nothing will stop us from doing a similar method but for iterators.

If this approach is used in `AbstractTestIndexStorage#get`, this does not mean that it is good, I think you should not foul the code with an anonymous class, but use a separate method / class.

ibessonov: Ok, I'll fix it then, not a big deal

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/TestSortedIndexStorage.java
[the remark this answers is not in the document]
ibessonov: That's too much work. I added a test and created issue to fix all storages: https://issues.apache.org/jira/browse/IGNITE-19422

sashapolo (approved):

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractSortedIndexStorageTest.java
```diff
@@ -215,12 +134,17 @@ protected SortedIndexStorage createIndexStorage(ColumnarIndexDefinition indexDef
...
+        return index.indexDescriptor();
+    }
+
     /**
      * Tests that columns of all types are correctly serialized and deserialized.
      */
     @Test
     void testRowSerialization() {
-        SortedIndexStorage indexStorage = createIndexStorage(ALL_TYPES_COLUMN_DEFINITIONS);
+        SortedIndexStorage indexStorage = createIndexStorage("TEST_IDX", ALL_TYPES_COLUMN_DEFINITIONS);
```
tkalkirill, line 147:
```suggestion
        SortedIndexStorage indexStorage = createIndexStorage(INDEX_NAME, ALL_TYPES_COLUMN_DEFINITIONS);
```

tkalkirill (approved):

PR #2029, opened 2023-05-05
IGNITE-18859 PendingComparableValuesTracker refactoring

sanpwc:
https://issues.apache.org/jira/browse/IGNITE-18859

1. PendingComparableValuesTracker was extended withPendingIndependentComparableValuesTracker in order to supportmultiple independent waiters instead of single one.
2. PendingComparableValuesTracker#waitFor() was updated in order to return gentrified version of CompletableFuture<R> instead of CompletableFuture<Void> in order to have an ability to associate arbitrary value with a waiter.
3. Related code refactoring PendingComparableValuesTracker<?, Void> in lots of places, tests updates.


modules/core/src/main/java/org/apache/ignite/internal/util/PendingComparableValuesTracker.java
```diff
@@ -102,6 +109,17 @@ public void update(T newValue) {
...
     }
 
+    /**
+     * Updates the internal state, if it is lower than {@code newValue} and completes all futures waiting for {@code newValue}
+     * that had been created for corresponding values that are lower than the given one.
+     *
+     * @param newValue New value.
+     * @throws TrackerClosedException if the tracker is closed.
+     */
+    public void update(T newValue) {
```
vldpyatkov, line 119: This method is looking risky, because you don't know exactly the type of value.

modules/core/src/main/java/org/apache/ignite/internal/util/PendingComparableValuesTracker.java
```diff
@@ -22,34 +22,39 @@
...
 
     /** Map of comparable values to corresponding futures. */
-    private final ConcurrentSkipListMap<T, CompletableFuture<Void>> valueFutures = new ConcurrentSkipListMap<>();
+    private final ConcurrentSkipListMap<T, CompletableFuture<R>> valueFutures = new ConcurrentSkipListMap<>();
 
-    /** Current value. */
-    private volatile T current;
+    /** Current value along with associated result. */
+    @SuppressWarnings("FieldMayBeFinal") // Changed through CURRENT VarHandle.
+    private volatile Map.Entry<T, @Nullable R> current;
```
vldpyatkov, line 57: As I understand, the current value is the only is used. Because if you try to get future for the previous T, you will receive the value of this variable.

modules/core/src/main/java/org/apache/ignite/internal/util/PendingComparableValuesTracker.java
```diff
@@ -161,8 +171,38 @@ public void close() {
...
 
         valueFutures.clear();
     }
+
+    protected void completeWaitersOnUpdate(T newValue, @Nullable R futureResult) {
+        ConcurrentNavigableMap<T, CompletableFuture<R>> smallerFutures = valueFutures.headMap(newValue, true);
+
+        smallerFutures.forEach((k, f) -> f.complete(futureResult));
+
+        smallerFutures.clear();
```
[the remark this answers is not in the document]
sanpwc: Because all that futures are guaranteed to be completed and there's no sense in having them any longer. BTW it's a part of the original code.

modules/core/src/main/java/org/apache/ignite/internal/util/PendingIndependentComparableValuesTracker.java
```diff
@@ -0,0 +1,84 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Tracker that stores comparable value internally, this value can grow when {@link #update(Comparable)} method is called. The tracker gives
+ * ability to wait for certain value, see {@link #waitFor(Comparable)}. In addition to {@link PendingComparableValuesTracker}
+ * given tracker supports multiple independent waiters for same value.
+ */
+public class PendingIndependentComparableValuesTracker<T extends Comparable<T>, R> extends PendingComparableValuesTracker<T, R> {
+    /** Map of comparable values to corresponding futures. */
+    private final ConcurrentSkipListMap<T, Set<CompletableFuture<R>>> valueFutures = new ConcurrentSkipListMap<>();
```
denis-chudov, line 35: I just think about how can we get rid of doubling this map in parent and derived classes. It is just confusing a bit.

sanpwc: I don't think that it's possible.

denis-chudov (approved):

PR #2030, opened 2023-05-05
IGNITE-19177 Add assignments checks when building indexes

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19177


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItBuildIndexTest.java
```diff
@@ -63,8 +70,20 @@ TABLE_NAME, toValuesString(List.of(1, 1), List.of(2, 2), List.of(3, 3), List.of(
...
+
+        for (Entry<Integer, List<Ignite>> entry : nodesWithBuiltIndexesByPartitionId.entrySet()) {
+            assertEquals(
+                    replicas,
+                    entry.getValue().size(),
+                    IgniteStringFormatter.format("p={}, nodes={}", entry.getKey(), entry.getValue())
+            );
+        }
 
+        // Let's check the sql query.
```
sashapolo, line 86: This comment is redundant

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -100,10 +111,14 @@ void stop() {
      */
     void startIndexBuild(TableIndexView tableIndexView, TableImpl table) {
         for (int partitionId = 0; partitionId < table.internalTable().partitions(); partitionId++) {
+            // Let's check if there is a node in the assignments for the partition.
```
sashapolo, line 114:
```suggestion
            // Check if this node is present in the assignments for the partition.
```

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -100,10 +111,14 @@ void stop() {
...
+            if (!replicaManager.startedGroups().contains(new TablePartitionId(table.tableId(), partitionId))) {
```
sashapolo, line 115: Why do we need to iterate over all partitions and check if `startedGroups` contains a particular one? Can we simply iterate over `startedGroups` and get partitions from there?

vldpyatkov: The checking based on started replica does not look like a permanent solution. Because the replica is a consequence of assignments.
If it is a temporary, that required TODO here.

tkalkirill: I used replicas on the advice of @sanpwc.

sanpwc: That's definitely not what I've meant. What about having build index logic inside replica itself? The basic question here is whether we have any index related logic if there's no replica locally? 
If not, we should neither check assignments nor replicas but rather perform index related logic inside replica. Of course it's still possible to used IndexBuilder as an abstraction that will cover corresponding index related logic.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItBuildIndexTest.java
tkalkirill: Fix it

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -382,8 +382,17 @@ public boolean stopReplica(ReplicationGroupId replicaGrpId) throws NodeStoppingE
      * @param replicaGrpId Replication group id.
      * @return True if the replica is found and closed, false otherwise.
      */
+    // TODO: IGNITE-19494 We need to correctly stop the replica
     private boolean stopReplicaInternal(ReplicationGroupId replicaGrpId) {
-        return replicas.remove(replicaGrpId) != null;
+        CompletableFuture<Replica> removed = replicas.remove(replicaGrpId);
+
+        if (removed != null) {
+            removed.whenComplete((replica, throwable) -> replica.shutdown());
```
sashapolo, line 390: You need to check that `throwable` is `null`

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -382,8 +382,17 @@ public boolean stopReplica(ReplicationGroupId replicaGrpId) throws NodeStoppingE
      * @param replicaGrpId Replication group id.
      * @return True if the replica is found and closed, false otherwise.
      */
+    // TODO: IGNITE-19494 We need to correctly stop the replica
```
sashapolo, line 385: The ticket that is mentioned here has a terrible description, sorry

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ClusterPerClassIntegrationTest.java
```diff
@@ -515,14 +513,20 @@ public static void waitForIndexBuild(String tableName, String indexName) throws
                     continue;
                 }
 
+                TablesView tablesView = getTablesConfiguration(clusterNode).value();
+
                 IndexStorage index = internalTable.storage().getOrCreateIndex(
                         partitionId,
-                        createIndexDescription(getTablesConfiguration(clusterNode).value(), indexName)
+                        createIndexDescriptor(tablesView, tablesView.indexes().get(indexName.toUpperCase()).id())
```
sashapolo, line 520: please extract `tablesView.indexes().get(indexName.toUpperCase()).id()` into a variable

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
```diff
@@ -417,4 +417,12 @@ public void buildIndex(UUID indexId, List<UUID> rowUuids, boolean finish) {
     public PendingComparableValuesTracker<HybridTimestamp, Void> getSafeTimeTracker() {
         return safeTimeTracker;
     }
+
+    /**
+     * Waits for indexes to be created.
+     */
+    // TODO: IGNITE-19513 Fix it, we should have already waited for the indexes to be created
```
sashapolo, line 424: Again, the referred ticket description is basically nonexistent

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+        inBusyLock(() -> {
+            if (indexStorage.getNextRowIdToBuild() == null) {
+                return;
+            }
+
+            IndexBuildTaskId taskId = new IndexBuildTaskId(tableId, partitionId, indexId);
+
+            IndexBuildTask newTask = new IndexBuildTask(taskId, indexStorage, partitionStorage, raftClient, executor, busyLock, BATCH_SIZE);
+
+            IndexBuildTask previousTask = indexBuildTaskById.putIfAbsent(taskId, newTask);
```
sashapolo, line 106: you should use `computeIfAbsent` in order to avoid creating unnecessary task instances

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+    }
+
+    @Override
+    public void close() {
+        if (!closeGuard.compareAndSet(false, true)) {
+            return;
+        }
+
+        busyLock.block();
```
sashapolo, line 170: Do we need to stop all index building tasks?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * Stops building all indexes (for a table partition) if they are in progress.
+     *
+     * @param tableId Table ID.
+     * @param partitionId Partition ID.
+     */
+    public void stopBuildIndexes(UUID tableId, int partitionId) {
+        for (Iterator<Entry<IndexBuildTaskId, IndexBuildTask>> it = indexBuildTaskById.entrySet().iterator(); it.hasNext(); ) {
+            if (!busyLock.enterBusy()) {
```
sashapolo, line 144: Why do we take the lock inside the loop?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTaskId.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private final UUID indexId;
+
+    IndexBuildTaskId(UUID tableId, int partitionId, UUID indexId) {
+        this.tableId = tableId;
+        this.partitionId = partitionId;
+        this.indexId = indexId;
+    }
+
+    public UUID getTableId() {
```
sashapolo, line 39: we don't use `get` prefixes

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccessImpl.java
```diff
@@ -192,6 +192,8 @@ public long maxLastAppliedTerm() {
 
     @Override
     public CompletableFuture<Void> startRebalance() {
+        storageUpdateHandler.waitIndexes();
```
sashapolo, line 195: Please add a clarifying comment, why this line is here

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTask.java
```diff
@@ -0,0 +1,223 @@
...
+        this.busyLock = busyLock;
+        this.batchSize = batchSize;
+    }
+
+    /**
+     * Starts building the index.
+     */
+    void start() {
+        if (!enterBusy()) {
+            taskFuture.complete(null);
```
sashapolo, line 93: why not `completeExceptionally(NodeStoppingException)`?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTask.java
```diff
@@ -0,0 +1,223 @@
...
+
+            return;
+        }
+
+        LOG.info("Start building the index: [{}]", getCommonIndexInfo());
```
sashapolo, line 98: `getCommonIndexInfo` is not a free operation, you should wrap this logging with `isInfoEnabled`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTask.java
```diff
@@ -0,0 +1,223 @@
...
+
+        try {
+            supplyAsync(this::handleNextBatch, executor)
+                    .thenCompose(Function.identity())
+                    .whenComplete((unused, throwable) -> {
+                        if (throwable != null) {
+                            LOG.error("Index build error: [{}]", throwable, getCommonIndexInfo());
```
sashapolo, line 105: Same here

sashapolo: Also, the format doesn't look correct

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTask.java
```diff
@@ -0,0 +1,223 @@
...
+        if (!enterBusy()) {
+            taskFuture.complete(null);
+
+            return;
+        }
+
+        LOG.info("Start building the index: [{}]", getCommonIndexInfo());
+
+        try {
+            supplyAsync(this::handleNextBatch, executor)
```
sashapolo, line 101: Why do you call `supplyAsync` here? So that `createBatchRowIds` are computed on the executor? Can we do that explicitly instead?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTask.java
```diff
@@ -0,0 +1,223 @@
...
+                .finish(finish)
+                .build();
+    }
+
+    private boolean enterBusy() {
+        if (!taskBusyLock.enterBusy()) {
+            return false;
+        }
+
+        if (!busyLock.enterBusy()) {
```
sashapolo, line 203: Wouldn't it be easier to check the `busyLock` first and then the `taskBusyLock`?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
tkalkirill: Fix it

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ClusterPerClassIntegrationTest.java
tkalkirill: fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuilder.java
tkalkirill: think that your suggestion does not quite suit me, because then the code for starting the task and subscribing to the future will be inside the closure, which can lead to a lengthening of the closure execution, although it should be fast, and the deletion of the task can immediately be performed in this closure and we can freeze.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuilder.java
tkalkirill: If the node stops then we must stop building all indexes.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuilder.java
tkalkirill: This is an optimization, if we stop the node, then it no longer makes sense to stop building the partition indexes.

If you don't like it, I can remove this optimization.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTaskId.java
tkalkirill: This rule is valid for 2.0 in 3.0 we don't have it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionAccessImpl.java
tkalkirill: Fix it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTask.java
tkalkirill: I'm not sure what is appropriate, since we can stop building indexes both at node shutdown and, for example, after a change assignments.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTask.java
tkalkirill: fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTask.java
tkalkirill: The error level has no such check, what's wrong with the format?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTask.java
tkalkirill: I would not like to load a thread that will start building the index, so I think it's right to do it in a separate pool.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTask.java
tkalkirill: It doesn't matter to me - I'll change it.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
tkalkirill: I tried to paraphrase, but I don’t know what else to add, do you advise something?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
tkalkirill: Try to fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTask.java
sashapolo: your version is correct then

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTask.java
sashapolo: > The error level has no such check

indeed, forgot about that

> what's wrong with the format

originally I thought you would need another `{}` for the throwable, but that's not true, so everything is correct

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTask.java
sashapolo: I agree that it is right, I'm proposing to do `runAsync(createBatchRowIds)` inside the `handleNextBatch` method, not here. This will also help to get rid of `Funciton.identity` below

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuilder.java
sashapolo: ok, I agree

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
sashapolo: it's better now

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuilder.java
tkalkirill: No, I meant that index building tasks will eventually be stopped because this busyLock is passed to all tasks.

sashapolo: Shall we still clear the map?

tkalkirill: Yes, when this **busyLock** is blocked, the tasks will complete their work and be removed from the map on their own.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/index/IndexBuildTask.java
tkalkirill: In personal correspondence, we decided to leave it as it is.

sashapolo (approved):

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
sanpwc: Instead of waiting we should complete the future exceptionally in order to interrupt replica await process. Semantically, only if it's not completed already, but who cares completeExceptionally will do the trick. We may introduce ReplicaStoppingException or similar and verify that such kind of exception is properly handled in `ReplicaService#sendToReplica`.
@vldpyatkov @sergeyuttsel Guys you've touched replica await code recently, do you have any objections?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/listener/ReplicaListener.java
```diff
@@ -31,4 +33,20 @@ public interface ReplicaListener {
      * @return Listener response.
      */
     CompletableFuture<?> invoke(ReplicaRequest request);
+
+    /**
+     * Callback on leader election.
+     *
+     * @param clusterNode Leader node.
+     */
+    default void onLeaderElected(ClusterNode clusterNode) {
```
sanpwc, line 42: We may have leaderless replication protocols. I'd say that we should use onBecomePrimary instead.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
tkalkirill: I have already created IGNITE-19494 in which it will be possible to implement any behavior, I propose this within wdyt?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/listener/ReplicaListener.java
tkalkirill: fix it

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
sergeyuttsel: I have no objections. I agree that we should complete the future exceptionally in order to interrupt replica await process in ReplicaService#sendToReplica.

sanpwc: Yep. We(TX teem) will do this.

sanpwc (approved):

PR #2050, opened 2023-05-10
IGNITE-19439 Fixed bug in VersionChainIo that lead to tree corruption

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-19439


ibessonov:
https://github.com/apache/ignite-3/pull/2054

PR #2055, opened 2023-05-10
IGNITE-19228 Schema validation during tx processing: common framework

rpuch:
https://issues.apache.org/jira/browse/IGNITE-19228

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/CollectionUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/CollectionUtils.java
@@ -22,6 +22,7 @@ import static java.util.Collections.emptyIterator;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.unmodifiableCollection;
 import static java.util.Collections.unmodifiableSet;
+import static java.util.stream.Collectors.toSet;
 
 import it.unimi.dsi.fastutil.ints.IntOpenHashSet;
 import it.unimi.dsi.fastutil.ints.IntSet;
@@ -595,5 +596,16 @@ public final class CollectionUtils {
      */
     public static IntSet setOf(Collection<Integer> coll) {
         return IntSets.unmodifiable(new IntOpenHashSet(coll));
+    }
+
+    /**
+     * Returns an intersection of two sets.
+     *
+     * @param op1 First operand.
+     * @param op2 Second operand.
+     * @return Result of the intersection.
+     */
+    public static <T> Set<T> intersect(Set<T> op1, Set<T> op2) {
+        return op1.stream().filter(op2::contains).collect(toSet());
     }
 }
--- a/modules/core/src/test/java/org/apache/ignite/internal/util/CollectionUtilsTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/util/CollectionUtilsTest.java
@@ -365,5 +365,6 @@ public class CollectionUtilsTest {
     @Test
     void testIntersect() {
         assertThat(intersect(Set.of(1, 2), Set.of(2, 3)), is(Set.of(2)));
+        assertThat(intersect(Set.of(1, 2), Set.of(3, 4)), is(Set.of()));
     }
 }
--- a/modules/core/src/test/java/org/apache/ignite/internal/util/CollectionUtilsTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/util/CollectionUtilsTest.java
@@ -22,11 +22,13 @@ import static java.util.function.Function.identity;
 import static java.util.stream.Collectors.toList;
 import static org.apache.ignite.internal.util.CollectionUtils.concat;
 import static org.apache.ignite.internal.util.CollectionUtils.difference;
+import static org.apache.ignite.internal.util.CollectionUtils.intersect;
 import static org.apache.ignite.internal.util.CollectionUtils.setOf;
 import static org.apache.ignite.internal.util.CollectionUtils.union;
 import static org.apache.ignite.internal.util.CollectionUtils.viewReadOnly;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
@@ -358,5 +360,10 @@ public class CollectionUtilsTest {
 
         assertThrows(UnsupportedOperationException.class, () -> copy.add(42));
         assertThrows(UnsupportedOperationException.class, () -> copy.remove(3));
+    }
+
+    @Test
+    void testIntersect() {
+        assertThat(intersect(Set.of(1, 2), Set.of(2, 3)), is(Set.of(2)));
     }
 }
--- a/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/asserts/CompletableFutureAssert.java
+++ b/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/asserts/CompletableFutureAssert.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.testframework.asserts;
+
+import static org.junit.jupiter.api.Assertions.fail;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Assertions related to {@link CompletableFuture}.
+ */
+public class CompletableFutureAssert {
+    /**
+     * Asserts that the given future completes with an exception being an instance of the given class and returns
+     * that exception for further examination.
+     *
+     * <p>Unlike {@link org.apache.ignite.internal.testframework.matchers.CompletableFutureExceptionMatcher#willThrowFast(Class)},
+     * this method allows to examine the actual exception thrown further in the test.
+     *
+     * @param future Future to work on.
+     * @param expectedExceptionClass Expected class of the exception.
+     * @return Matched exception.
+     * @param <X> Exception type.
+     */
+    public static <X extends Throwable> X assertWillThrowFast(
+            CompletableFuture<?> future,
+            Class<X> expectedExceptionClass
+    ) {
+        try {
+            Object normalResult = future.get(1, TimeUnit.SECONDS);
+
+            return fail("Expected the future to be completed with an exception of class " + expectedExceptionClass
+                    + ", but it completed normally with result " + normalResult);
+        } catch (TimeoutException e) {
+            return fail("Expected the future to be completed with an exception of class in 1 second, but it did not complete in time");
+        } catch (Throwable e) {
+            Throwable unwrapped = unwrapThrowable(e);
+
+            if (expectedExceptionClass.isInstance(unwrapped)) {
+                return (X) unwrapped;
+            } else {
+                return fail(
+                        "Expected the future to be completed with an exception of class " + expectedExceptionClass.getName() + ", but got "
+                                + e.getClass().getName(),
+                        e
+                );
+            }
+        }
+    }
+
+    private static Throwable unwrapThrowable(Throwable e) {
+        while (e instanceof ExecutionException || e instanceof CompletionException) {
+            e = e.getCause();
+        }
+
+        return e;
+    }
+}
--- a/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/asserts/CompletableFutureAssert.java
+++ b/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/asserts/CompletableFutureAssert.java
@@ -38,8 +38,8 @@ public class CompletableFutureAssert {
      *
      * @param future Future to work on.
      * @param expectedExceptionClass Expected class of the exception.
-     * @return Matched exception.
      * @param <X> Exception type.
+     * @return Matched exception.
      */
     public static <X extends Throwable> X assertWillThrowFast(
             CompletableFuture<?> future,
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
@@ -24,7 +24,6 @@ import static org.apache.ignite.internal.metastorage.dsl.Operations.ops;
 import static org.apache.ignite.internal.metastorage.dsl.Operations.put;
 import static org.apache.ignite.internal.metastorage.dsl.Operations.remove;
 import static org.apache.ignite.internal.metastorage.dsl.Statements.iif;
-import static org.apache.ignite.internal.utils.RebalanceUtil.intersect;
 import static org.apache.ignite.internal.utils.RebalanceUtil.pendingPartAssignmentsKey;
 import static org.apache.ignite.internal.utils.RebalanceUtil.plannedPartAssignmentsKey;
 import static org.apache.ignite.internal.utils.RebalanceUtil.stablePartAssignmentsKey;
@@ -62,6 +61,7 @@ import org.apache.ignite.internal.schema.configuration.ExtendedTableChange;
 import org.apache.ignite.internal.schema.configuration.TableConfiguration;
 import org.apache.ignite.internal.table.distributed.PartitionMover;
 import org.apache.ignite.internal.util.ByteUtils;
+import org.apache.ignite.internal.util.CollectionUtils;
 import org.apache.ignite.internal.util.IgniteSpinBusyLock;
 import org.apache.ignite.lang.ByteArray;
 
@@ -347,12 +347,12 @@ public class RebalanceRaftGroupEventsListener implements RaftGroupEventsListener
             // For further addition
             Set<Assignment> calculatedSwitchAppend = union(retrievedSwitchAppend, reducedNodes);
             calculatedSwitchAppend = subtract(calculatedSwitchAppend, addedNodes);
-            calculatedSwitchAppend = intersect(calculatedAssignments, calculatedSwitchAppend);
+            calculatedSwitchAppend = CollectionUtils.intersect(calculatedAssignments, calculatedSwitchAppend);
 
             Set<Assignment> calculatedPendingReduction = subtract(stable, retrievedSwitchReduce);
 
             Set<Assignment> calculatedPendingAddition = union(stable, reducedNodes);
-            calculatedPendingAddition = intersect(calculatedAssignments, calculatedPendingAddition);
+            calculatedPendingAddition = CollectionUtils.intersect(calculatedAssignments, calculatedPendingAddition);
 
             // eq(revision(assignments.stable), retrievedAssignmentsStable.revision)
             SimpleCondition con1 = stableEntry.empty()
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
@@ -24,6 +24,7 @@ import static org.apache.ignite.internal.metastorage.dsl.Operations.ops;
 import static org.apache.ignite.internal.metastorage.dsl.Operations.put;
 import static org.apache.ignite.internal.metastorage.dsl.Operations.remove;
 import static org.apache.ignite.internal.metastorage.dsl.Statements.iif;
+import static org.apache.ignite.internal.util.CollectionUtils.intersect;
 import static org.apache.ignite.internal.utils.RebalanceUtil.pendingPartAssignmentsKey;
 import static org.apache.ignite.internal.utils.RebalanceUtil.plannedPartAssignmentsKey;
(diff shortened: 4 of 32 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/SchemaCompatValidator.java
```diff
@@ -0,0 +1,119 @@
...
+            boolean commit,
+            @Nullable HybridTimestamp commitTimestamp
+    ) {
+        if (!commit) {
+            return completedFuture(ForwardValidationResult.success());
+        }
+
+        HybridTimestamp beginTimestamp = TransactionIds.beginTimestamp(txId);
+
+        List<UUID> tableIds = enlistedGroupIds.stream()
```
sashapolo, line 69: I think this intermediate variable is redundant, you can use the stream directly inside `thenApply` below

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/NonHistoricSchemas.java
```diff
@@ -0,0 +1,160 @@
...
+
+    @Override
+    public List<FullTableSchema> tableSchemaVersionsBetween(UUID tableId, HybridTimestamp fromIncluding, HybridTimestamp toIncluding) {
+        SchemaRegistry schemaRegistry = schemaManager.schemaRegistry(tableId);
+
+        SchemaDescriptor schemaDescriptor = schemaRegistry.schema();
+        FullTableSchema fullSchema = new FullTableSchema(
+                1,
+                1,
+                schemaDescriptor.columnNames().stream()
```
sashapolo, line 72: Please extract this into a variable

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/NonHistoricSchemas.java
```diff
@@ -0,0 +1,160 @@
...
+        return getColumnType(type.spec());
+    }
+
+    /**
+     * Detects {@link ColumnType} by {@link NativeTypeSpec}.
+     *
+     * @param spec Native type spec.
+     * @return Detected {@link ColumnType}.
+     */
+    public static ColumnType getColumnType(NativeTypeSpec spec) {
```
sashapolo, line 106: There exists an exact copy of this method in `ClientTableCommon`, I would suggest to extract it into a common place

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/NonHistoricSchemas.java
```diff
@@ -0,0 +1,160 @@
...
+                return ColumnType.TIME;
+
+            case DATETIME:
+                return ColumnType.DATETIME;
+
+            case TIMESTAMP:
+                return ColumnType.TIMESTAMP;
+
+            default:
+                throw new IgniteException(PROTOCOL_ERR, "Unsupported native type: " + spec);
```
sashapolo, line 157: Wait a second, you indeed copy-pasted it from `ClientTableCommon`, shame on you!

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/NonHistoricSchemas.java
```diff
@@ -0,0 +1,160 @@
...
+    private static TableColumnDescriptor columnType(String colName, Column column) {
+        return new TableColumnDescriptor(
+                colName,
+                columnTypeFromNativeType(column.type()),
+                column.nullable(),
+                DefaultValue.constant(column.defaultValue())
+        );
+    }
+
+    private static ColumnType columnTypeFromNativeType(NativeType type) {
```
sashapolo, line 96: This method seems redundant to me

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/NonHistoricSchemas.java
```diff
@@ -0,0 +1,160 @@
...
+                            return columnType(colName, column);
+                        })
+                        .collect(toList()),
+                List.of()
+        );
+
+        return List.of(fullSchema);
+    }
+
+    private static TableColumnDescriptor columnType(String colName, Column column) {
```
sashapolo, line 87: Should be called `columnDescriptor`

sashapolo, line 87: `colName` parameter is redundant, `column` already contains a name

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/NonHistoricSchemas.java
```diff
@@ -0,0 +1,160 @@
...
+        return new TableColumnDescriptor(
+                colName,
+                columnTypeFromNativeType(column.type()),
+                column.nullable(),
+                DefaultValue.constant(column.defaultValue())
```
sashapolo, line 92: Are sure this is correct? There are multiple types of default values

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/replication/PartitionReplicaListenerTest.java
```diff
@@ -290,6 +314,8 @@ public void beforeTest(@InjectConfiguration DataStorageConfiguration dsCfg) {
 
         lenient().when(safeTimeClock.waitFor(any())).thenReturn(completedFuture(null));
 
+        lenient().when(schemas.waitForSchemasAvailability(any())).thenReturn(completedFuture(null));
```
sashapolo, line 317: I know that it's not related to this PR, but all these `lenient()` calls can be removed using the `MockitoSettings` annotation

modules/table/src/testFixtures/java/org/apache/ignite/internal/table/impl/DummySchemas.java
```diff
@@ -0,0 +1,86 @@
...
+        return completedFuture(null);
+    }
+
+    @Override
+    public List<FullTableSchema> tableSchemaVersionsBetween(UUID tableId, HybridTimestamp fromIncluding, HybridTimestamp toIncluding) {
+        SchemaDescriptor schemaDescriptor = schemaRegistry.schema();
+        FullTableSchema fullSchema = new FullTableSchema(
+                1,
+                1,
+                schemaDescriptor.columnNames().stream()
```
sashapolo, line 59: Please extract this into a variable

modules/table/src/testFixtures/java/org/apache/ignite/internal/table/impl/DummySchemas.java
```diff
@@ -0,0 +1,86 @@
...
+        return new TableColumnDescriptor(
+                colName,
+                columnTypeFromNativeType(column.type()),
+                column.nullable(),
+                DefaultValue.constant(column.defaultValue())
+        );
+    }
+
+    private static ColumnType columnTypeFromNativeType(NativeType type) {
+        return NonHistoricSchemas.getColumnType(type.spec());
```
sashapolo, line 84: Why do you reuse this method from `NonHistoricSchemas` and not the method that creates a `TableColumnDescriptor` from a `Column`?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -976,13 +983,34 @@ private CompletableFuture<Void> processTxFinishAction(TxFinishReplicaRequest req
         List<TablePartitionId> aggregatedGroupIds = request.groups().values().stream()
                 .flatMap(List::stream)
                 .map(IgniteBiTuple::get1)
-                .collect(Collectors.toList());
+                .collect(toList());
 
         UUID txId = request.txId();
 
-        boolean commit = request.commit();
+        return schemaCompatValidator.validateForwards(txId, aggregatedGroupIds, request.commit(), request.commitTimestamp())
```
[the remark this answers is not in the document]
rpuch: I see two alternatives to the current approach.

One is to branch on `request.commit()` in the `PartitionReplicaListener` and, if it's a commit, do what is done there now (validate, then invoke `finishAndCleanup()`, then maybe throw); if it's an abort, just call `finishAndCleanup()`. But the result is that `finishAndCleanup()` is called twice in different branches, and this seems weird.

Another alternative is to build a method like `validateForwardsIfCommit()`, but it would still produce a future of `ForwardValidationResult`, so we would have to return `success()` if it's an abort. Still looks weird: we kinda did not validate, but we returned a validation result. We could also add a third state to the validation result, like 'no need for validation', but it would work exactly like 'success', so this also looks weird.

What changes do you envision here?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -976,13 +983,34 @@ private CompletableFuture<Void> processTxFinishAction(TxFinishReplicaRequest req
...
+                .thenCompose(validationResult -> {
+                    boolean stillCommit = request.commit() && validationResult.isOk();
 
-        CompletableFuture<Object> changeStateFuture = finishTransaction(aggregatedGroupIds, txId, commit);
+                    return finishAndCleanup(request, stillCommit, aggregatedGroupIds, txId)
+                            .thenCompose(result -> {
```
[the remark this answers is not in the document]
rpuch: I return a future to signal an exception, so `thenAccept()` does not fit. Do you suggest throwing the exception?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -976,13 +983,34 @@ private CompletableFuture<Void> processTxFinishAction(TxFinishReplicaRequest req
...
+                                if (validationResult.isOk()) {
+                                    return completedFuture(result);
+                                } else {
+                                    return failedFuture(new AbortDueToIncompatibleSchemaException("Commit failed because schema "
```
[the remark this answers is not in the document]
rpuch: The idea is that, if validation fails, we change the intention from commit to abort, so finishTransaction() will actually abort the transaction, and only then will we signal the exception

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/SchemaCompatValidator.java
rpuch: It looks too cumbersome inside `for`. Introducing a variable allows to easier read the `for` loop, imho.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/replication/PartitionReplicaListenerTest.java
rpuch: I'm not sure it's a good thing to just make all stubbings lenient. I would prefer to be explicit.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/NonHistoricSchemas.java
rpuch: I'm sorry :(

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/asserts/CompletableFutureAssert.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Assertions related to {@link CompletableFuture}.
+ */
+public class CompletableFutureAssert {
```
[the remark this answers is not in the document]
sashapolo: Can we add a `Matcher` and an overload that allows us to inspect the exception? This way we can write something like `assertThat(f, willThrow(e -> ...)`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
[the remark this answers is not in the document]
sashapolo: yes

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/NonHistoricSchemas.java
sashapolo: Because looks like there is also a functional default value type, will it work in this case?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/RebalanceRaftGroupEventsListener.java
```diff
@@ -347,12 +347,12 @@ private void doOnNewPeersConfigurationApplied(PeersAndLearners configuration) {
             // For further addition
             Set<Assignment> calculatedSwitchAppend = union(retrievedSwitchAppend, reducedNodes);
             calculatedSwitchAppend = subtract(calculatedSwitchAppend, addedNodes);
-            calculatedSwitchAppend = intersect(calculatedAssignments, calculatedSwitchAppend);
+            calculatedSwitchAppend = CollectionUtils.intersect(calculatedAssignments, calculatedSwitchAppend);
```
sashapolo, line 350: I propose to add a static import here, looks strange otherwise

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/FullTableSchema.java
```diff
@@ -125,34 +117,24 @@ public TableDefinitionDiff diffFrom(FullTableSchema prevSchema) {
...
+    private static <T> Map<String, T> toMapByName(List<T> elements, Function<T, String> nameExtractor) {
+        return elements.stream().collect(toMap(nameExtractor, identity()));
     }
 
-    private static Set<String> intersect(Set<String> set1, Set<String> set2) {
-        Set<String> result = new HashSet<>(set1);
-        result.retainAll(set2);
-        return result;
+    private static <T> List<T> subtractKeyed(Map<String, T> diminuend, Map<String, T> subtrahend) {
+        return diminuend.keySet().stream()
```
sashapolo, line 134: You can use `entrySet` here to avoid calling `get` afterwards

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/FullTableSchema.java
```diff
@@ -125,34 +117,24 @@ public TableDefinitionDiff diffFrom(FullTableSchema prevSchema) {
...
-        return result;
+    private static <T> Map<String, T> toMapByName(List<T> elements, Function<T, String> nameExtractor) {
+        return elements.stream().collect(toMap(nameExtractor, identity()));
     }
 
-    private static Set<String> intersect(Set<String> set1, Set<String> set2) {
-        Set<String> result = new HashSet<>(set1);
-        result.retainAll(set2);
-        return result;
+    private static <T> List<T> subtractKeyed(Map<String, T> diminuend, Map<String, T> subtrahend) {
```
sashapolo, line 133: `diminuend`? What does that mean?

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
```diff
@@ -468,7 +468,7 @@ public static CompletableFuture<Void> handleReduceChanged(MetaStorageManager met
      * @return Result of the subtraction.
      */
     public static <T> Set<T> subtract(Set<T> minuend, Set<T> subtrahend) {
```
sashapolo, line 470: Can we get rid of this method?

modules/core/src/main/java/org/apache/ignite/internal/util/CollectionUtils.java
```diff
@@ -596,4 +597,15 @@ public static <T> Set<T> difference(@Nullable Set<T> a, @Nullable Set<T> b) {
...
     }
+
+    /**
+     * Returns an intersection of two sets.
+     *
+     * @param op1 First operand.
+     * @param op2 Second operand.
+     * @return Result of the intersection.
+     */
+    public static <T> Set<T> intersect(Set<T> op1, Set<T> op2) {
```
sashapolo, line 608: No subtrahend =(

modules/core/src/test/java/org/apache/ignite/internal/util/CollectionUtilsTest.java
```diff
@@ -359,4 +361,9 @@ private void testSetOf(Collection<Integer> data) {
         assertThrows(UnsupportedOperationException.class, () -> copy.add(42));
         assertThrows(UnsupportedOperationException.class, () -> copy.remove(3));
     }
+
+    @Test
+    void testIntersect() {
```
sashapolo, line 366: Let's add a test for en empty intersection as well

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/replication/PartitionReplicaListenerTest.java
sashapolo: ok

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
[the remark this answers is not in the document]
sashapolo: > But the result is that finishAndCleanup() is called twice in different branches, and this seems weird.

I don't know but it looks perfectly fine to me:

```
if (request.commit()) {
    return schemaCompatValidator.validateForwards(txId, aggregatedGroupIds, request.commitTimestamp())
            .thenCompose(validationResult -> {
                return finishAndCleanup(request, validationResult.isSuccessful(), aggregatedGroupIds, txId)
                        .thenCompose(result -> failureIfValidationFailed(validationResult));
            });
} else {
    return finishAndCleanup(request, false, aggregatedGroupIds, txId);
}
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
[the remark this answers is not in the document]
sashapolo: oops, I missed the `stillCommit` part

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/SchemaCompatValidator.java
sashapolo: I think it looks ok if extracted into a method:
```
ForwardValidationResult validateTablesSchemas(
        List<TablePartitionId> enlistedGroupIds,
        HybridTimestamp beginTimestamp,
        HybridTimestamp commitTimestamp
) {
    return enlistedGroupIds.stream()
            .map(TablePartitionId::tableId)
            .distinct()
            .map(tableId -> validateSchemaCompatibility(beginTimestamp, commitTimestamp, tableId))
            .filter(validationResult -> !validationResult.isSuccessful())
            .findAny()
            .orElse(ForwardValidationResult.success());
}
```

sashapolo: Or, if you still like your `for` cycle, I think it's better to use a `Set` instead of `distinct().toList()`

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/asserts/CompletableFutureAssert.java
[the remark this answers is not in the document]
rpuch: A matcher gives a binary answer (matches/does not match) and a description of what does not match. This is a pretty rigid format if you want to make a few assertions. On the other hand, if we just add assertions (throwing an AssertionError) 'inside' a matcher, this looks like an abuse of the idea of a matcher.

Each tool is good for its situation. Sometimes a matcher is the right tool, sometimes direct inspection is better. It seems correct to have both tools and use them accordingly. Moreover, this does not bloat the code too much, the class we are discussing is rather small and simple.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/NonHistoricSchemas.java
rpuch: This is a temporary code that will be thrown away when we finish the switch. The column definition that is produced is actually 'fake', it's not used anywhere because this class is only used for schema validation, and for now it always returns just one schema, so no actual column validation happens.

Also, it's impossible to directly convert `Column`'s functional default value provider (which is a `DefaultValueGenerator`) to `TableColumnDefinition`'s functional default value (which is a string - just a function name).

I amended the comment to reflect that the conversion is approximate.

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
rpuch: Not in this PR probably

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/NonHistoricSchemas.java
sashapolo: got it

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
sashapolo: why? Is it used a lot?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/SchemaCompatValidator.java
rpuch: Comparing the two options, I still prefer the `for` loop as the one with the stream makes you think a bit harder to understand it. But I like the ideas of using `Set` and extracting the method.

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/asserts/CompletableFutureAssert.java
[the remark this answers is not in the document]
sashapolo: ok, my suggestion was a little bit too eager and does not cover your case, I agree. I just don't like this logic duplication between two classes, but ok, let's keep this class(

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/SchemaCompatValidator.java
sashapolo: ok

modules/table/src/main/java/org/apache/ignite/internal/utils/RebalanceUtil.java
rpuch: You are right, it's not a big deal. Removed it.

sashapolo (approved):

PR #2056, opened 2023-05-10
IGNITE-15507: correct type mapping for JDBC metadata.

ygerzhedovich:


lowka (approved):

modules/api/src/main/java/org/apache/ignite/sql/ColumnType.java
```diff
@@ -164,4 +164,23 @@ public static Class<?> columnTypeToClass(ColumnType type) {
                 throw new IllegalArgumentException("Unsupported type " + type);
         }
     }
+
+    /**
+     * Column type to Java class for JDBC.
+     */
+    public static Class<?> columnTypeToJdbcClass(ColumnType type) {
```
AMashenkov, line 171: Can we move this method along with columnTypeToClass() from public api to the ignite-core module?

AMashenkov (approved):

modules/api/src/main/java/org/apache/ignite/sql/ColumnType.java
AMashenkov: As I see, methods 'columnTypeToJdbcClass' and 'columnTypeToClass' are used in 
* clients, client-handler modules to return correct java class for particular client
* sql-engine module to return correct java class in case of embedded node
The common modules for all of them is ignite-client-common module.

ygerzhedovich: @AMashenkov thanks for pointing me. Fixed.

PR #2064, opened 2023-05-11
IGNITE-19081 Introduce Catalog events

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-19081

### DIFF
--- a/modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
+++ b/modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
@@ -48,6 +48,8 @@ import org.apache.ignite.internal.catalog.storage.UpdateEntry;
 import org.apache.ignite.internal.catalog.storage.UpdateLog;
 import org.apache.ignite.internal.catalog.storage.UpdateLog.OnUpdateHandler;
 import org.apache.ignite.internal.catalog.storage.VersionedUpdate;
+import org.apache.ignite.internal.logger.IgniteLogger;
+import org.apache.ignite.internal.logger.Loggers;
 import org.apache.ignite.internal.manager.Producer;
 import org.apache.ignite.internal.util.ArrayUtils;
 import org.apache.ignite.internal.util.PendingComparableValuesTracker;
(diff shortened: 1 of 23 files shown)


modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -270,7 +323,7 @@ public void handle(VersionedUpdate update) {
 
             registerCatalog(catalog);
 
-            versionTracker.update(catalog.version(), null);
+            CompletableFuture.allOf(eventFutures.toArray(CompletableFuture[]::new)).thenRun(() -> versionTracker.update(version, null));
```
korlov42, line 326: `thenRun` is invoked only in case of successful completion, but what if any listener will throw an exception? We will get a frozen tests without notion about what went wrong. Let's at least log the exception

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ddl/DdlCommandHandlerWrapper.java
```diff
@@ -53,8 +57,15 @@ public DdlCommandHandlerWrapper(
     @Override
     public CompletableFuture<Boolean> handle(DdlCommand cmd) {
         if (cmd instanceof CreateTableCommand) {
```
korlov42, line 59: since we always have to call super, does it make sense to move `if` statement inside `thenCompose`?

modules/catalog/src/test/java/org/apache/ignite/internal/catalog/CatalogServiceSelfTest.java
```diff
@@ -247,6 +253,35 @@ public void catalogServiceManagesUpdateLogLifecycle() throws Exception {
         verify(updateLogMock).stop();
     }
 
+    @Test
+    public void testCreateTableEvents() {
```
korlov42, line 257: let's test to validate CatalogEvent.TABLE_DROP

korlov42 (approved):

PR #2066, opened 2023-05-11
IGNITE-19464 Move out logic of applying cluster configuration from ClusterManagementGroupManager

ivgag:


modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/Action.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.cluster.management;
+
+public interface Action<A1, A2> {
```
ibessonov, line 20: I would love to see a Javadoc. Looks like some primitive version of a state machine, I wonder what it's for.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -828,6 +818,10 @@ public CompletableFuture<Void> onJoinReady() {
         }
     }
 
+    public CompletableFuture<UpdateDistributedConfigurationAction> clusterConfigurationToUpdate() {
+        return updateDistributedConfigurationActionFuture;
```
ibessonov, line 822: We should complete this future on the node stop. What if somebody's waiting and we stop the node too early?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/UpdateDistributedConfigurationAction.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package org.apache.ignite.internal.cluster.management;
+
+import java.util.concurrent.CompletableFuture;
+import org.jetbrains.annotations.Nullable;
+
+public class UpdateDistributedConfigurationAction implements Action<String, CompletableFuture<Void>> {
+
+    private final String cfgToUpdate;
```
ibessonov, line 25: I know this will sound picky, but I don't think that it's correct to name this thing "configuration to update", because "object to update" usually means that the object itself will be updated, not that it contains the update for some other object.
Could it simple be "configurationString" or "configuration"? I'd be less confused in that case

ibessonov (commented):
Generally speaking, I agree with the approach, but what I don't understand is why we need an `Action` abstraction, if there's a single implementation and it's only used by configuration updater.
Are there plans for more actions? Is there a bigger design goal that I don't see?
If not, then please remove the interface. "Entities are not to be multiplied without necessity", you know

modules/runner/src/main/java/org/apache/ignite/internal/configuration/DistributedConfigurationUpdater.java
```diff
@@ -31,37 +31,38 @@ public class DistributedConfigurationUpdater implements IgniteComponent {
...
-     */
-    public CompletableFuture<Void> updateConfiguration(String configurationToApply) {
-        return clusterCfgPresentation.thenCompose(presentation -> presentation.update(configurationToApply))
-                .whenComplete((v, e) -> {
+    @Override
+    public void start() {
+        cmgMgr.clusterConfigurationToUpdate()
+                .thenApply(action -> {
+                    if (action.configuration() != null) {
+                        presentation.update(action.configuration());
```
sashapolo, line 48: `presentation.update` returns a future, which we ignore here, that's a bug

modules/cluster-management/src/integrationTest/java/org/apache/ignite/internal/cluster/management/ItClusterManagerTest.java
```diff
@@ -414,6 +457,18 @@ void nonCmgNodeAddedLaterGetsLogicalTopologyChanges(TestInfo testInfo) throws Ex
         assertTrue(waitForCondition(() -> nonCmgTopology.getLogicalTopology().nodes().size() == 2, 10_000));
     }
 
+    private Optional<MockNode> findLeaderNode(List<MockNode> cluster) {
```
sashapolo, line 460: Since you have extracted a method, please use it in other places in this class as well

modules/cluster-management/src/integrationTest/java/org/apache/ignite/internal/cluster/management/ItClusterManagerTest.java
```diff
@@ -440,11 +495,17 @@ private void waitForLogicalTopology() throws InterruptedException {
         }, 10000));
     }
 
+
     private void initCluster(String[] metaStorageNodes, String[] cmgNodes) throws NodeStoppingException {
+        initCluster(metaStorageNodes, cmgNodes, null);
+    }
+
+    private void initCluster(String[] metaStorageNodes, String[] cmgNodes, String clusterConfiguration) throws NodeStoppingException {
```
sashapolo, line 503: `clusterConfiguration` should be annotated as `@Nullable`

modules/cluster-management/src/integrationTest/java/org/apache/ignite/internal/cluster/management/ItClusterManagerTest.java
```diff
@@ -351,6 +355,45 @@ void testLeaderChangeDuringJoin(TestInfo testInfo) throws Exception {
...
+        // Read cluster configuration from the cluster state and remove it.
+        UpdateDistributedConfigurationAction configurationAction = leaderNode.clusterManager()
+                .clusterConfigurationToUpdate()
+                .get();
+
+        assertThat(configurationAction.configuration(), is(clusterConfiguration));
+        configurationAction.nextAction().apply(CompletableFuture.completedFuture(null)).join();
+
+        // Stop the cluster leader.
+        stopNodes(List.of(leaderNode));
```
sashapolo, line 381: What scenario are you testing here? Why do we need to stop the leader?

modules/runner/src/test/java/org/apache/ignite/internal/configuration/DistributedConfigurationUpdaterTest.java
```diff
@@ -0,0 +1,107 @@
...
+        UpdateDistributedConfigurationAction updateDistributedConfigurationAction =
+                new UpdateDistributedConfigurationAction(
+                        configuration,
+                        (result) -> result.whenComplete((v, e) -> nextAction.complete(null)));
+
+        when(cmgMgr.clusterConfigurationToUpdate())
+                .thenReturn(CompletableFuture.completedFuture(updateDistributedConfigurationAction));
+
+        // Run updater.
+        DistributedConfigurationUpdater distributedConfigurationUpdater = new DistributedConfigurationUpdater(
```
sashapolo, line 64: `distributedConfigurationUpdater` can be extracted into a field

modules/cluster-management/src/integrationTest/java/org/apache/ignite/internal/cluster/management/ItClusterManagerTest.java
sashapolo: Please leave a clarifying comment

modules/runner/src/main/java/org/apache/ignite/internal/configuration/DistributedConfigurationUpdater.java
```diff
@@ -31,37 +30,27 @@ public class DistributedConfigurationUpdater implements IgniteComponent {
...
-     *
-     * @param configurationToApply Cluster configuration that should be applied.
-     * @return Future that will be completed when cluster configuration is updated.
-     */
-    public CompletableFuture<Void> updateConfiguration(String configurationToApply) {
-        return clusterCfgPresentation.thenCompose(presentation -> presentation.update(configurationToApply))
+    @Override
+    public void start() {
+        cmgMgr.clusterConfigurationToUpdate()
+                .thenApply(action -> action.execute(presentation::update))
```
sashapolo, line 45: this must be `thenCompose`

modules/cluster-management/src/integrationTest/java/org/apache/ignite/internal/cluster/management/ItClusterManagerTest.java
sashapolo: This comment has not been fixed

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -389,45 +388,44 @@ private void onElectedAsLeader(long term) {
...
-                        return completedFuture(null);
-                    } else {
-                        LOG.info("Cluster configuration is found in the Raft service, going to apply it");
-                        return distributedConfigurationUpdater.updateConfiguration(state.clusterConfigurationToApply())
-                                .thenCompose(unused -> removeClusterConfigFromClusterState(service));
-                    }
-                });
+        raftServiceAfterJoin().whenComplete((service, e) -> {
+            if (e != null) {
+                LOG.error("Error when joining to the raft service", e);
```
sashapolo, line 393: `joining to the raft service` ? What does that mean?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -389,45 +388,44 @@ private void onElectedAsLeader(long term) {
...
+                updateDistributedConfigurationActionFuture.completeExceptionally(e);
+            } else {
+                service.readClusterState()
+                        .thenAccept(state -> {
```
sashapolo, line 397: I'll copy my previous comment here: I think we need to use `whenComplete` here to complete `updateDistributedConfigurationActionFuture` with exceptions in case previous operations fail

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -389,45 +388,44 @@ private void onElectedAsLeader(long term) {
...
+                                updateDistributedConfigurationActionFuture.cancel(true);
+                            }
+                        });
+            }
+        });
     }
 
     private CompletableFuture<Void> removeClusterConfigFromClusterState(CmgRaftService service) {
         return service.readClusterState()
                 .thenCompose(state -> {
```
sashapolo, line 415: I think we should check that configuration is actually present in the cluster state (i.e. it is not null) and only then update it

modules/runner/src/main/java/org/apache/ignite/internal/configuration/DistributedConfigurationUpdater.java
```diff
@@ -31,37 +31,34 @@ public class DistributedConfigurationUpdater implements IgniteComponent {
...
+    public void start() {
+        cmgMgr.clusterConfigurationToUpdate()
+                .thenCompose(action -> {
+                    if (action.configuration() != null) {
+                        return presentation.update(action.configuration()).thenApply(ignored -> action);
+                    } else {
+                        return CompletableFuture.completedFuture(action);
+                    }
+                })
+                .thenCompose(action -> action.nextAction().get())
```
sashapolo, line 53: Why do we need to run the `nextAction` even if no configuration was provided in the previous step?

modules/runner/src/test/java/org/apache/ignite/internal/configuration/DistributedConfigurationUpdaterTest.java
```diff
@@ -0,0 +1,116 @@
...
+        when(presentation.update(anyString())).thenReturn(completedFuture(null));
+
+        CompletableFuture<Void> nextAction = new CompletableFuture<>();
+        String configuration = "security.authentication.enabled:true";
+        UpdateDistributedConfigurationAction updateDistributedConfigurationAction =
+                new UpdateDistributedConfigurationAction(
+                        configuration,
+                        () -> {
+                            nextAction.complete(null);
+                            return null;
```
sashapolo, line 61: Wouldn't this lead to an NPE? Why not return `nextAction`?

modules/runner/src/test/java/org/apache/ignite/internal/configuration/DistributedConfigurationUpdaterTest.java
```diff
@@ -0,0 +1,116 @@
...
+                presentation
+        );
+
+        distributedConfigurationUpdater.start();
+
+        // Verify that configuration wasn't updated.
+        verify(presentation, never()).update(any());
+
+        // Verify that next action is completed.
+        nextAction.join();
```
sashapolo, line 113: please use `willCompleteSuccessfully()`

modules/runner/src/test/java/org/apache/ignite/internal/configuration/DistributedConfigurationUpdaterTest.java
```diff
@@ -0,0 +1,116 @@
...
+                presentation
+        );
+
+        distributedConfigurationUpdater.start();
+
+        // Verify that configuration was updated.
+        verify(presentation, times(1)).update(configuration);
+
+        // Verify that next action is completed.
+        nextAction.join();
```
sashapolo, line 80: please use `willCompleteSuccessfully()`

modules/runner/src/main/java/org/apache/ignite/internal/configuration/DistributedConfigurationUpdater.java
```diff
@@ -31,37 +31,37 @@ public class DistributedConfigurationUpdater implements IgniteComponent {
...
+    public DistributedConfigurationUpdater(ClusterManagementGroupManager cmgMgr, ConfigurationPresentation<String> presentation) {
+        this.cmgMgr = cmgMgr;
+        this.presentation = presentation;
     }
 
     @Override
     public void start() {
-
+        cmgMgr.clusterConfigurationToUpdate()
+                .thenCompose(action -> {
```
sashapolo, line 46: Why do you need to use `thenCompose` here? This future is not returned anywhere. `thenAccept` looks more suitable

modules/runner/src/test/java/org/apache/ignite/internal/configuration/DistributedConfigurationUpdaterTest.java
sashapolo: This comment was not fixed

modules/runner/src/main/java/org/apache/ignite/internal/configuration/DistributedConfigurationUpdater.java
```diff
@@ -31,37 +31,37 @@ public class DistributedConfigurationUpdater implements IgniteComponent {
...
+    public DistributedConfigurationUpdater(ClusterManagementGroupManager cmgMgr, ConfigurationPresentation<String> presentation) {
+        this.cmgMgr = cmgMgr;
+        this.presentation = presentation;
     }
 
     @Override
     public void start() {
-
+        cmgMgr.clusterConfigurationToUpdate()
+                .thenApply(action -> {
```
sashapolo, line 46: This should be `thenAccept`, not `thenApply` (as was written in my original comment), what's the point of creating a future in the `else` branch?

sashapolo (approved):

PR #2070, opened 2023-05-14
IGNITE-19405 Remove assignments from table configuration

kgusakov:
https://issues.apache.org/jira/browse/IGNITE-19405


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1277,6 +1296,7 @@ private void dropTableLocally(long causalityToken, String name, int tblId, List<
 
     private Set<Assignment> calculateAssignments(TableConfiguration tableCfg, int partNum) {
         return AffinityUtils.calculateAssignmentForPartition(
+                // TODO: https://issues.apache.org/jira/browse/IGNITE-19506 we must use distribution zone keys here
```
sergeyuttsel, line 1299: Why did you add this TODO? Actually we will not use data nodes from zone manager here when  IGNITE-19506 will be implemented. We want to use data nodes from zone manager here when https://issues.apache.org/jira/browse/IGNITE-19425 will be implemented.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -544,16 +548,60 @@ private CompletableFuture<?> onTableCreate(ConfigurationNotificationEvent<TableV
         }
 
         try {
-            return createTableLocally(
+            DistributionZoneView zone =
+                    getZoneById(distributionZonesConfiguration, (ctx.newValue()).zoneId()).value();
+
+            List<Set<Assignment>> assignments = AffinityUtils.calculateAssignments(
+                    baselineMgr.nodes().stream().map(ClusterNode::name).collect(toList()),
```
alievmirza, line 555: I think that we shouldn't use here baseline manger, what do you think about `DistributionZoneManager#topologyVersionedDataNodes` while proper solution with causality token is not implemented? 


In any cases, we need to add todo here to use proper method

sergeyuttsel:
Previously it was discussed that usage of DistributionZoneManager#topologyVersionedDataNodes method will be removed in this pr. So why it is not removed?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
alievmirza: I see that you've removed usage of `DistributionZoneManager#topologyVersionedDataNodes` in a further commit, so in general I'm ok with using `baselineMgr.nodes()` and adding to do

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
```diff
@@ -1038,7 +1038,9 @@ public void testOneNodeRestartWithGap() throws InterruptedException {
     /**
      * Checks that the table created in cluster of 2 nodes, is recovered on a node after restart of this node.
      */
-    @Test
+    // TODO: https://issues.apache.org/jira/browse/IGNITE-19506 here we receive the empty nodes from baseline manager,
+    // but should get the nodes from metastore instead
+    @Disabled("https://issues.apache.org/jira/browse/IGNITE-19506")
```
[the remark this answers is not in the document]
sergeyuttsel: Ok. But still this test can be fixed by using startNode instead of startPartialNode.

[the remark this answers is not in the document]
kgusakov: Yep, fixed, thanks a lot for the advice!

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/DistributionZoneRebalanceEngine.java
```diff
@@ -187,39 +187,47 @@ public CompletableFuture<Void> onUpdate(WatchEvent evt) {
...
-                            List<Set<Assignment>> tableAssignments = assignmentsBytes == null
-                                    ? Collections.emptyList()
-                                    : ByteUtils.fromBytes(assignmentsBytes);
+                            CompletableFuture<List<Set<Assignment>>> tableAssignmentsFut = tableAssignments(
+                                    metaStorageManager,
+                                    tableId, distributionZoneConfiguration.partitions().value());
 
-                            for (int part = 0; part < distributionZoneConfiguration.partitions().value(); part++) {
-                                int tableId = tableCfg.id().value();
+                            tableAssignmentsFut.thenAccept(tbls -> {
```
vldpyatkov, line 196: Why do you call assignments tbls?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/DistributionZoneRebalanceEngine.java
```diff
@@ -187,39 +187,47 @@ public CompletableFuture<Void> onUpdate(WatchEvent evt) {
...
+
+                                    updatePendingAssignmentsKeys(
+                                            tableView.name(),
+                                            replicaGrpId,
+                                            filteredDataNodes,
+                                            replicas,
+                                            evt.entryEvent().newEntry().revision(),
+                                            metaStorageManager,
+                                            partId,
+                                            tbls.isEmpty() ? Collections.emptySet() : tbls.get(partId)
```
vldpyatkov, line 214: This is a suspicious condition.
What happens if one assignment is not created in MS?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -574,63 +623,38 @@ private CompletableFuture<?> onTableDelete(ConfigurationNotificationEvent<TableV
...
-        } finally {
-            busyLock.leaveBusy();
-        }
-    }
-
     /**
      * Updates or creates partition raft groups.
      *
-     * @param assignmentsCtx Change assignment event.
+     * @param evt Change assignment event.
```
vldpyatkov, line 644: Please, describe all method parameters.

kgusakov:
> Previously it was discussed that usage of DistributionZoneManager#topologyVersionedDataNodes method will be removed in this pr. So why it is not removed?

I checked, that's not so simple (not just remove the one method). It looks like we need to remove the logic from the PR, whichi introduced the method too and fix some tests. I created the separate issue for it https://issues.apache.org/jira/browse/IGNITE-19600

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/DistributionZoneRebalanceEngine.java
kgusakov: Just a note - this code was here before my PR.
But anyway, at the moment, we have no places where we can drop 1 partition assignment:
- we have an atomic invoke, which will write all assignments or nothing
- rebalance logic can change the stable partition for 1 partition, but not removing it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
kgusakov: Fixed

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/DistributionZoneRebalanceEngine.java
kgusakov: Fixed

vldpyatkov (approved):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/distribution/zones/ItDistributionZonesFilterTest.java
```diff
@@ -126,14 +124,6 @@ void testFilteredDataNodesPropagatedToStable() throws Exception {
 
         TablePartitionId partId = new TablePartitionId(table.tableId(), 0);
 
-        assertValueInStorage(
```
alievmirza, line 129: We somehow how to check that stable has been actually rewritten. Now we only check it at the end

modules/runner/src/integrationTest/java/org/apache/ignite/internal/distribution/zones/ItDistributionZonesFilterTest.java
```diff
@@ -220,15 +199,15 @@ void testFilteredEmptyDataNodesDoNotTriggerRebalance() throws Exception {
         TablePartitionId partId = new TablePartitionId(table.tableId(), 0);
 
         // Table was created after both nodes was up, so there wasn't any rebalance.
-        assertPendingStableAreNull(metaStorageManager, partId);
+        assertPendingsAreNull(metaStorageManager, partId);
```
alievmirza, line 202: This method was needed to check there wasn't any rebalance at all, now we check only pending, is this correct to check like this the absence of a rebalance?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/distribution/zones/ItDistributionZonesFilterTest.java
kgusakov: Added the check with appropriate todo

modules/runner/src/integrationTest/java/org/apache/ignite/internal/distribution/zones/ItDistributionZonesFilterTest.java
```diff
@@ -298,7 +295,16 @@ void testEmptyDataNodesDoNotPropagatedToStableAfterAlteringFilter() throws Excep
 
         waitDataNodeAndListenersAreHandled(metaStorageManager, 2);
 
-        assertPendingStableAreNull(metaStorageManager, partId);
+        // TODO: https://issues.apache.org/jira/browse/IGNITE-19425 here we should have no nodes,
+        // which pass the filter, when dataNodes from DistributionZoneManager will be used
+        assertValueInStorage(
```
alievmirza, line 300: why not `assertPendingWasNeverExsists`?

kgusakov: The name and semantic of the test is testEmptyDataNodesDoNotPropagatedToStableAfterAlteringFilter, so, I thought it will be strange to check the pendings in the end

sergeyuttsel (approved):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/distribution/zones/ItDistributionZonesFilterTest.java
alievmirza: Yes, you're right

alievmirza (approved):
Generally,  I'm ok with the changes, left some minor comments regarding the code formatting

PR #2072, opened 2023-05-15
IGNITE-19342 Some tests from DistributionZoneManagerScaleUpTest were fixed and enabled.

sergeyuttsel:
https://issues.apache.org/jira/browse/IGNITE-19342

### DIFF
--- a/modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerScaleUpTest.java
+++ b/modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerScaleUpTest.java
@@ -41,6 +41,7 @@ import static org.apache.ignite.internal.util.ByteUtils.toBytes;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
@@ -87,7 +88,6 @@ public class DistributionZoneManagerScaleUpTest extends BaseDistributionZoneMana
     private long prerequisiteRevision;
 
     @Test
-    @Disabled("https://issues.apache.org/jira/browse/IGNITE-19255")
     void testDataNodesPropagationAfterScaleUpTriggered() throws Exception {
         topology.putNode(NODE_1);
 
@@ -114,7 +114,13 @@ public class DistributionZoneManagerScaleUpTest extends BaseDistributionZoneMana
 
         assertDataNodesForZone(1, clusterNodes2.stream().map(ClusterNode::name).collect(Collectors.toSet()), keyValueStorage);
 
-        assertZoneScaleUpChangeTriggerKey(1, 1, keyValueStorage);
+        topology.putNode(NODE_3);
+
+        Set<LogicalNode> clusterNodes3 = Set.of(NODE_1, NODE_2, NODE_3);
+
+        assertLogicalTopology(clusterNodes3, keyValueStorage);
+
+        assertDataNodesForZone(1, clusterNodes3.stream().map(ClusterNode::name).collect(Collectors.toSet()), keyValueStorage);
     }
 
     @Disabled("https://issues.apache.org/jira/browse/IGNITE-19255")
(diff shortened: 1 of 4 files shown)


modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerScaleUpTest.java
```diff
@@ -114,7 +114,13 @@ void testDataNodesPropagationAfterScaleUpTriggered() throws Exception {
...
         assertDataNodesForZone(1, clusterNodes2.stream().map(ClusterNode::name).collect(Collectors.toSet()), keyValueStorage);
 
-        assertZoneScaleUpChangeTriggerKey(1, 1, keyValueStorage);
+        topology.putNode(NODE_3);
+
+        Set<LogicalNode> clusterNodes3 = Set.of(NODE_1, NODE_2, NODE_3);
+
+        assertLogicalTopology(clusterNodes3, keyValueStorage);
+
+        assertDataNodesForZone(1, clusterNodes3.stream().map(ClusterNode::name).collect(Collectors.toSet()), keyValueStorage);
```
sanpwc, line 123: It worth to mention that

- It's better to have `private static final int ZONE_1_ID = 1;` instead of 1.
- Almost everywhere were we assert assertDataNodesForZone for zone1 we may also assert data nodes for the default one, I'd rather add following method to `DistributionZonesTestUtil`
```
    public static void assertDataNodesForZones(
            @Nullable Set<String> clusterNodes,
            KeyValueStorage keyValueStorage,
            int... zoneIds
    ) throws InterruptedException {
        for (int zoneId : zoneIds) {
            assertDataNodesForZone(zoneId, clusterNodes, keyValueStorage);
        }
    }
```

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerScaleUpTest.java
```diff
@@ -260,14 +265,20 @@ void testDropZoneDoNotPropagateDataNodesAfterScaleUp() throws Exception {
 
         assertDataNodesForZone(1, clusterNodesNames2, keyValueStorage);
 
-        assertZoneScaleUpChangeTriggerKey(1, 1, keyValueStorage);
+        assertNotNull(keyValueStorage.get(zoneDataNodesKey(1).bytes()).value(),
```
sanpwc, line 268: Seems that we already checked exact equality of data nodes. See line above. So that NotNull check check dataNodes is useless.

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerScaleUpTest.java
```diff
@@ -293,11 +304,18 @@ void testDropZoneDoNotPropagateDataNodesAfterScaleDown() throws Exception {
 
         assertDataNodesForZone(1, clusterNodesNames2, keyValueStorage);
 
-        assertZoneScaleDownChangeTriggerKey(1, 1, keyValueStorage);
+        assertNotNull(keyValueStorage.get(zoneDataNodesKey(1).bytes()).value(),
```
sanpwc, line 307: Same as above.

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerWatchListenerTest.java
```diff
@@ -36,74 +36,18 @@
...
 import org.junit.jupiter.api.Test;
 
 /**
  * Tests distribution zones logical topology changes and reaction to that changes.
  */
 //TODO: IGNITE-18564 Add tests with not default distribution zones, when distributionZones.change.trigger per zone will be created.
 public class DistributionZoneManagerWatchListenerTest extends BaseDistributionZoneManagerTest {
-    @Disabled("https://issues.apache.org/jira/browse/IGNITE-19255")
-    @Test
-    void testDataNodesOfDefaultZoneUpdatedOnWatchListenerEvent() throws Exception {
```
sanpwc, line 55: Why it's valid to remove given test?

sanpwc (approved):

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerScaleUpTest.java
sanpwc: Fixed.

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerScaleUpTest.java
sanpwc: Fixed.

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerScaleUpTest.java
```diff
@@ -672,7 +689,7 @@ void testUpdateZoneScaleUpTriggersDataNodePropagation() throws Exception {
 
         assertDataNodesForZone(1, Set.of(), keyValueStorage);
 
-        assertZoneScaleDownChangeTriggerKey(1, 1, keyValueStorage);
+        assertZoneScaleDownChangeTriggerKey(3L, 1, keyValueStorage);
```
alievmirza: This check is useless. Let's check that this key become non-null after zone creation or remove check completely

alievmirza: please check also the same logic below

sergeyuttsel: Fixed. I decided to remove this check.

PR #2075, opened 2023-05-16
IGNITE-19454 Validate backwards schema compatibility

rpuch:
https://issues.apache.org/jira/browse/IGNITE-19454

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/SchemaCompatValidator.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/SchemaCompatValidator.java
@@ -116,9 +116,14 @@ class SchemaCompatValidator {
     }
 
     /**
-     * Performs backward compatibility validation. That is, for a tuple that was just read in a transaction,
-     * checks that, if the tuple was written with a schema version later than the initial schema version
-     * of the transaction, the initial schema version is backward compatible with the tuple schema version.
+     * Performs backward compatibility validation of a tuple that was just read in the transaction.
+     *
+     * <ul>
+     *     <li>If the tuple was written with a schema version earlier or same as the initial schema version of the transaction,
+     *     the read is valid.</li>
+     *     <li>If the tuple was written with a schema version later than the initial schema version of the transaction,
+     *     the read is valid only if the initial schema version is backward compatible with the tuple schema version.</li>
+     * </ul>
      *
      * @param tupleSchemaVersion Schema version ID of the tuple.
      * @param tableId ID of the table to which the tuple belongs.


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/SchemaCompatValidator.java
```diff
@@ -49,9 +49,9 @@ class SchemaCompatValidator {
      * @param txId ID of the transaction that gets validated.
      * @param enlistedGroupIds IDs of the partitions that are enlisted with the transaction.
      * @param commitTimestamp Commit timestamp (or {@code null} if it's an abort).
-     * @return Future completed with validation result.
+     * @return Future of validation result.
      */
-    CompletableFuture<ForwardValidationResult> validateForwards(
+    CompletableFuture<CompatValidationResult> validateForwards(
```
SammyVimes, line 54: Should it be "validateForward" instead of validateForward**s**?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/SchemaCompatValidator.java
```diff
@@ -98,20 +98,66 @@ private ForwardValidationResult validateSchemaCompatibility(
...
 
-        // TODO: IGNITE-19229 - more sophisticated logic
+        // TODO: IGNITE-19229 - more sophisticated logic.
         return diff.isEmpty();
     }
+
+    /**
+     * Performs backward compatibility validation. That is, for a tuple that was just read in a transaction,
+     * checks that, if the tuple was written with a schema version later than the initial schema version
+     * of the transaction, the initial schema version is backward compatible with the tuple schema version.
```
SammyVimes, line 121:
```suggestion
    /**
     * Performs backward compatibility validation for a tuple that was just read in a transaction.
     * This method checks if the tuple was written using a schema version later than the initial schema version
     * of the transaction, and verifies that the initial schema version is backward compatible with the tuple's schema version.
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/SchemaCompatValidator.java
```diff
@@ -98,20 +98,66 @@ private ForwardValidationResult validateSchemaCompatibility(
...
+                .thenCompose(ignored -> schemas.waitForSchemaAvailability(tableId, tupleSchemaVersion))
+                .thenApply(ignored -> validateBackwardSchemaCompatibility(tupleSchemaVersion, tableId, beginTimestamp));
+    }
+
+    private CompatValidationResult validateBackwardSchemaCompatibility(
+            int tupleSchemaVersion,
+            UUID tableId,
+            HybridTimestamp beginTimestamp
+    ) {
+        List<FullTableSchema> tableSchemas = schemas.tableSchemaVersionsBetween(tableId, beginTimestamp, tupleSchemaVersion);
```
SammyVimes, line 141: Wait, we use both timestamp and version? What does 'between' mean then?

SammyVimes (changes requested):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/SchemaCompatValidator.java
rpuch: We can identify the version that corresponds to the end of the interval by either a ts or the version ID explicitly. In the case of the backwards validation, we already know the ID, and it seems safer to use the real version ID instead of the tuple's commitTs (as the real schema ID might be different from the one that corresponds to the commitTs).

Between means that we first identify version IDs corresponding to the start and the end of the interval, then we return the list of all versions between them (including), or an empty list if start ID turns out to be greater than end ID.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/SchemaCompatValidator.java
rpuch: I rephrased the javadoc, please take a look, is it better now?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/SchemaCompatValidator.java
SammyVimes: Hm, ok

SammyVimes (approved):
LGTM

PR #2078, opened 2023-05-17
IGNITE-19428 PlacementDriver API implemented.

sanpwc:
https://issues.apache.org/jira/browse/IGNITE-19428

Following API was introduced
```
/**
 * Service that provides an ability to await and retrieve primary replicas for replication groups.
 */
public interface PlacementDriver {
    /**
     * Returns a future for the primary replica for the specified replication group whose expiration time (the right border of the
     * corresponding lease interval) is greater than or equal to the timestamp passed as a parameter. Please pay attention that there are
     * no restriction on the lease start time (left border), it can either be less or greater than or equal to proposed timestamp.
     * Given method will await for an appropriate primary replica appearance if there's no already existing one. Such awaiting logic is
     * unbounded, so it's mandatory to use explicit await termination like {@code orTimeout}.
     *
     * @param groupId Replication group id.
     * @param timestamp Timestamp reference value.
     * @return Primary replica future.
     */
    CompletableFuture<LeaseMeta> awaitPrimaryReplica(ReplicationGroupId groupId, HybridTimestamp timestamp);

    /**
     * Same as {@link #awaitPrimaryReplica(ReplicationGroupId, HybridTimestamp)} despite the fact that given method await logic is bounded.
     * It will wait for a primary replica for a reasonable period of time, and complete a future with {@code TimeoutException} if a matching
     * lease isn't found. Generally speaking reasonable here means enough for distribution across cluster nodes.
     *
     * @param replicationGroupId Replication group id.
     * @param timestamp Timestamp reference value.
     * @return Primary replica future.
     */
    CompletableFuture<LeaseMeta> getPrimaryReplica(ReplicationGroupId replicationGroupId, HybridTimestamp timestamp);
}
```


modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/LeaseTracker.java
```diff
@@ -65,12 +84,22 @@ public class LeaseTracker {
      *
      * @param vaultManager Vault manager.
      * @param msManager Meta storage manager.
+     * @param longLeaseInterval Long lease interval in {@code TimeUnit.MILLISECONDS}.  The interval is used between lease granting attempts.
```
denis-chudov, line 87: It's not obvious from this description that this is the interval for newly granted lease that is not accepted yet, and is never used for actual leases.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseMeta.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package org.apache.ignite.internal.placementdriver;
+
+import org.apache.ignite.internal.hlc.HybridTimestamp;
+
+/**
+ * Replica lease meta.
+ */
+public interface LeaseMeta {
```
denis-chudov, line 25: Shouldn't this interface be in `placement-driver-api`?

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/Lease.java
```diff
@@ -23,17 +23,19 @@
...
  * A lease representation in memory.
  * The real lease is stored in Meta storage.
  */
-public class Lease {
+public class Lease implements LeaseMeta {
     /** The object is used when nothing holds the lease. Empty lease is always expired. */
     public static Lease EMPTY_LEASE = new Lease(null, MIN_VALUE, MIN_VALUE);
 
-    /** A node that holds a lease until {@code stopLeas}. */
+    /** A node that holds a lease until {@code stopLeas}LeaseMeta. */
```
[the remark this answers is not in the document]
sanpwc: It's an Ctrl+C/Ctrl+V artifact. LeaseMeta should be removed.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/LeaseTracker.java
sanpwc: Because same value is used in LeaseUpdater for the purposes you've mentioned. Here it's used in order to specify max time to await for lease to be populated locally within lease across nodes distribution process. So, to be honest I believe that comment is fine.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/LeaseTracker.java
```diff
@@ -152,4 +195,46 @@ public CompletableFuture<Void> onUpdate(WatchEvent event) {
...
+    }
+
+    @Override
+    public CompletableFuture<LeaseMeta> getPrimaryReplica(ReplicationGroupId replicationGroupId, HybridTimestamp timestamp) {
+        if (!busyLock.enterBusy()) {
+            return failedFuture(new NodeStoppingException("Component is stopping."));
+        }
+        try {
+            // There's no sense in awaiting previously detected primary replica more than lease interval.
+            return awaitPrimaryReplica(replicationGroupId, timestamp).orTimeout(longLeaseInterval, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
sanpwc: Because it's maximum reasonable time to await for lease appearance based on lease propagation.
First of all, from the one hand leaseInterval should be gt the MAX_CLOCK_SKEW by design thus we've safe with clock drift here, on the other hand if distribution duration will be more than leaseInterval we may see an inappropriate gap.
Another approach with awaiting ms safe time, seems to be to invasive, code will become ugly, so I've finished with await bounding with given lease interval.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/Lease.java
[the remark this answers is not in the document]
denis-chudov: Anyway, this javadoc seems to be broken. At least we can remove the part " until {@code stopLeas}".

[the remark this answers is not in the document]
sanpwc: Well, it wasn't part of my changes.
@vldpyatkov Could you please clarify what does it mean? Do you have any objections for {@code stopLeas} removal?

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/LeaseTracker.java
sanpwc: Same as above, that wasn't part of my changes, I've just used comment from original code. However I don't have any objections to your proposal. Will use the suggestion. Fixed.

vldpyatkov: I do not have any objection. I think the both definition are clear.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/LeaseTracker.java
```diff
@@ -92,6 +121,11 @@ public void startTrack() {
                 Lease lease = fromBytes(entry.value());
 
                 leases.put(grpId, lease);
+
+                assert lease.getExpirationTime() != MAX_VALUE : "INFINITE lease expiration time isn't expected";
```
vldpyatkov, line 125: I do not understand which is the reason to check expiration time to MAX_VALUE. We never use the constant to define a lease.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/LeaseTracker.java
```diff
@@ -138,10 +175,16 @@ public CompletableFuture<Void> onUpdate(WatchEvent event) {
 
                 if (msEntry.empty()) {
                     leases.remove(grpId);
+                    tryRemoveTracker(grpId);
                 } else {
                     Lease lease = fromBytes(msEntry.value());
 
+                    assert lease.getExpirationTime() != MAX_VALUE : "INFINITE lease expiration time isn't expected";
```
vldpyatkov, line 182: The same as above, using of MAX_VALUE is unexpected.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/LeaseTracker.java
```diff
@@ -65,12 +84,22 @@ public class LeaseTracker {
...
+    public LeaseTracker(
+            VaultManager vaultManager,
+            MetaStorageManager msManager,
+            long longLeaseInterval,
+            IgniteSpinBusyLock busyLock
+    ) {
         this.vaultManager = vaultManager;
         this.msManager = msManager;
+        this.longLeaseInterval = longLeaseInterval;
+        this.busyLock = busyLock;
```
vldpyatkov, line 99: It is forbidden to take the lock object from where it was created.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/Lease.java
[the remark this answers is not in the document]
sanpwc: Fixed.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/LeaseTracker.java
sanpwc: Fixed,

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
```diff
@@ -74,7 +73,7 @@ public class LeaseUpdater {
 
     private final AtomicBoolean active = new AtomicBoolean();
 
-    /** Long lease interval. The interval is used between lease granting attempts. */
+    /** Long lease interval in {@code TimeUnit.MILLISECONDS}. The interval is used between lease granting attempts. */
```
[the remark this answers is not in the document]
sanpwc: Shame on me))

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/LeaseTracker.java
sanpwc: Removed.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/LeaseTracker.java
sanpwc: Also removed.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/LeaseTracker.java
sanpwc: Fixed.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/LeaseTracker.java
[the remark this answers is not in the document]
vldpyatkov: I disagree that we can use long lease interval to limited waiting time. Timeout transaction is possible to use here, if I correctly understand how it is used in production, but the limitation is imposed externally.
Here, safe time is only possible to use. If it is not possible for now, you will be able to write TODO.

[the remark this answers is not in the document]
denis-chudov: Consider the following example:
- longLeaseInterval is 2 min
- cluster is experiencing some problems
- user runs very important transaction with a timeout 10 min, they are aware that there are some problems but have nothing to do with it right now, that's why the transaction timeout is rather long
- transaction starts, successfully performs some operation within, let say, 1 minute, and is going to be committed with commit timestamp 1000
- commit operation hangs for 2 minutes and then the transaction is rolled back at timestamp 1120 (let max clock skew be 0), because it was unable to get lease for timestamp 1000 withing this time interval. Actually lease for the interval (500, 1002) was received on corresponding node at local time 1121, but the transaction is already rolled back at this time.

So we have the transaction with timeout 10 minutes that was committed by user but rolled back by cluster after ~3 minutes. It is actually implicit commit timeout, which serves also as implicit and non-configurable transaction timeout. This looks at least weird from the user's point of view.

I understand that longLeaseInterval is long enough to count it as a reasonable time of data propagation, but the consequence is that we can roll back any transaction if some of its operations exceeds "reasonable" time which is not related to transaction timeout, and therefore, arbitrarily roll back any transaction before it reaches its timeout. This behavior impacts user experience, seems to be invalid and shouldn't exist.

So, the transaction timeout instead of safe time seems to be more correct here.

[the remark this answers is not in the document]
vldpyatkov: To use safe time is absolutely correct, but any timeout cannot be used as a replacement, because any timeout might be violated. Time to spread information across cluster does not have time limitation.
We have only guaranties are related to clock of cluster, but this cannot be applied to safe time. Safe time on different nodes might be arbitrary different. Hence, we cannot to replace safe time check by checking timeout.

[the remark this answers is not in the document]
sanpwc: > Timeout transaction is possible to use here
Nope

> To use safe time is absolutely correct
Yes. I've reworked getPrimaryReplica to use the safe time. And created a ticket in order to fix the race in meta storage safe time publication https://issues.apache.org/jira/browse/IGNITE-19532 @SammyVimes is notified and hopefully will fix it within next week. It worth to mention that I've run corresponding test 1000 times - no failures, so I event didn't mute it with the issue.

denis-chudov (approved):

vldpyatkov (approved):

PR #2079, opened 2023-05-17
IGNITE-19278 Multiple partition scan requests return an unexpected number of rows.

xtern:
Due to an implementation bug the following 2 requests to partition scan may produce more rows then expected (6 rows instead of 4).
```
publisher = internalTable.scan(...)
Subscription subscription = ...subscribe to publisher...

subscription.request(3); // call scanBatch(3)
subscription.request(1); // call scanBatch(3) instead of scanBatch(1)
```

https://issues.apache.org/jira/browse/IGNITE-19278


modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
```diff
@@ -336,6 +337,36 @@ public void testGetAndDeleteDuringPureTableScan() throws Exception {
...
+
+            subscription.request(input[0]);
+            subscription.request(input[1]);
+
+            int total = input[0] + input[1];
+            waitForCondition(() -> scannedRows.size() == total, 10_000);
+            assertEquals(total, scannedRows.size(), "input=" + Arrays.toString(input));
+
+            // Close the publisher.
+            subscription.request(1_000);
```
zstan, line 365: why do we need such request ? can we just cancel it here ?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
```diff
@@ -336,6 +337,36 @@ public void testGetAndDeleteDuringPureTableScan() throws Exception {
...
+            CompletableFuture<Void> scanned = new CompletableFuture<>();
+
+            Subscription subscription = subscribeToPublisher(scannedRows, publisher, scanned);
+
+            subscription.request(input[0]);
+            subscription.request(input[1]);
+
+            int total = input[0] + input[1];
+            waitForCondition(() -> scannedRows.size() == total, 10_000);
+            assertEquals(total, scannedRows.size(), "input=" + Arrays.toString(input));
```
zstan, line 362: we already check rows in previous line, isn`t it ?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
xtern: Done, thanks.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
xtern: In the previous line, we just wait for a condition. The condition may fail.
But `assertTrue(waitForCondition(...))` will not display the actual number of rows received. Thus, I prefer to do the final check with an explicit `assertEquals(...)`. If you insist on adding `assertTrue` I can redo it.

xtern: Reworked to single line condition.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -1408,8 +1408,12 @@ private void scanBatch(int n) {
 
                     if (binaryRows.size() < n) {
                         cancel();
-                    } else if (requestedItemsCnt.addAndGet(Math.negateExact(binaryRows.size())) > 0) {
-                        scanBatch(Math.min(n, INTERNAL_BATCH_SIZE));
+                    } else {
+                        long remaining = requestedItemsCnt.addAndGet(Math.negateExact(binaryRows.size()));
+
+                        if (remaining > 0) {
+                            scanBatch((int) Math.min(remaining, INTERNAL_BATCH_SIZE));
```
sanpwc, line 1415: I believe that it's not correct to request remaining if user specified n < INTERNAL_BATCH_SIZE, so it's actually should be `Math.min(Math.min(remaining, n), INTERNAL_BATCH_SIZE);` What do you think?

xtern: No, I don't think so, RequestItemsCnt represents the number of requested rows left.

For example, user expects 513 rows total.:
```
request(1)  // requestedItemsCnt=1, then call scanBatch(1)
request(512) // just set requestedItemsCnt=513, don't call scanBatch(512)
```
When the first scanBatch (n = 1) completes we have
```
n = 1 (still)
remaining = requestedItemsCnt - processed = 512
```
then (with your suggestion) we will repeat it with n = 1 
```
scanBatch(Math.min(n, remaining)) = scanBatch(Math.min(1, 512)) = scanBatch(1)
```

In other worlds we'll do
```
while (remaining-- > 0)
 scanBatch(n=1)
```
instead of single `scanBatch(512)`

sanpwc: Yep, you are right.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
```diff
@@ -580,6 +581,34 @@ public void testPhantomReads() throws Exception {
...
 
+    /**
+     * Ensures that multiple consecutive scan requests with different requested rows amount
+     * return the expected total number of requested rows.
+     *
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testCompositeScanRequest() throws Exception {
+        int[][] demandPairs = {{3, 1}, {1, 3}};
```
sanpwc, line 592: Parametrized test? Up to you though.

sanpwc (approved):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
xtern: Done, thanks

PR #2082, opened 2023-05-17
IGNITE-19509 Get rid of reading configuration when calling methods MvTableStorage#getOrCreate*Index

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19509

### DIFF
--- a/modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
+++ b/modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
@@ -39,6 +39,7 @@ import org.apache.ignite.internal.raft.service.RaftGroupService;
 import org.apache.ignite.internal.schema.configuration.index.TableIndexView;
 import org.apache.ignite.internal.storage.MvPartitionStorage;
 import org.apache.ignite.internal.storage.RowId;
+import org.apache.ignite.internal.storage.index.IndexDescriptor;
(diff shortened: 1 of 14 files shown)


modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuilder.java
```diff
@@ -98,10 +99,10 @@ void stop() {
     /**
      * Initializes the build of the index.
      */
-    void startIndexBuild(TableIndexView tableIndexView, TableImpl table) {
+    void startIndexBuild(TableIndexView tableIndexView, TableImpl table, IndexDescriptor indexDescriptor) {
```
rpuch, line 102: It looks like `tableIndexView` is only used to obtain index ID, which we can now take from `indexDescriptor`. I suggest to remove `tableIndexView` from parameters and pass index ID directly to the components that need it (like `BuildIndexTask`). This will allow us to reduce the usage of configuration even further.

tkalkirill: I think that in the near future this code will disappear, the building of indexes has changed a little in the https://github.com/apache/ignite-3/pull/2030

rpuch: Ok

rpuch (approved):

PR #2083, opened 2023-05-17
IGNITE-19381 TimestampAware messages sometimes lacks timestamps

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-19381

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
@@ -41,7 +41,7 @@ import org.jetbrains.annotations.Nullable;
  */
 public class SortedIndexLocker implements IndexLocker {
     /** Index INF+ value object. */
-    private final Integer positiveInf;
+    private final Object positiveInf;
 
     private final UUID indexId;
     private final LockManager lockManager;
@@ -67,8 +67,7 @@ public class SortedIndexLocker implements IndexLocker {
         this.lockManager = lockManager;
         this.storage = storage;
         this.indexRowResolver = indexRowResolver;
-
-        this.positiveInf = new Integer(partId);
+        this.positiveInf = Integer.valueOf(partId);
     }
 
     /** {@inheritDoc} */
@@ -152,16 +151,16 @@ public class SortedIndexLocker implements IndexLocker {
         BinaryTuplePrefix prefix = BinaryTuplePrefix.fromBinaryTuple(key);
 
         // Find next key.
-        Cursor<IndexRow> cursor = storage.scan(prefix, null, SortedIndexStorage.GREATER);
+        try (Cursor<IndexRow> cursor = storage.scan(prefix, null, SortedIndexStorage.GREATER)) {
+            IndexRow nextRow = cursor.hasNext() ? cursor.next() : null;
 
-        IndexRow nextRow = cursor.hasNext() ? cursor.next() : null;
+            var nextLockKey = new LockKey(indexId, indexKey(nextRow));
 
-        var nextLockKey = new LockKey(indexId, indexKey(nextRow));
-
-        return lockManager.acquire(txId, nextLockKey, LockMode.IX).thenCompose(shortLock ->
-                lockManager.acquire(txId, new LockKey(indexId, key.byteBuffer()), LockMode.X).thenApply((lock) ->
-                        new Lock(nextLockKey, LockMode.IX, txId)
-                ));
+            return lockManager.acquire(txId, nextLockKey, LockMode.IX).thenCompose(shortLock ->
+                    lockManager.acquire(txId, new LockKey(indexId, key.byteBuffer()), LockMode.X).thenApply((lock) ->
+                            new Lock(nextLockKey, LockMode.IX, txId)
+                    ));
+        }
     }
 
     /** {@inheritDoc} */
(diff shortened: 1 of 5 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
```diff
@@ -45,18 +40,8 @@
  * <p>Simply acquires lock on a given row for lookup and remove, acquires lock on a next key for insert.
  */
 public class SortedIndexLocker implements IndexLocker {
```
sanpwc, line 42: What about HashIndexLocker?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
```diff
@@ -45,18 +40,8 @@
...
-            new Column[]{
-                    new Column("indexId", NativeTypes.UUID, false),
-                    new Column("partId", NativeTypes.INT32, false)
-            },
-            new Column[0]
-    );
-
     /** Index INF+ value object. */
-    private final BinaryTuple positiveInf;
+    private final Integer positiveInf;
```
sanpwc, line 44: Well, to be honest, despite the fact that currently it should work because index keys are Bytes, I'll still favor custom InfWrapper here.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
```diff
@@ -83,26 +68,7 @@ public SortedIndexLocker(UUID indexId, int partId, LockManager lockManager, Sort
...
-        var binarySchema = BinaryTupleSchema.createSchema(INFINITY_TUPLE_SCHEMA, new int[]{
-                INFINITY_TUPLE_SCHEMA.column("indexId").schemaIndex(),
-                INFINITY_TUPLE_SCHEMA.column("partId").schemaIndex()
-        });
-
-        return new BinaryTuple(
-                binarySchema,
-                new BinaryTupleBuilder(binarySchema.elementCount(), false).appendUuid(indexId).appendInt(partId).build()
-        );
+        this.positiveInf = new Integer(partId);
```
sanpwc, line 71: @Deprecated(since="9", forRemoval = true)
    public Integer(int value) {

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
```diff
@@ -188,14 +154,9 @@ public CompletableFuture<Lock> locksForInsert(UUID txId, BinaryRow tableRow, Row
         // Find next key.
         Cursor<IndexRow> cursor = storage.scan(prefix, null, SortedIndexStorage.GREATER);
```
sanpwc, line 155: Not related to your changes, but we should close the cursor. try-with-resources?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
korlov42: +1 for custom wrapper

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
korlov42: good catch!

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
vldpyatkov: Is it enough to use object here, or you want to have a specific class to define infinite infinity?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
vldpyatkov: Replace to valueOf()

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/SortedIndexLocker.java
```diff
@@ -186,21 +151,16 @@ public CompletableFuture<Lock> locksForInsert(UUID txId, BinaryRow tableRow, Row
         BinaryTuplePrefix prefix = BinaryTuplePrefix.fromBinaryTuple(key);
 
         // Find next key.
-        Cursor<IndexRow> cursor = storage.scan(prefix, null, SortedIndexStorage.GREATER);
+        try (Cursor<IndexRow> cursor = storage.scan(prefix, null, SortedIndexStorage.GREATER)) {
+            IndexRow nextRow = cursor.hasNext() ? cursor.next() : null;
```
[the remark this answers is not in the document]
vldpyatkov: Why? The rest part is async and does not influence the time when the cursor is being open.
In the other hand, we do not know what we will deal with without cursor there.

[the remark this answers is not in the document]
sanpwc: Thanks!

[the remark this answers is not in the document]
vldpyatkov: Fixed.

sanpwc (approved):

korlov42 (approved):
LGTM!

PR #2088, opened 2023-05-18
IGNITE-19525 Disable flaky test

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-19525


sk0x50 (approved):

modules/rest/src/integrationTest/java/org/apache/ignite/internal/rest/deployment/DeploymentManagementControllerTest.java
```diff
@@ -94,6 +95,7 @@ public void cleanup(TestInfo testInfo) throws Exception {
     }
 
     @Test
+    @Disabled("IGNITE-19526")
```
sk0x50, line 98: IMHO, it would be better to provide a full link to the ticket "https://issues.apache.org/jira/browse/IGNITE-19526", but it's up to you.

PR #2089, opened 2023-05-18
IGNITE-19443 Empty data nodes are not propagated to pending assignments after filtering

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-19443


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ddl/DdlCommandHandler.java
```diff
@@ -230,6 +230,10 @@ private CompletableFuture<Boolean> handleAlterZone(AlterZoneSetCommand cmd) {
             zoneCfgBuilder.dataNodesAutoAdjustScaleUp(cmd.dataNodesAutoAdjustScaleUp());
         }
 
+        if (cmd.replicas() != null) {
```
sanpwc, line 233: What's that?

alievmirza: I've added this for supporting altering replicas for a zone through SQL.

sanpwc (approved):

PR #2091, opened 2023-05-19
IGNITE-19396 Move work with indexes from StorageUpdateHandler to a separate class

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19396


modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/IndexStorage.java
```diff
@@ -67,7 +67,7 @@ public interface IndexStorage {
     @Nullable RowId getNextRowIdToBuild() throws StorageException;
 
     /**
-     * Sets the row ID for which the index needs to be built, {@code null} means that the index is built.
+     * Sets the row ID for which the index needs to be build, {@code null} means that the index is build.
```
rpuch, line 70: Let's revert this change, because 'to be built' seems to be the correct form ('built' is the 3rd form for 'to build')

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/StorageUpdateHandler.java
```diff
@@ -195,11 +193,11 @@ public void handleUpdateAll(
     void executeBatchGc() {
         HybridTimestamp lwm = lowWatermark.getLowWatermark();
 
-        if (lwm == null || safeTimeTracker.current().compareTo(lwm) < 0) {
+        if (lwm == null || gcUpdateHandler.getSafeTimeTracker().current().compareTo(lwm) < 0) {
```
rpuch, line 196: How about extracting the check to a method on GcUpdateHandler? It could be called `safeTimeBelow(HybridTimestamp lwm)` or similarly

tkalkirill: If you insist, I can add, but at the moment it seems that for a single method it is a bit redundant.

rpuch: It's up to you

rpuch (approved):

PR #2096, opened 2023-05-22
IGNITE-19472 Make fsync flag configurable in RAFT

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19472

### DIFF
--- a/modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
+++ b/modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
@@ -149,6 +149,8 @@ public class Loza implements RaftManager {
     public void start() {
         opts.setRpcInstallSnapshotTimeout(raftConfiguration.rpcInstallSnapshotTimeout().value());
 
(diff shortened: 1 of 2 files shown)


modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
```diff
@@ -149,6 +149,8 @@ public Loza(
     public void start() {
         opts.setRpcInstallSnapshotTimeout(raftConfiguration.rpcInstallSnapshotTimeout().value());
 
+        opts.getRaftOptions().setSync(raftConfiguration.fsync().value());
```
rpuch, line 152: Should we first take `RaftView` and then use it to obtain the 2 settings (to make sure that we always get a consistent snapshot of the configuration)?

tkalkirill: Fix it

rpuch (approved):

PR #2102, opened 2023-05-24
IGNITE-19476 On demand deploy units API

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-19476

### DIFF
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/deployment/ItDeploymentUnitTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/deployment/ItDeploymentUnitTest.java
@@ -46,10 +46,8 @@ import org.apache.ignite.internal.deployunit.IgniteDeployment;
 import org.apache.ignite.internal.deployunit.UnitStatuses;
 import org.apache.ignite.internal.deployunit.UnitStatuses.UnitStatusesBuilder;
 import org.apache.ignite.internal.deployunit.configuration.DeploymentConfiguration;
-import org.apache.ignite.internal.deployunit.version.Version;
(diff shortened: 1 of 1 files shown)


modules/code-deployment/src/test/java/org/apache/ignite/deployment/metastore/DeploymentUnitStoreImplTest.java
```diff
@@ -48,39 +54,46 @@
  */
 @ExtendWith(WorkDirectoryExtension.class)
 public class DeploymentUnitStoreImplTest {
+    private static final String LOCAL_NODE = "localNode";
 
     private final VaultManager vaultManager = new VaultManager(new InMemoryVaultService());
 
+    private final List<UnitNodeStatus> history = new ArrayList<>();
```
PakhomovAlexander, line 61: The usage of ArrayList requires tests to run from a single thread otherwise it is not thread-safe. Could you mention this in the comments or use thread-safe collection here?

modules/code-deployment/src/test/java/org/apache/ignite/deployment/metastore/DeploymentUnitStoreImplTest.java
```diff
@@ -58,7 +58,7 @@ public class DeploymentUnitStoreImplTest {
 
     private final VaultManager vaultManager = new VaultManager(new InMemoryVaultService());
 
-    private final List<UnitNodeStatus> history = new ArrayList<>();
+    private final Set<UnitNodeStatus> history = ConcurrentHashMap.newKeySet();
```
PakhomovAlexander, line 61: Set is not the best solution here. It might hide the duplicates. I think duplicates are an important part to check here.

PakhomovAlexander (approved):

PR #2103, opened 2023-05-24
IGNITE-18443 Sql. Provide extend commands and handlers for distributed zones operation

lowka:


korlov42:
@lowka , the patch looks good. It would be nice to add a few tests on parsing though

lowka:
> @lowka , the patch looks good. It would be nice to add a few tests on parsing though

Fixed disabled tests.

PR #2105, opened 2023-05-24
IGNITE-19407 Sql. Introduce DDL command for ALTER COLUMN TYPE.

xtern:
https://issues.apache.org/jira/browse/IGNITE-19407

Command syntax:
```
ALTER TABLE <table_name> ALTER COLUMN <col_name> <action>;
<action> ::= SET DATA TYPE <new_type> [NOT NULL | NULLABLE] [DEFAULT <default value>]| SET NOT NULL | DROP NOT NULL | SET DEFAULT <default value> | DROP DEFAULT
```

Example  of changing the data type and dropping NOT NULL at the same time:
```
SET DATA TYPE BIGINT NULLABLE
```

Example  of changing the data type and dropping default at the same time:
```
SET DATA TYPE BIGINT DEFAULT NULL
```

List of possible actions to change the column.

- NOT NULL -> NULLABLE allowed for any non-PK column.
- NULLABLE -> NOT NULL forbidden.

- DEFAULT -> ANY LITERAL DEFAULT allowed for any column.
- DROP DEFAULT / SET DEFAULT NULL allowed for any column.

- TYPE SCALE change forbidden
- a. TYPE PRECISION increase allowed for DECIMAL non PK column
- b. TYPE LENGTH increase allowed for STRING and BYTE_ARRAY non PK column.
- The following type transitions allowed for TYPE of non PK column:
   FLOAT -> DOUBLE
   INT8 -> INT16 -> INT32 -< INT64
   Any other type transitions are forbidden.


korlov42:
I don't agree the original syntax has any ambiguity. SQL is a declarative language, so you only need to do what is declared in the statement. The statement `ALTER COLUMN VAL SET DATA TYPE VARCHAR(20)` declares that you only what to change type of the column, while leaving rest of the attributes as is.

xtern:
Maybe you don't quite understand what exactly has changed.

The standard SQL syntax is as follows:
```
ALTER TABLE <table_name> ALTER COLUMN <col_name> <action>
<action> ::= SET DATA TYPE <new_type> | SET NOT NULL | SET DEFAULT <default value>;
```
For example, it doesn't allow you to change the type and add NOT NULL constraint at the same time: `ALTER COLUMN SET DATA TYPE INT NOT NULL` - will not work.
You have to drop NOT NULL contraint separately and that makes sense.

Our previous syntax allows you to change the data type and add a NOT NULL constraint, but it does not allow you to change the type and remove NOT NULL, for example.

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -214,6 +222,57 @@ public CompletableFuture<Void> dropColumn(AlterTableDropColumnParams params) {
...
+            String schemaName = Objects.requireNonNullElse(params.schemaName(), CatalogService.PUBLIC);
+
+            SchemaDescriptor schema = Objects.requireNonNull(catalog.schema(schemaName), "No schema found: " + schemaName);
+
+            TableDescriptor table = schema.table(params.tableName());
+
+            if (table == null) {
+                if (params.ifTableExists()) {
+                    return Collections.emptyList();
+                }
```
AMashenkov, line 238: "CREATE TABLE IF EXISTS" always throws an exception in such case.
@korlov42 , do you agree to pass 'if exists' flags inside the parameters?

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/commands/altercolumn/AlterColumnNotNull.java
```diff
@@ -0,0 +1,57 @@
...
+    private final boolean notNull;
+
+    public AlterColumnNotNull(boolean notNull) {
+        this.notNull = notNull;
+    }
+
+    @Override
+    public @Nullable TableColumnDescriptor apply(TableColumnDescriptor origin, boolean isPkColumn) {
+        if (notNull == !origin.nullable()) {
+            return null;
```
AMashenkov, line 40: Null value semantic is unclear for me. E.g. I'd expect the origin descriptor should be dropped when null was returned.

Why don't just return origin if there is nothing to do? and drop @Nullable annotation at all.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/sql/SqlAlterColumnDdlParserTest.java
```diff
@@ -44,69 +46,92 @@ public class SqlAlterColumnDdlParserTest extends AbstractDdlParserTest {
...
+        assertThrows(SqlException.class, () -> parse(QUERY_PREFIX + "SET DEFAULT FUNC"));
     }
 
     /**
      * Verifies parsing of {@code ALTER TABLE ... ALTER COLUMN ... SET DATA TYPE} statement.
      *
-     * <p>The parser is expected to produce a node of {@link IgniteSqlAlterColumnType} class with the specified {@link
-     * IgniteSqlAlterColumnType#name() table name}, {@link IgniteSqlAlterColumnType#columnName() column name}, column {@link
-     * IgniteSqlAlterColumnType#dataType() data type} and an optional {@link IgniteSqlAlterColumnType#expression() default expression}.
+     * <p>Parser must support the following syntax {@code SET DATA TYPE &lt;new_type&gt; [NOT NULL | NULL] [DEFAULT &lt;default value&gt;]}.
```
AMashenkov, line 87:
```suggestion
     * <p>Parser must support the following syntax {@code SET DATA TYPE &lt;new_type&gt; [NOT NULL | NULLABLE] [DEFAULT &lt;default value&gt;]}.
```

AMashenkov: `SET DATA TYPE INTEGER NULL`
NULL here looks ambiguous.

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/commands/altercolumn/AlterColumnNotNull.java
xtern: Fixed, thanks.

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
xtern: This check has been removed.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/sql/SqlAlterColumnDdlParserTest.java
xtern: Grammar chaned: `NULL` changed to `NULLABLE`.... but may be it worth to keep NULL valid as well :thinking:

xtern: Reverted after private discussion :roll_eyes:

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -321,6 +341,89 @@ public CompletableFuture<Void> dropColumn(AlterTableDropColumnParams params) {
...
+                throw new TableNotFoundException(schemaName, params.tableName());
+            }
+
+            String columnName = params.columnName();
+
+            Optional<TableColumnDescriptor> source = table.columns().stream().filter(desc -> desc.name().equals(columnName)).findFirst();
+
+            if (source.isEmpty()) {
+                throw new ColumnNotFoundException(columnName);
+            }
```
AMashenkov, line 364:
```suggestion
            TableColumnDescriptor source = table.columns().stream()
                .filter(desc -> desc.name().equals(columnName))
                .findFirst()
                .orElseThrow(() ->  new ColumnNotFoundException(columnName));
```

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -321,6 +341,89 @@ public CompletableFuture<Void> dropColumn(AlterTableDropColumnParams params) {
...
+
+            boolean nullable = !Objects.requireNonNullElse(params.notNull(), !origin.nullable());
+            DefaultValue defaultValue = Objects.requireNonNullElse(params.defaultValue(origin.type()), origin.defaultValue());
+            ColumnType type = Objects.requireNonNullElse(params.type(), origin.type());
+            int scale = Objects.requireNonNullElse(params.scale(), origin.scale());
+
+            boolean varLenType = type == ColumnType.STRING || type == ColumnType.BYTE_ARRAY;
+            int precision = varLenType ? origin.precision() : Objects.requireNonNullElse(params.precision(), origin.precision());
+            int length = varLenType ? Objects.requireNonNullElse(params.precision(), origin.length()) : origin.length();
```
AMashenkov, line 385: Why do we need a different logic for varlen types?
This logic is incorrect when you change fixlen to varlen, but this transition is forbidden now.
Why do you ignore precision and length from params? Maybe there should be an assert, that we don't expect any changes here?

Also, `Objects.requireNonNullElse(params.precision(), origin.length())` - I guess, here should be `params.length()`

modules/catalog/src/test/java/org/apache/ignite/internal/catalog/CatalogServiceSelfTest.java
```diff
@@ -491,6 +504,319 @@ public void testAddDropMultipleColumns() {
...
+        assertThat(changeColumn(TABLE_NAME, "VAL", null, null, () -> DefaultValue.constant(1)),
+                willBe((Object) null));
+        assertNull(service.schema(schemaVer + 1));
+
+        // 1 -> 2 : Ok.
+        assertThat(changeColumn(TABLE_NAME, "VAL", null, null, () -> DefaultValue.constant(2)),
+                willBe((Object) null));
+        assertNotNull(service.schema(++schemaVer));
+
+        // 2 -> NULL : Ok (for nullable column).
```
AMashenkov, line 540:
```suggestion
        // 2 -> NULL : Ok.
```
I think changing column default to `null` is allowed for any columns.
For non-nullable columns that means user must set value explicitly.

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
xtern: > Why do you ignore precision and length from params?

Currently I don't have length in params. After parsing we have RelDataType, which doesn't have a length.
But since there is a length in the TableColumnDescriptor - for VARCHAR/VARBINARY precision is used as the length, for the rest it is in precision.

If you think this is the wrong place for this transformation, I can add length to the params and set it when converting, WDYT?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ddl/DdlSqlToCommandConverter.java
```diff
@@ -415,6 +423,41 @@ private DefaultValueDefinition convertDefault(SqlNode expression, RelDataType re
...
+        }
+
+        if (alterColumnNode.expression() != null) {
+            SqlNode expr = alterColumnNode.expression();
+
+            Function<ColumnType, DefaultValue> resolveDfltFunc;
+
+            if (expr instanceof SqlLiteral) {
+                resolveDfltFunc = type -> DefaultValue.constant(fromLiteral(type, (SqlLiteral) expr));
+            } else if (expr == null) {
```
AMashenkov, line 449: `expr` is always != null here

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
xtern: Added `length` to params.

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -89,6 +99,14 @@
  */
 public class CatalogServiceImpl extends Producer<CatalogEvent, CatalogEventParameters> implements CatalogManager {
     private static final int MAX_RETRY_COUNT = 10;
+    private static final Map<ColumnType, Set<ColumnType>> ALTER_COLUMN_TYPE_TRANSITIONS = new EnumMap<>(ColumnType.class);
```
AMashenkov, line 102: Let's move this to CatalogUtils

AMashenkov (approved):

PR #2107, opened 2023-05-25
IGNITE-19534 Fix error code duplication in the error message

sk0x50:
https://issues.apache.org/jira/browse/IGNITE-19534

### DIFF
--- a/modules/api/src/main/java/org/apache/ignite/lang/IgniteException.java
+++ b/modules/api/src/main/java/org/apache/ignite/lang/IgniteException.java
@@ -20,7 +20,6 @@ package org.apache.ignite.lang;
 import static org.apache.ignite.lang.ErrorGroup.ERR_PREFIX;
 import static org.apache.ignite.lang.ErrorGroup.errorGroupByCode;
 import static org.apache.ignite.lang.ErrorGroup.errorMessage;
-import static org.apache.ignite.lang.ErrorGroup.errorMessageFromCause;
 import static org.apache.ignite.lang.ErrorGroup.extractErrorCode;
 import static org.apache.ignite.lang.ErrorGroup.extractGroupCode;
 import static org.apache.ignite.lang.ErrorGroups.Common.INTERNAL_ERR;
@@ -110,8 +109,6 @@ public class IgniteException extends RuntimeException {
      * @param code Full error code.
      */
     public IgniteException(UUID traceId, int code) {
-        super(errorMessage(traceId, code, null));
-
         this.traceId = traceId;
         this.groupName = errorGroupByCode((extractGroupCode(code))).name();
         this.code = code;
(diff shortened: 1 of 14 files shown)


modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionTest.java
```diff
@@ -32,34 +32,89 @@
...
+    }
+
+    @Test
+    public void testWrapInternalException() {
+        var originalMessage = "Unexpected error.";
+        var originalTraceId = UUID.randomUUID();
+
+        var originalEx = new IgniteInternalException(originalTraceId, Common.INTERNAL_ERR, originalMessage);
+        var wrappedEx = new CompletionException(originalEx);
+        var res = IgniteException.wrap(wrappedEx);
```
sanpwc, line 78: Not sure, but formally speaking I believe it's not valid to use var in such cases. However, from my point of view it's still fine in given case.

ptupitsyn (approved):
Looks good to me.

I've pushed a small commit with assertion improvements in .NET `SqlTests` - please have a look.

modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionTest.java
sk0x50: Ok. this makes sense to me. I will change

modules/client-handler/src/integrationTest/java/org/apache/ignite/client/handler/ItClientHandlerMetricsTest.java
```diff
@@ -147,7 +147,7 @@ void testBytesSentReceived(TestInfo testInfo) throws Exception {
         ItClientHandlerTestUtils.connectAndHandshake(serverModule, false, true);
 
         assertTrue(
-                IgniteTestUtils.waitForCondition(() -> testServer.metrics().bytesSent() == 216, 1000),
+                IgniteTestUtils.waitForCondition(() -> testServer.metrics().bytesSent() == 158, 1000),
```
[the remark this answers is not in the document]
sk0x50: This particular line of code checks the size of an exception transferred from the server node. After the changes proposed in this PR, the size (in bytes) was reduced just because the exception message is not extended by the error code and trace identifier.

modules/api/src/main/java/org/apache/ignite/lang/IgniteException.java
```diff
@@ -255,6 +252,14 @@ public UUID traceId() {
         return traceId;
     }
 
+    /** {@inheritDoc} */
+    @Override
+    public String toString() {
```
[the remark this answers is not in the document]
sk0x50: we are on the same page. :)

modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionTest.java
sk0x50: As was discussed in private, we decided to stick with the current code without any changes for this particular case

PR #2112, opened 2023-05-29
IGNITE-19547 Switch index IDs from UUID to int

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19547


SammyVimes (approved):
LGTM

PR #2115, opened 2023-05-29
IGNITE-19478 Possible excess size of table entities written into meta storage

denis-chudov:
https://issues.apache.org/jira/browse/IGNITE-19478


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2004,7 +2005,7 @@ private CompletableFuture<Void> handleChangePendingAssignmentEvent(
 
         Set<Assignment> stableAssignments = stableAssignmentsBytes == null
                 // This is for the case when the first rebalance occurs.
-                ? ((List<Set<Assignment>>) ByteUtils.fromBytes(tblCfg.assignments().value())).get(partId)
+                ? Assignments.fromBytes(tblCfg.assignments().value()).get(partId)
                 : ByteUtils.fromBytes(stableAssignmentsBytes);
```
ibessonov, line 2009: This one still uses ByteUtils, is this necessary?

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -0,0 +1,228 @@
...
+        int idx = 0;
+        for (Set<Assignment> assignmentSet : assignments) {
+            for (Assignment a : assignmentSet) {
+                if (consistentIds.putIfAbsent(a.consistentId(), idx) == null) {
+                    idx++;
+                }
+            }
+        }
+
+        byte[] consistentIdsBytes = collectionToBytes(consistentIds.keySet(), String::getBytes);
```
ibessonov, line 94: You should always pass encoding into `getBytes`

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -0,0 +1,228 @@
...
+        byte[] assignmentsBytes = collectionToBytes(
+                assignments,
+                set -> collectionToBytes(set, a -> assignmentToBytes(a, consistentIds::get))
+        );
+
+        ByteBuffer buf = ByteBuffer.allocate(Integer.BYTES * 2 + consistentIdsBytes.length + assignmentsBytes.length);
+
+        buf.putInt(consistentIdsBytes.length);
+        buf.put(consistentIdsBytes);
+        buf.putInt(assignmentsBytes.length);
```
ibessonov, line 105: I feel like this (`buf.putInt(assignmentsBytes.length)`) is unnecessary, you can calculate the remaining size from what you have already read

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -0,0 +1,228 @@
...
+            byte[] b = transform.apply(o);
+            objects.add(b);
+            bytesObjects += b.length;
+        }
+
+        bytesObjects += Short.BYTES * (objects.size() + 1);
+
+        ByteBuffer buf = ByteBuffer.allocate(bytesObjects);
+
+        buf.putShort((short) objects.size());
```
ibessonov, line 159: Like here, there must be a check that the conversion is safe, otherwise the structure is broken

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /**
+     * Serializes collection to bytes.
+     *
+     * @param collection Collection.
+     * @param transform Tranform function for the collection element.
+     * @return Byte array.
+     */
+    private <T> byte[] collectionToBytes(Collection<T> collection, Function<T, byte[]> transform) {
```
sanpwc, line 145: Looks like an utility function.

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
ibessonov: For example, you're assuming that there can't be more than 65k partitions, which will not be true in the future. So we should use int for the header instead of short

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /**
+     * Deserializes the list from byte array.
+     *
+     * @param bytes Byte array.
+     * @param transform Transform function to create list element.
+     * @return List.
+     */
+    private static <T> List<T> bytesToList(byte[] bytes, Function<byte[], T> transform) {
```
sanpwc, line 176: Same as above, looks like an utility function.

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -0,0 +1,228 @@
...
+
+        short length = buf.getShort();
+        assert length >= 0 : "Negative collection size: " + length;
+
+        List<T> result = new ArrayList<>(length);
+
+        for (int i = 0; i < length; i++) {
+            short size = buf.getShort();
+            assert size >= 0 : "Negative object size: " + size + ", index=" + i;
+            byte[] arr = new byte[size];
```
ibessonov, line 187: Again, I recommend using byte buffers and not waste memory

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -0,0 +1,228 @@
...
+            result.add(transform.apply(arr));
+        }
+
+        return result;
+    }
+
+    private static byte[] assignmentToBytes(Assignment assignment, Function<String, Integer> consistentIdToIndex) {
+        ByteBuffer buf = ByteBuffer.allocate(Integer.BYTES + Byte.BYTES);
+        buf.putInt(consistentIdToIndex.apply(assignment.consistentId()));
+        buf.put((byte) (assignment.isPeer() ? 1 : 0));
```
ibessonov, line 198: Seems a bit wasteful. We could encode the "peer" bit in the consistent id itself.
By the way, do we use this class for CMG/Meta-Storage or not? They are replicated, in a nutshell, so this encoding would result in huge payloads anyway

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        Assignments that = (Assignments) o;
+
+        return assignments != null ? assignments.equals(that.assignments) : that.assignments == null;
+    }
+
+    @Override
+    public int hashCode() {
+        return assignments != null ? assignments.hashCode() : 0;
```
ibessonov, line 226: How can it be null?

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -0,0 +1,228 @@
...
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        Assignments that = (Assignments) o;
+
+        return assignments != null ? assignments.equals(that.assignments) : that.assignments == null;
```
ibessonov, line 221: There's a method in `Objects` class that does null-checks. Anyway, the same question, how can it be null?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
denis-chudov: `stableAssignmentsBytes` is taken from different meta storage entries, managed by rebalance. If it is `null`, assignments for partition are taken from table configuration. Serialization was reworked only for table configuration, not for rebalance keys.

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
denis-chudov: thanks, fixed

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * Bytes representation of the assignments.
+     *
+     * @return Bytes representation of the assignments.
+     */
+    public byte[] bytes() {
+        Map<String, Integer> consistentIds = new LinkedHashMap<>();
+
+        int idx = 0;
```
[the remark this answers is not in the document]
denis-chudov: done

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -0,0 +1,228 @@
...
+        int bytesObjects = 0;
+        List<byte[]> objects = new ArrayList<>();
+
+        for (T o : collection) {
+            byte[] b = transform.apply(o);
+            objects.add(b);
+            bytesObjects += b.length;
+        }
+
+        bytesObjects += Short.BYTES * (objects.size() + 1);
```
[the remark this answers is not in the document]
denis-chudov: maybe, if we make this an utility method, as @sanpwc suggested, there should be no such assumption, size should be Integer instead

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
denis-chudov: agree

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
denis-chudov: It can't. Fixed.

PR #2119, opened 2023-05-30
IGNITE-19592 Implement ADD/DROP COLUMN DDL commands with using Catalog

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-19592

### DIFF
--- a/modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
+++ b/modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
@@ -275,8 +275,8 @@ public class CatalogServiceImpl extends Producer<CatalogEvent, CatalogEventParam
 
             Arrays.stream(schema.indexes())
                     .filter(index -> index.tableId() == table.id())
-                    .forEach(index -> index.columns().stream()
-                            .filter(col -> params.columns().contains(col))
+                    .forEach(index -> params.columns().stream()
+                            .filter(index::hasColumn)
                             .findAny()
                             .ifPresent(columnName -> {
                                 throw new IllegalArgumentException("Can't drop indexed column: columnName=" + columnName + ", indexName=" +
--- a/modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
+++ b/modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
@@ -65,9 +65,11 @@ import org.apache.ignite.internal.util.PendingComparableValuesTracker;
 import org.apache.ignite.lang.ColumnAlreadyExistsException;
 import org.apache.ignite.lang.ColumnNotFoundException;
 import org.apache.ignite.lang.ErrorGroups.Common;
+import org.apache.ignite.lang.ErrorGroups.Sql;
 import org.apache.ignite.lang.IgniteInternalException;
 import org.apache.ignite.lang.TableAlreadyExistsException;
 import org.apache.ignite.lang.TableNotFoundException;
+import org.apache.ignite.sql.SqlException;
 import org.jetbrains.annotations.Nullable;
 
 /**
@@ -269,7 +271,10 @@ public class CatalogServiceImpl extends Producer<CatalogEvent, CatalogEventParam
                     throw new ColumnNotFoundException(columnName);
                 }
                 if (table.isPrimaryKeyColumn(columnName)) {
-                    throw new IllegalArgumentException("Can't drop primary key column: column=" + columnName);
+                    throw new SqlException(
+                            Sql.DROP_IDX_COLUMN_CONSTRAINT_ERR,
+                            "Can't drop primary key column: column=" + columnName
+                    );
                 }
             }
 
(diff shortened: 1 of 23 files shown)


modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -205,13 +215,84 @@ public CompletableFuture<Void> dropTable(DropTableParams params) {
...
+
+            if (table == null) {
+                throw new TableNotFoundException(schemaName, params.tableName());
+            }
+
+            List<TableColumnDescriptor> columnDescriptors = new ArrayList<>();
+
+            for (ColumnParams col : params.columns()) {
+                if (table.column(col.name()) != null) {
+                    throw new ColumnAlreadyExistsException(col.name());
```
xtern, line 237: - From the user's point of view, I would expect that for a table with an existing column "column1" the following command "ADD COLUMN IF NOT EXISTS (column1 int, column2 int)" will silently create 'column2'.
But as far as I understand the current behavior, we just ignore the exception. It seems to me that it's better to completely forbid 'IF NOT EXISTS' options for multiple columns if we can't support this (as was in previous PR).
- Why we need `ifColumnNotExists` field in `AlterTableAddColumnParams` if we don't use it?
- Also, current behavior for ADD COLUMN differs from DROP COLUMN, which seems strange to me.

I also don't quite understand - why we can't return an empty list here to prevent catalog version change?

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -205,13 +215,84 @@ public CompletableFuture<Void> dropTable(DropTableParams params) {
...
+                throw new TableNotFoundException(schemaName, params.tableName());
+            }
+
+            Arrays.stream(schema.indexes())
+                    .filter(index -> index.tableId() == table.id())
+                    .flatMap(index -> index.columns().stream())
+                    .filter(col -> params.columns().contains(col))
+                    .findAny()
+                    .ifPresent(columnName -> {
+                        throw new IllegalArgumentException("Can't drop indexed column: column=" + columnName);
```
xtern, line 275: It seems to me that in such a situation it would be interesting for a "simple user" :roll_eyes:  to know which index uses this column :thinking:

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -205,13 +215,84 @@ public CompletableFuture<Void> dropTable(DropTableParams params) {
...
+                    });
+
+            for (String columnName : params.columns()) {
+                if (table.column(columnName) == null) {
+                    throw new ColumnNotFoundException(columnName);
+                }
+                if (table.isPrimaryKeyColumn(columnName)) {
```
xtern, line 282: Since the primary key column is always indexed column - this condition will be hidden by the previous check. I suggest to move it upper.

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -205,13 +215,84 @@ public CompletableFuture<Void> dropTable(DropTableParams params) {
...
     }
 
     /** {@inheritDoc} */
     @Override
     public CompletableFuture<Void> dropColumn(AlterTableDropColumnParams params) {
-        return failedFuture(new UnsupportedOperationException("Not implemented yet."));
+        if (params.columns().isEmpty()) {
+            return completedFuture(null);
+        } else if (params.columns().size() > 1 && params.ifColumnExists()) {
+            return failedFuture(new UnsupportedOperationException("Clause 'IF NOT EXISTS' is not supported when adding multiple columns."));
```
xtern, line 255: current behavior for ADD COLUMN differs from DROP COLUMN, which seems strange to me. From my point of view it's better to keep the same for `ADD COLUMN` :roll_eyes:

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/descriptors/IndexDescriptor.java
```diff
@@ -17,6 +17,8 @@
 
 package org.apache.ignite.internal.catalog.descriptors;
 
+import java.util.List;
```
korlov42, line 20: the same: index-related changes should be moved to IGNITE-19460

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -205,13 +215,83 @@ public CompletableFuture<Void> dropTable(DropTableParams params) {
...
+            if (table == null) {
+                throw new TableNotFoundException(schemaName, params.tableName());
+            }
+
+            for (String columnName : params.columns()) {
+                if (table.column(columnName) == null) {
+                    throw new ColumnNotFoundException(columnName);
+                }
+                if (table.isPrimaryKeyColumn(columnName)) {
+                    throw new IllegalArgumentException("Can't drop primary key column: column=" + columnName);
```
korlov42, line 272: let's throw IgniteInternalException instead of IllegalArgumentException

korlov42 (approved):

PR #2120, opened 2023-05-30
IGNITE-19053 Handle the case of changing the leader of the raft group when building the index

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19053


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItBuildIndexTest.java
```diff
@@ -46,13 +65,103 @@ public class ItBuildIndexTest extends ClusterPerClassIntegrationTest {
...
+        // Let's make sure that the indexes are eventually built.
+        checkIndexBuild(1, nodes(), INDEX_NAME);
+    }
+
+    /**
+     * Prepares an index build for a primary replica change.
+     * <ul>
+     *     <li>Creates a table (replicas = {@link #nodes()}, partitions = 1) and populates it;</li>
+     *     <li>Creates an index;</li>
+     *     <li>Drop send {@link BuildIndexCommand} from the primary replica.</li>
```
ibessonov, line 127:
```suggestion
     *     <li>Drops {@link BuildIndexCommand} messages from the primary replica.</li>
```

ibessonov (approved):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItBuildIndexTest.java
tkalkirill: I'll leave it as is.

PR #2128, opened 2023-06-01
IGNITE-19540 Add Basic Data Streamer

ptupitsyn:
Add basic data streamer (without receiver) to Java client (Record and KV views).

This covers the first use case from https://cwiki.apache.org/confluence/display/IGNITE/IEP-102%3A+Data+Streamer.

* No changes to client protocol: `UPSERT_ALL` is used to upsert batches.
* Refactor `PartitionAwarenessProvider` to use `nodeId` instead of `channel`, so we can reuse per-node buffers after reconnect.
* The bulk of the batching and streaming logic is in `StreamerSubscriber` and `StreamerBuffer`.

https://issues.apache.org/jira/browse/IGNITE-19540


modules/client/src/main/java/org/apache/ignite/internal/client/table/StreamerBuffer.java
```diff
@@ -0,0 +1,98 @@
...
+
+        return flushFut == null ? CompletableFuture.completedFuture(null) : flushFut;
+    }
+
+    synchronized void flush(long period) {
+        if (closed || buf.isEmpty()) {
+            return;
+        }
+
+        if (System.currentTimeMillis() - lastFlushTime > period) {
```
isapego, line 82: AFAIK, `System.currentTimeMillis()` is not monotonic, so there may be an issue (for example, when system time is changed 1 hour due to seasonal time change). On the other hand, I'm not sure there are any monotonic clocks in Java at all, though some says that `System.nanoTime()` is monotonic on some platforms.

isapego (approved):

modules/api/src/main/java/org/apache/ignite/table/DataStreamerOptions.java
```diff
@@ -0,0 +1,176 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.table;
+
+/**
+ * Data streamer options. See {@link DataStreamerTarget} for more information.
+ */
+public class DataStreamerOptions {
```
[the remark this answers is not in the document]
ptupitsyn: Hmm, we have `TransactionOptions`, but that was also introduced by me. I though `Configuration` is for things like node configuration. And `Options` betters suits runtime APIs like `Transactions` and `Streamer`. But it looks like those terms are interchangeable and we might want to use `Configuration` everywhere, as you say.

modules/api/src/main/java/org/apache/ignite/table/DataStreamerOptions.java
```diff
@@ -0,0 +1,176 @@
...
+     * Builder.
+     */
+    public static class Builder {
+        private int batchSize = 1000;
+
+        private int perNodeParallelOperations = 4;
+
+        private int autoFlushFrequency = 5000;
+
+        private int retryLimit = 16;
```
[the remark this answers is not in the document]
ptupitsyn: No, we can discuss here I guess?

modules/client/src/main/java/org/apache/ignite/internal/client/table/StreamerBuffer.java
ptupitsyn: Very good catch. Replaced with `nanoTime`.

modules/client/src/main/java/org/apache/ignite/internal/client/table/ClientTable.java
```diff
@@ -72,9 +72,9 @@ public class ClientTable implements Table {
 
     private final Object partitionAssignmentLock = new Object();
 
-    private CompletableFuture<List<String>> partitionAssignment = null;
+    private volatile CompletableFuture<List<String>> partitionAssignment = null;
 
-    private long partitionAssignmentVersion = -1;
+    private volatile long partitionAssignmentVersion = -1;
```
zstan, line 77: why do you append volatile ? as i can see all usage is already under monitor.

ptupitsyn: `getPartitionAssignment` uses double-checked locking, which requires `volatile`.

PR #2131, opened 2023-06-01
IGNITE-19559 NPE in deploy/undeploy calls in non-REPL mode

valepakh:
https://issues.apache.org/jira/browse/IGNITE-19559


modules/cli/src/main/java/org/apache/ignite/internal/cli/call/cluster/unit/DeployUnitReplCall.java
```diff
@@ -0,0 +1,45 @@
...
+package org.apache.ignite.internal.cli.call.cluster.unit;
+
+import java.util.concurrent.CompletableFuture;
+import org.apache.ignite.internal.cli.core.call.CallOutput;
+import org.apache.ignite.internal.cli.core.call.ProgressTracker;
+import org.apache.ignite.internal.cli.core.repl.registry.UnitsRegistry;
+import org.apache.ignite.internal.cli.core.rest.ApiClientFactory;
+
+/** Call to deploy a unit and refresh units registry. */
+public class DeployUnitReplCall extends DeployUnitCall {
```
PakhomovAlexander, line 27: Why don't use an aggregation? You can pass `DeployUnitCall` to the constructor. And if you want to use polymorphism, it's better to declare an interface `DeployUnitCall ` and have two separate implementations like  `DeployUnitReplCall` and `DeployUnitNonReplCall`.

modules/cli/src/main/java/org/apache/ignite/internal/cli/call/cluster/unit/UndeployUnitCallFactory.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private final ApiClientFactory factory;
+
+    public UndeployUnitCallFactory(ApiClientFactory factory) {
+        this.factory = factory;
+    }
+
+    @Bean
+    @Singleton
+    public UndeployUnitCall create() {
```
PakhomovAlexander, line 37: its better to call factory methods the same name you would like to name a bean. Here `undeployUnitCall() ` is ok.

modules/cli/src/main/java/org/apache/ignite/internal/cli/call/cluster/unit/UndeployUnitReplCallFactory.java
```diff
@@ -0,0 +1,44 @@
...
+    private final UnitsRegistry registry;
+
+    public UndeployUnitReplCallFactory(ApiClientFactory factory, UnitsRegistry registry) {
+        this.factory = factory;
+        this.registry = registry;
+    }
+
+    @Bean
+    @Singleton
+    public UndeployUnitReplCall create() {
```
PakhomovAlexander, line 41: same for naming

PR #2134, opened 2023-06-02
IGNITE-19596 Partition creation fails with NSFE

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19596

### DIFF
--- a/modules/page-memory/src/test/java/org/apache/ignite/internal/pagememory/persistence/store/FilePageStoreManagerTest.java
+++ b/modules/page-memory/src/test/java/org/apache/ignite/internal/pagememory/persistence/store/FilePageStoreManagerTest.java
@@ -17,6 +17,7 @@
 
 package org.apache.ignite.internal.pagememory.persistence.store;
 
+import static java.util.concurrent.TimeUnit.MINUTES;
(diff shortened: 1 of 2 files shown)


modules/page-memory/src/test/java/org/apache/ignite/internal/pagememory/persistence/store/FilePageStoreManagerTest.java
```diff
@@ -440,13 +449,68 @@ void testFullyRemovePartitionOnStart() throws Exception {
         );
     }
 
+    @ParameterizedTest
+    @ValueSource(ints = {1, 2, 4, 10, 25, 50})
```
rpuch, line 453: Let's add a sane time out (like `@Timeout(value = 1, unit = MINUTE)`) just in case as racers use potentially infinite loops (if something gets broken in the future, they could hang forever)

tkalkirill: Why not

rpuch (approved):

PR #2142, opened 2023-06-02
IGNITE-19494 Correctly stop a replica

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-19494

### DIFF
--- a/modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
+++ b/modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
@@ -214,7 +214,7 @@ public class StripedDisruptor<T extends NodeIdAware> {
             assert handler != null : format("Group of the event is unsupported [nodeId={}, event={}]", event.nodeId(), event);
 
             //TODO: IGNITE-15568 endOfBatch should be set to true to prevent caching tasks until IGNITE-15568 has fixed.
-            handler.onEvent(event, sequence, true);
+            handler.onEvent(event, sequence, subscribers.size() > 1 ? true : endOfBatch);
         }
     }
 
--- a/modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
+++ b/modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
@@ -394,7 +394,6 @@ public class ReplicaManager implements IgniteComponent {
      * @param replicaGrpId Replication group id.
      * @return True if the replica is found and closed, false otherwise.
      */
-    // TODO: IGNITE-19494 We need to correctly stop the replica
     private boolean stopReplicaInternal(ReplicationGroupId replicaGrpId) {
         CompletableFuture<Replica> removed = replicas.remove(replicaGrpId);
 
--- a/modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
+++ b/modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
@@ -403,12 +403,10 @@ public class ReplicaManager implements IgniteComponent {
                         replicaGrpId,
                         clusterNetSvc.topologyService().localMember()
                 ));
-            } else {
-                removed.whenComplete((replica, throwable) -> {
-                    if (throwable != null) {
-                        replica.shutdown();
-                    }
-                });
+            }
+
+            if (!removed.isCompletedExceptionally()) {
+                removed.join().shutdown();
             }
 
             return true;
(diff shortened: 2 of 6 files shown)


modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
```diff
@@ -214,7 +214,7 @@ void unsubscribe(NodeId nodeId) {
             assert handler != null : format("Group of the event is unsupported [nodeId={}, event={}]", event.nodeId(), event);
 
             //TODO: IGNITE-15568 endOfBatch should be set to true to prevent caching tasks until IGNITE-15568 has fixed.
-            handler.onEvent(event, sequence, true);
+            handler.onEvent(event, sequence, subscribers.size() > 1 ? true : endOfBatch);
```
sanpwc, line 217: Is it an intended change?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/exception/ReplicaStoppingException.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * The constructor is used for creating an exception instance that is thrown from a remote server.
+     *
+     * @param traceId Trace id.
+     * @param code Error code.
+     * @param message Error message.
+     * @param cause Cause exception.
+     */
+    public ReplicaStoppingException(UUID traceId, int code, String message, Throwable cause) {
```
sanpwc, line 48: Where do we use such constructor? Within deserialization process?

modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
vldpyatkov: Not intended, but I decided to leave it after commit. Because this allows to do slightly better of disruptor for Meta storage and Cluster management group and I don't want to create dedicate ticket for this small change.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/exception/ReplicaStoppingException.java
vldpyatkov: I think, any exception should have constructor like this, even if it is not used in network just now.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -387,11 +399,18 @@ private boolean stopReplicaInternal(ReplicationGroupId replicaGrpId) {
         CompletableFuture<Replica> removed = replicas.remove(replicaGrpId);
 
         if (removed != null) {
-            removed.whenComplete((replica, throwable) -> {
-                if (throwable != null) {
-                    replica.shutdown();
-                }
-            });
+            if (!removed.isDone()) {
```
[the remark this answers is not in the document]
vldpyatkov: Removed this note.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
sanpwc: Sounds reasonable, however I had to ask. How did you understand that it works better after given change?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -387,11 +399,18 @@ private boolean stopReplicaInternal(ReplicationGroupId replicaGrpId) {
...
+                removed.completeExceptionally(new ReplicaStoppingException(
```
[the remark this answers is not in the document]
sanpwc: I'd still call shutdown just in case. Seems that it should be safe, what do you think?

modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
vldpyatkov: We don't have a test that will be able to demonstrate that. I assume, that update is not doing worse at least.
Theoretically, RAFT log may fsync less often in the circumstance when several RAFT commands are executing simultaniously.

alievmirza: I don't think that this change is reasonable, because we have a plan to port disruptor changes in a upcoming sprint, potentially this fix will live a few weeks

vldpyatkov: We will work a little better just now. Today, the change worth nothing for us.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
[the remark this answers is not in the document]
vldpyatkov: I changed the behavior and add shutdown invocation in any case when the future is not completed with exception.

sanpwc (approved):

alievmirza (approved):

PR #2152, opened 2023-06-06
IGNITE-19522 Modify deploy unit command

valepakh:
https://issues.apache.org/jira/browse/IGNITE-19522


modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/Options.java
```diff
@@ -240,6 +241,15 @@ public static final class Constants {
         /** Unit path option description. */
         public static final String UNIT_PATH_OPTION_DESC = "Path to deployment unit file or directory";
 
+        /** Unit nodes option long name. */
+        public static final String UNIT_NODES_OPTION = "--nodes";
+
+        /** Unit nodes option short name. */
+        public static final String UNIT_NODES_OPTION_SHORT = "-un";
```
PakhomovAlexander, line 248: if I am not mistaken, we decided not to use the two-letter symbols for short options. It might be confusing to understand if it means -u -n or -un.

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/unit/ItDeploymentUnitTest.java
```diff
@@ -156,6 +156,98 @@ void deployDirectory() {
...
+            // Unit is deployed on all requested nodes
+            assertAll(
+                    this::assertExitCodeIsZero,
+                    this::assertErrOutputIsEmpty,
+                    () -> assertOutputContains("1.0.0"),
+                    () -> assertOutputContains("DEPLOYED")
+            );
+
+            resetOutput();
+            execute("node", "unit", "list", "--node-url", "http://localhost:10301", "test.unit.id.7");
```
PakhomovAlexander, line 186: could you check that it is also deployed to the CMG group?

modules/code-deployment/src/main/java/org/apache/ignite/internal/deployunit/DeploymentManagerImpl.java
```diff
@@ -151,32 +155,72 @@ private void onUnitRegister(UnitNodeStatus status, Set<String> deployedNodes) {
...
+            List<String> initialNodes
+    ) {
+        checkId(id);
+        Objects.requireNonNull(version);
+        Objects.requireNonNull(deploymentUnit);
+
+        return extractNodes(initialNodes)
+                .thenCompose(nodesToDeploy ->
+                        doDeploy(id, version, force, deploymentUnit, nodesToDeploy,
+                                undeployed -> deployAsync(id, version, deploymentUnit, initialNodes)
```
PakhomovAlexander, line 192: Looks like a recursion, am I right?

modules/code-deployment/src/main/java/org/apache/ignite/internal/deployunit/IgniteDeployment.java
```diff
@@ -28,29 +28,80 @@
...
+     * @param version Unit version.
+     * @param deploymentUnit Unit content.
+     * @param initialNodes List of nodes to deploy to initially.
+     * @return Future with success or not result.
+     */
+    default CompletableFuture<Boolean> deployAsync(
+            String id,
+            Version version,
+            DeploymentUnit deploymentUnit,
+            List<String> initialNodes
```
PakhomovAlexander, line 82: Maybe just `nodes` is enogh?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/raft/CmgRaftService.java
```diff
@@ -228,6 +228,38 @@ public Set<String> nodeNames() {
...
+
+        int peersCount = peers.size();
+        String leaderId = leader.consistentId();
+
+        // Take half of the voting peers without the leader.
+        Set<String> result = peers.stream()
+                .map(Peer::consistentId)
+                .filter(consistentId -> !consistentId.equals(leaderId))
+                .limit(peersCount / 2)
+                .collect(toSet());
```
sashapolo, line 255: It is generally a bad idea to modify a set, returned by `toSet()`, there's no guarantee on its mutability

sashapolo: Also, why is it a set and not a list?

modules/rest-api/openapi/openapi.yaml
```diff
@@ -469,29 +469,42 @@ paths:
...
+          type: string
+          description: The ID of the deployment unit.
+      - name: unitVersion
+        in: path
+        required: true
+        schema:
+          type: string
+          description: The version of the deployment unit.
+      - name: initialNodes
+        in: query
```
PakhomovAlexander, line 493: I think it might be moved to the body

PR #2154, opened 2023-06-06
IGNITE-19642 CatalogService should use HybridClock to register new Catalog version

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-19642


lowka (approved):

AMashenkov:
@rpuch , could you please review the PR.

rpuch:
Sure, I'll do it tomorrow

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -519,7 +536,7 @@ public void handle(VersionedUpdate update) {
                 if (entry instanceof NewTableEntry) {
                     catalog = new Catalog(
                             version,
-                            System.currentTimeMillis(),
+                            activationTimestamp(),
```
rpuch, line 625: If I'm not mistaken, `UpdateLog` acts like a WAL, and this code executes the commands contained in the `UpdateLog`.

1. It looks like a command might be executed a few times during the lifetime of the cluster (and probably it will be executed on each node separately), so same schema update might get different activation ts throught its existence. This is not right: each schema update must have an immutable activation moment, and it must be the same on any node of the cluster.
2. Activation timestamp must be bound to the Metastorage SafeTime; namely, an MS 'message' (from the leader to any cluster node) that transports a schema update must contain the 'message ts' (which, when consumed and fully processed by a node [so, after it makes all modifications to the in-memory Catalog representation], gets assigned to the node's local MS SafeTime variable) and also it must contain the schema update activation ts, and these two timestamps must be connected by the following equality: messageTs+DD=activationTs.

So the activation moment should not be computed here, but in `saveUpdate()` (and it should probably be a part of `VersionedUpdate`).

rpuch (changes requested):
Also, `CatalogServiceImpl:702`

versionTracker.update(version, null);

seems to be redundant as it's already called on line 697.

Another question is about `CatalogService` API: it's still using `long` to accept/pass timestamps, not `HybridTimestamp`. Is it planned to change this? Is it intentional that in the current PR this has not been changed?

rpuch (approved):
It seems that there still remains an issue that activation time in the MS entry is not strictly equal to MS entry ts+DD, but it's tricky to implement this and this probably has to be done in another ticket, probably, in https://issues.apache.org/jira/browse/IGNITE-19209

Could you please add a TODO mentioning this issue so that we don't forget this?

rpuch (approved):

PR #2157, opened 2023-06-06
IGNITE-18692 ItRebalanceTest fix

kgusakov:
https://issues.apache.org/jira/browse/IGNITE-18692


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -565,11 +571,17 @@ private CompletableFuture<?> onTableCreate(ConfigurationNotificationEvent<TableV
...
                     getZoneById(distributionZonesConfiguration, (ctx.newValue()).zoneId()).value();
 
-            List<Set<Assignment>> assignments = AffinityUtils.calculateAssignments(
-                    // TODO: https://issues.apache.org/jira/browse/IGNITE-19425 use data nodes from DistributionZoneManager instead.
-                    baselineMgr.nodes().stream().map(ClusterNode::name).collect(toList()),
-                    zone.partitions(),
-                    zone.replicas());
+            List<Set<Assignment>> assignments;
+
+            if (partitionAssignments(vaultManager, ctx.newValue().id(), 0) != null) {
```
sergeyuttsel, line 576: Looks like some kind of magic. Please add a comment if possible. Is this a temporary or permanent solution? Getting nodes from zone manager will fix this or not?

kgusakov: Added a comment. Not sure about the temporary or not, but in any case - if we already has assignments we shouldn't recalculate it again on recovery.

sergeyuttsel: Thanks

sergeyuttsel (approved):

vldpyatkov (approved):

PR #2162, opened 2023-06-07
IGNITE-19493 Sql. Change query execution flow

zstan:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -425,6 +421,18 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
                 .thenCompose(sqlNode -> {
                     boolean rwOp = dataModificationOp(sqlNode);
 
+                    boolean implicitTxRequired = outerTx == null;
+
+                    tx.set(implicitTxRequired ? txManager.begin(!rwOp) : outerTx);
+
+                    SchemaPlus schema = sqlSchemaManager.schema(schemaName);
```
AMashenkov, line 428:
```suggestion
                    SchemaPlus schema = sqlSchemaManager.activeSchema(schemaName, tx.startTimestamp().longValue());
```

AMashenkov: SqlSchemaManager methods without version or timestamp must not be used, and it is better to drop them at all.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -425,6 +421,18 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
...
+
+                    if (schema == null) {
+                        return CompletableFuture.failedFuture(
+                                new IgniteInternalException(SCHEMA_NOT_FOUND_ERR,
```
AMashenkov, line 432: Let's introduce a top-level class SchemaNotFoundException in public API.

modules/runner/src/integrationTest/sql/types/timestamp/test_incorrect_timestamp.test
```diff
@@ -1,7 +1,6 @@
 # name: test/sql/types/timestamp/test_incorrect_timestamp.test
 # description: Test out of range/incorrect timestamp formats
 # group: [timestamp]
-# Ignore https://issues.apache.org/jira/browse/IGNITE-15623
```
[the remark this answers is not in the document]
zstan: yep, but can it be here ? issue already fixed but mention still exist

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
zstan: done

AMashenkov (approved):

PR #2163, opened 2023-06-07
IGNITE-19451 Prohibit some types from being Marshallable

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-19451

### DIFF
--- a/modules/network-annotation-processor/src/main/java/org/apache/ignite/internal/network/processor/messages/MarshallableTypesBlackList.java
+++ b/modules/network-annotation-processor/src/main/java/org/apache/ignite/internal/network/processor/messages/MarshallableTypesBlackList.java
@@ -30,6 +30,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.ArrayType;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeMirror;
@@ -89,7 +90,7 @@ public class MarshallableTypesBlackList {
     private static class TypeVisitor extends SimpleTypeVisitor9<Boolean, Void> {
         private final TypeUtils typeUtils;
 
-        private final List<Class<?>> resourcesBlacklist;
+        private final List<TypeMirror> resourcesBlacklist;
 
         TypeVisitor(TypeUtils typeUtils) {
             super(false);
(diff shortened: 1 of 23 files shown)


modules/network-annotation-processor/src/main/java/org/apache/ignite/internal/network/processor/messages/MarshallableTypesBlackList.java
```diff
@@ -97,11 +126,17 @@ public Boolean visitDeclared(DeclaredType t, Void unused) {
                 return t.getTypeArguments().stream().anyMatch(this::visit);
             }
 
-            return !isSameType(NATIVE_TYPES, t) && !typeUtils.isSubType(t, NetworkMessage.class);
+            return !isSameType(NATIVE_TYPES, t)
+                    && !typeUtils.isSubType(t, NetworkMessage.class)
```
ibessonov, line 130: Would it make sense to implement NetworkMassage and default types as a blacklist file themselves?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/IgniteTableImpl.java
```diff
@@ -369,6 +370,16 @@ public <RowT> CompletableFuture<?> upsertAll(
         return CompletableFuture.allOf(futures);
     }
 
+    private static List<ByteBuffer> serializeBinaryRows(Collection<BinaryRow> rows) {
+        var result = new ArrayList<ByteBuffer>(rows.size());
+
+        for (BinaryRow row : rows) {
+            result.add(row.byteBuffer());
```
ibessonov, line 377: In the future,  I guess we would like to remove this "byteBuffer" method. In that case it would be more convenient to have specific BinaryRowMessage class, or something similar, that would have schema version and the payload in different fields.
Should we do it right now, or let's wait until the method is removed?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -1429,6 +1454,7 @@ public void close() {
     }
 
     // TODO: IGNITE-17963 Use smarter logic for recipient node evaluation.
+
```
ibessonov, line 1457: What's the meaning of this empty line and why the next javadoc is re-formatted? Was that done by IDEA?

ibessonov (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
sashapolo: yes, I've applied auto-formatting

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/IgniteTableImpl.java
ibessonov: Because what's the point of making a copy of an array if all we want is to slap a single integer to it. That's extra copying that most likely affects the performance in negative way.

PR #2164, opened 2023-06-07
IGNITE-17765 Sql. Avoid parsing queries that already have cached plans.

xtern:
https://issues.apache.org/jira/browse/IGNITE-17765

InsertBenchmark on local machine shows the following results:

|cache type| disabled | query plan optimization |query parse + plan optimization|
| --- | ----------- |-|-|
| ms/op | 2.441 | 0.924 |    0.659 |

### DIFF
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
@@ -410,15 +410,12 @@ public class SqlQueryProcessor implements QueryProcessor {
         AtomicReference<InternalTransaction> tx = new AtomicReference<>();
 
         CompletableFuture<AsyncSqlCursor<List<Object>>> stage = start
-                .thenApply(v -> {
+                .thenCompose(v -> {
                     StatementParseResult parseResult = IgniteSqlParser.parse(sql, StatementParseResult.MODE);
                     SqlNode sqlNode = parseResult.statement();
 
                     validateParsedStatement(context, outerTx, parseResult, sqlNode, params);
 
-                    return sqlNode;
-                })
-                .thenCompose(sqlNode -> {
                     boolean rwOp = dataModificationOp(sqlNode);
 
                     boolean implicitTxRequired = outerTx == null;
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
@@ -17,6 +17,9 @@
 
 package org.apache.ignite.internal.sql.engine;
 
+import static org.apache.ignite.internal.sql.engine.SqlQueryType.DML;
+import static org.apache.ignite.internal.sql.engine.SqlQueryType.QUERY;
+import static org.apache.ignite.internal.sql.engine.prepare.CacheKey.EMPTY_CLASS_ARRAY;
 import static org.apache.ignite.internal.sql.engine.util.Commons.FRAMEWORK_CONFIG;
 import static org.apache.ignite.lang.ErrorGroups.Sql.OPERATION_INTERRUPTED_ERR;
 import static org.apache.ignite.lang.ErrorGroups.Sql.QUERY_INVALID_ERR;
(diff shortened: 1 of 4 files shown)


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -407,70 +423,83 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
...
-                                                return dataCursor.closeAsync();
-                                            }
-                                        }
-                                );
-                            });
+                            .plannerTimeout(PLANNER_TIMEOUT);
+
+                    CompletableFuture<PlanWithContext>[] newPlanHolder = new CompletableFuture[1];
+
+                    CompletableFuture<QueryPlan> cachedPlan = queryCache.computeIfAbsent(new CacheKey(schemaName, sql, params), (k) -> {
```
AMashenkov, line 439: The lamda that is passed to `queryCache.computeIfAbsent` is run inside the synchronized block. For queries, which shouldn't be cached (like DDL), this maybe overkill.
The lamda that is passed to `queryCache.computeIfAbsent` is run inside the synchronized block. For queries, which shouldn't be cached (like DDL), this maybe overkill.

`queryCache.computeIfAbsent` already returns the future. Thus, there is no need to do any long operation inside the synchronized block.

Instead, you can always return a future from lamda for QUERY and DML and put computation to another thread. For non-QUERY or non-DML just return null and do computation in the current thread. Or maybe avoid using compute at all.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -407,70 +423,83 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
...
+                        // Parse query.
+                        StatementParseResult parseResult = IgniteSqlParser.parse(sql, StatementParseResult.MODE);
+                        SqlNode sqlNode = parseResult.statement();
+
+                        validateParsedStatement(context, outerTx, parseResult, sqlNode, params);
+
+                        // Build context.
+                        tx.set(implicitTxRequired ? txManager.begin(!dataModificationOp(sqlNode)) : outerTx);
```
AMashenkov, line 447: Let's add an assert 
`assert outerTx == null || !outerTx.readOnly()`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/CacheKey.java
```diff
@@ -18,48 +18,34 @@
...
      *
      * @param schemaName Schema name.
      * @param query      Query string.
-     * @param contextKey Optional context key to differ queries with and without/different flags, having an impact on result plan (like
-     *                   LOCAL flag)
-     * @param paramTypes Types of all dynamic parameters, no any type can be {@code null}.
+     * @param params     Dynamic parameters.
      */
-    public CacheKey(String schemaName, String query, Object contextKey, Class[] paramTypes) {
+    public CacheKey(String schemaName, String query, Object[] params) {
```
AMashenkov, line 43: Cache key must contains schema version.
Using integer CatalogSchemaDescriptor.id() instead of schemaName would be more optimal.

As of now, you can safely hardcode constants instead of version/id, which we will fix later.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/CacheKey.java
```diff
@@ -80,9 +66,6 @@ public boolean equals(Object o) {
         if (!query.equals(cacheKey.query)) {
             return false;
         }
-        if (!Objects.equals(contextKey, cacheKey.contextKey)) {
-            return false;
-        }
 
         return Arrays.deepEquals(paramTypes, cacheKey.paramTypes);
```
AMashenkov, line 70: Arrays.equals/hashCode can be used instead of deepEquals/deepHashCode.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -244,7 +260,7 @@ public synchronized void start() {
                 busyLock
         );
 
-        sqlSchemaManager.registerListener(prepareSvc);
+        sqlSchemaManager.registerListener(queryCache::clear);
```
AMashenkov, line 263: With versioned schema, there is no need to clear cache. Plans for the old schema versions could be used by time-traveling queries.
Let's avoid this, or add a TODO to drop this line later.

PR #2165, opened 2023-06-07
IGNITE-19674 introduce TestRocksDbKeyValueStorage

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-19674

### DIFF
--- a/modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
+++ b/modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
@@ -117,7 +117,7 @@ public class RocksDbKeyValueStorage implements KeyValueStorage {
     private static final long SYSTEM_REVISION_MARKER_VALUE = 0;
 
     /** Revision key. */
-    protected static final byte[] REVISION_KEY = keyToRocksKey(
+    private static final byte[] REVISION_KEY = keyToRocksKey(
             SYSTEM_REVISION_MARKER_VALUE,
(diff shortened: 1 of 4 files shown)


modules/metastorage/src/testFixtures/java/org/apache/ignite/internal/metastorage/server/TestRocksDbKeyValueStorage.java
```diff
@@ -0,0 +1,60 @@
...
+    @Override
+    public void start() {
+        try {
+            createDb();
+
+            byte[] revision = getData().get(REVISION_KEY);
+
+            if (revision != null) {
+                setRev(bytesToLong(revision));
+            }
```
ibessonov, line 55: You could move this to parent class and simply override "destroyRocksDb" with empty body. This way it would take less code and would be less confusing when it comes to abundance of "TestOnly" methods

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
```diff
@@ -116,7 +117,7 @@ public class RocksDbKeyValueStorage implements KeyValueStorage {
     private static final long SYSTEM_REVISION_MARKER_VALUE = 0;
 
     /** Revision key. */
-    private static final byte[] REVISION_KEY = keyToRocksKey(
+    protected static final byte[] REVISION_KEY = keyToRocksKey(
```
ibessonov, line 120: This no longer needs to be protected

ibessonov (approved):

PR #1729, opened 2023-02-28
IGNITE-18756 Awaiting for data nodes updating in distribution zone manager.

sergeyuttsel:
https://issues.apache.org/jira/browse/IGNITE-18756

PR #2171, opened 2023-06-08
IGNITE-19600 Removed topologyVersionedDataNodes

kgusakov:
https://issues.apache.org/jira/browse/IGNITE-19600


sergeyuttsel (approved):

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerLogicalTopologyEventsTest.java
```diff
@@ -45,6 +48,9 @@ void testMetaStorageKeysInitializedOnStartWhenTopVerEmpty() throws Exception {
 
         distributionZoneManager.start();
 
+        verify(keyValueStorage, timeout(1000).times(1))
```
alievmirza, line 51: why do we need changed in this class?

kgusakov: It was here before the topologyVersioneDataNodes logic https://github.com/apache/ignite-3/pull/1729/files#diff-eee737dc0330b6a383f01df9ee1d1ae3135655204bdb1a261ff874674d4a6a78L190 - I just thought, that the author of this test earlier wanted to check that the call of keyValueStorage occuried during init.

kgusakov: Sorry, wrong test, I think I put it here by mistake, will remove it

PR #2172, opened 2023-06-08
IGNITE-19574 restore global states of DistributionZoneManager after restart

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-19574


modules/runner/src/integrationTest/java/org/apache/ignite/internal/distribution/zones/ItIgniteDistributionZoneManagerNodeRestartTest.java
```diff
@@ -0,0 +1,313 @@
...
+                        .dataNodesAutoAdjustScaleDown(IMMEDIATE_TIMER_VALUE)
+                        .build()
+        ).get(10_000, TimeUnit.MILLISECONDS);
+
+        Set<String> nodes = distributionZoneManager.topologyVersionedDataNodes(
+                1,
+                partialNode.logicalTopology().getLogicalTopology().version()
+        ).get(10_000, TimeUnit.MILLISECONDS);
+
+        assertEquals(Set.of(A, B, C).stream().map(ClusterNode::name).collect(Collectors.toSet()), nodes);
```
sergeyuttsel, line 295: `Set.of(A, B, C).stream().map(ClusterNode::name).collect(Collectors.toSet())`
Need to save it in the variable.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/distribution/zones/ItIgniteDistributionZoneManagerNodeRestartTest.java
```diff
@@ -0,0 +1,313 @@
...
+    public void testLogicalTopologyRestoredAfterRestart() throws Exception {
+        PartialNode partialNode = startPartialNode(0);
+
+        DistributionZoneManager distributionZoneManager = findComponent(partialNode.startedComponents(), DistributionZoneManager.class);
+
+        partialNode.logicalTopology().putNode(A);
+        partialNode.logicalTopology().putNode(B);
+        partialNode.logicalTopology().putNode(C);
+
+        distributionZoneManager.createZone(
```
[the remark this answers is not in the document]
alievmirza: good catch, fixed!

modules/runner/src/integrationTest/java/org/apache/ignite/internal/distribution/zones/ItIgniteDistributionZoneManagerNodeRestartTest.java
```diff
@@ -0,0 +1,313 @@
...
+
+        Map<String, Map<String, String>> nodeAttributesAfterRestart = distributionZoneManager.nodesAttributes();
+
+        assertEquals(3, nodeAttributesAfterRestart.size());
+
+        assertEquals(nodeAttributesBeforeRestart, nodeAttributesAfterRestart);
+    }
+
+    @Test
+    public void testLogicalTopologyRestoredAfterRestart() throws Exception {
```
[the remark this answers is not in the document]
alievmirza: Test was simplified, please check

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZonesUtil.java
```diff
@@ -73,6 +73,12 @@ public class DistributionZonesUtil {
     /** Key prefix for zones' logical topology nodes and logical topology version. */
     private static final String DISTRIBUTION_ZONES_LOGICAL_TOPOLOGY_PREFIX = "distributionZones.logicalTopology.";
 
+    /** Key value for zones' nodes' attributes in vault. */
+    private static final String DISTRIBUTION_ZONES_NODES_ATTRIBUTES_VAULT = "vault.distributionZones.nodesAttributes";
+
+    /** Key prefix for zones' logical topology nodes in vault. */
+    private static final String DISTRIBUTION_ZONES_LOGICAL_TOPOLOGY_VAULT = "vault.distributionZones.logicalTopology.nodes";
```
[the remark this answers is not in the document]
alievmirza: Thanks, changed

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1340,6 +1349,26 @@ public void onError(Throwable e) {
         };
     }
 
+    /**
+     * Saves logical topology and nodes' attributes map to vault atomically in one batch.
+     * After restart it could be used to restore these fields.
+     *
+     * @param newLogicalTopology Logical topology.
+     */
+    private void saveLogicalTopologyNodeAttributesToVault(Set<NodeWithAttributes> newLogicalTopology) {
```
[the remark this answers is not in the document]
alievmirza: renamed

sergeyuttsel (approved):

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1027,44 +1030,52 @@ private void initDataNodesFromVaultManager() {
...
 
-            VaultEntry topologyEntry = vaultMgr.get(zonesLogicalTopologyKey()).join();
-
-            if (topologyEntry != null && topologyEntry.value() != null) {
-                assert  appliedRevision > 0 : "The meta storage last applied revision is 0 but the logical topology is not null.";
-
-                logicalTopology = fromBytes(topologyEntry.value());
-
-                logicalTopology.forEach(n -> nodesAttributes.put(n.nodeId(), n.nodeAttributes()));
+            restoreGlobalStateFromVault();
```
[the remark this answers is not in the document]
sanpwc: I've tried to understand what `initLogicalTopologyAndVersionInMetaStorageOnStart` should do though it's javadoc and failed with this.
```
    /**
     * Initialises {@link DistributionZonesUtil#zonesLogicalTopologyKey()} and
     * {@link DistributionZonesUtil#zonesLogicalTopologyVersionKey()} from meta storage on the start of {@link DistributionZoneManager}.
     */
```
What do you mean by `Initialises {@link DistributionZonesUtil#zonesLogicalTopologyKey()}` for example?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1027,44 +1030,52 @@ private void initDataNodesFromVaultManager() {
...
+
+        VaultEntry nodeAttributesEntry = vaultMgr.get(zonesNodesAttributesVault()).join();
+
+        if (topologyEntry != null && topologyEntry.value() != null) {
+            assert nodeAttributesEntry != null : "Nodes' attributes cannot be null when logical topology is not null.";
+            assert nodeAttributesEntry.value() != null : "Nodes' attributes cannot be null when logical topology is not null.";
+
+            logicalTopology = fromBytes(topologyEntry.value());
+
+            nodesAttributes = fromBytes(nodeAttributesEntry.value());
```
sanpwc, line 1067: What kind of Map class will be resolved from fromBytes()?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1119,9 +1130,8 @@ public CompletableFuture<Void> onUpdate(WatchEvent evt) {
                                     .map(NodeWithAttributes::node)
                                     .collect(toSet());
 
-                    newLogicalTopology.forEach(n -> nodesAttributes.put(n.nodeId(), n.nodeAttributes()));
+                    updateLogicalTopologyNodeAttributesAndSaveToVault(newLogicalTopology);
```
sanpwc, line 1133: So, as a result we may finish with an inconsistent vault state, if we will fail right after `updateLogicalTopologyNodeAttributesAndSaveToVault` when applied revision will be x but logicalTopology and node attributes will match x + 1. It's not a problem right?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/BaseIgniteRestartTest.java
```diff
@@ -0,0 +1,369 @@
...
+import org.intellij.lang.annotations.Language;
+import org.jetbrains.annotations.Nullable;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.TestInfo;
+
+/**
+ * Base class for node's restart tests.
+ */
+public abstract class BaseIgniteRestartTest extends IgniteAbstractTest {
```
sanpwc, line 68: So, you are going to extend given class by all node restart tests, is that correct? What about `ItIgniteInMemoryNodeRestartTest`?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/BaseIgniteRestartTest.java
```diff
@@ -0,0 +1,369 @@
...
+    /** Default node port. */
+    protected static final int DEFAULT_NODE_PORT = 3344;
+
+    @Language("HOCON")
+    protected static final String RAFT_CFG = "{\n"
+            + "  fsync: false,\n"
```
sanpwc, line 74: Why? AFAIK default is true.

alievmirza: I haven't changed the defaults from the original `ItIgniteNodeRestartTest`, just moved to abstract class

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
alievmirza: Thanks for noticing! IMHO right behaviour here is to call `updateLogicalTopologyNodeAttributesAndSaveToVault` right after the WatchListener was successfully handled. In that case, your case won't occur, so data in vault will be consistent with `WatchListener`.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/BaseIgniteRestartTest.java
alievmirza: Extended

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
```diff
@@ -113,7 +116,12 @@ public void setUp() throws Exception {
 
         // Mock logical topology for distribution zone.
         vaultMgr = new VaultManager(new InMemoryVaultService());
-        assertThat(vaultMgr.put(zonesLogicalTopologyKey(), toBytes(nodes)), willCompleteSuccessfully());
+        assertThat(vaultMgr.put(zonesLogicalTopologyVault(), toBytes(nodes)), willCompleteSuccessfully());
+
+        Map<String, Map<String, String>> nodesAttributes = new HashMap<>();
```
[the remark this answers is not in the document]
alievmirza: changed to CHM

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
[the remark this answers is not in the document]
alievmirza: `initLogicalTopologyAndVersionInMetaStorageOnStart` was removed

modules/runner/src/integrationTest/java/org/apache/ignite/internal/BaseIgniteRestartTest.java
sanpwc: Ok, I've forwarded given message to @ibessonov. My point here, is that you may need fsync in restart tests.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -946,81 +955,11 @@ private void updateLogicalTopologyInMetaStorage(LogicalTopologySnapshot newTopol
         }
     }
 
-    /**
-     * Initialises {@link DistributionZonesUtil#zonesLogicalTopologyKey()} and
-     * {@link DistributionZonesUtil#zonesLogicalTopologyVersionKey()} from meta storage on the start of {@link DistributionZoneManager}.
-     */
-    private void initLogicalTopologyAndVersionInMetaStorageOnStart() {
```
sanpwc, line 953: So, in order to remove given method and still preserve all guarantees you've added
```
            if (newTopology.version() == 1) {
                // Very first start of the cluster, so we just initialize zonesLogicalTopologyVersionKey
                updateCondition = notExists(zonesLogicalTopologyVersionKey());
            }
```
Is that correct?
Did you add anything else besides aforementioned code?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/BaseIgniteRestartTest.java
sanpwc: My fault, fsync: false is fine here as in all other tests actually.

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerConfigurationChangesTest.java
```diff
@@ -113,7 +116,12 @@ public void setUp() throws Exception {
 
         // Mock logical topology for distribution zone.
         vaultMgr = new VaultManager(new InMemoryVaultService());
-        assertThat(vaultMgr.put(zonesLogicalTopologyKey(), toBytes(nodes)), willCompleteSuccessfully());
+        assertThat(vaultMgr.put(zonesLogicalTopologyVault(), toBytes(nodes)), willCompleteSuccessfully());
```
sanpwc, line 119: Here and in other tests, what about `zonesGlobalStateRevision`, should we also mock it?

sanpwc (approved):

PR #2173, opened 2023-06-09
IGNITE-19690 Get rid of configuration in MvTableStorage and related classes

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19690

### DIFF
--- a/modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestStorageEngine.java
+++ b/modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestStorageEngine.java
@@ -39,19 +39,16 @@ public class TestStorageEngine implements StorageEngine {
         return ENGINE_NAME;
     }
 
-    /** {@inheritDoc} */
     @Override
     public void start() throws StorageException {
         // No-op.
     }
 
-    /** {@inheritDoc} */
     @Override
     public void stop() throws StorageException {
         // No-op.
     }
 
-    /** {@inheritDoc} */
     @Override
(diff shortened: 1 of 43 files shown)


AMashenkov (approved):

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/impl/TestStorageEngine.java
```diff
@@ -39,19 +39,16 @@ public String name() {
...
 
-    /** {@inheritDoc} */
     @Override
     public void stop() throws StorageException {
         // No-op.
     }
 
-    /** {@inheritDoc} */
     @Override
     public MvTableStorage createMvTable(TableConfiguration tableCfg, TablesConfiguration tablesCfg,
```
rpuch, line 53: Do we still need `tableCfg` parameter here? It looks like we are only interested in table ID from it.

Also, do we need `tablesCfg`?

And what about distribution zone cfg: would it be better to pass some kind of config-agnostic object (zone descriptor?) containing partitions count (and something else, probably)?

tkalkirill: For now, I suggest leaving it as is in IGNITE-19483 will be changed to pass catalog descriptors instead of configuration.

Instead of **tablesCfg** there will be either some temporary interface or a configuration will remain, I think it's more of an interface.

rpuch: Ok

rpuch (approved):

PR #2176, opened 2023-06-09
IGNITE-18856

sanpwc:


modules/replicator/src/main/java/org/apache/ignite/internal/replicator/exception/PrimaryReplicaMissException.java
```diff
@@ -22,6 +22,7 @@
 import org.apache.ignite.lang.IgniteInternalException;
 import org.apache.ignite.lang.IgniteStringFormatter;
 
+// TODO: sanpwc rework
```
vldpyatkov, line 25: TODO without a ticket number.

PR #2181, opened 2023-06-12
IGNITE-19709 Sql. Get rid of starting implicit transactions for SELECT statements without FROM keyword, change mapping

zstan:
…T statements without FROM keyword, change mapping

### DIFF
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/IgniteFragmentMapping.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/IgniteFragmentMapping.java
@@ -72,7 +72,7 @@ public class IgniteFragmentMapping {
      * @param mq  Metadata query instance.
      * @return Fragment meta information.
      */
-    public static FragmentMapping fragmentMappingForMetadataQuery(RelNode rel, RelMetadataQuery mq, MappingQueryContext ctx) {
+    public static FragmentMapping calculateMapping(RelNode rel, RelMetadataQuery mq, MappingQueryContext ctx) {
         assert mq instanceof RelMetadataQueryEx;
 
         IgniteRelVisitor<FragmentMapping> visitor = new IgniteRelVisitor<>() {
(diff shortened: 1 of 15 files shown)


zstan:
@korlov42 can u make a review ? There are exist some questionable places as for me.
It`s not clear question - can Boolean hold ternary logic or only binary, i think it can )

zstan:
@korlov42 i revert tx fix and change only mapping part, can u take a look ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/IgniteFragmentMapping.java
```diff
@@ -85,33 +231,19 @@ public MetadataDef<FragmentMappingMetadata> getDef() {
...
-     * See {@link IgniteMdFragmentMapping#fragmentMapping(RelNode, RelMetadataQuery, MappingQueryContext)}.
+     * See {@link IgniteFragmentMapping#fragmentMapping(SingleRel, RelMetadataQuery, MappingQueryContext)}.
      *
      * <p>{@link ColocationMappingException} may be thrown on two children nodes locations merge. This means that the fragment
      * (which part the parent node is) cannot be executed on any node and additional exchange is needed. This case we throw {@link
      * NodeMappingException} with an edge, where we need the additional exchange. After the exchange is put into the fragment and the
      * fragment is split into two ones, fragment meta information will be recalculated for all fragments.
      */
-    public FragmentMapping fragmentMapping(BiRel rel, RelMetadataQuery mq, MappingQueryContext ctx) {
+    private static FragmentMapping fragmentMapping(BiRel rel, RelMetadataQuery mq, MappingQueryContext ctx) {
```
korlov42, line 246: why do we need all those static fields? It's better to inline them

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/IgniteFragmentMapping.java
```diff
@@ -69,13 +75,153 @@ public class IgniteMdFragmentMapping implements MetadataHandler<FragmentMappingM
     public static FragmentMapping fragmentMappingForMetadataQuery(RelNode rel, RelMetadataQuery mq, MappingQueryContext ctx) {
```
korlov42, line 75: it worth to rename this method

korlov42 (approved):

PR #2183, opened 2023-06-13
IGNITE-19606 Linearize metaStorageManager.deployWatches and metaStora…

vldpyatkov:
…geManager.start()
https://issues.apache.org/jira/browse/IGNITE-19606

### DIFF
--- a/modules/distribution-zones/src/testFixtures/java/org/apache/ignite/internal/distributionzones/DistributionZonesTestUtil.java
+++ b/modules/distribution-zones/src/testFixtures/java/org/apache/ignite/internal/distributionzones/DistributionZonesTestUtil.java
@@ -57,7 +57,6 @@ import org.apache.ignite.internal.schema.configuration.storage.DataStorageChange
 import org.apache.ignite.internal.util.ByteUtils;
 import org.apache.ignite.internal.vault.VaultManager;
 import org.apache.ignite.lang.ByteArray;
-import org.apache.ignite.lang.NodeStoppingException;
 import org.jetbrains.annotations.Nullable;
 
 
@@ -306,11 +305,9 @@ public class DistributionZonesTestUtil {
      * TODO: IGNITE-19403 Watch listeners must be deployed after the zone manager starts.
      *
      * @param metaStorageManager Meta storage manager.
-     * @throws NodeStoppingException If node is stopping.
      * @throws InterruptedException If thread was interrupted.
      */
-    public static void deployWatchesAndUpdateMetaStorageRevision(MetaStorageManager metaStorageManager)
-            throws NodeStoppingException, InterruptedException {
+    public static void deployWatchesAndUpdateMetaStorageRevision(MetaStorageManager metaStorageManager) throws InterruptedException {
         // Watches are deployed before distributionZoneManager start in order to update Meta Storage revision before
         // distributionZoneManager's recovery.
         CompletableFuture<Void> deployWatchesFut = metaStorageManager.deployWatches();
--- a/modules/distribution-zones/src/testFixtures/java/org/apache/ignite/internal/distributionzones/DistributionZonesTestUtil.java
+++ b/modules/distribution-zones/src/testFixtures/java/org/apache/ignite/internal/distributionzones/DistributionZonesTestUtil.java
@@ -313,7 +313,7 @@ public class DistributionZonesTestUtil {
             throws NodeStoppingException, InterruptedException {
         // Watches are deployed before distributionZoneManager start in order to update Meta Storage revision before
         // distributionZoneManager's recovery.
-        var deployWatchesFut = metaStorageManager.deployWatches();
+        CompletableFuture<Void> deployWatchesFut = metaStorageManager.deployWatches();
 
         // Bump Meta Storage applied revision by modifying a fake key. DistributionZoneManager breaks on start if Vault is not empty, but
         // Meta Storage revision is equal to 0.
--- a/modules/distribution-zones/src/testFixtures/java/org/apache/ignite/internal/distributionzones/DistributionZonesTestUtil.java
+++ b/modules/distribution-zones/src/testFixtures/java/org/apache/ignite/internal/distributionzones/DistributionZonesTestUtil.java
@@ -313,17 +313,17 @@ public class DistributionZonesTestUtil {
             throws NodeStoppingException, InterruptedException {
         // Watches are deployed before distributionZoneManager start in order to update Meta Storage revision before
         // distributionZoneManager's recovery.
-        metaStorageManager.deployWatches();
+        var deployWatchesFut = metaStorageManager.deployWatches();
 
         // Bump Meta Storage applied revision by modifying a fake key. DistributionZoneManager breaks on start if Vault is not empty, but
         // Meta Storage revision is equal to 0.
         var fakeKey = new ByteArray("foobar");
 
-        CompletableFuture<Boolean> invokeFuture = metaStorageManager.invoke(
+        CompletableFuture<Boolean> invokeFuture = deployWatchesFut.thenCompose(unused -> metaStorageManager.invoke(
                 Conditions.notExists(fakeKey),
                 Operations.put(fakeKey, fakeKey.bytes()),
                 Operations.noop()
-        );
+        ));
 
         assertThat(invokeFuture, willBe(true));
 
--- a/modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
+++ b/modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
@@ -173,7 +173,7 @@ public interface MetaStorageManager extends IgniteComponent {
      *
      * @return Future which completes when Meta storage manager is started and deploying watches is finished.
      */
-    CompletableFuture<Void> deployWatches() throws NodeStoppingException;
+    CompletableFuture<Void> deployWatches();
 
     /**
      * Returns cluster time with a hybrid clock instance and access to safe time.
--- a/modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
+++ b/modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
@@ -171,7 +171,7 @@ public interface MetaStorageManager extends IgniteComponent {
      *
      * <p>Should be called after all Ignite components have registered required watches and they are ready to process Meta Storage events.
      *
-     * @return Future to complete.
+     * @return Future which completes when Meta storage manager is started and deploying watches is finished.
      */
     CompletableFuture<Void> deployWatches() throws NodeStoppingException;
 
--- a/modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
+++ b/modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
@@ -170,8 +170,10 @@ public interface MetaStorageManager extends IgniteComponent {
      * Starts all registered watches.
      *
      * <p>Should be called after all Ignite components have registered required watches and they are ready to process Meta Storage events.
+     *
+     * @return Future to complete.
      */
-    void deployWatches() throws NodeStoppingException;
+    CompletableFuture<Void> deployWatches() throws NodeStoppingException;
 
     /**
      * Returns cluster time with a hybrid clock instance and access to safe time.
--- a/modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItMetaStorageManagerImplTest.java
+++ b/modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItMetaStorageManagerImplTest.java
@@ -61,7 +61,6 @@ import org.apache.ignite.internal.vault.VaultEntry;
 import org.apache.ignite.internal.vault.VaultManager;
 import org.apache.ignite.internal.vault.inmemory.InMemoryVaultService;
 import org.apache.ignite.lang.ByteArray;
-import org.apache.ignite.lang.NodeStoppingException;
 import org.apache.ignite.network.ClusterService;
 import org.apache.ignite.network.NetworkAddress;
 import org.apache.ignite.network.StaticNodeFinder;
@@ -87,7 +86,7 @@ public class ItMetaStorageManagerImplTest extends IgniteAbstractTest {
     private MetaStorageManagerImpl metaStorageManager;
 
     @BeforeEach
-    void setUp(TestInfo testInfo, @InjectConfiguration RaftConfiguration raftConfiguration) throws NodeStoppingException {
+    void setUp(TestInfo testInfo, @InjectConfiguration RaftConfiguration raftConfiguration) {
         var addr = new NetworkAddress("localhost", 10_000);
 
         clusterService = clusterService(testInfo, addr.port(), new StaticNodeFinder(List.of(addr)));
--- a/modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItMetaStorageManagerImplTest.java
+++ b/modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItMetaStorageManagerImplTest.java
@@ -118,7 +118,7 @@ public class ItMetaStorageManagerImplTest extends IgniteAbstractTest {
         raftManager.start();
         metaStorageManager.start();
 
-        metaStorageManager.deployWatches();
+        metaStorageManager.deployWatches().join();
     }
 
     @AfterEach
--- a/modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItMetaStorageManagerImplTest.java
+++ b/modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItMetaStorageManagerImplTest.java
@@ -24,6 +24,7 @@ import static org.apache.ignite.internal.testframework.flow.TestFlowUtils.subscr
 import static org.apache.ignite.internal.testframework.matchers.CompletableFutureExceptionMatcher.willThrowFast;
 import static org.apache.ignite.internal.testframework.matchers.CompletableFutureMatcher.will;
 import static org.apache.ignite.internal.testframework.matchers.CompletableFutureMatcher.willBe;
+import static org.apache.ignite.internal.testframework.matchers.CompletableFutureMatcher.willCompleteSuccessfully;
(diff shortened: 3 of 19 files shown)


modules/distribution-zones/src/testFixtures/java/org/apache/ignite/internal/distributionzones/DistributionZonesTestUtil.java
```diff
@@ -313,17 +313,17 @@ public static void deployWatchesAndUpdateMetaStorageRevision(MetaStorageManager
             throws NodeStoppingException, InterruptedException {
         // Watches are deployed before distributionZoneManager start in order to update Meta Storage revision before
         // distributionZoneManager's recovery.
-        metaStorageManager.deployWatches();
+        var deployWatchesFut = metaStorageManager.deployWatches();
```
sanpwc, line 316: AFAIK it's not valid to use var here.

modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
```diff
@@ -170,8 +170,10 @@ public interface MetaStorageManager extends IgniteComponent {
      * Starts all registered watches.
      *
      * <p>Should be called after all Ignite components have registered required watches and they are ready to process Meta Storage events.
+     *
+     * @return Future to complete.
```
sanpwc, line 174: Please add meaningful comment here.

modules/distribution-zones/src/testFixtures/java/org/apache/ignite/internal/distributionzones/DistributionZonesTestUtil.java
```diff
@@ -313,17 +313,17 @@ public static void deployWatchesAndUpdateMetaStorageRevision(MetaStorageManager
...
 
         // Bump Meta Storage applied revision by modifying a fake key. DistributionZoneManager breaks on start if Vault is not empty, but
         // Meta Storage revision is equal to 0.
         var fakeKey = new ByteArray("foobar");
 
-        CompletableFuture<Boolean> invokeFuture = metaStorageManager.invoke(
+        CompletableFuture<Boolean> invokeFuture = deployWatchesFut.thenCompose(unused -> metaStorageManager.invoke(
```
sanpwc, line 322: Is it the only place we should await deployWatches?

modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/impl/MetaStorageDeployWatchesCorrectnessTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.ignite.internal.raft.RaftManager;
+import org.apache.ignite.internal.testframework.IgniteAbstractTest;
+import org.apache.ignite.internal.vault.VaultManager;
+import org.apache.ignite.internal.vault.inmemory.InMemoryVaultService;
+import org.apache.ignite.lang.NodeStoppingException;
+import org.apache.ignite.network.ClusterService;
+import org.junit.jupiter.api.Test;
+
+/**
+ * There are tests of correctness invocation {@link MetaStorageManager#deployWatches()}.
```
sanpwc, line 43: It's just "Test that checks correctness of an invocation ..." without "There are".

modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/impl/MetaStorageDeployWatchesCorrectnessTest.java
```diff
@@ -0,0 +1,95 @@
...
+        ClusterService clusterService = mock(ClusterService.class);
+        LogicalTopologyService logicalTopologyService = mock(LogicalTopologyService.class);
+        RaftManager raftManager = mock(RaftManager.class);
+
+        when(cmgManager.metaStorageNodes()).thenReturn(completedFuture(Set.of(mcNodeName)));
+        when(clusterService.nodeName()).thenReturn(mcNodeName);
+        when(raftManager.startRaftGroupNodeAndWaitNodeReadyFuture(any(), any(), any(), any(), any())).thenReturn(completedFuture(null));
+
+        var metastore = new MetaStorageManagerImpl(
+                new VaultManager(new InMemoryVaultService()),
```
sanpwc, line 61: VaultManager is an Ignite service, so please either start/stop it or use mock instead.

modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/impl/MetaStorageDeployWatchesCorrectnessTest.java
```diff
@@ -0,0 +1,95 @@
...
+                clusterService,
+                cmgManager,
+                logicalTopologyService,
+                raftManager,
+                new SimpleInMemoryKeyValueStorage(mcNodeName),
+                clock
+        );
+
+        checkCorrectness(metastore);
```
sanpwc, line 70: I'd rather use parametrized test here with method source (MethodSource) that will either return StandaloneMetaStorageManager or General meta storage. In that case checkCorrectness should be refactored to parametrized test itself.

modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/impl/MetaStorageDeployWatchesCorrectnessTest.java
```diff
@@ -0,0 +1,95 @@
...
+     *
+     * @param metastore Meta storage.
+     * @throws NodeStoppingException If failed.
+     */
+    private static void checkCorrectness(MetaStorageManager metastore) throws NodeStoppingException {
+        var deployWatchesFut = metastore.deployWatches();
+
+        assertFalse(deployWatchesFut.isDone());
+
+        metastore.start();
```
sanpwc, line 91: metastore.start() is  async thus `deployWatchesFut.isDone()` isn't guaranteed without some awaiting logic. Please use 
`org.apache.ignite.internal.testframework.matchers.CompletableFutureMatcher#willSucceedFast` or similar instead.

modules/distribution-zones/src/testFixtures/java/org/apache/ignite/internal/distributionzones/DistributionZonesTestUtil.java
vldpyatkov: This is a test classe. I think it is possible for test purposes.

modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/impl/MetaStorageDeployWatchesCorrectnessTest.java
vldpyatkov: It is just a unit test, I know the behavior here.
I am sure we will invoke the listener in the same thread where the start future is completed.

modules/distribution-zones/src/testFixtures/java/org/apache/ignite/internal/distributionzones/DistributionZonesTestUtil.java
vldpyatkov: I do not know. I add explicit waiting here because it is a common test class, which might be used in different circumstances.
Other places probably have internal guaranties or should wait to this future.

modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/impl/MetaStorageDeployWatchesCorrectnessTest.java
sanpwc: Why should we exposes such implementation details (e.g. the one that there's no executors or similar inside metaStorage.start) instead of writing guaranteed code based on willSucceedFast?

modules/distribution-zones/src/testFixtures/java/org/apache/ignite/internal/distributionzones/DistributionZonesTestUtil.java
vldpyatkov: I think, if before, we have no guarantees that MC start earlier than watches subscribe, it means thus tests are buggy.
Ok, add synchronous waiting in every invocation.

sanpwc: > Ok, add synchronous waiting in every invocation.
Well, in tests I'd rather add assertThat(willSucceedFast()) with proper message in order to have a meaningful explanation if synchronous waiting will hang instead of simple join.

modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
sanpwc: You're joking right? The core idea of the current ticket (IGNITE-19606) is to linearise ms.start() and ms.deployWatches().
Do we still have an implementation with unconnected start and deployWatches?

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItMetaStorageMultipleNodesAbstractTest.java
```diff
@@ -228,6 +233,8 @@ void testLearnerJoin(TestInfo testInfo) throws NodeStoppingException {
 
         firstNode.cmgManager.initCluster(List.of(firstNode.name()), List.of(firstNode.name()), "test");
 
+        firstNode.deployWatches();
```
sanpwc, line 236: It seems incorrect. deployWatches is a part of node start procedure.

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItMetaStorageManagerImplTest.java
```diff
@@ -118,7 +118,7 @@ void setUp(TestInfo testInfo, @InjectConfiguration RaftConfiguration raftConfigu
         raftManager.start();
         metaStorageManager.start();
 
-        metaStorageManager.deployWatches();
+        assertThat("Watches were not deployed", metaStorageManager.deployWatches(), willCompleteSuccessfully());
```
sanpwc, line 121: Please add full stop at the end of the sentence. Related to all such messages.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageManagerImpl.java
```diff
@@ -277,14 +278,16 @@ public void unregisterWatch(WatchListener lsnr) {
     }
 
     @Override
-    public void deployWatches() throws NodeStoppingException {
+    public CompletableFuture<Void> deployWatches() {
         if (!busyLock.enterBusy()) {
-            throw new NodeStoppingException();
+            return CompletableFuture.failedFuture(new NodeStoppingException());
```
sanpwc, line 283: Good catch!

modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/impl/MetaStorageDeployWatchesCorrectnessTest.java
```diff
@@ -0,0 +1,115 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+/**
+ * Tests that check correctness of an invocation {@link MetaStorageManager#deployWatches()}.
+ */
+public class MetaStorageDeployWatchesCorrectnessTest extends IgniteAbstractTest {
+
```
sanpwc, line 50: Unnecessary blank line.

modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/impl/MetaStorageDeployWatchesCorrectnessTest.java
```diff
@@ -0,0 +1,115 @@
...
+    @BeforeAll
+    public static void init() {
+        vaultManager = new VaultManager(new InMemoryVaultService());
+
+        vaultManager.start();
+    }
+
+    @AfterAll
+    public static void deInit() {
+        vaultManager.stop();
```
sanpwc, line 63: Despite the fact that Vault.beforeNodeStop() is no-op, it's still required to call it before calling vaultManager.stop();

modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/impl/MetaStorageDeployWatchesCorrectnessTest.java
```diff
@@ -0,0 +1,115 @@
...
+     * @param metastore Meta storage.
+     */
+    @ParameterizedTest
+    @MethodSource("metaStorageProvider")
+    public void testCheckCorrectness(MetaStorageManager metastore) {
+        var deployWatchesFut = metastore.deployWatches();
+
+        assertFalse(deployWatchesFut.isDone());
+
+        metastore.start();
```
sanpwc, line 111: Same as for vault, MetaStorageManager is an IgniteCompoment thus it's required not only to start it but also call beforeNodeStop and stop. All in all, I'd suggest adding vault and ms to some components list and close it (beforeNodeStop + stop). From my point of view, it also may worth to change @Before**All**/@After**All** to Before**Each**/After**Each**.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/ItDistributedConfigurationPropertiesTest.java
```diff
@@ -286,6 +287,8 @@ void setUp(@WorkDirectory Path workDir, TestInfo testInfo) throws Exception {
         Stream.of(firstNode, secondNode).parallel().forEach(Node::start);
 
         firstNode.cmgManager.initCluster(List.of(firstNode.name()), List.of(), "cluster");
+
+        Stream.of(firstNode, secondNode).parallel().forEach(Node::deployWatches);
```
sanpwc, line 291: Same as above. It's required to call deployWatches within node's startup process and assert deployWatches' future completion after init.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItDistributedConfigurationStorageTest.java
```diff
@@ -216,6 +223,8 @@ void testRestartWithPds(@WorkDirectory Path workDir, TestInfo testInfo) throws E
         try {
             node2.start();
 
+            node2.deployWatches();
```
sanpwc, line 226: Same as above.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
```diff
@@ -514,6 +516,8 @@ void testDestroyPartitionStoragesOnRestartEvictedNode(TestInfo testInfo) throws
 
         newNode.start();
 
+        newNode.deployWatches();
```
sanpwc, line 519: Same as above.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
```diff
@@ -494,24 +492,15 @@ private PartialNode startPartialNode(
...
-                        metaStorageMgr.deployWatches();
-                    } catch (NodeStoppingException e) {
-                        throw new CompletionException(e);
-                    }
-
-                    return configurationCatchUpFuture;
-                });
-
-        assertThat(startFuture, willCompleteSuccessfully());
+        assertThat("Partial node was not started", startFuture, willCompleteSuccessfully());
```
sanpwc, line 503: Why you've changed the order of the processes?

vldpyatkov: I just add a message here, because the assertion relates to deploy watches future.

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItMetaStorageMultipleNodesAbstractTest.java
vldpyatkov: Fixed

modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/impl/MetaStorageDeployWatchesCorrectnessTest.java
vldpyatkov: I added the methods' invocation for MC. But I do not think that have to be implemented in suggested manner.

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItMetaStorageManagerImplTest.java
sanpwc: AFAIK that was revised from Ignite 2 times, and now we follow common grammar rules.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
sanpwc: Really? Seems that originally there was notificationFuture(notifyCurrentConfigurationListeners, notifyCurrentConfigurationListeners).thenCompose(deployWatches) and now it's startFuture(notifyCurrentConfigurationListeners, notifyCurrentConfigurationListeners, deployWatches) in other words originally deployWatches was called after notificationFutures and now it's called concurrently. Am I mistaken?

modules/metastorage/src/test/java/org/apache/ignite/internal/metastorage/impl/MetaStorageDeployWatchesCorrectnessTest.java
vldpyatkov: No, I add this in the test body.

sanpwc: Not sure that I got you. I will add, or you've already added? I'm asking because currently I can see only vault's stop.

PR #2186, opened 2023-06-13
IGNITE-19536 Added RecoverableException marker interface indicates recoverable error

sk0x50:
https://issues.apache.org/jira/browse/IGNITE-19536

### DIFF
--- a/modules/api/src/main/java/org/apache/ignite/lang/RecoverableException.java
+++ b/modules/api/src/main/java/org/apache/ignite/lang/RecoverableException.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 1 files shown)


modules/api/src/main/java/org/apache/ignite/lang/RecoverableException.java
```diff
@@ -0,0 +1,38 @@
...
+ *         ...
+ *         tx.commit();
+ *     } catch (TransactionException t) {
+ *         if (t instanceof RecoverableException) {
+ *             // put your retry logic here.
+ *         }
+ *     }
+ * </code>
+ */
+public interface RecoverableException {
```
sanpwc, line 37: We already have recoverable exceptions, e.g. PrimaryReplicaMissException. Do we have a ticket (team-based tickets) to check existing ones and mark them as RecoverableException?

sanpwc, line 37: To be honest I'm not sure whether proposed approach with RecoverableException interface is good enough. In many cases recoverability is context-depended.

- It's possible to retry tx request after lock exception if nothing was shared with a user, so formally LockExpcetion is RecoverableException, however in real life it is "sometimes recoverable".
- Same is about PrimaryReplicaMissException. If there wasn't any successful touches of given partition it's possible to retry the invoke, otherwise it's not.
 All in all, RecoverableException might bring even more mess.

sk0x50: additional details were added to the ticket

PR #2196, opened 2023-06-15
IGNITE-19615: Index is not used while performing SELECT over an indexed column.

lowka:
Removed int type / interval type checks from TypeCoercion::needToCast(Type, Type) - it behaves exactly the same way as TypeUtils::needToCast (which was removed in https://issues.apache.org/jira/browse/IGNITE-19128).


AMashenkov (approved):

korlov42 (commented):
After this patch an index is used indeed. But resulting plan is rather invalid. For a query like `SELECT val FROM my WHERE id = 100500`, where `id` is of type TINYINT, I've got following plan:

```
IgniteExchange<...>
  IgniteIndexScan(<...>, searchBounds=[[ExactBounds [bound=CAST(100500):TINYINT NOT NULL]]], filters=[=(CAST($t0):INTEGER NOT NULL, 100500)]<...>
```

this results in RuntimeException, since 100500 is out of range of TINYINT

lowka:
> After this patch an index is used indeed. But resulting plan is rather invalid. For a query like `SELECT val FROM my WHERE id = 100500`, where `id` is of type TINYINT, I've got following plan:
> 
> ```
> IgniteExchange<...>
>   IgniteIndexScan(<...>, searchBounds=[[ExactBounds [bound=CAST(100500):TINYINT NOT NULL]]], filters=[=(CAST($t0):INTEGER NOT NULL, 100500)]<...>
> ```
> 
> this results in RuntimeException, since 100500 is out of range of TINYINT
@korlov42 
I going to look into it. I looked at the result of `IgniteSqlValidator` and it add `INTEGER` casts to both id and literal (this is expected). But somehow the bounds are still use `TINYINT`.

lowka:
Optimiser starts from:

> LogicalProject(VAL=[$1]), id = 116
>   LogicalFilter(condition=[=(CAST($0):INTEGER NOT NULL, 100500)]), id = 114
>     IgniteLogicalTableScan(table=[[PUBLIC, MY]]), id = 113
    
And produces:

> IgniteExchange(distribution=[single]), id = 144
>  IgniteTableScan(table=[[PUBLIC, MY]], filters=[=(CAST($t0):INTEGER NOT NULL, 100500)], projects=[[$t1]], requiredColumns=[{0, 1}]), id = 143

lowka:
Yep. Types are correct.

> SqlValidator BINARY_COMPARISON: CAST(`MY`.`ID` AS INTEGER) = 100500 types: INTEGER INTEGER

So the problem should be in SearchBound assembly code.

korlov42 (changes requested):
First problem with this patch is that now IgniteTypeCoercion has two similar method with different behaviour:
```
boolean needToCast(RelDataType fromType, RelDataType toType)
boolean needToCast(SqlValidatorScope scope, SqlNode node, RelDataType toType)
```
The former returns `false` for (TINYINT, INTEGER), while the latter returns `true` in the very similar case (node::TINYINT, INTEGER).

The second problem is that it deals with literals only, while leaving dynamic params and columns references out of scope for some reason

lowka:
>The second problem is that it deals with literals only, while leaving dynamic params and columns references out of scope for some reason
What makes you think so? Search bounds are built from arbitrary expression, optimizer wouldn't be able to select an index.
But nonetheless I have reverted the changes to the state as were copied from ignite2.

AMashenkov (approved):

lowka:
>I have reverted the changes to the state as were copied from ignite2.
That would not work. Left all as it is - just renamed a method to `needToCastInIndex`.

lowka:
* Partially reverted changes, that were made in https://issues.apache.org/jira/browse/IGNITE-19128 and moved type cast checks to TypeUtils.
* Added test cases with indexes to ImplicitCastsTest.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/TypeUtils.java
```diff
@@ -533,6 +535,55 @@ public static NativeType columnType2NativeType(ColumnType columnType, int precis
...
+            return false;
+        }
+        // Should keep sync with rules in SqlTypeCoercionRule.
+        assert SqlTypeUtil.canCastFrom(toType, fromType, true);
+        return true;
+    }
+
+    /**
+     * Checks whether one type can be casted to another if one of type is a custom data type.
+     * This method expects at least one of its arguments to be a custom data type.
```
korlov42, line 570:
```suggestion
     * Checks whether one type can be casted to another if one of type is a custom data type.
     *
     * <p>This method expects at least one of its arguments to be a custom data type.
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/TypeUtils.java
```diff
@@ -533,6 +535,55 @@ public static NativeType columnType2NativeType(ColumnType columnType, int precis
...
+     */
+    public static boolean customDataTypeNeedCast(IgniteTypeFactory factory, RelDataType fromType, RelDataType toType) {
+        assert fromType.getSqlTypeName() == SqlTypeName.ANY || toType.getSqlTypeName() == SqlTypeName.ANY :
```
korlov42, line 573: why do we need both this assertion and one on line 584? Let's leave only one on line 584 with improved message

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/TypeUtils.java
```diff
@@ -533,6 +535,55 @@ public static NativeType columnType2NativeType(ColumnType columnType, int precis
         }
     }
 
+    /** Checks whether cast operation is necessary in {@code SearchBound}. */
+    public static boolean needCastInSearchBounds(IgniteTypeFactory typeFactory, RelDataType fromType, RelDataType toType) {
+        // Check custom data types first.
+        if (toType.getSqlTypeName() == SqlTypeName.ANY || fromType.getSqlTypeName() == SqlTypeName.ANY) {
```
korlov42, line 541: this predicate doesn't seem right... Method `customDataTypeNeedCast` actually expects that at least one of the argument is instance of IgniteCustomType. If there is an invariant that there should not be any ANY type but IgniteCustomType, it's better to check it explicitly

lowka: @korlov42 yeap. I removed the first assertion since the second one is the only one that is needed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/TypeUtils.java
```diff
@@ -533,6 +535,55 @@ public static NativeType columnType2NativeType(ColumnType columnType, int precis
...
+            return customDataTypeNeedCast(typeFactory, fromType, toType);
```
korlov42, line 542: why do even need call in needCastInSearchBounds? This is the very same rule as in TypeCoercion

lowka: Fixed.

PR #2197, opened 2023-06-15
IGNITE-19737 TestMvStorageUpdateHandlerTest.testConcurrentExecuteBatchGc is flacky

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19737


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/gc/GcUpdateHandler.java
```diff
@@ -69,41 +68,74 @@ public PendingComparableValuesTracker<HybridTimestamp, Void> getSafeTimeTracker(
...
             if (gcEntry == null) {
-                return false;
+                return VacuumResult.NO_GARBAGE_LEFT;
             }
 
             RowId rowId = gcEntry.getRowId();
 
-            if (!locker.tryLock(rowId)) {
-                return true;
+            if (!useTryLock) {
```
rpuch, line 133: How about inverting the condition (and swapping the branches)? It's a bit easier to understand a condition without a negation

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/gc/GcUpdateHandler.java
```diff
@@ -69,41 +68,74 @@ public PendingComparableValuesTracker<HybridTimestamp, Void> getSafeTimeTracker(
...
+    }
+
+    private VacuumResult internalVacuumBatch(HybridTimestamp lowWatermark, IntHolder countHolder) {
         return storage.runConsistently(locker -> {
+            int count = countHolder.get();
+
             for (int i = 0; i < count; i++) {
-                if (!internalVacuum(lowWatermark, locker)) {
-                    return false;
+                // It is safe for the first iteration to use a lock instead of tryLock.
```
rpuch, line 109: Why is it safe specifically for the first iteration?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/gc/GcUpdateHandler.java
```diff
@@ -69,41 +68,74 @@ public PendingComparableValuesTracker<HybridTimestamp, Void> getSafeTimeTracker(
...
+     *      exit is allowed when it is not possible to acquire a lock for the {@link RowId}.
      * @return {@code False} if there is no garbage left in the storage.
      */
-    public boolean vacuumBatch(HybridTimestamp lowWatermark, int count) {
+    public boolean vacuumBatch(HybridTimestamp lowWatermark, int count, boolean strict) {
+        if (count <= 0) {
+            return true;
+        }
+
+        IntHolder countHolder = new IntHolder(count);
```
rpuch, line 80: It looks like `IntHolder` is always used for counting down. How about renaming it to `CountDown`? It could also have a method like `finished()` that would check that the count reached zero

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/gc/GcUpdateHandler.java
tkalkirill: Fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/gc/GcUpdateHandler.java
tkalkirill: Tried to expand the comment.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/gc/GcUpdateHandler.java
tkalkirill: I'd rather leave it as is because it's closer (as api usage) to using **AtomicInteger**.

rpuch: Why is being closer to `AtomicInteger` is an advantage? `AtomicInteger` is a general-purpose util, and `IntHolder` is in our specific package, and only used in just one place, so it could be specialized

rpuch (approved):

PR #2201, opened 2023-06-15
IGNITE-19736 Do not cancel tasks in DistributionZoneManager#executor if they were created by immediate scaleUp/scaleDown events. Avoid concurrent executing several tasks for the same zone.

sergeyuttsel:
https://issues.apache.org/jira/browse/IGNITE-19736

### DIFF
--- a/modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
+++ b/modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
@@ -21,6 +21,7 @@ import static java.util.Collections.emptySet;
 import static java.util.concurrent.CompletableFuture.completedFuture;
 import static java.util.concurrent.CompletableFuture.failedFuture;
 import static java.util.concurrent.CompletableFuture.supplyAsync;
+import static java.util.concurrent.TimeUnit.SECONDS;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 import static org.apache.ignite.internal.distributionzones.DistributionZonesUtil.deleteDataNodesAndUpdateTriggerKeys;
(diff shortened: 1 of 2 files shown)


modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerScaleUpTest.java
```diff
@@ -307,312 +301,6 @@ void testDropZoneDoNotPropagateDataNodesAfterScaleDown() throws Exception {
         assertZoneScaleDownChangeTriggerKey(null, ZONE_1_ID, keyValueStorage);
     }
 
-    @Test
-    void testTwoScaleUpTimersSecondTimerRunFirst() throws Exception {
```
alievmirza, line 311: Is this a temporary solution? If yes, why did we delete this tests, not mute?

sergeyuttsel: This is a permanent solution. My algorithm needs a guarantee that tasks of each zone will be executed in the order they were added to the executor. These tests check that multiple tasks for a zone are running at the same time, so now such tests will not pass successfully. Even if I make a striped executor, it won't change this.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -282,8 +282,10 @@ public DistributionZoneManager(
 
         nodesAttributes = new ConcurrentHashMap<>();
 
+        // Use a single thread executor to avoid concurrent executing several tasks for the same zone.
+        // TODO: IGNITE-19783 Need to use a striped executor.
```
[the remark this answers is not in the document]
sergeyuttsel: I agree. Moved.

alievmirza (approved):

PR #2202, opened 2023-06-16
IGNITE-19621 Sql. Shrink planner search space by removing pre known not applicable indexes

zstan:


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rule/logical/ExposeIndexRule.java
```diff
@@ -79,6 +83,14 @@ public void onMatch(RelOptRuleCall call) {
         call.transformTo(indexes.get(0), equivMap);
     }
 
+    /** Filter pre known not applicable variants. Significant shrink search space in some cases. */
+    private static boolean filter(IgniteTable table, String idxName, List<SearchBounds> searchBounds) {
```
korlov42, line 87: With such approach pk index won't be chosen for CNLJ

korlov42: seems I was wrong about pk index would not being chosen for CNLJ... Please ignore message above...

korlov42 (approved):
@zstan LGTM in general. It would be nice to see results of the benchmark though

PR #2203, opened 2023-06-16
IGNITE-19744 Clean up IEP-54 leftovers

korlov42:
https://issues.apache.org/jira/browse/IGNITE-19744


modules/schema/src/main/java/org/apache/ignite/internal/schema/SchemaDescriptor.java
```diff
@@ -84,71 +82,118 @@ public class SchemaDescriptor {
...
+        List<Column> orderedColumns = new ArrayList<>(columns.size());
+
+        Object2IntMap<String> columnNameToPositionInKey = new Object2IntOpenHashMap<>();
+        int idx = 0;
+        for (String name : keyColumns) {
+            assert !columnNameToPositionInKey.containsKey(name)
+                    : "Key column must not have duplicates: " + name;
+
+            columnNameToPositionInKey.put(name, idx++);
+        }
```
AMashenkov, line 114: This can be an utility method.

modules/schema/src/main/java/org/apache/ignite/internal/schema/SchemaDescriptor.java
```diff
@@ -84,71 +82,118 @@ public class SchemaDescriptor {
...
-
-            for (int i = 0; i < colocationCols.length; i++) {
-                Column col = colMap.get(colocationCols[i]);
-                this.colocationCols[i] = col;
-                this.colocationColIndexes.put(col, i);
+        if (colocationColumns == null) {
+            colocationColumns = keyColumns;
+        }
+
+        Object2IntMap<String> columnNameToPositionInColocation = new Object2IntOpenHashMap<>();
```
AMashenkov, line 120:
```suggestion
        if (colocationColumns == null) {
            columnNameToPositionInColocation = columnNameToPositionInKey;
        }
        else {
        Object2IntMap<String> columnNameToPositionInColocation = new Object2IntOpenHashMap<>();
```

modules/schema/src/main/java/org/apache/ignite/internal/schema/SchemaDescriptor.java
```diff
@@ -84,71 +82,118 @@ public class SchemaDescriptor {
...
+            if (column.type() instanceof TemporalNativeType) {
+                hasTemporalColumns = true;
             }
         }
 
-        // TODO: Move keyIndex and colocationIndex to Column class for faster and simpler access?
-        keyColIndexes = new HashMap<>(keyCols.length);
+        this.ver = ver;
+        this.columns = List.copyOf(orderedColumns);
+        this.columnsByName = Map.copyOf(columnsByName);
```
AMashenkov, line 155: Can a copy be replaced with immutable collection?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/TableRowConverterImpl.java
```diff
@@ -46,66 +41,36 @@ public class TableRowConverterImpl implements TableRowConverter {
...
     ) {
         this.schemaRegistry = schemaRegistry;
         this.schemaDescriptor = schemaDescriptor;
-        this.fullRowColocationIndexes = fullRowColocationIndexes;
-        this.keyOnlyColocationIndexes = keyOnlyColocationIndexes;
-        this.colocationColumnTypes = colocationColumnTypes;
         this.fullTupleSchema = fullTupleSchema;
 
-        this.skipReshuffling = requiredColumns == null && schemaDescriptor.physicalOrderMatchesLogical();
+        this.skipTrimming = requiredColumns == null;
```
AMashenkov, line 62: Maybe it is better make the class package-private and have 2 constructor for both cases: `requiredColumns` is null and not null; instead of having bunch of `if (requiredColumns == null)`? Or maybe different implementations?
In fact, the object is always created via factory and it is clear which constructor must be used.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/TableRowConverterImpl.java
```diff
@@ -115,15 +80,7 @@ public class TableRowConverterImpl implements TableRowConverter {
...
-        if (!key) {
-            InternalTuple tuple = schemaDescriptor.physicalOrderMatchesLogical()
-                    ? binaryTuple
-                    : new FormatAwareProjectedTuple(binaryTuple, fullMapping);
-
-            return SqlOutputBinaryRow.newRow(schemaDescriptor.version(), fullRowColocationIndexes, colocationColumnTypes, tuple);
-        } else {
-            return SqlOutputBinaryRow.newRow(schemaDescriptor.version(), keyOnlyColocationIndexes, colocationColumnTypes, binaryTuple);
-        }
+        return SqlOutputBinaryRow.newRow(schemaDescriptor, key, binaryTuple);
```
AMashenkov, line 83: Is it possible to get rid of `boolean key` flag?
Having a special method usually better than using a flag, when the caller always knows the case.

It is ok to do in a separate ticket.

AMashenkov (approved):

modules/schema/src/main/java/org/apache/ignite/internal/schema/SchemaDescriptor.java
korlov42: according to javadoc, `Map.copyOf` returns unmodifiable Map

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/TableRowConverterImpl.java
korlov42: make sense. I filed an issue https://issues.apache.org/jira/browse/IGNITE-21731

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/TableRowConverterImpl.java
korlov42: good point! Filed an issue https://issues.apache.org/jira/browse/IGNITE-21732

modules/schema/src/main/java/org/apache/ignite/internal/schema/SchemaDescriptor.java
korlov42: the goal of this patch is to remove code rather than adding new one, so I tried not add new lines until it's absolutely required in order to proceed. There was no validation before, but after refactoring this object become more sensible to unspoken invariants like "no duplicate columns". In order to find and all the places that violates such invariant I've added bunch of assertions.

I think, the proper fix of validation should be done under separate issue (https://issues.apache.org/jira/browse/IGNITE-21733).

modules/schema/src/main/java/org/apache/ignite/internal/schema/SchemaDescriptor.java
korlov42: fixed, thanks

PR #2215, opened 2023-06-19
IGNITE-19617 Embedded basic Data Streamer

ptupitsyn:
Implement data streamer in the embedded API (e.g. to use from a Compute task).

* Extract common streamer logic into core module for reuse across client and embedded streamer.
* Add partition-based `InternalTable.upsertAll` overload: existing `upsertAll` splits the input data into per-partition batches, which is already handled by the streamer in this case.
* Batch by partition and not by node (like client does), because this is how table internals work.
* Fix error handling in `StreamerSubscriber.sendBatch`.
* Add common test class for client and embedded APIs to ensure consistent behavior.


sanpwc (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -648,18 +648,26 @@ public CompletableFuture<Void> upsertAll(Collection<BinaryRowEx> rows, InternalT
...
+    public CompletableFuture<Void> upsertAll(Collection<BinaryRowEx> rows, int partition) {
+        InternalTransaction tx = txManager.begin();
+        TablePartitionId partGroupId = new TablePartitionId(tableId, partition);
+
+        CompletableFuture<Void> fut = enlistWithRetry(
+                tx,
+                partition,
+                (commitPart, term) -> upsertAllInternal(commitPart, rows, tx, partGroupId, term),
+                ATTEMPTS_TO_ENLIST_PARTITION
+        );
```
isapego, line 666: What is going to happen, if we provide non-matching rows and partition? AFAIR our streamer gives no guarantees for the right partition, only best effort.

ptupitsyn: There is no strong guarantee on the client. But embedded mode is deterministic:
* We batch the data by partition, not by node
* Out-of-date partition assignment is not a concern
* Crashing nodes are not a concern

ptupitsyn: We use the same logic as `upsertAll` here, only the batching part is in a different place.

isapego (approved):

PR #2218, opened 2023-06-19
IGNITE-19700 Added several optimizations for meta-storage reads.

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-19700


modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
```diff
@@ -239,20 +243,25 @@ public void start() {
     }
 
     private static List<ColumnFamilyDescriptor> cfDescriptors() {
-        Options dataOptions = new Options().setCreateIfMissing(true)
+        Options sharedOptions = new Options()
+                .setCreateIfMissing(true)
+                .setNumLevels(4)
```
rpuch, line 248: I think it will be useful to comment the options that were chosen and write a justification for the specific options and values chosen; otherwise, in the future, it will become difficult to remember why the configuration is as it is now.

I'm not sure whether we need to make these configurable by the user. Maybe we should leave the values hard-coded for now as they seem to be very specific to our Meta-Storage scenarios (and the load profile will not necessarily change from cluster to cluster).

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
```diff
@@ -239,20 +243,25 @@ public void start() {
     }
 
     private static List<ColumnFamilyDescriptor> cfDescriptors() {
-        Options dataOptions = new Options().setCreateIfMissing(true)
+        Options sharedOptions = new Options()
```
rpuch, line 246: BasicOptions/BaseOptions?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
```diff
@@ -427,7 +436,7 @@ private void updateKeysIndex(WriteBatch batch, byte[] key, long curRev) {
      * @throws RocksDBException If failed.
      */
     private void fillAndWriteBatch(WriteBatch batch, long newRev, long newCntr, @Nullable HybridTimestamp ts) throws RocksDBException {
-        try (WriteOptions opts = new WriteOptions()) {
+        try (WriteOptions opts = new WriteOptions().setDisableWAL(true)) {
```
rpuch, line 439: Could you please leave a comment on why we need (and can) disable the WAL to make it explicit?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
ibessonov: "baseOptions", thank you

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
ibessonov: Done

rpuch (approved):

PR #2219, opened 2023-06-19
IGNITE-19708 Check refcounter of unit before undeploy

Flaugh24:
https://issues.apache.org/jira/browse/IGNITE-19708


PakhomovAlexander (changes requested):

modules/code-deployment/src/main/java/org/apache/ignite/internal/deployunit/DeploymentUnitAccessorImpl.java
```diff
@@ -36,7 +37,7 @@ public DeploymentUnitAccessorImpl(FileDeployerService deployer) {
      * {@inheritDoc}
      */
     @Override
-    public DisposableDeploymentUnit acquire(DeploymentUnit unit) {
+    public synchronized DisposableDeploymentUnit acquire(DeploymentUnit unit) {
```
PakhomovAlexander, line 40: You synchronize by accessor instance which is not the best choice here. The completely independent process that is needed `unit1` will be blocked by another process that runs `computeIfNotAcquired` for `unit2`.

PakhomovAlexander (changes requested):

modules/code-deployment/src/main/java/org/apache/ignite/internal/deployunit/DeploymentUnitAccessorImpl.java
Flaugh24: I'm not sure that is going to be a bottle neck

modules/code-deployment/src/main/java/org/apache/ignite/internal/deployunit/DeploymentUnitAccessorImpl.java
```diff
@@ -27,6 +30,8 @@
 public class DeploymentUnitAccessorImpl implements DeploymentUnitAccessor {
     private final RefCountedObjectPool<DeploymentUnit, DisposableDeploymentUnit> pool = new RefCountedObjectPool<>();
 
+    private final RefCountedObjectPool<DeploymentUnit, Lock> locks = new RefCountedObjectPool<>();
```
PakhomovAlexander, line 33: Why do you use `RefCountedObjectPool` for locks? Do you need to count them?

Flaugh24: Yep, we have to remove unused locks

PR #2224, opened 2023-06-20
IGNITE-17770 ItIgniteNodeRestartTest.testCfgGap is flaky

denis-chudov:
https://issues.apache.org/jira/browse/IGNITE-17770


modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
```diff
@@ -1229,9 +1228,20 @@ private static void checkTableWithData(Ignite ignite, String name) {
...
 
-            assertEquals(VALUE_PRODUCER.apply(i), row.stringValue("name"));
+            try {
+                row = table.keyValueView().get(null, Tuple.create().set("id", i));
+
+                assertEquals(VALUE_PRODUCER.apply(i), row.stringValue("name"));
+            } catch (TransactionException te) {
+                try {
+                    // There may be an exception if the primary replica node was stopped. We should wait for a new primary to appear.
+                    Thread.sleep(100);
```
sk0x50, line 1240: Does it mean that we just skip a row and move on to the next one? Is it correct test behavior?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -1268,10 +1270,18 @@ protected CompletableFuture<IgniteBiTuple<ClusterNode, Long>> enlist(int partId,
                 throw new TransactionException(REPLICA_UNAVAILABLE_ERR, "Failed to get the primary replica.");
             }
 
+            String consistentId = primaryPeerAndTerm.leader().consistentId();
+
+            ClusterNode node = clusterNodeResolver.apply(consistentId);
+
+            if (node == null) {
+                throw new TransactionException(REPLICA_UNAVAILABLE_ERR, "Failed to resolve the primary replica node, consistentId="
```
sk0x50, line 1278: I would change the error message a bit: `Failed to resolve the primary replica node [consistentId=" + consistentId + ']`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -446,14 +446,16 @@ private <R> CompletableFuture<R> enlistWithRetry(
                                         requestFunction.apply(tx.commitPartition(), primaryReplicaAndTerm.get2())
                                 );
                             } catch (PrimaryReplicaMissException e) {
-                                throw new TransactionException(e);
+                                throw new TransactionException(INTERNAL_ERR, e);
```
sk0x50, line 449: It seems to me, that using `TransactionException` with the `INTERNAL_ERR` code is weird. Let's check and rework (if it is needed) our TX exception handling in a separate ticket.

denis-chudov: I did it just because the constructor with only Throwable as argument is deprecated. Let's remove the error code and rework this later.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
denis-chudov: My bad, missed retry code.

sk0x50 (approved):

PR #2225, opened 2023-06-20
IGNITE-19539 Added exception mapper utility

sk0x50:
https://issues.apache.org/jira/browse/IGNITE-19539


modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperFuture.java
```diff
@@ -0,0 +1,436 @@
...
+
+    /**
+     * Creates a new instance {@link IgniteExceptionMapperFuture} with the given {@code delegate}
+     * as an underlying future that serves all requests.
+     *
+     * @param delegate Future to be wrapped.
+     * @param <U> The result Type.
+     * @return New instance {@link IgniteExceptionMapperFuture}.
+     */
+    public static <U> IgniteExceptionMapperFuture<U> of(CompletableFuture<U> delegate) {
```
ibessonov, line 54: Why not returning `delegate.handle(...)`? This entire subclass does nothing useful as far as I see.
If it does, it must be documented. Please consider removing the inheritance and making this class utilitarian.

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
```diff
@@ -0,0 +1,149 @@
...
+    /**
+     * Add a new mapping to already registered ones.
+     *
+     * @param mapper Exception mapper from internal exception to a public one.
+     * @param registeredMappings Already registered mappings.
+     * @throws IgniteException If a mapper for the given {@code clazz} already registered,
+     *      or {@code clazz} represents Java standard exception like {@link NullPointerException}, {@link IllegalArgumentException}.
+     */
+    static void registerMapping(
+            IgniteExceptionMapper<? extends Exception, ? extends Exception> mapper,
```
ibessonov, line 54: Given that `Exception` is already an upped bound for type parameters, this could be safely replaced with `IgniteExceptionMapper<?, ?>`.

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
```diff
@@ -0,0 +1,149 @@
...
+            Map<Class<? extends Exception>, IgniteExceptionMapper<? extends Exception, ? extends Exception>> registeredMappings) {
```
ibessonov, line 55: Same here, shorter types are easier to read

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
```diff
@@ -0,0 +1,149 @@
...
+        registeredMappings.put(mapper.mappingFrom(), mapper);
+    }
+
+    /**
+     * This method provides a mapping from internal exception to Ignite public ones.
+     *
+     * <p>The rules of mapping are the following:</p>
+     * <ul>
+     *     <li>any instance of {@link Error} is returned as is, except {@link AssertionError}
+     *     that always be mapped to {@link IgniteException} with the {@link ErrorGroups.Common#INTERNAL_ERR} error code.</li>
```
ibessonov, line 78: "...that _will_ always be mapped..."? I think that some word is missing here

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
```diff
@@ -0,0 +1,149 @@
...
+            Exception mapped = map(m, origin);
+
+            assert mapped instanceof IgniteException || mapped instanceof IgniteCheckedException :
+                    "Unexpected mapping of internal exception to a public one [origin=" + origin + ", mapped=" + mapped + ']';
+
+            return mapped;
+        }
+
+        // There are no exception mappings for the given exception. This case should be considered as internal error.
+        return new IgniteException(INTERNAL_ERR, origin);
```
ibessonov, line 116: Shouldn't we have a specific exception class for each error code?

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
```diff
@@ -0,0 +1,149 @@
...
+        return mapper.map(mapper.mappingFrom().cast(t));
+    }
+
+    /**
+     * Returns {@code true} if the given exception {@code t} is instance of {@link NullPointerException}, {@link IllegalArgumentException}.
+     *
+     * @return {@code true} if the given exception is Java standard exception.
+     */
+    private static boolean isJavaStandardException(Throwable t) {
+        return t instanceof NullPointerException || t instanceof IllegalArgumentException;
```
ibessonov, line 138: Interesting. Is this all that we support?

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMappersProvider.java
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.lang;
+
+import java.util.Collection;
+
+/**
+ * This interface provides the ability to register specific mappers from Ignite internal exceptions a public ones.
```
ibessonov, line 23: ''...exceptions _to_ a public ones", I guess. I think a word is missing here as well

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMappersProvider.java
```diff
@@ -0,0 +1,48 @@
...
+ *     <li>mapper should not provide any mappings for errors {@link Error}.</li>
+ * </ul>
+ */
+public interface IgniteExceptionMappersProvider {
+    /**
+     * Returns a collection of mappers to be used to map internal exceptions to public ones.
+     *
+     * @return Collection of mappers.
+     */
+    Collection<IgniteExceptionMapper<? extends Exception, ? extends Exception>> mappers();
```
ibessonov, line 47: Same not about the signature. I get that explicit is better than implicit, but this name is too long to justify the argument

modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionMapperUtilTest.java
```diff
@@ -0,0 +1,238 @@
...
+
+/**
+ * Tests mapping internal exceptions to public ones.
+ */
+public class IgniteExceptionMapperUtilTest {
+    /** Internal collection of mappers for tests. */
+    private Map<Class<? extends Exception>, IgniteExceptionMapper<? extends Exception, ? extends Exception>> mappers = new HashMap<>();
+
+    /**
+     * Tests a simple scenario od registering mapper for internal exceptions.
```
ibessonov, line 50:
```suggestion
     * Tests a simple scenario of registering mapper for internal exceptions.
```

modules/api/src/test/java/org/apache/ignite/lang/IgniteExceptionMapperUtilTest.java
```diff
@@ -0,0 +1,238 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+/**
+ * Tests mapping internal exceptions to public ones.
+ */
+public class IgniteExceptionMapperUtilTest {
+    /** Internal collection of mappers for tests. */
+    private Map<Class<? extends Exception>, IgniteExceptionMapper<? extends Exception, ? extends Exception>> mappers = new HashMap<>();
```
ibessonov, line 47: I get my point, please make types shorter

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperFuture.java
sk0x50: This makes sense to me. I will provide a utility method instead of this future.

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
sk0x50: I don't think so. Let's assume that we have hundreds of codes (it is an absolutely possible case). Should we create hundreds of exception classes? IMHO, it is OK to map several error codes to one exception class, assuming that these error codes relate to the one error group of course.

In this particular case, we can introduce a new type with a proper name, I think.

ibessonov (approved):

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
```diff
@@ -0,0 +1,170 @@
...
+
+        if (origin instanceof IgniteException || origin instanceof IgniteCheckedException) {
+            return origin;
+        }
+
+        if (isJavaStandardException(origin)) {
+            return origin;
+        }
+
+        IgniteExceptionMapper<? extends Exception, ? extends Exception> m = EXCEPTION_CONVERTERS.get(origin.getClass());
```
sanpwc, line 107: Why do we check only exact matching? Let's say InternalTransactionException -> PublicTransactionException mapper was registered. And we have SomeSpecficInternalTransactionException extends InternalTransactionException. Do we expect this SomeSpecficInternalTransactionException to be converted to PublicTransactionException? I believe we should, because from the hierarchy point of view SomeSpecficInternalTransactionException is an InternalTransactionException. 
Thus I believe we should find the most specific mapper with a top level IgniteException(INTERNAL_ERR, origin);

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
```diff
@@ -0,0 +1,170 @@
...
+     *
+     * @param origin The future to use to create a new stage.
+     * @param <T> Type os result.
+     * @return New CompletableFuture.
+     */
+    public static <T> CompletableFuture<T> convertToPublicFuture(CompletableFuture<T> origin) {
+        return origin
+                .handle((res, err) -> {
+                    if (err != null) {
+                        throw new CompletionException(mapToPublicException(unwrapCause(err)));
```
sanpwc, line 133: There's still a gap here. In case of Future of future we may have CompletionException of similar with InternalException inside. So I believe we should flat all top level futures here.

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
sanpwc: Not sure that I got an idea. What about all that common IllegalStateException, IllegalArgumentException etc? You've mentioned that 

> I think that we should not support any of them

But still we consider NPE as valid internalException to throw in public.

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionUtils.java
```diff
@@ -177,14 +177,15 @@ public static int extractCodeFrom(Throwable t) {
         return INTERNAL_ERR;
     }
 
-    // TODO: https://issues.apache.org/jira/browse/IGNITE-19539 this method should be removed or re-worked.
+    // TODO: This method should be removed or re-worked and usages should be changed to IgniteExceptionMapperUtil.mapToPublicException.
```
sanpwc, line 180: I don't think that it's valid to have todo without ignite link.

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapper.java
```diff
@@ -0,0 +1,95 @@
...
+        return from;
+    }
+
+    /**
+     * Maps the provided internal exception to a public one.
+     *
+     * @param error Exception instance to be mapped.
+     * @return Public exception instance.
+     */
+    public R map(T error) {
```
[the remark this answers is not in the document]
sk0x50: Ok, will change.

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapper.java
```diff
@@ -0,0 +1,95 @@
...
+     * Creates a new exception mapper from an internal exception {@code T} to a public runtime exception {@code R}.
+     *
+     * @param from Class instance that represents a class of internal exception.
+     * @param mapper Mapping function to map an internal exception to a public one.
+     * @param <T> Internal exception type.
+     * @param <R> Public runtime exception type.
+     *
+     * @return New instance of {@link IgniteExceptionMapper}.
+     */
+    public static <T extends Exception, R extends IgniteException> IgniteExceptionMapper<T, R> unchecked(
```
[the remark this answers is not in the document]
sk0x50: The main goal is to restrict the type of mapper. The exception mapper should provide either a mapping Exception -> IgniteException (unchecked) or Exception -> IgniteCheckedException (checked). The mapper should not try to provide a mapping Exception -> IgniteInternalException, for instance.
Java generics do not allow using something as follows:
```
    <T, R extends IgniteException> Function<T, R> mapper();
    <T, R extends IgniteCheckedException> Function<T, R> mapper();
```
just because these methods have exactly the same `erasure`

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapper.java
```diff
@@ -0,0 +1,95 @@
...
+     * Creates a new exception mapper from an internal exception {@code T} to a public checked exception {@code R}.
+     *
+     * @param from Class instance that represents a class of internal exception.
+     * @param mapper Mapping function to map an internal exception to a public one.
+     * @param <T> Internal exception type.
+     * @param <R> Public checked exception type.
+     *
+     * @return New instance of {@link IgniteExceptionMapper}.
+     */
+    public static <T extends Exception, R extends IgniteCheckedException> IgniteExceptionMapper<T, R> checked(
```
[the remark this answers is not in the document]
sk0x50: Please take a look at the comment above.

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
sk0x50: Well, It is arguable. IMHO, a particular component should provide a comprehensive list of mappers. Let's consider the following case:
```
    public class ComponentException {};
    public class SpecificExceptionA extends ComponentException {};

    // @throws SpecificExceptionA If ...
    publicApiCall() throws SpecificExceptionA
```
and after that, we decided to introduce a new type of exception to clarify the behavior in some particular case:
```
    public class SpecificExceptionB extends ComponentException {};
    or
    public class SpecificExceptionB extends SpecificExceptionA{};

    // @throws SpecificExceptionA If ...
    // @throws SpecificExceptionB If ...
    publicApiCall() throws SpecificExceptionA
```
and our component does not provide a mapping from InternalSpecificExceptionB to SpecificExceptionB.

In this case, the user will observe SpecificExceptionA or  ComponentException instead of SpecificExceptionB.
IMHO, this behavior would not be correct.

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
sk0x50: Could you please provide an example? I didn't get the idea.

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionUtils.java
sk0x50: Yep, you are right. Will create a ticket.

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
sk0x50: Well, after some thought, I think that NPE should not be considered a valid case.
```
    /**
     * @throws NullPointerException If the provided name is {@code null}.
     */
    public CompletableFuture<Res> asycnOp(String name) throws NullPointerException {
        Object.requireNonNull(name);

        // This future should not be completed with NPE
        CompletableFuture<Res> r = internalApiCall();
        ...
        return res;
    }
```

sk0x50: All in all, this particular step of mapping should be removed from the PR.

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapper.java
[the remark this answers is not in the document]
sanpwc: Yep, ok.

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
sanpwc: It's correct from the substitution principle. Alternative here, the one you've implemented, is converting SpecificExceptionB to root IgniteException, that should be considered incorrect according to aforementioned logic, meaning we have A extends B extends C, and B to B_public mapper along with anything to C_public. You claim that it's incorrect to consider A as B and thus convert it to B_public however it's correct to consider A as C and thus convert it to C_public, where C_public is IgniteException in this particular case.That seems confusing.
In real life that would actually mean that in case of TransactionInternalCheckedException->TransactionException mapper we will convert LockException extends TransactionInternalCheckedException to IgniteException instead of TransactionException, that is more specific and thus better.

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
sanpwc: I mean that it's possible to have CompletionException(ExecutionException(NPE)) or similar. However unwrapCause actually flattens stack on it's own:
```
    public static Throwable unwrapCause(Throwable e) {
        while ((e instanceof CompletionException || e instanceof ExecutionException) && e.getCause() != null) {
            e = e.getCause();
        }

        return e;
    }
```

modules/api/src/main/java/org/apache/ignite/lang/IgniteExceptionMapperUtil.java
sk0x50: Well, I don't think so, just because the logic you proposed breaks handling exceptions on the user's side, and even cannot be compiled in case of checked exceptions.

Let's consider the following examples:
(1)
```
Public hierarchy:
    ComponentException extends IgniteException;
    SpecificExceptionA extends ComponentException
    SpecificExceptionB extends ComponentException

Internal hierarchy:
    InternalComponentException extends IgniteInternalException;
    InternalSpecificExceptionA extends InternalComponentException
    InternalSpecificExceptionB extends InternalComponentException

Mappings:
    InternalComponentException -> ComponentException
    InternalSpecificExceptionA -> SpecificExceptionA
    mapping for InternalSpecificExceptionB is absent due to a bug
```
```
Public api:
    public void apiCall() throws SpecificExceptionA, SpecificExceptionB;

    try {
        apiCall();
    catch (SpecificExceptionB e) {
        // handle specific case (B)
    } catch (SpecificExceptionA e) {
        // handle specific case (A)
    } catch (ComponentException e) {
        // optional catch block, just to illustrate the issue (C)
    }
```

If I understand your proposal correctly, the mapper should convert InternalSpecificExceptionB to ComponentException, because the most specific mapping is InternalComponentException -> ComponentException and InternalSpecificExceptionB extends InternalComponentException.

In this case, exception handler (B) will be skipped at all, only (C) will be executed. I would say, this is a bug caused by the absence of the required mapping and we should clearly state that it is incorrect behavior via throwing `IgniteException(INTERNAL_ERR)` I think.

When ComponentException is a checked exception, the situation is even worse - the code should not be compiled:
```
    public static void apiCall() throws SpecificExceptionA, SpecificExceptionB {
        ...
        throw new ComponentException();
    }
```

(2) 
```
Public hierarchy:
    ComponentException extends IgniteException;
    SpecificExceptionA extends ComponentException
    SpecificExceptionB extends SpecificExceptionA
```

The fact, that the right handler is skipped, is still in play.

sanpwc: > I would say, this is a bug caused by the absence of the required mapping and we should clearly state that it is incorrect behavior via throwing IgniteException(INTERNAL_ERR) I think.

Why it's a bug? Is it mandatory to have explicit internal to public mappers for all internal exceptions? Why?

> When ComponentException is a checked exception, the situation is even worse - the code should not be compiled:

What do you mean? You either expect ComponentException to be thrown and thus add it to throws clause, or provide required mapping.

sanpwc (approved):

PR #2226, opened 2023-06-20
IGNITE-17765 Sql. Introduce cache for parsed statements

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-17765


AMashenkov:
- Implemented indirect cache for query plans: query string -> normalized query srting -> query plan
- Fixed EXPLAIN to use cached plan
- Tests added

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImpl.java
```diff
@@ -141,8 +164,59 @@ public void stop() throws Exception {
...
+            }
+
+            normalizedQueryCache.putIfAbsent(cacheKey, sqlNode.toString());
+
+            return planCache.computeIfAbsent(createCacheKey(sqlNode.toString(), ctx), k -> prepareAsync(sqlNode, ctx))
+                    .thenApply(QueryPlan::copy);
+        }
+
+        return planCache.computeIfAbsent(createCacheKey(normalizedQuery, ctx), k ->
+                        supplyAsync(() -> parse(query, queryContext, ctx)).thenCompose(sqlNode -> prepareAsync(sqlNode, ctx)))
```
lowka, line 213: Here `supplyAsync` is called w/o an executor so it uses common ForkJoinPool.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareService.java
```diff
@@ -29,5 +29,5 @@ public interface PrepareService extends LifecycleAware {
     /**
      * Prepare query plan.
      */
-    CompletableFuture<QueryPlan> prepareAsync(SqlNode sqlNode, BaseQueryContext ctx);
+    CompletableFuture<QueryPlan> prepareAsync(String query,  QueryContext queryContext, BaseQueryContext ctx);
```
xtern, line 32:
```suggestion
    CompletableFuture<QueryPlan> prepareAsync(String query, QueryContext queryContext, BaseQueryContext ctx);
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/PrepareServiceSelfTest.java
```diff
@@ -0,0 +1,273 @@
...
+        assertEquals(1, service.planCacheSize());
+    }
+
+    @Test
+    public void explainUsesCachedPlans() {
+        String query = "SELECT * FROM tbl WHERE id > 0";
+        String explainQuery = "EXPLAIN PLAN FOR SELECT * FROM tbl WHERE id > 0";
+        String explainQuery2 = "EXPLAIN PLAN FOR SELECT * /* comment */ FROM tbl WHERE id > 0";
+
+        // Ensure explain don't cache anything.
```
xtern, line 196:
```suggestion
        // Ensure explain doesn't cache anything.
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImpl.java
```diff
@@ -296,4 +364,67 @@ private ResultSetMetadata resultSetMetadata(
...
+        return sqlNode;
+    }
+
+    private static boolean skipCache(SqlNode sqlNode) {
+        SqlKind kind = sqlNode.getKind();
+
+        switch (kind) {
+            case SELECT:
+            case INSERT:
+                return false;
```
xtern, line 384: I don't quite understand why we don't caching only 'INSERT' instead of caching all DML queries?

For example, I have the following results of `InsertBenchmark` if we changing `insert` query to something with UPDATE: `update usertable set field3='a' where ycsb_key=?`

Patch not applied:
```
Benchmark                  (fsync)  Mode  Cnt  Score   Error  Units
InsertBenchmark.sqlInsert    false  avgt       1.288          ms/op
```

Patch applied:
```
Benchmark                  (fsync)  Mode  Cnt  Score   Error  Units
InsertBenchmark.sqlInsert    false  avgt       3.553          ms/op
```

AMashenkov: There is an issue with DELETE statement https://issues.apache.org/jira/browse/IGNITE-19866.
Allow caching for the rest of DML queries.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestNode.java
```diff
@@ -81,7 +80,7 @@
 public class TestNode implements LifecycleAware {
     private final String nodeName;
     private final SchemaPlus schema;
-    private final PrepareService prepareService;
+    private final PrepareServiceImpl prepareService;
```
korlov42, line 83: TestNode must use interface, not implementation

AMashenkov: fixed

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImpl.java
```diff
@@ -72,10 +84,13 @@ public class PrepareServiceImpl implements PrepareService, SchemaUpdateListener
 
     private final DdlSqlToCommandConverter ddlConverter;
 
-    private final ConcurrentMap<CacheKey, CompletableFuture<QueryPlan>> cache;
+    private final ConcurrentMap<CacheKey, CompletableFuture<QueryPlan>> planCache;
+    private final ConcurrentMap<CacheKey, String> normalizedQueryCache;
```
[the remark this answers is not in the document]
AMashenkov: fixed

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/PrepareServiceSelfTest.java
```diff
@@ -0,0 +1,294 @@
...
+        return Stream.of(
+                // Query
+                Arguments.of("SELECT * FROM tbl WHERE id > 0", OBJECT_EMPTY_ARRAY),
+                // Query with args
+                Arguments.of("SELECT * FROM tbl WHERE id > ?", new Object[]{1}),
+                // DML
+                Arguments.of("INSERT INTO tbl VALUES (1, '42')", OBJECT_EMPTY_ARRAY),
+                Arguments.of("UPDATE tbl SET VAL = '42' WHERE id = 1", OBJECT_EMPTY_ARRAY),
+                // TODO IGNITE-19866: uncomment DELETE statement
+                // Arguments.of("DELETE FROM tbl WHERE id = 1", OBJECT_EMPTY_ARRAY),
```
xtern, line 97: This issue will not reproduce if you add the PK column to the test table.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestNode.java
korlov42: well, no. Now it's even worse: interface PrepareService has two similar method, but one of them uses cache while other doesn't. If you don't believe in 0-sized caffeine cache, then introduce an abstraction for cache and provide different implementations in tests and production configuration

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/PrepareServiceSelfTest.java
AMashenkov: I believe this should work without PK.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestNode.java
```diff
@@ -203,7 +200,7 @@ public QueryPlan prepare(String query) {
     public QueryPlan prepare(SqlNode queryAst) {
         assertThat(queryAst, not(instanceOf(SqlNodeList.class)));
 
-        return await(prepareService.prepareAsync(queryAst, createContext()));
+        return await(((PrepareServiceImpl) prepareService).prepareAsync(queryAst, createContext()));
```
korlov42, line 203: once again: TestNode _must_ (I mean even cast like this is no allowed) use interface, not implementation

AMashenkov: Benchmarks need a way to prepare plan from AST, to avoid parsing stage from benchmark.
There are next ways to fix this:
* get prepare service from node (e.g. via 'unwrap' method), then cast to specific implementation
* pull `prepareAsync` method into the interface
* introduce the interface(s) with separate methods for parsing and optimization stages, introduce handler objects for these interface(s), and pass then to PrepareService constructor. Then either call `unwrap` method on prepare service to get these handlers.
* Rework benchmark to make it uses PrepareServiceImpl directly.

PR #2229, opened 2023-06-21
IGNITE-19560 Java client: fix Netty buffer leak

ptupitsyn:
* Fix certain leak in `TcpClientChannel`
* Fix potential leak in `ClientInboundMessageHandler`
* Improve leak detection further


isapego (approved):
Nice catch

PR #2232, opened 2023-06-21
IGNITE-19735 Created interface for interaction with local KeyValueStorage of the meta storage

sergeyuttsel:
https://issues.apache.org/jira/browse/IGNITE-19735


modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/MetaStorageManager.java
```diff
@@ -181,4 +181,11 @@ public interface MetaStorageManager extends IgniteComponent {
      * @return Cluster time.
      */
     ClusterTime clusterTime();
+
+    /**
+     * Returns the local meta storage.
+     *
+     * @return Storage.
+     */
+    LocalMetaStorageManager getLocalStorage();
```
rpuch, line 190: This seems to go against the Dependency Injection principle. Why can't local metastorage manager be an independent component/service that gets injected (via constructor) to each component that needs it (including the MetaStorageManager)?

PR #2240, opened 2023-06-22
IGNITE-19812 Do not swallow exceptions during start of a partition

rpuch:
https://issues.apache.org/jira/browse/IGNITE-19812


SammyVimes (approved):
LGTM

PR #2248, opened 2023-06-23
IGNITE-19626 .NET: Propagate compute deployment units

ptupitsyn:
* Propagate deployment units to .NET Compute API
* Fix NPE in `DeploymentManagerImpl.detectLatestDeployedVersion`
* Fix compute job hang in `ComputeComponentImpl.processExecuteRequest`


modules/platforms/dotnet/Apache.Ignite/Compute/ICompute.cs
```diff
@@ -15,60 +15,81 @@
...
-        /// Executes a compute job represented by the given class on one of the specified nodes.
-        /// </summary>
-        /// <param name="nodes">Nodes to use for the job execution.</param>
-        /// <param name="jobClassName">Java class name of the job to execute.</param>
-        /// <param name="args">Job arguments.</param>
-        /// <typeparam name="T">Job result type.</typeparam>
-        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
-        Task<T> ExecuteAsync<T>(IEnumerable<IClusterNode> nodes, string jobClassName, params object?[]? args);
+    /// <param name="nodes">Nodes to use for the job execution.</param>
+    /// <param name="units">Deployment units.</param>
```
isapego, line 34: Can we provide an empty list? Let's add these details to comments.

isapego (approved):

modules/compute/src/main/java/org/apache/ignite/internal/compute/ComputeComponentImpl.java
```diff
@@ -238,12 +238,15 @@ private void processExecuteRequest(ExecuteRequest executeRequest, String senderC
...
-                                        .handle((result, ex) -> sendExecuteResponse(result, ex, senderConsistentId, correlationId));
-                            }
-                    );
+                    .whenComplete((context, err) -> {
+                        if (err != null) {
+                            sendExecuteResponse(null, err, senderConsistentId, correlationId);
+                        }
+
+                        doExecuteLocally(jobClass(context.classLoader(), executeRequest.jobClassName()), executeRequest.args())
+                                .whenComplete((r, e) -> context.close())
```
rpuch, line 247: After this change, it seems that there is a possibility that the `context` will not be closed (if first `sendExecuteResponse` on line 243 throws an exception). How about moving the context closure after `whenComplete()`?

modules/platforms/dotnet/Apache.Ignite/Compute/ICompute.cs
ptupitsyn: Added "can be empty" note. Ideally we should explain what that means, but I don't have deep enough understanding of https://cwiki.apache.org/confluence/display/IGNITE/IEP-103%3A+Code+Deployment to do that.

modules/compute/src/main/java/org/apache/ignite/internal/compute/ComputeComponentImpl.java
ptupitsyn: We can't close the `context` in `whenComplete` continuation, because it completes earlier than `doExecuteLocally`. So I've added `context.close()` to the error handling block.

rpuch (approved):

PR #2249, opened 2023-06-23
IGNITE-19828 Optimize RAFT log encoder/decoder data format

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-19828


modules/raft/src/main/java/org/apache/ignite/raft/jraft/entity/codec/v1/V1Decoder.java
```diff
@@ -51,98 +52,127 @@ public LogEntry decode(final byte[] content) {
...
+    // Refactored to look closer to Ignites code style.
     public void decode(final LogEntry log, final byte[] content) {
-        // 1-5 type
-        final int iType = Bits.getInt(content, 1);
-        log.setType(EnumOutter.EntryType.forNumber(iType));
-        // 5-13 index
-        // 13-21 term
-        final long index = Bits.getLong(content, 5);
-        final long term = Bits.getLong(content, 13);
+        var reader = new Reader(content);
```
SammyVimes, line 57: Probably some of the comments should be retained. Or maybe we can describe the format of the entry here (like we do in storages)

modules/raft/src/main/java/org/apache/ignite/raft/jraft/entity/codec/v1/V1Decoder.java
```diff
@@ -51,98 +52,127 @@ public LogEntry decode(final byte[] content) {
...
+                ByteBuffer data = ByteBuffer.wrap(content, pos, len).slice();
+
+                log.setData(data);
             }
-            log.setOldLearners(oldLearners);
         }
+    }
+
+    private static int readNodesList(int pos, byte[] content, int count, List<PeerId> nodes) {
+        while (count --> 0) {
```
SammyVimes, line 133:
```suggestion
        while (count-- > 0) {
```

modules/raft/src/main/java/org/apache/ignite/raft/jraft/entity/codec/v1/V1Encoder.java
```diff
@@ -46,130 +48,137 @@ public byte[] encode(final LogEntry log) {
...
-        if (oldLearners != null) {
-            oldLearnerCount = oldLearners.size();
-            for (final PeerId oldLearner : oldLearners) {
-                final String learnerStr = oldLearner.toString();
-                // oldLearner len (short in 2 bytes)
-                // oldLearner str
-                totalLen += 2 + learnerStr.length();
-                oldLearnerStrs.add(learnerStr);
-            }
+        int totalLen = LogEntryV1CodecFactory.PAYLOAD_OFFSET;
```
SammyVimes, line 51: Same about comments/entry format

SammyVimes (changes requested):

modules/raft/src/main/java/org/apache/ignite/raft/jraft/entity/codec/v1/V1Decoder.java
ibessonov: Ok, I'll explain data format in documentation of the _factory_

modules/raft/src/main/java/org/apache/ignite/raft/jraft/entity/codec/v1/V1Decoder.java
ibessonov: Come on man!

modules/raft/src/main/java/org/apache/ignite/raft/jraft/entity/codec/v1/V1Encoder.java
```diff
@@ -46,130 +48,137 @@ public byte[] encode(final LogEntry log) {
...
+                return 7;
+            } else if (val < (1L << 56)) {
+                return 8;
+            } else {
+                return 9;
+            }
+        } else {
+            return 10;
+        }
     }
```
SammyVimes, line 183:
```suggestion
    private static int sizeInBytes(long val) {
        if (val >= 0) {
            int leadingZeros = Long.numberOfLeadingZeros(val);
            int bits = 64 - leadingZeros;
            return (bits + 6) / 7;
        } else {
            return 10;
        }
    }
```

SammyVimes (changes requested):

modules/raft/src/main/java/org/apache/ignite/raft/jraft/entity/codec/v1/V1Encoder.java
ibessonov: It's going to be slower on smaller values

SammyVimes (approved):
LGTM

PR #2251, opened 2023-06-25
IGNITE-19831 Rename RaftManager methods to better reflect their behavior

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-19831

### DIFF
--- a/modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/OnRevisionAppliedCallback.java
+++ b/modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/OnRevisionAppliedCallback.java
@@ -29,7 +29,7 @@ public interface OnRevisionAppliedCallback {
     /**
      * Notifies of completion of processing of Meta Storage watches for a particular revision.
      *
-     * @param watchEvent Event with modified Meta Storage entries processed at least one Watch.
+     * @param watchEvent Event with modified Meta Storage entries processed by at least one Watch.
      * @param newSafeTime Safe time of the applied revision.
      * @return Future that represents the state of the execution of the callback.
      */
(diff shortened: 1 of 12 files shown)


modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/OnRevisionAppliedCallback.java
```diff
@@ -29,7 +29,7 @@ public interface OnRevisionAppliedCallback {
     /**
      * Notifies of completion of processing of Meta Storage watches for a particular revision.
      *
-     * @param watchEvent Event with modified Meta Storage entries processed at least one Watch.
```
tkalkirill, line 32: According to the ticket description, this does not need to be corrected in this ticket.

modules/raft-api/src/main/java/org/apache/ignite/internal/raft/RaftManager.java
```diff
@@ -26,61 +26,91 @@
 
 /**
  * Raft manager.
+ *
+ * <p>This class contains two groups of methods for starting Raft nodes: {@code #startRaftGroupNode} and
+ * {@code startRaftGroupNodeAndWaitNodeReadyFuture} (and its overloads). When using {@code #startRaftGroupNode}, Raft log re-application
```
tkalkirill, line 31: Give an example when it is not necessary to wait for reapplying the log.

modules/raft-api/src/main/java/org/apache/ignite/internal/raft/RaftManager.java
```diff
@@ -26,61 +26,91 @@
...
+ * does not get performed and external synchronisation methods must be used to avoid observing a Raft node in inconsistent state. The other
+ * group of methods synchronously waits for the Raft log to be re-applied, so no external synchronisation is required.
  */
 public interface RaftManager extends IgniteComponent {
+    /**
+     * Starts a Raft group and a Raft service on the current node, using the given service factory.
```
tkalkirill, line 37: Add to the description that it doesn't wait for the raft log to be applied.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/OnRevisionAppliedCallback.java
sashapolo: And what do you propose? To remove this change?

tkalkirill: yep

sashapolo: Why?

PR #2254, opened 2023-06-26
IGNITE-19500 IndexManager should listen CatalogService events instead of configuration

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19500


modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -108,226 +111,164 @@ public class IndexManager extends Producer<IndexEvent, IndexEventParameters> imp
...
+            assert exception == null;
 
-            return completedFuture(false);
+            return onIndexDrop((DropIndexEventParameters) parameters);
         });
 
-        LOG.info("Index manager started");
+        if (LOG.isInfoEnabled()) {
+            LOG.info("Index manager started");
+        }
```
AMashenkov, line 148: Log level is checked in the call in depth.

modules/catalog/src/test/java/org/apache/ignite/internal/catalog/CatalogServiceSelfTest.java
```diff
@@ -263,6 +268,13 @@ public void testCreateTable() {
         assertEquals(table.primaryKeyColumns(), pkIndex.columns());
         assertTrue(pkIndex.unique());
 
+        // Validate newly created pk index
+        assertEquals(3L, pkIndex.id());
```
AMashenkov, line 272: We shouldn't rely on certain ID in tests.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -96,414 +86,180 @@ public class IndexManager extends Producer<IndexEvent, IndexEventParameters> imp
...
     }
 
-    /** {@inheritDoc} */
     @Override
     public void stop() throws Exception {
-        LOG.debug("Index manager is about to stop");
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("Index manager is about to stop");
+        }
```
AMashenkov, line 158: Why do we need an `if` statement here?  
isDebugEnabled is checked by IgniteLogger.
```
 private void logInternal(Level level, String msg, Throwable th, Object... params) {
        Objects.requireNonNull(level);

        if (!delegate.isLoggable(level)) {
            return;
        }
```

tkalkirill: Yes, you are right, but there is one caveat, a line will be created in the heap that we will not write to the log.

We discussed this issue within the team for a long time, and decided that it is better to use such a rule to check through a predicate.

Maybe this should be moved to code style.

modules/index/src/test/java/org/apache/ignite/internal/index/IndexManagerTest.java
```diff
@@ -104,74 +128,59 @@ public void setUp() {
...
-                indexManager.dropIndexAsync("sName", "nonExisting", true),
-                willThrowFast(IndexNotFoundException.class, "Index does not exist [name=\"sName\".\"nonExisting\"]")
+    @AfterEach
+    void tearDown() throws Exception {
+        IgniteUtils.closeAll(
+                vaultManager == null ? null : vaultManager::stop,
+                metaStorageManager == null ? null : metaStorageManager::stop,
+                clockWaiter == null ? null : clockWaiter::stop,
+                catalogManager == null ? null : catalogManager::stop,
+                indexManager == null ? null : indexManager::stop
```
AMashenkov, line 183: Let's add a Utility method to stop components, 
```
IgniteUtils.stopComponents(IgniteComponent... comps) {
    for (IgniteComponent c : comps) {
        if (c != null) {
            try {
                c.stop();
            } catch(Throwable th) {
                th.printStacktrace(); // log properly
            }
        }
    }
}
```
to make code looks like
```suggestion
        IgniteUtils.stopComponents(
                vaultManager,
                metaStorageManager,
                clockWaiter,
                catalogManager,
                indexManager,
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItTablesApiTest.java
```diff
@@ -255,21 +256,44 @@ public void testGetTableFromLaggedNode() {
...
     @Test
     public void testAddIndex() {
-        clusterNodes.forEach(ign -> assertNull(ign.tables().table(TABLE_NAME)));
+        Ignite ignite = clusterNodes.get(0);
 
-        Ignite ignite0 = clusterNodes.get(0);
+        createTable(ignite, TABLE_NAME);
 
-        createTable(ignite0, TABLE_NAME);
+        String indexName = "TEST_INDEX";
```
AMashenkov, line 263: Let's make it constant as it done for "TABLE_NAME".

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItTablesApiTest.java
```diff
@@ -582,29 +596,36 @@ protected void addColumnIfNotExists(Ignite node, String tableName) {
...
      *
      * @param node Cluster node.
      * @param tableName Table name.
+     * @param indexName Index name.
+     * @param failIfNotExist Throw an exception if the index does not exist.
      */
-    protected void tryToCreateIndex(Ignite node, String tableName, boolean failIfNotExist) {
-        var tmpl  = "CREATE INDEX %s testHI ON %s (valInt, valStr)";
-        var sql = String.format(tmpl, failIfNotExist ? "" : "IF NOT EXISTS", tableName);
+    protected static void tryToCreateIndex(Ignite node, String tableName, String indexName, boolean failIfNotExist) {
```
AMashenkov, line 606: We have no public methods on Table interface for creating tables\indices anymore.
Test for creating a table make sense, to check if Table is (non-)available from any node right after creation (removal).
But tests for creating/dropping indices looks uselsess here. They should be moved to another integration test and shoudl looks like `ItCreateTableDdlTest`

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItCommonApiTest.java
```diff
@@ -123,37 +115,25 @@ public void testSessionExpiration() throws Exception {
...
-                SchemaBuilders.column(keyCol, ColumnType.INT64).build(),
-                SchemaBuilders.column("time", ColumnType.time(maxTimePrecision)).asNullable(true).build(),
-                SchemaBuilders.column("timestamp", ColumnType.timestamp(maxTimePrecision)).asNullable(true).build(),
-                SchemaBuilders.column("datetime", ColumnType.datetime(maxTimePrecision)).asNullable(true).build()
-        ).withPrimaryKey(keyCol).build();
-
-        await(((TableManager) node.tables()).createTableAsync(schTblAllSql.name(), DEFAULT_ZONE_NAME, tblCh ->
-                SchemaConfigurationConverter.convert(schTblAllSql, tblCh)
-        ));
+        sql(String.format("CREATE TABLE %s(\"%s\" INTEGER PRIMARY KEY, \"TIMESTAMP\" TIMESTAMP(9))", kvTblName, keyCol));
```
AMashenkov, line 130: Why do you drop DATETIME column?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ddl/DdlCommandHandler.java
```diff
@@ -352,59 +343,42 @@ private CompletableFuture<Boolean> handleCreateIndex(CreateIndexCommand cmd) {
...
-                    createSortedIndexInternal(cmd, tableIndexChange.convert(SortedIndexChange.class));
+        boolean failIfExists = !cmd.ifNotExists();
 
-                    break;
-                case HASH:
-                    createHashIndexInternal(cmd, tableIndexChange.convert(HashIndexChange.class));
+        return catalogCreateIndexAsync(cmd)
+                .handle((unused, throwable) -> {
+                    if (throwable != null) {
+                        throwable = ExceptionUtils.unwrapCause(throwable);
```
AMashenkov, line 351:
```suggestion
                .handle(handleModificationResult(!failIfExists, IndexAlreadyExistsException.class));
```

AMashenkov: We already have postprocessing method `handleModificationResult`.
Please, reuse it.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/StopCalciteModuleTest.java
AMashenkov, line 1: I'd drop this test class, as we don't check real behaviour of SqlQueryProcessor.
There are bunch of mocks, without making clear how they affect the SqlQueryProcessor.

If we want to check all nodes stop their threads in various scenarious, then it would be more affective making this check at tearDown stage for every integration test.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItCommonApiTest.java
```diff
@@ -123,37 +115,25 @@ public void testSessionExpiration() throws Exception {
...
-        String keyCol = "key";
-        int maxTimePrecision = TemporalColumnType.MAX_TIME_PRECISION;
+        String keyCol = "KEY";
 
         Ignite node = CLUSTER_NODES.get(0);
 
         // TODO: https://issues.apache.org/jira/browse/IGNITE-19162 Trim all less than millisecond information from timestamp
         //String tsStr = "2023-03-29T08:22:33.005007Z";
-        String tsStr = "2023-03-29T08:22:33.005Z";
+        String tsStr = "2023-03-29T08:22:33Z";
```
AMashenkov, line 124: There is a bug in` CatalogUtils.fromParams(ColumnParams)` method.
Precision, scale and length are set to `0` by default, but we expected using default values described in SQL standard.
So now, table descriptor, which was built from Catalog, differs from descriptor, which was built from Configuration.

To fix the issue, next default should be used.
```
/**
     * Default TIMESTAMP type precision: microseconds.
     *
     * <p>SQL`16 part 2 section 6.1 syntax rule 36
     */
    public static final int DEFAULT_TIMESTAMP_PRECISION = 6;
    /**
     * Default TIME type precision: seconds.
     *
     * <p>SQL`16 part 2 section 6.1 syntax rule 36
     */
    public static final int DEFAULT_TIME_PRECISION = 0;
    /**
     * Default DECIMAL precision is implementation-defined.
     *
     * <p>SQL`16 part 2 section 6.1 syntax rule 25
     */
    public static final int DEFAULT_DECIMAL_PRECISION = 19;
    /**
     * Default DECIMAL scale is implementation-defined.
     *
     * <p>SQL`16 part 2 section 6.1 syntax rule 25
     */
    public static final int DEFAULT_DECIMAL_SCALE = 3;
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2718,4 +2752,39 @@ private PartitionUpdateHandlers createPartitionUpdateHandlers(
...
+        return tables.stream().filter(table -> table.name().equals(name)).findAny().orElse(null);
+    }
+
+    /**
+     * Fires table creation events so that indexes can be correctly created at IndexManager startup.
+     *
+     * <p>NOTE: This is a temporary solution that must be get rid/remake/change.
+     */
+    // TODO: IGNITE-19499 Need to get rid/remake/change
+    private void fireCreateTablesOnStartManager() {
```
rpuch, line 2766:
```suggestion
    private void fireCreateTablesOnManagerStart() {
```

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
rpuch: This is a `stop()` method that is called once per JVM. It makes no sense to save one object here, but a line of code might make a difference. Same thing for `start()`: it's not on a hot path.

tkalkirill: Okay, got rid of the checks

modules/index/src/test/java/org/apache/ignite/internal/index/IndexManagerTest.java
tkalkirill: fix it

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItTablesApiTest.java
tkalkirill: fix it

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/StopCalciteModuleTest.java
tkalkirill: Removed this class.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItTablesApiTest.java
tkalkirill: Moved tests to **org.apache.ignite.internal.sql.engine.ItIndexDdlTest**

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ddl/DdlCommandHandler.java
tkalkirill: fix it

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItCommonApiTest.java
tkalkirill: created a https://issues.apache.org/jira/browse/IGNITE-20105

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItCommonApiTest.java
tkalkirill: There is a nuance here, here I left `org.apache.ignite.sql.ColumnType#DATETIME` which in SQL has the `org.apache.calcite.sql.type.SqlTypeName#TIMESTAMP`, but I can’t use `org.apache.ignite.sql.ColumnType#TIMESTAMP` because in SQL it is type `org.apache.calcite.sql.type.SqlTypeName#TIMESTAMP_WITH_LOCAL_TIME_ZONE` which does not currently work here is the IGNITE-17376 and IGNITE-19274.

I looked at the mapping in `org.apache.ignite.internal.sql.engine.util.TypeUtils#columnType`.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItIndexDdlTest.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Tries to create the index.
+     *
+     * @param tableName Table name.
+     * @param indexName Index name.
+     * @param failIfNotExist Throw an exception if the index does not exist.
+     */
+    private static void tryToCreateIndex(String tableName, String indexName, boolean failIfNotExist) {
```
rpuch, line 84: `failIfNotExist` seems to have inverted meaning. If the parameter is true, we'll fail if the index EXISTS, not if it DOES NOT

tkalkirill: fix it

rpuch (approved):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItCommonApiTest.java
AMashenkov: Ok. Let's add these tickets to comment as well, or disable the test with these tickets numbers.

AMashenkov (approved):

PR #2259, opened 2023-06-27
IGNITE-19688 use waitAndGetLeader instead of waitLeader and getLeader

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-19688


modules/raft/src/testFixtures/java/org/apache/ignite/raft/jraft/core/TestCluster.java
```diff
@@ -408,6 +408,26 @@ public void waitLeader() throws InterruptedException {
...
 
+    /**
+     * Wait until a leader is elected.
+     * @throws InterruptedException
+     * @return Leader
+     */
+    public Node waitAndGetLeader() throws InterruptedException {
+        Node node;
+
+        while (true) {
```
vldpyatkov, line 419: Do you sure about need an infinity loop here?

alievmirza: changed to an assertion

vldpyatkov (approved):

PR #2263, opened 2023-06-27
IGNITE-19843 Unify code that triggers rebalance

rpuch:
https://issues.apache.org/jira/browse/IGNITE-19843

### DIFF
--- a/modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/DistributionZoneRebalanceEngine.java
+++ b/modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/DistributionZoneRebalanceEngine.java
@@ -175,9 +175,9 @@ public class DistributionZoneRebalanceEngine {
                         return completedFuture(null);
                     }
 
-                    for (TableView tableView : findTablesByZoneId(zoneId)) {
+                    for (TableView tableConfig : findTablesByZoneId(zoneId)) {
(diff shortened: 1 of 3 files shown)


modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/DistributionZoneRebalanceEngine.java
```diff
@@ -153,80 +149,48 @@ public CompletableFuture<Void> onUpdate(WatchEvent evt) {
...
-                                        );
-
-                                        return null;
-                                    });
+                        Set<Throwable> exceptions = newSetFromMap(new ConcurrentHashMap<>());
+                        for (int partId = 0; partId < partitionFutures.length; partId++) {
+                            int finalPartId = partId;
 
+                            partitionFutures[partId].exceptionally(e -> {
+                                if (exceptions.add(e)) {
```
sanpwc, line 188: Do we use exceptions set anywhere?

sanpwc (approved):

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/DistributionZoneRebalanceEngine.java
rpuch: I added a couple of comments to clarify the situation

PR #2265, opened 2023-06-27
IGNITE-19775 Add index builder to TestBuilders

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-19775


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestClusterTest.java
```diff
@@ -67,4 +74,19 @@ public void testSimpleQuery() {
             assertNotNull(row);
         }
     }
+
+    /**
+     * Runs a SELECT query with condition.
+     */
+    @Test
+    public void testQueryWithCondition() {
```
korlov42, line 82: Not sure I understand why do we need this test

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestClusterTest.java
```diff
@@ -45,6 +46,12 @@ public class TestClusterTest {
             .addColumn("VAL", NativeTypes.stringOf(64))
             .defaultDataProvider(dataProvider)
             .end()
+            .addHashIndex()
```
korlov42, line 49: It would be better to make index builder as child of table builder: 

```
TestBuilders.cluster()
    .nodes("N1")
    .addTable()
        .name("T1")
        .distribution(IgniteDistributions.hash(List.of(0)))
            .addColumn("ID", NativeTypes.INT32)
            .addColumn("VAL", NativeTypes.stringOf(64))
            .defaultDataProvider(dataProvider)
            .addHashIndex()
                .name("IDX_ID")
                .addColumn("ID")
                .defaultDataProvider(dataProvider)
            .end()
        .end()
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestClusterTest.java
AMashenkov: The query in the test uses index.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestClusterTest.java
AMashenkov: fixed

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestClusterTest.java
korlov42: I removed the index definition, and this test still passes. Let's make it more robust then

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestClusterTest.java
korlov42: You've changed indexes in cluster builder flow, but leave top level builder indexes flow as is. Let's make it consistent

korlov42 (approved):

PR #2271, opened 2023-06-30
IGNITE-19862: Revise error codes for SQL module

lowka:
Updates error codes used by SQL engine.

First commit:
- Adds new error codes and uses those code to represent existing code (most of them read comments in ErrorGroups).
- Adds explanations on where the existing codes are being used.
- Replaces `INVALID_QUERY_ERR` it into `PARSE_ERR` and `VALIDATION_ERR` to simplify splitting the codes.
- Adds `EXECUTION_ERR` that represent generic errors during execution + there is `INVALID_DATA_ERR` for invalid data error in runtime (such numeric overflows etc).

Second commit completely replaces old codes with new ones.


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
```diff
@@ -330,7 +329,13 @@ public Publisher<Long> executeBatchReactive(@Nullable Transaction transaction, S
     /** {@inheritDoc} */
     @Override
     public void close() {
-        await(closeAsync());
+        try {
+            closeAsync().toCompletableFuture().get();
+        } catch (ExecutionException e) {
+            throw new SqlException(RUNTIME_EXECUTION_ERR, e.getCause());
```
AMashenkov, line 335: SqlException may rewrite code from cause.
Is it done intentionally?

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -117,92 +117,35 @@ public static class Sql {
...
-        public static final int UNSUPPORTED_DDL_OPERATION_ERR = SQL_ERR_GROUP.registerErrorCode(14);
-
-        /** Query validation error. */
-        public static final int QUERY_VALIDATION_ERR = SQL_ERR_GROUP.registerErrorCode(15);
+        public static final int CURSOR_CLOSED_ERR = SQL_ERR_GROUP.registerErrorCode(6);
 
-        /** Object is not found in schema. */
-        public static final int OBJECT_NOT_FOUND_ERR = SQL_ERR_GROUP.registerErrorCode(16);
+        /** Parsing error. */
+        public static final int PARSE_ERR = SQL_ERR_GROUP.registerErrorCode(7);
```
AMashenkov, line 130:
```suggestion
        public static final int QUERY_PARSE_ERR = SQL_ERR_GROUP.registerErrorCode(7);
```

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -117,92 +117,35 @@ public static class Sql {
...
 
-        /** Object already exists in schema. */
-        public static final int OBJECT_ALREADY_EXISTS_ERR = SQL_ERR_GROUP.registerErrorCode(17);
+        /** Validation error. */
+        public static final int VALIDATION_ERR = SQL_ERR_GROUP.registerErrorCode(8);
```
AMashenkov, line 133:
```suggestion
        public static final int QUERY_VALIDATION_ERR = SQL_ERR_GROUP.registerErrorCode(8);
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -311,11 +310,11 @@ private static RuntimeException convertDdlException(Throwable e) {
         }
 
         if (e instanceof IgniteInternalCheckedException) {
-            return new IgniteInternalException(DDL_EXEC_ERR, "Failed to execute DDL statement [stmt=" /*+ qry.sql()*/
+            return new IgniteInternalException(RUNTIME_EXECUTION_ERR, "Failed to execute DDL statement [stmt=" /*+ qry.sql()*/
                     + ", err=" + e.getMessage() + ']', e);
         }
 
-        return (e instanceof RuntimeException) ? (RuntimeException) e : new SqlException(DDL_EXEC_ERR, e);
+        return (e instanceof RuntimeException) ? (RuntimeException) e : new SqlException(RUNTIME_EXECUTION_ERR, e);
```
AMashenkov, line 317: Part of stacktrace could be lost if `e` was unwrapped from CompletionException few lines above `e = e.getCause();`
BTW, ExecutionException is never unwrapped. Seems, it is just missed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/AbstractNode.java
```diff
@@ -161,7 +161,7 @@ protected void checkState() throws Exception {
             throw new ExecutionCancelledException();
         }
         if (Thread.interrupted()) {
-            throw new IgniteInternalCheckedException(OPERATION_INTERRUPTED_ERR, "Thread was interrupted.");
+            throw new IgniteInternalCheckedException(RUNTIME_EXECUTION_ERR, "Thread was interrupted.");
```
AMashenkov, line 164: Execution nodes are run in SQL query pool.
If thread is interrupted that means pool has been stopped and query was cancelled.

So, we either never get here because of `context.isCancelled()` returns `true`, or we should throw ExecutionCancelledException, isn't it?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/HashAggregateNode.java
```diff
@@ -77,11 +75,9 @@ public HashAggregateNode(
         this.accFactory = accFactory;
         this.rowFactory = rowFactory;
 
-        ImmutableBitSet.Builder b = ImmutableBitSet.builder();
+        assert grpSets.size() <= Byte.MAX_VALUE : "Too many grouping sets";
 
-        if (grpSets.size() > Byte.MAX_VALUE) {
-            throw new IgniteInternalException(TOO_MANY_GROUPING_EXPRESSIONS_ERR, "Too many groups");
-        }
+        ImmutableBitSet.Builder b = ImmutableBitSet.builder();
```
AMashenkov, line 80: Why did you changed an Exception to assertion?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/Inbox.java
```diff
@@ -365,7 +365,9 @@ private void onNodeLeft0(String nodeName) throws Exception {
         checkState();
 
         if (perNodeBuffers.get(nodeName).check() != State.END) {
-            throw new IgniteInternalCheckedException(NODE_LEFT_ERR, "Failed to execute query, node left [nodeName=" + nodeName + ']');
+            throw new IgniteInternalCheckedException(
```
AMashenkov, line 368: Let's introduce NodeLeftException and pass it as cause.
Thus, in trace user will see `RUNTIME_EXECUTION_ERR` code and `NODE_LEFT_ERR` code as a root cause.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgnitePlanner.java
```diff
@@ -264,7 +264,7 @@ public RelRoot expandView(
             sqlNode = parser.parseQuery();
         } catch (SqlParseException e) {
             //            throw new IgniteSQLException("parse failed", IgniteQueryErrorCode.PARSING, e);
-            throw new IgniteException(QUERY_INVALID_ERR, "parse failed", e);
+            throw new IgniteException(VALIDATION_ERR, "parse failed", e);
```
AMashenkov, line 267: This looks to be a parser error.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/HashAggregateNode.java
lowka: If we get here and an aggregate is invalid, we had an error at validation/planning stage.
If we want to keep an error here, we should also replace a lot of assertions in other ExecutionNodes with Exceptions as well.

lowka: I tried to reproduce this case at the validation stage (to reject queries with more than 128 groups sets), but irregardless of how many grouping sets a query has, grpSets.size() is always 1.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/Inbox.java
lowka: Yeah, such error is raised three times.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
lowka: The code behaves exactly as it has been prior to this change.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
lowka: `CompletableFuture` wraps all exceptions in `CompletionException`, so we should not get `ExecutionException ` here, unless of course we have a custom `CompletableFuture/CompletionStage` that does bother itself such details.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/AbstractNode.java
lowka: Replaced with `ExecutionCancelledException `.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
lowka: Do not have strong option on this but QUERY is unnecessary because SQL engine executes both queries and commands that do not produce results.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -117,92 +117,35 @@ public static class Sql {
...
 
-        /** DDL execution error. */
-        public static final int DDL_EXEC_ERR = SQL_ERR_GROUP.registerErrorCode(20);
+        /** Internal SQL engine error. */
+        public static final int INTERNAL_EXECUTION_ERR = SQL_ERR_GROUP.registerErrorCode(10);
 
-        /** DML result error. */
-        public static final int INVALID_DML_RESULT_ERR = SQL_ERR_GROUP.registerErrorCode(21);
+        /** Runtime SQL engine error. */
+        public static final int RUNTIME_EXECUTION_ERR = SQL_ERR_GROUP.registerErrorCode(11);
```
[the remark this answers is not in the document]
lowka: We can combine all error to internal errors but if you look more closely on where these code they are used for different situations. I do not have strong opinion on this.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
lowka: Same as `VALIDATION_ERR`. SQL executes both queries and commands that produce no results, so I think `QUERY` prefix  is unnecessary.

modules/client-handler/src/test/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImplTest.java
```diff
@@ -299,4 +299,13 @@ private long acquireConnectionId() {
 
         return result.connectionId();
     }
+
+    @SuppressWarnings("UnusedReturnValue")
```
zstan, line 303: use : IgniteTestUtils#await

lowka: Fixed

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
AMashenkov: Just to be clear that we failed to parse query string, but not cluster configuration or any JSON result or whatever.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
AMashenkov: It's not clear what validation do you mean. Query validation phase or session validation or user parameters validation or whatever?

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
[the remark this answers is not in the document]
AMashenkov: I looked at a places where INTERNAL errors are thrown. It looks like unrecoverable internal errors.
I mean that there is no way to fix the issue at runtime and possibly grid should be stopped.

On the other hand, RUNTIME exception allow user to fix query string, create missed entities or just retry because of new topology established, without grid maintenance.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
AMashenkov: Ok, ExecutionException can be thrown from `future.get()` method, but we never intend calling synchronous methods in thread pools.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
lowka: I see neither user parameter, nor session validation at the moment. I can can rename both `PARSE_ERR` and `VALIDATION_ERR` to `STMT_PARSE_ERR`, `STMT_VALIDATION_ERR`.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
lowka: Updated javadoc.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
lowka: Fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/NodeLeftException.java
```diff
@@ -0,0 +1,35 @@
...
+package org.apache.ignite.internal.sql.engine;
+
+import static org.apache.ignite.lang.ErrorGroups.Sql.RUNTIME_EXECUTION_ERR;
+
+import org.apache.ignite.lang.IgniteException;
+
+/**
+ * The exception is thrown when SQL engine can not process an operation because a node has a left cluster.
+ */
+public class NodeLeftException extends IgniteException {
```
AMashenkov, line 27: Can we move this class to api module and use error code from Common group?

AMashenkov (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/NodeLeftException.java
lowka: Sure we can but I think do we really need NodeLeftException? Maybe we can use `IgniteException` with `NODE_LEFT_ERR` code.

lowka: Moved NODE_LEFT to common error group.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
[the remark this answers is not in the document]
lowka: Renamed to `RUNTIME_EXECUTION_ERR` to `EXECUTION_ERR`.  Replaced `INTERNAL_EXECUTION_ERR`s with `INTERNAL_ERR`s

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
lowka: Fixed.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
korlov42, line 1: 1) I would say, we need neither `SESSION_NOT_FOUND_ERR` nor `SESSION_EXPIRED_ERR`. But I would preserve `EXECUTION_CANCELLED_ERR` code. The idea is to distinguish cursor closed explicitly (by invocation of closeAsync method) from those closed implicitly (by timeout, by admin request, etc).

2) For every remaining code in SQL group, we need to improve javadoc. Things to highlight: general description, some example when it's legit to throw this exception (if applicable), action points. Here is an example for `VALIDATION_ERR`:

```
Although statement is grammatically correct, the semantic is in question. This error may appears in following cases:
    * the statement refer to relation that doesn't exists
    * the statement describes action that is prohibited by the system, like changing columns belonging to primary keys
    * the statement contains operation that is not defined for given operands' types, like addition of DATE and DECIMAL
    * etc

See message for details.
```

3) rename `RUNTIME_EXECUTION_ERR` --> `RUNTIME_ERR`. This code should be used whenever the exception caused by user (like in the case `CAST('asd' as INTEGER)`)

4) rest of the runtime exceptions should be covered by INTERNAL error code

lowka: @korlov42 by INTERNAL code you mean `INTERNAL_ERR` from the common error group?

lowka: @korlov42 What should be done with these:

>if (!busyLock.enterBusy()) {
>  return CompletableFuture.failedFuture(new SqlException(SESSION_NOT_FOUND_ERR, "Session is closed."));
>}

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ddl/DdlSqlToCommandConverter.java
```diff
@@ -772,7 +766,7 @@ private <S, T> void updateCommandOption(
         try {
             value0 = value.getValueAs(optInfo.type);
         } catch (Throwable e) {
-            throw new IgniteException(QUERY_VALIDATION_ERR, String.format(
+            throw new IgniteException(STMT_VALIDATION_ERR, String.format(
```
AMashenkov, line 769:
```suggestion
            throw new SqlException(STMT_VALIDATION_ERR, String.format(
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ddl/DdlSqlToCommandConverter.java
```diff
@@ -278,7 +272,7 @@ private CreateTableCommand convertCreateTable(IgniteSqlCreateTable createTblNode
                     updateCommandOption("Table", optionKey, (SqlLiteral) option.value(), tblOptionInfo, ctx.query(), createTblCmd);
                 } else {
                     throw new IgniteException(
```
AMashenkov, line 274:
```suggestion
                    throw new SqlException(
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ddl/DdlSqlToCommandConverter.java
```diff
@@ -785,7 +779,7 @@ private <S, T> void updateCommandOption(
             try {
                 optInfo.validator.accept(value0);
             } catch (Throwable e) {
-                throw new IgniteException(QUERY_VALIDATION_ERR, String.format(
+                throw new IgniteException(STMT_VALIDATION_ERR, String.format(
```
AMashenkov, line 782:
```suggestion
                throw new SqlException(STMT_VALIDATION_ERR, String.format(
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -312,11 +311,11 @@ private static RuntimeException convertDdlException(Throwable e) {
         }
 
         if (e instanceof IgniteInternalCheckedException) {
-            return new IgniteInternalException(DDL_EXEC_ERR, "Failed to execute DDL statement [stmt=" /*+ qry.sql()*/
+            return new IgniteInternalException(INTERNAL_ERR, "Failed to execute DDL statement [stmt=" /*+ qry.sql()*/
```
AMashenkov, line 314:
```suggestion
            return new SqlException(INTERNAL_ERR, "Failed to execute DDL statement [stmt=" /*+ qry.sql()*/
```

AMashenkov (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ddl/DdlSqlToCommandConverter.java
```diff
@@ -800,7 +794,7 @@ private <S, T> void updateCommandOption(
 
     private void checkPositiveNumber(int num) {
         if (num < 0) {
-            throw new IgniteException(QUERY_VALIDATION_ERR, "Must be positive:" + num);
+            throw new IgniteException(STMT_VALIDATION_ERR, "Must be positive:" + num);
```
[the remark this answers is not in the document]
lowka: Fixed

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ddl/DdlSqlToCommandConverter.java
lowka: Fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ddl/DdlSqlToCommandConverter.java
lowka: Fixed.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
korlov42: > by INTERNAL code you mean INTERNAL_ERR from the common error group?

yep

> What should be done with these blocks

I think we need SessionClosedException in this case with dedicated error code

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/Commons.java
```diff
@@ -392,7 +392,7 @@ public static <T> T compile(Class<T> interfaceType, String body) {
 
             return (T) cbe.createInstance(new StringReader(body));
         } catch (Exception e) {
-            throw new SqlException(EXPRESSION_COMPILATION_ERR, e);
+            throw new SqlException(INTERNAL_ERR, "Unable to compile expression", e);
```
korlov42, line 395: any code from error group `Common` should be used with Ignite[Internal][Checked]Exception

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/session/SessionNotFoundException.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Thrown when session has been expired AND collected by cleaner thread.
+ */
+public class SessionNotFoundException extends IgniteInternalException {
+
+    private static final long serialVersionUID = 0L;
+
+    /** Constructor. */
+    public SessionNotFoundException(SessionId sessionId) {
+        super(format("Session not found [{}]", sessionId));
```
korlov42, line 33: constructor `IgniteInternalException(String msg)` is deprecated

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/session/SessionExpiredException.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Thrown when session has been expired but not yet been collected by cleaner thread.
+ */
+public class SessionExpiredException extends IgniteInternalException {
+
+    private static final long serialVersionUID = 0L;
+
+    /** Constructor. */
+    public SessionExpiredException(SessionId sessionId, Throwable cause) {
+        super(format("Session has been expired [{}]", sessionId), cause);
```
korlov42, line 33: the same: this constructor is deprecated

lowka: SessionExpiredException now extends RuntimeException.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/session/SessionNotFoundException.java
lowka: SessionNotFoundException now extends RuntimeException.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/Commons.java
lowka: Fixed.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
lowka: Fixced

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
lowka: Fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/session/SessionExpiredException.java
lowka: Replaced SessionExpiredException  with SessionNotFoundException

xtern: As I see - this class can be removed.

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/exception/handler/SqlExceptionHandler.java
```diff
@@ -57,11 +57,11 @@ public class SqlExceptionHandler implements ExceptionHandler<SQLException> {
...
     public SqlExceptionHandler() {
         sqlExceptionMappers.put(Client.CONNECTION_ERR, this::connectionErrUiComponent);
-        sqlExceptionMappers.put(Sql.QUERY_INVALID_ERR, this::invalidQueryErrUiComponent);
+        sqlExceptionMappers.put(Sql.STMT_PARSE_ERR, this::sqlParseErrUiComponent);
     }
 
-    private ErrorComponentBuilder invalidQueryErrUiComponent(IgniteException e) {
-        return fromExWithHeader(PARSING_ERROR_MESSAGE, e.code(), e.traceId(), e.getMessage());
+    private ErrorComponentBuilder sqlParseErrUiComponent(IgniteException e) {
+        return fromExWithHeader(PARSING_ERROR_MESSAGE, e.errorCode(), e.traceId(), e.getMessage());
```
xtern, line 64: (just for clarification) this is expected to be `e.codeAsString()` (and not just code within a group), but it doesn't work right now anyway. It just looks weird that you are changing this code and not changing it in neighboring methods :sunglasses:  

Also not so long ago a similar change was made, but vice versa https://github.com/apache/ignite-3/commit/93ebaf441b0a3b761d2ae48547d4d03d64f86b61#diff-3a374590342b51f519dc35c4b74f357241cf5ad1c6889c0f05a77ec5fd1dbf3e and it looks even more stranger...

lowka: Thanks. Fixed. I submitted this PR prior to this change of error code API and missed it after merging.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDynamicParameterTest.java
```diff
@@ -94,13 +94,18 @@ public void testDynamicParameters() {
...
 
+        t.printStackTrace(System.err);
+
         LocalDate date1 = LocalDate.parse("2022-01-01");
         LocalDate date2 = LocalDate.parse("2022-01-31");
 
         assertQuery("SELECT LAST_DAY(?)").withParams(date1).returns(date2).check();
+
+        RuntimeException x = assertThrows(RuntimeException.class, () -> sql("SELECT CAST(? AS INTEGER)", "lame"));
+        x.printStackTrace(System.err);
```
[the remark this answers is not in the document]
lowka: Fixed

modules/client-handler/src/test/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImplTest.java
```diff
@@ -123,23 +122,21 @@ public void sessionRecreatedIfExpired() throws IgniteInternalCheckedException {
 
         JdbcConnectionContext context = resourceRegistry.get(connectionId).get(JdbcConnectionContext.class);
 
-        when(queryProcessor.createSession(any())).thenAnswer(inv -> new SessionId(UUID.randomUUID()));
+        SessionId sessionId = new SessionId(UUID.randomUUID());
 
-        List<Integer> errorCodes = List.of(Sql.SESSION_EXPIRED_ERR, Sql.SESSION_NOT_FOUND_ERR);
+        when(queryProcessor.createSession(any())).thenAnswer(inv -> sessionId);
```
AMashenkov, line 127:
```suggestion
        when(queryProcessor.createSession(any())).thenReturn(sessionId);
```

xtern: Do we really need this mock?

lowka: It was there before but it is not used. Removed, thanks.

xtern (approved):

PR #2274, opened 2023-06-30
IGNITE-19892 Fix catalog module dependencies

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-19892


AMashenkov:
This refactoring is required to pass CatalogManager into various components, which should subscribe to catalog events.

As for now, I've no clue how to resolve few weird dependencies because of component configuration classes structure, like:
storage-api -> distribution-zones, distribution-zones -> schema
I hope with will be fixed after https://issues.apache.org/jira/browse/IGNITE-19752

modules/distribution-zones/build.gradle
```diff
@@ -26,6 +26,7 @@ dependencies {
     annotationProcessor project(":ignite-configuration-annotation-processor")
     annotationProcessor libs.auto.service
 
+    implementation project(':ignite-catalog')
```
korlov42, line 29: why do we need this dependency? I would postpone introducing any dependency till the moment whet it actually required

korlov42 (approved):

PR #2275, opened 2023-06-30
IGNITE-19644 Change DROP|ADD COLUMN IF (NOT) EXISTS syntax

zstan:


zstan:
@AMashenkov can u make a review plz ?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/internal/InternalSchemaTest.java
```diff
@@ -93,12 +92,23 @@ public void testDropColumns() {
...
                 null,
-                "SELECT c1, c3 FROM my"
+                "SELECT c2, c4 FROM my WHERE c1=2"
         );
 
-        assertNotNull(res.next());
+        SqlRow result = res.next();
+
+        assertNotNull(result);
+        System.err.println(result.metadata().columns());
```
lowka, line 108: I think you forgot to remove this line.

zstan: yep, fixed

zstan:
@AMashenkov done

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItCreateTableDdlTest.java
```diff
@@ -122,6 +129,55 @@ public void implicitColocationColumns() {
...
+
+        assertFalse(res.isEmpty());
+
+        sql("ALTER TABLE my DROP COLUMN c2");
+
+        res = sql("SELECT c1, c3 FROM my");
+
+        assertFalse(res.isEmpty());
+
+        sql("ALTER TABLE my ADD COLUMN (c2 INT, c4 VARCHAR)");
```
AMashenkov, line 149: Let's have a test for adding/dropping of individual column as well as adding/dropping of multiple columns.

AMashenkov (approved):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItCreateTableDdlTest.java
```diff
@@ -122,6 +129,63 @@ public void implicitColocationColumns() {
...
+        Ignite node = CLUSTER_NODES.get(0);
+
+        ConfigurationManager cfgMgr = IgniteTestUtils.getFieldValue(node, "clusterCfgMgr");
+
+        TablesConfiguration tablesConfiguration = cfgMgr.configurationRegistry().getConfiguration(TablesConfiguration.KEY);
+
+        int schIdBefore = ((ExtendedTableView) tablesConfiguration.tables().get("TEST").value()).schemaId();
+
+        sql("ALTER TABLE TEST ADD COLUMN (VAL1 INT)");
+
```
lowka, line 183: Maybe I am missing something. But nonetheless what could possibly go wrong in this test case (VAL0` and `VAL1` are distinct names).
Why this test case should be removed after some general issue related to catalog is resolved?
Maybe the link is not correct (https://issues.apache.org/jira/browse/IGNITE-19082 it's general issue for catalog migration).

zstan: this test is uninformative as for me, but Andrey say not touch it, issue is correct, after migration there is no version logic as present in reflection.

PR #2280, opened 2023-07-03
IGNITE-17765 Sql. Introduce cache for parsed statements

korlov42:
Before:
```
Benchmark                  (fsync)  Mode  Cnt   Score   Error  Units
InsertBenchmark.sqlInsert    false  avgt   20   1.937 ± 0.286  ms/op
```

After
```
Benchmark                  (fsync)  Mode  Cnt   Score   Error  Units
InsertBenchmark.sqlInsert    false  avgt   20   1.550 ± 0.204  ms/op
```

### DIFF
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImpl.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImpl.java
@@ -21,7 +21,6 @@ import static org.apache.ignite.internal.sql.engine.prepare.CacheKey.EMPTY_CLASS
 import static org.apache.ignite.internal.sql.engine.prepare.PlannerHelper.optimize;
 import static org.apache.ignite.internal.sql.engine.trait.TraitUtils.distributionPresent;
 import static org.apache.ignite.lang.ErrorGroups.Sql.QUERY_VALIDATION_ERR;
-import static org.apache.ignite.lang.ErrorGroups.Sql.UNSUPPORTED_SQL_OPERATION_KIND_ERR;
 
 import com.github.benmanes.caffeine.cache.Caffeine;
 import java.util.ArrayList;
@@ -46,12 +45,11 @@ import org.apache.ignite.internal.logger.IgniteLogger;
 import org.apache.ignite.internal.logger.Loggers;
 import org.apache.ignite.internal.sql.api.ColumnMetadataImpl;
 import org.apache.ignite.internal.sql.api.ResultSetMetadataImpl;
-import org.apache.ignite.internal.sql.engine.SqlQueryType;
 import org.apache.ignite.internal.sql.engine.prepare.ddl.DdlSqlToCommandConverter;
 import org.apache.ignite.internal.sql.engine.rel.IgniteRel;
 import org.apache.ignite.internal.sql.engine.schema.SchemaUpdateListener;
+import org.apache.ignite.internal.sql.engine.sql.ParsedResult;
 import org.apache.ignite.internal.sql.engine.util.BaseQueryContext;
-import org.apache.ignite.internal.sql.engine.util.Commons;
 import org.apache.ignite.internal.sql.engine.util.TypeUtils;
 import org.apache.ignite.internal.storage.DataStorageManager;
 import org.apache.ignite.internal.thread.NamedThreadFactory;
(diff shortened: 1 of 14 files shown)


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/sql/ParserServiceImpl.java
```diff
@@ -0,0 +1,132 @@
...
+        SqlQueryType queryType = Commons.getQueryType(parsedTree);
+
+        assert queryType != null : parsedTree.toString();
+
+        result = new ParsedResultImpl(
+                queryType,
+                query,
+                parsedTree.toString(),
+                parsedStatement.dynamicParamsCount(),
+                () -> IgniteSqlParser.parse(query, StatementParseResult.MODE).statement()
```
xtern, line 67: Is it a good idea to parse each non-cached query twice? :thinking:

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImpl.java
```diff
@@ -183,19 +174,25 @@ public void onSchemaUpdated() {
...
             IgnitePlanner planner = ctx.planner();
 
+            SqlNode sqlNode = parsedResult.parsedTree();
+
+            assert single(sqlNode);
+
             // Validate
             // We extract query subtree inside the validator.
-            SqlNode explainNode = planner.validate(explain);
+            SqlNode explainNode = planner.validate(sqlNode);
```
AMashenkov, line 195: Explain never uses cached plan.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/sql/ParserServiceImpl.java
AMashenkov: The very first query will be parsed twice, which looks acceptable. 
The next queries will be parsed once if either parse result will be evicted from parser cache or a plan was evicted from planner cache. 

I think we can reuse `parsedTree` for the queries, which plans shouldn't be cached, because `parsedTree` is not shared object in that case.

xtern: > The very first query will be parsed twice.
It seems to me that this can be fixed with simple atomic bool/ref, something like
```
() -> staleTreeFlag.compareAndSet(false, true) ? parsedTree : IgniteSqlParser.parse(query, StatementParseResult.MODE).statement()
```

AMashenkov: > It seems to me that this can be fixed with simple atomic bool/ref, something like

Or have 2 implementations of ParsedResult.

korlov42: Folks, I agree that this looks weird at the first glance. But here is my justification of current approach:

1. we have 4 types of queries: QUERY, DML, DDL, and EXPLAIN
2. first two should pass an optimisation step, which takes from dozens to thousands milliseconds. I believe an extra millisecond won't make any difference here.
3. Both DDL and EXPLAIN  neither hot nor recurrent. Besides, DDL processing times mostly dictated by Delay Duration time, which is hundreds of milliseconds. Thus I believe an extra millisecond won't make any difference here as well.
4. An extra IF branches | ternary operators | any exceptional handling increases complexity of the code while brings (almost) no value

If you still believe it does make sense to avoid double parsing, I will be happy to improve this

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImpl.java
korlov42: This patch was never meant to address this issue. I've filed separate JIRA case for this https://issues.apache.org/jira/browse/IGNITE-19908

AMashenkov (approved):

korlov42:
@AMashenkov, @xtern, guys, I've made some refactoring and introduced tests for ParserServiceImpl. Take a look once again please

AMashenkov (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/sql/ParserServiceImpl.java
```diff
@@ -68,7 +66,7 @@ public ParsedResult parse(String query) {
         );
 
         if (shouldBeCached(queryType)) {
-            queryToParsedResultCache.put(query, result);
+            queryToParsedResultCache.computeIfAbsent(query, k -> result);
```
xtern, line 69: Why do we use `compute` instead of `put`?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/sql/ParserServiceImplTest.java
```diff
@@ -0,0 +1,172 @@
...
+                }
+
+                @Override
+                public void clear() {
+
+                }
+            };
+        }
+    }
+}
```
xtern, line 172:
```suggestion
}

```

korlov42: fixed

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/sql/ParserServiceImpl.java
korlov42: I tried to make Cache interface minimal yet sufficient. But if it causes confusion, it's better to replace this call with `put`. 

Fixed.

xtern (approved):

PR #2283, opened 2023-07-03
IGNITE-19903 Fix recovery descriptor race condition

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-19903

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/future/OrderingFuture.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/future/OrderingFuture.java
@@ -29,6 +29,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 import org.jetbrains.annotations.Nullable;
 
@@ -299,8 +300,8 @@ public class OrderingFuture<T> {
     private static <T> void acceptQuietly(BiConsumer<? super T, ? super Throwable> action, T result, Throwable ex) {
         try {
             action.accept(result, ex);
-        } catch (Exception e) {
-            // ignore
+        } catch (Exception ignored) {
+            // No-op.
         }
     }
 
@@ -328,6 +329,41 @@ public class OrderingFuture<T> {
 
             if (dependent == null) {
                 dependent = new ThenComposeToCompletable<>(new CompletableFuture<>(), mapper);
+            }
+            State<T> newState = prevState.enqueueDependent(dependent);
+
+            if (replaceState(prevState, newState)) {
+                return dependent.resultFuture;
+            }
+        }
+    }
+
+    /**
+     * Adds a mapping function that gets executed as soon as this future gets completed for any reason. The function will accept both result
+     * and exception and return a future with the result of the function's execution.
+     *
+     * @param mapper The function to use to compute the value of the returned CompletionStage.
+     * @return The new CompletionStage.
+     * @see CompletableFuture#handle(BiFunction)
+     */
+    public <U> OrderingFuture<U> handle(BiFunction<? super T, Throwable, ? extends U> mapper) {
+        Handle<T, U> dependent = null;
+
+        while (true) {
+            State<T> prevState = state;
+
+            if (prevState.completionQueueProcessed()) {
+                try {
+                    U mappingResult = mapper.apply(prevState.result, prevState.exception);
+
+                    return completedFuture(mappingResult);
+                } catch (Throwable t) {
+                    return failedFuture(t);
+                }
+            }
+
+            if (dependent == null) {
+                dependent = new Handle<>(new OrderingFuture<>(), mapper);
             }
             State<T> newState = prevState.enqueueDependent(dependent);
 
@@ -463,6 +499,25 @@ public class OrderingFuture<T> {
         }
     }
 
+    private static class Handle<T, U> implements DependentAction<T> {
+        private final OrderingFuture<U> resultFuture;
+        private final BiFunction<? super T, Throwable, ? extends U> action;
+
+        private Handle(OrderingFuture<U> resultFuture, BiFunction<? super T, Throwable, ? extends U> action) {
+            this.resultFuture = resultFuture;
+            this.action = action;
+        }
+
+        @Override
+        public void onCompletion(@Nullable T result, @Nullable Throwable ex, NotificationContext context) {
+            try {
+                resultFuture.complete(action.apply(result, ex));
+            } catch (Throwable t) {
+                resultFuture.completeExceptionally(t);
+            }
+        }
+    }
+
     private static class ThenComposeToCompletable<T, U> implements DependentAction<T>, BiConsumer<U, Throwable> {
         private final CompletableFuture<U> resultFuture;
         private final Function<? super T, ? extends CompletableFuture<U>> mapper;
@@ -563,8 +618,8 @@ public class OrderingFuture<T> {
 
                 try {
                     node.dependent.onCompletion(result, exception, context);
-                } catch (Exception e) {
-                    // ignore
+                } catch (Exception ignored) {
+                    // No-op.
                 }
             }
         }
--- a/modules/core/src/main/java/org/apache/ignite/internal/future/OrderingFuture.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/future/OrderingFuture.java
@@ -342,8 +342,8 @@ public class OrderingFuture<T> {
      * Adds a mapping function that gets executed as soon as this future gets completed for any reason. The function will accept both result
      * and exception and return a future with the result of the function's execution.
      *
-     * @param mapper The function to use to compute the value of the returned CompletionStage.
-     * @return The new CompletionStage.
+     * @param mapper The function to use to compute the value of the returned OrderingFuture.
+     * @return The new OrderingFuture.
      * @see CompletableFuture#handle(BiFunction)
      */
     public <U> OrderingFuture<U> handle(BiFunction<? super T, Throwable, ? extends U> mapper) {
--- a/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
+++ b/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
@@ -53,6 +53,9 @@ import org.apache.ignite.internal.util.ExceptionUtils;
 import org.apache.ignite.lang.IgniteInternalException;
 import org.apache.ignite.lang.IgniteStringFormatter;
 import org.hamcrest.CustomMatcher;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeMatcher;
 import org.jetbrains.annotations.Nullable;
 import org.junit.jupiter.api.TestInfo;
 
@@ -890,5 +893,41 @@ public final class IgniteTestUtils {
         public boolean matches(Object o) {
             return predicate.test((DataT) o);
         }
+    }
+
+    /**
+     * Creates a matcher that tests if a CompletableFuture has completed successfully.
+     *
+     * @param <T> the type of the CompletableFuture
+     * @return a matcher for a successfully completed CompletableFuture
+     */
+    public static <T> Matcher<CompletableFuture<T>> completedSuccessfully() {
+        return new TypeSafeMatcher<>() {
+            @Override
+            protected boolean matchesSafely(CompletableFuture<T> future) {
+                return future.isDone() && !future.isCompletedExceptionally() && !future.isCancelled();
+            }
+
+            @Override
+            public void describeTo(Description description) {
+                description.appendText("is a successfully completed CompletableFuture");
+            }
+
+            @Override
+            public void describeMismatchSafely(CompletableFuture<T> item, Description mismatchDescription) {
+                if (!item.isDone()) {
+                    mismatchDescription.appendText("was not completed");
+                } else {
+                    if (item.isCompletedExceptionally()) {
+                        mismatchDescription.appendText("completed exceptionally");
+                    } else if (item.isCancelled()) {
+                        mismatchDescription.appendText("was cancelled");
+                    } else {
+                        // It might be successfully done now, but it wasn't at the moment of matchesSafely execution.
+                        mismatchDescription.appendText("was not completed");
+                    }
+                }
+            }
+        };
     }
 }
--- a/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
+++ b/modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
@@ -53,9 +53,6 @@ import org.apache.ignite.internal.util.ExceptionUtils;
 import org.apache.ignite.lang.IgniteInternalException;
 import org.apache.ignite.lang.IgniteStringFormatter;
 import org.hamcrest.CustomMatcher;
-import org.hamcrest.Description;
-import org.hamcrest.Matcher;
-import org.hamcrest.TypeSafeMatcher;
 import org.jetbrains.annotations.Nullable;
 import org.junit.jupiter.api.TestInfo;
 
@@ -893,41 +890,5 @@ public final class IgniteTestUtils {
         public boolean matches(Object o) {
             return predicate.test((DataT) o);
         }
-    }
-
-    /**
-     * Creates a matcher that tests if a CompletableFuture has completed successfully.
-     *
-     * @param <T> the type of the CompletableFuture
-     * @return a matcher for a successfully completed CompletableFuture
-     */
-    public static <T> Matcher<CompletableFuture<T>> completedSuccessfully() {
-        return new TypeSafeMatcher<>() {
-            @Override
-            protected boolean matchesSafely(CompletableFuture<T> future) {
-                return future.isDone() && !future.isCompletedExceptionally() && !future.isCancelled();
-            }
-
-            @Override
-            public void describeTo(Description description) {
-                description.appendText("is a successfully completed CompletableFuture");
-            }
-
-            @Override
-            public void describeMismatchSafely(CompletableFuture<T> item, Description mismatchDescription) {
-                if (!item.isDone()) {
-                    mismatchDescription.appendText("was not completed");
-                } else {
-                    if (item.isCompletedExceptionally()) {
-                        mismatchDescription.appendText("completed exceptionally");
-                    } else if (item.isCancelled()) {
-                        mismatchDescription.appendText("was cancelled");
-                    } else {
-                        // It might be successfully done now, but it wasn't at the moment of matchesSafely execution.
-                        mismatchDescription.appendText("was not completed");
-                    }
-                }
-            }
-        };
     }
 }
--- a/modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
+++ b/modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
@@ -34,7 +34,6 @@ import static org.mockito.Mockito.when;
 import io.netty.handler.codec.DecoderException;
 import java.io.IOException;
 import java.net.InetSocketAddress;
-import java.net.SocketAddress;
 import java.nio.channels.ClosedChannelException;
 import java.util.ArrayList;
 import java.util.Collection;
--- a/modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
+++ b/modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
@@ -32,6 +32,7 @@ import static org.mockito.Mockito.when;
 import io.netty.handler.codec.DecoderException;
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.nio.channels.ClosedChannelException;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -59,6 +60,7 @@ import org.apache.ignite.network.NetworkMessage;
 import org.apache.ignite.network.OutNetworkObject;
 import org.apache.ignite.network.serialization.MessageSerializationRegistry;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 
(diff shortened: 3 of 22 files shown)


modules/network/src/main/java/org/apache/ignite/internal/network/netty/ChannelCreationListener.java
```diff
@@ -0,0 +1,39 @@
...
+/** Channel creation listener. */
+public interface ChannelCreationListener {
+    /**
+     * Notifies of creation of a new channel before the handshake.
+     *
+     * @param consistentId Remote node's consistent id.
+     * @param channelId Channel id.
+     * @return A future which is completed when listener's action is done.
+     */
+    CompletableFuture<Void> notifyInboundChannelCreation(String consistentId, short channelId);
```
ibessonov, line 31: I don't see a unique launch ID. Is it not available yet?
If it's not available, then how do you know node's consistent ID?
Or maybe it's not required here. You see, it's hard to understand your intentions with this interface.

modules/network/src/main/java/org/apache/ignite/internal/network/netty/ChannelCreationListener.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Notifies of the handshake's finish.
+     *
+     * @param channel Opened channel.
+     */
+    void handshakeFinished(NettySender channel);
```
ibessonov, line 38: Off-topic: name "sender" is very confusing, don't you think so?

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
```diff
@@ -136,14 +162,29 @@ public void onMessage(NetworkMessage message) {
             this.remoteLaunchId = msg.launchId();
             this.remoteConsistentId = msg.consistentId();
 
-            this.recoveryDescriptor = recoveryDescriptorProvider.getRecoveryDescriptor(
-                    remoteConsistentId,
-                    remoteLaunchId,
-                    connectionId,
-                    false
-            );
+            channelCreationListener.notifyInboundChannelCreation(this.remoteConsistentId, this.connectionId)
```
ibessonov, line 165: Maybe all of this should be extracted to `onHandshakeStartMessage` instead of being inside of a single big method

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
```diff
@@ -136,14 +162,29 @@ public void onMessage(NetworkMessage message) {
...
+                                connectionId,
+                                false
+                        );
+
+                        if (!descriptor.acquire(ctx)) {
+                            String err = "Failed to acquire recovery descriptor during handshake, it is held by: " + descriptor.holder();
+
+                            LOG.warn(err);
+
+                            handshakeCompleteFuture.completeExceptionally(new HandshakeException(err));
```
ibessonov, line 179: Maybe you should also close the connection or send some response, like "already opened"

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryDescriptor.java
```diff
@@ -124,4 +127,30 @@ public long onReceive() {
...
+     *
+     * @param ctx Channel handler context.
+     */
+    public boolean acquire(ChannelHandlerContext ctx) {
+        return ctxHolder.compareAndSet(null, ctx.channel());
+    }
+
+    public String holder() {
+        return ctxHolder.get().toString();
+    }
```
ibessonov, line 155: I'm sorry, but what is this?

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryServerHandshakeManager.java
```diff
@@ -148,24 +174,55 @@ public void onMessage(NetworkMessage message) {
...
-            this.remoteLaunchId = msg.launchId();
-            this.remoteConsistentId = msg.consistentId();
-            this.receivedCount = msg.receivedCount();
-            this.remoteChannelId = msg.connectionId();
+            this.remoteLaunchId = remoteLaunchId;
+            this.remoteConsistentId = remoteConsistentId;
+            this.receivedCount = remoteReceivedCount;
+            this.remoteChannelId = remoteChannelId;
+
+            channelCreationListener.notifyInboundChannelCreation(this.remoteConsistentId, this.remoteChannelId)
```
ibessonov, line 193: This looks copy-pasted. Also, both Client and Server notify that inbound channel is created. Which cannot be true, maybe that method needs to be renamed.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/LeaseTracker.java
```diff
@@ -214,7 +214,6 @@ public CompletableFuture<LeaseMeta> getPrimaryReplica(ReplicationGroupId replica
                 return completedFuture(lease);
             }
 
-            // TODO: https://issues.apache.org/jira/browse/IGNITE-19532 Race between meta storage safe time publication and listeners.
```
ibessonov, line 217: Did you fix it? Or it was here by mistake?

SammyVimes: This is a leftover from other ticket

modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
```diff
@@ -391,22 +392,12 @@ public Collection<NettyClient> clients() {
...
-        public RecoveryClientHandshakeManager create(UUID launchId,
-                String consistentId,
-                short connectionId,
-                RecoveryDescriptorProvider recoveryDescriptorProvider
-        ) {
-            return new RecoveryClientHandshakeManager(launchId, consistentId, connectionId, recoveryDescriptorProvider, staleIdDetector);
-        }
+    @TestOnly
+    public Map<ConnectorKey<String>, NettySender> channels() {
+        return new HashMap<>(channels);
```
ibessonov, line 401: Maybe `java.util.Map#copyOf` would look more appropriate, but it's up to you

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
```diff
@@ -136,14 +158,35 @@ public void onMessage(NetworkMessage message) {
...
                     remoteConsistentId,
                     remoteLaunchId,
-                    connectionId,
-                    false
+                    connectionId
             );
 
-            handshake(recoveryDescriptor);
+            while (!descriptor.acquire(ctx)) {
+                if (launchId.compareTo(remoteLaunchId) > 0) {
```
ibessonov, line 168: Please add clarifying comment, that this is a tie breaker in case of simultaneous connection openings from both side. It's not obvious from the code

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
```diff
@@ -136,14 +158,35 @@ public void onMessage(NetworkMessage message) {
...
+                    Channel holderChannel = descriptor.holderChannel();
+
+                    if (holderChannel == null) {
+                        continue;
+                    }
+
+                    holderChannel.close().awaitUninterruptibly();
```
ibessonov, line 175: Is it OK to block the thread? How fast is this operation?
Anyway, uninterruptible awaiting sometimes causes tests to timeout, and you can't kill them by interruption...

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryDescriptor.java
```diff
@@ -124,4 +128,45 @@ public long onReceive() {
...
+
+    private final AtomicReference<Channel> ctxHolder = new AtomicReference<>();
+
+    /**
+     * Release this descriptor.
+     *
+     * @param ctx Channel handler context.
+     */
+    public void release(ChannelHandlerContext ctx) {
+        ctxHolder.compareAndSet(ctx.channel(), null);
```
ibessonov, line 140: Should we assert the result, just in case?

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryServerHandshakeManager.java
```diff
@@ -148,24 +170,51 @@ public void onMessage(NetworkMessage message) {
...
+            this.remoteChannelId = remoteChannelId;
+
+            RecoveryDescriptor descriptor = recoveryDescriptorProvider.getRecoveryDescriptor(
+                    this.remoteConsistentId,
+                    this.remoteLaunchId,
+                    this.remoteChannelId
+            );
+
+            while (!descriptor.acquire(ctx)) {
+                if (launchId.compareTo(remoteLaunchId) <= 0) {
```
ibessonov, line 196: Why do we have the opposite condition here?
Should we extract this code into a common place?
How do we tell the other side that we reject the connection?

modules/core/src/main/java/org/apache/ignite/internal/future/OrderingFuture.java
```diff
@@ -337,6 +338,41 @@ public <U> CompletableFuture<U> thenComposeToCompletable(Function<? super T, ? e
         }
     }
 
+    /**
+     * Adds a mapping function that gets executed as soon as this future gets completed for any reason. The function will accept both result
+     * and exception and return a future with the result of the function's execution.
+     *
+     * @param mapper The function to use to compute the value of the returned CompletionStage.
+     * @return The new CompletionStage.
```
rpuch, line 346:
```suggestion
     * @param mapper The function to use to compute the value of the returned OrderingFuture.
     * @return The new OrderingFuture.
```

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
```diff
@@ -891,4 +894,40 @@ public boolean matches(Object o) {
...
         }
     }
+
+    /**
+     * Creates a matcher that tests if a CompletableFuture has completed successfully.
+     *
+     * @param <T> the type of the CompletableFuture
+     * @return a matcher for a successfully completed CompletableFuture
+     */
+    public static <T> Matcher<CompletableFuture<T>> completedSuccessfully() {
```
rpuch, line 904: How about moving this to `CompletableFutureMatcher`?

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
```diff
@@ -295,6 +299,99 @@ public void testStopTwice() throws Exception {
...
+            CompletableFuture<NettySender> fut1 = manager1.openChannelTo(manager2).toCompletableFuture();
+            CompletableFuture<NettySender> fut2 = manager2.openChannelTo(manager1).toCompletableFuture();
+
+            NettySender sender1 = null;
+            NettySender sender2 = null;
+
+            try {
+                sender1 = fut1.join();
+            } catch (Exception ignored) {
+                // No-op.
```
rpuch, line 322: Is it correct to suggest that for exactly one `.join()` a `ChannelAlreadyExistsException` will be thrown? If yes, let's assert that the exception is what we expect it to be

modules/network/src/integrationTest/java/org/apache/ignite/network/scalecube/ItScaleCubeNetworkMessagingTest.java
```diff
@@ -355,25 +357,44 @@ public void nodeCannotReuseOldId(TestInfo testInfo) throws Exception {
...
+        };
 
-        testCluster.members.stream()
+        List<ClusterService> notOutcasts = testCluster.members.stream()
                 .filter(service -> !outcastName.equals(service.nodeName()))
-                .forEach(service -> {
-                    DefaultMessagingService messagingService = (DefaultMessagingService) service.messagingService();
-                    messagingService.dropMessages((recipientConsistentId, message) -> outcastName.equals(recipientConsistentId));
-                });
+                .collect(Collectors.toList());
```
rpuch, line 373: Let's statically import `toList()`

modules/network/src/integrationTest/java/org/apache/ignite/network/scalecube/ItScaleCubeNetworkMessagingTest.java
```diff
@@ -355,25 +357,44 @@ public void nodeCannotReuseOldId(TestInfo testInfo) throws Exception {
...
+        DefaultMessagingService messagingService = (DefaultMessagingService) testCluster.members.stream()
+                .filter(service -> outcastName.equals(service.nodeName()))
+                .findFirst()
+                .get().messagingService();
+
+        ConnectionManager cm = messagingService.connectionManager();
+
+        // Forcefully close channels, so that on reanimation nodes will create new channels.
+        cm.channels().forEach((stringConnectorKey, nettySender) -> {
+            nettySender.close().awaitUninterruptibly();
```
rpuch, line 396: I wonder should we do this (close channels) in the production code when we understand that the node has left (or is going to leave)? Otherwise, it looks suspicious: maybe it's a bug?

modules/network/src/main/java/org/apache/ignite/internal/network/handshake/ChannelAlreadyExistsException.java
```diff
@@ -0,0 +1,35 @@
...
+public class ChannelAlreadyExistsException extends RuntimeException {
+    private static final long serialVersionUID = 0L;
+
+    private final String consistentId;
+
+    public ChannelAlreadyExistsException(String consistentId) {
+        this.consistentId = consistentId;
+    }
+
+    public String consistentId() {
```
rpuch, line 32: Let's add a javadoc explaining what consistentId is returned here

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
rpuch, line 168: Could you please add a comment about this comparison (or extract it to a method with a clarifying name)?

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryDescriptor.java
```diff
@@ -124,4 +128,45 @@ public long onReceive() {
     public String toString() {
         return S.toString(RecoveryDescriptor.class, this);
     }
+
+    private final AtomicReference<Channel> ctxHolder = new AtomicReference<>();
```
rpuch, line 132: Why is it called `ctxHolder` if it holds a Channel?

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryServerHandshakeManager.java
rpuch, line 196: Please add a comment or clarify this by extracting a method

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
```diff
@@ -295,6 +299,99 @@ public void testStopTwice() throws Exception {
...
+     * @throws Exception If failed.
+     */
+    @RepeatedTest(100)
+    public void testOneChannelLeftIfConnectToEachOther() throws Exception {
+        try (
+                ConnectionManagerWrapper manager1 = startManager(4000);
+                ConnectionManagerWrapper manager2 = startManager(4001)
+        ) {
+            CompletableFuture<NettySender> fut1 = manager1.openChannelTo(manager2).toCompletableFuture();
+            CompletableFuture<NettySender> fut2 = manager2.openChannelTo(manager1).toCompletableFuture();
```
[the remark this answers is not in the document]
SammyVimes: I ran it 1000 times, it's good, no flakiness

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
```diff
@@ -295,6 +299,99 @@ public void testStopTwice() throws Exception {
...
+
+            NettySender sender1 = null;
+            NettySender sender2 = null;
+
+            try {
+                sender1 = fut1.join();
```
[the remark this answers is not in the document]
SammyVimes: It may or may not complete successfully. But I will change to get with a timeout, sure

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
```diff
@@ -295,6 +299,99 @@ public void testStopTwice() throws Exception {
...
+            } catch (Exception ignored) {
+                // No-op.
+            }
```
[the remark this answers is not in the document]
SammyVimes: We are explicitly testing the result later, we don't need to test the specific exception, it will just bloat test even more

modules/network/src/integrationTest/java/org/apache/ignite/network/scalecube/ItScaleCubeNetworkMessagingTest.java
```diff
@@ -355,25 +357,44 @@ public void nodeCannotReuseOldId(TestInfo testInfo) throws Exception {
...
             public void onDisappeared(ClusterNode member) {
                 if (Objects.equals(member.name(), outcastName)) {
                     disappeared.countDown();
                 }
             }
-        });
+        };
 
-        testCluster.members.stream()
+        List<ClusterService> notOutcasts = testCluster.members.stream()
```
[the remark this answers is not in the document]
SammyVimes: I prefer setting event handler prior to dropping messages, because dropping messages from node1 to node2 may trigger "Node left event" on node3 because of the scalecube's nature

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
SammyVimes: Sure, I forgor

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryDescriptor.java
SammyVimes: Not really, we don't care if release was called more than once

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/IgniteTestUtils.java
SammyVimes: I moved it to a separate class

modules/network/src/integrationTest/java/org/apache/ignite/network/scalecube/ItScaleCubeNetworkMessagingTest.java
SammyVimes: I'm not sure. We will close a channel if it was closed. In this case, we artificially drop messages, but the channel is still alive, that's why channel is not closed. In normal conditions, I expect channel to be closed automatically either due to a channel reset or a timeout

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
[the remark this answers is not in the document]
ibessonov: I meant the rate before the fix. Of course it must be 0 after

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
```diff
@@ -295,6 +299,95 @@ public void testStopTwice() throws Exception {
...
+                    manager2.connectionManager.localAddress()
+            ).toCompletableFuture();
+
+            CompletableFuture<NettySender> channelFut2 = manager2.connectionManager.channel(
+                    manager1.connectionManager.consistentId(),
+                    ChannelType.DEFAULT,
+                    manager1.connectionManager.localAddress()
+            ).toCompletableFuture();
+
+            assertThat(channelFut1, futureCompleted());
```
ibessonov, line 372: I should go like "assert that <object> <verb> <noun>" or something.
`assertThat(channelFut, is(completedFuture()))` would look more natural. Current version just doesn't seem right. May I ask you to rename it?

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
ibessonov: Considering that this code is duplicated 2 times, I strongly suggest extracting it into a single place.

rpuch (commented):
Looks great, but it would be nice to have those `compareTo()`-related comments addressed :)

modules/network/src/main/java/org/apache/ignite/internal/network/netty/ChannelCreationListener.java
SammyVimes: Removed the method altogether

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
SammyVimes: Done

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryDescriptor.java
SammyVimes: Updated javadoc

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
SammyVimes: It is a good question, however in general this should not occur very often, only on first connections. I'd leave it like that for the sake of simplicity. Otherwise we'd need to introduce a future here and handle it in the retry style, which looks horrible

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryServerHandshakeManager.java
SammyVimes: See comment

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
SammyVimes: Done

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryServerHandshakeManager.java
SammyVimes: Done

modules/network/src/integrationTest/java/org/apache/ignite/internal/network/netty/ItConnectionManagerTest.java
SammyVimes: what is `is`?

ibessonov (approved):

rpuch (approved):

PR #2284, opened 2023-07-04
IGNITE-19212 ODBC: Implement basic query execution

isapego:
- Ported basic query execution functionality from Ignite 2;
- Ported tests;
- Adapted to use new protocol;


modules/platforms/cpp/ignite/client/detail/sql/result_set_impl.h
```diff
@@ -154,15 +154,15 @@ class result_set_impl : public std::enable_shared_from_this<result_set_impl> {
      *
      * @return @c true if there are more pages with results and @c false otherwise.
      */
-    [[nodiscard]] IGNITE_API bool has_more_pages() { return m_resource_id.has_value() && m_has_more_pages; }
+    [[nodiscard]] bool has_more_pages() { return m_resource_id.has_value() && m_has_more_pages; }
```
ptupitsyn, line 157: Is `IGNITE_API` removed because this is `_imp.h`?

modules/platforms/cpp/ignite/odbc/query/cursor.h
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
ptupitsyn, line 2: Incorrect header.

ptupitsyn (changes requested):

modules/platforms/cpp/ignite/client/detail/sql/result_set_impl.h
isapego: Yeah, those methods should not be in API

modules/platforms/cpp/ignite/odbc/query/cursor.h
isapego: Fixed.

ptupitsyn (approved):

PR #2286, opened 2023-07-04
IGNITE-19499 TableManager should listen CatalogService events instead of configuration

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-19499


AMashenkov:
Fixed the issue, when wrong default precision and scale values were passed to Catalog.
DistributionZoneManager push changes to Catalog as well as to configuration.
TableManager push changes to Catalog as well as to Configuration.
TableManager accepts DDL command parameters instead of configuration changer object.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1276,18 +1282,22 @@ private CompletableFuture<?> createTableLocally(
...
 
-        tablesById(causalityToken).thenAccept(ignored -> inBusyLock(busyLock, () -> {
+        createPartsFut.thenAccept(ignored -> inBusyLock(busyLock, () -> {
             pendingTables.remove(tableId);
         }));
 
-        tablesById(causalityToken)
-                .thenRun(() -> inBusyLock(busyLock, () -> completeApiCreateFuture(table)));
+        createPartsFut
+                .thenRunAsync(() -> inBusyLock(busyLock, () -> completeApiCreateFuture(table)));
```
tkalkirill, line 1290: Missing thread pool

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1462,32 +1472,53 @@ private Set<Assignment> calculateAssignments(TablePartitionId tablePartitionId)
...
      */
-    public CompletableFuture<Table> createTableAsync(String name, String zoneName, Consumer<TableChange> tableInitChange) {
+    public CompletableFuture<Table> createTableAsync(CreateTableParams parameters) {
         if (!busyLock.enterBusy()) {
             throw new IgniteException(new NodeStoppingException());
         }
         try {
-            return createTableAsyncInternal(name, zoneName, tableInitChange);
+            String tableName = parameters.tableName();
+            String zoneName = Objects.requireNonNullElse(parameters.zone(), DEFAULT_ZONE_NAME);
```
tkalkirill, line 1487: I think the check is not needed and the default zone should be set on the buid of the `CreateTableParams`.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1462,32 +1472,53 @@ private Set<Assignment> calculateAssignments(TablePartitionId tablePartitionId)
...
+
+            // Copied from DdlCommandHandler
+            Consumer<TableChange> tblChanger = tableChange -> {
```
tkalkirill, line 1490: Maybe we should get rid of the configuration?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2758,4 +2826,78 @@ private static CatalogDataStorageDescriptor toDataStorageDescriptor(DataStorageV
...
+    @Deprecated(forRemoval = true)
+    private CompletableFuture<Void> addColumnInternal(String fullName, List<ColumnParams> columnParams) {
+        return alterTableAsyncInternal(
+                fullName,
+                chng -> {
+                    if (columnParams.isEmpty()) {
+                        return false;
+                    }
+
+                    chng.changeColumns(cols -> {
```
tkalkirill, line 2840: Shouldn't we get rid of the configuration?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2758,4 +2826,78 @@ private static CatalogDataStorageDescriptor toDataStorageDescriptor(DataStorageV
...
+    // Copied from DdlCommandHandler
+    // TODO: IGNITE-19082 Drop unused temporary method.
+    @Deprecated(forRemoval = true)
+    private CompletableFuture<Void> dropColumnInternal(String tableName, Set<String> colNames) {
+        AtomicBoolean ret = new AtomicBoolean(true);
+
+        return alterTableAsyncInternal(
+                tableName,
+                chng -> {
+                    chng.changeColumns(cols -> {
```
tkalkirill, line 2869: Shouldn't we get rid of the configuration?

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -150,18 +151,18 @@ public CatalogServiceImpl(UpdateLog updateLog, ClockWaiter clockWaiter, LongSupp
 
     @Override
     public void start() {
-        int objectIdGen = 0;
+        int objectIdGen = 1;
 
         // TODO: IGNITE-19082 Move default schema objects initialization to cluster init procedure.
         CatalogSchemaDescriptor schemaPublic = new CatalogSchemaDescriptor(
-                objectIdGen++,
+                0,
```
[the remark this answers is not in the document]
AMashenkov: Why not?

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -150,18 +151,18 @@ public CatalogServiceImpl(UpdateLog updateLog, ClockWaiter clockWaiter, LongSupp
...
                 DEFAULT_SCHEMA_NAME,
                 new CatalogTableDescriptor[0],
                 new CatalogIndexDescriptor[0]
         );
 
         CatalogZoneDescriptor defaultZone = new CatalogZoneDescriptor(
-                objectIdGen++,
+                0,
```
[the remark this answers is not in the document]
AMashenkov: Because DistributedZoneManager and test expects zero as if for default zone.

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -389,6 +401,13 @@ public CompletableFuture<Void> alterColumn(AlterColumnParams params) {
 
     @Override
     public CompletableFuture<Void> createIndex(CreateHashIndexParams params) {
+        if (StringUtils.nullOrEmpty(params.indexName())) {
```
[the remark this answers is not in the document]
AMashenkov: Removed. This validation can be done in builder.

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/BaseDistributionZoneManagerTest.java
```diff
@@ -126,6 +126,7 @@ void setUp() {
         distributionZoneManager = new DistributionZoneManager(
                 zonesConfiguration,
                 tablesConfiguration,
+                DistributionZonesTestUtil.mockCatalog(),
```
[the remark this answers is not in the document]
AMashenkov: I think we shouldn't mock catalog the way we do.
Instead I'd create a test implementation, which allow us to check if component subscribes to some events,
and allow us to trigger any event directly (without any mocking internal stuff like metastorage).

modules/runner/src/integrationTest/java/org/apache/ignite/internal/index/ItIndexManagerTest.java
```diff
@@ -69,11 +71,12 @@ public void eventsAreFiredWhenIndexesCreatedAndDropped() {
         CompletableFuture<IndexEventParameters> indexCreatedFuture = registerListener(indexManager, IndexEvent.CREATE);
 
         await(indexManager.createIndexAsync(
-                "PUBLIC",
-                "INAME",
-                "TNAME",
-                true,
-                tableIndexChange -> tableIndexChange.convert(HashIndexChange.class).changeColumnNames("C3", "C2")
+                CreateHashIndexParams.builder()
+                        .schemaName("PUBLIC")
```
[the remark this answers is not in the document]
AMashenkov: Agree, tests should be rework regarding the new reality.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
AMashenkov: What a thread pool we use to complete user futures?
Why we should execute user stuff (other user futures, which may be chained to the current one) in a system thread pool?

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogService.java
```diff
@@ -56,8 +58,11 @@ public interface CatalogService {
 
     CatalogZoneDescriptor zone(String zoneName, long timestamp);
 
+    @Deprecated
     CatalogZoneDescriptor zone(int zoneId, long timestamp);
 
+    CatalogZoneDescriptor zone(int zoneId, int version);
```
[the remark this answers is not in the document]
AMashenkov: Maybe we should rethink interface methods. 
Accessing catalog methods with object name and timestamp, seems, never happens.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
AMashenkov: Then direct proxy will not work and we have no other way to wait for actual cluster state on lagging node.

PR #2295, opened 2023-07-06
IGNITE-19810 Remove excessive BinaryRow buffer copying

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-19810


modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/io/RowVersionDataIo.java
```diff
@@ -50,46 +52,102 @@ protected RowVersionDataIo(int ver) {
...
+        putInt(addr, 0, rowVersion.valueSize());
         addr += Integer.BYTES;
 
-        putByteBuffer(addr, 0, row.value());
+        BinaryRow row = rowVersion.value();
+
+        if (row != null) {
+            // Write schema version in LE order.
+            putByte(addr, 0, (byte) row.schemaVersion());
+            putByte(addr, 1, (byte) (row.schemaVersion() >> Byte.SIZE));
```
ibessonov, line 75: What's wrong with `putShort`? We use native ordering here, that's fine

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/io/RowVersionDataIo.java
```diff
@@ -50,46 +52,102 @@ protected RowVersionDataIo(int ver) {
...
+            putByte(addr, 2, (byte) (row.hasValue() ? 1 : 0));
```
ibessonov, line 76: We should never write this value, it's for clients only

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/io/RowVersionDataIo.java
```diff
@@ -50,46 +52,102 @@ protected RowVersionDataIo(int ver) {
...
-    protected void writeFragmentData(RowVersion row, ByteBuffer pageBuf, int rowOff, int payloadSize) {
+    protected void writeFragmentData(RowVersion rowVersion, ByteBuffer pageBuf, int rowOff, int payloadSize) {
         assertPageType(pageBuf);
 
         if (rowOff == 0) {
             // first fragment
-            assert row.headerSize() <= payloadSize : "Header must entirely fit in the first fragment, but header size is "
-                    + row.headerSize() + " and payload size is " + payloadSize;
+            assert rowVersion.headerSize() <= payloadSize : "Header must entirely fit in the first fragment, but header size is "
+                    + rowVersion.headerSize() + " and payload size is " + payloadSize;
```
ibessonov, line 88: What is the value for header size? Maybe it must be updated

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/io/RowVersionDataIo.java
```diff
@@ -50,46 +52,102 @@ protected RowVersionDataIo(int ver) {
...
+        BinaryRow row = rowVersion.value();
+
+        if (payloadSize == 0 || row == null) {
+            return;
+        }
+
+        writeFragmentedBinaryRow(row, pageBuf, rowOff, payloadSize);
+    }
+
+    private void writeFragmentedBinaryRow(BinaryRow row, ByteBuffer pageBuf, int offset, int payloadSize) {
```
ibessonov, line 113: I think you made it overly complicated. You could write "version" right before `pageBuf.putInt(rowVersion.valueSize());`, for example, and then all you had to do is just leave the old code, basically (considering that you'll make sure to store version before size in other places)

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -114,20 +105,22 @@
...
- * Please refer to {@link PartitionDataHelper#putTimestampDesc(ByteBuffer, HybridTimestamp)}
- * to see how it's achieved. Missing timestamp could be interpreted as a moment infinitely far away in the future.
+ * Please refer to {@link PartitionDataHelper#putTimestampDesc(ByteBuffer, HybridTimestamp)} to see how it's achieved. Missing timestamp
+ * could be interpreted as a moment infinitely far away in the future.
  */
 public class RocksDbMvPartitionStorage implements MvPartitionStorage {
     /** Thread-local on-heap byte buffer instance to use for key manipulations. */
     private static final ThreadLocal<ByteBuffer> HEAP_KEY_BUFFER = withInitial(() -> allocate(MAX_KEY_SIZE).order(KEY_BYTE_ORDER));
 
+    private static final ThreadLocal<ByteBuffer> DIRECT_KEY_BUFFER = withInitial(() -> allocateDirect(MAX_KEY_SIZE).order(KEY_BYTE_ORDER));
```
ibessonov, line 115: There is a `org.apache.ignite.internal.storage.rocksdb.PartitionDataHelper#MV_KEY_BUFFER`, can it be reused?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -374,40 +366,47 @@ private void saveGroupConfiguration(AbstractWriteBatch writeBatch, byte[] config
...
+                if (previousValueBytes != null) {
+                    ByteBuffer previousValue = ByteBuffer.wrap(previousValueBytes);
+
                     validateTxId(previousValue, txId);
 
                     res = wrapValueIntoBinaryRow(previousValue, true);
                 }
 
                 if (row == null) {
+                    ByteBuffer value = allocateDirect(VALUE_HEADER_SIZE);
```
ibessonov, line 395: I would avoid frequent usage of allocateDirect, it will create unnecessary GC pressure, as far as I know. That's why in the old code all direct buffers are cached as thread-locals.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -420,33 +419,46 @@ private void saveGroupConfiguration(AbstractWriteBatch writeBatch, byte[] config
...
      * @param txId Transaction id.
      * @throws RocksDBException If write failed.
      */
-    private void writeUnversioned(byte[] keyArray, BinaryRow row, UUID txId, int commitTableId, int commitPartitionId)
+    private void writeUnversioned(ByteBuffer directKeyBuffer, BinaryRow row, UUID txId, int commitTableId, int commitPartitionId)
             throws RocksDBException {
         @SuppressWarnings("resource") WriteBatchWithIndex writeBatch = PartitionDataHelper.requireWriteBatch();
 
-        byte[] rowBytes = rowBytes(row);
+        ByteBuffer value = allocateDirect(row.length() + VALUE_HEADER_SIZE);
```
ibessonov, line 431: Same here. We could use `allocate` + `wrapPointer`, and then free it manually, That would be better I think.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -420,33 +419,46 @@ private void saveGroupConfiguration(AbstractWriteBatch writeBatch, byte[] config
...
-    private static byte[] rowBytes(BinaryRow row) {
-        //TODO IGNITE-16913 Add proper way to write row bytes into array without allocations.
-        return row.bytes();
+    private static ByteBuffer writeBinaryRow(ByteBuffer dest, BinaryRow row) {
+        assert dest.order() == ByteOrder.BIG_ENDIAN;
+
+        return dest
+                .order(ByteOrder.LITTLE_ENDIAN)
+                .putShort((short) row.schemaVersion())
+                .put(row.hasValue() ? (byte) 1 : (byte) 0)
```
ibessonov, line 459: Again, I'm not convinced that this data belongs to the storage.

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -559,8 +575,21 @@ public void addWriteCommitted(RowId rowId, @Nullable BinaryRow row, HybridTimest
             // We only write tombstone if the previous value for the same row id was not a tombstone.
             // So there won't be consecutive tombstones for the same row id.
             if (!newAndPrevTombstones) {
+                //TODO IGNITE-16913 Add proper way to write row bytes into array without allocations.
+                ByteBuffer rowBytes;
+
+                if (row == null) {
+                    rowBytes = allocateDirect(0);
```
ibessonov, line 582: Hm, this just looks wrong, allocating a buffer with size 0. Maybe it could be a constant

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
```diff
@@ -657,21 +684,25 @@ private ReadResult readResultFromKeyAndValue(boolean isWriteIntent, ByteBuffer k
...
+
+        keyBuf.rewind();
 
         return handleReadByTimestampIterator(seekIterator, rowId, timestamp, keyBuf);
     }
 
     /**
-     * Walks "version chain" via the iterator to find a row by timestamp.
-     * See {@link MvPartitionStorage#read(RowId, HybridTimestamp)} for details.
+     * Walks "version chain" via the iterator to find a row by timestamp. See {@link MvPartitionStorage#read(RowId, HybridTimestamp)} for
```
ibessonov, line 704: Why would you re-format perfectly fine-looking comment, that specifically showed different sentences on different lines? Please don't do that, and return the old formatting back.

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryRowImpl.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    public boolean hasValue() {
+        return hasValue;
+    }
+
+    @Override
+    public ByteBuffer tupleSlice() {
+        return binaryTuple.slice().order(ORDER);
```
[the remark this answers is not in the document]
sashapolo: I agree, `duplicate` is better here

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
sashapolo: Unfortunately, no, for two reasons: `MV_KEY_BUFFER` is used by the Garbage Collector, which can mess up our keys. There's also a situation, when we need *two* buffers at once, so we use both this one and `MV_KEY_BUFFER` (it's confusing I know, but I don't know how to simplify that)

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
sashapolo: > it will create unnecessary GC pressure

Can you please clarify? What's the difference between using a direct buffer or copying a byte array inside JNI?

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
sashapolo: Because that's what IDEA does with its autoformatting. If you want to have two sentences on separate lines, you should use `<p>` tags

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/io/RowVersionDataIo.java
sashapolo: That's not fine, because we wrap bytes that we read from the storage into a `ByteBufferRow` which is little endian. However, I think that we should remove `ByteBufferRow` altogether and parse it manually into a `BinaryRowImpl`, so I'll rework this logic and will bring `putShort` back

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/mv/io/RowVersionDataIo.java
sashapolo: What do you mean? I didn't change the header...

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
sashapolo: ok, I did some research, looks like this is indeed a bad idea. I'll rollback this change

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryRow.java
```diff
@@ -44,12 +47,11 @@ public interface BinaryRow {
...
+    int length();
+
     /** Get ByteBuffer slice representing the binary tuple. */
     ByteBuffer tupleSlice();
 
-    /** Get byte array of the row. */
-    byte[] bytes();
-
     /** Returns the representation of this row as a Byte Buffer. */
     ByteBuffer byteBuffer();
```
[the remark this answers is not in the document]
ibessonov: > It is also used to serialize rows as parts of Raft commands.

This part we can remove. It would be better to convert these objects into a BinaryRowMessage. Less memory manipulations, better performance. Ideally, there should zero memory copying between layers. It's not achievable, bet we should try to be as close to is as possible.

modules/schema/src/main/java/org/apache/ignite/internal/schema/ByteBufferRow.java
```diff
@@ -18,14 +18,11 @@
 package org.apache.ignite.internal.schema;
 
 import java.nio.ByteBuffer;
-import java.nio.ByteOrder;
 
 /**
  * Heap byte buffer-based row.
  */
 public class ByteBufferRow implements BinaryRow {
```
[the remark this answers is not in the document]
ibessonov: Can we get rid of it in a separate issue? Seems obsolete to me

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
ibessonov: Ok, fair

modules/storage-rocksdb/src/main/java/org/apache/ignite/internal/storage/rocksdb/RocksDbMvPartitionStorage.java
ibessonov: As far as I know, most people read javadocs in code, without HTML rendering. I meant that.

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryRow.java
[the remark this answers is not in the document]
sashapolo: Do you want me to do that in this PR? That's quite a big change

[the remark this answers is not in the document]
ibessonov: No, we should do it in a separate issue

ibessonov (approved):

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryRow.java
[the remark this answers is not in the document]
sashapolo: I've created a ticket and added TODOs: https://issues.apache.org/jira/browse/IGNITE-19937

modules/schema/src/main/java/org/apache/ignite/internal/schema/ByteBufferRow.java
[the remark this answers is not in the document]
sashapolo: I've created a ticket and added TODOs: https://issues.apache.org/jira/browse/IGNITE-19937

PR #2296, opened 2023-07-06
IGNITE-19924 Test shouldn't guess id of newly created catalog object

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-19924

### DIFF
--- a/modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
+++ b/modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
@@ -148,18 +148,18 @@ public class CatalogServiceImpl extends Producer<CatalogEvent, CatalogEventParam
 
     @Override
     public void start() {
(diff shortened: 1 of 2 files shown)


modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogServiceImpl.java
```diff
@@ -148,18 +148,18 @@ public CatalogServiceImpl(UpdateLog updateLog, ClockWaiter clockWaiter, LongSupp
...
-                objectIdGen++,
+                0,
                 DEFAULT_SCHEMA_NAME,
                 new CatalogTableDescriptor[0],
                 new CatalogIndexDescriptor[0]
         );
 
         CatalogZoneDescriptor defaultZone = new CatalogZoneDescriptor(
-                objectIdGen++,
+                0,
```
korlov42, line 162: why schema and zone share the same id?

korlov42 (approved):

PR #2300, opened 2023-07-07
IGNITE-19928 Change error and group codes from int to short

sk0x50:
The error group and the error code in the group are defined by two bytes each (see [IEP-84 Error Handling](https://cwiki.apache.org/confluence/display/IGNITE/IEP-84%3A+Error+handling)). Fix the code to use `short` instead of `int`.

https://issues.apache.org/jira/browse/IGNITE-19928


modules/core/src/main/java/org/apache/ignite/lang/ErrorGroup.java
```diff
@@ -86,14 +87,9 @@ public int code() {
...
-     * @throws IllegalArgumentException If the given {@code errorCode} is already registered
-     *      or {@code errorCode} is greater than 0xFFFF or less than or equal to 0.
+     * @throws IllegalArgumentException If the given {@code errorCode} is already registered.
      */
-    public int registerErrorCode(int errorCode) {
-        if (errorCode <= 0 || errorCode > 0xFFFF) {
-            throw new IllegalArgumentException("Error code should be greater than 0 and less than or equal to 0xFFFF");
-        }
-
+    public int registerErrorCode(short errorCode) {
```
sanpwc, line 92: Minor. Seems that there's no consistency in errorCode naming. Here it's errorCode, but in `public boolean isRegistered(ErrorGroup group, short code)` it's just code.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroup.java
```diff
@@ -106,10 +102,10 @@ public int registerErrorCode(int errorCode) {
     /**
      * Checks that the given {@code code} is registered for this error group.
      *
-     * @param code Full error code to be tested.
```
sanpwc, line 109: So, initial javadoc was incorrect, and it wasn't "Full error code", but "group specific code", right?

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroup.java
```diff
@@ -177,10 +172,20 @@ public static int extractErrorCode(int code) {
...
         return registeredGroups.get(groupCode);
     }
 
+    /**
+     * Returns error group identified by the given error {@code code}.
+     *
+     * @param code Full error code
+     * @return Error Group.
+     */
+    public static ErrorGroup errorGroupByErrorCode(int code) {
```
sanpwc, line 185: Same as above. ErrorCode seems to be code within group, e.g. according to
`public int registerErrorCode(short errorCode)`. Maybe in addition to consistency with naming we even need better naming itself. public int registerErrorCode(short errorCode) {

sanpwc (changes requested):

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroup.java
sk0x50: You are right, the parameter names should be the same.
I decided to remove the `isRegistered` method just because it is not used anywhere.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroup.java
sk0x50: Yes, the initial javadoc was incorrect. Moreover, there is no test for this method... (my fault).
In any way, I removed this method because there are no tests/prod code that use it.

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroup.java
sk0x50: changed to `erroGroupByCode`

sk0x50: Unfortunately, I don't have any ideas how it can be improved right now. I will think about it in the background.

sanpwc (approved):

PR #2301, opened 2023-07-07
IGNITE-19671 Save default configuration values to configuration storage

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-19671

### DIFF
--- a/modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationChanger.java
+++ b/modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationChanger.java
@@ -94,6 +94,7 @@ public abstract class ConfigurationChanger implements DynamicConfigurationChange
     /** Storage trees. */
     private volatile StorageRoots storageRoots;
 
+    /** Future that resolves after the defaults are persisted to the storage. */
     private final CompletableFuture<Void> defaultsPersisted = new CompletableFuture<>();
 
     /** Configuration listener notification counter, must be incremented before each use of {@link #configurationUpdateListener}. */
@@ -260,12 +261,16 @@ public abstract class ConfigurationChanger implements DynamicConfigurationChange
 
         SuperRoot superRootNoDefaults = superRoot.copy();
 
-        // Workaround for distributed configuration.
         addDefaults(superRoot);
 
         // Validate the restored configuration.
         validateConfiguration(superRoot);
-
+        // We store two configuration roots, one with the defaults set and another one without them.
+        // The root WITH the defaults is used when we calculate who to notify of a configuration change or
+        // when we provide the configuration outside.
+        // The root WITHOUT the defaults is used to calculate which properties to write to the underlying storage,
+        // in other words it allows us to persist the defaults from the code.
+        // After the storage listener fires for the first time both roots are supposed to become equal.
         storageRoots = new StorageRoots(superRootNoDefaults, superRoot, data.changeId());
 
         storage.registerConfigurationListener(configurationStorageListener());
@@ -279,11 +284,8 @@ public abstract class ConfigurationChanger implements DynamicConfigurationChange
      * <p>Specifically, this method runs a check whether there are
      * default values that are not persisted to the storage and writes them if there are any.
      */
-    private CompletableFuture<Void> persistDefaults() {
-        if (storageRoots == null) {
-            throw new ComponentNotStartedException();
-        }
-        return changeInternally(ConfigurationUtil.EMPTY_CFG_SRC, true)
+    private void persistDefaults() {
+        changeInternally(ConfigurationUtil.EMPTY_CFG_SRC, true)
                 .whenComplete((v, e) -> {
                     if (e == null) {
                         defaultsPersisted.complete(null);
@@ -514,6 +516,7 @@ public abstract class ConfigurationChanger implements DynamicConfigurationChange
      * Entry point for configuration changes.
      *
      * @param src Configuration source.
+     * @param onStartup if {@code true} this change is triggered right after startup
      * @return Future that will be completed after changes are written to the storage.
      * @throws ComponentNotStartedException if changer is not started.
      */
@@ -540,6 +543,7 @@ public abstract class ConfigurationChanger implements DynamicConfigurationChange
      *
      * @param src Configuration source.
      * @param storageRevision Latest storage revision from the metastorage.
+     * @param onStartup if {@code true} this change is triggered right after startup
      * @return Future that will be completed after changes are written to the storage.
      */
     private CompletableFuture<Void> changeInternally0(ConfigurationSource src, long storageRevision, boolean onStartup) {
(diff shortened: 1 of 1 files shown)


modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationChanger.java
```diff
@@ -263,40 +271,63 @@ public void start() {
...
+                    }
+                });
+    }
 
-            if (cause instanceof ConfigurationValidationException) {
-                throw (ConfigurationValidationException) cause;
+    /**
+     * Internal configuration persist method.
+     */
+    private CompletableFuture<Void> persistDefaultsInternal(long storageRevision) {
```
ibessonov, line 306: I have doubts in the approach you've chosen. Remember that `//Workaround for distributed configuration.` comment in the code? It's there for a reason. And the reason is - it's a workaround, that must be removed. I believe it must be removed in this issue. Sorry for not having proper TODO with the description in that comment, that's on me. Let me explain my position.

Now, this is going to be a pretty long comment, so bear with me.
1. Can this CAS fail? I see that literally every node tries to execute the operation.So it seems like it can. On the other hand, incoming updates should make the map empty. Can we have a concurrent update of the configuration, that will come from a user? Maybe. You probably meant the upgrade from older to a newer version of Ignite.
Are there plans for `init` operation upon upgrade, or we will just quietly add default, invisible to the user? This is a complicated question.
2. I think there might be a simpler way to do things, it's removing that code with a "workaround" comment. This way, all changers will always explicitly write data, that was previously `null`, to the storage (with some tweaks). There are advantages to such approach:
 - Upon init, we would only need a single configuration update, instead of two.
 - We will reuse already existing code, making the procedure much simpler.
One thing to consider - we may still need to call similar method on each start, but having the update with empty map, that will have a specific property of not incrementing the storage revision if there are no updates, would solve the issue, with much less complexity in code.
The only possible issue here, that I see, is that some validators would get `null` from "old" values of configuration, and that must be handled properly. This is also a little inconvenient, because primitive values would throw `NullPointerException` instead of returning nulls, because you know, primitives can't be nulls. This issue can be eliminated, if we will have both "tree with defaults" and "tree without defaults" in the `storageRoots`.
Otherwise, it seems like a better approach to me.
3. Maybe this is not relevant, but this is, most likely, a misuse of the read lock. It has a very specific purpose of protecting a single field. I guess it's not documented properly. But the idea is - it protected the state of the `storageRoots` field while executing `src.descend(changes);`. Your code doesn't use any external closures, and the only access you have to `storageRoots` is a single volatile read, which would still be correct without the read lock.
Considering that I propose deleting the method entirely, this comment was purely informational.

Overall, I would like to discuss my alternative approach with you. Let's do it, please also ask clarifying questions about what I've said, it may be a little confusing, I know that.

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationRegistry.java
```diff
@@ -137,16 +131,10 @@ public void stop() throws Exception {
...
+     * Persist default values to the storage.
      */
-    public void initializeDefaults() {
-        changer.initializeDefaults();
-
-        for (RootKey<?, ?> rootKey : rootKeys) {
-            DynamicConfiguration<?, ?> dynCfg = configs.get(rootKey.key());
-
-            touch(dynCfg);
-        }
```
ibessonov, line 149: Why have you removed this code? I believe it may be necessary for notifications to work properly, but I have no tests to support the statement. Did you fully understand the intention of the code before removing it?

modules/configuration/src/test/java/org/apache/ignite/internal/configuration/ConfigurationChangerTest.java
```diff
@@ -131,6 +131,7 @@ public static void afterAll() {
     public void testSimpleConfigurationChange() throws Exception {
         ConfigurationChanger changer = createChanger(KEY);
         changer.start();
+        changer.persistDefaults().get(1, SECONDS);
```
ibessonov, line 134: Please use `assertThat(..., willCompleteSuccessfully())`, it looks nicer.

modules/configuration/src/test/java/org/apache/ignite/internal/configuration/ConfigurationChangerTest.java
```diff
@@ -275,7 +282,10 @@ public void testFailedToWrite() {
 
         CompletableFuture<Map<String, ? extends Serializable>> dataFuture = storage.readDataOnRecovery().thenApply(Data::values);
 
-        assertThat(dataFuture, willBe(anEmptyMap()));
+        //these are the default values taken from FirstConfigurationSchema
```
ibessonov, line 285:
```suggestion
        // These are the default values taken from FirstConfigurationSchema.
```

modules/configuration/src/test/java/org/apache/ignite/internal/configuration/tree/InternalIdTest.java
```diff
@@ -98,8 +98,7 @@ void setUp() {
         );
 
         registry.start();
-
-        registry.initializeDefaults();
+        registry.persistDefaults().get(1, TimeUnit.SECONDS);
```
ibessonov, line 101: Why don't we move this call inside of the `start()`? It may return the future as well. I see that they are always called together, that's a good argument for combining them.

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerTest.java
```diff
@@ -92,8 +92,6 @@ public void setUp() {
 
         registry.start();
 
-        registry.initializeDefaults();
```
ibessonov, line 95: Right after I said _"they are always called together"_, I see that they're not. Is this intentional, or you simply forgot to call the method?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/ddl/DdlCommandHandlerExceptionHandlingTest.java
```diff
@@ -100,8 +100,6 @@ public class DdlCommandHandlerExceptionHandlingTest extends IgniteAbstractTest {
     void before() {
         registry.start();
 
-        registry.initializeDefaults();
```
ibessonov, line 103: Same question here

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationRegistry.java
ibessonov: It was used in production before. This is interesting, I guess our tests are insufficient, this sucks. I clearly remember, that there was a reason for this code. Let's discuss it right after current issue and maybe try writing a test that would fail without this code.

ibessonov (commented):
Thank you! I think I'll approve the changes when all minor are resolved

ibessonov (approved):

PR #2302, opened 2023-07-10
IGNITE-19935 Allow combining individual colocation hashes

ptupitsyn:
Refactor hashing to change combined hash formula from 
```
hash(col2, seed: hash(col1))
```

to

```
hashCombine(hash(col1), hash(col2))
```

* This allows us to compute individual hash codes in any order, then combine them in correct order (according to colocation key)
* Also this means we can combine hashing and serialization and avoid duplicate value extraction (especially for strings, which require array allocation to get the bytes)
* Collisions: 125 per 1M entries (`Row(0..99, 0..99, 0..99)`), up from 118 reported in [IGNITE-14769 comments](https://issues.apache.org/jira/browse/IGNITE-14769)
* Distribution: 
image


modules/core/src/main/java/org/apache/ignite/internal/util/HashCalculator.java
```diff
@@ -170,8 +244,19 @@ public void appendNumber(BigInteger v) {
...
+    /**
+     * Get value hash.
+     *
+     * @param v Value to hash.
+     */
+    public static int hashUuid(UUID v) {
+        int hash1 = hashLong(v.getMostSignificantBits());
+        int hash2 = hashLong(v.getLeastSignificantBits());
+
+        return combine(hash1, hash2);
```
isapego, line 259: Do we need to use `combine` here? I'd avoid using it where it is possible.
```suggestion
        int hash = HashUtils.hash32(v.getMostSignificantBits());
        hash = HashUtils.hash32(v.getLeastSignificantBits(), hash);

        return hash;
```

modules/core/src/main/java/org/apache/ignite/internal/util/HashUtils.java
```diff
@@ -120,12 +105,50 @@ public static int hash32(int data, int seed) {
...
+     * Combines two hash values.
+     *
+     * @param hash1 The first hash.
+     * @param hash2 The second hash.
+     * @return The combined 32-bit hash.
+     */
+    public static int combine(int hash1, int hash2) {
+        long hash = hash64(hash1, hash2);
+
+        return (int) (hash ^ (hash >>> 32));
```
isapego, line 124: Are there any reasons why it is implemented this way?
Also, does `combine(a,b) == combine(b,a)`?
More descriptive javadoc probably is needed here.

modules/core/src/main/java/org/apache/ignite/internal/util/HashCalculator.java
ptupitsyn: Good point, fixed.

modules/core/src/main/java/org/apache/ignite/internal/util/HashUtils.java
ptupitsyn: Refactored, added comments.

isapego (approved):

korlov42 (approved):

PR #2303, opened 2023-07-11
IGNITE-19205 ODBC data buffers

isapego:
- Implemented data buffers;
- Added tests;
- Unmuted related old tests.

### DIFF
--- a/modules/platforms/cpp/ignite/odbc/app/application_data_buffer.cpp
+++ b/modules/platforms/cpp/ignite/odbc/app/application_data_buffer.cpp
@@ -630,8 +630,12 @@ conversion_result application_data_buffer::put_decimal(const big_decimal& value)
         {
             auto* numeric = reinterpret_cast<SQL_NUMERIC_STRUCT*>(get_data());
 
+            auto sign = value.is_negative() ? 0 : 1;
             big_decimal zero_scaled;
             value.set_scale(0, zero_scaled);
+
+            if (zero_scaled.is_negative())
+                zero_scaled.negate();
 
             const big_integer& unscaled = zero_scaled.get_unscaled_value();
             std::vector<std::byte> bytes_buffer = unscaled.to_bytes();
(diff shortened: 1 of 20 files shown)


modules/platforms/cpp/ignite/odbc/app/application_data_buffer.cpp
```diff
@@ -892,22 +841,27 @@ conversion_result application_data_buffer::put_time(const ignite_time& value)
         case odbc_native_type::AI_TTIMESTAMP:
         {
             auto* buffer = reinterpret_cast<SQL_TIMESTAMP_STRUCT*>(data_ptr);
+            memset(buffer, 0, sizeof(*buffer));
```
ptupitsyn, line 844: Is this cleanup necessary? We seemingly override the entire buffer below.

modules/platforms/cpp/ignite/tuple/tuple_test.cpp
```diff
@@ -26,6 +26,7 @@
 #include <tuple>
 #include <type_traits>
 #include <vector>
+#include <optional>
```
ptupitsyn, line 29: The rest of the file has no changes, is this needed?

ptupitsyn (approved):

modules/platforms/cpp/ignite/odbc/app/application_data_buffer.cpp
isapego: Such a cleanup considered to be a good practice in C world, because you can not be sure that for example on other platform or in the next version of standard they will not add some new fields.

modules/platforms/cpp/ignite/tuple/tuple_test.cpp
isapego: Yeah. It seems that I removed `<optional>` from some other header that was included in this one and everything stopped compiling. It's better to explicitly include all used standard headers in the file anyway.

PR #2305, opened 2023-07-11
IGNITE-19941 Sql. Test ExecutionServiceImplTest.testCancelOnInitiator is flaky

korlov42:


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImplTest.java
```diff
@@ -170,6 +172,14 @@ public void tearDown() throws Exception {
         }
 
         mailboxes.clear();
+
+        executers.forEach(executer -> {
```
ygerzhedovich, line 176: should we cleanup the list of executors after each test?

ygerzhedovich, line 176: should we clean up the list of executors after each of tests?

korlov42: such a stupid bug... yes, we definitely should clean up this collection, thanks!

fixed

PR #2310, opened 2023-07-12
IGNITE-19963 Move completion of ClockWaiter futures to a special thread pool

rpuch:
https://issues.apache.org/jira/browse/IGNITE-19963

### DIFF
--- a/modules/catalog/src/main/java/org/apache/ignite/internal/catalog/ClockWaiter.java
+++ b/modules/catalog/src/main/java/org/apache/ignite/internal/catalog/ClockWaiter.java
@@ -169,7 +169,9 @@ public class ClockWaiter implements IgniteComponent {
                 .handle((res, ex) -> {
                     scheduledFuture.cancel(true);
 
-                    translateTrackerClosedException(ex);
+                    if (ex != null) {
+                        translateTrackerClosedException(ex);
+                    }
 
                     return res;
                 })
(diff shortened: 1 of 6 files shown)


SammyVimes (changes requested):

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/ClockWaiter.java
```diff
@@ -125,47 +152,45 @@ public CompletableFuture<Void> waitFor(HybridTimestamp targetTimestamp) {
...
-                long millisToWait = targetTimestamp.getPhysical() - now.getPhysical() + 1;
+        ScheduledFuture<?> scheduledFuture = scheduler.schedule(triggerClockUpdate, millisToWait, TimeUnit.MILLISECONDS);
 
-                scheduledFuture = scheduler.schedule(this::triggerClockUpdate, millisToWait, TimeUnit.MILLISECONDS);
-            }
-        } else {
-            scheduledFuture = null;
-        }
+        return future
+                .handle((res, ex) -> {
```
SammyVimes, line 169: Should it be handleAsync then, if future can complete on any thread?

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/ClockWaiter.java
```diff
@@ -125,47 +152,45 @@ public CompletableFuture<Void> waitFor(HybridTimestamp targetTimestamp) {
...
+                    scheduledFuture.cancel(true);
 
-        return future.handle((res, ex) -> {
-            if (scheduledFuture != null) {
-                scheduledFuture.cancel(true);
-            }
+                    translateTrackerClosedException(ex);
```
SammyVimes, line 172: This looks like we are handling the error every time, but it might not exist. Maybe pull `if (ex != null)` out of the method?

modules/core/src/main/java/org/apache/ignite/internal/hlc/HybridClockImpl.java
```diff
@@ -71,15 +75,23 @@ public long nowLong() {
...
 
     private void notifyUpdateListeners(long newTs) {
-        updateListeners.forEach(listener -> listener.onUpdate(newTs));
+        for (ClockUpdateListener listener : updateListeners) {
+            try {
+                listener.onUpdate(newTs);
+            } catch (Throwable e) {
+                LOG.error("ClockUpdateListener#onUpdate() failed for {} at {}", e, listener, newTs);
+
+                if (e instanceof Error && !(e instanceof AssertionError)) {
```
SammyVimes, line 90: why `!(e instanceof AssertionError)`?

SammyVimes (changes requested):

SammyVimes (approved):
LGTM

PR #2312, opened 2023-07-12
[IGNITE-19785] Sql. Rework execution of 2-phase aggregates

lowka:
Interaction between map/reduce phases of two phase aggregates no longer uses complex objects. Prior to that it was group_key, list of accumulators, group id (optional for SortAggregates) .

Removes `apply` method from accumulators. Prior to this commit on map phase accumulator were using `add` method and `apply` method on reduce. Now it is only one method `apply` and aggregates are supposed to be rewritten for reduce phase. (1)

Adds classes to transform aggregate its into map/reduce form (see `MapReduceAggregates`) for simple cases such as COUNT (COUNT on MAP and SUM of COUNTs on REDUCE), and SUM, MIN, MAX, etc.

`AVG` aggregate is now only works for collocated aggregates. Because it should be rewritten as `SUM(x) as avg_x_sum`, `COUNT(x) as avg_count_x` on map and `avg_x_sum/avg_count_x` on reduce phase as a separate ticket - https://issues.apache.org/jira/browse/IGNITE-20009.

Adds tests some cases for `GROUP BY GROUPING SET`  because `HashAggregateNode` is implemented by using grouping sets.

https://issues.apache.org/jira/browse/IGNITE-19785


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/HashAggregateNode.java
```diff
@@ -255,126 +254,74 @@ private void reset() {
...
-            }
-
-            return res;
-        }
-
-        private List<RowT> getOnReducer(int cnt) {
-            Iterator<Map.Entry<GroupKey, List<AccumulatorWrapper<RowT>>>> it = groups.entrySet().iterator();
+        /**
+         * Get rows.
+         * TODO Documentation https://issues.apache.org/jira/browse/IGNITE-15859
```
korlov42, line 278: let's improve this javadoc

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/BaseQueryContext.java
```diff
@@ -132,7 +133,7 @@ public List<MetadataHandler<?>> handlers(Class<? extends MetadataHandler<?>> hnd
         );
 
         cluster.setMetadataQuerySupplier(() -> {
-            throw new AssertionError(cantBeUsedMsg);
+            return RelMetadataQueryEx.create();
```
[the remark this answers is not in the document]
lowka: @korlov42 Thanks. Should have been removed. This is code is used to debug tests in case of serialisation mismatches (because when it happens you get this error instead of a mismatch).

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/agg/AggregateRow.java
```diff
@@ -0,0 +1,219 @@
...
+public final class AggregateRow<RowT> {
+    /** A placeholder of absent group id. */
+    public static final byte NO_GROUP_ID = -1;
+
+    private final List<AccumulatorWrapper<RowT>> accs;
+
+    private final AggregateType type;
+
+    /** Constructor. */
+    public AggregateRow(List<AccumulatorWrapper<RowT>> accs, IgniteTypeFactory typeFactory,
```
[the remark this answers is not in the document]
lowka: Thanks. Some leftovers from feature flags.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/sql/fun/IgniteSqlOperatorTable.java
```diff
@@ -163,6 +165,20 @@ public boolean isDeterministic() {
                 }
             };
 
+    /** Reduce phase of COUNT aggregate which sums all counts. */
+    public static final SqlAggFunction REDUCE_COUNT =
```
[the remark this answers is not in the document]
lowka: @korlov42 OK. I will replace it with projection because vanilla SUM won't do.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/HashAggregateNode.java
lowka: Done.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/agg/AggregateRow.java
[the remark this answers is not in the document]
lowka: Done.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/agg/GroupKey.java
```diff
@@ -91,7 +92,7 @@ private Builder(int rowLen) {
          * Add.
          * TODO Documentation https://issues.apache.org/jira/browse/IGNITE-15859
          */
-        public Builder add(Object val) {
```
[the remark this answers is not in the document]
lowka: Done.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/agg/IgniteMapSortAggregate.java
```diff
@@ -31,13 +31,13 @@
 import org.apache.calcite.rel.core.AggregateCall;
 import org.apache.calcite.rel.metadata.RelMetadataQuery;
 import org.apache.calcite.rel.type.RelDataType;
-import org.apache.calcite.rel.type.RelDataTypeFactory;
-import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.util.ImmutableBitSet;
+import org.apache.ignite.internal.sql.engine.exec.exp.agg.AggregateRow;
```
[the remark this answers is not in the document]
lowka: Agree. Moved to PlanUtils (maybe there is a better place for these).

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/BaseQueryContext.java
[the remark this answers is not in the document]
lowka: Fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/sql/fun/IgniteSqlOperatorTable.java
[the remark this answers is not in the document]
lowka: Fixed.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItAggregatesTest.java
```diff
@@ -485,24 +496,90 @@ public void checkEmptyTable(String[] rules) {
...
+                    .returns("s2", 10L, 1L)
+                    .returns("s3", 40L, 1L)
+                    .returns("s1", 30L, 2L)
+                    .returns(null, 40L, 1L)
+                    .returns(null, 20L, 2L)
+                    .returns(null, 20L, 1L)
+                    .check();
+
+        } finally {
+            sql("DROP TABLE test1");
```
AMashenkov, line 526: Let's move to `tearDown` method, with "IF EXISTS" flag pf course.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/agg/Accumulators.java
```diff
@@ -118,6 +117,7 @@ private Supplier<Accumulator> avgFactory(AggregateCall call) {
     private Supplier<Accumulator> sumFactory(AggregateCall call) {
         switch (call.type.getSqlTypeName()) {
             case BIGINT:
+                return () -> new Sum(new LongSumEmptyIsZero());
```
AMashenkov, line 120: What if accumulator value overflows? Can we detect this?
Does SQL standard defines aggregate return types? or it's implementation defined?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AbstractAggregatePlannerTest.java
```diff
@@ -79,29 +79,29 @@ public abstract class AbstractAggregatePlannerTest extends AbstractPlannerTest {
     @SuppressWarnings("NonSerializableFieldInSerializableClass")
     enum TestCase {
         /**
-         * Query: SELECT AVG(val0) FROM test.
+         * Query: SELECT SUM(val0) FROM test.
```
AMashenkov, line 82: Why do you change aggregate function in tests?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/MapReduceHashAggregatePlannerTest.java
```diff
@@ -318,11 +324,9 @@ public void expandDistinctAggregates() throws Exception {
...
                                 .and(not(hasAggregate()))
                                 .and(hasGroups())
-                                .and(input(isInstanceOf(IgniteExchange.class)
-                                        .and(input(isInstanceOf(IgniteMapHashAggregate.class)
-                                                .and(not(hasAggregate()))
-                                                .and(hasGroups())
-                                        ))
+                                .and(input(isInstanceOf(IgniteMapHashAggregate.class)
+                                        .and(not(hasAggregate()))
+                                        .and(hasGroups())
```
AMashenkov, line 329: I guess, before the change, the plan assumes grouping on affinity columns.
Thus, we can check "DISTINCT" condition on each node locally.

After the fix - plan check the common case, when all the data fetched to single node.
Does it means this case is broken?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/agg/Accumulators.java
lowka: At the moment COUNT returns BIGINT. It MAP/REDUCE version of it should also produce the same type.

AMashenkov: Ok, you are right.
Same questions. What about "SUM" aggregate below?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItAggregatesTest.java
lowka: I have no problem with this change, but there are a lot of test cases in this class that use this approach I just followed the suit, it would be strange modify only one case.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/MapReduceHashAggregatePlannerTest.java
lowka: @AMashenkov Yes, It does. And we should discuss this.

lowka: As I explained earlier the number of projections/aggregate affects cost of a query plan.  
At the moment MAP/REDUCE use the fixed number of columns (GROUP_ID, LIST, ). After this patch this is no longer holds, add all intermediate expressions are sent as is.

lowka:
Updated MapReduce aggregate building code  by moving trait building back to ConverterRules.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/MapReduceHashAggregatePlannerTest.java
lowka: Changing the number of columns between affects the cost of `Exchange` operator, because it uses that number (computeSelfCost):
> double rowCount = mq.getRowCount(getInput());
> double bytesPerRow = getRowType().getFieldCount() * IgniteCost.AVERAGE_FIELD_SIZE;

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/JoinCommutePlannerTest.java
```diff
@@ -123,7 +123,8 @@ public void testEnforceJoinOrderHint() throws Exception {
 
     @Test
     public void testOuterCommute() throws Exception {
-        String sql = "SELECT COUNT(*) FROM SMALL s RIGHT JOIN HUGE h on h.id = s.id";
+        // Use aggregates that are the some for both MAP and REDUCE phases.
```
AMashenkov, line 126:
```suggestion
        // Use aggregates that are the same for both MAP and REDUCE phases.
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/JoinCommutePlannerTest.java
```diff
@@ -173,7 +174,8 @@ public void testOuterCommute() throws Exception {
 
     @Test
     public void testInnerCommute() throws Exception {
-        String sql = "SELECT COUNT(*) FROM SMALL s JOIN HUGE h on h.id = s.id";
+        // Use aggregates that are the some for both MAP and REDUCE phases.
```
AMashenkov, line 177:
```suggestion
        // Use aggregates that are the same for both MAP and REDUCE phases.
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/agg/AggregateRow.java
```diff
@@ -0,0 +1,85 @@
...
+
+        return new Object[rowSize];
+    }
+
+    /** Writes aggregate state of the given row to given array. */
+    public void writeTo(Object[] output, ImmutableBitSet allFields, byte groupId) {
+        for (int i = 0; i < accs.size(); i++) {
+            AccumulatorWrapper<RowT> wrapper = accs.get(i);
+            output[i + allFields.cardinality()] = wrapper.end();
+        }
```
AMashenkov, line 78:
```suggestion
        int cardinality = allFields.cardinality();
        for (int i = 0; i < accs.size(); i++) {
            AccumulatorWrapper<RowT> wrapper = accs.get(i);
            output[i + cardinality] = wrapper.end();
        }
```

AMashenkov: I wonder, why ImmutableBitSet doesn't cache cardinality.

lowka: Because Integer/Long ::bitCount is translated into a `POPCNT` instruction that takes only 2 or 3 cycles.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/PlanUtils.java
```diff
@@ -46,4 +62,124 @@ public static boolean complexDistinctAgg(List<AggregateCall> aggCalls) {
...
+
+        if (aggregateType == AggregateType.REDUCE) {
+            Mapping mapping = Mappings.create(MappingType.INVERSE_SURJECTION, fieldIndices.length(), fieldIndices.cardinality());
+            int[] position = new int[1];
+
+            fieldIndices.stream().forEach(b -> {
+                int i = position[0];
+                mapping.set(b, i);
+                position[0] = i + 1;
+            });
```
AMashenkov, line 163:
```suggestion
            int position = 0;

            for(int i : fieldIndices) {
                mapping.set(i, pos++);
            }
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/PlanUtils.java
```diff
@@ -46,4 +62,124 @@ public static boolean complexDistinctAgg(List<AggregateCall> aggCalls) {
...
+
+        for (ImmutableBitSet groupingSet : groupingSets) {
+            for (int field : groupingSet) {
+                fieldIndices.set(field);
+            }
+        }
+
+        if (aggregateType == AggregateType.REDUCE) {
+            Mapping mapping = Mappings.create(MappingType.INVERSE_SURJECTION, fieldIndices.length(), fieldIndices.cardinality());
+            int[] position = new int[1];
```
AMashenkov, line 157: Does it do the same as `Common.inverseTrimmingMapping`?
Maybe we can reuse method from common. 

Agree, you implementation is better as it avoids boxing. But it's better to use utility method here, and think about avoiding boxing/unboxing in Commons later.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/PlanUtils.java
AMashenkov, line 163: Does it do the same as `Common.inverseTrimmingMapping`?
Maybe we can reuse method from common. 

Agree, you implementation is better as it avoids boxing. But it's better to use utility method here, and think about avoiding boxing/unboxing in Commons later.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/PlanUtils.java
```diff
@@ -46,4 +62,124 @@ public static boolean complexDistinctAgg(List<AggregateCall> aggCalls) {
...
+
+    private static void addAccumulatorFields(IgniteTypeFactory typeFactory, List<AggregateCall> aggregateCalls, Builder builder) {
+        Accumulators accumulators = new Accumulators(typeFactory);
+
+        for (int i = 0; i < aggregateCalls.size(); i++) {
+            AggregateCall call = aggregateCalls.get(i);
+
+            Accumulator acc = accumulators.accumulatorFactory(call).get();
+            RelDataType fieldType = acc.returnType(typeFactory);
+            String fieldName = format("_ACC{}", i);
```
AMashenkov, line 180: It is not good idea to parse 'format' in a loop.
```suggestion
            String fieldName = "_ACC{}" + i;
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/PlanUtils.java
lowka: >Does it do the same as Common.inverseTrimmingMapping?
>Maybe we can reuse method from common.
No it is not. Because Mapping returned from inverseTrimmingMapping does not support `getSource`.

lowka: Rewrote stream as a while loop.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/PlanUtils.java
lowka: Fixed.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/JoinCommutePlannerTest.java
lowka: Fixed

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/JoinCommutePlannerTest.java
lowka: Fixed.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/MapReduceHashAggregatePlannerTest.java
AMashenkov: Seems, `IgniteMdDistinctRowCount.getDistinctRowCount` is used for calculating aggregate nodes costs.
It's code looks weird. BTW, I have no idea why don't we use default implementation.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItAggregatesTest.java
AMashenkov: I think it would be ok to fix only changes and new tests.

AMashenkov:
I'm ok with changes, in general.

However, there is an issue when Exchange pushed down to under MapAggregate.
Most likely, the reason is overriden cost calculation function (IgniteMdDistinctRowCount.getDistinctRowCount) in Ignite.
Let's create ticket to fix and investigate, if default function brokes smth and doesn't fit our needs due to some reason.

AMashenkov (approved):

AMashenkov:
@korlov42, do you have any concerns?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/MapReduceHashAggregatePlannerTest.java
```diff
@@ -553,11 +620,13 @@ private void checkGroupsWithOrderByGroupColumnsHash(TestCase testCase, RelCollat
...
                         .and(s -> s.collation().equals(collation))
-                        .and(input(isInstanceOf(IgniteReduceHashAggregate.class)
-                                .and(input(isInstanceOf(IgniteMapHashAggregate.class)
-                                        //TODO: Why can't Map be pushed down to under 'exchange'.
-                                        .and(input(isInstanceOf(IgniteExchange.class)
-                                                .and(input(isTableScan("TEST")))
+                        .and(input(isInstanceOf(IgniteProject.class)
+                                .and(input(isInstanceOf(IgniteReduceHashAggregate.class)
+                                        .and(input(isInstanceOf(IgniteMapHashAggregate.class)
+                                                //TODO: Why can't Map be pushed down to under 'exchange'.
```
AMashenkov, line 626: TODO without a ticket link

lowka: Fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/PlanUtils.java
```diff
@@ -46,4 +60,126 @@ public static boolean complexDistinctAgg(List<AggregateCall> aggCalls) {
         }
         return false;
     }
+
+    /**
+     * Creates a row type for REDUCE phase of a two phase aggregate used by hash-based map/reduce implementation.
```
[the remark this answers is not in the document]
lowka: Fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/HashAggregateNode.java
```diff
@@ -51,10 +51,14 @@ public class HashAggregateNode<RowT> extends AbstractNode<RowT> implements Singl
...
     private final RowFactory<RowT> rowFactory;
 
-    private final ImmutableBitSet grpSet;
+    /** A bit set that contains fields included in all grouping sets. */
+    private final ImmutableBitSet allFields;
 
     private final List<Grouping> groupings;
 
+    /** Mapping between group columns and their positions inside an input row.*/
+    private final Mapping mapping;
```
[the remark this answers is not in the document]
korlov42: Can't find a single test in which this mapping is not an identity mapping (0->0, 1->1, etc)... Do you have such test in mind? 

The reason I want to see it in action is that I have feeling you are trying to fix broken plan in runtime with this mapping. We definitely don't want to do this.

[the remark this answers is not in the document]
lowka: @korlov42 

We need this mapping because the positions between the same columns on MAP phase and REDUCE phase are different. Example.

> sql("CREATE TABLE testMe (a INTEGER, b INTEGER, s VARCHAR);");
> sql("INSERT INTO testMe VALUES (11, 1, 'hello'), (12, 2, 'world'), (11, 3, NULL)");
> sql("INSERT INTO testMe VALUES (11, 3, 'hello'), (12, 2, 'world'), (10, 5, 'ahello'), (13, 6, 'world')");
            
Query:

>SELECT COUNT(a), COUNT(DISTINCT(b)) FROM testMe

Logical plan:
> LogicalAggregate(group=[{}], EXPR$0=[COUNT($0)], EXPR$1=[COUNT(DISTINCT $1)]), id = 5
>  LogicalProject(A=[$0], B=[$1]), id = 4
>    IgniteLogicalTableScan(table=[[PUBLIC, TESTME]]), id = 1

Optimized plan:

>IgniteProject(EXPR$0=[CAST($0):BIGINT NOT NULL], EXPR$1=[$1]), id = 128
>  IgniteProject(EXPR$0=[$0], EXPR$1=[CAST($1):BIGINT NOT NULL]), id = 127
>    IgniteReduceSortAggregate(group=[{}], EXPR$0=[$SUM0($0)], agg#1=[$SUM0($1)], collation=[[]]), id = 126
>      IgniteMapSortAggregate(group=[{}], EXPR$0=[$SUM0($1)], EXPR$1=[COUNT($0)], collation=[[]]), id = 125
>        IgniteProject(B=[$0], EXPR$0=[CAST($1):BIGINT NOT NULL]), id = 124
>          IgniteReduceHashAggregate(group=[{1}], agg#0=[$SUM0($1)]), id = 123
>            IgniteMapHashAggregate(group=[{1}], EXPR$0=[COUNT($0)]), id = 122
>              IgniteTableScan(table=[[PUBLIC, TESTME]], requiredColumns=[{0, 1}]), id = 59


If we look at the first Aggregate (the aggregate that takes a table as it's input)
>        IgniteReduceHashAggregate(group=[{1}], agg#0=[$SUM0($1)]), id = 123
>            IgniteMapHashAggregate(group=[{1}], EXPR$0=[COUNT($0)]), id = 122

Aggregate:

> input: RecordType(INTEGER A, INTEGER B)
> grouping sets: [{1}]
> output:  RecordType(INTEGER B, BIGINT _ACC0, TINYINT _GROUP_ID)

MAP 
>input: RecordType(INTEGER A, INTEGER B)
>output: RecordType(INTEGER B, BIGINT _ACC0, TINYINT _GROUP_ID)
>grouping sets: [{1}]

REDUCE 
>input:  RecordType(INTEGER B, BIGINT _ACC0, TINYINT _GROUP_ID)
>output: RecordType(INTEGER f0, BIGINT f1)
>grouping sets: [{1}]

We see that Column A is assigned index to 0, column B is assigned to 1. Grouping sets are `{{1}} ` where `1` is a position in MAP's input. 
The ticket states that MAP/REDUCE should use the same logic (because they are almost identical). To achieve that we need a mapping between for columns used on REDUCE phase and for MAP phase we use identity mapping.
Output of this aggregate does not produce `A` but REDUCE phase uses `B` on the second phase in both it's input and its output. So we map A to 0, 1 to B on MAP phase, and B to 0 on map phase.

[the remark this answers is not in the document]
lowka: Updated grouping set field index computation on REDUCE phase and removed `mapping` property of `HashAggregateNode`.

korlov42 (approved):

PR #2316, opened 2023-07-13
IGNITE-19979 Table start might fail due to an outdated causality token

rpuch:
https://issues.apache.org/jira/browse/IGNITE-19979

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
@@ -909,27 +909,27 @@ public class TableManager extends Producer<TableEvent, TableEventParameters> imp
             return allOf(futures);
         };
 
-        return localPartsByTableIdVv.update(causalityToken, (previous, throwable) -> inBusyLock(busyLock, () -> {
-            return getOrCreatePartitionStorages(table, parts).thenApply(u -> {
-                var newValue = new HashMap<>(previous);
+        // NB: all vv.update() calls must be made from the synchronous part of the method (not in thenCompose()/etc!).
+        CompletableFuture<?> localPartsUpdateFuture = localPartsByTableIdVv.update(causalityToken,
+                (previous, throwable) -> inBusyLock(busyLock, () -> {
+                    return getOrCreatePartitionStorages(table, parts).thenApply(u -> {
+                        var newValue = new HashMap<>(previous);
 
-                newValue.put(tableId, parts);
+                        newValue.put(tableId, parts);
 
-                return newValue;
-            });
-        })).thenCompose(unused -> {
-            CompletableFuture<Void> updateAssignmentsFuture = tablesByIdVv.get(causalityToken).thenComposeAsync(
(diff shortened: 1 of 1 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -909,27 +909,27 @@ private CompletableFuture<?> createTablePartitionsLocally(
...
-                    return failedFuture(e);
-                }
+        return assignmentsUpdatedVv.update(causalityToken, (token, e) -> {
+            if (e != null) {
+                return failedFuture(e);
+            }
 
-                return updateAssignmentsFuture;
-            });
+            return localPartsUpdateFuture.thenCompose(unused ->
```
ibessonov, line 929: Assuming that `assignmentsUpdatedVv` depends on `localPartsByTableIdVv`, I think it's safe to assume that this particular `thenCompose` is redundant.

ibessonov: We could also assert that `tablesByIdVv.get(causalityToken)` is already completed at this point

rpuch: If I change the code to the following

```
        return assignmentsUpdatedVv.update(causalityToken, (token, e) -> {
            if (e != null) {
                return failedFuture(e);
            }

            return tablesByIdVv.get(causalityToken)
                    .thenComposeAsync(tablesById -> inBusyLock(busyLock, updateAssignmentsClosure), ioExecutor);
        });
```
, I get an assertion:

```
Caused by: java.lang.AssertionError
	at org.apache.ignite.internal.table.distributed.TableManager.getPartitionStorages(TableManager.java:2502)
	at org.apache.ignite.internal.table.distributed.TableManager.lambda$createTablePartitionsLocally$13(TableManager.java:763)
	at org.apache.ignite.internal.util.IgniteUtils.inBusyLock(IgniteUtils.java:893)
	at org.apache.ignite.internal.table.distributed.TableManager.lambda$createTablePartitionsLocally$17(TableManager.java:930)
	at java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1072)
	... 4 more
```

Also, `tablesByIdVv.get(causalityToken)` is not done when `assignmentsUpdatedVv`-related callback is called, so the assertion cannot be added.

rpuch: Looks like VV.complete() is not guaranteed to be called strictly after all VV.update-related futures complete

ibessonov (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
ibessonov: Ok, considering that you're fixing a flaky test, this is urgent, so let's investigate the assertion later

PR #2319, opened 2023-07-14
IGNITE-19984 Improve "dependingOn" documentation and add clarifying tests.

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-19984

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/causality/IncrementalVersionedValue.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/causality/IncrementalVersionedValue.java
@@ -63,7 +63,14 @@ public class IncrementalVersionedValue<T> implements VersionedValue<T> {
 
     /**
      * This registry chains two versioned values. The value, that uses this registry in the constructor, will be completed strictly after
-     * the value, passed into this method.
+     * the value, passed into this method, meaning that {@code resultVv.get(token).isDone();} will always imply imply
+     * {@code vv.get(token).isDone();} for the same token value.
+     *
+     * <p>While affecting the state of resulting futures, this dependency doesn't affect the order of {@link #update(long, BiFunction)}
+     * closures execution. These closures will still be called independently once the required parameter value is available.
+     *
+     * <p>In the case of "fresh" VV with no updates, first closure is always being executed synchronously inside of the
+     * {@link #update(long, BiFunction)} call.
      */
     public static Consumer<LongFunction<CompletableFuture<?>>> dependingOn(IncrementalVersionedValue<?> vv) {
         return callback -> vv.whenComplete((causalityToken, value, ex) -> callback.apply(causalityToken));
--- a/modules/core/src/test/java/org/apache/ignite/internal/causality/IncrementalVersionedValueTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/causality/IncrementalVersionedValueTest.java
@@ -19,6 +19,7 @@ package org.apache.ignite.internal.causality;
 
 import static java.util.concurrent.CompletableFuture.completedFuture;
 import static java.util.concurrent.CompletableFuture.failedFuture;
+import static org.apache.ignite.internal.causality.IncrementalVersionedValue.dependingOn;
 import static org.apache.ignite.internal.testframework.IgniteTestUtils.assertThrowsWithCause;
 import static org.apache.ignite.internal.testframework.IgniteTestUtils.runRace;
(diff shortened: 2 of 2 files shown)


modules/core/src/main/java/org/apache/ignite/internal/causality/IncrementalVersionedValue.java
```diff
@@ -63,7 +63,14 @@ public class IncrementalVersionedValue<T> implements VersionedValue<T> {
 
     /**
      * This registry chains two versioned values. The value, that uses this registry in the constructor, will be completed strictly after
-     * the value, passed into this method.
+     * the value, passed into this method, meaning that {@code resultVv.get(token).isDone();} will always imply imply
```
rpuch, line 66:
```suggestion
     * the value, passed into this method, meaning that {@code resultVv.get(token).isDone();} will always imply
```

modules/core/src/test/java/org/apache/ignite/internal/causality/IncrementalVersionedValueTest.java
```diff
@@ -341,6 +344,51 @@ void testCompleteMultipleFutures() {
...
+        int token = 1;
+
+        vv0.update(token, (i, e) -> completedFuture(i + 1));
+
+        vv1.update(token, (i, e) -> completedFuture(i + 1));
+
+        register.moveRevision(token);
+
+        // Will always complete in time.
+        vv1.get(token).join();
```
rpuch, line 365: How about using `assertThat(..., willCompleteSuccessfully())` just in case there is a bug preventing the completion?

modules/core/src/test/java/org/apache/ignite/internal/causality/IncrementalVersionedValueTest.java
```diff
@@ -341,6 +344,51 @@ void testCompleteMultipleFutures() {
...
+
+        var vv1 = new IncrementalVersionedValue<>(dependingOn(vv0), () -> 1);
+
+        int token = 1;
+
+        vv0.update(token, (i, e) -> completedFuture(i + 1));
+
+        vv1.update(token, (i, e) -> completedFuture(i + 1));
+
+        register.moveRevision(token);
```
rpuch, line 362: Maybe I'm looking in the wrong direction, but I don't see any concurrency here (no async calls), so everything will be executed in the same thread and the order of execution will be deterministic. Would it make sense to make the test more interesting by using something like `supplyAsync()` instead of `completedFuture()`?

modules/core/src/test/java/org/apache/ignite/internal/causality/IncrementalVersionedValueTest.java
```diff
@@ -341,6 +344,51 @@ void testCompleteMultipleFutures() {
...
+
+        assertTrue(vv0.get(token).isDone());
+    }
+
+    /**
+     * Tests that {@link IncrementalVersionedValue#update(long, BiFunction)} closure is called immediately when underlying value is
+     * accessible, i.e. when there were no other updates.
+     */
+    @Test
+    public void testImmediateUpdate() {
```
rpuch, line 375: Should this be tested? Is this property a desired property important for the contract of `IncrementalVersionedValue`, or it's just a coincidence that the current implementation behaves like this?

modules/core/src/test/java/org/apache/ignite/internal/causality/IncrementalVersionedValueTest.java
ibessonov: Ok, why not. I just wanted to avoid as much CPU activity as I could for the next assertion

modules/core/src/test/java/org/apache/ignite/internal/causality/IncrementalVersionedValueTest.java
ibessonov: Ok, I'll do it

modules/core/src/test/java/org/apache/ignite/internal/causality/IncrementalVersionedValueTest.java
ibessonov: Technically, it's undefined, but I'd rather have this as an explicit contract. It is also the sole reason why "dependingOn" doesn't work the way I expected it to work earlier

ibessonov: Done

rpuch (approved):

PR #2321, opened 2023-07-14
IGNITE-19960 Backport 1.3.10 release form the JRaft

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-19960


modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
```diff
@@ -1857,36 +1856,30 @@ public void apply(final Task task) {
...
+
+        final EventTranslator<LogEntryAndClosure> translator = (event, sequence) -> {
+            event.reset();
+            event.nodeId = getNodeId();
+            event.done = task.getDone();
+            event.entry = entry;
+            event.expectedTerm = task.getExpectedTerm();
+        };
+        switch (this.options.getApplyTaskMode()) {
+            case Blocking:
```
ascherbakoff, line 1868: I'm not sure if a blocking mode has any use for us. Probably shouldn't be ported at all.
@alievmirza Do you know a single use case ?

ascherbakoff (approved):

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
alievmirza: For example, it could help us to not fail with overload exception in scenarios like this https://issues.apache.org/jira/browse/IGNITE-19419

Also in the current implementation after overload of disk queue in LogManagerImpl and overload of FSMCallerImpl they enter erroneous state that is never reset, which makes impossible the further work of the component. It was fixed also

ascherbakoff: It is forbidden to use blocking code in the worker threads. The patch adds retries on a busy state, this should be enough.

PR #2326, opened 2023-07-17
IGNITE-19009 Introduce file transfer support in messaging

ivgag:
https://issues.apache.org/jira/browse/IGNITE-19009

### DIFF
--- a/modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
+++ b/modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
@@ -35,17 +35,15 @@ import static org.apache.ignite.internal.util.GridUnsafe.IS_BIG_ENDIAN;
 import static org.apache.ignite.internal.util.GridUnsafe.LONG_ARR_OFF;
 import static org.apache.ignite.internal.util.GridUnsafe.SHORT_ARR_OFF;
 
-import java.io.BufferedInputStream;
-import java.io.BufferedOutputStream;
 import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.lang.reflect.Array;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
+import java.nio.channels.FileChannel;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.BitSet;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
```diff
@@ -1662,12 +1682,167 @@ public <C extends Set<?>> C readSet(MessageCollectionItemType itemType, MessageR
...
+
+                    if (fileReader == null) {
+                        fileReader = new BufferedInputStream(new FileInputStream(file));
+                    }
+
+                    int pos = buf.position();
+                    int toWrite = Math.min(buf.remaining(), fileReader.available());
+
+                    if (toWrite > 0) {
+                        byte[] arr = new byte[toWrite];
```
SammyVimes, line 1728: I think it is possible to read file directly to the output buffer to avoid allocations

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
```diff
@@ -1662,12 +1682,167 @@ public <C extends Set<?>> C readSet(MessageCollectionItemType itemType, MessageR
...
+
+                    //noinspection fallthrough
+                case 2:
+                    lastFinished = false;
+
+                    int pos = buf.position();
+                    int remaining = buf.remaining();
+                    int toRead = (int) Math.min(remaining, fileSize - bytesWritten);
+
+                    byte[] arr = new byte[toRead];
```
SammyVimes, line 1801: Same here, we should write to writer from bytebuffer without allocations

modules/network/src/main/java/org/apache/ignite/internal/network/direct/stream/DirectByteBufferStreamImplV1.java
```diff
@@ -1662,12 +1682,167 @@ public <C extends Set<?>> C readSet(MessageCollectionItemType itemType, MessageR
...
+
+                    //noinspection fallthrough
+                case 1:
+                    String fileName = readString();
+                    if (!lastFinished) {
+                        return null;
+                    }
+
+                    filePath = Files.createTempDirectory("").resolve(fileName);
+                    fileWriter = new BufferedOutputStream(new FileOutputStream(filePath.toFile(), true));
```
SammyVimes, line 1788: I presume, this doesn't retain file attributes? I'm not sure if they are needed though

modules/network/src/test/resources/files/string.json
```diff
@@ -0,0 +1,5 @@
+{
```
SammyVimes, line 1: Not sure if we need these files, it's easier and cleaner to generate them

SammyVimes (changes requested):

PR #2335, opened 2023-07-19
IGNITE-19999 Sql. Add planner test to verify numeric type coercion in binary comparison

korlov42:


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AbstractPlannerTest.java
```diff
@@ -570,6 +570,10 @@ protected <T extends RelNode> void assertPlan(
                 return;
             }
         } catch (Throwable th) {
+            if (lastErrorMsg == null) {
```
ygerzhedovich, line 573: is it last or first?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/datatypes/NumericComparisonTypeCoercionTest.java
```diff
@@ -0,0 +1,965 @@
...
+                        .secondOpMatches(castTo(NativeTypes.DOUBLE)),
+
+                forTypePair(NumericPair.DECIMAL_5_0_DOUBLE)
+                        .firstOpMatches(castTo(NativeTypes.DOUBLE))
+                        .secondOpMatches(ofTypeWithoutCast(NativeTypes.DOUBLE)),
+
+
+                forTypePair(NumericPair.DECIMAL_6_1_DECIMAL_6_1)
+                        .firstOpMatches(ofTypeWithoutCast(Types.DECIMAL_6_1))
+                        .secondOpMatches(ofTypeWithoutCast(Types.DECIMAL_6_1)),
```
AMashenkov, line 724: This looks similar to NumericPair.DECIMAL_8_3_DECIMAL_8_3 and NumericPair.DECIMAL_5_3_DECIMAL_5_3.

AMashenkov (approved):

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/AbstractPlannerTest.java
korlov42: well, it's a bit messy... This variable is just a field of an abstract class. Every test in test class extending `AbstractPlannerTest` will overwrite this variable if validation has failed. So, this is the last error message for test class, but this is a first and only error message for test.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/datatypes/NumericComparisonTypeCoercionTest.java
korlov42: that's true, but I would prefer to leave this as is. 

To wrote all those test cases, first, I defined a list of types that I would like to verify. Among others, I definitely did want to verify different DECIMAL types: exactly the same types, types with equal precision but different scale, types with equal scale but different precision, completely different types. Then, I made a Cartesian join of the type list a leave only those pairs where first type is lesser then or equals to second. That's how I end up with *duplicates. Technically, it's possible to exclude such pairs, but it much easier and  safer just to include them as valid test cases. After all, every test case takes couple of milliseconds only. Not a big deal, I believe.

PR #2339, opened 2023-07-21
IGNITE-19936  Add username and password parameters to the connect command

nevdmitry:
https://issues.apache.org/jira/browse/IGNITE-19936


modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/connect/ConnectOptions.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.jetbrains.annotations.Nullable;
+import picocli.CommandLine.Option;
+
+/**
+ * Mixin class for connect command options.
+ */
+public class ConnectOptions {
+
+
```
PakhomovAlexander, line 18: two empty lines

PakhomovAlexander:
```bash 
[apakh:defaultNode]>
[apakh:defaultNode]> disconnect
Disconnected from http://localhost:10300
[disconnected]> connect http://localhost:10300
Connected to http://localhost:10300
[apakh:defaultNode]> disconnect
Disconnected from http://localhost:10300
[disconnected]> connect defaultNode
Invalid value for positional parameter at index 0 (<nodeNameOrUrl>): Node defaultNode not found. Provide valid name or use URL
```

if this case also does not work on main, create a ticket, please.

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/questions/ConnectToClusterQuestion.java
```diff
@@ -108,6 +112,25 @@ public FlowBuilder<Void, String> askQuestionIfConnected(String nodeUrl) {
...
+     * Ask if the user wants to store credentials in config.
+     *
+     * @param username username.
+     * @param password password
+     * @return {@link FlowBuilder} instance which provides result.
+     */
+    public void askQuestionToStoreCredentials(@Nullable String username, @Nullable String password) {
+        if (!nullOrBlank(username) && !nullOrBlank(password)) {
+            QuestionUiComponent question = fromYesNoQuestion(
+                    "Do you want to store username and password in cli configuration config?"
```
PakhomovAlexander, line 125: I think a question like: "Remember current credentials?" is shorter and less-implementation-specific.

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/connect/ConnectReplCommand.java
```diff
@@ -50,10 +57,11 @@ public class ConnectReplCommand extends BaseCommand implements Runnable {
     @Override
     public void run() {
         question.askQuestionIfConnected(nodeNameOrUrl.stringUrl())
-                .map(UrlCallInput::new)
+                .map(url -> new ConnectCallInput(url, connectOptions.username(), connectOptions.password()))
                 .then(Flows.fromCall(connectCall))
                 .verbose(verbose)
                 .print()
+                .onSuccess(() -> question.askQuestionToStoreCredentials(connectOptions.username(), connectOptions.password()))
```
PakhomovAlexander, line 64: I did not manage to be asked whatever I tried.

PakhomovAlexander (changes requested):

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/connect/ItConnectWithBasicAuthenticationCommandTest.java
```diff
@@ -99,4 +103,155 @@ void failToConnectWithWrongCredentials() {
...
+    @Test
+    void connectFailIfPasswordNotDefined() {
+        // Given basic authentication is NOT configured in config file
+        configManagerProvider.setConfigFile(createIntegrationTestsConfig());
+
+        // Given prompt before connect
+        assertThat(getPrompt()).isEqualTo("[disconnected]> ");
+
+        // When connect with auth parameters
+        //execute("connect", nodeName(), "--username", "admin1", "--password", "password1");
```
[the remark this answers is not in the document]
PakhomovAlexander: Instead of a direct message to the obvious review requests we just hit "Resolve conversation" button if we are agreed with the comments.

modules/cli/src/main/java/org/apache/ignite/internal/cli/call/connect/ConnectCallInput.java
```diff
@@ -12,23 +12,54 @@ public class ConnectCallInput implements CallInput {
...
         return username;
     }
 
     @Nullable
-    String getPassword() {
+    String password() {
         return password;
     }
+
+    public static class ConnectCallInputBuilder {
```
PakhomovAlexander, line 35: When we add a builder to the class we tend to create a static method `builder()` and close a builder's constructor.

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/Options.java
```diff
@@ -261,7 +261,7 @@ public static final class Constants {
 
         public static final String PASSWORD_OPTION = "--password";
 
-        public static final String PASSWORD_OPTION_SHORT = "--pwd";
+        public static final String PASSWORD_OPTION_SHORT = "--p";
```
PakhomovAlexander, line 264: Short options should be defined with single dash `-p`

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/Options.java
```diff
@@ -258,5 +258,23 @@ public static final class Constants {
...
+
+        public static final String PASSWORD_OPTION = "--password";
+
+        public static final String PASSWORD_OPTION_SHORT = "--p";
+
+        public static final String PASSWORD_OPTION_DESC = "Password to connect to cluster";
+
+        public static final String USERNAME_OPTION = "--username";
+
+        public static final String USERNAME_OPTION_SHORT = "--u";
```
PakhomovAlexander, line 270: Same here for single dash

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/connect/ConnectOptions.java
```diff
@@ -0,0 +1,39 @@
...
+    @Option(names = {USERNAME_OPTION, USERNAME_OPTION_SHORT}, description = USERNAME_OPTION_DESC)
+    private String username;
+
+    @Option(names = {PASSWORD_OPTION, PASSWORD_OPTION_SHORT}, description = PASSWORD_OPTION_DESC)
+    private String password;
```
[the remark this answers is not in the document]
PakhomovAlexander: I mean I should not be able to run command like `connect --username <username` but I can. I expect a message from Picocli explaining that I should use both `--username` and `--password` or non of them.

PakhomovAlexander (changes requested):

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/connect/ItConnectWithBasicAuthenticationCommandTest.java
```diff
@@ -99,4 +109,178 @@ void failToConnectWithWrongCredentials() {
...
+        assertThat(getPrompt()).isEqualTo("[disconnected]> ");
+
+        // When connect with auth parameters
+        execute("connect", "--username", "admin", "--password", "wrong-password");
+
+        // Then
+        assertAll(
+                this::assertOutputIsEmpty,
+                () -> assertErrOutputIs("Authentication error" + System.lineSeparator()
+                        + "Could not connect to node with URL http://localhost:10300. Check authentication configuration"
```
PakhomovAlexander, line 151: `Check authentication configuration` were fine when we used a configuration to provide usename/password but now a user passes username and password explicitly with command line arguments. What "configuration" is supposed to mean in this case?

I would say that we either divide these two cases (config provided creds and explicit ones) or just rephrase the general error to commit "configuration" word.

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/connect/ItConnectWithBasicAuthenticationCommandTest.java
```diff
@@ -99,4 +109,178 @@ void failToConnectWithWrongCredentials() {
...
```
PakhomovAlexander, line 224: Ah, "Check authentication configuration" here is misleading information. In fact, configuration is ok but provided options are not.

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/connect/ItConnectWithBasicAuthenticationCommandTest.java
```diff
@@ -99,4 +109,178 @@ void failToConnectWithWrongCredentials() {
...
+                        "Config saved" + System.lineSeparator() + "Connected to http://localhost:10300" + System.lineSeparator())
+        );
+
+        // And prompt shows user name and node name
+        assertThat(getPrompt()).isEqualTo("[admin:" + nodeName() + "]> ");
+    }
+
+    @Test
+    @DisplayName("Should create correct api client even if user doesn't store credentials in settings.")
+    void sessionListenersShouldBeInvokedWithCorrectCredentials() throws IOException {
```
PakhomovAlexander, line 262: I think yuo mix different levels of abstraction in test and naming. The test itself is on the highest abstraction level and operated with entities like "connect, prompt, disconnect, configuration, answer". But test name (both method signature and DisplayName) operates with entities like: API client and listeners. This is a lower level of abstraction. 

I think it is better not to mix them. The test name should use the same vocabulary that test uses.

modules/cli/src/main/java/org/apache/ignite/internal/cli/call/connect/ConnectCallInput.java
```diff
@@ -0,0 +1,106 @@
...
+
+package org.apache.ignite.internal.cli.call.connect;
+
+import org.apache.ignite.internal.cli.core.call.CallInput;
+import org.jetbrains.annotations.Nullable;
+
+/** Input for the {@link ConnectCall} call. */
+public class ConnectCallInput implements CallInput {
+
+
```
PakhomovAlexander, line 26: two empty lines

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/Options.java
```diff
@@ -266,5 +266,23 @@ public static final class Constants {
...
+
+        public static final String PASSWORD_OPTION = "--password";
+
+        public static final String PASSWORD_OPTION_SHORT = "-p";
+
+        public static final String PASSWORD_OPTION_DESC = "Password to connect to cluster";
+
+        public static final String USERNAME_OPTION = "--username";
+
+        public static final String USERNAME_OPTION_SHORT = "-u";
```
PakhomovAlexander, line 278: We use -u and -p for URL and Profile thighs. Can you please create a ticket for revising all short options and reworking this concept?

modules/cli/src/main/java/org/apache/ignite/internal/cli/commands/questions/ConnectToClusterQuestion.java
```diff
@@ -109,6 +113,29 @@ public FlowBuilder<Void, String> askQuestionIfConnected(String nodeUrl) {
...
+     *
+     * @param username username.
+     * @param password password
+     */
+    public void askQuestionToStoreCredentials(@Nullable String username, @Nullable String password) {
+        if (!nullOrBlank(username) && !nullOrBlank(password)) {
+            String storedUsername = configManagerProvider.get().getCurrentProperty(BASIC_AUTHENTICATION_USERNAME.value());
+            String storedPassword = configManagerProvider.get().getCurrentProperty(BASIC_AUTHENTICATION_PASSWORD.value());
+
+            //Ask question only if cli config has different values
```
PakhomovAlexander, line 127:
```suggestion
            // Ask question only if cli config has different values.
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/rest/ApiClientFactory.java
```diff
@@ -98,27 +111,49 @@ private ApiClient getClientFromSettings(ApiClientSettings settings) {
...
-
-        return builder.build();
     }
 
-    public String basicAuthenticationUsername() {
+    private ApiClientSettingsBuilder setupAuthentication(ApiClientSettingsBuilder builder) {
         ConfigManager configManager = configManagerProvider.get();
-        return configManager.getCurrentProperty(BASIC_AUTHENTICATION_USERNAME.value());
+
+        // Use credentials from current session settings if exist
```
PakhomovAlexander, line 144:
```suggestion
        // Use credentials from current session settings if exist.
```

PakhomovAlexander (approved):

PR #2342, opened 2023-07-21
IGNITE-18453: Query executor thread synchronously waits for leaders on query mapping.

lowka:
Removes colocation group lookups from `IgniteTable`. All `ColocationGroup` lookups are now done via `ExecutableTable::primaryReplicas` - updated `InternalTable::primaryReplicas` to return a `CompletableFuture`.  

- Converts `IgniteFragmentMapping` into a visitor.
- Moves fragment mapping code from `Fragment` to `FragmentMapping` to make it possible to make fragment mapping code asynchronous.
- Merges MultiStepQuery/Ddl Plans into one, removes `QueryTemplate`, `ExecutionPlan` because as they are simple wrappers for a list of fragments.

- Updates `TestNode` now uses colocation group from created from node names specified by data providers (node name, data variant) or if they has not been specified uses all node names.

https://issues.apache.org/jira/browse/IGNITE-18453


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -820,6 +830,27 @@ private void enlist(SourceAwareIgniteRel rel) {
             }.visit(fragment.root());
         }
 
+        private CompletableFuture<MultiStepPlan> mapFragments(MultiStepPlan plan) {
+            List<IgniteRel> fragments = plan.fragments()
+                    .stream()
+                    .map(Fragment::root)
+                    .collect(Collectors.toList());
```
AMashenkov, line 837: Does it make sense to avoid collection copying?

We could use smth like `org.apache.ignite.internal.utilTransformingIterator`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -952,4 +983,30 @@ public interface ImplementorFactory<RowT> {
...
+
+            CompletableFuture<Void> all = CompletableFuture.allOf(list.toArray(new CompletableFuture[0]));
+
+            return all.thenApply(
+                    v -> list.stream()
+                            .map(CompletableFuture::join)
+                            .collect(Collectors.toMap(Pair::getFirst, Pair::getSecond)));
+
+        }
+    }
```
AMashenkov, line 1011: Seems, this method can be static helper method and class can be dropped.
Or maybe the whole method can be moved to `ResolvedDependencies` class.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
lowka: Sure. I thought of it but I forgot about `TransformingIterator`.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
lowka: Sure. This class can be replaced with a simple function that does exactly what execute does.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
lowka: Well, here is a problem it is a bad idea to accept an iterator because is modifiable. We need something like transforming iterable here.

lowka: Added `TransformingIterator::newIterable`. And replaced `List<IgniteRel>` with `Iterable<IgniteRel>` in DepedencyResolver's signature.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
lowka: Fixed.

AMashenkov (approved):

PR #2345, opened 2023-07-24
IGNITE-14865 Improve error handling in SQL module.

xtern:
https://issues.apache.org/jira/browse/IGNITE-14865

We have two main entry points in the SQL module for all public API

- `AsyncSqlCursorImpl#requestNextAsync`
- `SqlQueryProcessor#querySingleAsync`

**requestNextAsync** converts internal exceptions to public by itself.

**querySingleAsync**, cannot translate internal exceptions because SessionImpl#execAsync (and others) handles an internal `SessionNotFoundException` that currently has no public equivalent (or at least separate sql code).

 So for **querySingleAsync** exception mapping is made in 3 places
 (right after `SessionNotFoundException` handling):

- `SessionImpl#executeAsync`
- `SessionImpl#executeBatchAsync`
- `JdbcConnectionContext#doInSession`

We also have a `RemoteFragmentExecutionException` which must have a valid traceId and error code, so we convert the internal exception to public and sending traceId and error code in `ExchangeServiceImpl#sendError`. Those. in this case the exception is converted twice. First time on the remote host, and then the resulting `RemoteFragmentExecutionException` converts into a `SqlException` with specified error code.


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDynamicParameterTest.java
```diff
@@ -159,8 +160,11 @@ public void testWithDifferentParametersTypes() {
      */
     @Test
     public void testWithDifferentParametersTypesMismatch() {
-        assertThrows(IgniteException.class, () -> assertQuery("SELECT COALESCE(12.2, ?)").withParams("b").check());
-        assertThrows(IgniteException.class, () -> assertQuery("SELECT COALESCE(?, ?)").withParams(12.2, "b").check());
+        SqlException t = assertThrows(SqlException.class, () -> assertQuery("SELECT COALESCE(12.2, ?)").withParams("b").check());
```
ygerzhedovich, line 163: there is org.apache.ignite.internal.sql.engine.util.SqlTestUtils#assertThrowsSqlException

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
```diff
@@ -327,7 +346,7 @@ public Publisher<Long> executeBatchReactive(@Nullable Transaction transaction, S
     @Override
     public void close() {
         try {
-            closeAsync().toCompletableFuture().get();
+            closeAsync().get();
         } catch (ExecutionException e) {
             sneakyThrow(IgniteExceptionUtils.copyExceptionWithCause(e));
```
AMashenkov, line 351: Shoud we map to public exception here?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExchangeServiceImpl.java
```diff
@@ -112,45 +109,28 @@ public CompletableFuture<Void> request(String nodeName, UUID queryId, long fragm
     /** {@inheritDoc} */
     @Override
     public CompletableFuture<Void> sendError(String nodeName, UUID queryId, long fragmentId, Throwable error) {
-        IgniteException errorWithCode = wrapIfNecessary(error);
+        Throwable errorWithCode = IgniteExceptionMapperUtil.mapToPublicException(ExceptionUtils.unwrapCause(error));
```
AMashenkov, line 112: Why do we convert exception here?
Seems, we can convert internal Traceble exception to default one with INTERNAL_ERR code and get lost traceID.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDynamicParameterTest.java
xtern: Thanks, I've fixed all the places in the PR where applicable.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExchangeServiceImpl.java
xtern: Thanks, seems there is no need to convert exception here. fixed.

AMashenkov (approved):

PR #2349, opened 2023-07-24
IGNITE-19566 Remove ignite-core dependency from ignite-api module

sk0x50:
https://issues.apache.org/jira/browse/IGNITE-19566


modules/api/src/main/java/org/apache/ignite/lang/util/ExceptionUtils.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
+import org.apache.ignite.lang.TraceableException;
+
+/** Temp class. */
```
ibessonov, line 28: Can you please elaborate? Is there a JIRA issue that would describe the deletion process of this class?

modules/api/src/main/java/org/apache/ignite/sql/SqlBatchException.java
```diff
@@ -30,6 +29,9 @@
  *
  */
 public class SqlBatchException extends SqlException {
+    /** Empty array of long. */
+    public static final long[] LONG_EMPTY_ARRAY = new long[0];
```
ibessonov, line 33: Are you sure it's supposed to be public?

modules/api/src/main/java/org/apache/ignite/table/TupleImpl.java
```diff
@@ -385,10 +384,4 @@ private void readObject(java.io.ObjectInputStream in) throws IOException, ClassN
             colMapping.put(colNames.get(i), i);
         }
     }
-
-    /** {@inheritDoc} */
-    @Override
-    public String toString() {
```
ibessonov, line 391: Should we preserve some version of "toString"? It is possible to format it manually

modules/client/src/main/java/org/apache/ignite/internal/client/ClientClusterNode.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /** {@inheritDoc} */
+    @Override
+    public String name() {
+        return name;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public NetworkAddress address() {
```
ibessonov, line 83: I'm not sure that this method should be in the interface, because we may have detached nodes in the future, as in Ignite 2

modules/client/src/main/java/org/apache/ignite/internal/client/ClientClusterNode.java
```diff
@@ -0,0 +1,120 @@
...
+     *
+     * @param id      Local ID that changes between restarts.
+     * @param name    Unique name of a cluster member.
+     * @param address Node address.
+     */
+    public ClientClusterNode(String id, String name, NetworkAddress address) {
+        this(id, name, address, null);
+    }
+
+    /** {@inheritDoc} */
```
ibessonov, line 69: By the way, we don't have to have these comments

modules/core/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneAlreadyExistsException.java
```diff
@@ -15,12 +15,14 @@
...
 
-package org.apache.ignite.lang;
+package org.apache.ignite.internal.distributionzones;
 
 import static org.apache.ignite.lang.ErrorGroups.DistributionZones.ZONE_ALREADY_EXISTS_ERR;
 
 import java.util.UUID;
+import org.apache.ignite.lang.IgniteInternalException;
 
+// TODO
```
ibessonov, line 25: Please add a link

modules/core/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneBindTableException.java
```diff
@@ -15,13 +15,15 @@
...
-package org.apache.ignite.lang;
+package org.apache.ignite.internal.distributionzones;
 
 import static org.apache.ignite.lang.ErrorGroups.DistributionZones.ZONE_DROP_ERR;
 
 import java.util.UUID;
+import org.apache.ignite.lang.IgniteInternalException;
 import org.jetbrains.annotations.Nullable;
 
+// TODO
```
ibessonov, line 26: Please add a link

modules/core/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneNotFoundException.java
```diff
@@ -15,13 +15,15 @@
...
-package org.apache.ignite.lang;
+package org.apache.ignite.internal.distributionzones;
 
 import static org.apache.ignite.lang.ErrorGroups.DistributionZones.ZONE_NOT_FOUND_ERR;
 
 import java.util.UUID;
+import org.apache.ignite.lang.IgniteInternalException;
 import org.jetbrains.annotations.Nullable;
 
+// TODO
```
ibessonov, line 26: Please add a link

modules/core/src/main/java/org/apache/ignite/lang/IgniteExceptionUtils.java
```diff
@@ -37,7 +37,6 @@
 
 /**
  * Utility class to wrap Ignite exceptions.
- * TODO: https://issues.apache.org/jira/browse/IGNITE-19566 this class should be merged with ExceptionUtils.
```
ibessonov, line 40: Was it merged?

modules/network-api/src/main/java/org/apache/ignite/network/ClusterNodeImpl.java
```diff
@@ -0,0 +1,118 @@
...
+        return result;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public String toString() {
+        return S.toString(ClusterNodeImpl.class, this);
+    }
+}
+
```
ibessonov, line 118: This line is not needed here

modules/api/src/main/java/org/apache/ignite/table/TupleImpl.java
ptupitsyn: Agree, we should have a user-friendly `toString` for Tuple.

ptupitsyn (changes requested):

modules/api/src/main/java/org/apache/ignite/lang/util/ExceptionUtils.java
sk0x50: My fault. This class was renamed to `TraceIdGenerator` and it only provides the following method `getOrCreateTraceId`.
This class allows to extract a trace identifier from an instance of `TraceableException`, or creating a new one.

modules/api/src/main/java/org/apache/ignite/sql/SqlBatchException.java
sk0x50: You are right, it should be private. Fixed

modules/api/src/main/java/org/apache/ignite/table/TupleImpl.java
sk0x50: My goal is to avoid using internal utilities in the `api` module at all (perhaps, this idea is pure purism that does not bring valuable benefits, but anyway...).

`IgniteToStringBuilder.toString` is an internal utility that is based on some environment variables to hide sensitive information. I think, that this behavior should be controlled via configuration instead of environment variables, so it will be hard to move this utility into the `api` module in the future. However, I agree that we should have a user-friendly method to print tuples.

I see the following alternatives:
 - move `IgniteToStringBuilder` utility to the `api` module (including annotations like `IgniteToStringExclude`, `IgniteToStringInclude`, `IgniteToStringOrder`, and so on). The behavior is still controlled by env variables (perhaps, it is possible to offer a wiser approach).
 - add simple implementation `toString` to the `TupleImpl` class and improve `IgniteToStringBuilder` implementation to handle such cases taking into account sensitive information.
 - move `TupleImpl` implementation into an internal package and provide new methods to create instances via `Ignite` instance.

For now, I would add a simple implementation `toString` to the `TupleImpl` and after that, it would be nice to discuss the alternatives.

modules/client/src/main/java/org/apache/ignite/internal/client/ClientClusterNode.java
sk0x50: I would not change the `ClusterNode` interface in this pull request. It should be a separate task that can be done in the future when all requirements will be clear.

modules/client/src/main/java/org/apache/ignite/internal/client/ClientClusterNode.java
sk0x50: Yep, you are right. In accordance with https://google.github.io/styleguide/javaguide.html#s7.3.2-javadoc-exception-overrides there is no special need to explicitly mark a method with `inheritDoc` tag. Will remove redundant javadocs.

modules/network-api/src/main/java/org/apache/ignite/network/ClusterNodeImpl.java
sk0x50: Thanks. Fixed

modules/client/src/main/java/org/apache/ignite/internal/client/ClientClusterNode.java
```diff
@@ -0,0 +1,120 @@
...
+        return name.equals(that.name) && address.equals(that.address);
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public int hashCode() {
+        int result = name.hashCode();
+        result = 31 * result + address.hashCode();
+        return result;
+    }
```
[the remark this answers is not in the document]
sk0x50: It looks like these methods are not needed. You are right. Removed.

modules/core/src/main/java/org/apache/ignite/lang/IgniteExceptionUtils.java
sk0x50: Thanks. Merged

modules/core/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneAlreadyExistsException.java
sk0x50: Done

modules/core/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneBindTableException.java
sk0x50: Done

modules/core/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneNotFoundException.java
sk0x50: Done

modules/client/src/main/java/org/apache/ignite/internal/client/ClientClusterNode.java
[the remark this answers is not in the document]
sk0x50: Well, I was wrong. `ClusterNode` is used as a key of `Map`. For example, `IgniteCompute.broadcast` method. So, `equals` and `hashCode` cannot not be removed.

modules/client/src/main/java/org/apache/ignite/internal/client/ClientClusterNode.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.ignite.internal.tostring.S;
+import org.apache.ignite.network.ClusterNode;
+import org.apache.ignite.network.NetworkAddress;
+import org.apache.ignite.network.NodeMetadata;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Client representation of a node in a cluster.
+ */
+public class ClientClusterNode implements ClusterNode {
```
ptupitsyn, line 29: `ClientClusterNode` fully duplicates `ClusterNodeImpl`, there is no difference at all. We could move it to `core` module. But this is not a big deal, let's keep it for now.

ptupitsyn (approved):
Client-related changes look good to me.

modules/core/src/main/java/org/apache/ignite/internal/util/ExceptionUtils.java
```diff
@@ -443,28 +452,499 @@ private static <T extends Exception> T withCauseInternal(
...
+     *
+     *     CompletionException c = new CompletionException(err);
+     *
+     *     // The following line prints: IGN_SPECIFIC_ERR_CODE TraceId XYZ wrong argument
+     *     c.getMessage();
+     *
+     *     CustomException copy = createCopyWithCause(c);
+     *
+     *     // The following line prints: CustomException IGN_SPECIFIC_ERR_CODE TraceId XYZ wrong argument
+     *     // but it should be just as follows: XYZ wrong argument
```
korlov42, line 937:
```suggestion
     *     // but it should be just as follows: wrong argument
```

korlov42 (approved):

modules/api/src/main/java/org/apache/ignite/lang/util/TraceIdGenerator.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.IdentityHashMap;
+import java.util.Set;
+import java.util.UUID;
+import org.apache.ignite.lang.TraceableException;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * This utility class allows to extract trace identifier from an instance of {@link TraceableException} or creating a new trace identifier.
+ **/
+public class TraceIdGenerator {
```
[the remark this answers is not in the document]
sk0x50: Yep, it makes sense to me. Will rename it.

PR #2350, opened 2023-07-24
IGNITE-20036 Fix resource leaks in client tests and streamer

sashapolo:
* Use `IgniteUtils.closeAll` to close multiple resources reliably
* Use try-with-resources for `SubmissionPublisher` in all tests
* Use `scheduleAtFixedRate` instead of `Timer` in client streamer ([recommended approach for modern code](https://stackoverflow.com/questions/409932/java-timer-vs-executorservice))

https://issues.apache.org/jira/browse/IGNITE-20036

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/streamer/StreamerSubscriber.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/streamer/StreamerSubscriber.java
@@ -19,15 +19,19 @@ package org.apache.ignite.internal.streamer;
 
 import java.util.Collection;
 import java.util.Set;
-import java.util.Timer;
-import java.util.TimerTask;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
 import java.util.concurrent.Flow;
 import java.util.concurrent.Flow.Subscriber;
 import java.util.concurrent.Flow.Subscription;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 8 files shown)


modules/core/src/main/java/org/apache/ignite/internal/streamer/StreamerSubscriber.java
```diff
@@ -241,18 +251,18 @@ private void requestMore() {
...
         }
 
-        Timer timer = new Timer("client-data-streamer-flush-" + hashCode());
+        String threadPrefix = "client-data-streamer-flush-" + hashCode();
 
-        timer.schedule(new PeriodicFlushTask(), interval, interval);
+        flushTimer = Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory(threadPrefix, log));
 
-        return timer;
+        flushTask = flushTimer.scheduleAtFixedRate(new PeriodicFlushTask(), interval, interval, TimeUnit.MILLISECONDS);
```
ptupitsyn, line 265: 1. Can you please explain this change?
2. PR title says "in client tests", but this code is not test

ptupitsyn (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/streamer/StreamerSubscriber.java
sashapolo: > PR title says "in client tests", but this code is not test

Yes, but client tests were affected, because the timer sometimes didn't stop because of a race between `requestMore` and timer initialization. That's why this thread was sometimes showing up in my debugging sessions.

> Can you please explain this change?

This particular change is not related to the original problem, `Timer` is simply an older API and it is recommended to use `SheduledExecutorService` instead. I got lazy and didn't want to create a separate PR just for this.

ptupitsyn (approved):

PR #2356, opened 2023-07-25
IGNITE-19501 SchemaManager should use CatalogService for building BinaryRow descriptors

rpuch:
https://issues.apache.org/jira/browse/IGNITE-19501

### DIFF
--- a/modules/schema/src/main/java/org/apache/ignite/internal/schema/CatalogSchemaManager.java
+++ b/modules/schema/src/main/java/org/apache/ignite/internal/schema/CatalogSchemaManager.java
@@ -177,7 +177,8 @@ public class CatalogSchemaManager extends Producer<SchemaEvent, SchemaEventParam
                     );
                 }
 
-                return registerSchema(registries, tableId, newSchema);
+                return saveSchema(tableId, newSchema)
+                        .thenApply(t -> registerSchema(registries, tableId, newSchema));
             })).thenApply(ignored -> false);
         } finally {
             busyLock.leaveBusy();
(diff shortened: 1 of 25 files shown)


modules/schema/src/test/java/org/apache/ignite/internal/schema/CatalogSchemaManagerTest.java
```diff
@@ -0,0 +1,377 @@
...
+    void schemaRegistryByCausalityTokenIsAvailable() {
+        createSomeTable();
+
+        CompletableFuture<SchemaRegistry> future = schemaManager.schemaRegistry(CAUSALITY_TOKEN_1, TABLE_ID);
+        assertThat(future, willCompleteSuccessfully());
+
+        SchemaRegistry schemaRegistry = future.join();
+
+        assertThat(schemaRegistry.schema().version(), is(1));
+        assertThat(schemaRegistry.schema(1).version(), is(1));
```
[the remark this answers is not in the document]
rpuch: I added such a test. Please take a look if this is what you wanted

modules/schema/src/main/java/org/apache/ignite/internal/schema/CatalogSchemaManager.java
```diff
@@ -0,0 +1,487 @@
...
+            Map<Integer, SchemaRegistryImpl> registries,
+            int tableId,
+            SchemaDescriptor schema
+    ) {
+        ByteArray key = schemaWithVerHistKey(tableId, schema.version());
+
+        byte[] serializedSchema = SchemaSerializerImpl.INSTANCE.serialize(schema);
+
+        return metastorageMgr.invoke(notExists(key), put(key, serializedSchema), noop())
+                .thenApply(t -> {
```
AMashenkov, line 224: Let's move this to the outside, and rename method to smth like `saveSchema`.
Thus, the method will represent a single action, its name will corresponds the content, and will `registries` changes will be localized.

modules/schema/src/main/java/org/apache/ignite/internal/schema/CatalogSchemaManager.java
```diff
@@ -0,0 +1,487 @@
...
+        return latestVersionFuture;
+    }
+
+    /**
+     * Gets the defined version of the table schema which available in Metastore.
+     *
+     * @param tblId Table id.
+     * @return Schema representation if schema found, {@code null} otherwise.
+     */
+    private CompletableFuture<SchemaDescriptor> schemaByVersion(int tblId, int ver) {
```
AMashenkov, line 448:
```suggestion
    private CompletableFuture<SchemaDescriptor> loadSchema(int tblId, int ver) {
```

modules/schema/src/main/java/org/apache/ignite/internal/schema/CatalogSchemaManager.java
```diff
@@ -0,0 +1,487 @@
...
+            }
+        });
+
+        return latestVersionFuture;
+    }
+
+    /**
+     * Gets the defined version of the table schema which available in Metastore.
+     *
+     * @param tblId Table id.
```
AMashenkov, line 445:
```suggestion
     * Loads the table schema descriptor by version from Metastore.
     *
     * @param tblId Table id.
     * @param ver Schema version.
```

modules/schema/src/main/java/org/apache/ignite/internal/schema/CatalogSchemaManager.java
```diff
@@ -0,0 +1,487 @@
...
+
+                SchemaDescriptor desc = registry0.schemaCached(schemaVer);
+
+                assert desc != null : "Unexpected empty schema description.";
+
+                fut.complete(desc);
+            }
+        };
+
+        registriesVv.whenComplete(schemaListener);
```
AMashenkov, line 284: The current method is called only from inside the SchemaRegistryImpl, and can be moved into the registry class. This listener can be replaced with VersionTracker to wait for the next versions (like CatalogService.versionTracker).

Thus, code will be much simple. RegistersVv calls will gone as registry is already available.
Joggling with subscriptions will be replaced with versionTracker future.

modules/schema/src/main/java/org/apache/ignite/internal/schema/CatalogSchemaManager.java
```diff
@@ -0,0 +1,487 @@
...
+    /**
+     * Gets the latest version of the table schema which is available in Metastore.
+     *
+     * @param tblId Table id.
+     * @return The latest schema version.
+     */
+    private CompletableFuture<Integer> latestSchemaVersion(int tblId) {
+        var latestVersionFuture = new CompletableFuture<Integer>();
+
+        metastorageMgr.prefix(schemaHistPrefix(tblId)).subscribe(new Subscriber<>() {
```
AMashenkov, line 409: I think we can update 2 keys in single transaction
* write a new descriptor to the history
* update latest version key with a version from written descriptor.

modules/schema/src/main/java/org/apache/ignite/internal/schema/CatalogSchemaManager.java
```diff
@@ -0,0 +1,487 @@
...
+        return registriesVv.latest().get(tableId);
+    }
+
+    /**
+     * Drop schema registry for the given table id.
+     *
+     * @param causalityToken Causality token.
+     * @param tableId Table id.
+     */
+    public CompletableFuture<?> dropRegistry(long causalityToken, int tableId) {
```
AMashenkov, line 376: Seems, we also should drop history from metastorage for the given table along with its registry.

modules/schema/src/main/java/org/apache/ignite/internal/schema/CatalogSchemaManager.java
rpuch: Fixed in https://issues.apache.org/jira/browse/IGNITE-20071

AMashenkov (approved):

PR #2357, opened 2023-07-25
IGNITE-19291 Generate default node configuration file at compile-time

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-19291


modules/configuration/src/main/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorage.java
ibessonov, line 1: Why do we move it to the configuration module? Don't we have `runner` in classpath? I thought that this module should have basic abstract entities, not specific implementations

ibessonov (approved):

PR #2358, opened 2023-07-25
IGNITE-20045 Use classes from Catalog domain in assignments recalculation code

rpuch:
https://issues.apache.org/jira/browse/IGNITE-20045

### DIFF
--- a/modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/ZoneCatalogDescriptorUtils.java
+++ b/modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/ZoneCatalogDescriptorUtils.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.distributionzones.rebalance;
+
+import java.lang.reflect.Method;
+import org.apache.ignite.internal.catalog.descriptors.CatalogDataStorageDescriptor;
+import org.apache.ignite.internal.catalog.descriptors.CatalogZoneDescriptor;
+import org.apache.ignite.internal.distributionzones.configuration.DistributionZoneView;
+import org.apache.ignite.internal.schema.configuration.storage.DataStorageView;
+
+/**
+ * Utils used to convert configuration representations of distribution zones to catalog representations.
+ */
+public class ZoneCatalogDescriptorUtils {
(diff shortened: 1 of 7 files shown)


modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/rebalance/DistributionZoneRebalanceEngineTest.java
```diff
@@ -443,6 +451,23 @@ void replicasTriggersAssignmentsChangingOnDefaultZone(
         }
     }
 
+    private static void completeTablesConfigs(TablesConfiguration tablesConfiguration) {
```
sashapolo, line 454: Why is this method needed?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/ZoneCatalogDescriptorUtils.java
```diff
@@ -0,0 +1,67 @@
...
+ * Utils used to convert configuration representations of distribution zones to catalog representations.
+ */
+public class ZoneCatalogDescriptorUtils {
+    // TODO: IGNITE-19719 Fix it
+    /**
+     * Converts a distribution zone configuration to a Distribution zone descriptor.
+     *
+     * @param config Distribution zone configuration.
+     */
+    @Deprecated(forRemoval = true)
```
sashapolo, line 36: Looks like the whole class should be deprecated and removed in the future?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/ZoneCatalogDescriptorUtils.java
```diff
@@ -0,0 +1,67 @@
...
+                toDataStorageDescriptor(config.dataStorage())
+        );
+    }
+
+    @Deprecated(forRemoval = true)
+    private static CatalogDataStorageDescriptor toDataStorageDescriptor(DataStorageView config) {
+        String dataRegion;
+
+        try {
+            Method dataRegionMethod = config.getClass().getMethod("dataRegion");
```
sashapolo, line 56: oh god...

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/rebalance/DistributionZoneRebalanceEngineTest.java
rpuch: The code that processes rebalance converts `TableView` to `CatalogTableDescriptor` and fails if there is no primary key or columns, so it is a temporary ceremony that will go away when we switch to the Catalog completely

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/ZoneCatalogDescriptorUtils.java
rpuch: I agree!

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/ZoneCatalogDescriptorUtils.java
rpuch: Deprecated the class too

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/rebalance/DistributionZoneRebalanceEngineTest.java
sashapolo: can it be provided as part of the initial configuration inside `@InjectConfiguration`?

sashapolo (approved):

PR #2359, opened 2023-07-25
IGNITE-19108 Check connection during REPL session in CLI

nevdmitry:
https://issues.apache.org/jira/browse/IGNITE-19108

- start connection heartbeat thread on 'connect' command
- stop heartbeat on 'disconnect'

### DIFF
--- a/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/ConnectionHeartBeat.java
+++ b/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/ConnectionHeartBeat.java
@@ -27,7 +27,7 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import javax.annotation.Nullable;
 import org.apache.ignite.internal.cli.core.rest.ApiClientFactory;
-import org.apache.ignite.internal.cli.event.AsyncConnectionEventListener;
+import org.apache.ignite.internal.cli.event.ConnectionEventListener;
 import org.apache.ignite.internal.cli.event.EventPublisher;
 import org.apache.ignite.internal.cli.event.Events;
 import org.apache.ignite.internal.cli.logger.CliLoggers;
@@ -40,7 +40,7 @@ import org.apache.ignite.rest.client.invoker.ApiException;
  * Connection to node heart beat.
  */
 @Singleton
-public class ConnectionHeartBeat extends AsyncConnectionEventListener {
+public class ConnectionHeartBeat implements ConnectionEventListener {
 
     private static final IgniteLogger LOG = CliLoggers.forClass(ConnectionHeartBeat.class);
 
@@ -79,7 +79,7 @@ public class ConnectionHeartBeat extends AsyncConnectionEventListener {
      * @param sessionInfo session info with node url
      */
     @Override
-    protected void onConnect(SessionInfo sessionInfo) {
+    public void onConnect(SessionInfo sessionInfo) {
         if (connected.compareAndSet(false, true)) {
             eventPublisher.publish(Events.connectionRestored());
         }
@@ -107,7 +107,7 @@ public class ConnectionHeartBeat extends AsyncConnectionEventListener {
      * Stops connection heartbeat.
      */
     @Override
-    protected void onDisconnect() {
+    public void onDisconnect() {
         lock.lock();
         try {
             if (scheduledConnectionHeartbeatExecutor != null) {
--- a/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/PeriodicSessionTaskExecutor.java
+++ b/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/PeriodicSessionTaskExecutor.java
@@ -24,7 +24,7 @@ import java.util.List;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
-import org.apache.ignite.internal.cli.event.AsyncConnectionEventListener;
+import org.apache.ignite.internal.cli.event.ConnectionEventListener;
 import org.apache.ignite.internal.cli.logger.CliLoggers;
 import org.apache.ignite.internal.logger.IgniteLogger;
 import org.apache.ignite.internal.thread.NamedThreadFactory;
@@ -32,7 +32,7 @@ import org.jetbrains.annotations.Nullable;
 
 /** Executes tasks periodically while the session is connected. */
 @Singleton
-public class PeriodicSessionTaskExecutor extends AsyncConnectionEventListener {
+public class PeriodicSessionTaskExecutor implements ConnectionEventListener {
     private static final IgniteLogger LOG = CliLoggers.forClass(PeriodicSessionTaskExecutor.class);
 
     @Nullable
--- a/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/Session.java
+++ b/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/Session.java
@@ -20,13 +20,13 @@ package org.apache.ignite.internal.cli.core.repl;
 import jakarta.inject.Singleton;
 import java.util.concurrent.atomic.AtomicReference;
 import org.apache.ignite.internal.cli.core.exception.ConnectionException;
-import org.apache.ignite.internal.cli.event.AsyncConnectionEventListener;
+import org.apache.ignite.internal.cli.event.ConnectionEventListener;
 
 /**
  * Connection session that in fact is holder for state: connected or disconnected. Also has session info if the state is connected.
  */
 @Singleton
-public class Session extends AsyncConnectionEventListener {
+public class Session implements ConnectionEventListener {
 
     private final AtomicReference<SessionInfo> info = new AtomicReference<>();
 
@@ -34,14 +34,14 @@ public class Session extends AsyncConnectionEventListener {
     }
 
     @Override
-    protected void onConnect(SessionInfo sessionInfo) {
+    public void onConnect(SessionInfo sessionInfo) {
         if (!info.compareAndSet(null, sessionInfo)) {
             throw new ConnectionException("Already connected to " + info.get().nodeUrl());
         }
     }
 
     @Override
-    protected void onDisconnect() {
+    public void onDisconnect() {
         info.getAndSet(null);
     }
 
--- a/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/prompt/ReplPromptProvider.java
+++ b/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/prompt/ReplPromptProvider.java
@@ -25,13 +25,13 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.ignite.internal.cli.core.repl.Session;
 import org.apache.ignite.internal.cli.core.repl.SessionInfo;
 import org.apache.ignite.internal.cli.core.style.AnsiStringSupport.Color;
-import org.apache.ignite.internal.cli.event.AsyncConnectionEventListener;
+import org.apache.ignite.internal.cli.event.ConnectionEventListener;
 
 /**
  * Provider for prompt in REPL.
  */
 @Singleton
-public class ReplPromptProvider extends AsyncConnectionEventListener implements PromptProvider {
+public class ReplPromptProvider implements PromptProvider, ConnectionEventListener {
     private final Session session;
 
     private final AtomicBoolean connected = new AtomicBoolean(true);
@@ -61,12 +61,12 @@ public class ReplPromptProvider extends AsyncConnectionEventListener implements 
     }
 
     @Override
-    protected void onConnectionLost() {
+    public void onConnectionLost() {
         connected.getAndSet(false);
     }
 
     @Override
-    protected void onConnectionRestored() {
+    public void onConnectionRestored() {
         connected.getAndSet(true);
     }
 }
--- a/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/registry/impl/ClusterConfigRegistryImpl.java
+++ b/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/registry/impl/ClusterConfigRegistryImpl.java
@@ -26,12 +26,12 @@ import org.apache.ignite.internal.cli.call.configuration.JsonString;
 import org.apache.ignite.internal.cli.core.call.DefaultCallOutput;
 import org.apache.ignite.internal.cli.core.repl.SessionInfo;
 import org.apache.ignite.internal.cli.core.repl.registry.ClusterConfigRegistry;
-import org.apache.ignite.internal.cli.event.AsyncConnectionEventListener;
+import org.apache.ignite.internal.cli.event.ConnectionEventListener;
 import org.jetbrains.annotations.Nullable;
 
 /** Implementation of {@link ClusterConfigRegistry}. */
 @Singleton
-public class ClusterConfigRegistryImpl extends AsyncConnectionEventListener implements ClusterConfigRegistry {
+public class ClusterConfigRegistryImpl implements ClusterConfigRegistry, ConnectionEventListener {
 
     private final ClusterConfigShowCall clusterConfigShowCall;
 
--- a/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/registry/impl/MetricRegistryImpl.java
+++ b/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/registry/impl/MetricRegistryImpl.java
@@ -27,13 +27,13 @@ import org.apache.ignite.internal.cli.core.call.CallOutput;
 import org.apache.ignite.internal.cli.core.call.UrlCallInput;
 import org.apache.ignite.internal.cli.core.repl.SessionInfo;
 import org.apache.ignite.internal.cli.core.repl.registry.MetricRegistry;
-import org.apache.ignite.internal.cli.event.AsyncConnectionEventListener;
+import org.apache.ignite.internal.cli.event.ConnectionEventListener;
 import org.apache.ignite.rest.client.model.MetricSource;
 import org.jetbrains.annotations.Nullable;
 
 /** Implementation of {@link MetricRegistry}. */
 @Singleton
-public class MetricRegistryImpl extends AsyncConnectionEventListener implements MetricRegistry {
+public class MetricRegistryImpl implements MetricRegistry, ConnectionEventListener {
 
     @Inject
     private NodeMetricSourceListCall metricSourceListCall;
--- a/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/registry/impl/NodeConfigRegistryImpl.java
+++ b/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/registry/impl/NodeConfigRegistryImpl.java
@@ -24,11 +24,11 @@ import org.apache.ignite.internal.cli.call.configuration.NodeConfigShowCall;
 import org.apache.ignite.internal.cli.call.configuration.NodeConfigShowCallInput;
 import org.apache.ignite.internal.cli.core.repl.SessionInfo;
 import org.apache.ignite.internal.cli.core.repl.registry.NodeConfigRegistry;
-import org.apache.ignite.internal.cli.event.AsyncConnectionEventListener;
+import org.apache.ignite.internal.cli.event.ConnectionEventListener;
 
 /** Implementation of {@link NodeConfigRegistry}. */
 @Singleton
-public class NodeConfigRegistryImpl extends AsyncConnectionEventListener implements NodeConfigRegistry {
+public class NodeConfigRegistryImpl implements NodeConfigRegistry, ConnectionEventListener {
 
     private final NodeConfigShowCall nodeConfigShowCall;
 
--- a/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/registry/impl/UnitsRegistryImpl.java
+++ b/modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/registry/impl/UnitsRegistryImpl.java
@@ -28,13 +28,13 @@ import org.apache.ignite.internal.cli.call.unit.ListUnitCallInput;
 import org.apache.ignite.internal.cli.core.call.CallOutput;
 import org.apache.ignite.internal.cli.core.repl.SessionInfo;
 import org.apache.ignite.internal.cli.core.repl.registry.UnitsRegistry;
-import org.apache.ignite.internal.cli.event.AsyncConnectionEventListener;
+import org.apache.ignite.internal.cli.event.ConnectionEventListener;
 import org.apache.ignite.rest.client.model.UnitStatus;
 import org.apache.ignite.rest.client.model.UnitVersionStatus;
 
 /** Implementation of {@link UnitsRegistry}. */
 @Singleton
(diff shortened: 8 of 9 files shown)


modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/ItConnectionStatusTest.java
```diff
@@ -0,0 +1,67 @@
...
+        execute("connect");
+
+        // Then
+        assertAll(
+                this::assertErrOutputIsEmpty,
+                () -> assertOutputContains("Connected to http://localhost:10300")
+        );
+
+        // And prompt is changed to green (connection status OPEN)
+        assertEquals(ConnectionStatus.OPEN, session.info().connectionStatus());
```
PakhomovAlexander, line 61: If you check `ConnectionStatus` after an action it would be nice to check `ConnectionStatus` before, not the prompt.

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/core/repl/executor/ItIgnitePicocliCommandsTest.java
```diff
@@ -245,7 +246,7 @@ void nodeConfigShowSuggested(ParsedLine givenParsedLine) {
     }
 
     private void connected() {
-        session.connect(new SessionInfo(DEFAULT_REST_URL, null, null, null));
+        session.connect(new SessionInfo(DEFAULT_REST_URL, null, null, null, ConnectionStatus.OPEN));
```
PakhomovAlexander, line 249: That is getting weird. Could you please get rid of a constructor in SessionInfo and create a builder?

modules/cli/src/main/java/org/apache/ignite/internal/cli/call/connect/ConnectCall.java
```diff
@@ -81,8 +86,8 @@ public CallOutput<String> execute(UrlCallInput input) {
             stateConfigProvider.get().setProperty(CliConfigKeys.LAST_CONNECTED_URL.value(), nodeUrl);
 
             String jdbcUrl = jdbcUrlFactory.constructJdbcUrl(configuration, nodeUrl);
-            session.connect(new SessionInfo(nodeUrl, fetchNodeName(nodeUrl), jdbcUrl, username));
-
+            session.connect(new SessionInfo(nodeUrl, fetchNodeName(nodeUrl), jdbcUrl, username, ConnectionStatus.OPEN));
+            connectionHeartBeat.start(session, clientFactory);
```
PakhomovAlexander, line 90: We have `AsyncSessionEventListener`, I think you can use it for connect/disconnect logic.

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/ConnectionHeartBeat.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.ignite.rest.client.api.NodeManagementApi;
+import org.apache.ignite.rest.client.invoker.ApiException;
+
+/**
+ * Connection to node heart beat.
+ */
+@Singleton
+public class ConnectionHeartBeat {
+
+    private static final IgniteLogger log = CliLoggers.forClass(Session.class);
```
PakhomovAlexander, line 38:
```suggestion
    private static final IgniteLogger log = CliLoggers.forClass(ConnectionHeartBeat.class);
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/ConnectionHeartBeat.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /** CLI check connection period period. */
+    private static final int CLI_CHECK_CONNECTION_PERIOD_SECONDS = 10;
```
PakhomovAlexander, line 41: 10 seconds is too long a time period. The peak time user might see the wrong state is (period + period-1) = 19 sec. I suggest having at lead 5 sec.

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/ConnectionHeartBeat.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public void stop() {
+        scheduledIdleSafeTimeSyncExecutor.shutdownNow();
+    }
+
+    private static void pingConnection(Session session, ApiClientFactory clientFactory) {
+        try {
+            new NodeManagementApi(clientFactory.getClient(session.info().nodeUrl())).nodeState();
+            SessionInfo currentSessionInfo = session.info();
+            session.updateSessionInfo(new SessionInfo(
```
PakhomovAlexander, line 70: That is not a thread-safe solution. You can get `session.info()` and get it updated by disconnect/connect or some other things. I mean if you read some state, do action, and update state – this should be thread safe.

PakhomovAlexander (changes requested):

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/ItConnectionHeartbeatTest.java
```diff
@@ -0,0 +1,102 @@
...
+    @BeforeEach
+    void setUp() {
+        //Set connection check timeout to 1 sec to make test fast
+        ConnectionHeartBeat.CLI_CHECK_CONNECTION_PERIOD_SECONDS = 1L;
+    }
+
+    @AfterEach
+    void teardown() {
+        // Start all nodes
+        allNodeNames().forEach(this::startNode);
```
PakhomovAlexander, line 57: I don't think you want to start all nodes on the test tear-down.

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/ItConnectionHeartbeatTest.java
```diff
@@ -0,0 +1,102 @@
...
+    Session session;
+
+    @Spy
+    @Inject
+    private ReplPromptProvider replPromptProvider;
+
+    @BeforeEach
+    void setUp() {
+        //Set connection check timeout to 1 sec to make test fast
+        ConnectionHeartBeat.CLI_CHECK_CONNECTION_PERIOD_SECONDS = 1L;
```
PakhomovAlexander, line 51: We tend not to use global static mutable variables (only in extreme cases). I would ask you to put the configuration value into `ConnectionHeartBeat` via a constructor. In test, you can create the instance directly or using Micronaut's configuration property.

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/ItConnectionHeartbeatTest.java
```diff
@@ -0,0 +1,102 @@
...
+        // Then
+        assertAll(
+                this::assertErrOutputIsEmpty,
+                () -> assertOutputContains("Connected to http://localhost:10300")
+        );
+
+        // When stop all nodes
+        allNodeNames().forEach(IgnitionManager::stop);
+
+        verify(replPromptProvider, after(ConnectionHeartBeat.CLI_CHECK_CONNECTION_PERIOD_SECONDS * 2L).times(1)).onConnectionLost();
```
PakhomovAlexander, line 99: It would be safer to use the `Awaitility` library here and above. Something can go wrong on CI and we'll get one more flaky test.

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/ConnectionHeartBeat.java
```diff
@@ -0,0 +1,126 @@
...
+                    CLI_CHECK_CONNECTION_PERIOD_SECONDS,
+                    TimeUnit.SECONDS
+            );
+        }
+    }
+
+    /**
+     * Stops connection heartbeat.
+     */
+    public void stop() {
```
PakhomovAlexander, line 85: Can you make start/stop methods thread safe? I think synchronized will be ok.

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/ItConnectionHeartbeatTest.java
```diff
@@ -0,0 +1,147 @@
...
+    private final AtomicInteger connectionLost = new AtomicInteger(0);
+    private final AtomicInteger connectionRestored = new AtomicInteger(0);
+
+    @BeforeEach
+    void setUp() {
+        //ToDo: Set connection check timeout to 1 sec to make test fast
+        connectionLost.set(0);
+        connectionRestored.set(0);
+        EventListener eventListener = (eventType, event) -> {
+            if (EventType.CONNECTION_LOST == eventType) {
```
PakhomovAlexander, line 59: It is better not to use if statement but register two separate event listeners

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/ItConnectionHeartbeatTest.java
```diff
@@ -0,0 +1,147 @@
...
+class ItConnectionHeartbeatTest extends CliCommandTestInitializedIntegrationBase {
+
+    @Inject
+    Session session;
+
+    @Inject
+    EventFactory eventFactory;
+
+    @Value("${cli.check.connection.period.second}")
+    private long CLI_CHECK_CONNECTION_PERIOD_SECONDS;
```
PakhomovAlexander:
```suggestion
    private final long CLI_CHECK_CONNECTION_PERIOD_SECONDS = 1;
```

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/ItConnectionHeartbeatTest.java
```diff
@@ -0,0 +1,147 @@
...
+        assertAll(
+                this::assertErrOutputIsEmpty,
+                () -> assertOutputContains("Connected to http://localhost:10300")
+        );
+
+        // When stop node
+        String nodeName = session.info().nodeName();
+        this.stopNode(nodeName);
+
+        //Then
```
PakhomovAlexander, line 136:
```suggestion
        // Then
```

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/ItConnectionHeartbeatTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+        // When stop node
+        String nodeName = session.info().nodeName();
+        this.stopNode(nodeName);
+
+        //Listener was invoked
+        await().timeout(CLI_CHECK_CONNECTION_PERIOD_SECONDS * 2, TimeUnit.SECONDS).until(() -> connectionRestored.get() == 1);
+        await().timeout(CLI_CHECK_CONNECTION_PERIOD_SECONDS * 2, TimeUnit.SECONDS).until(() -> connectionLost.get() == 1);
+
+        //Tear down
```
PakhomovAlexander, line 116:
```suggestion
        // Tear down
```

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/ItConnectionHeartbeatTest.java
```diff
@@ -0,0 +1,147 @@
...
+        assertAll(
+                this::assertErrOutputIsEmpty,
+                () -> assertOutputContains("Connected to http://localhost:10300")
+        );
+
+        // When stop node
+        String nodeName = session.info().nodeName();
+        this.stopNode(nodeName);
+
+        //Listener was invoked
```
PakhomovAlexander, line 112:
```suggestion
        // Listener was invoked
```

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/ItConnectionHeartbeatTest.java
```diff
@@ -0,0 +1,147 @@
...
+        await().timeout(CLI_CHECK_CONNECTION_PERIOD_SECONDS * 2, TimeUnit.SECONDS).until(() -> connectionRestored.get() == 1);
+        await().timeout(CLI_CHECK_CONNECTION_PERIOD_SECONDS * 2, TimeUnit.SECONDS).until(() -> connectionLost.get() == 1);
+
+        //Tear down
+        this.startNode(nodeName);
```
PakhomovAlexander, line 117: I think stop should be invoked instead.

PakhomovAlexander: As far as I see it is already stopped.

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/ItConnectionHeartbeatTest.java
```diff
@@ -0,0 +1,147 @@
...
+        this.stopNode(nodeName);
+
+        //Then
+        await().timeout(CLI_CHECK_CONNECTION_PERIOD_SECONDS * 2, TimeUnit.SECONDS).until(() -> connectionRestored.get() == 1);
+        await().timeout(CLI_CHECK_CONNECTION_PERIOD_SECONDS * 2, TimeUnit.SECONDS).until(() -> connectionLost.get() == 1);
+
+        // When
+        this.startNode(nodeName);
+
+        //Then
```
PakhomovAlexander, line 143:
```suggestion
        // Then
```

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/ItConnectionHeartbeatTest.java
```diff
@@ -0,0 +1,147 @@
...
+        await().timeout(CLI_CHECK_CONNECTION_PERIOD_SECONDS * 2, TimeUnit.SECONDS).until(() -> connectionRestored.get() == 2);
+        await().timeout(CLI_CHECK_CONNECTION_PERIOD_SECONDS * 2, TimeUnit.SECONDS).until(() -> connectionLost.get() == 1);
```
PakhomovAlexander: We've already waited for this condition.

modules/cli/src/main/java/org/apache/ignite/internal/cli/call/connect/ConnectCall.java
```diff
@@ -56,15 +61,18 @@ public class ConnectCall implements Call<UrlCallInput, String> {
...
     private final JdbcUrlFactory jdbcUrlFactory;
 
+    private final EventFactory eventFactory;
+
     /**
      * Constructor.
      */
     public ConnectCall(Session session, StateConfigProvider stateConfigProvider, ApiClientFactory clientFactory,
-            JdbcUrlFactory jdbcUrlFactory) {
+            JdbcUrlFactory jdbcUrlFactory, ConnectionHeartBeat connectionHeartBeat, EventFactory eventFactory) {
```
PakhomovAlexander, line 70:
```suggestion
            JdbcUrlFactory jdbcUrlFactory, EventFactory eventFactory) {
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/call/connect/DisconnectCall.java
```diff
@@ -17,35 +17,40 @@
...
  */
 @Singleton
 public class DisconnectCall implements Call<EmptyCallInput, String> {
-    @Inject
     private final Session session;
 
-    public DisconnectCall(Session session) {
+    private final EventFactory eventFactory;
+
+    public DisconnectCall(Session session, ConnectionHeartBeat connectionHeartBeat, EventFactory eventFactory) {
```
PakhomovAlexander, line 43:
```suggestion
    public DisconnectCall(Session session, EventFactory eventFactory) {
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/ConnectionHeartBeat.java
```diff
@@ -0,0 +1,129 @@
...
+    private ScheduledExecutorService scheduledConnectionHeartbeatExecutor;
+
+    private final ApiClientFactory clientFactory;
+
+    private final EventFactory eventFactory;
+
+    private final AtomicBoolean connected = new AtomicBoolean(false);
+
+    /**
+     * Created instance of connection heartbeat.
```
PakhomovAlexander, line 60:
```suggestion
     * Creates the instance of connection heartbeat.
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/ConnectionHeartBeat.java
```diff
@@ -0,0 +1,129 @@
...
+     * @param sessionInfo session info with node url
+     */
+    private void onConnect(SessionInfo sessionInfo) {
+        //eventFactory.fireEvent(EventType.CONNECTION_RESTORED, new ConnectionStatusEvent());
+
+        if (scheduledConnectionHeartbeatExecutor == null) {
+            scheduledConnectionHeartbeatExecutor =
+                    Executors.newScheduledThreadPool(1, new NamedThreadFactory("cli-check-connection-thread", log));
+
+            //Start connection heart beat
```
PakhomovAlexander, line 85:
```suggestion
            // Start connection heart beat.
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/ConnectionHeartBeat.java
```diff
@@ -0,0 +1,129 @@
...
+        this.cliCheckConnectionPeriodSecond = cliCheckConnectionPeriodSecond;
+    }
+
+    /**
+     * Starts connection heartbeat. By default connection will be checked every 5 sec.
+     *
+     * @param sessionInfo session info with node url
+     */
+    private void onConnect(SessionInfo sessionInfo) {
+        //eventFactory.fireEvent(EventType.CONNECTION_RESTORED, new ConnectionStatusEvent());
```
PakhomovAlexander, line 79: commented code

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/PeriodicSessionTaskExecutor.java
```diff
@@ -24,35 +24,48 @@
...
-public class PeriodicSessionTaskExecutor implements AsyncSessionEventListener {
+public class PeriodicSessionTaskExecutor implements EventListener {
     private static final IgniteLogger LOG = CliLoggers.forClass(PeriodicSessionTaskExecutor.class);
 
     @Nullable
     private ScheduledExecutorService executor;
 
     private final List<? extends PeriodicSessionTask> tasks;
 
+    //private final EventFactory eventFactory;
```
PakhomovAlexander, line 45: commented

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/ReplBuilder.java
```diff
@@ -152,4 +155,15 @@ public ReplBuilder withAutosuggestionsWidgets() {
         this.autosuggestionsWidgetsEnabled = true;
         return this;
     }
+
+    /**
+     * Builder setter of {@code aliases} field.
+     *
+     * @param eventSubscriber ???? map of aliases for commands.
```
PakhomovAlexander, line 162:
```suggestion
     * @param eventSubscriber map of aliases for commands.
```

modules/cli/src/main/java/org/apache/ignite/internal/cli/event/Event.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.cli.event;
+
+/**
+ * The event cas which is produced by event producer component.
+ */
+public interface Event {
+}
```
PakhomovAlexander, line 24: missed eof

PakhomovAlexander (changes requested):

modules/cli/src/main/java/org/apache/ignite/internal/cli/event/EventListener.java
```diff
@@ -0,0 +1,27 @@
...
+
+package org.apache.ignite.internal.cli.event;
+
+/**
+ * The listener handles events from a producer.
+ */
+@FunctionalInterface
+public interface EventListener {
+
+    void onEvent(EventType eventType, Event event);
```
PakhomovAlexander, line 26: Could you add `eventType` method here?

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/ItConnectionHeartbeatTest.java
PakhomovAlexander: Could you mention this in the comment?

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/ConnectionHeartBeat.java
```diff
@@ -0,0 +1,130 @@
...
+     * Starts connection heartbeat. By default connection will be checked every 5 sec.
+     *
+     * @param sessionInfo session info with node url
+     */
+    private void onConnect(SessionInfo sessionInfo) {
+        if (connected.compareAndSet(false, true)) {
+            eventPublisher.fireEvent(Events.connectionRestored());
+        }
+
+        if (scheduledConnectionHeartbeatExecutor == null) {
```
PakhomovAlexander, line 84: I think a null check on `scheduledConnectionHeartbeatExecutor ` should be done under the lock in `onConnect` and in `onDisconnect` methods.

modules/cli/src/main/java/org/apache/ignite/internal/cli/core/repl/EventListeningActivationPoint.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.ignite.internal.cli.event.EventType;
+
+/**
+ * Subscribes beans on appropriate events.
+ */
+@Singleton
+public class EventListeningActivationPoint {
+
+    @Inject
+    private EventSubscriber eventSubscriber;
```
PakhomovAlexander, line 37: Can you please rename `EventSubscriber` to something like `EventSubscribtionManager`? Because I'm confused a little. When I read "Event Subscriber" I think of it like this object is subscribed to event but not subscribes others.

modules/cli/src/main/java/org/apache/ignite/internal/cli/event/Event.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package org.apache.ignite.internal.cli.event;
+
+/**
+ * The event cas which is produced by event producer component.
+ */
+public interface Event {
+    EventType eventType();
+}
```
PakhomovAlexander, line 25: Missed eof.  It is better to configure your IDEA to do it automatically.

modules/cli/src/main/java/org/apache/ignite/internal/cli/event/EventFactory.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * Registers an event listener.
+     *
+     * @param eventType     type of event to listen.
+     * @param eventListener event listener.
+     */
+    @Override
+    public void listen(EventType eventType, EventListener eventListener) {
```
PakhomovAlexander, line 48: I think it is better to use "subscribe" everywhere. Sometimes you mix "subscribe" and "listen" but in fact, there is no difference.

PakhomovAlexander (approved):

PR #2368, opened 2023-07-27
IGNITE-19877 Sql. Fix not allowed CAST with dynamic params

zstan:
https://issues.apache.org/jira/browse/IGNITE-19877

### DIFF
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItPublicApiColocationTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItPublicApiColocationTest.java
@@ -284,7 +284,6 @@ public class ItPublicApiColocationTest extends ClusterPerClassIntegrationTest {
             case INT32:
                 return "integer";
             case INT64:
-            case NUMBER:
                 return "bigint";
             case FLOAT:
                 return "real";
@@ -300,6 +299,8 @@ public class ItPublicApiColocationTest extends ClusterPerClassIntegrationTest {
                 return "varbinary";
             case BITMASK:
                 return "bitmap";
+            case NUMBER:
+                return "number";
             case DATE:
                 return "date";
             case TIME:
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItPublicApiColocationTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItPublicApiColocationTest.java
@@ -284,6 +284,7 @@ public class ItPublicApiColocationTest extends ClusterPerClassIntegrationTest {
             case INT32:
                 return "integer";
             case INT64:
+            case NUMBER:
                 return "bigint";
             case FLOAT:
                 return "real";
@@ -299,8 +300,6 @@ public class ItPublicApiColocationTest extends ClusterPerClassIntegrationTest {
                 return "varbinary";
             case BITMASK:
                 return "bitmap";
-            case NUMBER:
-                return "number";
             case DATE:
                 return "date";
             case TIME:
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
@@ -17,6 +17,7 @@
 
 package org.apache.ignite.internal.sql.engine.prepare;
 
+import static org.apache.calcite.sql.type.SqlTypeUtil.isNull;
 import static org.apache.calcite.util.Static.RESOURCE;
 import static org.apache.ignite.lang.IgniteStringFormatter.format;
 
@@ -27,6 +28,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import org.apache.calcite.avatica.SqlType;
 import org.apache.calcite.plan.RelOptTable;
 import org.apache.calcite.prepare.CalciteCatalogReader;
 import org.apache.calcite.prepare.Prepare;
@@ -423,6 +425,14 @@ public class IgniteSqlValidator extends SqlValidatorImpl {
                 firstType = super.deriveType(scope, first);
             }
 
+
+            boolean nullType = isNull(returnType) || isNull(firstType);
+
+            // propagate null type validation
+            if (nullType) {
+                return;
+            }
+
             RelDataType returnCustomType = returnType instanceof IgniteCustomType ? returnType : null;
             RelDataType fromCustomType = firstType instanceof IgniteCustomType ? firstType : null;
 
@@ -453,8 +463,6 @@ public class IgniteSqlValidator extends SqlValidatorImpl {
                         var ex = IgniteResource.INSTANCE.operationRequiresExplicitCast(operator.getName());
                         throw SqlUtil.newContextException(expr.getParserPosition(), ex);
                     } else {
-/*                        SqlCallBinding callBinding = new SqlCallBinding(this, scope, (SqlCall) expr);
-                        throw callBinding.newValidationSignatureError();*/
                         var ex = RESOURCE.incompatibleValueType(operator.getName());
                         throw SqlUtil.newContextException(expr.getParserPosition(), ex);
                     }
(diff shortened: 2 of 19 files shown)


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/DynamicParametersTest.java
```diff
@@ -44,6 +52,41 @@ public void testInferencePreservesNullability() throws Throwable {
...
+            /*sql(String.format("SELECT CAST(time '02:30:00' AS %s)", typeName)).fails("Cast function cannot convert value of type")
+                    .getExecutable().execute();*/
+        } else {
+            sql(String.format("SELECT CAST(? AS %s)", typeName), now).ok()
+                    .getExecutable().execute();
+        }
+
+        // not supported type.
+        Object custom = new Object();
+        sql(String.format("SELECT CAST(? AS %s)", typeName), custom).fails("Cast function cannot convert value of type")
```
lowka, line 86: You can replace .getExecutable().execute();
with 
>Stream.of(
> sql(...).check(),
> sql(...).check(),
>)
+ Change `testAllowedCasts` to return `Stream<DynamicTest>`
After that each test case runs separately.

zstan: thanks, done

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -358,11 +358,22 @@ public void validateAggregateParams(SqlCall aggCall,
     public RelDataType deriveType(SqlValidatorScope scope, SqlNode expr) {
         RelDataType dataType = super.deriveType(scope, expr);
 
+        // If type of dynamic parameter has not been inferred, use a type of its value.
```
lowka, line 361: @zstan I think we should check whether `expr` is a cast expr, and if it `true`, and then validate its arguments ensure that it is legal cast. Because at the moment it looks like the check is applied to dynamic parameters in any expression or maybe I am missing something here?

zstan: make sense, done.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -399,6 +403,54 @@ public RelDataType deriveType(SqlValidatorScope scope, SqlNode expr) {
...
+                customTypeCheck = returnType.toString().equals(firstType.toString());
+
+                if (!customTypeCheck) {
+                    IgniteCustomTypeCoercionRules coercionRules = ((IgniteTypeFactory) typeFactory).getCustomTypeCoercionRules();
+
+                    String returnTypeName = returnType.toString();
+                    SqlTypeName firstTypeName = firstType.getSqlTypeName();
+
+                    customTypeCheck = SqlTypeUtil.inCharFamily(returnType)
+                            || coercionRules.canCastFrom(returnTypeName).contains(firstTypeName);
```
lowka, line 437: `returnType.toString` returns more than a type name, so `coercionRules.canCastFrom(returnTypeName)` might fail when it shouldn't. 

In addition to that comparing string representation of types != comparing types, so it does not guarantee to work.

korlov42 (changes requested):
I run following test

```
    @Test
    public void test() {
        assertQuery("SELECT CAST(? AS DECIMAL(5, 1))")
                .withParams(LocalDateTime.now())
                .returns(1)
                .check();
    }
```
(the very same test from ticket description) and got "Numeric field overflow" (the very same exception from ticket description).


Please, write exhaustive set of tests. Make test simple as possible (perhaps, it would be better to avoid autogeneration). Make sure the case from description is actually fixed. Make self review. And only after all these steps submit patch for review.

korlov42:
> another existing issue covers highlighted problem [IGNITE-20069](https://issues.apache.org/jira/browse/IGNITE-20069)

why do we need this patch, if we will have general solution in IGNITE-20069? Why do we need 2 JIRA cases for the same problem in the first place?

> issue describes a bit different problem, i will fix it.

so, you've changed `LocalDateTime` --> `LocalTime`, but what is the difference? According to table above, both conversions `T->EN` and `TS->EN` are not allowed

zstan:
you are rigth here too, but for now we have only one place in code which need to be elaborated, issue  IGNITE-20069 was covered in case of working under this one, this is not to equal issues, from my point of view this is two sequential issues

zstan:
@korlov42 @lowka i hope all possible types casting is fixed by now

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItPublicApiColocationTest.java
```diff
@@ -284,6 +284,7 @@ private static String sqlTypeName(NativeTypeSpec type) {
             case INT32:
                 return "integer";
             case INT64:
+            case NUMBER:
```
lowka, line 287: I thin this is not correct because there is no SQL type for NativeType.NUMBER + a number is BigInteger

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/util/StatementChecker.java
```diff
@@ -371,14 +387,16 @@ private IgniteSchema createSchema() {
...
-    private DynamicTest shouldPass(String name, Throwable exception, Consumer<IgniteRel> check) {
+    private DynamicTest shouldPass(String name, Throwable exception, Consumer<IgniteRel> check, boolean relCheck) {
         return DynamicTest.dynamicTest(name, () -> {
             IgniteSchema schema = initSchema(exception);
             IgniteRel root;
 
             try {
                 root = (IgniteRel) sqlPrepare.prepare(schema, sqlStatement, dynamicParams);
-                checkRel(root, schema);
+                if (relCheck) {
```
lowka, line 397: Why do we need this?

zstan: yep, i fill appropriate issue https://issues.apache.org/jira/browse/IGNITE-20170

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItPublicApiColocationTest.java
zstan: @lowka did i need to change here smth ?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/util/StatementChecker.java
zstan: Ok i can change it for extended ver, but it need to be implemented in 2-3 classes, i fill the issue and fill apropriate comment to remove it after fix, is it ok ?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItPublicApiColocationTest.java
lowka: There is no need for this change because NativeType::Number is part of EXCLUDED_TYPES defined by this class.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/util/StatementChecker.java
lowka: Sure.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/Commons.java
```diff
@@ -169,6 +180,59 @@ public final class Commons {
     private Commons() {
     }
 
+    private static SqlTypeCoercionRule standardCompatibleCoercionRules() {
+        Map<SqlTypeName, ImmutableSet<SqlTypeName>> mappings = new EnumMap<>(SqlTypeCoercionRule.instance()
```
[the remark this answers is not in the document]
zstan: @AMashenkov is it still relevant  ?

AMashenkov (approved):

PR #2370, opened 2023-07-27
IGNITE-20075 Get rid of potential thread leak in RocksDbTxStateStorageTest

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-20075

### DIFF
--- a/modules/transactions/src/test/java/org/apache/ignite/internal/tx/storage/state/rocksdb/RocksDbTxStateStorageTest.java
+++ b/modules/transactions/src/test/java/org/apache/ignite/internal/tx/storage/state/rocksdb/RocksDbTxStateStorageTest.java
@@ -24,14 +24,19 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 import org.apache.ignite.internal.testframework.WorkDirectory;
 import org.apache.ignite.internal.testframework.WorkDirectoryExtension;
 import org.apache.ignite.internal.tx.TxMeta;
 import org.apache.ignite.internal.tx.storage.state.AbstractTxStateStorageTest;
 import org.apache.ignite.internal.tx.storage.state.TxStateStorage;
+import org.apache.ignite.internal.util.IgniteUtils;
 import org.apache.ignite.lang.IgniteBiTuple;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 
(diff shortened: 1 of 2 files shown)


modules/transactions/src/test/java/org/apache/ignite/internal/tx/storage/state/rocksdb/RocksDbTxStateStorageTest.java
```diff
@@ -43,18 +48,28 @@ public class RocksDbTxStateStorageTest extends AbstractTxStateStorageTest {
     @WorkDirectory
     private Path workDir;
 
+    private final ScheduledExecutorService scheduledExecutor = new ScheduledThreadPoolExecutor(1);
```
sashapolo, line 51: Not really an issue, but I would suggest to use `Executors.newSingleThreadScheduledExecutor`

modules/transactions/src/test/java/org/apache/ignite/internal/tx/storage/state/rocksdb/RocksDbTxStateStorageTest.java
```diff
@@ -43,18 +48,28 @@ public class RocksDbTxStateStorageTest extends AbstractTxStateStorageTest {
...
+
+    private final ExecutorService executor = Executors.newFixedThreadPool(1);
```
sashapolo, line 53: Same here, we can change it to `Executors.newSingleThreadExecutor` for consistency sake

sashapolo (approved):

modules/transactions/src/test/java/org/apache/ignite/internal/tx/storage/state/rocksdb/RocksDbTxStateStorageTest.java
```diff
@@ -43,18 +48,28 @@ public class RocksDbTxStateStorageTest extends AbstractTxStateStorageTest {
...
-                new ScheduledThreadPoolExecutor(1),
-                Executors.newFixedThreadPool(1),
+                scheduledExecutor,
+                executor,
                 () -> 1_000
         );
     }
 
+    @AfterEach
+    void shutdownExecutors() {
```
sashapolo, line 68: I'm a little bit worried that we stop the executors before we close the storage, maybe we should override the parent's `AfterEach` instead.

modules/transactions/src/test/java/org/apache/ignite/internal/tx/storage/state/rocksdb/RocksDbTxStateStorageTest.java
tkalkirill: fix it

modules/transactions/src/test/java/org/apache/ignite/internal/tx/storage/state/rocksdb/RocksDbTxStateStorageTest.java
tkalkirill: fix it

modules/transactions/src/test/java/org/apache/ignite/internal/tx/storage/state/rocksdb/RocksDbTxStateStorageTest.java
tkalkirill: fix it

modules/transactions/src/testFixtures/java/org/apache/ignite/internal/tx/storage/state/AbstractTxStateStorageTest.java
```diff
@@ -75,8 +74,7 @@ void beforeTest() {
         tableStorage.start();
     }
 
-    @AfterEach
-    void afterTest() {
+    protected void afterTest() {
```
sashapolo, line 77: Why did you remove `@AfterEach` here?

tkalkirill: I tried to use the annotation `@Override` and `@AfterEach` at the same time and found that none of them is called, I read [here](https://stackoverflow.com/questions/48864716/before-beforeeach-inheritance-behaviour-change-junit4-junit5) that in JUnit5 you need to do it differently than in JUnit4, I did it.

tkalkirill: fix it

sashapolo (approved):

PR #2371, opened 2023-07-27
IGNITE-19888 Java client: Track observable timestamp

ptupitsyn:
* Add observable timestamp to standard response header in client protocol (including errors, because data modification can occur before the error, like in a compute task - so we still need to update client-side timestamp)
* Add observable timestamp to TX_BEGIN and SQL_EXEC client requests
* Propagate latest known timestamp in Java client
* Fix .NET and C++ clients to ignore the timestamp for now (separate tickets exist)


isapego (approved):

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
```diff
@@ -385,6 +394,7 @@ private void writeError(long requestId, int opCode, Throwable err, ChannelHandle
             packer.packInt(ServerMessageType.RESPONSE);
             packer.packLong(requestId);
             writeFlags(packer, ctx);
+            packer.packLong(observableTimestamp(null));
```
denis-chudov, line 397: Does it make sense? It will just write the current timestamp, unrelated to the error itself

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/IgniteTransactionsImpl.java
```diff
@@ -39,25 +41,31 @@ public IgniteTransactionsImpl(TxManager txManager) {
...
     }
 
-    /** {@inheritDoc} */
-    @Override
-    public Transaction begin(@Nullable TransactionOptions options) {
+    /**
+     * Begins a transaction.
+     *
+     * @param options Transaction options.
+     * @param observableTimestamp Timestamp.
```
denis-chudov, line 48: Let's minimize the usage of such placeholder descriptions :) you can easily reuse the description of observable timestamp from javadoc from here: `TxManager#begin(boolean, org.apache.ignite.internal.hlc.HybridTimestamp)`

modules/client/src/test/java/org/apache/ignite/client/ObservableTimestampPropagationTest.java
```diff
@@ -0,0 +1,97 @@
...
+ * Tests that observable timestamp (causality token) is propagated from server to client and back.
+ */
+public class ObservableTimestampPropagationTest {
+    private static TestServer testServer;
+
+    private static FakeIgnite ignite;
+
+    private static IgniteClient client;
+
+    private static final AtomicLong observableTimestamp = new AtomicLong(1);
```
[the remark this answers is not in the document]
ptupitsyn: Fixed

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
ptupitsyn: We should write `now()` with every error, because we don't know if an operation has modified any data before it failed.

For example, a compute job can update one row and fail on another. As a user, I would expect to see the first update from the client anyway, so we need a recent timestamp.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/IgniteTransactionsImpl.java
ptupitsyn: Good point, fixed.

denis-chudov (approved):

PR #2372, opened 2023-07-27
IGNITE-18875 Sql. Drop AbstractPlannerTest.TestTable.

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-18875

### DIFF
--- a/modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestBuilders.java
+++ b/modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestBuilders.java
@@ -37,6 +37,7 @@ import org.apache.ignite.internal.sql.engine.exec.ArrayRowHandler;
 import org.apache.ignite.internal.sql.engine.exec.ExecutionContext;
 import org.apache.ignite.internal.sql.engine.exec.QueryTaskExecutor;
 import org.apache.ignite.internal.sql.engine.exec.TxAttributes;
+import org.apache.ignite.internal.sql.engine.metadata.ColocationGroup;
 import org.apache.ignite.internal.sql.engine.metadata.FragmentDescription;
 import org.apache.ignite.internal.sql.engine.schema.ColumnDescriptor;
 import org.apache.ignite.internal.sql.engine.schema.ColumnDescriptorImpl;
@@ -132,6 +133,8 @@ public class TestBuilders {
 
         /** Returns a builder of the test hash-index object. */
         public HashIndexBuilder hashIndex();
+
+        TableBuilder colocationGroup(ColocationGroup colocationGroup);
 
         /**
          * Builds a table.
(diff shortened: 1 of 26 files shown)


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestStatistic.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.calcite.rel.RelCollation;
+import org.apache.calcite.rel.RelDistribution;
+import org.apache.calcite.rel.RelReferentialConstraint;
+import org.apache.calcite.schema.Statistic;
+import org.apache.calcite.util.ImmutableBitSet;
+
+/**
+ * A test statistic that implements all the necessary for the optimizer methods to be used to prepare a query.
+ */
+public class TestStatistic implements Statistic {
```
lowka, line 31: Why not to use the same Statistics object as production code?

lowka: IgniteStatistics only requires Distribution but it is accessible in constructor of TestTable via TableDescriptor.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestBuilders.java
```diff
@@ -138,6 +138,8 @@ public interface TableBuilder extends TableBuilderBase<TableBuilder> {
         /** Returns a builder of the test hash-index object. */
         public HashIndexBuilder hashIndex();
 
+        TableBuilder colocationGroup(ColocationGroup colocationGroup);
```
korlov42, line 141: why do we need colocation group in TableBuilder? I mean, TableBuilder is meant to build table for planner tests, they don't need colocation group to be provided

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/CorrelatedNestedLoopJoinPlannerTest.java
```diff
@@ -112,51 +83,29 @@ public IgniteDistribution distribution() {
...
-                                .build(), "T1") {
-
-                    @Override
-                    public IgniteDistribution distribution() {
-                        return IgniteDistributions.broadcast();
-                    }
-                }
-                        .addIndex("t1_jid_idx", 1, 0)
+        IgniteSchema publicSchema = createSchemaFrom(
+                tableA("T0").andThen(addHashIndex("JID", "ID")),
```
korlov42, line 87: > tableA("T0").andThen(addHashIndex("JID", "ID"))

This notation looks cool!

korlov42 (approved):

PR #2375, opened 2023-07-28
IGNITE-20076 Improve networking shutdown implementation

rpuch:
https://issues.apache.org/jira/browse/IGNITE-20076

### DIFF
--- a/modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
+++ b/modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
@@ -26,6 +26,7 @@ import io.netty.channel.ChannelHandlerContext;
 import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.ignite.internal.logger.IgniteLogger;
 import org.apache.ignite.internal.logger.Loggers;
 import org.apache.ignite.internal.network.NetworkMessagesFactory;
@@ -45,7 +46,6 @@ import org.apache.ignite.internal.network.recovery.message.HandshakeStartRespons
 import org.apache.ignite.lang.IgniteException;
 import org.apache.ignite.network.NetworkMessage;
 import org.apache.ignite.network.OutNetworkObject;
-import org.jetbrains.annotations.TestOnly;
 
 /**
  * Recovery protocol handshake manager for a client.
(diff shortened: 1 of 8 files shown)


modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryServerHandshakeManager.java
```diff
@@ -177,12 +182,20 @@ public void onMessage(NetworkMessage message) {
         if (message instanceof HandshakeRejectedMessage) {
             HandshakeRejectedMessage msg = (HandshakeRejectedMessage) message;
 
-            LOG.warn("Handshake rejected by client: {}", msg.reason());
+            boolean ignorable = stopping.get() || !msg.critical();
```
sashapolo, line 185: Why do we check the stopping flag here?

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
```diff
@@ -156,12 +161,20 @@ public void onMessage(NetworkMessage message) {
         if (message instanceof HandshakeRejectedMessage) {
             HandshakeRejectedMessage msg = (HandshakeRejectedMessage) message;
 
-            LOG.warn("Handshake rejected by server: {}", msg.reason());
+            boolean ignorable = stopping.get() || !msg.critical();
```
sashapolo, line 164: Same question

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryServerHandshakeManager.java
rpuch: If a node are stopping, then the situation 'I was not let by another node to do a handshake' is expected and should neither cause anything (serious) be logged nor fail the node

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
rpuch: The explanation is same :)

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryServerHandshakeManager.java
sashapolo: Agree

sashapolo (approved):

PR #2390, opened 2023-08-01
IGNITE-19009 Introduce file transfer support in messaging

ivgag:
https://issues.apache.org/jira/browse/IGNITE-19009

### DIFF
--- a/modules/file-transfer/src/test/java/org/apache/ignite/internal/network/file/FileTransferServiceImplTest.java
+++ b/modules/file-transfer/src/test/java/org/apache/ignite/internal/network/file/FileTransferServiceImplTest.java
@@ -27,6 +27,7 @@ import static org.apache.ignite.internal.testframework.matchers.CompletableFutur
 import static org.awaitility.Awaitility.await;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.emptyArray;
+import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.nullValue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyList;
@@ -43,7 +44,7 @@ import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executors;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.ignite.internal.network.file.messages.FileDownloadRequest;
 import org.apache.ignite.internal.network.file.messages.FileDownloadResponse;
 import org.apache.ignite.internal.network.file.messages.FileHeader;
@@ -266,19 +267,18 @@ class FileTransferServiceImplTest {
     @Test
     void transferIsRegisteredBeforeResponseIsSent() {
         // Set file receiver to complete transfer registration.
-        AtomicBoolean transferRegistered = new AtomicBoolean(false);
+        AtomicInteger transferLifecycleState = new AtomicInteger(0);
         doAnswer(invocation -> {
-            transferRegistered.set(true);
+            // Update transfer lifecycle state - transfer is registered.
+            transferLifecycleState.compareAndSet(0, 1);
             return (TransferredFilesCollector) () -> completedFuture(List.of());
         }).when(fileReceiver).registerTransfer(anyString(), any(UUID.class), anyList(), any(Path.class));
 
         // Set messaging service to fail to send upload response if transfer is not registered.
         doAnswer(invocation -> {
-            if (!transferRegistered.get()) {
-                throw new RuntimeException("Transfer is not registered");
-            } else {
-                return completedFuture(null);
-            }
+            // Check lifecycle state - transfer is registered.
+            transferLifecycleState.compareAndSet(1, 2);
+            return completedFuture(null);
         }).when(messagingService).respond(anyString(), eq(FILE_TRANSFER_CHANNEL), any(FileTransferInitResponse.class), anyLong());
 
         // Create file transfer request.
@@ -293,5 +293,8 @@ class FileTransferServiceImplTest {
 
         // Send file transfer request.
         messagingService.fairMessage(fileTransferInitMessage, TARGET_CONSISTENT_ID, 1L);
+
+        // Check that transfer was registered before response was sent.
+        await().untilAtomic(transferLifecycleState, equalTo(2));
     }
 }


modules/file-transferring/src/main/java/org/apache/ignite/internal/network/file/messages/ChunkedFile.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Comparator;
+import java.util.UUID;
+import org.apache.ignite.network.NetworkMessage;
+import org.apache.ignite.network.annotations.Transferable;
+
+/**
+ * Chunked file.
+ */
+@Transferable(FileTransferringMessageType.CHUNKED_FILE)
+public interface ChunkedFile extends NetworkMessage {
```
rpuch, line 29: I suggest renaming it to `FileChunk`

modules/network-api/src/main/java/org/apache/ignite/network/MessagingService.java
```diff
@@ -76,6 +76,26 @@ default CompletableFuture<Void> send(ClusterNode recipient, NetworkMessage msg)
...
+     *
+     * <p>Please note that the guarantees only work for same (sender, receiver) pairs. That is, if A sends m1 and m2
+     * to B, then the guarantees are maintained. If, on the other hand, A sends m1 to B and m2 to C, then no guarantees
+     * exist.
+     *
+     * @param recipientConsistentId Consistent ID of the recipient of the message.
+     * @param msg       Message which should be delivered.
+     * @return Future of the send operation.
+     */
+    CompletableFuture<Void> send(String recipientConsistentId, ChannelType channelType, NetworkMessage msg);
```
rpuch, line 97: Documentation about `ChannelType` is missing.

Also, it's not clear what will happen when using old `send()` without this parameter. Some default channel wlll be used? No channel will be used?

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -400,4 +400,15 @@ public static class Compute {
         /** Class loader error. */
         public static final int CLASS_LOADER_ERR = COMPUTE_ERR_GROUP.registerErrorCode((short) 2);
     }
+
+    /**
+     * File transfer error group.
+     */
+    public static class FileTransfer {
```
sk0x50, line 407: Do we really need a dedicated error group for this? Perhaps, it should be a part of the network error group. WDYT?

modules/core/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -400,4 +400,15 @@ public static class Compute {
...
+        /** File transfer error group. */
+        public static final ErrorGroup FILE_TRANSFER_ERR_GROUP = ErrorGroup.newGroup("FILE_TRANSFER", (short) 17);
+
+        /** Generic error. */
+        public static final int GENERIC_ERR = FILE_TRANSFER_ERR_GROUP.registerErrorCode((short) 1);
```
sk0x50, line 412: Please, avoid using generic error codes, just because it does not make much sense to the end user. This must be used with great care.

modules/file-transfer/src/integrationTest/java/org/apache/ignite/internal/network/file/ItFileTransferTest.java
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    @AfterEach
+    void tearDown() throws Exception {
+        cluster.shutdown();
+    }
+
+    @Test
+    void download() throws IOException {
+        Node node0 = cluster.members.get(0);
```
PakhomovAlexander, line 83: Could you name `node0` like `sourceNode` and `node1` -> `targetNode`.

modules/file-transfer/src/integrationTest/java/org/apache/ignite/internal/network/file/ItFileTransferTest.java
```diff
@@ -0,0 +1,336 @@
...
+
+        Node node1 = cluster.members.get(1);
+        CompletableFuture<List<File>> downloadedFilesFuture = node1.fileTransferringService().download(
+                node0.nodeName(),
+                MetadataImpl.builder().build()
+        );
+
+        assertThat(
+                downloadedFilesFuture.thenAccept(files -> {
+                    assertContentEquals(Set.of(file1, file2, file3, file4), new HashSet<>(files));
```
PakhomovAlexander, line 109: Why do you use sets in the assertion? This might hide duplicates.

modules/file-transfer/src/integrationTest/java/org/apache/ignite/internal/network/file/TestCluster.java
```diff
@@ -0,0 +1,191 @@
...
+    /**
+     * Starts and waits for the cluster to come up.
+     *
+     * @throws InterruptedException If failed.
+     * @throws AssertionError If the cluster was unable to start in 3 seconds.
+     */
+    void startAwait() throws InterruptedException {
+        members.forEach(Node::start);
+
+        if (!startupLatch.await(3, TimeUnit.SECONDS)) {
```
PakhomovAlexander, line 127: Why is `3` hardcoded?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/ChunkedFileReader.java
```diff
@@ -0,0 +1,68 @@
...
+        this.chunkSize = chunkSize;
+    }
+
+    static ChunkedFileReader open(File file, int chunkSize) throws IOException {
+        RandomAccessFile raf = new RandomAccessFile(file, "r");
+        return new ChunkedFileReader(file.getName(), raf, chunkSize);
+    }
+
+    byte[] readNextChunk() throws IOException {
+        int toRead = (int) Math.min(chunkSize, raf.length() - raf.getFilePointer());
```
PakhomovAlexander, line 42: I think you can use methods `length()` and `offset()`.

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileHandler.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.ignite.internal.network.file.messages.Metadata;
+
+/**
+ * Handler for the uploaded file.
+ */
+public interface FileHandler<M extends Metadata> {
+    /**
+     * Handles the uploaded files.
+     *
+     * @param uploadedFiles The temporary files that were uploaded. These files will be deleted after the method returns.
```
PakhomovAlexander, line 32: Metadata parameter description is missed.

modules/file-transfer/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+
+apply from: "$rootDir/buildscripts/java-core.gradle"
+apply from: "$rootDir/buildscripts/publishing.gradle"
+apply from: "$rootDir/buildscripts/java-junit5.gradle"
+apply from: "$rootDir/buildscripts/java-integration-test.gradle"
+apply from: "$rootDir/buildscripts/java-test-fixtures.gradle"
+
+dependencies {
```
PakhomovAlexander, line 24: This is a huge module and the README would be usefull.

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/RateLimiter.java
```diff
@@ -0,0 +1,43 @@
...
+    private final int maxConcurrentRequests;
+    private final AtomicInteger counter = new AtomicInteger(0);
+
+    RateLimiter(int maxConcurrentRequests) {
+        this.maxConcurrentRequests = maxConcurrentRequests;
+    }
+
+    boolean tryAcquire() {
+        int count = counter.incrementAndGet();
+        if (count <= maxConcurrentRequests) {
```
PakhomovAlexander, line 32: I think using explicit lock is abetter choice here.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/ClusterServiceFactory.java
```diff
@@ -184,6 +184,11 @@ public CompletableFuture<Void> send(ClusterNode recipient, ChannelType channelTy
             return CompletableFuture.completedFuture(null);
         }
 
+        @Override
+        public CompletableFuture<Void> send(String recipientConsistentId, ChannelType channelType, NetworkMessage msg) {
+            throw new AssertionError("Not implemented yet");
```
PakhomovAlexander, line 189: Do we have a ticket for the implementation?

modules/core/src/main/java/org/apache/ignite/internal/util/FilesUtils.java
```diff
@@ -0,0 +1,60 @@
...
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+
+/**
+ * Files utilities.
+ */
+public class FilesUtils {
```
rpuch, line 30: It looks like there are not tests for this class. Let's add them.

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileSender.java
```diff
@@ -0,0 +1,114 @@
...
+        AtomicReference<Throwable> error = new AtomicReference<>();
+        try (FileTransferMessagesStream stream = new FileTransferMessagesStream(id, files, chunkSize)) {
+            while (stream.hasNextMessage() && error.get() == null && !Thread.currentThread().isInterrupted()) {
+                if (rateLimiter.tryAcquire()) {
+                    CompletableFuture.completedFuture(stream.nextMessage())
+                            .thenCompose(message -> send.apply(receiverConsistentId, message))
+                            .whenComplete((res, e) -> {
+                                try {
+                                    if (e != null) {
+                                        LOG.error("Failed to send message to node: {}, transfer id: {}. Exception: {}",
```
rpuch, line 85: Let's use another variant of `IgniteLogger.error()`: the one that takes a Throwable as a second argument. Then the exception does not need to be mentioned explicitly in the log message.

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileSender.java
```diff
@@ -0,0 +1,114 @@
...
+                                        error.compareAndSet(null, e);
+                                    }
+                                } finally {
+                                    rateLimiter.release();
+                                }
+                            });
+                }
+            }
+
+            if (error.get() != null) {
```
rpuch, line 99: Do we wait for the sends to complete somewhere? If not, we'll probably miss an error if it happens

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferMessagesHandler.java
```diff
@@ -0,0 +1,171 @@
...
+        } catch (IOException e) {
+            handleFileTransferError(e);
+        }
+    }
+
+    void handleFileChunk(FileChunkMessage fileChunk) {
+        if (result.isDone()) {
+            throw new IllegalStateException("Received chunked file after result is already done.");
+        }
+        doInLock(fileChunk.fileName(), () -> handleFileChunk0(fileChunk));
```
rpuch, line 97: This is called from a network thread. No blocking/IO operations should be run on this thread

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferMessagesHandler.java
```diff
@@ -0,0 +1,171 @@
...
+        } catch (IOException e) {
+            handleFileTransferError(e);
+        }
+    }
+
+    void handleFileHeader(FileHeaderMessage header) {
+        if (result.isDone()) {
+            throw new IllegalStateException("Received file header after result is already done.");
+        }
+        doInLock(header.fileName(), () -> handleFileHeader0(header));
```
rpuch, line 69: This is called from a network thread. No blocking/IO operations should be run on this thread

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/ChunkedFileWriter.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    static ChunkedFileWriter open(Path path, long fileSize) throws FileNotFoundException {
+        return new ChunkedFileWriter(new RandomAccessFile(path.toFile(), "rw"), fileSize);
+    }
+
+    void write(FileChunkMessage chunk) throws IOException {
+        chunks.add(chunk);
+
+        while (!chunks.isEmpty() && chunks.peek().offset() == raf.getFilePointer()) {
```
rpuch, line 48: Why is this queue needed? Can the chunks be written out-of-order?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/ChunkedFileWriter.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Queue;
+import org.apache.ignite.internal.close.ManuallyCloseable;
+import org.apache.ignite.internal.network.file.messages.FileChunkMessage;
+
+class ChunkedFileWriter implements ManuallyCloseable {
+    private final RandomAccessFile raf;
+
+    private long fileSize;
+
+    private final Queue<FileChunkMessage> chunks = new PriorityQueue<>(FileChunkMessage.COMPARATOR);
```
rpuch, line 34: `PriorityQueue` is not thread-safe. If there a guarantee that the queue is always accessed from the same thread?

modules/file-transfer/src/testFixtures/java/org/apache/ignite/internal/network/file/FileUtils.java
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * Sorts files by names.
+     *
+     * @param files Files.
+     * @return Sorted files.
+     */
+    public static List<File> sortByNames(List<File> files) {
+        return files.stream()
+                .sorted(Comparator.comparing(File::getName))
+                .collect(Collectors.toList());
```
rpuch, line 42: Let's import `toList()` statically, this would make the line read as an English phrase

modules/network/src/main/java/org/apache/ignite/internal/network/configuration/NetworkConfigurationSchema.java
```diff
@@ -78,4 +78,8 @@ public class NetworkConfigurationSchema {
     @ConfigValue
     @SslConfigurationValidator
     public SslConfigurationSchema ssl;
+
+    /** File transferring configuration. */
+    @ConfigValue
+    public FileTransferConfigurationSchema fileTransferring;
```
rpuch, line 84:
```suggestion
    public FileTransferConfigurationSchema fileTransfer;
```

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferMessagesHandler.java
rpuch: My bad

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferMessagesHandler.java
rpuch: My bad

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/ChunkedFileWriter.java
rpuch: So there is synchronization that is external to the class which method uses a non-thread-safe data structure. It should probably be mentioned in a comment as otherwise it's easy to miss (and break in the future)

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/exception/FileHandlerNotFoundException.java
```diff
@@ -15,30 +15,22 @@
...
 /**
- * File utils.
+ * Exception thrown when file handler is not found for a metadata message.
  */
-public class FileUtils {
-    public static List<File> sortByNames(File... files) {
-        return sortByNames(List.of(files));
-    }
-
+public class FileHandlerNotFoundException extends IgniteException {
```
rpuch, line 27: Isn't it `FileConsumer` now?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileReceiver.java
```diff
@@ -0,0 +1,196 @@
...
+    private final ExecutorService executorService;
+
+    private final Map<UUID, FileTransferMessagesHandler> transferIdToHandler = new ConcurrentHashMap<>();
+
+    private final Map<String, Set<UUID>> senderConsistentIdToTransferIds = new ConcurrentHashMap<>();
+
+    private final RefCountedObjectPool<String, ReentrantLock> senderConsistentIdToLock = new RefCountedObjectPool<>();
+
+    FileReceiver(String nodeName, int threadPoolSize) {
+        this.executorService = new ThreadPoolExecutor(
```
ibessonov, line 59: Is there a matching method in `Executors` class?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileReceiver.java
```diff
@@ -0,0 +1,196 @@
...
+                }
+                v.add(transferId);
+                return v;
+            });
+            handler.result().whenComplete((files, throwable) -> deregisterTransfer(senderConsistentId, transferId));
+            return handler;
+        } finally {
+            lock.unlock();
+        }
+    }
```
ibessonov, line 86: Where are empty lines? This is very hard to read, and it violates code style guidelines. Please format your code properly

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileReceiver.java
```diff
@@ -0,0 +1,196 @@
...
+    private void deregisterTransfer(String senderConsistentId, UUID transferId) {
+        ReentrantLock lock = senderConsistentIdToLock.acquire(senderConsistentId, ignored -> new ReentrantLock());
+        lock.lock();
+        try {
+            transferIdToHandler.remove(transferId);
+            Set<UUID> uuids = senderConsistentIdToTransferIds.get(senderConsistentId);
+            uuids.remove(transferId);
+            if (uuids.isEmpty()) {
+                senderConsistentIdToTransferIds.remove(senderConsistentId);
+            }
```
ibessonov, line 97: In previous method you use `compute`, but here you use get/remove. What's the reason for discrepancy?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileReceiver.java
```diff
@@ -0,0 +1,196 @@
...
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    void cancelTransfersFromSender(String senderConsistentId) {
+        ReentrantLock lock = senderConsistentIdToLock.acquire(senderConsistentId, ignored -> new ReentrantLock());
+        lock.lock();
+        try {
+            new HashSet<>(senderConsistentIdToTransferIds.get(senderConsistentId))
```
ibessonov, line 107: I'm sorry, but why do you create a new hash set here? Also, why not using a `for ( : )` instead of streams, what are the benefits?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileReceiver.java
```diff
@@ -0,0 +1,196 @@
...
+import org.apache.ignite.internal.network.file.messages.FileTransferErrorMessage;
+import org.apache.ignite.internal.network.file.messages.FileTransferInfoMessage;
+import org.apache.ignite.internal.thread.NamedThreadFactory;
+import org.apache.ignite.internal.util.IgniteUtils;
+import org.apache.ignite.internal.util.RefCountedObjectPool;
+
+/**
+ * File receiver.
+ */
+class FileReceiver implements ManuallyCloseable {
```
ibessonov, line 47: What class is responsible for handling "node left" events and cancelling all operations?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileSender.java
```diff
@@ -0,0 +1,117 @@
...
+    private final BiFunction<String, NetworkMessage, CompletableFuture<Void>> send;
+
+    private final ExecutorService executorService;
+
+    FileSender(
+            String nodeName,
+            int chunkSize,
+            int threadPoolSize,
+            RateLimiter rateLimiter,
+            BiFunction<String, NetworkMessage, CompletableFuture<Void>> send) {
```
ibessonov, line 55: Why not just providing a messaging service here?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileSender.java
```diff
@@ -0,0 +1,117 @@
...
+        return CompletableFuture.runAsync(() -> send0(receiverConsistentId, id, files), executorService);
+    }
+
+    private void send0(String receiverConsistentId, UUID id, List<File> files) {
+        AtomicReference<Throwable> error = new AtomicReference<>();
+        try (FileTransferMessagesStream stream = new FileTransferMessagesStream(id, files, chunkSize)) {
+            while (stream.hasNextMessage() && error.get() == null && !Thread.currentThread().isInterrupted()) {
+                if (rateLimiter.tryAcquire()) {
+                    CompletableFuture.completedFuture(stream.nextMessage())
+                            .thenCompose(message -> send.apply(receiverConsistentId, message))
```
ibessonov, line 81: This approach is very non-intuitive to me. Why did you chose it instead of good old `send.apply(receiverConsistentId, stream.nextMessage())`?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileSender.java
ibessonov: I think the question was "this error will be set in the future, asynchronously, why do we check it right here?"

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileSender.java
```diff
@@ -0,0 +1,117 @@
...
+     * Adds files to the queue to be sent to the receiver.
+     */
+    CompletableFuture<Void> send(String receiverConsistentId, UUID id, List<File> files) {
+        return CompletableFuture.runAsync(() -> send0(receiverConsistentId, id, files), executorService);
+    }
+
+    private void send0(String receiverConsistentId, UUID id, List<File> files) {
+        AtomicReference<Throwable> error = new AtomicReference<>();
+        try (FileTransferMessagesStream stream = new FileTransferMessagesStream(id, files, chunkSize)) {
+            while (stream.hasNextMessage() && error.get() == null && !Thread.currentThread().isInterrupted()) {
```
ibessonov, line 78: Is this OK that we block the sending thread with the rate limiter? Doesn't look like a good solution to me

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferMessagesHandler.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.ignite.internal.network.file.messages.FileChunkMessage;
+import org.apache.ignite.internal.network.file.messages.FileHeaderMessage;
+import org.apache.ignite.internal.network.file.messages.FileTransferInfoMessage;
+import org.apache.ignite.internal.util.IgniteUtils;
+import org.apache.ignite.lang.IgniteInternalException;
+
+class FileTransferMessagesHandler {
```
ibessonov, line 39: Again, a total lack of comments in package-private classes. Do you assume that the code is trivial and doesn't require any comment at all?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferMessagesHandler.java
```diff
@@ -0,0 +1,171 @@
...
+    private void closeAllWriters() {
+        try {
+            IgniteUtils.closeAllManually(fileNameToWriter.values().stream());
+        } catch (Exception e) {
+            throw new IgniteInternalException(e);
+        }
+    }
+
+    private ChunkedFileWriter writer(String fileName) {
+        // set -1 as file size to indicate that file size is unknown and will be set later
```
ibessonov, line 158:
```suggestion
        // Set -1 as file size to indicate that file size is unknown and will be set later.
```

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferMessagesHandler.java
```diff
@@ -0,0 +1,171 @@
...
+        return writer(fileName, -1);
+    }
+
+    private ChunkedFileWriter writer(String fileName, long fileSize) {
+        try {
+            Path path = Files.createFile(dir.resolve(fileName));
+            return ChunkedFileWriter.open(path, fileSize);
+        } catch (IOException e) {
+            handleFileTransferError(e);
+            throw new IgniteInternalException(e);
```
ibessonov, line 168: Please don't do that, throw proper exception instead

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferMessagesStream.java
```diff
@@ -0,0 +1,226 @@
...
+        if (chunkSize <= 0) {
+            throw new IllegalArgumentException("Chunk size must be positive.");
+        }
+
+        if (filesToSend.isEmpty()) {
+            throw new IllegalArgumentException("Files to send must not be empty.");
+        }
+
+        this.transferId = transferId;
+        this.filesToSend = new LinkedList<>(filesToSend);
```
ibessonov, line 79: Very interesting, could you comment the reason for this choice?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferMessagesStream.java
```diff
@@ -0,0 +1,226 @@
...
+        this.fileTransferInfoMessage = fileTransferInfo();
+    }
+
+    /**
+     * Returns true if there are more messages to send.
+     *
+     * @return true if there are more messages to send.
+     */
+    boolean hasNextMessage() throws IOException {
+        // check that the stream is not closed.
```
ibessonov, line 90:
```suggestion
        // Check that the stream is not closed.
```

ibessonov: Please check all your comments

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -0,0 +1,402 @@
...
+import org.apache.ignite.network.annotations.Transferable;
+import org.jetbrains.annotations.TestOnly;
+
+/**
+ * Implementation of {@link FileTransferService}.
+ */
+public class FileTransferServiceImpl implements FileTransferService {
+    private static final IgniteLogger LOG = Loggers.forClass(FileTransferServiceImpl.class);
+
+    private static final ChannelType FILE_TRANSFERRING_CHANNEL = ChannelType.register((short) 1, "FileTransferring");
```
ibessonov, line 67: Don't we already have a channel type 1?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -0,0 +1,402 @@
...
+
+    private static final long RESPONSE_TIMEOUT = 10_000;
```
ibessonov, line 69: Should this be configurable? I guess so

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -0,0 +1,402 @@
...
+
+    /**
+     * Topology service.
+     */
+    private final TopologyService topologyService;
+    /**
```
ibessonov, line 75:
```suggestion

    /**
```

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -0,0 +1,402 @@
...
+     * File transfer factory.
+     */
+    private final FileTransferFactory factory = new FileTransferFactory();
+
+    /**
+     * Constructor.
+     *
+     * @param messagingService Messaging service.
+     * @param tempDirectory Temporary directory.
```
ibessonov, line 114: What about other 3 parameters?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -0,0 +1,402 @@
...
+                    }
+                });
+    }
+
+    private void processFileTransferErrorMessage(FileTransferErrorMessage message) {
+        fileReceiver.receiveFileTransferErrorMessage(message);
+    }
+
+    private void processUploadRequest(FileUploadRequest message, String senderConsistentId, long correlationId) {
+        completedFuture(UUID.randomUUID())
```
ibessonov, line 188: Once again I see this pattern. What's the reason? Writing
```
UUID transferId = UUID.randomUUID();
```
looks absolutely fine. Why would you need to pollute your code with extra paddings? To avoid writing the type explicitly?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -0,0 +1,402 @@
...
+                .thenCompose(transferId -> {
+                    Path directory = createTransferDirectory(transferId);
+                    FileTransferMessagesHandler handler = fileReceiver.registerTransfer(
+                            senderConsistentId,
+                            transferId,
+                            directory
+                    );
+                    FileUploadResponse response = factory.fileUploadResponse()
```
ibessonov, line 196: It would be nice to name it "messageFactory", because "factory" itself is too broad of a term

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -0,0 +1,402 @@
...
+                            .transferId(transferId)
+                            .build();
+                    Metadata metadata = message.metadata();
+                    return messagingService.respond(senderConsistentId, FILE_TRANSFERRING_CHANNEL, response, correlationId)
```
ibessonov, line 200: Threading model here is not clear to me. All "compose" are not "async", how do you guarantee that network thread won't be executing your disk IO operations?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -0,0 +1,402 @@
...
+                });
+    }
+
+    private void processDownloadRequest(FileDownloadRequest message, String senderConsistentId, Long correlationId) {
+        supplyAsync(() -> getFileProvider(message.metadata()))
+                .thenCompose(provider -> provider.files(message.metadata()))
+                .whenComplete((files, e) -> {
+                    if (e != null) {
+                        LOG.error("Failed to get files for download. Metadata: {}", message.metadata(), e);
+                        FileDownloadResponse response = factory.fileDownloadResponse()
```
ibessonov, line 225: I will ask you one more time here to use more empty lines in your code. I didn't write this comment every time. Code this dense is difficult to read and it doesn't look like the rest of Ignite code. Please fix it everywhere in your code.

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -0,0 +1,402 @@
...
+                            directory
+                    );
+                    FileUploadResponse response = factory.fileUploadResponse()
+                            .transferId(transferId)
+                            .build();
+                    Metadata metadata = message.metadata();
+                    return messagingService.respond(senderConsistentId, FILE_TRANSFERRING_CHANNEL, response, correlationId)
+                            .thenCompose(ignored -> {
+                                return handler.result()
+                                        .thenCompose(files -> getFileHandler(metadata).handleUpload(metadata, files));
```
ibessonov, line 203: Is `getFileHandler(metadata)` the same as `handler` here? It's hard to follow, which one is which :)

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -0,0 +1,402 @@
...
+                        FileTransferErrorMessage message = factory.fileTransferErrorMessage()
+                                .transferId(transferId)
+                                .error(unwrapFileTransferException(e).getMessage())
+                                .build();
+                        messagingService.send(recipientConsistentId, FILE_TRANSFERRING_CHANNEL, message);
+                    }
+                });
+    }
+
+    private void processFileTransferInfo(FileTransferInfoMessage info) {
```
ibessonov, line 261: Interesting, why not just inline these single-liners?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -0,0 +1,402 @@
...
+                        return failedFuture(new FileTransferException("No files to upload"));
+                    } else {
+                        return messagingService.invoke(nodeConsistentId, FILE_TRANSFERRING_CHANNEL, message, RESPONSE_TIMEOUT)
+                                .thenApply(FileUploadResponse.class::cast)
+                                .thenCompose(response -> sendFiles(nodeConsistentId, response.transferId(), files));
+                    }
+                });
+    }
+
+    private static short getMessageType(Class<?> metadata) {
```
ibessonov, line 352: Maybe we should register handlers by the numeric message types, rather than classes? Cause it looks a bit hacky right now

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -0,0 +1,402 @@
...
+        if (handler == null) {
+            throw new FileHandlerNotFoundException(metadata.getClass());
+        } else {
+            return (FileConsumer<M>) handler;
+        }
+    }
+
+    private Path createTransferDirectory(UUID transferId) {
+        try {
+            return Files.createDirectories(tempDirectory.resolve(transferId.toString()));
```
ibessonov, line 380: Why do we use temp directories? Can't we ask handler for the destination?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -0,0 +1,402 @@
...
+            FilesUtils.deleteDirectoryIfExists(directory);
+        } catch (IOException e) {
+            LOG.error("Failed to delete directory: {}. Exception: {}", directory, e);
+        }
+    }
+
+    private static Throwable unwrapFileTransferException(Throwable throwable) {
+        Throwable unwrapped = ExceptionUtils.unwrapCause(throwable);
+        if (unwrapped instanceof FileTransferException) {
+            return unwrapped.getCause() == null ? unwrapped : unwrapped.getCause();
```
ibessonov, line 397: Wow, why so complicated? I would love to have a comment here, or a simpler solution when it comes to exceptions

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/RateLimiter.java
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.network.file;
+
+class RateLimiter {
```
ibessonov, line 20: Why don't you use a semaphore instead?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/exception/FileProviderNotFoundException.java
```diff
@@ -0,0 +1,36 @@
...
+ * Exception thrown when file provider is not found for a metadata message.
+ */
+public class FileProviderNotFoundException extends IgniteException {
+    /**
+     * Constructor.
+     *
+     * @param message Message.
+     */
+    public FileProviderNotFoundException(Class<? extends Metadata> message) {
+        super(Network.FILE_PROVIDER_NOT_FOUND_ERR, "File provider not found for message: " + message.getName());
```
ibessonov, line 34: How is this a public exception for the user? Missing handler is a bug in the product, not an API misuse. Please remove these error codes from the "api" module

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/messages/FileTransferMessageType.java
```diff
@@ -0,0 +1,42 @@
...
+package org.apache.ignite.internal.network.file.messages;
+
+import org.apache.ignite.network.annotations.MessageGroup;
+
+/**
+ * File transfer message types.
+ */
+@MessageGroup(groupType = 200, groupName = "FileTransfer")
+public final class FileTransferMessageType {
+    public static final short FILE_TRANSFER_INFO = 0;
```
ibessonov, line 27: Please add a link to the class in javadocs for every constant, like we do it in other classes.

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/messages/FileTransferMessageType.java
```diff
@@ -0,0 +1,42 @@
...
+    public static final short FILE_DOWNLOAD_RESPONSE = 4;
+    public static final short FILE_UPLOAD_REQUEST = 5;
+    public static final short FILE_UPLOAD_RESPONSE = 6;
+    public static final short FILE_TRANSFER_ERROR = 7;
+    public static final short METADATA = 8;
+
+    /**
+     * File transferring metadata.
+     */
+    public static final class Metadata {
```
ibessonov, line 40: What's this class for? I assume you forgot to delete it

modules/file-transfer/src/test/java/org/apache/ignite/internal/network/file/FileReceiverTest.java
```diff
@@ -0,0 +1,173 @@
...
+@ExtendWith(WorkDirectoryExtension.class)
+class FileReceiverTest {
+    private static final int CHUNK_SIZE = 1024;
+
+    @WorkDirectory
+    private Path workDir;
+
+    @Test
+    void receiveSingleFile() throws IOException {
+        // when a single file is sent
```
ibessonov, line 54:
```suggestion
        // When a single file is sent.
```

ibessonov: You get the idea

modules/file-transfer/src/test/java/org/apache/ignite/internal/network/file/FileReceiverTest.java
```diff
@@ -0,0 +1,173 @@
...
+        List<File> filesToSend = List.of(randomFile(workDir, CHUNK_SIZE));
+        FileReceiver receiver = new FileReceiver("node1", 10);
+        UUID transferId = UUID.randomUUID();
+
+        Path path = Files.createDirectory(workDir.resolve(transferId.toString()));
+        path.toFile().deleteOnExit();
```
ibessonov, line 60: But why? Work dir should be deleted after tests. If it's not, we should fix the extension itself

modules/file-transfer/src/test/java/org/apache/ignite/internal/network/file/FileTransferMessagesHandlerTest.java
```diff
@@ -0,0 +1,145 @@
...
+import org.apache.ignite.internal.testframework.WorkDirectoryExtension;
+import org.apache.ignite.network.NetworkMessage;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+@ExtendWith(WorkDirectoryExtension.class)
+class FileTransferMessagesHandlerTest {
+    private static final int CHUNK_SIZE = 1024; // 1 KB
```
ibessonov, line 47: We know it's 1 KB :)
In every other place there's no such comment, why here?

modules/file-transfer/src/testFixtures/java/org/apache/ignite/internal/network/file/FileGenerator.java
```diff
@@ -0,0 +1,48 @@
...
+     * @param dir Directory to create the file in.
+     * @param contentSize Size of the file content.
+     * @return Generated file.
+     */
+    public static File randomFile(Path dir, int contentSize) {
+        try {
+            Path tempFile = Files.createTempFile(dir, null, null);
+            byte[] bytes = new byte[contentSize];
+            new Random().nextBytes(bytes);
+            tempFile.toFile().deleteOnExit();
```
ibessonov, line 42: Once again, what's the reason?

modules/file-transfer/src/testFixtures/java/org/apache/ignite/internal/network/file/FileGenerator.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public class FileGenerator {
+    /**
+     * Generates a random file with the given content size. The file be deleted on exit.
+     *
+     * @param dir Directory to create the file in.
+     * @param contentSize Size of the file content.
+     * @return Generated file.
+     */
+    public static File randomFile(Path dir, int contentSize) {
```
ibessonov, line 37: Why do you use both old and new file API in the same signature?

modules/core/src/main/java/org/apache/ignite/internal/util/FilesUtils.java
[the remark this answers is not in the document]
ibessonov: it throws "file not found" exception instead of ignoring missing files, as it should according to its name

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileConsumer.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Handler for the uploaded file.
+ */
+public interface FileConsumer<M extends Metadata> {
+    /**
+     * Handles the uploaded files.
+     *
+     * @param metadata The metadata of the uploaded file.
+     * @param uploadedFiles The temporary files that were uploaded. These files will be deleted after the method returns.
```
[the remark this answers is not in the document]
ibessonov: Is that a good idea to use temp dir? It may not have enough space, we should always use explicitly configured directories to write any data, like "workDir", for example. You can't make such assumption about user's environment.

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
ibessonov: What I see right now is two channels with matching type, it's a bug, that's why I asked. If it will be replaced in the future - the type will also be changed in the futures accordingly

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
ibessonov: I don't see how this clarifies my question. You're doing "messagingService.respond(...).thenCompose(...)", how is it related to executors in FileSender/FileReceiver? Do we guarantee that "handler.result()" will never be completed in that case, so that "handleUpload" is in its proper pool? I'd love to see that in a comment

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/MessagesUtils.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /**
+     * Extracts file headers from files.
+     *
+     * @param factory File transfer factory.
+     * @param paths List of paths.
+     * @return List of file headers.
+     */
+    static List<FileHeader> getHeaders(FileTransferFactory factory, List<Path> paths) {
```
ibessonov, line 43: Is this reasonable to extract such small method into utilities?
And why do we store the index of an element inside of the element? What's its purpose, it can be derived from the list

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/RateLimiter.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.network.file;
+
+/**
+ * Rate limiter interface.
+ */
+public interface RateLimiter {
```
ibessonov, line 23: I don't think that this interface is necessary at all

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/messages/FileHeader.java
```diff
@@ -0,0 +1,53 @@
...
+     * Returns the size of the file in bytes.
+     *
+     * @return File size.
+     */
+    long length();
+
+    /**
+     * Creates a new {@link FileHeader} instance from the given {@link File}.
+     */
+    static FileHeader fromPath(FileTransferFactory factory, int id, Path path) {
```
ibessonov, line 47: Wait a minute, you don't even use this "id". Please remove it then

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/messages/Identifier.java
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.ignite.internal.network.file.FileProvider;
+import org.apache.ignite.network.NetworkMessage;
+import org.apache.ignite.network.annotations.Transferable;
+
+/**
+ * Files identifier. This interface is used to mark all identifier messages. Identifier messages are used to retrieve files from
+ * {@link FileProvider} and handle them on the receiving side by {@link FileConsumer}.
+ */
+@Transferable(FileTransferMessageType.FILE_IDENTIFIER)
+public interface Identifier extends NetworkMessage {
```
ibessonov, line 30: Marker interface doesn't need to be transferable itself

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileSender.java
```diff
@@ -0,0 +1,218 @@
...
+     */
+    private CompletableFuture<Void> sendMessagesFromStream(
+            String receiverConsistentId,
+            FileChunkMessagesStream stream,
+            AtomicBoolean shouldBeCancelled
+    ) {
+        try {
+            if (stream.hasNextMessage() && !shouldBeCancelled.get()) {
+                return messagingService.send(receiverConsistentId, FILE_TRANSFER_CHANNEL, stream.nextMessage())
+                        .thenCompose(ignored -> sendMessagesFromStream(receiverConsistentId, stream, shouldBeCancelled));
```
ibessonov, line 138: What thread will execute this method?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileSender.java
```diff
@@ -0,0 +1,227 @@
...
+
+        CompletableFuture[] results = transfers.stream().map(it -> it.result)
+                .map(it -> it.whenComplete((v, e) -> {
+                    if (e != null) {
+                        shouldBeCancelled.set(true);
+                    }
+                }))
+                .toArray(CompletableFuture[]::new);
+
+        transfers.forEach(this::processTransfer);
```
ibessonov: This code could be much simpler if "processTransfer" returned the "result" future. That way you wouldn't need additional "transfers" collection. _I think_
It's actually pretty difficult to understand the entire thing)

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileSender.java
```diff
@@ -0,0 +1,227 @@
...
+        return allOf(results);
+    }
+
+    /**
+     * Processes the given transfer. If the rate limiter is not available, the transfer will be processed later.
+     *
+     * @param transfer The transfer to process.
+     */
+    private void processTransfer(FileTransfer transfer) {
+        synchronized (rateLimiter) {
```
ibessonov, line 103: Please document, how and why you use this synchronization. If it protects the "requests" queue, then why do you use rate limiter as a monitor? Not very intuitive to me

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileSender.java
```diff
@@ -0,0 +1,227 @@
...
+import org.apache.ignite.network.MessagingService;
+
+/**
+ * A class that sends files to a node. It uses a rate limiter to limit the bandwidth used. It also uses a thread pool to send the files in
+ * parallel.
+ */
+class FileSender {
+    private final int chunkSize;
+
+    private final Semaphore rateLimiter;
```
ibessonov, line 47: So, what exactly are we limiting? The number of files that we send concurrently?
Can we send a single file concurrently (sending N chunks in parallel)?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileSender.java
```diff
@@ -0,0 +1,227 @@
...
+     * Sends the file to the node with the given consistent id.
+     *
+     * @param receiverConsistentId The consistent id of the node to send the file to.
+     * @param id The id of the file transfer.
+     * @param path The path of the file to send.
+     * @return A future that will be completed when the file is sent.
+     */
+    private CompletableFuture<Void> sendFile(String receiverConsistentId, UUID id, Path path, AtomicBoolean shouldBeCancelled) {
+        if (path.toFile().length() == 0) {
+            return completedFuture(null);
```
ibessonov, line 164: Why don't we send empty files?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileSender.java
```diff
@@ -0,0 +1,227 @@
...
+        } else {
+            return supplyAsync(() -> {
+                try {
+                    return FileChunkMessagesStream.fromPath(chunkSize, id, path);
+                } catch (IOException e) {
+                    throw new FileTransferException("Failed to create a file transfer stream", e);
+                }
+            }, executorService)
+                    .thenCompose(stream -> {
```
ibessonov, line 173: What's the deal with "thenCompose"? You could put this code right into "supplyAsync"?
Please explain, don't just close the conversation.

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -242,87 +249,104 @@ private void processUploadRequest(FileUploadRequest message, String senderConsis
...
                 });
     }
 
-    private void processFileChunkMessage(FileChunkMessage message) {
-        runAsync(() -> fileReceiver.receiveFileChunk(message), executorService);
+    private void processFileChunkMessage(FileChunkMessage message, String senderConsistentId, long correlationId) {
+        runAsync(() -> fileReceiver.receiveFileChunk(message), executorService)
+                .whenComplete((v, e) -> {
+                    if (e != null) {
+                        LOG.error("Failed to receive file chunk. Transfer ID: {}", e, message.transferId());
```
ibessonov, line 332: What does it mean, "failed to receive"? You have received it, but wasn't able to process the data. It's not the same

ibessonov: By the way, I don't think that this is how we should format our log messages, usually we have all the extra data in brackets, not as a separate sentence. This comment applied to a lot of your log messages

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -370,7 +394,7 @@ public <M extends Identifier> void addFileConsumer(
             Class<M> identifier,
             FileConsumer<M> consumer
     ) {
-        metadataToHandler.compute(
+        metadataToConsumer.compute(
```
ibessonov, line 397: Why did you renamed "handler" to "consumer"? Consumer is too broad of a term

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -390,75 +414,64 @@ public CompletableFuture<List<Path>> download(String sourceNodeConsistentId, Ide
...
+    }
+
+    private CompletableFuture<Void> transferFilesToNode(
+            String targetNodeConsistentId,
+            UUID transferId,
+            Identifier identifier,
+            List<Path> paths
+    ) {
+        if (paths.isEmpty()) {
+            return failedFuture(new FileTransferException("No files to upload"));
```
ibessonov, line 448: Why is this a problem? What happens if user tries to do this, will he have an error in his log? It's not what he would expect

modules/file-transfer/src/test/java/org/apache/ignite/internal/network/file/FileSenderTest.java
```diff
@@ -106,25 +116,61 @@ void sendMultipleFiles() {
...
 
-        });
+        // When.
+        Path randomFile = FileGenerator.randomFile(workDir, CHUNK_SIZE * 5);
+        UUID transferId = UUID.randomUUID();
+        FileSender sender = new FileSender(
+                CHUNK_SIZE,
+                new Semaphore(4),
+                RESPONSE_TIMEOUT, messagingService,
+                Executors.newSingleThreadExecutor()
```
ibessonov, line 139: Where's the code that stops these executors?

modules/file-transfer/src/test/java/org/apache/ignite/internal/network/file/FileSenderTest.java
```diff
@@ -170,7 +219,7 @@ void maxConcurrentRequestsLimitIsNotExceeded() {
         FileSender sender = new FileSender(
                 CHUNK_SIZE,
                 new Semaphore(maxConcurrentRequests),
-                messagingService,
+                RESPONSE_TIMEOUT, messagingService,
```
ibessonov, line 222: Every parameter is on a new line, please fox other such places as well

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/ChunkedFileWriter.java
```diff
@@ -17,99 +17,87 @@
...
+            throw new FileValidationException("Chunk number mismatch: expected " + expectedNextChunkNumber + ", actual " + chunk.number());
         }
 
-        if (fileSize != UNKNOWN_FILE_SIZE && raf.getFilePointer() > fileSize) {
-            throw new FileValidationException("File size exceeded: expected " + fileSize + ", actual " + raf.getFilePointer());
-        }
-    }
+        lock.lock();
+        try {
+            stream.write(chunk.data());
```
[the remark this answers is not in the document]
ibessonov: Ok, my bad, I mistaken this method for another, I apologize

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
ibessonov: What error will a user see in this case? If it's an error message in a server log, then it's a problem. If it's an exception message in his application, then I'm fine with it

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/ChunkedFileReader.java
```diff
@@ -81,17 +81,21 @@ boolean hasNextChunk() {
...
         }
 
         int toRead = (int) Math.min(chunkSize, length - offset);
         byte[] data = new byte[toRead];
-        stream.read(data);
+        int read = stream.read(data);
+
+        if (read != toRead) {
+            throw new IOException("Failed to read chunk data from file: expected " + toRead + ", actual " + read + "]");
+        }
```
ibessonov, line 97: This is not what I meant, you should read in a loop.
-1 would mean an unexpected end of file, but you don't check for it exclusively. Please use the API properly

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/FileTransferServiceImpl.java
```diff
@@ -244,69 +254,84 @@ private void processFileTransferInitMessage(FileTransferInitMessage message, Str
...
-            }
-        });
-
-        FileTransferInitResponse response = messageFactory.fileTransferInitResponse().build();
-
-        messagingService.respond(senderConsistentId, FILE_TRANSFER_CHANNEL, response, correlationId)
+        ));
+
+        // Check that directory was created successfully and send response to the sender.
+        directoryFuture.handle((directory, e) -> {
```
ibessonov, line 272: I know you're tired of me at this point, but does it really have to be one huge statement in the code?
Why not split it into several statements, by extracting temporary results into variables?
Why not extract some parts into methods?
Are you comfortable yourself with the code of such complexity?

modules/file-transfer/src/main/java/org/apache/ignite/internal/network/file/ChunkedFileReader.java
```diff
@@ -88,18 +81,12 @@ byte[] readNextChunk() throws IOException {
...
-            throw new IOException("Failed to read chunk data from file: expected " + toRead + ", actual " + read + "]");
-        }
+        byte[] data = stream.readNBytes(chunkSize);
 
-        offset += toRead;
         nextChunkNumber++;
+        hasNextChunk = stream.available() > 0;
 
-        if (offset == length) {
+        if (stream.available() == 0) {
```
ibessonov, line 89: I wonder why it's not a `!hasNextChunk`, but whatever.
I've been looking at this code for too long. If there are more problems, I probably won't notice them at this point

ibessonov (approved):

PR #2395, opened 2023-08-02
IGNITE-20129 Fix CMake configuration

isapego:
- Removed explicit CMake generator from Gradle config. This way it's up to the user system to decide wthich generator to use - now it works on Windows as well;
- Added `CMAKE_CONFIGURATION_TYPES` for multi config generators (this is apparently a root cause of the initial issue);
- GTest now only downloaded and used if tests are enabled.


ptupitsyn (approved):

PakhomovAlexander (approved):
Checked on my laptop, LGTM;

PR #2413, opened 2023-08-04
IGNITE-20167 Always use version to get SQL schema

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-20167

In order to streamline the review of the patch and ensure better code quality
we ask both an author and a reviewer to check the following requirements:

### The Review Checklist
- [ ] **Formal criteria:** TC status, codestyle, mandatory documentation. Also check the following list:  
\- The description of the JIRA ticket explains WHAT and WHY was made and HOW.  
- [ ] **Patch quality:** patch cannot be split into a smaller pieces, its size is reasonable.
- [ ] **Tests code quality:** tests set covers positive/negative scenarios, happy/edge cases. Tests are effective in terms of execution time and resources.

### Notes


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestNode.java
```diff
@@ -103,7 +103,7 @@ public class TestNode implements LifecycleAware {
         this.nodeName = nodeName;
         var ps = new PrepareServiceImpl(nodeName, 0, mock(DdlSqlToCommandConverter.class), PLANNING_TIMEOUT, mock(MetricManager.class));
         this.prepareService = registerService(ps);
-        this.schema = schemaManager.schema("PUBLIC");
+        this.schema = schemaManager.schema("PUBLIC", -1);
```
tkalkirill, line 106: Please replace with `CatalogService#DEFAULT_SCHEMA_NAME`

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/PredefinedSchemaManager.java
```diff
@@ -61,35 +60,23 @@ public class PredefinedSchemaManager implements SqlSchemaManager {
                     schema.getTableNames().stream()
                             .map(schema::getTable)
                             .map(IgniteTable.class::cast)
-                            .collect(Collectors.toMap(IgniteTable::id, Function.identity()))
+                            .collect(CollectionUtils.toIntMapCollector(IgniteTable::id, Function.identity()))
```
tkalkirill, line 63: please use static imports

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/PredefinedSchemaManager.java
```diff
@@ -61,35 +60,23 @@ public class PredefinedSchemaManager implements SqlSchemaManager {
...
 
-    /** {@inheritDoc} */
-    @Override
-    public SchemaPlus schema(@Nullable String schema) {
-        return schema == null ? root : root.getSubSchema(schema);
-    }
-
     /** {@inheritDoc} */
     @Override
     public SchemaPlus schema(String name, int version) {
```
tkalkirill, line 70: Please note that the **name** may be `null` in the base class and its implementations.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/SqlSchemaManager.java
```diff
@@ -39,18 +34,15 @@ public interface SqlSchemaManager {
...
-
-    /**
-     * Returns a required schema if specified, or default schema otherwise.
+     * Returns a future to wait for given SQL schema version readiness.
+     *
+     * @param version SQL schema version to wait.
+     * @return Operation future.
      */
-    SchemaPlus activeSchema(@Nullable String name, long timestamp);
+    CompletableFuture<?> schemaReadyFuture(long version);
```
tkalkirill, line 47: I am confused by the type of the return value of the future, I propose to change it to `Void`, we will wait for the version of the scheme and then we will receive it separately as we need.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -443,9 +444,15 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
...
+                    InternalTransaction currentTx = implicitTxRequired ? txManager.begin(!rwOp, null) : outerTx;
 
-                    SchemaPlus schema = sqlSchemaManager.schema(schemaName);
+                    tx.set(currentTx);
+
+                    // TODO IGNITE-18733: wait for actual metadata for TX.
+                    HybridTimestamp txTimestamp = currentTx.startTimestamp();
+                    int observableCatalogVersion = catalogManager.activeCatalogVersion(txTimestamp.longValue());
+
+                    SchemaPlus schema = sqlSchemaManager.schema(schemaName, observableCatalogVersion);
```
tkalkirill, line 455: This is where I get confused too.
We get the catalog version and use it as the schema version, which may not be true.
Maybe we need, by analogy with the catalog, to wait for the schema version by `txTimestamp`?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -239,13 +239,13 @@ private AsyncCursor<List<Object>> executeQuery(
         return queryManager.execute(tx, plan);
     }
 
-    private BaseQueryContext createQueryContext(UUID queryId, @Nullable String schema, Object[] params) {
+    private BaseQueryContext createQueryContext(UUID queryId, long version, @Nullable String schema, Object[] params) {
```
tkalkirill, line 242: `long version` -> `long schemaVersion`

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/schema/SqlSchemaManagerTest.java
```diff
@@ -147,7 +143,7 @@ public void testOnTableDroppedHandler() {
 
         testRevisionRegister.moveForward();
 
-        Table schemaTable = sqlSchemaManager.schema("PUBLIC").getTable("T");
+        Table schemaTable = sqlSchemaManager.latestSchema("PUBLIC").getTable("T");
```
tkalkirill, line 146: Please use `CatalogService#DEFAULT_SCHEMA_NAME`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
AMashenkov: Ok, we wait for catalog version and want to get corresponding SQL schema.
SchemaPlus is Calcite class and we can' rename it.
I can add a comment.... WDYT?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/CatalogSqlSchemaManager.java
```diff
@@ -70,7 +71,7 @@ public CatalogSqlSchemaManager(CatalogManager catalogManager, int cacheSize) {
 
     /** {@inheritDoc} */
     @Override
-    public SchemaPlus schema(String name, int version) {
+    public @Nullable SchemaPlus schema(String name, int version) {
```
tkalkirill, line 74: `String name` can be `null` not return value =)

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/PredefinedSchemaManager.java
```diff
@@ -60,21 +63,21 @@ public class PredefinedSchemaManager implements SqlSchemaManager {
                     schema.getTableNames().stream()
                             .map(schema::getTable)
                             .map(IgniteTable.class::cast)
-                            .collect(CollectionUtils.toIntMapCollector(IgniteTable::id, Function.identity()))
+                            .collect(toIntMapCollector(IgniteTable::id, Function.identity()))
```
tkalkirill, line 66: maybe also `Function.identity()` -> `identity()` wdyt?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
tkalkirill: Whether correctly I understand that the version of the scheme and the catalog are equal?

If yes, then it is necessary to correct the documentation of the method and the naming of variables, for example, obtaining the schema by the name and version of the directory.

If not, then you probably need to fix or create a ticket (indicate the existing one) that this will be fixed in the future.

At the moment, I assume that the versions of the schemas and the catalog will be different, since an entity not related to the schema, such as a zone, table, and index, may change in the catalog.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/schema/CatalogSqlSchemaManager.java
AMashenkov: Ok.
Return value also can be null if someone use non-existed schema name.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
AMashenkov: 1. `SqlSchemaManager` is a general interface, which provides method to get schema by name and version. It doesn't assumed any connection with Catalog, and whether version should match catalog version or not. 
I think we shouldn't fix any naming here.
2. `CatalogSqlSchemaManager is an implementation that backed by catalog. We assume schema version is mapped to catalog version as 1:1.
3. Actually, current implementation `SqlSchemaManager` does not support versions at all. So, any naming is possible here, as it make no sense.

Maybe, we should either use `CatalogSqlSchemaManager` as `sqlSchemaManager` type, or add a method to `SqlSchemaManager` interface that will resolve a version by tx timestamp. WDYT?

tkalkirill: > Maybe, we should either use CatalogSqlSchemaManager as sqlSchemaManager type, or add a method to SqlSchemaManager interface that will resolve a version by tx timestamp. WDYT?

Yes, I like the option with an additional method.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -330,7 +330,10 @@ private AsyncCursor<List<Object>> executeExplain(ExplainPlan plan) {
     private void onMessage(String nodeName, QueryStartRequest msg) {
         assert nodeName != null && msg != null;
 
-        CompletableFuture<?> fut = sqlSchemaManager.actualSchemaAsync(msg.schemaVersion());
+        //TODO IGNITE-18733: We should use txTimestamp from message tx attributes to wait for metadata ready,
+        // then get actual version for txTimestamp and wait for catalog version ready.
+        // As optimization, we can lookup for latest available schema and go sync way if it's version >= desirable.
```
korlov42, line 335: in my opinion, this is not quite correct. It's better to use catalog version explicitly whenever possible

korlov42 (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
AMashenkov: Do you mean we shouldn't get rid of catalogVersion in message?
We already have an TxTimestmap that unambiguously maps to catalog version.

PR #2415, opened 2023-08-07
IGNITE-20114 DistributionZoneManager should listen CatalogService events instead of configuration

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-20114

### The Review Checklist

### Notes


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/ddl/DistributionZoneDdlCommandHandlerTest.java
```diff
@@ -51,21 +47,14 @@ public class DistributionZoneDdlCommandHandlerTest extends IgniteAbstractTest {
...
 
-    private DistributionZoneManager distributionZoneManager;
-
     private CatalogManager catalogManager;
 
     @BeforeEach
     void before() {
-        distributionZoneManager = mock(DistributionZoneManager.class, invocation -> completedFuture(null));
+        catalogManager = mock(CatalogManager.class, invocation -> completedFuture(null));
         catalogManager = mock(CatalogManager.class, invocation -> completedFuture(null));
```
AMashenkov, line 55:
```suggestion
        catalogManager = mock(CatalogManager.class, invocation -> completedFuture(null));
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
```diff
@@ -1076,29 +1069,41 @@ private static void createZone(Node node, String zoneName, int partitions, int r
...
+                zoneName,
+                tableName,
+                List.of(
+                        ColumnParams.builder().name("key").type(org.apache.ignite.sql.ColumnType.INT64).build(),
+                        ColumnParams.builder().name("val").type(org.apache.ignite.sql.ColumnType.INT32).nullable(true).build()
+                ),
+                List.of("key")
+        );
+
+        TableDefinition tableDefinition = SchemaBuilders.tableBuilder(DEFAULT_SCHEMA_NAME, tableName).columns(
```
AMashenkov, line 1098: Why do you use Utility class to create a table instead of using CatalogManager directly?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/ddl/DistributionZoneDdlCommandHandlerTest.java
tkalkirill: fix it

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
tkalkirill: For convenience, when I started doing the task, I discovered that many tests created the zone directly through the manager instead of using auxiliary methods, which introduced quite a lot of unpleasant refactorings.

To reduce refactoring in the future, I created this method.

ibessonov (approved):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
AMashenkov: I don't understand how this approach could help if smth will be changed in future.
Having a shortcut method for building CreateTableParams only looks ok.
But, let's call CatalogManager explicitly.

```suggestion
        CreateTableParams params = TableTestUtils.createTable(         
                DEFAULT_SCHEMA_NAME,
                zoneName,
                tableName,
                List.of(
                        ColumnParams.builder().name("key").type(org.apache.ignite.sql.ColumnType.INT64).build(),
                        ColumnParams.builder().name("val").type(org.apache.ignite.sql.ColumnType.INT32).nullable(true).build()
                ),
                List.of("key")
        );
        node.catalogManager().createTable(params);
      
        TableDefinition tableDefinition = SchemaBuilders.tableBuilder(DEFAULT_SCHEMA_NAME, tableName).columns(
```

tkalkirill: It took me quite a bit of effort to fix the tests and change them to the new API.
I was helped in this by using methods instead of explicitly calling the API.
Since this method has proven to be more effective so far (it will also reduce the pain of trying to merge `catalog-feature` into `main`), I will stick with it and try to stick with it.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -621,174 +299,110 @@ public CompletableFuture<Set<String>> dataNodes(long causalityToken, int zoneId)
...
-        public CompletableFuture<?> onDelete(ConfigurationNotificationEvent<DistributionZoneView> ctx) {
-            int zoneId = ctx.oldValue().zoneId();
-
-            long revision = ctx.storageRevision();
-
-            ZoneState zoneState = zonesState.get(zoneId);
+        vaultMgr.put(zonesFilterUpdateRevision(), longToBytes(causalityToken)).join();
 
-            zoneState.stopTimers();
+        saveDataNodesToMetaStorageOnScaleUp(zoneId, causalityToken);
```
sanpwc, line 390: Well, it's not yours however in order to prevent ms invocations reordering we should return saveDataNodesToMetaStorageOnScaleUp result future as onAnyCallback result.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -798,11 +412,11 @@ private void createOrRestoreZoneState(DistributionZoneView zone, long revision)
 
             ZoneState prevZoneState = zonesState.putIfAbsent(zoneId, zoneState);
 
-            assert prevZoneState == null : "Zone's state was created twice [zoneId = " + zoneId + ']';
+            assert prevZoneState == null : "Zone's state was created twice: zoneId=" + zoneId;
```
sanpwc, line 415: Probably I've missed that. We should not use square braces any longer? Same is about spaces before and after '='.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -282,25 +253,10 @@ public DistributionZoneManager(
...
-            zonesConfiguration.distributionZones().any().filter().listen(onUpdateFilter());
-
-            zonesConfiguration.defaultDistributionZone().listen(zonesConfigurationListener);
-            zonesConfiguration.defaultDistributionZone().dataNodesAutoAdjustScaleUp().listen(onUpdateScaleUp());
-            zonesConfiguration.defaultDistributionZone().dataNodesAutoAdjustScaleDown().listen(onUpdateScaleDown());
-            zonesConfiguration.defaultDistributionZone().filter().listen(onUpdateFilter());
+        inBusyLock(busyLock, () -> {
+            registerCatalogEventListenersOnStartManagerBusy();
 
             rebalanceEngine.start();
```
sanpwc, line 261: I would say that we should start rebalanceEngine at the very end of DZM.start, or more formally after DZM local state recovery.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1945,29 +1365,84 @@ public Set<NodeWithAttributes> logicalTopology() {
...
+
+        catalogManager.listen(ZONE_ALTER, new ManagerCatalogAlterZoneEventListener());
+    }
+
+    private void startZonesOnStartManagerBusy() {
+        CompletableFuture<Long> recoveryFinishedFuture = metaStorageManager.recoveryFinishedFuture();
+
+        assert recoveryFinishedFuture.isDone();
+
+        int catalogVersion = catalogManager.latestCatalogVersion();
```
sanpwc, line 1395: Is it possible for latestCatalogVersion to be increased during DZM startup or it might be considered as an immutable value during startup period?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -1945,29 +1365,84 @@ public Set<NodeWithAttributes> logicalTopology() {
...
+        long causalityToken = recoveryFinishedFuture.join();
+
+        for (CatalogZoneDescriptor zone : catalogManager.zones(catalogVersion)) {
```
sanpwc, line 1398: We should also cleanup all abandoned resources. Both local (vault) and distributed (ms based keys). I would rather create two separate tickets for both that items in order not to clutter up given PR, however they should be ranked as important ones.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
tkalkirill: fix it

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
tkalkirill: returned it(

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
tkalkirill: fix it

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
tkalkirill: Like manager `VaultManager`, `CatalogManager` starts before manager `TableManager`/`IndexManager`/`DistributionZoneManager` and can be changed (not immutable).

To avoid manager changes, you can do the following:

1. Slightly rework `IgniteComponent`, for managers that depend on the catalog it will add the version of the catalog on which we made the recovery;

2. Add a `CatalogService#catalogVersionAfterRecovery` that will throw an exception before recovery, and after that it will return an immutable value;

3. Make the `CatalogManager` immutable at the start phase of all other components, and then make it mutable.

I like point 2 and 3.
It might be worth discussing this with @AMashenkov as well.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
tkalkirill: add todo

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
sanpwc: That seems as an important one. From my point of view latestCatalogVersion retrieved within recovery should be consistent with metaStrorage."recoveryRevision" and thus it should be immutable because recovery revision is immutable. Are you going to fix it in this ticket or create another one?

tkalkirill: I suggest fixing it in another ticket.

PR #2429, opened 2023-08-09
IGNITE 19148 Switch tests from using JUL to log4j2

sk0x50:
https://issues.apache.org/jira/browse/IGNITE-19148

### The Review Checklist

### Notes


modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/log4j2/TestLogChecker.java
```diff
@@ -0,0 +1,321 @@
...
+     *
+     * @param loggerName Logger name.
+     * @param handlers List of handlers to be added.
+     */
+    public TestLogChecker(String loggerName, Handler... handlers) {
+        this(loggerName);
+
+        lock.writeLock().lock();
+        try {
+            for (Handler handler : handlers) {
```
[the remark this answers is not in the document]
sk0x50: Thanks. Fixed

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/log4j2/TestLogChecker.java
```diff
@@ -0,0 +1,324 @@
...
+     * Adds a new handler with the given {@code predicate} and {@code action}.
+     *
+     * @param predicate Predicate to check log messages.
+     * @param action Action to be executed when the {@code predicate} is matched.
+     * @return New instance of {@link Handler}.
+     */
+    public Handler addHandler(Predicate<String> predicate, Runnable action) {
+        Handler handler = new Handler(predicate, action);
+
+        lock.writeLock().lock();
```
rpuch, line 160: It seems we can just invoke `addHandler(handler)` here

modules/network/src/integrationTest/java/org/apache/ignite/network/scalecube/ItScaleCubeNetworkMessagingTest.java
```diff
@@ -87,10 +86,15 @@ class ItScaleCubeNetworkMessagingTest {
     /** Message factory. */
     private final TestMessagesFactory messageFactory = new TestMessagesFactory();
 
+    /** List of test loggers. */
+    private final List<TestLogChecker> loggers = new ArrayList<>();
```
rpuch, line 90: Those are not loggers, they are checkers

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
```diff
@@ -734,18 +728,14 @@ private BiPredicate<String, NetworkMessage> dropSnapshotMetaResponse(Completable
     void snapshotInstallTimeoutDoesNotBreakSubsequentInstallsWhenSecondAttemptIsIdenticalToFirst() throws Exception {
         AtomicBoolean snapshotInstallFailedDueToIdenticalRetry = new AtomicBoolean(false);
 
-        Logger snapshotExecutorLogger = Logger.getLogger(SnapshotExecutorImpl.class.getName());
+        TestLogChecker snapshotExecutorLogger = TestLogChecker.create(SnapshotExecutorImpl.class);
```
rpuch, line 731: This is not a logger

rpuch (commented):
What's the reason to change from JUL to log4j2? The JIRA only says about logging to stderr, but JUL can be configured to use any appenders (or how are they called there?) we want. So why?

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/log4j2/TestLogChecker.java
```diff
@@ -0,0 +1,324 @@
...
+import org.apache.logging.log4j.core.Logger;
+import org.apache.logging.log4j.core.appender.AbstractAppender;
+import org.apache.logging.log4j.core.config.Configuration;
+import org.apache.logging.log4j.core.config.LoggerConfig;
+import org.apache.logging.log4j.core.config.Property;
+
+/**
+ * Test log checker.
+ */
+public class TestLogChecker {
```
rpuch, line 40: The name does not seem perfect. 'Check' is too abstract; also this class not just 'checks' incoming log messages, but it also reacts to them somehow. Maybe a better name should be devised (but I cannot come up with anything good now)

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/commands/ItClusterCommandTest.java
```diff
@@ -70,42 +67,34 @@ class ItClusterCommandTest extends AbstractCliTest {
...
 
-    private static final Logger topologyLogger = Logger.getLogger("org.apache.ignite.network.scalecube.ScaleCubeTopologyService");
-
     @BeforeEach
     void setup(@WorkDirectory Path workDir, TestInfo testInfo) throws Exception {
         CountDownLatch allNodesAreInPhysicalTopology = new CountDownLatch(1);
 
-        Handler physicalTopologyWaiter = physicalTopologyWaiter(allNodesAreInPhysicalTopology);
-        topologyLogger.addHandler(physicalTopologyWaiter);
+        TestLogChecker topologyLogger = new TestLogChecker(
```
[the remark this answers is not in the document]
sk0x50: You are absolutely right. Will fix it here and in other places.

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/log4j2/TestLogChecker.java
```diff
@@ -0,0 +1,324 @@
...
+import org.apache.logging.log4j.core.Appender;
+import org.apache.logging.log4j.core.LogEvent;
+import org.apache.logging.log4j.core.Logger;
+import org.apache.logging.log4j.core.appender.AbstractAppender;
+import org.apache.logging.log4j.core.config.Configuration;
+import org.apache.logging.log4j.core.config.LoggerConfig;
+import org.apache.logging.log4j.core.config.Property;
+
+/**
+ * Test log checker.
```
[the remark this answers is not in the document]
sk0x50: Javadoc was added.

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/log4j2/TestLogChecker.java
```diff
@@ -0,0 +1,324 @@
...
+        this.appender = new TestLogAppender(loggerName);
+    }
+
+    /**
+     * Creates a new instance of {@link TestLogChecker} with the given {@code predicate}.
+     *
+     * @param loggerName Logger name.
+     * @param predicate Predicate to check log messages.
+     */
+    public TestLogChecker(String loggerName, Predicate<String> predicate) {
```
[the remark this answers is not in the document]
sk0x50: > What happens if the predicate returns true or false?
updated javadoc in the following way:
```
    /**
     * Creates a new instance of {@link TestLogChecker} with the given {@code predicate}.
     * In order to check that the required log event was logged, use {@link #isMatched()} method.
     *
     * @param loggerName Logger name.
     * @param predicate Predicate to check log messages.
     */
```

> Also, in some cases it is required to verify other message attributes, not only its text
It makes sense to me. Predicate should accept `LogEvent` instead of `String`

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/log4j2/TestLogChecker.java
sk0x50: Thanks. Fixed.

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/log4j2/TestLogChecker.java
```diff
@@ -0,0 +1,324 @@
...
+     * @param loggerName Logger name.
+     * @param predicate Predicate to check log messages.
+     * @param action Action to be executed when the {@code predicate} is matched.
+     */
+    public TestLogChecker(String loggerName, Predicate<String> predicate, Runnable action) {
+        this(loggerName, new Handler(predicate, action));
+    }
+
+    /**
+     * Creates a new instance of {@link TestLogChecker} with the given {@code predicate} and {@code action}.
```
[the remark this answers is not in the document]
sk0x50: Thanks. Fixed.

sk0x50 (commented):
Hi @rpuch,

I agree that this is my fault and I had to describe the motivation in the ticket.

IMHO
 - It would be better to use a well-known and widely-used logging framework instead of JUL, which should not require any additional code like a custom appender, etc. I don't think that JUL is used in real-life solutions, so this patch can be considered a test scenario to use 3rd party implementation of the logging framework (For example, scalecube, which is an internal thing, uses slf4j and so we need to provide a bridge from slf4j to log4j2/logback. this step is not so obvious and has to be described in our documentation).
 - log4j2 provides a color schema for the logs, which is a nice feature :)
 
By the way, I don't like our approach of using logs to check something like cluster initialization, snapshots, etc. It should be done via special lifecycle events, callbacks, whatever... but using logs for this looks weird to me.

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/log4j2/TestLogChecker.java
```diff
@@ -35,7 +35,51 @@
...
+ *     try {
+ *         // do something
+ *     } finally {
+ *         checker.stop();
+ *     }
+ *
+ *     assertThat(messageCounter.get(), is(42));
+ * </code></pre>
+ * <p>
+ * It is possible to add a new pair of of predicate and action using {@link #addHandler(Predicate, Runnable)},
```
rpuch, line 80:
```suggestion
 * It is possible to add a new pair of predicate and action using {@link #addHandler(Predicate, Runnable)},
```

rpuch (approved):

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/log4j2/TestLogChecker.java
rpuch: > By the way, I don't like our approach of using logs to check something like cluster initialization, snapshots, etc. It should be done via special lifecycle events, callbacks, whatever... but using logs for this looks weird to me.

I agree, this sometimes looks weird. But at times we need to know something very specific, and we don't always have (and should not have, probably) such a specific event. Also, logs can be considered an 'event recording' system, so this approach does not look too innatural :)

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/log4j2/TestLogChecker.java
sk0x50: Renamed to `LogInspector`

PR #2437, opened 2023-08-10
IGNITE-20035 Fixed handling maxrows jdbc query parameter.

xtern:
https://issues.apache.org/jira/browse/IGNITE-20035

### DIFF
--- a/modules/client-handler/src/test/java/org/apache/ignite/client/handler/requests/jdbc/JdbcQueryCursorSelfTest.java
+++ b/modules/client-handler/src/test/java/org/apache/ignite/client/handler/requests/jdbc/JdbcQueryCursorSelfTest.java
@@ -34,7 +34,7 @@ import org.junit.jupiter.params.provider.ValueSource;
 /**
  * Test class for {@link JdbcQueryCursor}.
  */
-public class JdbcQueryCursorTest {
+public class JdbcQueryCursorSelfTest {
     private static final List<Integer> ROWS = List.of(1, 2, 3);
 
     private static final int TOTAL_ROWS_COUNT = ROWS.size();
@@ -58,7 +58,7 @@ public class JdbcQueryCursorTest {
         assertEquals(ROWS, results);
     }
 
-    private List<Integer> fetchFullBatch(int maxRows, int fetchSize) {
+    private static List<Integer> fetchFullBatch(int maxRows, int fetchSize) {
         JdbcQueryCursor<Integer> cursor = new JdbcQueryCursor<>(maxRows,
                 new AsyncSqlCursorImpl<>(SqlQueryType.QUERY, null, null,
                         new AsyncWrapper<>(CompletableFuture.completedFuture(ROWS.iterator()), Runnable::run)));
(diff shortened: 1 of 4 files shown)


modules/client-handler/src/test/java/org/apache/ignite/client/handler/requests/jdbc/JdbcQueryCursorSelfTest.java
```diff
@@ -0,0 +1,91 @@
...
+        do {
+            requestResult = cursor.requestNextAsync(fetchSize).join();
+
+            results.addAll(requestResult.items());
+        } while (requestResult.hasMore());
+
+        if (maxRows == 0 || maxRows >= TOTAL_ROWS_COUNT) {
+            assertEquals(rows, results);
+        } else {
+            assertEquals(rows.subList(0, maxRows), results);
```
lowka, line 63: I think It would be better to split if this test case were split into two - one for maxRows = 0 and another for maxRows != 0 by extracting a driver method that accepts (rows for iteration + max rows and returns a `BatchedResult` ) and does all the necessary setup + loops over  `requestResult`. Because right know this test case is almost identical to implementation.

modules/client-handler/src/test/java/org/apache/ignite/client/handler/requests/jdbc/JdbcQueryCursorSelfTest.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+    private static List<Arguments> maxRowsTestParameters() {
+        int[] fetchSizes = {1024, 1, 0, -1};
```
lowka, line 68: Negative number of `rows` parameter looks like something that should not be allowed. As we have a counter that limits the number of returned records, it seem that we should to allow negative values for `rows` otherwise maxRows makes a little sense: 
initially: maxRows = 10, fetched = 0

0: fetch(8): fetched= 8
1: fetch(-5): fetched=3 

Or I am missing something here?

modules/client-handler/src/test/java/org/apache/ignite/client/handler/requests/jdbc/JdbcQueryCursorSelfTest.java
xtern: Split into two tests, thanks.

modules/client-handler/src/test/java/org/apache/ignite/client/handler/requests/jdbc/JdbcQueryCursorSelfTest.java
xtern: I agree that negative fetch size looks very weird.
I've added the test parameters manually to make them more clear.

modules/client-handler/src/test/java/org/apache/ignite/client/handler/requests/jdbc/JdbcQueryCursorTest.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.ignite.internal.sql.engine.exec.AsyncWrapper;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.junit.jupiter.params.provider.ValueSource;
+
+/**
+ * Tests for for {@link JdbcQueryCursor}.
+ */
+public class JdbcQueryCursorTest {
```
AMashenkov, line 37:
```suggestion
/**
 * Test class for {@link JdbcQueryCursor}.
 */
public class JdbcQueryCursorSelfTest {
```

AMashenkov (approved):

lowka (approved):

PR #2449, opened 2023-08-15
IGNITE-19898 Sql. Added ability to use different transaction facades to start an implicit transaction.

xtern:
https://issues.apache.org/jira/browse/IGNITE-19898

ItSqlAsynchronousApiTest and ItSqlSynchronousApiTest have been fixed a bit to run sql queries from the same host, otherwise they will fail after propagating the observable timestamp from the client.


modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/jdbc/JdbcQueryCursor.java
```diff
@@ -86,4 +88,10 @@ public SqlQueryType queryType() {
     public ResultSetMetadata metadata() {
         return cur.metadata();
     }
+
+    /** {@inheritDoc} */
+    @Override
+    public @Nullable HybridTimestamp observableTimestamp() {
+        return cur.observableTimestamp();
```
AMashenkov, line 95: I think leaking internals into public API is not a good idea.
What should use do with this?

AMashenkov (changes requested):

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/jdbc/JdbcQueryCursor.java
xtern: I see two options:
- remove inheritance from SqlAsyncCursor
- add something like BaseSqlAsyncCursor interface with public methods.

What do you think?
p.s. I updated PR with option 1 (removed inheritance).

xtern: As was discussed privately - JdbcQueryCursor is not a part of the public API.

ygerzhedovich (approved):

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/jdbc/JdbcQueryCursor.java
ygerzhedovich: @AMashenkov can the 'change request' closed?

modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
```diff
@@ -76,7 +76,7 @@ protected IgniteException convertException(Throwable th) {
             return (IgniteException) th;
         }
 
-        //TODO: IGNITE-14500 Replace with public exception with an error code (or unwrap?).
+        //TODO: IGNITE-20181 KV/Binary view public API should only throw public exceptions for the end user.
```
xtern, line 79: The change is not related to the current pull request - just a fix to a wrong issue link.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -432,26 +424,16 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
...
-
-                    boolean rwOp = dataModificationOp(result);
-
-                    boolean implicitTxRequired = outerTx == null;
-
-                    InternalTransaction currentTx = implicitTxRequired ? txManager.begin(!rwOp, null) : outerTx;
+                    validateParsedStatement(context, result, params);
 
-                    tx.set(currentTx);
+                    HybridTimestamp txStartTs = txWrapper.beginTxIfNeeded(result.queryType());
```
AMashenkov, line 433:
```suggestion
                    Transaction tx = txWrapper.beginTxIfNeeded(result.queryType());
                    HybridTimestamp currentTx = tx == null ? clock.now() : tx.beginTimestamp();
```

xtern: Reworked

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/datatypes/tests/BaseDataTypeTest.java
```diff
@@ -125,6 +126,11 @@ protected QueryProcessor getEngine() {
                 return ((IgniteImpl) CLUSTER_NODES.get(0)).queryEngine();
             }
 
+            @Override
+            protected IgniteTransactions transactions() {
+                return CLUSTER_NODES.get(0).transactions();
```
ygerzhedovich, line 131:
```suggestion
                return igniteTx();
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -518,6 +505,25 @@ public CompletableFuture<Void> closeAsync() {
         return stage;
     }
 
+    private static QueryTransactionWrapper beginImplicitTxIfNeeded(
```
ygerzhedovich, line 508: I think such method should be part of QueryTransactionWrapper. And the QueryTransactionWrapper shouldn't have public constructors

xtern: :weary: But as far as I remember, we decided that this is wrong from an architectural point of view.

We decided that wrapper should not do any unclear things :roll_eyes: .

For example what is the return type of the following call?

wrapper.beginImplicitTxIfNeeded(..

`void` or `InternalTransaction` or... 'QueryTransactionWrapper`?

ygerzhedovich: it should be factory method `QueryTransactionWrapper`

xtern: Done https://github.com/apache/ignite-3/pull/2449/files#diff-d92eb2bb714b57a42aa2c745c30005875861cc19234e6e29cf5ed0a1b0973bdaR37

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -414,6 +413,8 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
 
         InternalTransaction outerTx = context.unwrap(InternalTransaction.class);
 
+        AtomicReference<Runnable> implicitTxRollbackAction = new AtomicReference<>(() -> {});
```
AMashenkov, line 416: Let's try to drop the AtomicReference. If impossible, then replace with tx wrapper (as it was in prev PR version)

ygerzhedovich (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -429,93 +430,120 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> querySingle0(
...
-                .thenCompose(ignored -> {
-                    ParsedResult result = parserService.parse(sql);
+            validateParsedStatement(context, result, params);
 
-                    validateParsedStatement(context, outerTx, result, params);
+            QueryTransactionWrapper txWrapper = wrapTxOrStartImplicit(result.queryType(), transactions, outerTx);
 
-                    boolean rwOp = dataModificationOp(result);
+            return waitForActualSchema(schemaName, txWrapper.unwrap().startTimestamp())
+                    .thenCompose(schema -> {
```
AMashenkov, line 443: `waitForActualSchema` future may completes in Metastorage thread.
So, `prepareAsync` or/and `executePlan` may be processed in Metastorage thread.
Should we use `thenComposeAsync` here?

AMashenkov (approved):

PR #2471, opened 2023-08-22
IGNITE-20057  Track observable timestamp in C++ client and ODBC

isapego:
Added storing and sending of observable timestamp to both ODBC and C++ client.


ptupitsyn (approved):
Looks good to me.

Can we add tests for this? Either integration causality (modify value, check within RO TX), or round-trip check?

isapego:
> Looks good to me.
> 
> Can we add tests for this? Either integration causality (modify value, check within RO TX), or round-trip check?

I'm not sure any of these tests will show anything.

PR #2473, opened 2023-08-22
IGNITE-20256 Refuse to install Raft snapshots on partitions when not enough schemas are available

rpuch:
https://issues.apache.org/jira/browse/IGNITE-20256

### The Review Checklist

### Notes


modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
```diff
@@ -792,6 +816,116 @@ void testChangeLeaderDuringSnapshotInstallationToLeaderWithEnoughLog() throws Ex
...
+
+        copierLogInspector.addHandler(
+                event -> event.getMessage().getFormattedMessage().startsWith("Metadata not yet available, URI '"),
+                installationRejected::countDown
+        );
+
+        return installationRejected;
+    }
+
+    private BlockingAppendEntriesRequestProcessor installBlockingAppendEntriesProcessor(int nodeIndex) {
```
ibessonov, line 908: These tests become more and more complicated with time, any ideas of improvements?
Too much reflection for my taste

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/PartitionSnapshotStorage.java
```diff
@@ -90,6 +94,7 @@ public PartitionSnapshotStorage(
             String snapshotUri,
             RaftOptions raftOptions,
             PartitionAccess partition,
+            CatalogService catalogService,
```
ibessonov, line 97: Why do we have a `CatalogAccess` abstraction, if we still use `CatalogService` instead? I don't really get it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -68,6 +71,9 @@ public class IncomingSnapshotCopier extends SnapshotCopier {
 
     private static final int MAX_TX_DATA_BATCH_SIZE = 1000;
 
+    /** Number of milliseconds that the follower is allowed to try to catch up the required catalog version. */
+    private static final int WAIT_FOR_METADATA_CATCHUP_MS = 3000;
```
ibessonov, line 75: Why don't we create proper configuration? What's up with all these constants?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -121,13 +127,41 @@ public void start() {
...
-                            .thenCompose(unused1 -> loadSnapshotMvData(snapshotSender, executor))
-                            .thenCompose(unused1 -> loadSnapshotTxData(snapshotSender, executor));
+                            // Give metadata some time to catch up as it's very probable that the leader is ahead metadata-wise.
+                            .thenCompose(unused2 -> waitForMetadataWithTimeout())
+                            .thenCompose(unused2 -> {
+                                if (metadataIsSufficientlyComplete()) {
+                                    return loadSnapshotMvData(snapshotSender, executor)
+                                            .thenCompose(unused1 -> loadSnapshotTxData(snapshotSender, executor));
+                                } else {
+                                    logMetadataIsInsufficiencyAndSetError();
```
ibessonov, line 137: "is insufficiency" - I think you don't need "is"

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -142,7 +176,7 @@ public void join() throws InterruptedException {
 
                 LOG.error("Error when completing the copier", cause);
 
-                if (!isOk()) {
+                if (isOk()) {
```
ibessonov, line 179: What?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -215,6 +249,28 @@ private CompletableFuture<?> loadSnapshotMeta(ClusterNode snapshotSender) {
         }
     }
 
+    private boolean metadataIsSufficientlyComplete() {
+        return partitionSnapshotStorage.catalogService().latestCatalogVersion() >= snapshotMeta.requiredCatalogVersion();
+    }
+
+    private void logMetadataIsInsufficiencyAndSetError() {
+        LOG.warn(
+                "Metadata not yet available, URI '{}', required level {}; rejecting snapshot installation.",
```
ibessonov, line 258: Is this the way we format logs? I thought it should be more like "Foo [a=1, b=2]", there a guidelines somewhere

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -215,6 +249,28 @@ private CompletableFuture<?> loadSnapshotMeta(ClusterNode snapshotSender) {
...
+                snapshotMeta.requiredCatalogVersion()
+        );
+
+        String errorMessage = String.format(
+                "Metadata not yet available, URI '%s', required level %s; rejecting snapshot installation.",
+                this.snapshotUri,
+                snapshotMeta.requiredCatalogVersion()
+        );
+
+        if (isOk()) {
```
ibessonov, line 269: Again, I don't quite get it, why do we print warnings/errors if everything's OK. Please comment this code, or fix it if it's wrong

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
```diff
@@ -792,6 +816,116 @@ void testChangeLeaderDuringSnapshotInstallationToLeaderWithEnoughLog() throws Ex
...
+     * rejected, and that, when metadata catches up, the snapshot gets successfully installed.
+     */
+    @Test
+    void laggingSchemasOnFollowerPreventSnapshotInstallation() throws Exception {
+        cluster.startAndInit(3);
+
+        createTestTableWith3Replicas(DEFAULT_STORAGE_ENGINE);
+
+        // Prepare the scene: force node 0 to be a leader, and node 2 to be a follower.
+
```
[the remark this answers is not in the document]
rpuch: Probably the idea was to give a hint that the comment is not about just next line, but about next fragment of method

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
rpuch: I wrote it as simply as I could. The test scenario is pretty twisted by its virtue. Any suggestions are welcome

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
rpuch: This one seems to be something really internal, I'm not sure it should be exposed and be tunable. Concerning other constants, we might create a JIRA ticket to think about them.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
rpuch: This is how it works in `LocalSnapshotCopier`. Otherwise, if no error gets registered from elsewhere, we will never register any error; but, if some error already happened somewhere, we will overwrite it. It seems better to only leave the first error.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
rpuch: It's something completely different. 'If there is no error registered yet, register an error; otherwise leave an earlier error to not overwrite it'. To be honest, I don't see what should be commented here: 'if still no error, raise error status'? But is seems obvious from the code :)

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
```diff
@@ -436,21 +457,23 @@ private void reanimateNode2AndWaitForSnapshotInstalled() throws InterruptedExcep
...
+        assertTrue(snapshotInstalledLatch.await(60, TimeUnit.SECONDS), "Did not install a snapshot in time");
+    }
+
+    private CountDownLatch snapshotInstalledLatch(int nodeIndex) {
         CountDownLatch snapshotInstalledLatch = new CountDownLatch(1);
 
-        Handler handler = replicatorLogInspector.addHandler(
+        replicatorLogInspector.addHandler(
                 evt -> evt.getMessage().getFormattedMessage().matches(
                         "Node .+ received InstallSnapshotResponse from .+_" + nodeIndex + " .+ success=true"),
```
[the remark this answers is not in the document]
rpuch: It's already not `\w+` as it might include colons (:). Changed it to `\S+`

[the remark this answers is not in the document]
ibessonov: Thank you! Completely forgot about colons

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
```diff
@@ -792,6 +816,116 @@ void testChangeLeaderDuringSnapshotInstallationToLeaderWithEnoughLog() throws Ex
...
+        // Prepare the scene: force node 0 to be a leader, and node 2 to be a follower.
+
+        final int leaderIndex = 0;
+        final int followerIndex = 2;
+
+        transferLeadershipOnSolePartitionTo(leaderIndex);
+        cluster.transferLeadershipTo(leaderIndex, MetastorageGroupId.INSTANCE);
+
+        // Block AppendEntries from being accepted on the follower so that the leader will have to use a snapshot.
+        blockIncomingAppendEntriesAt(followerIndex);
```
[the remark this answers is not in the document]
ibessonov: We could create a table on nodes `A` and `B`, and then increase the number of replicas, causing the rebalance process and creation of the table on node `C`. If we truncate the log before that, it would lead to the full state transfer. No need to block append entries requests, but it'll probably get harder to inhibit the meta-storage...
Anyway, that wasn't a suggestion, it was a conversation starter

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
ibessonov: Can we have some unit tests maybe? We're not even trying to write them for some reason

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
ibessonov: Ok, I completely misunderstood this construction. I apologize

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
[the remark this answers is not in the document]
rpuch: The problem is that if node C's watches are inhibited it will not see the assignments update, so the rebalance will not happen.

Also, using regular mechanisms in such a way looks as a hack itself. It seems that when we want to cripple the ability of a node to receive AppendEntries we should do exactly this, it makes the test a little bit understandable and less fragile.

ibessonov (approved):

ibessonov (approved):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/raftsnapshot/ItTableRaftSnapshotsTest.java
rpuch: Added a unit test

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/incoming/IncomingSnapshotCopier.java
```diff
@@ -332,7 +354,7 @@ private CompletableFuture<?> loadSnapshotMvData(ClusterNode snapshotSender, Exec
     /**
      * Requests and stores data into {@link TxStateStorage}.
      */
-    private CompletableFuture<?> loadSnapshotTxData(ClusterNode snapshotSender, Executor executor) {
+    private CompletableFuture<Void> loadSnapshotTxData(ClusterNode snapshotSender, Executor executor) {
```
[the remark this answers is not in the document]
rpuch: When a method is declared as returning a `CompletableFuture<?>`, you can actually return a future that completes with an int, or a String, or whatever, so it's less restricting type-wise. But when you declare it as returning a `CompletableFuture<Void>`, you can only return a future completed with a `null`, so it's stricter. I think it's better to be more strict: 'null-only' versus 'I don't care with  what this will be completed'.

PR #2475, opened 2023-08-22
IGNITE-19570 Write intent resolution for RW transactions

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-19570

### DIFF
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
@@ -32,6 +32,8 @@ import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.stream.IntStream;
 import org.apache.ignite.internal.app.IgniteImpl;
+import org.apache.ignite.internal.raft.Peer;
+import org.apache.ignite.internal.raft.service.RaftGroupService;
 import org.apache.ignite.internal.schema.BinaryRow;
 import org.apache.ignite.internal.schema.BinaryRowEx;
 import org.apache.ignite.internal.schema.marshaller.TupleMarshallerImpl;
@@ -183,6 +185,10 @@ public class ItRaftCommandLeftInLogUntilRestartTest extends ClusterPerClassInteg
 
             assertTrue(IgniteTestUtils.waitForCondition(() -> appliedIndexNode0.get() == appliedIndexNode1.get(), 10_000));
 
+            RaftGroupService raftGroupService = table.internalTable().partitionRaftGroupService(0);
+
+            raftGroupService.peers().forEach(peer -> raftGroupService.snapshot(peer).join());
+
             leaderAndGroupRef.set(new IgniteBiTuple<>(leader, table.tableId() + "_part_0"));
 
             afterBlock.accept(tx);
@@ -193,6 +199,9 @@ public class ItRaftCommandLeftInLogUntilRestartTest extends ClusterPerClassInteg
         }
 
         stopNodes();
+
+        log.info("Restart the cluster");
+
         startCluster();
 
         var node0Started = (IgniteImpl) CLUSTER_NODES.get(0);
@@ -294,6 +303,43 @@ public class ItRaftCommandLeftInLogUntilRestartTest extends ClusterPerClassInteg
                 new RuntimeException(IgniteStringFormatter.format("Cannot check a row {}", row), e);
             }
         }
+
+        transferLeadershipToLocalNode(ignite);
+
+        for (Object[] row : dataSet) {
+            try {
+                Tuple txTuple = table.keyValueView().get(null, Tuple.create().set("ID", row[0]));
+
+                assertNotNull(txTuple);
+
+                assertEquals(row[1], txTuple.value("NAME"));
+                assertEquals(row[2], txTuple.value("SALARY"));} catch (Exception e) {
+                new RuntimeException(IgniteStringFormatter.format("Cannot check a row {} when the local node leader", row), e);
+            }
+        }
+    }
+
+    /**
+     * Transfers the leader to the local node related to the Ignite instance.
+     *
+     * @param ignite Ignite instance.
+     */
+    private static void transferLeadershipToLocalNode(IgniteImpl ignite) {
+        TableImpl table = (TableImpl) ignite.tables().table(DEFAULT_TABLE_NAME);
+
+        RaftGroupService raftGroupService = table.internalTable().partitionRaftGroupService(0);
+
+        List<Peer> peers = raftGroupService.peers();
+        assertNotNull(peers);
+
+        Peer leader = raftGroupService.leader();
+        assertNotNull(leader);
+
+        Peer localPeer = peers.stream().filter(peer -> peer.consistentId().equals(ignite.name())).findFirst().get();
+
+        log.info("Leader is transferring [from={}, to={}]", leader, localPeer);
+
+        raftGroupService.transferLeadership(localPeer).join();
     }
 
     /**
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
@@ -313,7 +313,8 @@ public class ItRaftCommandLeftInLogUntilRestartTest extends ClusterPerClassInteg
                 assertNotNull(txTuple);
 
                 assertEquals(row[1], txTuple.value("NAME"));
-                assertEquals(row[2], txTuple.value("SALARY"));} catch (Exception e) {
+                assertEquals(row[2], txTuple.value("SALARY"));
+            } catch (Exception e) {
                 new RuntimeException(IgniteStringFormatter.format("Cannot check a row {} when the local node leader", row), e);
             }
         }
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
@@ -17,6 +17,8 @@
 
 package org.apache.ignite.internal.runner.app;
 
+import static org.apache.ignite.internal.testframework.matchers.CompletableFutureMatcher.willCompleteSuccessfully;
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
@@ -187,7 +189,7 @@ public class ItRaftCommandLeftInLogUntilRestartTest extends ClusterPerClassInteg
 
             RaftGroupService raftGroupService = table.internalTable().partitionRaftGroupService(0);
 
-            raftGroupService.peers().forEach(peer -> raftGroupService.snapshot(peer).join());
+            raftGroupService.peers().forEach(peer -> assertThat(raftGroupService.snapshot(peer), willCompleteSuccessfully()));
 
             leaderAndGroupRef.set(new IgniteBiTuple<>(leader, table.tableId() + "_part_0"));
 
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
@@ -342,7 +342,7 @@ public class ItRaftCommandLeftInLogUntilRestartTest extends ClusterPerClassInteg
 
         log.info("Leader is transferring [from={}, to={}]", leader, localPeer);
 
-        raftGroupService.transferLeadership(localPeer).join();
+        assertThat(raftGroupService.transferLeadership(localPeer), willCompleteSuccessfully());
     }
 
     /**
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
@@ -94,7 +94,10 @@ public class PartitionListener implements RaftGroupListener {
     /** Storage of transaction metadata. */
     private final TxStateStorage txStateStorage;
 
-    /** Rows that were inserted, updated or removed. All row IDs are sorted in natural order to prevent deadlocks upon commit/abort. */
+    /**
+     * Rows that were inserted, updated or removed. All row IDs are sorted in natural order to prevent deadlocks upon commit/abort.
+     * TODO: IGNITE-18502 Implement a pending update storage
+     */
     private final Map<UUID, SortedSet<RowId>> txsPendingRowIds = new HashMap<>();
 
     /** Safe time tracker. */
@@ -123,7 +126,6 @@ public class PartitionListener implements RaftGroupListener {
         this.safeTime = safeTime;
         this.storageIndexTracker = storageIndexTracker;
 
-        // TODO: IGNITE-18502 Implement a pending update storage
         try (PartitionTimestampCursor cursor = partitionDataStorage.scan(HybridTimestamp.MAX_VALUE)) {
             while (cursor.hasNext()) {
                 ReadResult readResult = cursor.next();
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
@@ -94,10 +94,7 @@ public class PartitionListener implements RaftGroupListener {
     /** Storage of transaction metadata. */
     private final TxStateStorage txStateStorage;
 
-    /**
-     * Rows that were inserted, updated or removed. All row IDs are sorted in natural order to prevent deadlocks upon commit/abort.
-     * TODO: IGNITE-18502 Implement a pending update storage
-     */
+    /** Rows that were inserted, updated or removed. All row IDs are sorted in natural order to prevent deadlocks upon commit/abort. */
     private final Map<UUID, SortedSet<RowId>> txsPendingRowIds = new HashMap<>();
 
     /** Safe time tracker. */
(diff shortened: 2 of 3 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -123,7 +126,6 @@ public PartitionListener(
         this.safeTime = safeTime;
         this.storageIndexTracker = storageIndexTracker;
 
-        // TODO: IGNITE-18502 Implement a pending update storage
```
sanpwc, line 126: Seems that given place is a proper one for this TODO.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -717,7 +717,14 @@ private CompletableFuture<List<BinaryRow>> continueScanBatchRetrieval(
             List<BinaryRow> batchRows
     ) {
         if (batchRows.size() < batchCount && cursor.hasNext()) {
-            return resolveAndCheckReadCompatibility(cursor.next(), txId)
+            ReadResult readResult = cursor.next();
+
+            HybridTimestamp newestCommitTimestamp = readResult.newestCommitTimestamp();
```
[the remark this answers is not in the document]
vldpyatkov: No, the logic is executed asynchronously. That means the cursor might move ahead when we try to find a candidate.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -970,28 +976,41 @@ private CompletableFuture<Void> continueIndexScan(
...
-                                                    indexLocker,
-                                                    indexCursor,
-                                                    batchSize,
-                                                    result,
-                                                    isUpperBoundAchieved
-                                            );
-                                        });
+                                ReadResult readResult = mvDataStorage.read(rowId, HybridTimestamp.MAX_VALUE);
+                                return resolveAndCheckReadCompatibility(readResult, txId, () -> {
+                                    if (readResult.newestCommitTimestamp() == null) {
```
sanpwc, line 985: Same as above. I'd rather hide given logic inside resolve... method if it's possible.

sanpwc, line 985: Why do we have different logic for candidate calculation. Within code above there's a check that given readResult is a writeIntent.
```
            BinaryRow candidate =
                    newestCommitTimestamp == null || !readResult.isWriteIntent() ? null : cursor.committed(newestCommitTimestamp);
```
However here we don't have such logic, why?

vldpyatkov: The logic is already written and used in other places. In cases like the above, we have to get the commit timestamp earlier than the resolution is finalized.

vldpyatkov: Because here we can be sure that is a write-intent. It happens after the resolution is finalized and the decision about getting committed value is applied.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2308,38 +2355,31 @@ private CompletableFuture<BinaryRow> resolveRoReadResult(
...
-            @Nullable Supplier<BinaryRow> lastCommitted
+            Supplier<BinaryRow> lastCommitted
     ) {
         if (readResult == null) {
             return completedFuture(null);
+        } else if (!readResult.isWriteIntent()) {
+            return completedFuture(readResult.binaryRow());
         } else {
+            // RW resolution.
             if (txId != null) {
```
sanpwc, line 2373: Do we assume that RO transaction has null txId?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1025,15 +1044,26 @@ private CompletableFuture<Void> continueIndexLookup(
         return lockManager.acquire(txId, new LockKey(tableId(), rowId), LockMode.S)
                 .thenComposeAsync(rowLock -> { // Table row S lock
                     ReadResult readResult = mvDataStorage.read(rowId, HybridTimestamp.MAX_VALUE);
-                    return resolveAndCheckReadCompatibility(readResult, txId)
-                            .thenCompose(resolvedReadResult -> {
-                                if (resolvedReadResult != null) {
-                                    result.add(resolvedReadResult);
-                                }
+                    return resolveAndCheckReadCompatibility(readResult, txId, () -> {
+                        if (readResult.newestCommitTimestamp() == null) {
```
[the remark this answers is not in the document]
vldpyatkov: We have a different resolution procedure for cursors and for storage values. If we hide this for storage, we will check it for cursors twice.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
vldpyatkov: Yes, it is an old comment

sanpwc: Is that actually true?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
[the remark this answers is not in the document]
sanpwc: I didn't get your point. Could you please elaborate it?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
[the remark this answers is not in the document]
sanpwc: Anyway it seems strange to have such duplications. If there are two different triggers, let's add two different resolve helpers one for cursor and one for storage, what do you think?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2382,18 +2422,18 @@ private CompletableFuture<Boolean> resolveTxState(
             UUID txId,
             HybridTimestamp timestamp
     ) {
-        requireNonNull(timestamp, "timestamp");
+        boolean readLatest = timestamp == null;
 
         return placementDriver.sendMetaRequest(commitGrpId, FACTORY.txStateReplicaRequest()
                         .groupId(commitGrpId)
-                        .readTimestampLong(timestamp.longValue())
+                        .readTimestampLong((readLatest ? HybridTimestamp.MIN_VALUE : timestamp).longValue())
```
sanpwc, line 2429: It seems incorrect. Why we propagate MIN_VALUE as a read timestamp?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
vldpyatkov: Of couse.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
vldpyatkov: I am sure it is correct, because I tested it. 
The timestamp is set MIN_VALUE , because it gives a guarantee that this request don't wait for safe time.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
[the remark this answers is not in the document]
vldpyatkov: It should be executed synchronously to get a committed row from the cursor. We cannot get the row after the resolution because the resolution is asynchronously completed.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
```diff
@@ -183,6 +185,10 @@ private void restartClusterWithNotAppliedCommands(
 
             assertTrue(IgniteTestUtils.waitForCondition(() -> appliedIndexNode0.get() == appliedIndexNode1.get(), 10_000));
 
+            RaftGroupService raftGroupService = table.internalTable().partitionRaftGroupService(0);
+
+            raftGroupService.peers().forEach(peer -> raftGroupService.snapshot(peer).join());
```
sanpwc, line 190: I though, that we've agreed, that we will use willCompleteSuccessfully or similar instead of join, haven't we?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
```diff
@@ -294,6 +303,44 @@ private void checkData(IgniteImpl ignite, Object[][] dataSet) {
                 new RuntimeException(IgniteStringFormatter.format("Cannot check a row {}", row), e);
             }
         }
+
+        transferLeadershipToLocalNode(ignite);
```
sanpwc, line 307: Why do you need to transfer the leadership? I mean that transactions use primary replicas and not the leaders.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
vldpyatkov: I am not sure it is needed. We already have the same logic for the resolution in other places. In the code where we check it for a cursor, we are doing this to avoid additional calculations on a commit candidate, but when the closure is invoked, we might be sure the committed value is really needed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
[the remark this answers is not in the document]
sanpwc: It's still not clear.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: According to the TxManagerImpl.begin() it's not true
```
    public InternalTransaction begin(boolean readOnly, @Nullable HybridTimestamp observableTimestamp) {
        assert readOnly || observableTimestamp == null : "Observable timestamp is applicable just for read-only transactions.";

        HybridTimestamp beginTimestamp = clock.now();
        UUID txId = transactionIdGenerator.transactionIdFor(beginTimestamp);
        changeState(txId, null, PENDING);

        if (!readOnly) {
            return new ReadWriteTransactionImpl(this, txId);
        }
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
vldpyatkov: Because currently we collaborate primarily with a leader, in the future this place should be reviewed. That is not the only place where we move a leader, but think about the primary.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
vldpyatkov: It is a different place, and we never pass the transaction id in resolution for RO transactions.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: > I am sure it is correct, because I tested it.
What exactly did you test? Where are the tests?

I mean that RW transaction has both startTimestamp and corresponding operations timestamps. Why we use MIN_VALUE for RW instead of actual RW related timestamps?

> because it gives a guarantee that this request don't wait for safe time.
Could you please share a code, where we wait for a safeTime for the writeIntent resolution?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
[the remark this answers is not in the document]
vldpyatkov: It was done early. I am not sure that it will be easier to read when I add two additional resolution methods. I can do it if you insist.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2382,18 +2422,18 @@ private CompletableFuture<Boolean> resolveTxState(
...
                         .txId(txId)
                         .build())
                 .thenApply(txMeta -> {
                     if (txMeta == null) {
                         return true;
                     } else if (txMeta.txState() == TxState.COMMITED) {
-                        return txMeta.commitTimestamp().compareTo(timestamp) > 0;
+                        return !readLatest && txMeta.commitTimestamp().compareTo(timestamp) > 0;
```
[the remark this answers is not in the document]
vldpyatkov: It makes this code more clear.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: So, you believe that multiple code duplications are better?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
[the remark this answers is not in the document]
sanpwc: I believe it's worth doing it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: It's very unreliable. Do we have any other way to distinguish RW from RO here?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
sanpwc: Why do you need to transfer the leadership? Could you please share test scenario?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
vldpyatkov: Maybe nowhere, but we also use this timestamp to update a clock on the other node. The minimum value protects us from a useless update.
Moreover, I do not see the need to propagate a specific timestamp for RW.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
vldpyatkov: The transfer leader is needed to ensure that we are reading values from a specific node and that this works correctly (resolving write intents).

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
vldpyatkov: Here we can compare the read timestamp with null. The null is only possible for RW transactions.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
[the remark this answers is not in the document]
sanpwc: Discussed in person. Let' it be !readLatest.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
sanpwc: Discussed in person. Let'it be transferLeader. WIll be covered in separate ticket with PD.transferPrimaryReplica

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: Discussed in person. Let' it be MIN_VALUE.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
vldpyatkov: Done.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
```diff
@@ -294,6 +305,44 @@ private void checkData(IgniteImpl ignite, Object[][] dataSet) {
...
+        assertNotNull(peers);
+
+        Peer leader = raftGroupService.leader();
+        assertNotNull(leader);
+
+        Peer localPeer = peers.stream().filter(peer -> peer.consistentId().equals(ignite.name())).findFirst().get();
+
+        log.info("Leader is transferring [from={}, to={}]", leader, localPeer);
+
+        raftGroupService.transferLeadership(localPeer).join();
```
sanpwc, line 345: Here as well - willCompleteSuccessfully()

sanpwc (approved):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItRaftCommandLeftInLogUntilRestartTest.java
vldpyatkov: Fixed.

PR #2490, opened 2023-08-23
IGNITE-20272 Clean up of DistributionZoneManagerWatchListenerTest

sergeyuttsel:
https://issues.apache.org/jira/browse/IGNITE-20272


modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/DistributionZoneManagerWatchListenerTest.java
```diff
@@ -18,61 +18,25 @@
...
-
-        Set<LogicalNode> nodes = Set.of(NODE_1, NODE_2);
-
-        setLogicalTopologyInMetaStorage(nodes, 100, metaStorageManager);
-
-        // TODO: IGNITE-18564 This is incorrect to check that data nodes are the same right after logical topology is changes manually.
-        assertDataNodesForZone(defaultZoneId, Set.of(NODE_1), keyValueStorage);
-    }
 
+    //TODO: IGNITE-19955 Check if this test is needed.
```
alievmirza, line 39: This test is kinda strange and tests some strange invariants, that will be changed after new restart mechanism will be implemented.  
For the sake of the saving a scenario, let's save this test, but move to the `DistributionZoneManagerLogicalTopologyEventsTest`, because this test is not related to ant watch listener.
As a result, `DistributionZoneManagerWatchListenerTest` could be removed

alievmirza (changes requested):

alievmirza (approved):

PR #2500, opened 2023-08-25
IGNITE-19499 TableManager should listen CatalogService events instead of configuration

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19499

### The Review Checklist

### Notes


modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -410,21 +394,18 @@ private void startIndexes() {
...
 
-            assert table != null : "tableId=" + index.tableId() + ", indexId=" + index.id();
+            CatalogTableDescriptor table = catalogManager.table(tableId, catalogVersion);
 
-            // TODO: IGNITE-19499 Only catalog should be used
-            int configTableId = getConfigTableId(table.name(), tableListView);
+            assert table != null : "tableId=" + tableId + ", indexId=" + index.id();
 
-            CompletableFuture<?> fireCreateIndexEventFuture = fireCreateIndexEvent(index, causalityToken, configTableId);
+            CompletableFuture<?> fireCreateIndexEventFuture = fireCreateIndexEvent(index, causalityToken, tableId);
```
AMashenkov, line 406: 1. Do we need IndexEvents? Seems, no one use them.
2. If yes, then will it be correctly to register index in table, then fire the event?

tkalkirill: Maybe we should get rid of index events altogether, in the sense of deleting them to hell?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -502,54 +465,44 @@ public TableManager(
...
-                return onTableDelete(ctx);
-            }
-        });
+                return onTableDelete(((DropTableEventParameters) parameters)).thenApply(unused -> false);
+            });
 
-        schemaManager.listen(SchemaEvent.CREATE, new EventListener<>() {
-            @Override
-            public CompletableFuture<Boolean> notify(SchemaEventParameters parameters, @Nullable Throwable exception) {
+            schemaManager.listen(SchemaEvent.CREATE, (parameters, exception) -> inBusyLockAsync(busyLock, () -> {
```
AMashenkov, line 498: Seems, no one listens for SchemaEvent and TableEvent.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1463,408 +1416,51 @@ private void dropTableLocally(long causalityToken, CatalogTableDescriptor tableD
...
-                        var tables = new ArrayList<Table>(tableIds.size());
-
-                        for (var fut : tableFuts) {
-                            var table = fut.join();
-
-                            if (table != null) {
-                                tables.add((Table) table);
-                            }
-                        }
+        return anyOf(schemaSyncService.waitForMetadataCompleteness(now), stopManagerFuture)
```
AMashenkov, line 1444: This looks like a hack.

I believe, SchemaSyncService should be responsible cancelling futures, which it creates.
And it does, because uses `ClusterTime`, which implementation cleanup resources on `close()`.

So, I guess the real issue is that we do `future.join()` under busylock.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1233,20 +1130,77 @@ public boolean removeAssignmentsChangeListener(Consumer<IgniteTablesInternal> li
...
+                    zoneDescriptor,
+                    assignmentsFuture,
+                    catalogVersion
+            ).whenComplete((v, e) -> {
+                if (e == null) {
+                    for (var listener : assignmentsChangeListeners) {
+                        listener.accept(this);
+                    }
+                }
+            }).thenCompose(ignored -> writeTableAssignmentsToMetastore(tableId, assignmentsFuture));
```
AMashenkov, line 1182: Will it be the same?
```suggestion
            ).thenCompose((v, e) -> {
                    for (var listener : assignmentsChangeListeners) {
                        listener.accept(this);
                    }
                return writeTableAssignmentsToMetastore(tableId, assignmentsFuture);
        });
```

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
```diff
@@ -245,37 +222,36 @@ public class TableManagerTest extends IgniteAbstractTest {
     /** Hybrid clock. */
     private final HybridClock clock = new HybridClockImpl();
 
+    /** Catalog vault. */
+    private VaultManager catalogVault;
```
AMashenkov, line 226: Catalog doesn't have own Value, and no Vault implementation is backed by Catalog.
```suggestion
    private VaultManager vaultManager;
```

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
```diff
@@ -245,37 +222,36 @@ public class TableManagerTest extends IgniteAbstractTest {
...
+
+    /** Catalog metastore. */
+    private MetaStorageManager catalogMetastore;
```
AMashenkov, line 229:
```suggestion
    private MetaStorageManager metastorageManager;
```

AMashenkov (changes requested):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1463,408 +1416,51 @@ private void dropTableLocally(long causalityToken, CatalogTableDescriptor tableD
...
-            } else {
-                return exTblCfg.id().value();
-            }
-        } catch (NoSuchElementException e) {
-            return null;
-        }
+                    return allOf(tableImplFutures)
+                            .thenApply(unused1 -> inBusyLock(
+                                    busyLock,
+                                    () -> Stream.of(tableImplFutures).map(CompletableFuture::join).collect(toList())
```
AMashenkov, line 1461: BusyLock is used to forbid accessing internals, when it is unsafe, e.g. on node stop.
Here, collecting tables doesn't need to be in busy lock.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
tkalkirill: In the current test, the catalog has its own valt and metastore, the one used in the test is mocked.

Then the question may arise, they say, why are they needed at all? 
We need a `Consumer<LongFunction<CompletableFuture<?>>> registry` (for vv entities inside **TableManager**) that used to be from the configuration, but now it needs to be provided somehow. If there is a more elegant way to do this, then welcome.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
tkalkirill: Same

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: @sanpwc, wdyt, should we get rid of the **busyLock** in `TableManager#join`?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: I'll get rid of them well.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
AMashenkov: BTW, we do nothing in case or error.
Also, @sanpwc, does the order looks correct? or maybe we should write to metastorage prior to notification?

I'm ok to make a separate fix if needed. I'm just curious and this place cause questions.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
tkalkirill: I looked at the opportunity to get rid of this, I realized that I need to make quite a few changes for this, I propose to do it in IGNITE-20339.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: I looked at the opportunity to get rid of this, I realized that I need to make quite a few changes for this, I propose to do it in IGNITE-20339.

modules/index/src/test/java/org/apache/ignite/internal/index/IndexManagerTest.java
```diff
@@ -105,12 +89,12 @@ public class IndexManagerTest extends BaseIgniteAbstractTest {
     public void setUp() {
         TableManager tableManagerMock = mock(TableManager.class);
 
-        when(tableManagerMock.tableAsync(anyLong(), anyString())).thenAnswer(inv -> {
+        when(tableManagerMock.tableAsync(anyLong(), anyInt())).thenAnswer(inv -> {
```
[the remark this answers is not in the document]
tkalkirill: fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: fix it

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
tkalkirill: Since there is no written answer (only emoji), I assume that Andrey agrees with my comment.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2013,94 +1589,67 @@ public CompletableFuture<TableImpl> tableImplAsync(String name) {
...
-    /**
-     * Checks that the table is configured with specific id.
-     *
-     * @param id Table id.
-     * @return True when the table is configured into cluster, false otherwise.
-     */
-    private boolean isTableConfigured(int id) {
-        return configuredTablesCache.isTableConfigured(id);
+        return anyOf(getLatestTableFuture, stopManagerFuture)
+                .thenComposeAsync(o -> getLatestTableFuture, ioExecutor)
```
sashapolo, line 1651: what's the point of using `thenComposeAsync` instead of `thenCompose` here? Also, what's the point of `anyOf` above, which uses the same future, if we still need to wait for it?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -240,8 +237,11 @@ public class PartitionReplicaListener implements ReplicaListener {
 
     private final CatalogService catalogService;
 
-    /** Listener for configuration indexes, {@code null} if the replica is not the leader. */
-    private final AtomicReference<ConfigurationNamedListListener<TableIndexView>> indexesConfigurationListener = new AtomicReference<>();
+    /** Listener for creating an index in catalog, {@code null} if the replica is not the leader. */
+    private final AtomicReference<EventListener<CreateIndexEventParameters>> createIndexListener = new AtomicReference<>();
```
sashapolo, line 241: Do we need to use `AtomicReferenceFieldUpdater` here? `PartitionReplicaListener` is created for every partition

modules/table/src/testFixtures/java/org/apache/ignite/internal/table/TableTestUtils.java
```diff
@@ -61,6 +62,46 @@ public static void createTable(
...
+    }
+
+    /**
+     * Returns table descriptor form catalog, {@code null} if table is absent.
+     *
+     * @param catalogService Catalog service.
+     * @param tableName Table name.
+     * @param timestamp Timestamp.
+     */
+    public static CatalogTableDescriptor getTableStrict(CatalogService catalogService, String tableName, long timestamp) {
```
sashapolo, line 97: Usually such methods are called something like `getRequiredTable`

modules/table/src/testFixtures/java/org/apache/ignite/internal/table/TableTestUtils.java
```diff
@@ -61,6 +62,46 @@ public static void createTable(
...
+     * @param catalogService Catalog service.
+     * @param tableName Table name.
+     * @param timestamp Timestamp.
+     */
+    public static @Nullable CatalogTableDescriptor getTable(CatalogService catalogService, String tableName, long timestamp) {
+        return catalogService.table(tableName, timestamp);
+    }
+
+    /**
+     * Returns table descriptor form catalog, {@code null} if table is absent.
```
sashapolo, line 91: This is not true

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
AMashenkov: I think you can pass a CatalogEvent directly to TableManager listeners.

And I don't understand the purpose of `mockManagersAndCreateTableWithDelay` test.
If the test check interactions between CatalogManager and TableManager, then it should be proper integration test.
If it just a unit, that checks guarantee the table will be properly initialized, then passing an event is enough.

sanpwc: First of all, why do you guys need vault here? MS doesn't use it to store key projections any longer. Seems that it's only used for persisting applied revision which is also effectively deprecated.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/replication/PartitionReplicaListenerTest.java
```diff
@@ -198,7 +197,7 @@ public class PartitionReplicaListenerTest extends IgniteAbstractTest {
     private static final int FUTURE_SCHEMA_ROW_INDEXED_VALUE = 0;
 
     /** Table id. */
-    private final int tblId = 1;
+    private static final int tblId = 1;
```
sanpwc, line 200: UPPER case.

sanpwc (changes requested):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2740,47 +2255,47 @@ private PartitionUpdateHandlers createPartitionUpdateHandlers(
...
+        List<CompletableFuture<?>> startTableFutures = new ArrayList<>();
 
-        for (TableView tableView : tablesCfg.tables().value()) {
-            fireEventFutures.add(fireEvent(TableEvent.CREATE, new TableEventParameters(causalityToken, tableView.id())));
+        for (CatalogTableDescriptor tableDescriptor : catalogService.tables(catalogVersion)) {
+            startTableFutures.add(createTableLocally(causalityToken, catalogVersion, tableDescriptor));
         }
 
-        startVv.update(causalityToken, (unused, throwable) -> allOf(fireEventFutures.toArray(CompletableFuture[]::new)))
+        startVv.update(causalityToken, (unused, throwable) -> allOf(startTableFutures.toArray(CompletableFuture[]::new)))
```
denis-chudov, line 2293: I am not sure that I understand the point of `startVv` here.  Looks like versioned value is not really needed, completable future would be ok.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/storage/ItRebalanceDistributedTest.java
```diff
@@ -608,6 +600,12 @@ private class Node {
         /** The future have to be complete after the node start and all Meta storage watches are deployd. */
         private CompletableFuture<Void> deployWatchesFut;
 
+        /** Hybrid clock. */
+        private final HybridClock clock = new HybridClockImpl();
```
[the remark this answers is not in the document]
tkalkirill: As far as I can see these hours are no longer shared.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: We discussed personally with @sanpwc and decided that in `TableManager#join` we will remove the use of the **busyLock**, and leave `anyOf`.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: I see that listeners are used in `org.apache.ignite.client.handler.ClientInboundMessageHandler`.
I will create a separate ticket for this problem.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
tkalkirill: > I think you can pass a CatalogEvent directly to TableManager listeners.

I don’t understand how this can help, I need exactly:
`revisionUpdater = (LongFunction<CompletableFuture<?>> function) -> catalogMetastore.registerRevisionUpdateListener(function::apply);`

> First of all, why do you guys need vault here? MS doesn't use it to store key projections any longer. Seems that it's only used for persisting applied revision which is also effectively deprecated.

I just want the tests to work when changing to a directory, then we can redo that later.

modules/schema/src/test/java/org/apache/ignite/internal/schema/configuration/SchemaDistributedConfigurationModuleTest.java
```diff
@@ -45,24 +41,22 @@ void typeIsDistributed() {
...
-        assertThat(module.schemaExtensions(), hasItem(ExtendedTableConfigurationSchema.class));
+        assertThat(module.schemaExtensions(), empty());
     }
 
     @Test
     void providesNoPolymorphicSchemaExtensions() {
-        assertThat(module.polymorphicSchemaExtensions(), hasItem(ConstantValueDefaultConfigurationSchema.class));
-        assertThat(module.polymorphicSchemaExtensions(), hasItem(FunctionCallDefaultConfigurationSchema.class));
-        assertThat(module.polymorphicSchemaExtensions(), hasItem(NullValueDefaultConfigurationSchema.class));
+        assertThat(module.polymorphicSchemaExtensions(), empty());
```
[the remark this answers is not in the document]
tkalkirill: Delete it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: > what's the point of using thenComposeAsync instead of thenCompose here?

I thought that it would be better when one of the futures is completed (from `anyOf`), the subsequent processing of the futures continued to be performed in the inner pool, so as not to block the threads that complete the futures, but maybe I'm really bothering, wdyt?

> Also, what's the point of anyOf above, which uses the same future, if we still need to wait for it?

I will try to explain: if we `busyLock#enterBusy` somewhere and do not release it for some time, then in order to avoid blocking on the `TableManager#stop`, we will simply immediately complete the future with an error.
We discussed the decision with @sanpwc and decided to keep it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
tkalkirill: `AtomicReferenceFieldUpdater` is an optimization, and at this stage of development we do not save on memory.
If we do optimization, then it is worth doing it under a separate ticket.

modules/table/src/testFixtures/java/org/apache/ignite/internal/table/TableTestUtils.java
tkalkirill: Thanks for the info, I'll leave it as it is =)

modules/table/src/testFixtures/java/org/apache/ignite/internal/table/TableTestUtils.java
tkalkirill: fix it

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogManagerImpl.java
```diff
@@ -350,7 +350,7 @@ public CompletableFuture<Void> dropTable(DropTableParams params) {
 
             Arrays.stream(schema.indexes())
                     .filter(index -> index.tableId() == table.id())
-                    .forEach(index -> updateEntries.add(new DropIndexEntry(index.id(), index.tableId(), table.name())));
+                    .forEach(index -> updateEntries.add(new DropIndexEntry(index.id(), index.tableId())));
```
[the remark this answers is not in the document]
tkalkirill: Yes, the busyLock is not used anywhere in this class, created [IGNITE-20381](https://issues.apache.org/jira/browse/IGNITE-20381)

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/AbstractSchemaChangeTest.java
```diff
@@ -129,50 +121,6 @@ void afterEach() throws Exception {
         IgniteUtils.closeAll(closeables);
     }
 
-    /**
-     * Check unsupported column type change.
-     */
-    @Test
-    public void testChangeColumnType() throws Exception {
```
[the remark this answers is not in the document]
tkalkirill: There are tests for validating catalog commands that have already arrived in separate classes [/ignite-3/pull/2547](https://github.com/apache/ignite-3/pull/2547).

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
```diff
@@ -1229,30 +1222,9 @@ private void createTableWithData(List<IgniteImpl> nodes, String name, int replic
...
-                .filter(Objects::nonNull)
-                .map(node -> node.clusterConfiguration().getConfiguration(TablesConfiguration.KEY));
-
-        List<TablesConfiguration> tablesConfigurations = Stream.concat(nodesTablesConfigurations, partialTablesConfiguration)
-                .collect(Collectors.toList());
-
         assertTrue(waitForCondition(
-                () -> tablesConfigurations.stream()
-                        .map(cfg -> cfg.indexes().get(indexName.toUpperCase()))
+                () -> nodes.stream()
```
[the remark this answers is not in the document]
tkalkirill: Since this is not a very useful code, there is already a TODO to fix it above.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItTablesApiTest.java
```diff
@@ -486,14 +431,17 @@ private void addColumnInternal(Ignite node, String tableName, ColumnDefinition c
...
-                .withDefaultValue("default").build();
-
+    private static void addColumnIfNotExists(Ignite node, String tableName) {
         try {
-            addColumnInternal(node, tableName, col);
-        } catch (ColumnAlreadyExistsException ex) {
-            log.info("Column already exists [naem={}]", col.name());
+            addColumn(node, tableName);
+        } catch (ColumnAlreadyExistsException e) {
+            log.info("Column already exists", e);
```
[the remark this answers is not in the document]
tkalkirill: I noticed that this check is useless, since in SQL we do not have support for the added column `IF NOT EXISTS`, I deleted it.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/schemasync/ItSchemaSyncAndReplicationTest.java
```diff
@@ -166,7 +167,8 @@ private static boolean solePartitionIsEmpty(IgniteImpl node) {
     }
 
     private static MvPartitionStorage solePartitionStorage(IgniteImpl node) {
-        TableImpl table = (TableImpl) node.tables().table(TABLE_NAME);
+        // We use this api because there is no waiting for schemas to sync.
```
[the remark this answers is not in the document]
tkalkirill: There is no workaround, we need to use the API to get the table without schema sync.
I just added a comment why I use this api.

modules/schema/src/main/java/org/apache/ignite/internal/schema/configuration/TableValidatorImpl.java
```diff
@@ -1,147 +0,0 @@
...
- * Table schema configuration validator implementation.
- */
-public class TableValidatorImpl implements Validator<TableValidator, NamedListView<TableView>> {
-    /** Static instance. */
-    public static final TableValidatorImpl INSTANCE = new TableValidatorImpl();
-
-    /** {@inheritDoc} */
-    @Override
-    public void validate(TableValidator annotation, ValidationContext<NamedListView<TableView>> ctx) {
-        TablesView tablesConfig = ctx.getNewRoot(TablesConfiguration.KEY);
```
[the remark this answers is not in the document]
tkalkirill: There are tests for validating catalog commands that have already arrived in separate classes in https://github.com/apache/ignite-3/pull/2547.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ClusterPerClassIntegrationTest.java
```diff
@@ -603,25 +585,26 @@ protected IgniteSql igniteSql() {
...
      * @param node Node.
      * @param tableName Table name.
-     * @return Table configuration.
      */
-    private static @Nullable TableConfiguration getTableConfiguration(Ignite node, String tableName) {
-        return getTablesConfiguration(node).tables().get(tableName.toUpperCase());
+    private static @Nullable CatalogTableDescriptor getTableDescriptor(Ignite node, String tableName) {
+        IgniteImpl nodeImpl = (IgniteImpl) node;
+
+        return getTable(nodeImpl.catalogManager(), tableName, nodeImpl.clock().nowLong());
```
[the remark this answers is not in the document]
tkalkirill: I think that for tests it will not be dangerous to use NOW, since most likely we will wait for the completion of the operation to create (or another operation) an index/table/zone for one node.
If something goes wrong, we can fix it point by point.

modules/table/build.gradle
```diff
@@ -66,6 +66,7 @@ dependencies {
     testImplementation(testFixtures(project(':ignite-metastorage')))
     testImplementation(testFixtures(project(':ignite-distribution-zones')))
     testImplementation(testFixtures(project(':ignite-catalog')))
+    testImplementation(testFixtures(project(':ignite-vault')))
```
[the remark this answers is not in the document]
tkalkirill: I use it in `org.apache.ignite.internal.table.distributed.TableManagerTest`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -219,17 +200,6 @@ public class TableManager extends Producer<TableEvent, TableEventParameters> imp
     private static final int TX_STATE_STORAGE_FLUSH_DELAY = 1000;
     private static final IntSupplier TX_STATE_STORAGE_FLUSH_DELAY_SUPPLIER = () -> TX_STATE_STORAGE_FLUSH_DELAY;
 
-    /**
-     * If this property is set to {@code true} then an attempt to get the configuration property directly from Meta storage will be skipped,
-     * and the local property will be returned.
-     * TODO: IGNITE-16774 This property and overall approach, access configuration directly through Meta storage,
-     * TODO: will be removed after fix of the issue.
-     */
-    private final boolean getMetadataLocallyOnly = IgniteSystemProperties.getBoolean("IGNITE_GET_METADATA_LOCALLY_ONLY");
```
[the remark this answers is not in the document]
tkalkirill: Removed use this sys_property in tests.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1233,20 +1130,77 @@ public boolean removeAssignmentsChangeListener(Consumer<IgniteTablesInternal> li
...
+     * @param causalityToken Causality token.
+     * @param catalogVersion Catalog version on which the table was created.
+     * @param tableDescriptor Catalog table descriptor.
+     * @return Future that will be completed when local changes related to the table creation are applied.
+     */
+    private CompletableFuture<?> createTableLocally(long causalityToken, int catalogVersion, CatalogTableDescriptor tableDescriptor) {
+        int tableId = tableDescriptor.id();
+
+        if (!busyLock.enterBusy()) {
+            fireEvent(TableEvent.CREATE, new TableEventParameters(causalityToken, tableId), new NodeStoppingException());
```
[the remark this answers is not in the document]
tkalkirill: Now no one is listening to this event, we will get rid of it in [IGNITE-20339](https://issues.apache.org/jira/browse/IGNITE-20339)

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/replication/PartitionReplicaListenerTest.java
tkalkirill: fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2740,47 +2255,47 @@ private PartitionUpdateHandlers createPartitionUpdateHandlers(
...
 
+        int catalogVersion = catalogService.latestCatalogVersion();
         long causalityToken = recoveryFinishFuture.join();
 
-        List<CompletableFuture<?>> fireEventFutures = new ArrayList<>();
+        List<CompletableFuture<?>> startTableFutures = new ArrayList<>();
 
-        for (TableView tableView : tablesCfg.tables().value()) {
-            fireEventFutures.add(fireEvent(TableEvent.CREATE, new TableEventParameters(causalityToken, tableView.id())));
+        for (CatalogTableDescriptor tableDescriptor : catalogService.tables(catalogVersion)) {
```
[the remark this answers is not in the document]
tkalkirill: Adds todo

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: We discussed it in person and removed use busyLock at `TableManager#join`.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1233,20 +1130,77 @@ public boolean removeAssignmentsChangeListener(Consumer<IgniteTablesInternal> li
...
+        try {
+            int zoneId = tableDescriptor.zoneId();
+
+            CatalogZoneDescriptor zoneDescriptor = getZoneDescriptor(tableDescriptor, catalogVersion);
+
+            CompletableFuture<List<Set<Assignment>>> assignmentsFuture;
+
+            // Check if the table already has assignments in the vault.
+            // So, it means, that it is a recovery process and we should use the vault assignments instead of calculation for the new ones.
+            if (partitionAssignments(vaultManager, tableId, 0) != null) {
```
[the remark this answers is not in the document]
tkalkirill: Add todo

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: In this case, it is important for us to avoid races when performing these future and catalog updates.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sashapolo: > I thought that it would be better when one of the futures is completed (from anyOf), the subsequent processing of the futures continued to be performed in the inner pool

This is ok, but I think it's a bit redundant, because looks like there's not a lot of things to execute, we wait for a future and then remove a listener

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sashapolo: Sure, I just thought that this optimization looks useful here and does not take a lot of code to implement.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
sanpwc: Fine to me

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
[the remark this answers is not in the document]
sanpwc: Before your changes within non-sql API we've checked tables explicitly though MS, meaning that we had metadata data sync, it's no longer true, right?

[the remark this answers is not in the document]
tkalkirill: We discussed it in person and the issue was dropped.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
denis-chudov: I am ok with comment.

sashapolo (approved):
I'm approving because my comments are resolved. This should not be treated as final approval

denis-chudov (approved):

sanpwc (approved):

PR #2508, opened 2023-08-28
IGNITE-20285 Move handler of CreateTable command from CatalogManager to command itself

korlov42:

### The Review Checklist

### Notes


modules/catalog/src/test/java/org/apache/ignite/internal/catalog/commands/CreateTableCommandValidationTest.java
```diff
@@ -0,0 +1,421 @@
...
+            mock(ClockWaiter.class)
+    );
+
+    @Test
+    void schemaNameMustBeNullOrBlank() {
+        CreateTableCommandBuilder builder = manager.createTableCommandBuilder();
+
+        builder = fillProperties(builder);
+
+        for (String name : new  String[] {null, "", " ", "  "}) {
```
lowka, line 68: Maybe test cases can be parameterised with `{null, "", " ", " "}`  and this collection of string can be used in other test cases (that check that table name, column name can not be null) as well?

modules/catalog/src/test/java/org/apache/ignite/internal/catalog/commands/CreateTableCommandValidationTest.java
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    @Test
+    @SuppressWarnings("unchecked")
+    void tableShouldHaveAtLeastOneColumn() {
+        CreateTableCommandBuilder builder = manager.createTableCommandBuilder();
+
+        builder = fillProperties(builder);
+
+        for (List<ColumnParams> columns : new List[] {null, List.of()}) {
```
lowka, line 103: Maybe this test case can also be parameterised with `null, List.of()` and `tableShouldHaveAtLeastOnePrimaryKeyColumn` as well?

korlov42: fixed

modules/catalog/src/test/java/org/apache/ignite/internal/catalog/commands/CreateTableCommandValidationTest.java
korlov42: fixed

lowka (approved):

modules/catalog/src/testFixtures/java/org/apache/ignite/internal/catalog/BaseCatalogManagerTest.java
```diff
@@ -159,24 +158,19 @@ protected static CreateSortedIndexParams createSortedIndexParams(
         return createSortedIndexParams(indexName, false, indexColumns, columnsCollations);
     }
 
-    protected static CreateTableParams createTableParams(
+    protected static CatalogCommand createTableCommand(
+            CreateTableCommandBuilder builder,
```
xtern, line 162: Personally, I don't see the point in adding a builder as a parameter and writing 40+ times `manager.createTableCommandBuilder()`  just to keep this method static, but it's up to you.

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogValidationException.java
```diff
@@ -26,6 +27,15 @@
 public class CatalogValidationException extends IgniteInternalException {
     private static final long serialVersionUID = 2692301541251354006L;
 
+    /**
+     * Constructor.
+     *
+     * @param message Error message.
+     */
+    public CatalogValidationException(String message) {
+        super(Common.INTERNAL_ERR, message);
```
xtern, line 36: It seems worth adding a comment why we don't care about the error code.

p.s. I didn't find where these exceptions are now mapped to STMT_VALIDATION_ERR code.
Maybe I didn't look well enough or it will be added in [1].

[1] https://issues.apache.org/jira/browse/IGNITE-20149

xtern (approved):

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogManager.java
```diff
@@ -24,25 +24,35 @@
...
+     * @throws IllegalArgumentException If given command was created not by this manager.
+     * @see #createTableCommandBuilder()
+     */
+    CompletableFuture<Void> execute(CatalogCommand command) throws IllegalArgumentException;
+
+    /**
+     * Returns builder to create a command to create a new table.
      */
-    CompletableFuture<Void> createTable(CreateTableParams params);
+    CreateTableCommandBuilder createTableCommandBuilder();
```
AMashenkov, line 55: We use CatalogManager for creating command builder.
Then we use CatalogManager again to apply the command, and expect the CatalogCommand is  an command instance, which is created with the builder.

Looks. like we expose some internals to the outside.
As we don't need a CatalogManager to create the builder, then maybe this method could be moved to a separate Factory class?

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/commands/CreateTableCommand.java
```diff
@@ -0,0 +1,187 @@
...
+     * @param schemaName Name of the schema to create table in. Should not be null or blank.
+     * @param primaryKeyColumns Name of columns which represent primary key.
+     *      Should be subset of columns in param `columns`.
+     * @param colocationColumns Name of the columns participating in distribution calculation.
+     *      Should be subset of the primary key columns.
+     * @param columns List of the columns containing by the table. There is should be at least one column.
+     * @param zoneName Name of the zone to create table in. Should not be null or blank.
+     * @throws CatalogValidationException if any of restrictions above is violated.
+     */
+    public CreateTableCommand(
```
xtern, line 78: We usually restrict the ability to create directly with the constructor those classes that the builder creates :roll_eyes: :thinking:

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogManager.java
korlov42: I agree that this looks rather strange. Besides, there is violation of Liskov Substitution Principle.

Considering only single command execution case, we actually may move method `execute` to the interface CatalogCommand, making a command self-contained. But I did not came up with a good solution to implement batched command. No matter what you do, you always need a guarantee that the command was created through the right builder.

One option is to introduce method `createBatch(List<Command> commands)` on a manager, another option is to introduce method `execute(List<CatalogCommand> commands)` on a manager. 

I decided to go with second option. With that said, having method `execute` on a manager for single command make api more consistent.

> maybe this method could be moved to a separate Factory class?

that's possible, but what will we do if we need access to some dynamic defaults from configuration, for instance? I think, static factory methods for such kind of objects not the best choice.

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/commands/builders/CreateTableCommandBuilderImpl.java
```diff
@@ -0,0 +1,105 @@
...
+        return this;
+    }
+
+    @SuppressWarnings("DataFlowIssue") // params are validated in constructor
+    @Override
+    public CatalogCommand build() {
+        String zoneName = requireNonNullElse(this.zoneName, CatalogService.DEFAULT_ZONE_NAME);
+
+        if (colocationColumns == null) {
+            colocationColumns = primaryKeyColumns;
```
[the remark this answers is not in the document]
korlov42: good catch! Fixed

modules/catalog/src/testFixtures/java/org/apache/ignite/internal/catalog/BaseCatalogManagerTest.java
korlov42: make sense, fixed

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogValidationException.java
korlov42: decided to introduce new code for this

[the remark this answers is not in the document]
korlov42: new code has been added

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/commands/CreateTableCommand.java
korlov42: there is validation in contractor, thus it should be safe to create object directly. But anyways, I've reworked this part a bit, now constructor is private, so there is only one option to create command now

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/commands/CreateTableCommand.java
```diff
@@ -0,0 +1,269 @@
...
+        @SuppressWarnings("DataFlowIssue") // params are validated in constructor
+        @Override
+        public CatalogCommand build() {
+            String zoneName = requireNonNullElse(this.zoneName, CatalogService.DEFAULT_ZONE_NAME);
+
+            List<String> colocationColumns = this.colocationColumns != null
+                    ? this.colocationColumns
+                    : primaryKeyColumns;
+
+            return new CreateTableCommand(
```
AMashenkov, line 259: Let's make a defensive copy for all the collections.

korlov42: good catch! Fixed

AMashenkov (approved):

PR #2512, opened 2023-08-29
IGNITE-19791 Introduce SqlRowHandler

xtern:
https://issues.apache.org/jira/browse/IGNITE-19791

Added `SqlRowHandler` that works with objects of `RowWrapper` type, which has two implementations - simple object array and binarytuple (see SqlRowHandler javadoc).

Known flaws that should be optimized, but it seems not in the current task.

- Currently `TableRowConverterImpl#toRow()` always rebuild binary tuple. It seems that this should be done only if the binary tuple schema does not match the row schema. Also the tuple reassembly (if needed) can be done more efficiently by copying required bytes and bypassing conversion to objects.

- We cannot read binary tuple directly in internal format. Therefore, when reading (using `handler#get()` method), a double conversion occurs. For example for DATE, first binary tuple reader converts bytes to localdate, then we convert localdate to integer.


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/row/SqlRowHandlerTest.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    private Object[] values(List<ColumnType> columnTypes) {
+        Object[] values = new Object[columnTypes.size()];
+        int baseValue = rnd.nextInt();
+
+        for (int i = 0; i < values.length; i++) {
+            ColumnType type = columnTypes.get(i);
+
+            values[i] = type == ColumnType.NULL || rnd.nextInt(4) == 1 ? null : generateValueByType(baseValue, type);
```
lowka, line 173: > rnd.nextInt(4) == 1
Why do we need this condition ? The chance of this being true is practically nil.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/row/SqlRowHandlerTest.java
```diff
@@ -0,0 +1,214 @@
...
+        RowSchema concatenatedSchema = builder.build();
+
+        // Serialize.
+        ByteBuffer buf = handler.toByteBuffer(concatenated);
+
+        // Wrap into row.
+        RowWrapper wrapped = handler.factory(concatenatedSchema).wrap(new BinaryTuple(totalElementsCount, buf));
+
+        for (int i = 0; i < Math.max(leftLen, rightLen); i++) {
+            if (i < leftLen) {
```
lowka, line 125: I think is code too complicate for what is supposed to do - concat(row1+row2) = row(cols::row1, cols::row2)
We can just concatenate leftData with rightData and check elements of concatenated row one by one.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/row/SqlRowHandlerTest.java
lowka: Every fourth type (assuming that distribution is gaussian) is expected to be nullable.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/row/RowSchemaTypes.java
```diff
@@ -82,4 +83,21 @@ public static BaseTypeSpec nativeTypeWithNullability(NativeType nativeType, bool
             return new BaseTypeSpec(nativeType, nullable);
         }
     }
+
+    /** Returns a {@link NativeType} extracted from the row schema type. */
```
lowka, line 87: I think it would be indicate in javadoc that this method returns null if is the given type spec represents null type and it throws exception if the given type spec is not `BaseTypeSpec`.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/row/RowSchema.java
```diff
@@ -39,6 +43,43 @@ public List<TypeSpec> fields() {
...
     }
 
+    /**
+     * Gets an object representation from a tuple's field.
+     *
+     * @param fieldIndex Field index to read.
+     * @param tuple Tuple to read the value from.
+     * @return An object representation of the value.
+     */
+    public @Nullable Object value(int fieldIndex, InternalTuple tuple) {
```
lowka, line 53: It would be better to swap arguments here. Almost every time field id follows the source data structure and not vice versa.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/TableRowConverterImpl.java
```diff
@@ -53,30 +55,37 @@ public <RowT> RowT toRow(
             RowHandler.RowFactory<RowT> factory,
             @Nullable BitSet requiredColumns
     ) {
-        RowHandler<RowT> handler = factory.handler();
+        Row row = schemaRegistry.resolve(binaryRow, schemaDescriptor);
+        BinaryTupleSchema binarySchema = BinaryTupleSchema.createRowSchema(schemaDescriptor);
```
lowka, line 59: I think we can create `BinaryTupleSchema` in constructor and reuse it.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
```diff
@@ -146,4 +169,55 @@ private static <RowT> ByteBuffer toByteBuffer(
 
         return tupleBuilder.build();
     }
+
+    private static BinaryTupleBuilder appendValue(BinaryTupleBuilder builder, TypeSpec schemaType, @Nullable Object value) {
```
lowka, line 173: I think return type is redundant here.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/SqlRowHandler.java
```diff
@@ -0,0 +1,260 @@
...
+        Object[] values = new Object[leftLen + rightLen];
+        Builder schemaBuilder = RowSchema.builder();
+
+        for (int i = 0; i < leftLen; i++) {
+            values[i] = left.get(i);
+            schemaBuilder.addField(leftTypes.get(i));
+        }
+
+        for (int i = 0; i < rightLen; i++) {
+            values[leftLen + i] = right.get(i);
```
zstan, line 85: no need to call additional offset here : leftLen + i
you just can initialize global int i = 0; and increment it in further for loops.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/SqlRowHandler.java
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    @Override
+    public String toString(RowWrapper row) {
+        return String.valueOf(row);
+    }
+
+    @Override
+    public ByteBuffer toByteBuffer(RowWrapper row) {
+        return row.toByteBuffer();
```
lowka, line 116: It would be better to provide implementation of `toString` method that is similar to ArrayRowHandler's. This one simply return `typeName@132` or something.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/ArrayRowHandler.java
```diff
@@ -50,6 +52,18 @@ public Object[] concat(Object[] left, Object[] right) {
         return ArrayUtils.concat(left, right);
     }
 
+    /** {@inheritDoc} */
+    @Override
+    public Object[] map(Object[] row, int[] mapping, int offset) {
```
zstan, line 57: can we gather new row without new allocations ? just using incoming row and cut all unused stuff at the end ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
```diff
@@ -277,8 +277,17 @@ public Supplier<RowT> rowSource(List<RexNode> values) {
     /** {@inheritDoc} */
     @Override
     public <T> Supplier<T> execute(RexNode node) {
-        RelDataType nodeType = node.getType();
-        RowSchema rowSchema = TypeUtils.rowSchemaFromRelTypes(List.of(nodeType));
+        RelDataType exprType = node.getType();
+        List<RelDataType> typesList;
+
+        if (exprType.getSqlTypeName() == SqlTypeName.ROW) {
```
zstan, line 283: we have no such a check previously, can you explain why do we need it for now ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/agg/Accumulators.java
```diff
@@ -88,11 +88,11 @@ private Supplier<Accumulator> accumulatorFunctionFactory(AggregateCall call) {
             case "SOME":
                 return minMaxFactory(false, call);
             case "SINGLE_VALUE":
-                return SingleVal.FACTORY;
-            case "LITERAL_AGG":
-                return LiteralVal.FACTORY;
+                return singleAnyValueFactory(true, call);
             case "ANY_VALUE":
-                return AnyVal.FACTORY;
+                return singleAnyValueFactory(false, call);
```
zstan, line 93: i think new division by singleAnyValueFactory(true|false become less understandable ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/ModifyNode.java
```diff
@@ -268,15 +268,7 @@ private void inlineUpdates(int offset, List<RowT> rows) {
...
-        for (RowT row : rows) {
-            for (int i = 0; i < mapping.length; i++) {
-                if (offset == 0 && i == mapping[i]) {
-                    continue;
-                }
-
-                handler.set(i, row, handler.get(mapping[i] + offset, row));
-            }
-        }
+        rows.replaceAll(row -> handler.map(row, mapping, offset));
```
zstan, line 271: previous implementation was a bit lighter, why do we need to change it ?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/row/SqlRowHandlerTest.java
```diff
@@ -0,0 +1,212 @@
...
+            ColumnType type = columnTypes.get(i);
+
+            values[i] = type == ColumnType.NULL ? null : generateValueByType(baseValue, type);
+        }
+
+        return values;
+    }
+
+    private List<ColumnType> columnTypes() {
+        List<ColumnType> columnTypes = new ArrayList<>(EnumSet.complementOf(EnumSet.of(ColumnType.PERIOD, ColumnType.DURATION)));
```
lowka, line 178: it would be better to add a link to an issue with INTERVAL type support: https://issues.apache.org/jira/browse/IGNITE-15200

xtern: Done, thanks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/SqlRowHandler.java
xtern: Forgot about it - thanks for pointing on this.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/row/RowSchema.java
xtern: Swapped, thanks.
p.s. I was guided by the order in which they go in RowHandler#get/set :roll_eyes:

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/row/RowSchemaTypes.java
xtern: Javadoc updated, thanks.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/row/SqlRowHandlerTest.java
xtern: Split into two cycles.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/row/SqlRowHandlerTest.java
xtern: Removed this confusing `random`.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/SqlRowHandler.java
xtern: We need two indices anyway.  I don't think it worth to replace `leftLen + i` to something like 'index++' and declare `index` out of cycle.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/agg/Accumulators.java
xtern: I divided it into two factories, if I understood your comment correctly

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
xtern: on the next line after this line there is a comment explaining why this condition is needed.

Previously, RowSchema was not used at all.
Currently it used to resolve object types in the row.
We faced with the problem that table function may return composite type.
As far as I understand, we do not and do not plan to support such types, and at the moment this composite type can only include one "inner" type. Therefore, we decided to convert such a composite type to single base type here.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
```diff
@@ -108,34 +121,44 @@ public static <RowT> BinaryTuple toBinaryTuple(
...
-        }
-
         BinaryTupleBuilder tupleBuilder = new BinaryTupleBuilder(rowColumnsCount);
 
-        return new BinaryTuple(rowColumnsCount, toByteBuffer(ectx, binarySchema, handler, tupleBuilder, searchRow));
+        return new BinaryTuple(rowColumnsCount, toByteBuffer(binarySchema, handler, tupleBuilder, rowColumnsCount, searchRow));
+    }
+
+    /**
+     * Converts an array of objects to a byte buffer representation of a {@link BinaryTuple binary tuple}. This method uses
```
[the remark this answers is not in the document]
xtern: Done, thanks.

lowka (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/SqlRowHandler.java
```diff
@@ -0,0 +1,272 @@
...
+
+                return new BinaryTupleRowWrapper(rowSchema, tuple);
+            }
+        };
+    }
+
+    /**
+     * Provides the ability for a single {@link RowHandler} instance to interact with multiple row implementations.
+     */
+    public abstract static class RowWrapper {
```
zstan, line 175: i see an interface, not abstract class here )

xtern: I would like to hide the methods of this class so that it is possible to work only through the RowHandler and not directly :thinking: :roll_eyes:

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/SqlRowHandler.java
```diff
@@ -0,0 +1,272 @@
...
+            }
+
+            NativeType nativeType = RowSchemaTypes.toNativeType(rowSchema.fields().get(field));
+
+            return TypeUtils.toInternal(value, Commons.nativeTypeToClass(nativeType));
+        }
+
+        @Override
+        void set(int field, @Nullable Object value) {
+            throw new UnsupportedOperationException();
```
korlov42, line 259: how to make sure no one will invoke `set` for row that is wrapper around BinaryTuple?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowHandler.java
```diff
@@ -44,6 +45,23 @@ public interface RowHandler<RowT> {
...
+     *    result mapping will be [1, 3]
+     *    result row will be [6, 8]
+     * </pre>
+     *
+     * @param row Source row.
+     * @param mapping Target field indexes.
+     * @param offset Target field indexes offset.
+     * @return A new row with fields from the specified mapping.
+     */
+    RowT map(RowT row, int[] mapping, int offset);
```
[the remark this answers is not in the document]
xtern: offset removed from the signature.

> why do we even need this method in the first place?
It seemed to me that depending on the implementation it can be implemented more efficiently.

> it does make sense to create new row explicitly
I didn't find how to get `rowschema` :pensive: to create a new row

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowHandler.java
```diff
@@ -87,5 +105,13 @@ interface RowFactory<RowT> {
          * @return Instantiation defined representation.
          */
         RowT create(ByteBuffer raw);
+
+        /**
+         * Wraps incoming binary tuple into a instantiation defined representation.
```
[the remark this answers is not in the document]
xtern: Reword a bit.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowHandler.java
```diff
@@ -87,5 +105,13 @@ interface RowFactory<RowT> {
...
+         *
+         * @param tuple {@link InternalTuple} representation.
+         * @return Instantiation defined representation.
+         */
+        RowT wrap(InternalTuple tuple);
```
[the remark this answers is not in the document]
xtern: Done

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/TableRowConverterImpl.java
```diff
@@ -53,30 +59,36 @@ public <RowT> RowT toRow(
             RowHandler.RowFactory<RowT> factory,
             @Nullable BitSet requiredColumns
     ) {
-        RowHandler<RowT> handler = factory.handler();
+        Row row = schemaRegistry.resolve(binaryRow, schemaDescriptor);
 
-        assert handler == ectx.rowHandler();
+        BinaryTupleBuilder builder = requiredColumns == null
```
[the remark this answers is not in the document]
xtern: Done, thanks

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
```diff
@@ -108,34 +121,44 @@ public static <RowT> BinaryTuple toBinaryTuple(
...
+
+    /**
+     * Converts an array of objects into a {@link ByteBuffer} representation of a {@link BinaryTuple binary tuple}. This method uses
+     * {@link RowSchema} to resolve object types.
+     *
+     * @param schema Row schema.
+     * @param values Array of objects to be stored in the buffer.
+     * @return Buffer in binary tuple format.
+     */
+    static ByteBuffer toByteBuffer(RowSchema schema, Object[] values) {
```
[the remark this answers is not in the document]
xtern: > why do you put this method here 
It seemed to me that this is a more suitable place, because. this class essentially does the same RowT -> BinaryTuple conversion.

anyway, moved to SqlRowHandler.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/row/RowSchema.java
```diff
@@ -39,6 +43,43 @@ public List<TypeSpec> fields() {
...
     }
 
+    /**
+     * Gets an object representation from a tuple's field.
+     *
+     * @param tuple Tuple to read the value from.
+     * @param fieldIndex Field index to read.
+     * @return An object representation of the value.
+     */
+    public @Nullable Object value(InternalTuple tuple, int fieldIndex) {
```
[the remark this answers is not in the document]
xtern: Moved to BinaryTupleRowWrapper :thinking:

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/SqlRowHandler.java
xtern: I agree that this method should be implemented or should not be provided at all.
From my point of view, the only right way is to replace this method with append(Object obj) and rework all related code.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/ModifyNode.java
```diff
@@ -268,15 +268,22 @@ private void inlineUpdates(int offset, List<RowT> rows) {
 
         RowHandler<RowT> handler = context().rowHandler();
 
-        for (RowT row : rows) {
-            for (int i = 0; i < mapping.length; i++) {
-                if (offset == 0 && i == mapping[i]) {
-                    continue;
-                }
+        rows.replaceAll(row -> handler.map(row, applyOffset(mapping, offset)));
```
korlov42, line 271: it's better to do computation outside of the loop

korlov42 (approved):

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/ArrayRowHandler.java
xtern: Yes we can, but this will go against the description of the method in the javadoc. This implementation will only be used in tests. I'm not sure how much can we gain from such optimization. But if you insist, I can change this method to return the original array with the values replaced.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/rel/ModifyNode.java
xtern: Done, thanks.

zstan (approved):

PR #2513, opened 2023-08-29
IGNITE-20300 Metastorage command reordering wrt Safe Time on Raft Group entry

rpuch:
https://issues.apache.org/jira/browse/IGNITE-20300

### The Review Checklist

### Notes


modules/raft-api/src/main/java/org/apache/ignite/internal/raft/service/RaftGroupListener.java
```diff
@@ -93,4 +93,18 @@ default void onConfigurationCommitted(CommittedConfiguration config) {
...
+     * Returns {@code true} if {@link #onBeforeApply(Command)} and saving of the command to the LogManager
+     * must be executed atomically (that is, these actions cannot intermingle for different commands
+     * on the same Raft group.
+     *
+     * <p>This does not influence read commands: for them no atomicity guarantees are provided.
+     *
+     * @param command The command.
+     * @return {@code true} if {@link #onBeforeApply(Command)} and saving must be executed atomically.
+     */
+    default boolean atomicOnBeforeApplyAndWrite(Command command) {
```
sashapolo, line 107: I would suggest to introduce a sub-class, that will declare the `onBeforeApply` method. This way we can always know if `onBeforeApply` is not a no-op and always execute it atomically with `write`

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/ActionRequestProcessor.java
```diff
@@ -71,13 +75,23 @@ public void handleRequest(RpcContext rpcCtx, ActionRequest request) {
 
         JraftServerImpl.DelegatingStateMachine fsm = (JraftServerImpl.DelegatingStateMachine) node.getOptions().getFsm();
 
-        // Apply a filter before committing to STM.
-        fsm.getListener().onBeforeApply(request.command());
+        if (fsm.getListener().atomicOnBeforeApplyAndWrite(request.command()) && request.command() instanceof WriteCommand) {
+            // Synchronizing per group.
+            synchronized (internizer.intern(request.groupId())) {
```
sashapolo, line 80: I really think that this "internizer" class is an overkill. Can we synchronize on the "node" object instead?

modules/raft-api/src/main/java/org/apache/ignite/internal/raft/service/RaftGroupListener.java
rpuch: But do we need to ALWAYS execute it atomically with the write? I think no

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/ActionRequestProcessor.java
rpuch: Yes, we could, but someone else could do the same in another part of the code which might lead to less parallelizm (unnecessarily). Another solution is to add a monitor object to the `NodeImpl` (one per instance), but we would have to add an accessor for it to the `Node` interface, and this also seems non-desirable.

modules/raft-api/src/main/java/org/apache/ignite/internal/raft/service/RaftGroupListener.java
sashapolo: Why not? I can't imagine a scenario when you need to implement `onBeforeApply` and not have it executed atomically. Having a separate boolean method seems to cryptic to me. Anyway, we can add it after we find out that we need it.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/ActionRequestProcessor.java
sashapolo: > but someone else could do the same in another part of the code which might lead to less parallelizm

I don't think it's an issue

Another solution would be to add an explicit string -> object map to hold per-group locks (which is what this internizer does)

modules/raft-api/src/main/java/org/apache/ignite/internal/raft/service/RaftGroupListener.java
sashapolo: ok, I think my proposed solution doesn't make sense, because we don't have separate listener instances for writes and reads (and we only need to call onBeforeApply for writes)

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/ActionRequestProcessor.java
rpuch: I moved the map from internizer to the processor and removed internizer implementation

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/ActionRequestProcessor.java
```diff
@@ -71,15 +81,32 @@ public void handleRequest(RpcContext rpcCtx, ActionRequest request) {
...
         }
     }
+    private static void callOnBeforeApply(ActionRequest request, DelegatingStateMachine fsm) {
+        ((BeforeApplyHandler) fsm.getListener()).onBeforeApply(request.command());
+    }
+
+    private Object groupIdSyncMonitor(String groupId) {
+        assert groupId != null;
+
+        return groupIdsToMonitors.computeIfAbsent(groupId, k -> groupId);
```
sashapolo, line 108: I would have expected a `new Object` call in case the key is not present, but I guess that's not a big deal. Also, we may use `Function.identity` here

sashapolo (approved):

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/ActionRequestProcessor.java
rpuch: `Function.identify()` would look a bit too cryptic in this particular case.

Also, why would we allocate a new object if we can use the key as a monitor? It's referenced by the map anyway.

sashapolo: That's just a matter of taste, that's why I have approved the PR regardless

alievmirza (approved):

PR #2518, opened 2023-08-30
IGNITE-19768 Use schema sync instead of explicit config check

AMashenkov:

### The Review Checklist

### Notes


AMashenkov:
@rpuch, would you please make a review.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1832,51 +1817,19 @@ public CompletableFuture<List<Table>> tablesAsync() {
...
-     *
-     * @param tblName Name of the table.
-     * @return Direct id of the table, or {@code null} if the table with the {@code tblName} has not been found.
-     */
-    @Nullable
-    private Integer directTableId(String tblName) {
+    private @Nullable Integer tableNameToId(String tableName) {
         try {
-            TableConfiguration exTblCfg = directProxy(tablesCfg.tables()).get(tblName);
+            TableConfiguration exTblCfg = tablesCfg.tables().get(tableName);
```
rpuch, line 1832: It's not an extended table config but usual one, so let's drop the `ex` prefix from the variable name

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2012,43 +1965,67 @@ public CompletableFuture<TableImpl> tableImplAsync(String name) {
...
-        CompletableFuture<Boolean> tblCfgFut = checkConfiguration
-                ? supplyAsync(() -> inBusyLock(busyLock, () -> isTableConfigured(id)), ioExecutor)
-                : completedFuture(true);
+    private CompletableFuture<TableImpl> tableReadyFuture(int id, boolean checkConfiguration) {
+        if (!busyLock.enterBusy()) {
+            throw new IgniteException(new NodeStoppingException());
+        }
+
+        try {
+            TableImpl table = tablesByIdVv.latest().get(id);
```
rpuch, line 2024: There are `table`, `tbl` and `table0` variables. Let's rename them to make it easier to understand the difference between them.

For instance:

`table` -> `possiblyNotReadyTable` (this is gross, probably something shorter should be devised)
`tbl` -> `readyTable`
`table0` -> `newTable`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1832,51 +1817,19 @@ public CompletableFuture<List<Table>> tablesAsync() {
...
-                            if (table != null) {
-                                tables.add((Table) table);
-                            }
-                        }
-
-                        return tables;
-                    }));
-                }));
+        return schemaSyncService.waitForMetadataCompleteness(clock.now())
+                .thenApply(ignore -> List.copyOf(latestTablesById().values()));
```
rpuch, line 1821: It looks like `tablesAsync()` works differently from `tableAsync(String)` in one aspect. The latter waits for the table that it sees as 'not yet ready' to become ready, so it will not return `null` right away for a non-ready table; but the formor will just return ready tables. So from the point of view of a user, this might look inconsistent: if there is just 1 table that is not ready yet, `tablesAsync()` will return an empty collection immediately after the metadata sync is over, but `tableAsync(String)` will wait and return the table once it's ready.

How about aligning the behaviors of the methods? Like making `tablesAsync()` wait till all tables are ready? I think this is how it worked before the changes proposed in this PR.

rpuch (commented):
The PR seems to do not what the issue says, but a different thing. The issue seems to be obsolete, so I suggest to rename and rewrite the issue to align with the changes in the PR (like 'Use schema sync instead of explicit config check')

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItDataSchemaSyncTest.java
```diff
@@ -363,4 +367,8 @@ protected ResultSet<SqlRow> sql(Ignite node, String query, Object... args) {
         }
         return rs;
     }
+
+    private static BooleanSupplier tableSchemaVersionEqPredicate(TableImpl table0, int version) {
```
rpuch, line 371: Let's rename table0 to just table

modules/runner/src/integrationTest/java/org/apache/ignite/internal/schemasync/ItSchemaSyncAndReplicationTest.java
```diff
@@ -145,12 +149,8 @@ private CompletableFuture<?> rejectionDueToMetadataLagTriggered() {
         return rejectionTriggered;
     }
 
-    private void putToTableAt(int nodeIndex) {
-        cluster.node(nodeIndex)
-                .tables()
-                .table(TABLE_NAME)
-                .keyValueView()
-                .put(null, Tuple.create().set("key", 1), Tuple.create().set("val", "one"));
+    private void putToTableAt(Table table) {
```
rpuch, line 152: `at` made sense when the parameter was an index, now I suggest to drop the suffix so that the method name becomes 'putToTable()` (or even `putTo()`)

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1837,18 +1821,20 @@ public CompletableFuture<List<Table>> tablesAsync() {
      * @return Future representing pending completion of the operation.
      */
     private CompletableFuture<List<Table>> tablesAsyncInternal() {
-        return supplyAsync(() -> inBusyLock(busyLock, this::directTableIds), ioExecutor)
-                .thenCompose(tableIds -> inBusyLock(busyLock, () -> {
-                    var tableFuts = new CompletableFuture[tableIds.size()];
+        return schemaSyncService.waitForMetadataCompleteness(clock.now())
+                .thenCompose(ignore -> inBusyLock(busyLock, () -> {
+                    NamedListView<TableView> configuredTables = tablesCfg.tables().value();
```
rpuch, line 1826: Ahem, now the method goes go the configuration. Wouldn't the approach 'take tables from tablesByIdVv, then for each of them wait till they are ready and then return the tables' work? It seems that it wouldn't require to go to the configuration

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2017,52 +1993,81 @@ public CompletableFuture<TableImpl> tableImplAsync(String name) {
...
-            TableImpl tbl = latestTablesById().get(id);
+                    if (table != null) {
+                        // Table was published.
+                        return completedFuture(table);
+                    }
 
-            if (tbl != null) {
-                return completedFuture(tbl);
-            }
+                    if (tablesCfg.tables().value().stream().noneMatch(tbl -> tbl.id() == tableId)) {
```
rpuch, line 2049: Why do we need to check in the configuration here? We've awaited for metadata completeness, the table should be in the catalog, in the config and in tablesByIdVv.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2017,52 +1993,81 @@ public CompletableFuture<TableImpl> tableImplAsync(String name) {
...
+                        return completedFuture(null);
+                    }
+
+                    // Wait for table initialization.
+                    return tableReadyFuture(tableId);
+                }, ioExecutor);
+    }
 
+    /**
+     * Internal method for getting a table future, which is completed when the table will be published.
```
rpuch, line 2060:
```suggestion
     * Internal method for getting a table future, which is completed when the table is published.
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1837,18 +1821,20 @@ public CompletableFuture<List<Table>> tablesAsync() {
      * @return Future representing pending completion of the operation.
      */
     private CompletableFuture<List<Table>> tablesAsyncInternal() {
-        return supplyAsync(() -> inBusyLock(busyLock, this::directTableIds), ioExecutor)
-                .thenCompose(tableIds -> inBusyLock(busyLock, () -> {
-                    var tableFuts = new CompletableFuture[tableIds.size()];
+        return schemaSyncService.waitForMetadataCompleteness(clock.now())
```
tkalkirill, line 1824: `TableManager#tables` may hang, and if you try to stop the `TableManager`, you will permanently hang the stop of the node.

See how I fixed this problem.
https://github.com/apache/ignite-3/blob/c63c4530f00ff67895c734e9221eac1189849d78/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java#L1444C26-L1444C26

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1864,29 +1850,19 @@ private CompletableFuture<List<Table>> tablesAsyncInternal() {
     }
 
     /**
-     * Collects a list of direct table ids.
+     * Return actual table id by given name or {@code null} if table doesn't exist.
      *
-     * @return A list of direct table ids.
+     * @param tableName Table name.
+     * @return Table id or {@code null} if not found.
```
tkalkirill, line 1856:
```suggestion
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1864,29 +1850,19 @@ private CompletableFuture<List<Table>> tablesAsyncInternal() {
...
-
-    /**
-     * Gets direct id of table with {@code tblName}.
-     *
-     * @param tblName Name of the table.
-     * @return Direct id of the table, or {@code null} if the table with the {@code tblName} has not been found.
-     */
-    @Nullable
-    private Integer directTableId(String tblName) {
+    private @Nullable Integer tableNameToId(String tableName) {
```
tkalkirill, line 1858: Why no schema synchronization?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2017,52 +1993,81 @@ public CompletableFuture<TableImpl> tableImplAsync(String name) {
...
         if (!busyLock.enterBusy()) {
             throw new IgniteException(new NodeStoppingException());
         }
 
         try {
-            return supplyAsync(() -> inBusyLock(busyLock, () -> directTableId(name)), ioExecutor)
-                    .thenCompose(tableId -> inBusyLock(busyLock, () -> {
+            HybridTimestamp now = clock.now();
+
+            return schemaSyncService.waitForMetadataCompleteness(now)
```
tkalkirill, line 2007: Same problem with node stop.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2017,52 +1993,81 @@ public CompletableFuture<TableImpl> tableImplAsync(String name) {
...
-                ? supplyAsync(() -> inBusyLock(busyLock, () -> isTableConfigured(id)), ioExecutor)
-                : completedFuture(true);
+    private CompletableFuture<TableImpl> tableAsyncInternal(int tableId) {
+        HybridTimestamp now = clock.now();
 
-        return tblCfgFut.thenCompose(isCfg -> inBusyLock(busyLock, () -> {
-            if (!isCfg) {
-                return completedFuture(null);
-            }
+        return schemaSyncService.waitForMetadataCompleteness(now)
```
tkalkirill, line 2040: Same problem with node stop.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2089,26 +2094,18 @@ public CompletableFuture<TableImpl> tableAsyncInternal(int id, boolean checkConf
...
 
-            if (tbl != null) {
+            if (table != null) {
                 assignmentsUpdatedVv.removeWhenComplete(tablesListener);
 
-                return completedFuture(tbl);
+                return completedFuture(table);
             }
 
             return getTblFut.whenComplete((unused, throwable) -> assignmentsUpdatedVv.removeWhenComplete(tablesListener));
```
tkalkirill, line 2105: Same problem with node stop.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2017,52 +1993,81 @@ public CompletableFuture<TableImpl> tableImplAsync(String name) {
...
+    }
 
+    /**
+     * Internal method for getting a table future, which is completed when the table will be published.
+     *
+     * @param id Table id.
+     * @return Future representing pending completion of the operation.
+     */
+    private CompletableFuture<TableImpl> tableReadyFuture(int id) {
+        if (!busyLock.enterBusy()) {
```
tkalkirill, line 2066: I think it's better to complete the future with an error instead of throwing exceptions.
see `org.apache.ignite.internal.util.IgniteUtils#inBusyLockAsync`

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
```diff
@@ -809,6 +810,9 @@ private TableManager createTableManager(CompletableFuture<TableManager> tblManag
         when(vaultManager.get(any(ByteArray.class))).thenReturn(completedFuture(null));
         when(vaultManager.put(any(ByteArray.class), any(byte[].class))).thenReturn(completedFuture(null));
 
+        SchemaSyncService schemaSyncService = mock(SchemaSyncService.class);
+        when(schemaSyncService.waitForMetadataCompleteness(any(HybridTimestamp.class))).thenReturn(completedFuture(null));
+
```
tkalkirill, line 815:
```suggestion
```

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/TableManagerTest.java
```diff
@@ -834,7 +838,7 @@ private TableManager createTableManager(CompletableFuture<TableManager> tblManag
                 vaultManager,
                 cmgMgr,
                 distributionZoneManager,
-                mock(SchemaSyncService.class),
+                schemaSyncService,
```
tkalkirill, line 841:
```suggestion
                mock(SchemaSyncService.class, invocation -> completedFuture(null)),
```

tkalkirill (commented):
[Here](https://github.com/apache/ignite-3/pull/2500
) I have already implemented a similar mechanism with schema synchronization and fixed all the failed tests using the catalog, maybe we finalize this PR faster?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItDataSchemaSyncTest.java
```diff
@@ -167,9 +170,11 @@ public void checkSchemasCorrectUpdate() throws Exception {
         table = (TableImpl) ignite2.tables().table(TABLE_NAME);
 
         TableImpl table0 = table;
-        assertTrue(waitForCondition(() -> table0.schemaView().schema().version() == 2, 5_000));
+        assertTrue(waitForCondition(tableSchemaVersionEqPredicate(table0, 2), 5_000));
 
-        table = (TableImpl) ignite1.tables().table(TABLE_NAME);
+        table = ((TableManager) ignite1.tables()).getTable(table.tableId());
+
+        assertFalse(waitForCondition(tableSchemaVersionEqPredicate(table, 2), 5_000));
```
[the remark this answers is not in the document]
AMashenkov: What do you mean?
Test is already fixed the way you refers to.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
AMashenkov: However, I believe, we should never do `future.join()` inside the busylock.

```
 private <T> T join(CompletableFuture<T> future) {
        if (!busyLock.enterBusy()) {
            throw new IgniteException(new NodeStoppingException());
        }

        try {
            return future.join();
        } catch (CompletionException ex) {
            throw convertThrowable(ex.getCause());
        } finally {
            busyLock.leaveBusy();
        }
    }
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
AMashenkov: What is the root cause?

`ClusterTime.waitFor` registers a future and should cancel futures on close.
This is the way all components must work - be responsible to cancel futures they returns.

The fact, we do uninterruptible awaiting`future.join()` inside the busy lock due to no reason, shouldn't force us to write 'workarounds'.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
AMashenkov: Agree, async method shouldn't throw exceptions.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: On the one hand, you are right, but on the other hand, they won’t do it in another place. 
In theory, if a component is stopped, then it is responsible to complete the future with an error, and not rely on another component.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
AMashenkov: Fixed.
Seems, we shouldn't either use configuration here or mute tests.
E.g. `TableManagerTest.testGetTableDuringCreation`failed on TC because TableManager triggered event before Catalog being updated.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2734,6 +2709,19 @@ private PartitionUpdateHandlers createPartitionUpdateHandlers(
         return startedTables.get(tableId);
     }
 
+    /**
+     * Returns a table instance if it exists, {@code null} otherwise.
+     *
+     * @param tableName Table name.
+     */
+    public @Nullable TableImpl getTable(String tableName) {
```
tkalkirill, line 2717: Please add `@TestOnly`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: You don't fix it (

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1864,38 +1848,27 @@ private CompletableFuture<List<Table>> tablesAsyncInternal() {
...
+            TableImpl table = findTableImplByName(tablesByIdVv.latest().values(), tableName);
 
-            if (exTblCfg == null) {
+            if (table == null) {
                 return null;
             } else {
-                return exTblCfg.id().value();
+                return table.tableId();
             }
         } catch (NoSuchElementException e) {
```
tkalkirill, line 1864: Can this exception happen now?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1837,18 +1819,20 @@ public CompletableFuture<List<Table>> tablesAsync() {
      * @return Future representing pending completion of the operation.
      */
     private CompletableFuture<List<Table>> tablesAsyncInternal() {
-        return supplyAsync(() -> inBusyLock(busyLock, this::directTableIds), ioExecutor)
-                .thenCompose(tableIds -> inBusyLock(busyLock, () -> {
-                    var tableFuts = new CompletableFuture[tableIds.size()];
+        return schemaSyncService.waitForMetadataCompleteness(clock.now())
+                .thenCompose(ignore -> inBusyLock(busyLock, () -> {
```
tkalkirill, line 1823: thenComposeAsync is better

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1837,18 +1819,20 @@ public CompletableFuture<List<Table>> tablesAsync() {
...
+                    for (TableImpl tbl : configuredTables) {
+                        tableFuts[i++] = tableReadyFuture(tbl.tableId());
                     }
 
                     return allOf(tableFuts).thenApply(unused -> inBusyLock(busyLock, () -> {
-                        var tables = new ArrayList<Table>(tableIds.size());
+                        var tables = new ArrayList<Table>(configuredTables.size());
 
                         for (var fut : tableFuts) {
                             var table = fut.join();
```
tkalkirill, line 1838: Below code table can be null?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1978,7 +1951,7 @@ public CompletableFuture<TableImpl> tableAsync(int id) {
             throw new IgniteException(new NodeStoppingException());
```
tkalkirill, line 1951: Again =)

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2017,52 +1990,81 @@ public CompletableFuture<TableImpl> tableImplAsync(String name) {
...
      * Gets a table by name, if it was created before. Doesn't parse canonical name.
      *
-     * @param name Table name.
+     * @param tableName Table name.
      * @return Future representing pending completion of the {@code TableManager#tableAsyncInternal} operation.
      */
-    public CompletableFuture<TableImpl> tableAsyncInternal(String name) {
+    public CompletableFuture<TableImpl> tableAsyncInternal(String tableName) {
         if (!busyLock.enterBusy()) {
             throw new IgniteException(new NodeStoppingException());
```
tkalkirill, line 1998: Again =)

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
AMashenkov: Can a table be removed concurrently?

PR #2520, opened 2023-08-30
IGNITE-20127 Implement delayed replication acks

ascherbakoff:
Implement delayed replication acks for tx commit.
limitation: doesn't work for sql - needs a separate ticket.
https://issues.apache.org/jira/browse/IGNITE-20127


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1810,10 +1845,19 @@ private CompletableFuture<Object> applyCmdWithExceptionHandling(Command cmd) {
      * Executes an Update command.
      *
      * @param cmd Update command.
-     * @return Raft future, see {@link #applyCmdWithExceptionHandling(Command)}.
+     * @return A local update ready future, possibly having a nested replication future as a result for delayed ack purpose.
      */
-    private CompletableFuture<Object> applyUpdateCommand(UpdateCommand cmd) {
+    private CompletableFuture<CompletableFuture<?>> applyUpdateCommand(UpdateCommand cmd) {
```
vldpyatkov, line 1850: I don't see it necessary to use this unclear logic. This method returns either a result or a replication future. A future wrapper is never required.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1853,36 +1902,68 @@ private CompletableFuture<Object> applyUpdateCommand(UpdateCommand cmd) {
      * @param cmd UpdateAll command.
      * @return Raft future, see {@link #applyCmdWithExceptionHandling(Command)}.
      */
-    private CompletableFuture<Object> applyUpdateAllCommand(UpdateAllCommand cmd) {
+    private CompletableFuture<CompletableFuture<?>> applyUpdateAllCommand(UpdateAllCommand cmd, boolean skipDelayedAck) {
```
vldpyatkov, line 1905: The same is before. The future wrapper is not need here.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -106,6 +117,8 @@
  * Storage of table rows.
  */
 public class InternalTableImpl implements InternalTable {
+    private static final IgniteLogger LOGGER = Loggers.forClass(InternalTableImpl.class);
```
vldpyatkov, line 120: We name it LOG, just because we use the same name everywhere.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -57,7 +65,12 @@
  *
  * <p>Uses 2PC for atomic commitment and 2PL for concurrency control.
  */
-public class TxManagerImpl implements TxManager {
+public class TxManagerImpl implements TxManager, NetworkMessageHandler {
+    private static final IgniteLogger LOGGER = Loggers.forClass(TxManagerImpl.class);
```
vldpyatkov, line 69: LOG

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -202,21 +215,85 @@ public CompletableFuture<Void> finish(
...
-                .groups(groups)
-                .commit(commit)
-                .commitTimestampLong(hybridTimestampToLong(commitTimestamp))
-                .term(term)
-                .build();
-
-        return replicaService.invoke(recipientNode, req)
-                // TODO: IGNITE-20033 TestOnly code, let's consider using Txn state map instead of states.
-                .thenRun(() -> changeState(txId, PENDING, commit ? COMMITED : ABORTED));
+        Function<Void, CompletableFuture<Void>> clo = ignored -> {
```
vldpyatkov, line 218: Why do not use `Supplier<CompletableFuture<Void>>`?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -202,21 +215,85 @@ public CompletableFuture<Void> finish(
...
+                    .commitTimestampLong(hybridTimestampToLong(commitTimestamp))
+                    .term(term)
+                    .build();
+
+            return replicaService.invoke(recipientNode, req)
+                    // TODO: IGNITE-20033 TestOnly code, let's consider using Txn state map instead of states.
+                    .thenRun(() -> changeState(txId, PENDING, commit ? COMMITED : ABORTED));
+        };
+
+        if (!commit) {
```
vldpyatkov, line 236: The logic for rollback and commit operations are duplicated. You can just make this on the commit flag:
```
// All inflights have been completed before the finish.
if (commit && tuple.inflights == 0) {
     tuple.waitRepFut.complete(null);
}
```

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -57,7 +65,12 @@
  *
  * <p>Uses 2PC for atomic commitment and 2PL for concurrency control.
  */
-public class TxManagerImpl implements TxManager {
+public class TxManagerImpl implements TxManager, NetworkMessageHandler {
```
vldpyatkov, line 68: I don't see where TxManager is used as a NetworkMessageHandler. Also, I do not understand why it is necessary to calculate inflight operations for the algorithm.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -179,16 +195,13 @@ public TxState state(UUID txId) {
...
-        TxState computeResult = states.compute(txId, (k, v) -> {
+        states.compute(txId, (k, v) -> {
             if (v == before) {
                 return after;
             } else {
                 return v;
             }
         });
-
-        assert computeResult == after : "Unable to change transaction state, expected = [" + before + "],"
```
vldpyatkov, line 190: I'm not sure if it is right to delete this asset because the redundant state changes invoked are not expected.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1396,42 +1407,61 @@ private <T> CompletableFuture<T> continueResolvingByPk(
...
-                }
+        txCleanupReadyFutures.compute(txId, (id, txOps) -> {
+            if (txOps == null) {
+                txOps = new TxCleanupReadyFutureList();
+            }
 
-                return txOps;
-            });
+            if (txOps.state == TxState.ABORTED || txOps.state == TxState.COMMITED) {
+                cleanupReadyFut.completeExceptionally(new Exception());
```
sanpwc, line 1428: new Exception()? That doesn't sounds meaningful enough.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1561,6 +1591,7 @@ private CompletableFuture<Object> processMultiEntryAction(ReadWriteMultiRowRepli
 
                 int i = 0;
 
+                // TODO sort keys ?
```
sanpwc, line 1594: Why?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1654,10 +1685,13 @@ private CompletableFuture<Object> processMultiEntryAction(ReadWriteMultiRowRepli
                         }
                     }
 
-                    CompletableFuture<Object> raftFut = rowIdsToDelete.isEmpty() ? completedFuture(null)
-                            : updateAllCommand(request, rowIdsToDelete).thenCompose(this::applyUpdateAllCommand);
+                    if (rowIdsToDelete.isEmpty()) {
+                        return completedFuture(new ReplicaResult(result, null));
```
sanpwc, line 1689: That's a tricky part. We still may want to update safeTime even if there are no rows to delete, otherwise we will need to await idle safe time propagation. We may cover this with separate optimization ticket, but for now I'd prefer calling applyUpdateAllCommand even for empty rows for the simplicity.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: Agree. Proposed API is overcomplicated.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc, line 1905: Code seems reasonably complicated. It worth adding comments explaining the flow.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1853,36 +1902,68 @@ private CompletableFuture<Object> applyUpdateCommand(UpdateCommand cmd) {
...
+                            return v;
+                        }),
+                        null);
+
+                return applyCmdWithExceptionHandling(cmd).thenApply(res -> null);
+            } else {
+                CompletableFuture<Object> fut = applyCmdWithExceptionHandling(cmd).thenApply(res -> {
+                    // Currently result is always null on a successfull execution of a replication command.
+                    // This check guaranties the result will never be lost.
+                    assert res == null : "Replication result is lost";
```
sanpwc, line 1928: Well, from my point of view it will always be null, because basically it's void.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxManager.java
```diff
@@ -149,4 +149,19 @@ CompletableFuture<Void> cleanup(
...
      */
     CompletableFuture<Void> updateLowWatermark(HybridTimestamp newLowWatermark);
+
+    /**
+     * Registers the infligh update for a transaction.
+     *
+     * @param tx The transaction id.
+     * @return {@code True} if the inflight was registered. The update must be failed on false.
+     */
+    boolean addInflight(UUID txId);
```
sanpwc, line 159: I'd say that add/removeInflight should be a part of InternalTransaction and not TxManager.
Besides that javadoc requires clarification. And what's more important sql currently totally miss the thing.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxManager.java
```diff
@@ -149,4 +149,19 @@ CompletableFuture<Void> cleanup(
      * @return Future of all read-only transactions with read timestamp less or equals the given new low watermark.
      */
     CompletableFuture<Void> updateLowWatermark(HybridTimestamp newLowWatermark);
+
+    /**
+     * Registers the infligh update for a transaction.
+     *
+     * @param tx The transaction id.
```
sanpwc, line 156: txId not tx, but as mentioned below probably not relevant because I believe we should move given method to InternalTransaction.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -434,41 +437,28 @@ private CompletableFuture<Collection<BinaryRow>> enlistCursorInTx(
...
     private <R> CompletableFuture<R> enlistWithRetry(
             InternalTransaction tx,
             int partId,
             Function<Long, ReplicaRequest> mapFunc,
-            int attempts
+            int attempts,
+            boolean full,
+            @Nullable BiPredicate<R, ReplicaRequest> noOpChecker
     ) {
+        // LOGGER.info("enlistWithRetry {} {}", partId, tx.id().toString());
```
sanpwc, line 452: Please remove.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -486,13 +476,54 @@ private <R> CompletableFuture<R> enlistWithRetry(
         return result;
     }
 
+    private <R> CompletableFuture<R> trackingInvoke(
```
sanpwc, line 479: Given method deserves javadoc.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -486,13 +476,54 @@ private <R> CompletableFuture<R> enlistWithRetry(
...
+            IgniteBiTuple<ClusterNode, Long> primaryReplicaAndTerm,
+            @Nullable BiPredicate<R, ReplicaRequest> noOpChecker
+    ) {
+        ReplicaRequest request = mapFunc.apply(primaryReplicaAndTerm.get2());
+
+        boolean write = request instanceof SingleRowReplicaRequest && ((SingleRowReplicaRequest) request).requestType() != RW_GET
+                || request instanceof MultipleRowReplicaRequest && ((MultipleRowReplicaRequest) request).requestType() != RW_GET_ALL;
+
+        if (write && !full) {
+            // Track only write requests from explicit transactions.
```
sanpwc, line 493: Why we track only write requests. AFAIK we should fail reading operations within finished transactions as well as modifications ones. WDYT?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -486,13 +476,54 @@ private <R> CompletableFuture<R> enlistWithRetry(
...
+                                partId,
+                                tx.state()
+                        )));
+            }
+
+            return replicaSvc.<R>invoke(primaryReplicaAndTerm.get1(), request).thenApply(res -> {
+                assert noOpChecker != null;
+
+                // Remove inflight if no replication was scheduled, otherwise inflight will be removed by delayed response.
+                if (noOpChecker.test(res, request)) {
```
sanpwc, line 509: noOpCheck naming seems confusing to me.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
sanpwc: Agree, assert seems useful.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -202,21 +215,85 @@ public CompletableFuture<Void> finish(
...
+            return replicaService.invoke(recipientNode, req)
+                    // TODO: IGNITE-20033 TestOnly code, let's consider using Txn state map instead of states.
+                    .thenRun(() -> changeState(txId, PENDING, commit ? COMMITED : ABORTED));
+        };
+
+        if (!commit) {
+            AtomicReference<CompletableFuture<Void>> ref = new AtomicReference<>();
+            TxContext tuple = txCtxMap.compute(txId, (uuid, tuple0) -> {
+                if (tuple0 == null) {
+                    tuple0 = new TxContext(); // No writes enlisted.
```
sanpwc, line 240: Let's be consistent with the naming. enlistment and in-flight operations have different meaning. It's minor of course.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -202,21 +215,85 @@ public CompletableFuture<Void> finish(
...
+            return tuple0;
+        });
+
+        if (ref.get() != null) { // This is committing thread.
+            // All inflights have been completed before the finish.
+            if (tuple.inflights == 0) {
+                tuple.waitRepFut.complete(null);
+            }
+
+            return tuple.waitRepFut.thenCompose(clo).handle((ignored, err) -> {
```
sanpwc, line 286: In order to be on the same page. We do wait for all in-flights to complete even in case of rollback, right?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -261,6 +261,37 @@ private void onReplicaMessageReceived(NetworkMessage message, String senderConsi
...
+                            LOG.warn("Failed to process delayed response [request={}]", ex, request);
+
+                            msg0 = prepareReplicaErrorResponse(finalSendTimestamp, ex);
+                        }
+
+                        ClusterNode sender = clusterNetSvc.topologyService().getByConsistentId(senderConsistentId);
+
+                        if (sender != null) {
+                            // Using strong send here is important to avoid a reordering with a normal response.
+                            clusterNetSvc.messagingService().send(sender, msg0);
```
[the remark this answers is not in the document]
ascherbakoff: The absence of reordering is guaranteed by current IO implementation. Both messages will use the same channel for sending a message, and a delayed message is send strictly after the initial.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1396,42 +1407,61 @@ private <T> CompletableFuture<T> continueResolvingByPk(
...
+                releaseTxLocks(txId);
+            });
+        }
 
-        if (!full) {
-            txCleanupReadyFutures.compute(txId, (id, txOps) -> {
-                if (txOps == null) {
-                    txOps = new TxCleanupReadyFutureList();
-                }
+        var cleanupReadyFut = new CompletableFuture<Void>();
```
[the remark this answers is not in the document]
ascherbakoff: This future is used to prevent races between inflight ops and commit/abort, as it was before my patch. No need to change anything here.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1810,10 +1845,19 @@ private CompletableFuture<Object> applyCmdWithExceptionHandling(Command cmd) {
...
         if (!cmd.full()) {
+            CompletableFuture<UUID> fut = applyCmdWithExceptionHandling(cmd).thenApply(res -> {
+                // This check guaranties the result will never be lost. Currently always null.
+                assert res == null : "Replication result is lost";
+
+                // Set context for delayed response.
+                return cmd.txId();
+            });
+
+            // TODO error handling similar to replication group.
```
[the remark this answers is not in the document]
ascherbakoff: No longer actual, removed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
ascherbakoff: This signature is required to implement correct chaining both for implicit and explicit tx flows. 
The result of a call either the immediate result with a delayed replication future, or immediate replication future.
Actually this is the simplest way to implement it. 
Can you propose simpler variant, which is also correct ?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
ascherbakoff: Because Supplier can not be used in completableFuture.thenCompose

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
ascherbakoff: I've removed code duplication.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
ascherbakoff: Here org.apache.ignite.internal.tx.impl.TxManagerImpl#start
Inflight calculation is required for safe committing a transaction, see ticket description.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
ascherbakoff: This assertion is no longer in the master.

modules/network/src/main/java/org/apache/ignite/network/DefaultMessagingService.java
```diff
@@ -115,6 +115,8 @@ public DefaultMessagingService(
         this.marshaller = marshaller;
 
         this.outboundExecutor = Executors.newSingleThreadExecutor(NamedThreadFactory.create(nodeName, "MessagingService-outbound-", LOG));
+        // TODO asch the implementation of delayed acks relies on absence of reordering on subsequent messages delivery.
```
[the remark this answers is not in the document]
ascherbakoff: Because if it is not held the commit can be called with some write operation in progress.
But we need to enforce the following invariant on commit - all inflight operations are finished and all corresponding replication is completed.
Actually, looks like we can relax this invariant to "all corresponding replication is completed", because it implies that all updates on primary replica is also applied.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -261,6 +261,35 @@ private void onReplicaMessageReceived(NetworkMessage message, String senderConsi
...
+                if (res.repFuture() != null) {
+                    assert request instanceof PrimaryReplicaRequest;
+
+                    res.repFuture().handle((res0, ex0) -> {
+                        NetworkMessage msg0;
+
+                        LOG.debug("Sending delayed response for replica request [request={}]", request);
+
+                        if (ex == null) {
+                            msg0 = prepareReplicaResponse(finalSendTimestamp, res0);
```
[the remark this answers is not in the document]
ascherbakoff: It is not always void. In delayed ack response txId is set as a result. See applyUpdateCommand

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -261,6 +261,35 @@ private void onReplicaMessageReceived(NetworkMessage message, String senderConsi
                     }
                 }
 
+                if (res.repFuture() != null) {
```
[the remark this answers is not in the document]
ascherbakoff: No special handling required, it just works as is. But you are correct - it may be additionally optimized. I suggest a separate ticket.

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestThreeNodesThreeReplicas.java
```diff
@@ -59,4 +69,36 @@ public void after() throws Exception {
...
+        });
+
+        assertTrue(IgniteTestUtils.waitForCondition(() -> server.blockedMessages(new RaftNodeId(groupId, leader)).size() == 2, 10000),
+                "Failed to wait for blocked messages");
+
+        ReadWriteTransactionImpl tx = (ReadWriteTransactionImpl) igniteTransactions.begin();
+        CompletableFuture<Void> fut = accounts.recordView().upsertAsync(tx, makeValue(1, 100.));
+        // Update must complete now despite the blocked replication protocol.
+        assertTrue(IgniteTestUtils.waitForCondition(fut::isDone, 5_000), "The update future is not completed within timeout");
+
```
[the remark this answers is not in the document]
ascherbakoff: I haven't

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -261,6 +261,35 @@ private void onReplicaMessageReceived(NetworkMessage message, String senderConsi
...
+                        NetworkMessage msg0;
+
+                        LOG.debug("Sending delayed response for replica request [request={}]", request);
+
+                        if (ex == null) {
+                            msg0 = prepareReplicaResponse(finalSendTimestamp, res0);
+                        } else {
+                            LOG.warn("Failed to process delayed response [request={}]", ex, request);
+
+                            msg0 = prepareReplicaErrorResponse(finalSendTimestamp, ex);
```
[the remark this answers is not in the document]
ascherbakoff: It doesn't matter here and can not break anything.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaResult.java
```diff
@@ -0,0 +1,61 @@
...
+
+package org.apache.ignite.internal.replicator;
+
+import java.util.concurrent.CompletableFuture;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Represents replica execution result.
+ */
+public class ReplicaResult {
```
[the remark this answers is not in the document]
ascherbakoff: I don't understand this. Which structure are you talking about ? ReplicaResult represent immediate and delayed results, as I previously replied.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/UpdatableTableImpl.java
```diff
@@ -176,6 +177,7 @@ public <RowT> CompletableFuture<?> upsertAll(
                     .term(nodeWithTerm.term())
                     .requestType(RequestType.RW_UPSERT_ALL)
                     .timestampLong(clock.nowLong())
+                    .skipDelayedAck(true)
```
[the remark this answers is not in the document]
ascherbakoff: SQL doesn't support this optimization yet. Without the flag, all sql txns will hang forever.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/UpdatableTableImpl.java
```diff
@@ -237,14 +239,15 @@ public <RowT> CompletableFuture<?> insertAll(
             TablePartitionId partGroupId = new TablePartitionId(tableId, partToRows.getIntKey());
             NodeWithTerm nodeWithTerm = ectx.description().mapping().updatingTableAssignments().get(partToRows.getIntKey());
 
-            ReplicaRequest request = MESSAGES_FACTORY.readWriteMultiRowReplicaRequest()
+            ReadWriteMultiRowReplicaRequest request = MESSAGES_FACTORY.readWriteMultiRowReplicaRequest()
```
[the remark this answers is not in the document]
ascherbakoff: Because of skipDelayedAck.

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestThreeNodesThreeReplicas.java
```diff
@@ -59,4 +69,36 @@ public void after() throws Exception {
             super.after();
         }
     }
+
+    @Test
+    public void testPrimaryReplicaDirectUpdateForExplicitTxn() throws InterruptedException {
+        Peer leader = accRaftClients.get(0).leader();
```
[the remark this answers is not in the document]
ascherbakoff: You are complicating things. Leader should not change during the test. There are many tests relaying on a stable leader.

modules/table/src/testFixtures/java/org/apache/ignite/internal/table/TxAbstractTest.java
```diff
@@ -1541,7 +1580,7 @@ public void run() {
                     }
 
                     while (!stop.get() && firstErr.get() == null) {
-                        InternalTransaction tx = txManager(accounts).begin();
+                        InternalTransaction tx = clientTxManager().begin();
```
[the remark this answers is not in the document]
ascherbakoff: Yes it is. Previously a transaction has been started using server txManager, instead of an client.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/LockManager.java
```diff
@@ -33,7 +33,6 @@ public interface LockManager {
      * @param lockKey The key.
      * @param lockMode Lock mode, for example shared, exclusive, intention-shared etc.
      * @return The future with gained lock that will be completed when a lock is successfully acquired.
-     * @throws LockException When a lock can't be taken due to possible deadlock.
```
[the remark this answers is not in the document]
ascherbakoff: Because this exception is never thrown.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -326,12 +331,18 @@ public PartitionReplicaListener(
...
-            return processTxStateReplicaRequest((TxStateReplicaRequest) request);
+            return processTxStateReplicaRequest((TxStateReplicaRequest) request).thenApply(res -> new ReplicaResult(res, null));
         }
 
-        return ensureReplicaIsPrimary(request).thenCompose(isPrimary -> processRequest(request, isPrimary));
+        return ensureReplicaIsPrimary(request).thenCompose(isPrimary -> processRequest(request, isPrimary).thenApply(res -> {
+            if (res instanceof ReplicaResult) {
+                return (ReplicaResult) res;
+            } else {
+                return new ReplicaResult(res, null);
```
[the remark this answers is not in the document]
ascherbakoff: This will not work, because ReplicaResult represents two results. See my previous comment about it.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
ascherbakoff: Not actual

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
ascherbakoff: Not sure if this is actual after merging.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
ascherbakoff: But the method signature says otherwise. The result can appear later, so I've added some protection.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxManager.java
ascherbakoff: I disagree. It makes sense only for RW transactions, so it conflicts with RO implementation.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
ascherbakoff: Tracking is required only for wait replication on commit purposes.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
ascherbakoff: Renamed to noWriteCheck

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
ascherbakoff: The naming is correct here - inflighs are caused by enlisted writes.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
ascherbakoff: No we do not.

sanpwc (approved):

vldpyatkov (approved):

PR #2537, opened 2023-09-01
IGNITE-20322 Add ability to pass an observable timestamp to an implicit transaction

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-20322

### DIFF
--- a/modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
+++ b/modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
@@ -661,49 +661,49 @@ public class ClientInboundMessageHandler extends ChannelInboundHandlerAdapter im
                 return ClientTableGetRequest.process(in, out, igniteTables);
 
             case ClientOp.TUPLE_UPSERT:
-                return ClientTupleUpsertRequest.process(in, out, igniteTables, resources);
+                return ClientTupleUpsertRequest.process(in, out, igniteTables, resources, igniteTransactions);
 
             case ClientOp.TUPLE_GET:
                 return ClientTupleGetRequest.process(in, out, igniteTables, resources);
 
             case ClientOp.TUPLE_UPSERT_ALL:
-                return ClientTupleUpsertAllRequest.process(in, out, igniteTables, resources);
+                return ClientTupleUpsertAllRequest.process(in, out, igniteTables, resources, igniteTransactions);
 
             case ClientOp.TUPLE_GET_ALL:
                 return ClientTupleGetAllRequest.process(in, out, igniteTables, resources);
 
             case ClientOp.TUPLE_GET_AND_UPSERT:
-                return ClientTupleGetAndUpsertRequest.process(in, out, igniteTables, resources);
+                return ClientTupleGetAndUpsertRequest.process(in, out, igniteTables, resources, igniteTransactions);
 
             case ClientOp.TUPLE_INSERT:
-                return ClientTupleInsertRequest.process(in, out, igniteTables, resources);
+                return ClientTupleInsertRequest.process(in, out, igniteTables, resources, igniteTransactions);
 
             case ClientOp.TUPLE_INSERT_ALL:
-                return ClientTupleInsertAllRequest.process(in, out, igniteTables, resources);
+                return ClientTupleInsertAllRequest.process(in, out, igniteTables, resources, igniteTransactions);
 
             case ClientOp.TUPLE_REPLACE:
-                return ClientTupleReplaceRequest.process(in, out, igniteTables, resources);
+                return ClientTupleReplaceRequest.process(in, out, igniteTables, resources, igniteTransactions);
 
             case ClientOp.TUPLE_REPLACE_EXACT:
-                return ClientTupleReplaceExactRequest.process(in, out, igniteTables, resources);
+                return ClientTupleReplaceExactRequest.process(in, out, igniteTables, resources, igniteTransactions);
 
             case ClientOp.TUPLE_GET_AND_REPLACE:
-                return ClientTupleGetAndReplaceRequest.process(in, out, igniteTables, resources);
+                return ClientTupleGetAndReplaceRequest.process(in, out, igniteTables, resources, igniteTransactions);
 
             case ClientOp.TUPLE_DELETE:
-                return ClientTupleDeleteRequest.process(in, out, igniteTables, resources);
+                return ClientTupleDeleteRequest.process(in, out, igniteTables, resources, igniteTransactions);
 
             case ClientOp.TUPLE_DELETE_ALL:
-                return ClientTupleDeleteAllRequest.process(in, out, igniteTables, resources);
+                return ClientTupleDeleteAllRequest.process(in, out, igniteTables, resources, igniteTransactions);
 
             case ClientOp.TUPLE_DELETE_EXACT:
-                return ClientTupleDeleteExactRequest.process(in, out, igniteTables, resources);
+                return ClientTupleDeleteExactRequest.process(in, out, igniteTables, resources, igniteTransactions);
 
             case ClientOp.TUPLE_DELETE_ALL_EXACT:
-                return ClientTupleDeleteAllExactRequest.process(in, out, igniteTables, resources);
+                return ClientTupleDeleteAllExactRequest.process(in, out, igniteTables, resources, igniteTransactions);
 
             case ClientOp.TUPLE_GET_AND_DELETE:
-                return ClientTupleGetAndDeleteRequest.process(in, out, igniteTables, resources);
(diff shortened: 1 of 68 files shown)


vldpyatkov:
Incorrect TC runs were attached.
https://ci.ignite.apache.org/buildConfiguration/ApacheIgnite3xGradle_Test_RunAllTests?branch=pull%2F2537&buildTypeTab=overview&mode=builds

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/client/ItThinClientObservationTsTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+        clientTable.keyValueView().put(null, Tuple.create().set(COLUMN_KEY, 42), Tuple.create().set(COLUMN_VAL, "client value"));
+
+        Transaction tx =  client().transactions().begin(new TransactionOptions().readOnly(true));
+
+        String clientValue = clientTable.keyValueView().get(tx,  Tuple.create().set(COLUMN_KEY, 42)).value(COLUMN_VAL);
+
+        tx.commit();
+
+        Transaction srvTx =  server().transactions().begin(new TransactionOptions().readOnly(true));
```
zstan, line 50: plz configure _idleSafeTimePropagationDuration_ before this and the same tests, for avoid potential flaky on jvm freeze

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/client/ItThinClientObservationTsTest.java
```diff
@@ -0,0 +1,65 @@
...
+ * This test is checking correctness of observation timestamp calculation.
+ */
+public class ItThinClientObservationTsTest extends ItAbstractThinClientTest {
+
+    @Test
+    public void test() {
+        Table srvTable = server().tables().table(TABLE_NAME);
+        Table clientTable = client().tables().table(TABLE_NAME);
+
+        log.info("Tables [client={}, srv={}]", clientTable.getClass().getSimpleName(), srvTable.getClass().getSimpleName());
```
zstan, line 38: remove or change log string

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
```diff
@@ -661,49 +661,49 @@ private void processOperation(ChannelHandlerContext ctx, ClientMessageUnpacker i
                 return ClientTableGetRequest.process(in, out, igniteTables);
 
             case ClientOp.TUPLE_UPSERT:
-                return ClientTupleUpsertRequest.process(in, out, igniteTables, resources);
+                return ClientTupleUpsertRequest.process(in, out, igniteTables, resources, igniteTransactions);
```
ascherbakoff, line 664: We don't need to track a client observable ts on server because it's already tracked on client for RO purposes:
`org.apache.ignite.internal.client.ReliableChannel#observableTimestamp`
This should be fixed. Instead, the commit ts should be propagated directly to client on commit.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/ReadWriteTransactionImpl.java
```diff
@@ -135,9 +140,24 @@ private void checkEnlistPossibility() {
...
+    @Override
+    public CompletableFuture<Void> rollbackAsync() {
+        return TransactionsExceptionMapperUtil.convertToPublicFuture(
+                finish(false, nullableHybridTimestamp(NULL_HYBRID_TIMESTAMP), false),
+                TX_ROLLBACK_ERR
+        );
+    }
+
+    @Override
+    public CompletableFuture<Void> finish(boolean commit, HybridTimestamp executionTimestamp, boolean full) {
```
ascherbakoff, line 160: Why execution timestamp is not nullable?
This is to avoid constructing useless object.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/IgniteTransactionsImpl.java
```diff
@@ -103,8 +103,18 @@ public CompletableFuture<Transaction> beginAsync(@Nullable TransactionOptions op
         return CompletableFuture.completedFuture(begin(options));
     }
 
+    /**
+     * Begins a transaction.
+     *
+     * @param readOnly {@code true} in order to start a read-only transaction, {@code false} in order to start read-write one.
+     * @return The started transaction.
+     */
+    public InternalTransaction beginImplicit(boolean readOnly) {
```
ascherbakoff, line 112: I believe it's not OK to add implicit flag to a transaction state as the part of this PR.
This introduces a log of unrelated bloating changes and better to be done as separate PR.

ascherbakoff (changes requested):

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TransactionIds.java
```diff
@@ -34,8 +37,8 @@ public class TransactionIds {
      * @param priority Transaction priority.
      * @return Transaction ID corresponding to the provided values.
      */
-    public static UUID transactionId(HybridTimestamp beginTimestamp, int nodeId, TxPriority priority) {
-        return transactionId(beginTimestamp.longValue(), nodeId, priority);
+    public static UUID transactionId(HybridTimestamp beginTimestamp, int nodeId, boolean implicit, TxPriority priority) {
```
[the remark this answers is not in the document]
vldpyatkov: Fortunately, at this moment we do not have to bother about compatibility, but I will try to move the flag out of transaction id.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/table/ClientTupleGetAllRequest.java
```diff
@@ -49,6 +49,7 @@ public static CompletableFuture<Void> process(
             ClientResourceRegistry resources
     ) {
         return readTableAsync(in, tables).thenCompose(table -> {
+            // TODO: IGNITE-23603 We have to create an implicit transaction, but leave a possibility to start RO direct.
```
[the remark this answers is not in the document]
vldpyatkov: I thought again, and currently we still have this issue only in batch get operation. 
We have to determine which transaction has to be started, RO or RW. It depends on if all keys are in one partition or not.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TransactionIds.java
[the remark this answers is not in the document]
vldpyatkov: I changed it to prived field in transaction object.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/table/ClientTupleDeleteAllExactRequest.java
```diff
@@ -35,20 +36,22 @@ public class ClientTupleDeleteAllExactRequest {
...
             ClientMessageUnpacker in,
             ClientMessagePacker out,
             IgniteTables tables,
-            ClientResourceRegistry resources
+            ClientResourceRegistry resources,
+            IgniteTransactionsImpl igniteTransactions
     ) {
         return readTableAsync(in, tables).thenCompose(table -> {
-            var tx = readTx(in, out, resources);
+            var tx = readOrStartImplicitTx(in, out, resources, igniteTransactions, false);
```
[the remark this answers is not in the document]
vldpyatkov: I fixed it in all places in the client handler where we are starting implicit transactions now.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
vldpyatkov: Currently, we can't do that. Probably we might think about this later.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/IgniteTransactionsImpl.java
vldpyatkov: The possibility to start an implicit transaction in another component is a main cause of this PR. Without this, we cannot handle an observable timestamp.

denis-chudov (approved):

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/IgniteTransactionsImpl.java
ascherbakoff: All beginXXX methods alredy accept HybridTimestampTracker as an argument.
To fix the issue in the ticket description it's not necessary to add `implicit `state field to a transaction.

ascherbakoff (approved):

PR #2581, opened 2023-09-12
IGNITE-20344 Fix rebalance recovery and add tests

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-20344

* Fix deadlock on TableManager start
* Add a test for pending assignments recovery
* Add a `copyStateTo` utility method and add some usages

---

### The Review Checklist

### Notes


modules/core/src/main/java/org/apache/ignite/internal/streamer/StreamerSubscriber.java
```diff
@@ -223,13 +225,7 @@ private void close(@Nullable Throwable throwable) {
...
             var futs = pendingRequests.values().toArray(new CompletableFuture[0]);
 
-            CompletableFuture.allOf(futs).whenComplete((res, err) -> {
-                if (err != null) {
-                    completionFut.completeExceptionally(err);
-                } else {
-                    completionFut.complete(null);
-                }
-            });
+            CompletableFuture.allOf(futs).whenComplete(copyStateTo(completionFut));
```
tkalkirill, line 228: Can use static import for `CompletableFuture#allOf`?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2202,7 +2220,11 @@ public CompletableFuture<Void> onUpdate(WatchEvent evt) {
                 try {
                     assert evt.single();
 
-                    return handleChangePendingAssignmentEvent(evt.entryEvent().newEntry());
+                    return recoveryFuture.thenComposeAsync(v -> {
```
tkalkirill, line 2223: Maybe it will be enough for us to wait for the `recoveryFuture` once during recovery and not use it again?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -589,10 +588,29 @@ private CompletableFuture<Void> performRebalanceOnRecovery(long revision) {
...
+                                    new String(pendingAssignmentEntry.key(), UTF_8),
+                                    revision
+                            );
+                        }
+
+                        // We use the Meta Storage recovery revision here instead of the entry revision, because
+                        // 'handleChangePendingAssignmentEvent' accesses some Versioned Values that only store values starting with
+                        // tokens equal to Meta Storage recovery revision. In other words, if the entry has a lower revision than the
+                        // recovery revision, there will never be a Versioned Value corresponding to its revision.
+                        return handleChangePendingAssignmentEvent(pendingAssignmentEntry, revision);
```
tkalkirill, line 604: Perhaps I'm missing something, but won't the `pendingAssignmentEntry` have the same revision as the `metaStorageMgr.recoveryFinishedFuture()`?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -1175,4 +1177,22 @@ public static <T> List<T> bytesToList(ByteBuffer buf, Function<ByteBuffer, T> tr
...
+
+    /**
+     * Creates a consumer that, when passed to a {@link CompletableFuture#whenComplete} call, will copy the outcome (either successful or
+     * not) of the target future to the given future.
+     *
+     * @param future Future to copy the outcome to.
+     * @param <T> Future result type.
+     * @return Consumer for transferring a future outcome to another future.
+     */
+    public static <T> BiConsumer<T, Throwable> copyStateTo(CompletableFuture<? super T> future) {
```
tkalkirill, line 1189: There aren't enough tests.

modules/core/src/main/java/org/apache/ignite/internal/streamer/StreamerSubscriber.java
sashapolo: This is an unrelated change overall, I don't want to add even more

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sashapolo: no, its revision will _at most_ be equal to the recovery revision, but it doesn't have to be equal (remember that revision gets changed for any key update)

modules/core/src/main/java/org/apache/ignite/internal/streamer/StreamerSubscriber.java
tkalkirill: Then, in theory, adding `copyStateTo` also does not apply to changes, since the ticket is about fixing a problem in the rebalance.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sashapolo: How will you do that? Versioned Values get completed after all components have started. I've modified the Watch code to be a little bit more optimal

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: I don't understand your comment a bit.
In your case, there is no need for an asynchronous start of recovery; you can read the keys from the metastore and then start processing them asynchronously, right?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
tkalkirill: Yes, at the same time it allows you to avoid races when updating the metastore revision, which can ruin someone’s life)

modules/core/src/main/java/org/apache/ignite/internal/streamer/StreamerSubscriber.java
sashapolo: that's my point exactly, I'm kinda doing a prohibited thing already, I don't want to make my case even worse)

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sashapolo: Please check the current approach, is it ok? 

> Yes, at the same time it allows you to avoid races when updating the metastore revision

What do you mean?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sashapolo: I need to read keys from the Meta Storage that were missed by Watch processing (e.g. while a node has been offline) and I can process them asynchonously

modules/core/src/main/java/org/apache/ignite/internal/streamer/StreamerSubscriber.java
tkalkirill: If you got in, then it’s better to do it, and not half-measure. 
You can do it on a separate ticket (trollface)

sashapolo: But what problem does a static import solve? I think it looks ok both ways

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2233,9 +2251,11 @@ public CompletableFuture<Void> onUpdate(WatchEvent evt) {
                 }
 
                 try {
-                    assert evt.single();
```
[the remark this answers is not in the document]
sashapolo: There's an identical assert inside `WatchEvent#entryEvent`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2273,7 +2300,8 @@ private CompletableFuture<Void> handleChangePendingAssignmentEvent(Entry pending
                                 replicaGrpId,
                                 tables.get(tblId),
```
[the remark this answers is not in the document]
sashapolo: I don't think we can receive null here, because we are using the *recovery revision* and, looking at the code, it seems that all VVs will be inititally completed with this revision (see `IgniteImpl#recoverComponentsStateOnStart`)

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2317,7 +2346,7 @@ private CompletableFuture<Void> handleChangePendingAssignmentEvent(
         CompletableFuture<Void> localServicesStartFuture;
 
         if (shouldStartLocalServices) {
-            localServicesStartFuture = localPartsByTableIdVv.get(pendingAssignmentsEntry.revision())
+            localServicesStartFuture = localPartsByTableIdVv.get(revision)
```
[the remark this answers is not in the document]
sashapolo: Can we? Shouldn't it be covered by the `tables.get(tblId)` check that will be introduced in a fix for the previous comment?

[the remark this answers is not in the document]
kgusakov: Yep, previous will be enough, you are right

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -496,19 +495,15 @@ public void start() {
...
-            } catch (ExecutionException e) {
-                throw new IgniteInternalException(INTERNAL_ERR, e);
-            }
+            assert recoveryFinishFuture.isDone();
+
+            long recoveryRevision = recoveryFinishFuture.join();
+
+            startTables(recoveryRevision);
+
+            performRebalanceOnRecovery(recoveryRevision);
```
tkalkirill, line 506: Can one of these methods return futures and add to `startVv.update(recoveryRevision, (v, e) -> allOf(startTablesFuture, performRebalanceOnRecoveryFuture));` once?

sashapolo: They can, but is it really better? The current approach is also fine, I think

tkalkirill (approved):

kgusakov (approved):

PR #2602, opened 2023-09-18
IGNITE-20019 Sql. Implement system views manager

xtern:
https://issues.apache.org/jira/browse/IGNITE-20019

System views are registered when the views manager starts.

In order to understand at the planning phase on which node which system views are available, a list of system views is added to the attributes of the logical node.

To collect attributes from system components before a node joins the cluster, the `NodeAttributesCollector` class was introduced, which gets attributes from component and writes them to a `ClusterNodeMessage` when a node is joined.

Public node attributes (from configuration) may be used for filtering when creating distribution zones (see [zones-filter](https://github.com/apache/ignite-3/blob/main/modules/distribution-zones/tech-notes/filters.md#zones-filter) ), in order to minimize the possible risks of mixing user attributes and internal system attributes, they are divided into two separate collections.

### DIFF
--- a/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/api/LogicalNode.java
+++ b/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/api/LogicalNode.java
@@ -36,7 +36,10 @@ public class LogicalNode extends ClusterNodeImpl {
      *         documentation</a>
      */
     @IgniteToStringInclude
-    private final Map<String, String> nodeAttributes;
+    private final Map<String, String> attributes;
+
+    @IgniteToStringInclude
+    private final Map<String, String> systemAttributes;
 
     /**
      * Constructor.
@@ -52,19 +55,32 @@ public class LogicalNode extends ClusterNodeImpl {
     ) {
         super(id, name, address);
 
-        this.nodeAttributes = Collections.emptyMap();
+        this.attributes = Collections.emptyMap();
+        this.systemAttributes = Collections.emptyMap();
     }
 
     /**
      * Constructor.
      *
-     * @param clusterNode    Represents a node in a cluster.
-     * @param nodeAttributes Node attributes.
+     * @param clusterNode Represents a node in a cluster.
+     * @param attributes  Node attributes defined in configuration.
      */
-    public LogicalNode(ClusterNode clusterNode, Map<String, String> nodeAttributes) {
+    public LogicalNode(ClusterNode clusterNode, Map<String, String> attributes) {
+        this(clusterNode, attributes, Collections.emptyMap());
+    }
+
+    /**
+     * Constructor.
+     *
+     * @param clusterNode Represents a node in a cluster.
+     * @param attributes Node attributes defined in configuration.
+     * @param systemAttributes Internal node attributes provided by system components at startup.
+     */
+    public LogicalNode(ClusterNode clusterNode, Map<String, String> attributes, Map<String, String> systemAttributes) {
         super(clusterNode.id(), clusterNode.name(), clusterNode.address(), clusterNode.nodeMetadata());
 
-        this.nodeAttributes = nodeAttributes;
+        this.attributes = attributes == null ? Collections.emptyMap() : attributes;
+        this.systemAttributes = systemAttributes == null ? Collections.emptyMap() : systemAttributes;
     }
 
     /**
@@ -73,18 +89,25 @@ public class LogicalNode extends ClusterNodeImpl {
      * @param clusterNode    Represents a node in a cluster.
      */
     public LogicalNode(ClusterNode clusterNode) {
-        super(clusterNode.id(), clusterNode.name(), clusterNode.address(), clusterNode.nodeMetadata());
-
-        this.nodeAttributes = Collections.emptyMap();
+        this(clusterNode, Collections.emptyMap(), Collections.emptyMap());
     }
 
     /**
-     * Returns node's attributes.
+     * Returns node's attributes defined in node's configuration.
      *
-     * @return Node's attributes.
+     * @return Node's attributes defined in node's configuration.
      */
-    public Map<String, String> nodeAttributes() {
-        return nodeAttributes;
+    public Map<String, String> attributes() {
+        return attributes;
+    }
+
+    /**
+     * Returns Internal node attributes provided by system components at startup.
+     *
+     * @return Internal node attributes provided by system components at startup.
+     */
+    public Map<String, String> systemAttributes() {
+        return systemAttributes;
     }
 
     /** {@inheritDoc} */
--- a/modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
+++ b/modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
@@ -47,6 +47,7 @@ import org.apache.ignite.internal.catalog.ClockWaiter;
 import org.apache.ignite.internal.catalog.configuration.SchemaSynchronizationConfiguration;
 import org.apache.ignite.internal.catalog.storage.UpdateLogImpl;
 import org.apache.ignite.internal.cluster.management.ClusterManagementGroupManager;
+import org.apache.ignite.internal.cluster.management.NodeAttributesCollector;
 import org.apache.ignite.internal.cluster.management.configuration.ClusterManagementConfiguration;
 import org.apache.ignite.internal.cluster.management.configuration.NodeAttributesConfiguration;
 import org.apache.ignite.internal.cluster.management.raft.ClusterStateStorage;
@@ -128,6 +129,7 @@ import org.apache.ignite.internal.sql.engine.SqlQueryProcessor;
 import org.apache.ignite.internal.storage.DataStorageManager;
 import org.apache.ignite.internal.storage.DataStorageModule;
 import org.apache.ignite.internal.storage.DataStorageModules;
+import org.apache.ignite.internal.systemview.SystemViewManagerImpl;
 import org.apache.ignite.internal.table.distributed.TableManager;
 import org.apache.ignite.internal.table.distributed.TableMessageGroup;
 import org.apache.ignite.internal.table.distributed.raft.snapshot.outgoing.OutgoingSnapshotsManager;
@@ -303,6 +305,9 @@ public class IgniteImpl implements Ignite {
     /** Timestamp tracker for embedded transactions. */
     private final HybridTimestampTracker observableTimestampTracker = new HybridTimestampTracker();
 
+    /** System views manager. */
+    private final SystemViewManagerImpl systemViewManager;
+
     /**
      * The Constructor.
      *
@@ -406,6 +411,9 @@ public class IgniteImpl implements Ignite {
         distributedConfigurationValidator =
                 ConfigurationValidatorImpl.withDefaultValidators(distributedConfigurationGenerator, modules.distributed().validators());
 
+        NodeAttributesCollector nodeAttributesCollector =
+                new NodeAttributesCollector(nodeConfigRegistry.getConfiguration(NodeAttributesConfiguration.KEY));
+
         cmgMgr = new ClusterManagementGroupManager(
                 vaultMgr,
                 clusterSvc,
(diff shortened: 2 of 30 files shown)


modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogManager.java
```diff
@@ -44,7 +44,7 @@ public interface CatalogManager extends IgniteComponent, CatalogService {
      * @param commands Commands to execute.
      * @return Future representing result of execution.
      */
-    CompletableFuture<Void> execute(List<CatalogCommand> commands);
+    CompletableFuture<Void> execute(Collection<CatalogCommand> commands);
```
lowka, line 47: It is better not to change this signature, because the order of commands matters.

xtern: Ok, fixed.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/api/LogicalNode.java
```diff
@@ -52,19 +56,33 @@ public LogicalNode(
...
-     * @param nodeAttributes Node attributes.
+     * @param clusterNode Represents a node in a cluster.
+     * @param attributes  Node attributes defined in configuration.
      */
-    public LogicalNode(ClusterNode clusterNode, Map<String, String> nodeAttributes) {
+    public LogicalNode(ClusterNode clusterNode, Map<String, String> attributes) {
+        this(clusterNode, attributes, Collections.emptyMap());
+    }
+
+
```
AMashenkov, line 73:
```suggestion

```

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManagerImpl.java
```diff
@@ -0,0 +1,163 @@
...
+    private final IgniteSpinBusyLock busyLock = new IgniteSpinBusyLock();
+
+    /** Prevents attempts to register a new system view after the component has started. */
+    private final AtomicBoolean startGuard = new AtomicBoolean();
+
+    /** Prevents double stopping of the component. */
+    private final AtomicBoolean stopGuard = new AtomicBoolean();
+
+    /** Future which is completed when system views are registered in the catalog. */
+    private final CompletableFuture<Void> schemaChangesFuture = new CompletableFuture<>();
```
AMashenkov, line 64:
```suggestion
    private final CompletableFuture<Void> viewsRegistrationFuture = new CompletableFuture<>();
```

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManagerImpl.java
```diff
@@ -0,0 +1,163 @@
...
+        });
+    }
+
+    @Override
+    public void stop() throws Exception {
+        if (!stopGuard.compareAndSet(false, true)) {
+            return;
+        }
+
+        busyLock.block();
```
AMashenkov, line 98: Should we try to complete `schemaChangesFuture` future exceptionally with e.g. `NodeStopException`?

xtern: Definitely should, thanks.

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManagerImpl.java
xtern: Done.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/api/LogicalNode.java
xtern: Fixed, thanks.

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManager.java
```diff
@@ -0,0 +1,35 @@
...
+public interface SystemViewManager extends IgniteComponent {
+    /**
+     * Registers a system view.
+     *
+     * <p>Registration of views is completed when the system view manager starts. Therefore,
+     * it is necessary for other components to register the views before the manager is started.
+     *
+     * @param view System view to register.
+     */
+    <T> void register(SystemView<T> view);
```
lowka, line 34: I thinks it is better to use wildcard here and remove T, since code does not use is.

modules/system-view/src/test/java/org/apache/ignite/internal/systemview/SystemViewManagerTest.java
```diff
@@ -0,0 +1,130 @@
...
+        viewMgr.start();
+
+        verify(catalog, only()).execute(anyList());
+        verifyNoMoreInteractions(catalog);
+
+        assertThat(viewMgr.nodeAttributes(), is(Map.of(NODE_ATTRIBUTES_KEY, String.join(NODE_ATTRIBUTES_LIST_SEPARATOR, name1, name2))));
+    }
+
+    @Test
+    @SuppressWarnings("ThrowableNotThrown")
```
lowka, line 102: I think `@SuppressWarnings("ThrowableNotThrown")` can be drop because Component::stop throws Exception?

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -774,7 +785,8 @@ public CompletableFuture<Ignite> start(Path configPath) {
                                     indexManager,
                                     qryEngine,
                                     clientHandlerModule,
-                                    deploymentManager
+                                    deploymentManager,
+                                    systemViewManager // Should be the last one.
```
korlov42, line 789: that's cool you put the note to keep view manager as the last one. It would be nice to add a few more words why exactly we have to keep it the last one though

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManagerImpl.java
```diff
@@ -0,0 +1,168 @@
...
+
+            if (views.isEmpty()) {
+                viewsRegistrationFuture.complete(null);
+
+                return;
+            }
+
+            catalogManager.execute(prepareCatalogCommands(views.values())).whenComplete(
+                    (r, t) -> viewsRegistrationFuture.complete(null)
+            );
```
korlov42, line 87: let's print WARN to the log in case the registration of views to the catalog has failed

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManager.java
xtern: Done, thanks.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -799,6 +811,7 @@ public CompletableFuture<Ignite> start(Path configPath) {
 
                         return cmgMgr.onJoinReady();
                     }, startupExecutor)
+                    .thenComposeAsync(v -> systemViewManager.completeRegistration())
```
korlov42, line 814: if my understanding correct, this is more like temporal solution... If so, let's add comment here describing why do we need this extra step

modules/system-view/src/test/java/org/apache/ignite/internal/systemview/SystemViewManagerTest.java
```diff
@@ -0,0 +1,130 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+/**
+ * Test class for {@link SystemViewManagerImpl}.
+ */
+@ExtendWith(MockitoExtension.class)
+public class SystemViewManagerTest extends BaseIgniteAbstractTest {
```
korlov42, line 51: let's also cover following cases:
- double start
- double stop
- start with no view registered
- register after stop 
- catalog responded with error

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManagerImpl.java
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        return catalogCommands;
+    }
+
+    private CatalogCommand prepareCatalogCommand(SystemView<?> view) {
+        List<ColumnParams> columnParams = new ArrayList<>(view.columns().size());
+
+        for (SystemViewColumn<?, ?> col : view.columns()) {
+            columnParams.add(prepareColumnParams(col));
```
lowka, line 148: I think we need to add an extra column of type `String`  for `nodeNameColumnAlias` in case of node system view.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
xtern: Done

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManagerImpl.java
xtern: Done, thanks

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManagerImpl.java
xtern: Column added, thanks.

modules/system-view/src/test/java/org/apache/ignite/internal/systemview/SystemViewManagerTest.java
xtern: Tests added.

AMashenkov: What about "double register"? 
Should it fail if 2 components registers different views with same name?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/api/LogicalNode.java
```diff
@@ -52,19 +56,32 @@ public LogicalNode(
...
+    /**
+     * Constructor.
+     *
+     * @param clusterNode Represents a node in a cluster.
+     * @param attributes Node attributes defined in configuration.
+     * @param systemAttributes Internal node attributes provided by system components at startup.
      */
-    public LogicalNode(ClusterNode clusterNode, Map<String, String> nodeAttributes) {
+    public LogicalNode(ClusterNode clusterNode, @Nullable Map<String, String> attributes, @Nullable Map<String, String> systemAttributes) {
```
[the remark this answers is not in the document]
xtern: Done, thanks.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
xtern: Added TODO)

korlov42 (approved):

lowka (approved):

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/utils/SystemViewUtils.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.ignite.internal.schema.NativeTypeSpec;
+import org.apache.ignite.internal.systemview.NodeSystemView;
+import org.apache.ignite.internal.systemview.SystemView;
+import org.apache.ignite.internal.systemview.SystemViewColumn;
+import org.apache.ignite.sql.ColumnType;
+
+/**
+ * System views utils.
+ */
+public class SystemViewUtils {
```
lowka, line 36: I think this can be package private.

lowka (approved):

korlov42 (approved):

PR #2538, opened 2023-09-02
IGNITE-20330 Create an abstraction for building indexes

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-20330

Thank you for submitting the pull request.

To streamline the review process of the patch and ensure better code quality
we ask both an author and a reviewer to verify the following:

### The Review Checklist
- [ ] **Formal criteria:** TC status, codestyle, mandatory documentation. Also make sure to complete the following:  
\- There is a single JIRA ticket related to the pull request.  
\- The web-link to the pull request is attached to the JIRA ticket.  
\- The JIRA ticket has the Patch Available state.  
\- The description of the JIRA ticket explains WHAT was made, WHY and HOW.  
\- The pull request title is treated as the final commit message. The following pattern must be used: IGNITE-XXXX Change summary where XXXX - number of JIRA issue.
- [ ] **Design:** new code conforms with the design principles of the components it is added to.
- [ ] **Patch quality:** patch cannot be split into smaller pieces, its size must be reasonable.
- [ ] **Code quality:** code is clean and readable, necessary developer documentation is added if needed.
- [ ] **Tests code quality:** test set covers positive/negative scenarios, happy/edge cases. Tests are effective in terms of execution time and resources.

### Notes
- [Apache Ignite Coding Guidelines](https://cwiki.apache.org/confluence/display/IGNITE/Java+Code+Style+Guide)

PR #2619, opened 2023-09-22
IGNITE-20339 Get rid of legacy IndexEvent, TableEvent and SchemaEvent

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-20339

### The Review Checklist

### Notes


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1329,10 +1316,11 @@ private CompletableFuture<Void> createTableLocally(
         tablesById(causalityToken).thenRun(() -> inBusyLock(busyLock, () -> completeApiCreateFuture(table)));
 
         // TODO should be reworked in IGNITE-16763
-        // We use the event notification future as the result so that dependent components can complete the schema updates.
+        // We used the event notification future as the result so that dependent components can complete the schema updates.
```
ygerzhedovich, line 1319: what event do you mena here?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -64,7 +61,7 @@
  * as well as managing indexes' lifecycle.
  */
 // TODO: IGNITE-19082 Delete this class
-public class IndexManager extends AbstractEventProducer<IndexEvent, IndexEventParameters> implements IgniteComponent {
+public class IndexManager implements IgniteComponent {
```
tkalkirill, line 64: There is no need to get rid of **AbstractEventProducer** it will be needed for IGNITE-20330

ygerzhedovich: Is there any arh document to see how it will be used?

tkalkirill: @ygerzhedovich I didn’t quite understand why an architectural document was needed, I always thought it didn’t go deep into the implementation but revealed a high-level algorithm.

AMashenkov: As I see IGNITE-20330 is opened, but has a PR with concern from @ibessonov about architecture [1]. Also, I see IndexEvent is not used in PR.

Usually, Events is used to notify external components when state was changed.
Here, IndexManager is fully responsible for index building flow. What external component needs subscribing to the IndexEvent?
As I see, Catalog can't do that because of position in dependency graph, it just know nothing about IndexManager. IndexBuilder looks like internal component relating to IndexManager, and I have no idea why it passed from the outside (however, this can be a IndexBuilder factory though). Thus, IndexManagere can call builder methods directly.

[1] https://github.com/apache/ignite-3/pull/2538/files#r1317100029

tkalkirill: > As I see IGNITE-20330 is opened, but has a PR with concern from @ibessonov about architecture [1]. Also, I see IndexEvent is not used in PR.

I closed the [PR](https://github.com/apache/ignite-3/pull/2538) because it is no longer valid because there will be a new implementation. We personally discussed (with @ibessonov and @sanpwc) the past decision and came to a new decision that will use the `IndexEvent` and `PrimaryReplicaEvent`.

> Usually, Events are used to notify external components (e.g. which we can't foresee in constructor) when state was changed.
> 
> Here, IndexManager is fully responsible for index building flow. What external component needs subscribing to the IndexEvent?

A new component `IndexBuildController` will appear, which will run the index building procedure, it will listen to events from `IndexManager` and `PlacementDriver` and react correctly to the its events.

This is necessary firstly for less influence on the code of other components and also for merging two types of events:
1. Index appeared/deleted.
2. Appearance/loss of primacy of a replica.

AMashenkov: Moreover, I think IgniteManager is a IgniteComponent in the first place, and not an event Producer. The current approach with extending Producer class is wrong.

If we decide at some point got extract common component behaviour into an abstract class, then a refactoring must be done. Let's rewrite event processing using incapsulation instead of inherintance.

ibessonov: > Thus, IndexManager can just call builder methods directly.

This is precisely my concern, I don't want index manager to depend on our internal stuff. Events allows us to remove such unnecessary dependencies.

tkalkirill: > Moreover, I think IgniteManager is a IgniteComponent in the first place, and not an event Producer. The current approach with extending Producer class is wrong.

Why wrong?

> If we decide at some point got extract common component behaviour into an abstract class, then a refactoring must be done. Let's rewrite event processing using incapsulation instead of inherintance.

I think that if we come up with something else, we will need to refactor something.
What do you propose it will look like with incapsulation? Will it be something like: `IndexManager#eventProduser` ?

We would also need to decide where the creation of index storage (`MvTableStorage#getOrCreateIndex`) will take place, since there is a [IGNITE-19112](https://issues.apache.org/jira/browse/IGNITE-19112) to fixing this.
The manager is in the right place for me at the moment - they listen the index creation event from catalog, created index storages.

AMashenkov: > A new component IndexBuildController will appear, which will run the index building procedure, it will listen to events from IndexManager and PlacementDriver and react correctly to the its events.

Should IndexBuildController listen for Catalog event directly, when an index is created.

tkalkirill: Look what happens, in the current ticket you will delete these events, then I will return them anyway, and I will need to deal with conflicts and return tests (it is clear that new ones will need to be added). Double work (even triple work) will be done, which would be better spent on other problems and/or bugs or tasks.

ibessonov: Generally speaking, it is a dead code, and it looks inappropriate at times. So let's see on our plans, and if there's no JIRA or its intentions are not clear, we may as well proceed with deleting these classes. Returning them back won't be that hard

tkalkirill: We discussed it personally (with @AMashenkov and @ibessonov) and decided to get rid of the events.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -147,35 +144,22 @@ public void stop() throws Exception {
     }
 
     private CompletableFuture<Boolean> onIndexDrop(DropIndexEventParameters parameters) {
-        int indexId = parameters.indexId();
-        int tableId = parameters.tableId();
-
-        long causalityToken = parameters.causalityToken();
-        int catalogVersion = parameters.catalogVersion();
-
         if (!busyLock.enterBusy()) {
```
tkalkirill, line 147: U can use `IgniteUtils#inBusyLockAsync`

modules/index/src/main/java/org/apache/ignite/internal/index/IndexManager.java
```diff
@@ -193,12 +177,6 @@ private CompletableFuture<Boolean> onIndexCreate(CreateIndexEventParameters para
         int catalogVersion = parameters.catalogVersion();
 
         if (!busyLock.enterBusy()) {
```
tkalkirill, line 179: same

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -319,16 +305,7 @@ public synchronized void stop() throws Exception {
 
         Stream<AutoCloseable> closableComponents = services.stream().map(s -> s::stop);
```
tkalkirill, line 306:
```suggestion
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -319,16 +305,7 @@ public synchronized void stop() throws Exception {
...
-                .map((p) -> () -> {
-                    if (p.left instanceof TableEvent) {
-                        tableManager.removeListener((TableEvent) p.left, p.right);
-                    } else {
-                        indexManager.removeListener((IndexEvent) p.left, p.right);
-                    }
-                });
-
-        IgniteUtils.closeAll(Stream.concat(closableComponents, closableListeners).collect(Collectors.toList()));
+        IgniteUtils.closeAll(closableComponents);
```
tkalkirill, line 308:
```suggestion
        IgniteUtils.closeAll(services.stream().map(s -> s::stop));
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -651,8 +642,6 @@ private CompletableFuture<Void> onTableDelete(DropTableEventParameters parameter
         int tableId = parameters.tableId();
 
         if (!busyLock.enterBusy()) {
```
tkalkirill, line 644: U can use IgniteUtils#inBusyLockAsync

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1217,8 +1206,6 @@ private CompletableFuture<?> createTableLocally(long causalityToken, int catalog
         int tableId = tableDescriptor.id();
 
         if (!busyLock.enterBusy()) {
```
tkalkirill, line 1208: Same

tkalkirill (approved):

ygerzhedovich (approved):

tkalkirill:
In my PR, I deleted `IndexEvent` from `IndexManager` and `IndexManagerTest` but added several new tests to test the `IndexManager#getMvTableStorage`, please do not delete `IndexManagerTest` in your PR.

PR #2622, opened 2023-09-24
IGNITE-20385 Sql. Fix handling of public exceptions on the remote node.

xtern:
https://issues.apache.org/jira/browse/IGNITE-20385

`NodeLeftException` is handled in the following places:

- _ExecutionServiceImpl#execute_
`NodeLeftException` is thrown above, then `mapToPublicSqlException` maps it to `SqlException` with code `NODE_LEFT_ERR`

- _Outbobx#sendBatches_
 here any exception will always be converted to a `RemoteFragmentExecutionException` with the original code preserved, then the `RemoteFragmentExecutionException` is mapped to a `SqlException`, the output is a `SqlException` with the code `NODE_LEFT_ERR`

- _Inbox#requestBatches_
same as previous
 
- _Outbobx#sendError_
just write to the log if you couldn’t send the original message with an error

The change from `withCauseAndCode` to `withCode` is required to preserve the original error code.
The change in `withCauseInternal` method is required to preserve the original error code for _public_ exception.


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/NodeLeftException.java
```diff
@@ -29,7 +29,7 @@ public class NodeLeftException extends IgniteException {
     private static final long serialVersionUID = 0L;
 
     /** Constructor. */
-    public NodeLeftException(String nodeName) {
-        super(NODE_LEFT_ERR, "Node left the cluster. Node: " + nodeName);
```
ygerzhedovich, line 33: What reason to move standart text message for the exception to invoke side? I prefer have message template inside a exceptions if it possible.

xtern: it is not recommended to change the exception message in the constructor, because of this issue: https://issues.apache.org/jira/browse/IGNITE-20140

modules/core/src/main/java/org/apache/ignite/internal/lang/IgniteExceptionMapperUtil.java
```diff
@@ -83,31 +86,39 @@ static void registerMapping(
...
+     * @return Public exception.
+     */
+    public static Throwable mapToPublicSqlException(Throwable origin) {
+        return mapToPublicExceptionInternal(origin, (e) -> {
+            if (e instanceof TraceableException) {
+                TraceableException traceable = (TraceableException) e;
 
-            return origin;
-        }
+                return new SqlException(traceable.traceId(), traceable.code(), e.getMessage(), e);
```
ygerzhedovich, line 117: it could be already SqlExceptin and should be just rethrow

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/NodeLeftException.java
xtern: Done

ygerzhedovich (approved):

ygerzhedovich (approved):

AMashenkov (approved):

PR #2628, opened 2023-09-25
IGNITE-20410 CheckCatalogVersionOnActionRequest breaks RAFT client leaderId refresh protocol

rpuch:
https://issues.apache.org/jira/browse/IGNITE-20410

### The Review Checklist

### Notes

### DIFF
--- a/modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
+++ b/modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
@@ -1592,14 +1592,9 @@ public class NodeImpl implements Node, RaftServerService {
         this.writeLock.lock();
         try {
             final int size = tasks.size();
-            if (this.state != State.STATE_LEADER) {
-                final Status st = new Status();
-                if (this.state != State.STATE_TRANSFERRING) {
-                    st.setError(RaftError.EPERM, "Is not leader.");
-                }
-                else {
-                    st.setError(RaftError.EBUSY, "Is transferring leadership.");
-                }
+            State nodeState = this.state;
+            if (nodeState != State.STATE_LEADER) {
+                final Status st = cannotApplyBecauseNotLeaderStatus(nodeState);
                 LOG.debug("Node {} can't apply, status={}.", getNodeId(), st);
                 final List<Closure> dones = tasks.stream().map(ele -> ele.done)
                         .filter(Objects::nonNull).collect(Collectors.toList());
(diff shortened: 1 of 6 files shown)


modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
```diff
@@ -1645,6 +1640,23 @@ private void executeApplyingTasks(final List<LogEntryAndClosure> tasks) {
         }
     }
 
+    /**
+     * Builds a status for 'Cannot apply because this node is not a leader' situation.
+     *
+     * @param nodeState Node state.
+     */
+    public static Status cannotApplyBecauseNotLeaderStatus(State nodeState) {
```
ibessonov, line 1648: There's nothing about "cannot apply" in the status itself. Why did you chose such a specific name?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/CheckCatalogVersionOnActionRequest.java
```diff
@@ -59,20 +72,53 @@ public CheckCatalogVersionOnActionRequest(CatalogService catalogService) {
...
+
+        if (state == State.STATE_LEADER) {
+            return null;
+        }
+
+        Status st = NodeImpl.cannotApplyBecauseNotLeaderStatus(state);
+
+        PeerId leaderId = node.getLeaderId();
+
+        if (st.getRaftError() == RaftError.EPERM && leaderId != null) {
```
ibessonov, line 113: What does this mean? Please leave a comment that it's "not a leader" status.
By the way, why is it necessary to check the code? Is it not enough to have a leaderId check?
Last question, why do we create responses differently in these cases?

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
rpuch: Because this method checks for this specific situation. When it finds out that 'it cannot apply because not a leader', it returns this status.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/CheckCatalogVersionOnActionRequest.java
rpuch: I added a comment and simplified the code.

As for 'why don't just return leaderId if we have it', take a look at `IgniteCliRpcRequestClosure:74`: they only enrich the response with leaderId for EPERM. I did not want to change this to avoid breaking this protocol.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
ibessonov: From what I see, method knows nothing about "cannot apply", it only converts one status to another, that's why I asked the question. But I see what you mean

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/schema/CheckCatalogVersionOnActionRequest.java
ibessonov: That's crazy, thank you for the explanation

ibessonov (approved):

alievmirza (approved):

PR #2641, opened 2023-09-28
IGNITE-19824 Implicit RO should be used in implicit single gets

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-19824

### DIFF
--- a/modules/replicator/src/main/java/org/apache/ignite/internal/replicator/message/ReadOnlyDirectReplicaRequest.java
+++ b/modules/replicator/src/main/java/org/apache/ignite/internal/replicator/message/ReadOnlyDirectReplicaRequest.java
@@ -20,8 +20,8 @@ package org.apache.ignite.internal.replicator.message;
 /**
  * Read-only request that sand to a specific node directly. This request has no read timestamp as other read-only requests because the
  * timestamp is calculated on the replica side. The linearization is guaranteed by sending the request directly to the primary node.
- * <p>
- * The requests are used to implement an implicit read-only transaction for a single partition.
+ *
+ * <p>The requests are used to implement an implicit read-only transaction for a single partition.
  */
 public interface ReadOnlyDirectReplicaRequest extends ReplicaRequest {
     /**
--- a/modules/replicator/src/main/java/org/apache/ignite/internal/replicator/message/ReadOnlyDirectReplicaRequest.java
+++ b/modules/replicator/src/main/java/org/apache/ignite/internal/replicator/message/ReadOnlyDirectReplicaRequest.java
@@ -15,12 +15,20 @@
  * limitations under the License.
  */
 
-package org.apache.ignite.internal.table.distributed.replication.request;
-
-import org.apache.ignite.internal.replicator.message.ReplicaRequest;
+package org.apache.ignite.internal.replicator.message;
 
 /**
- * Read only direct replica request.
+ * Read-only request that sand to a specific node directly. This request has no read timestamp as other read-only requests because the
+ * timestamp is calculated on the replica side. The linearization is guaranteed by sending the request directly to the primary node.
+ * <p>
+ * The requests are used to implement an implicit read-only transaction for a single partition.
  */
 public interface ReadOnlyDirectReplicaRequest extends ReplicaRequest {
+    /**
+     * Gets an enlistment consistency token.
+     * The token is used to check that the lease is still actual while the message goes to the replica.
+     *
+     * @return Enlistment consistency token.
+     */
+    Long enlistmentConsistencyToken();
 }
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
@@ -125,6 +125,8 @@ import org.apache.ignite.internal.table.distributed.replication.request.BinaryRo
 import org.apache.ignite.internal.table.distributed.replication.request.BinaryTupleMessage;
 import org.apache.ignite.internal.table.distributed.replication.request.BuildIndexReplicaRequest;
 import org.apache.ignite.internal.table.distributed.replication.request.CommittableTxRequest;
+import org.apache.ignite.internal.table.distributed.replication.request.ReadOnlyDirectMultiRowReplicaRequest;
+import org.apache.ignite.internal.table.distributed.replication.request.ReadOnlyDirectSingleRowReplicaRequest;
 import org.apache.ignite.internal.table.distributed.replication.request.ReadOnlyMultiRowPkReplicaRequest;
 import org.apache.ignite.internal.table.distributed.replication.request.ReadOnlyReplicaRequest;
 import org.apache.ignite.internal.table.distributed.replication.request.ReadOnlyScanRetrieveBatchReplicaRequest;
@@ -448,6 +450,10 @@ public class PartitionReplicaListener implements ReplicaListener {
             return processReplicaSafeTimeSyncRequest((ReplicaSafeTimeSyncRequest) request, isPrimary);
         } else if (request instanceof BuildIndexReplicaRequest) {
             return raftClient.run(toBuildIndexCommand((BuildIndexReplicaRequest) request));
+        } else if (request instanceof ReadOnlyDirectSingleRowReplicaRequest) {
+            return processReadOnlyDirectSingleEntryAction((ReadOnlyDirectSingleRowReplicaRequest) request);
+        } else if (request instanceof ReadOnlyDirectMultiRowReplicaRequest) {
+            return processReadOnlyDirectMultiEntryAction((ReadOnlyDirectMultiRowReplicaRequest) request);
         } else {
             throw new UnsupportedReplicaRequestException(request.getClass());
         }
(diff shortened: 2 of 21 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1987,6 +2031,24 @@ private CompletableFuture<Object> applyUpdateAllCommand(UpdateAllCommand cmd) {
...
+    private CompletableFuture<BinaryRow> processReadOnlyDirectSingleEntryAction(ReadOnlyDirectSingleRowReplicaRequest request) {
+        BinaryTuple primaryKey = resolvePk(request.primaryKey());
+        HybridTimestamp readTimestamp = hybridClock.now();
+
+        if (request.requestType() != RequestType.RO_GET) {
+            throw new IgniteInternalException(Replicator.REPLICA_COMMON_ERR,
+                    format("Unknown single request [actionType={}]", request.requestType()));
+        }
+
+        return resolveRowByPkForReadOnly(primaryKey, readTimestamp);
```
sanpwc, line 2049: We will wrap the result with TimestampAware response in order to adjust the clock on the client side?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -547,28 +549,163 @@ private <T> CompletableFuture<T> postEnlist(CompletableFuture<T> fut, boolean au
...
+        TablePartitionId tablePartitionId = new TablePartitionId(tableId, partId);
+
+        CompletableFuture<ReplicaMeta> primaryReplicaFuture = placementDriver.awaitPrimaryReplica(tablePartitionId, tx.startTimestamp())
+                .orTimeout(AWAIT_PRIMARY_REPLICA_TIMEOUT, TimeUnit.SECONDS);
+
+        CompletableFuture<R>  fut = primaryReplicaFuture.thenCompose(primaryReplica -> {
+            try {
+                ClusterNode node = clusterNodeResolver.apply(primaryReplica.getLeaseholder());
+
+                if (node == null) {
```
sanpwc, line 577: How it's possible?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -589,35 +726,68 @@ public CompletableFuture<BinaryRow> get(
         );
     }
 
+    /**
+     * Checks that the batch of rows belongs to one partition.
+     *
+     * @param rows Batch of rows.
+     * @return If all rows belong to one partition, the method returns true; otherwise, it returns false.
+     */
+    private boolean isOnePartitionBatch(Collection<BinaryRowEx> rows) {
```
sanpwc, line 735: It's better to use Single instead of One here. Same in javadoc.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1987,6 +2031,24 @@ private CompletableFuture<Object> applyUpdateAllCommand(UpdateAllCommand cmd) {
...
+     * Processes single entry direct request for read only transaction.
+     *
+     * @param request Read only single entry request.
+     * @return Result future.
+     */
+    private CompletableFuture<BinaryRow> processReadOnlyDirectSingleEntryAction(ReadOnlyDirectSingleRowReplicaRequest request) {
+        BinaryTuple primaryKey = resolvePk(request.primaryKey());
+        HybridTimestamp readTimestamp = hybridClock.now();
+
+        if (request.requestType() != RequestType.RO_GET) {
```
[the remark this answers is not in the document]
vldpyatkov: No

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1987,6 +2031,24 @@ private CompletableFuture<Object> applyUpdateAllCommand(UpdateAllCommand cmd) {
...
 
+    /**
+     * Processes single entry direct request for read only transaction.
+     *
+     * @param request Read only single entry request.
+     * @return Result future.
+     */
+    private CompletableFuture<BinaryRow> processReadOnlyDirectSingleEntryAction(ReadOnlyDirectSingleRowReplicaRequest request) {
+        BinaryTuple primaryKey = resolvePk(request.primaryKey());
+        HybridTimestamp readTimestamp = hybridClock.now();
```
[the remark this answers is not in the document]
vldpyatkov: It just because the request has no read timestamp.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
vldpyatkov: We do not adjust the clock for any read-only transactions. Why should we do it for this one?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
vldpyatkov: Each time when we find a node in topology by its identifier it is possible.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
vldpyatkov: Done.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
sanpwc: Yep. You are right.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/message/ReadOnlyDirectReplicaRequest.java
```diff
@@ -0,0 +1,34 @@
...
+ * <p>The requests are used to implement an implicit read-only transaction for a single partition.
+ */
+public interface ReadOnlyDirectReplicaRequest extends ReplicaRequest {
+    /**
+     * Gets an enlistment consistency token.
+     * The token is used to check that the lease is still actual while the message goes to the replica.
+     *
+     * @return Enlistment consistency token.
+     */
+    Long enlistmentConsistencyToken();
```
sanpwc, line 33: I don't think that we need to check whether it's the **same** primary replica. It's enough just to check whether we are on primary or not while processing RODirectRequest on server side.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/InternalTransaction.java
```diff
@@ -105,4 +105,15 @@ public interface InternalTransaction extends Transaction {
...
     boolean implicit();
+
+    /**
+     * Finishes a read-only transaction with a specific execution timestamp.
+     *
+     * @param commit Commit flag. The flag is ignored for read-only transactions.
+     * @param executionTimestamp The timestamp is the time when a read-only transaction is applied to the remote node.
+     * @return The future.
+     */
+    default CompletableFuture<Void> finish(boolean commit, HybridTimestamp executionTimestamp) {
```
sanpwc, line 116: What's that?
> executionTimestamp The timestamp is the time when a read-only transaction is applied to the remote node.
It's still not clear.
> And it's invalid to have read only specific mentions in common InternalTransaction.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/ReadOnlyTransactionImpl.java
```diff
@@ -39,17 +40,22 @@ class ReadOnlyTransactionImpl extends IgniteAbstractTransactionImpl {
...
+     * @param observableTsTracker Observable timestamp tracker.
      * @param id The id.
      * @param readTimestamp The read timestamp.
      */
-    ReadOnlyTransactionImpl(TxManagerImpl txManager, UUID id, HybridTimestamp readTimestamp) {
+    ReadOnlyTransactionImpl(TxManagerImpl txManager, HybridTimestampTracker observableTsTracker, UUID id, HybridTimestamp readTimestamp) {
         super(txManager, id);
 
         this.readTimestamp = readTimestamp;
+        this.observableTsTracker = observableTsTracker;
```
sanpwc, line 58: Why we need it?
How do we update observableTsTracker in case of implicit RW transcion? I'd say that we should do it same way for implicit RO.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/InternalTransaction.java
sanpwc, line 116: It's not a good idea to mention read-only specifics in common InternalTransacion.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/InternalTransaction.java
```diff
@@ -105,4 +105,15 @@ public interface InternalTransaction extends Transaction {
      * provided no transaction for an operation) or not.
      */
     boolean implicit();
+
+    /**
+     * Finishes a read-only transaction with a specific execution timestamp.
+     *
+     * @param commit Commit flag. The flag is ignored for read-only transactions.
+     * @param executionTimestamp The timestamp is the time when a read-only transaction is applied to the remote node.
```
sanpwc, line 113: It's not a good idea to mention read-only specifics in common InternalTransacion. It also worth to mention, that executionTimestamp is equals or **reasonably greater** to the time when a read-only transactions **read timestamp**.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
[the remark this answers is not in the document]
sanpwc: So, where's the explanation?

sanpwc (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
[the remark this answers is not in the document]
vldpyatkov: It is the ReadOnlyDirectReplicaRequest

modules/transactions/src/main/java/org/apache/ignite/internal/tx/InternalTransaction.java
vldpyatkov: Unfortunately, we already have methods for specific transaction types in this interface.

PR #2643, opened 2023-09-29
IGNITE-20518 Use CatalogService in JdbcMetadataCatalog

rpuch:
https://issues.apache.org/jira/browse/IGNITE-20518

### The Review Checklist

### Notes

### DIFF
--- a/modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/jdbc/JdbcMetadataCatalog.java
+++ b/modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/jdbc/JdbcMetadataCatalog.java
@@ -17,6 +17,10 @@
 
 package org.apache.ignite.client.handler.requests.jdbc;
 
+import static java.util.stream.Collectors.toCollection;
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toSet;
+
 import java.sql.DatabaseMetaData;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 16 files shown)


modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/jdbc/JdbcMetadataCatalog.java
```diff
@@ -182,28 +202,24 @@ public CompletableFuture<Collection<String>> getSchemasMeta(String schemaNamePtr
...
      *
      * @param tbl Table.
      * @return Jdbc primary key metadata.
      */
-    private JdbcPrimaryKeyMeta createPrimaryKeyMeta(Table tbl) {
-        final String keyName = PK + tbl.name();
-
-        SchemaRegistry registry = ((TableImpl) tbl).schemaView();
+    private JdbcPrimaryKeyMeta createPrimaryKeyMeta(CatalogTableDescriptor tbl) {
+        String keyName = PK + tbl.name();
```
ibessonov, line 220: Unrelated question: what happens after table rename? Do we lose the ability to access the PK index?

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/jdbc/JdbcMetadataCatalog.java
```diff
@@ -182,28 +202,24 @@ public CompletableFuture<Collection<String>> getSchemasMeta(String schemaNamePtr
...
 
-        List<String> keyColNames = Arrays.stream(registry.schema().keyColumns().columns())
-                .map(Column::name)
-                .collect(Collectors.toList());
+        List<String> keyColNames = new ArrayList<>(tbl.primaryKeyColumns());
```
ibessonov, line 222: Any particular reason of creating the new list?

ibessonov (approved):

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/jdbc/JdbcMetadataCatalog.java
rpuch: Probably, when we rename a table, we'll also (atomically) rename the corresponding PK. Should I file an issue?

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/jdbc/JdbcMetadataCatalog.java
rpuch: To make sure that, if `tbl.primaryKeyColumns()` is modifiable, we don't give it away

rpuch: Replaced it with `List.copyOf()` which will not allocate anything if the original list is immutable

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/jdbc/JdbcMetadataCatalog.java
ibessonov: I don't know, we should discuss this behavior with someone else too

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/jdbc/JdbcMetadataCatalog.java
ibessonov: I assume that if the list is returned from a method to external user, we can be user that it can be used for reading. Such issues are solved by documentation. Making copies of data, just to be sure, might not be the best solution

rpuch: I disagree. Noone reads documentation. The pattern is called 'defensive copy', it's widely used. Also, please note that this code is not on the hot path. A couple of instances on the heap will not crash the server.

ibessonov: > The pattern is called 'defensive copy', it's widely used.

You know that's bad for performance, right? How hot is this code?

PR #2650, opened 2023-10-03
IGNITE-20346 ODBC: Implement column metadata fetching

isapego:
- Ported code from Ignite 2;
- Adjusted to Calcite. Fixed some issues found in the process;
- Un-commented tests;
- Updated DEVNOTES.md.


modules/platforms/cpp/ignite/odbc/query/column_metadata_query.h
```diff
@@ -0,0 +1,188 @@
...
+    std::optional<std::string> schema;
+
+    /** Table name. */
+    std::optional<std::string> table;
+
+    /** Column name. */
+    std::optional<std::string> column;
+
+    /** Data type. */
+    ignite_type data_type{ignite_type::BOOLEAN};
```
ptupitsyn, line 51: Why `ignite_type::BOOLEAN`?

modules/platforms/cpp/ignite/odbc/type_traits.cpp
```diff
@@ -596,8 +596,36 @@ std::int32_t ignite_type_num_precision_radix(ignite_type typ) {
     return sql_type_num_precision_radix(sql_type);
 }
 
-std::int32_t sql_type_decimal_digits(std::int16_t) {
-    // TODO: IGNITE-19854 Remove once parameters meta fetching is implemented
+std::int32_t sql_type_decimal_digits(std::int16_t type) {
```
ptupitsyn, line 599: Should we rename this method? Currently it is confusing: for `DECIMAL` we return `-1`.

ptupitsyn (approved):
Looks good to me in general; please see a couple of minor comments.

modules/platforms/cpp/ignite/odbc/query/column_metadata_query.h
isapego: Good catch. I was going to put `UNDEFINED`. Probably just a slip.

modules/platforms/cpp/ignite/odbc/type_traits.cpp
isapego: I decided to add a `scale` parameter for this function instead. It's name confusing, but it's [ODBC terminology](https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/column-size-decimal-digits-transfer-octet-length-and-display-size), it's really very confusing in this aspect.

ptupitsyn (approved):

PR #2653, opened 2023-10-03
IGNITE-20240 C++ 3.0: Reject Tuples with unmapped fields

isapego:
- Checked that number of written columns is no less than total number of columns in tuple. Reject tuple if there are any extra columns;
- Added tests.

### DIFF
--- a/modules/platforms/cpp/ignite/client/detail/utils.cpp
+++ b/modules/platforms/cpp/ignite/client/detail/utils.cpp
@@ -237,7 +237,7 @@ std::vector<std::byte> pack_tuple(
         unmapped_columns_str.pop_back();
 
         assert(!unmapped_columns_str.empty());
-        throw ignite_error("Tuple contains columns that are not present in a table: " + unmapped_columns_str);
+        throw ignite_error("Tuple contains columns that are not present in the table: " + unmapped_columns_str);
     }
 
     return builder.build();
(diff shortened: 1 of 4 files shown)


modules/platforms/cpp/ignite/client/detail/utils.cpp
```diff
@@ -201,19 +201,45 @@ std::vector<std::byte> pack_tuple(
...
+            if (written_ind[i])
+                continue;
+            auto &name = tuple.column_name(i);
+            unmapped_columns << name << ",";
+        }
+        auto unmapped_columns_str = unmapped_columns.str();
+        unmapped_columns_str.pop_back();
+
+        assert(!unmapped_columns_str.empty());
+        throw ignite_error("Tuple contains columns that are not present in the table: " + unmapped_columns_str);
```
ptupitsyn, line 240: I think we should use the error message format from `TupleMarshallerImpl`: "Key tuple doesn't match schema: schemaVersion=%s, extraColumns=%s"

ptupitsyn (changes requested):

modules/platforms/cpp/ignite/client/detail/utils.cpp
isapego: Good point.

ptupitsyn (approved):

PR #2658, opened 2023-10-03
IGNITE-20457 Verify commitTimestamp against enlisted partitions expiration timestamps

sanpwc:
https://issues.apache.org/jira/browse/IGNITE-20457

- Added TxManagerImpl#verifyCommitTimestamp method that checks whether previously enlisted primary replicas aren't expired and that commit timestamp is less or equal than primary replicas expiration timestamp. Given method will either complete result future with void or PrimaryReplicaExpiredException 
- Added aforementioned verifyCommitTimestamp into TxManagerImpl#finish flow
- Added PlacementDriver as an additional TxManagerImpl constructor param in order to retrieve current primary replicas for the futher checks in verifyCommitTimestamp.


modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/ReadWriteTransactionImpl.java
```diff
@@ -96,44 +102,60 @@ public IgniteBiTuple<ClusterNode, Long> enlist(TablePartitionId tablePartitionId
...
                     id()
             );
         }
+
+        return finishFuture.exceptionally(
+                e -> {
+                    if (e instanceof IgniteInternalException) {
+                        // Transaction is already rolled back. Just wrapping internal exception with public one.
+                        throw withCause(TransactionException::new, TX_PRIMARY_REPLICA_EXPIRED_ERR,
+                                "Primary replica has expired, transaction will be rolled back", e);
```
vldpyatkov, line 152: Why are you shure about the reason, when we checked an exception type only?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -530,6 +558,41 @@ private TxStateMeta coordinatorFinalTxStateMeta(boolean commit, HybridTimestamp
...
+
+        for (Map.Entry<TablePartitionId, Long> enlistedGroup : enlistedGroups.entrySet()) {
+            TablePartitionId groupId = enlistedGroup.getKey();
+            Long expectedEnlistmentConsistencyToken = enlistedGroup.getValue();
+
+            verificationFutures[++cnt] = placementDriver.getPrimaryReplica(groupId, commitTimestamp)
+                    .thenAccept(currentPrimaryReplica -> {
+                        if (currentPrimaryReplica == null
+                                || !expectedEnlistmentConsistencyToken.equals(currentPrimaryReplica.getStartTime().longValue())
+                                || commitTimestamp.after(currentPrimaryReplica.getExpirationTime())
```
vldpyatkov, line 581: Is it really possible when you pass a commit timstamp in getPrimaryReplica method?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -290,37 +297,58 @@ public CompletableFuture<Void> finish(
...
+            CompletableFuture<Void> verificationFuture =
+                    commit ? verifyCommitTimestamp(enlistedGroups, commitTimestamp) : completedFuture(null);
+
+            return verificationFuture.handle(
+                    (unused, throwable) -> {
+                        // TODO: https://issues.apache.org/jira/browse/IGNITE-19170 Use ZonePartitionIdMessage and remove cast
+                        Collection<ReplicationGroupId> replicationGroupIds = new HashSet<>();
+                        replicationGroupIds.addAll((Collection<ReplicationGroupId>) (Collection<?>) enlistedGroups.keySet());
+
+                        boolean verifiedCommit = throwable == null && commit;
```
vldpyatkov, line 313: I do not understand where we will be able to see this exception, which changed the transaction state (to rollback) at the last moment.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/ReadWriteTransactionImpl.java
sanpwc: You're right, instead of 
`if (e instanceof IgniteInternalException)`
we should have
`if (e instanceof PrimaryReplicaExpiredException)`
My fault. Fixed.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
sanpwc: According to the current implementation - no. However, it seems more reliable and clear to explicitly check it. So, I'd rather left this as is.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
sanpwc: Please check
```
                    // verification future is added in order to share proper exception with the client
                    .thenCompose(r -> verificationFuture);
```
at the end of the future chaining.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/ReadWriteTransactionImpl.java
```diff
@@ -96,44 +101,60 @@ public IgniteBiTuple<ClusterNode, Long> enlist(TablePartitionId tablePartitionId
...
                     commit,
-                    groups,
+                    enlistedGroups,
                     id()
             );
         }
+
+        return finishFuture.exceptionally(
+                e -> {
+                    if (e instanceof PrimaryReplicaExpiredException) {
```
denis-chudov, line 148: This check doesn't work, because `PrimaryReplicaExpiredException` is always wrapped in `CompletionException` (thrown by`allOf` that is called in `TxManagerImpl#verifyCommitTimestamp`).

Also, I would recommend to check the error code of `TransactionException` in tests that you added - this would have indicated this problem.

denis-chudov (changes requested):

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -290,37 +297,58 @@ public CompletableFuture<Void> finish(
...
+            // given primaries are not expired or, in other words, whether commitTimestamp is less or equal to the enlisted primaries
+            // expiration timestamps.
+            CompletableFuture<Void> verificationFuture =
+                    commit ? verifyCommitTimestamp(enlistedGroups, commitTimestamp) : completedFuture(null);
+
+            return verificationFuture.handle(
+                    (unused, throwable) -> {
+                        // TODO: https://issues.apache.org/jira/browse/IGNITE-19170 Use ZonePartitionIdMessage and remove cast
+                        Collection<ReplicationGroupId> replicationGroupIds = new HashSet<>();
+                        replicationGroupIds.addAll((Collection<ReplicationGroupId>) (Collection<?>) enlistedGroups.keySet());
```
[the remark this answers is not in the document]
sanpwc: Fixed.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/ReadWriteTransactionImpl.java
sanpwc: You are right. I've removed exceptionally block and adjusted the tests.

denis-chudov (approved):

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
vldpyatkov: Why do you think it is not possible? This is a crucial part: can I trust that any (non-local) replica is primary or not in a particular moment just by using the placement driver API.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
vldpyatkov: Tricky way, I cannot insist, but I would make this code more clear.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
sanpwc: Because getPrimaryReplica(..., commitTimestamp) won't return expired lease
`if (lease.isAccepted() && lease.getExpirationTime().after(timestampWithClockSkew)) {`
thus it's not possible. So yep, it's possible to change given commitTimestamp check with an assert. However I believe that current approach is better cleaner.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
sanpwc: Well, it's a common pattern of "re-throwing" an exception from a future after handling it.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
vldpyatkov: Why does we consider a clock skew in placement driver API (and when I looked in the implementation we doing it twice)? I think we should describe this case in detail.
I always think, that the lease expiration date grater than lease start date, it is a contract, but here we are checking it....

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
vldpyatkov: I would create another future that is used to return.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -290,37 +297,56 @@ public CompletableFuture<Void> finish(
...
+
+                                        return finalTxStateMeta;
+                                    });
+
+                                    if (verifiedCommit) {
+                                        observableTimestampTracker.update(commitTimestamp);
+                                    }
+                                });
+                    })
+                    .thenCompose(Function.identity())
```
vldpyatkov, line 347: It this invokation required?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
sanpwc: I've removed the "twice" CLOCK_SKEW addition. And we use CLOCK_SKEW related logic because it's a contract of how await and get works.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
sanpwc: Again, proposed solution is a common pattern.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/ReadWriteTransactionImpl.java
```diff
@@ -96,20 +96,22 @@ public IgniteBiTuple<ClusterNode, Long> enlist(TablePartitionId tablePartitionId
     /** {@inheritDoc} */
     @Override
     protected CompletableFuture<Void> finish(boolean commit) {
-        Map<ClusterNode, List<IgniteBiTuple<TablePartitionId, Long>>> groups = new LinkedHashMap<>();
+        Map<TablePartitionId, Long> enlistedGroups = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
sanpwc: Fixed.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
sanpwc: Don't think so. Removed.

vldpyatkov (approved):

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/leases/LeaseTracker.java
```diff
@@ -222,7 +222,7 @@ public CompletableFuture<ReplicaMeta> awaitPrimaryReplica(
             long timeout,
             TimeUnit unit
     ) {
-        return inBusyLockAsync(busyLock, () -> getOrCreatePrimaryReplicaWaiter(groupId).waitFor(timestamp)
+        return inBusyLockAsync(busyLock, () -> getOrCreatePrimaryReplicaWaiter(groupId).waitFor(timestamp.addPhysicalTime(CLOCK_SKEW))
```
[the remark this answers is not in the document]
sanpwc: Removed. TODO added.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
sanpwc: My fault. It's required. Otherwise exceptions from original future won't be thrown. Lot's of tests including for example org.apache.ignite.internal.table.TxAbstractTest#testTransactionAlreadyCommitted will fail. Restored.

PR #2666, opened 2023-10-05
IGNITE-20437 Sql. Support system views in execution

korlov42:

### The Review Checklist

### Notes

### DIFF
--- a/modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManager.java
+++ b/modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManager.java
@@ -41,10 +41,9 @@ public interface SystemViewManager extends IgniteComponent {
      * Returns a list of nodes a view with given name can be found on.
      *
      * @param name Name of view of interest.
-     * @return A list of nodes owning a view. Never empty.
-     * @throws IgniteInternalException if we were unable to find nodes owning view with given name.
+     * @return A list of nodes owning a view, or empty list if there is no node in the cluster owning this view.
      */
-    List<String> owningNodes(String name) throws IgniteInternalException;
+    List<String> owningNodes(String name);
 
     /**
      * Opens a cursor over view with a given name and returns publisher emitting rows of the view.
--- a/modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManager.java
+++ b/modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManager.java
@@ -17,7 +17,11 @@
 
 package org.apache.ignite.internal.systemview;
 
+import java.util.List;
+import java.util.concurrent.Flow.Publisher;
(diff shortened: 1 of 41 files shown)


modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManager.java
```diff
@@ -32,4 +36,22 @@ public interface SystemViewManager extends IgniteComponent {
      * @param view System view to register.
      */
     void register(SystemView<?> view);
+
+    /**
+     * Returns a list of nodes a view with given name can be found on.
+     *
+     * @param name Name of view of interest.
+     * @return A list of nodes owning a view. Never empty.
```
ygerzhedovich, line 44: How we could achive it in case all nodes with the view leave a cluster? I think we just should return emty result for query which try to select fro such views.

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManagerImpl.java
```diff
@@ -136,10 +181,78 @@ public Map<String, String> nodeAttributes() {
...
+                assert value != null;
+
+                return List.copyOf(value);
+            });
+        }
+
+        this.owningNodesByViewName = Map.copyOf(owningNodesByViewName);
+    }
+
+    private static class EmptyPublisher<T> implements Publisher<T> {
```
ygerzhedovich, line 234: from my point of view it will be better to inline the publisher realisaton instead of create separate the new static class. The main reason - we can forget to remove it during amend code above to right publisher

AMashenkov (approved):

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManager.java
korlov42: as it stated on the very next line, exception will be thrown in that case

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManagerImpl.java
korlov42: > we can forget to remove it during amend code above to right publisher

I believe EmptyPublisher will be useful in any case

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManager.java
korlov42: changed this a bit: now manager returns empty list if no nodes were found

modules/system-view/src/main/java/org/apache/ignite/internal/systemview/SystemViewManagerImpl.java
korlov42: replaced EmptyPublisher with throwing exception

ygerzhedovich (approved):

PR #2671, opened 2023-10-10
IGNITE-20478 Sql. Get rid of UNSPECIFIED_VALUE_PLACEHOLDER

zstan:

### The Review Checklist

### Notes

### DIFF
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
@@ -30,7 +30,6 @@ import org.apache.ignite.internal.sql.engine.schema.TableDescriptor;
 import org.apache.ignite.internal.sql.engine.util.TypeUtils;
 import org.apache.ignite.internal.type.NativeTypeSpec;
 
-
 /**
  * Helper class provides method to convert binary tuple to rows and vice-versa.
  */
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
@@ -68,6 +68,8 @@ public final class RowConverter {
         int indexedColumnsCount = binarySchema.elementCount();
         int prefixColumnsCount = handler.columnCount(searchRow);
 
+        assert binarySchema.elementCount() >= prefixColumnsCount : "Invalid range condition";
+
         BinaryTuplePrefixBuilder tupleBuilder = new BinaryTuplePrefixBuilder(prefixColumnsCount, indexedColumnsCount);
 
         return new BinaryTuplePrefix(indexedColumnsCount, toByteBuffer(binarySchema, handler, tupleBuilder, searchRow));
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
@@ -19,7 +19,6 @@ package org.apache.ignite.internal.sql.engine.exec;
 
 import java.nio.ByteBuffer;
 import java.util.List;
-import org.apache.calcite.util.ImmutableBitSet;
 import org.apache.ignite.internal.binarytuple.BinaryTupleBuilder;
 import org.apache.ignite.internal.binarytuple.BinaryTuplePrefixBuilder;
 import org.apache.ignite.internal.schema.BinaryRowConverter;
@@ -56,30 +55,22 @@ public final class RowConverter {
      * @param factory Row handler factory.
      * @param searchRow Search row.
      * @param <RowT> Row type.
-     * @param unspecified Positions of unspecified nodes, i.e. nodes instantiated with {@code null}
      * @return Binary tuple.
      */
     static <RowT> BinaryTuplePrefix toBinaryTuplePrefix(
             ExecutionContext<RowT> ectx,
             BinaryTupleSchema binarySchema,
             RowHandler.RowFactory<RowT> factory,
-            RowT searchRow,
-            ImmutableBitSet unspecified
+            RowT searchRow
     ) {
         RowHandler<RowT> handler = factory.handler();
 
         int indexedColumnsCount = binarySchema.elementCount();
         int prefixColumnsCount = handler.columnCount(searchRow);
 
-        assert prefixColumnsCount == indexedColumnsCount : "Invalid range condition";
+        BinaryTuplePrefixBuilder tupleBuilder = new BinaryTuplePrefixBuilder(prefixColumnsCount, indexedColumnsCount);
 
-        int pos = unspecified.nextSetBit(0);
-
-        int rowColumnsCount = pos == -1 ? prefixColumnsCount : pos;
-
-        BinaryTuplePrefixBuilder tupleBuilder = new BinaryTuplePrefixBuilder(rowColumnsCount, indexedColumnsCount);
-
-        return new BinaryTuplePrefix(indexedColumnsCount, toByteBuffer(binarySchema, handler, tupleBuilder, searchRow, rowColumnsCount));
+        return new BinaryTuplePrefix(indexedColumnsCount, toByteBuffer(binarySchema, handler, tupleBuilder, searchRow));
     }
 
     /**
@@ -106,17 +97,18 @@ public final class RowConverter {
 
         BinaryTupleBuilder tupleBuilder = new BinaryTupleBuilder(rowColumnsCount);
 
-        return new BinaryTuple(rowColumnsCount, toByteBuffer(binarySchema, handler, tupleBuilder, searchRow, rowColumnsCount));
+        return new BinaryTuple(rowColumnsCount, toByteBuffer(binarySchema, handler, tupleBuilder, searchRow));
     }
 
     private static <RowT> ByteBuffer toByteBuffer(
             BinaryTupleSchema binarySchema,
             RowHandler<RowT> handler,
             BinaryTupleBuilder tupleBuilder,
-            RowT searchRow,
-            int colCount
+            RowT searchRow
     ) {
-        for (int i = 0; i < colCount; i++) {
+        int columnsCount = handler.columnCount(searchRow);
+
+        for (int i = 0; i < columnsCount; i++) {
             Object val = handler.get(i, searchRow);
 
             Element element = binarySchema.element(i);
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
@@ -73,14 +73,9 @@ public final class RowConverter {
 
         assert prefixColumnsCount == indexedColumnsCount : "Invalid range condition";
 
-        int rowColumnsCount = 0;
-        for (int i = 0; i < prefixColumnsCount; i++) {
-            if (unspecified.get(i)) {
-                break;
-            }
+        int pos = unspecified.nextSetBit(0);
 
-            rowColumnsCount++;
-        }
+        int rowColumnsCount = pos == -1 ? prefixColumnsCount : pos;
 
         BinaryTuplePrefixBuilder tupleBuilder = new BinaryTuplePrefixBuilder(rowColumnsCount, indexedColumnsCount);
 
@@ -118,7 +113,8 @@ public final class RowConverter {
             BinaryTupleSchema binarySchema,
             RowHandler<RowT> handler,
             BinaryTupleBuilder tupleBuilder,
-            RowT searchRow, int colCount
+            RowT searchRow,
+            int colCount
     ) {
         for (int i = 0; i < colCount; i++) {
             Object val = handler.get(i, searchRow);
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
@@ -17,10 +17,9 @@
 
 package org.apache.ignite.internal.sql.engine.exec;
 
-import static org.apache.ignite.internal.sql.engine.exec.exp.ExpressionFactoryImpl.UNSPECIFIED_VALUE_PLACEHOLDER;
-
 import java.nio.ByteBuffer;
 import java.util.List;
+import org.apache.calcite.util.ImmutableBitSet;
 import org.apache.ignite.internal.binarytuple.BinaryTupleBuilder;
 import org.apache.ignite.internal.binarytuple.BinaryTuplePrefixBuilder;
 import org.apache.ignite.internal.schema.BinaryRowConverter;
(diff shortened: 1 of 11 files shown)


zstan:
@xtern @korlov42 can you make a review ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
```diff
@@ -72,18 +72,18 @@ public static <RowT> BinaryTuplePrefix toBinaryTuplePrefix(
 
         assert prefixColumnsCount == indexedColumnsCount : "Invalid range condition";
 
-        int specifiedCols = 0;
+        int rowColumnsCount = 0;
```
AMashenkov, line 75:
```suggestion
        int rowColumnsCount = unspecified.nextSetBit(0);
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/RowConverter.java
```diff
@@ -108,35 +108,20 @@ public static <RowT> BinaryTuple toBinaryTuple(
...
+        return new BinaryTuple(rowColumnsCount, toByteBuffer(binarySchema, handler, tupleBuilder, searchRow, rowColumnsCount));
     }
 
     private static <RowT> ByteBuffer toByteBuffer(
-            ExecutionContext<RowT> ectx,
             BinaryTupleSchema binarySchema,
             RowHandler<RowT> handler,
             BinaryTupleBuilder tupleBuilder,
-            RowT searchRow
+            RowT searchRow, int colCount
```
AMashenkov, line 120:
```suggestion
            RowT searchRow,
            int colCount
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
```diff
@@ -93,9 +92,6 @@
 public class ExpressionFactoryImpl<RowT> implements ExpressionFactory<RowT> {
     private static final int CACHE_SIZE = 1024;
 
-    /** Placeholder for values, which expressions are not specified. */
-    public static final Object UNSPECIFIED_VALUE_PLACEHOLDER = Placeholder.UNSPECIFIED_VALUE;
-
     // We use enums for placeholders because enum serialization/deserialization guarantees to preserve object's identity.
     private enum Placeholder {
```
lowka, line 96: This enum is no longer used. I think it can be safely removed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ScannableTableImpl.java
```diff
@@ -105,8 +106,8 @@ public <RowT> Publisher<RowT> indexRangeScan(
             lower = null;
             upper = null;
         } else {
-            lower = toBinaryTuplePrefix(ctx, indexRowSchema, cond.lower(), rowFactory);
-            upper = toBinaryTuplePrefix(ctx, indexRowSchema, cond.upper(), rowFactory);
+            lower = toBinaryTuplePrefix(ctx, indexRowSchema, cond.lower(), cond.unspecifiedLower(), rowFactory);
```
zstan, line 109: i don`t like this place by myself, probably we will change it later ?)

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
```diff
@@ -487,18 +473,11 @@ private Scalar compile(List<RexNode> nodes, RelDataType type, boolean biInParams
 
         RexProgramBuilder programBuilder = new RexProgramBuilder(type, rexBuilder);
 
-        BitSet unspecifiedValues = new BitSet(nodes.size());
-
         for (int i = 0; i < nodes.size(); i++) {
             RexNode node = nodes.get(i);
 
             if (node != null) {
```
korlov42, line 479: we should not just ignore null nodes. Let's add an assertion for this.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/RangeCondition.java
```diff
@@ -29,9 +31,15 @@ public interface RangeCondition<RowT> {
     /** Upper search row. */
     RowT upper();
 
-    /** Inlusive search by lower row. */
+    /** Return positions of unspecified (i.e. null equal) nodes depends to lower bound. */
+    ImmutableBitSet unspecifiedLower();
```
korlov42, line 35: ticket description states `After reworking expandBounds the searchRow that comes to RowConverter#toBinaryTuplePrefix should already contain a prefix only.`. With such an approach, we still have nulls at the end of a searchRow.

I would suggest to rework this patch, so both `lower` and 'upper' will return only real prefixes of index keys. Here is an example: if table row is (a, b, c), we have index (c, a) and predicate `WHERE c > 0`, then `lower` should be `Row [0]` (currently it is `Row [0, null]`)

zstan:
@korlov42 thanks for clarifications, all done.

zstan:
@xtern is it line with expectations ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
```diff
@@ -765,36 +773,64 @@ private class RangeConditionImpl implements RangeCondition<RowT> {
...
+                    ? ctx.rowHandler().factory(TypeUtils.rowSchemaFromRelTypes(RelOptUtil.getFieldTypeList(rowTypeUpper)))
+                    : factory;
         }
 
         /** {@inheritDoc} */
         @Override
-        public RowT lower() {
-            return lowerRow != null ? lowerRow : (lowerRow = getRow(lowerBound));
+        public @Nullable RowT lower() {
+            return lowerRow != null ? lowerRow : getRow(lowerBound, true);
```
xtern, line 827: Is `lowerRow/upperRow` is always `null` now?
Also, instead of passing a boolean flag, might it be better to pass the target row factory?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
```diff
@@ -746,6 +735,25 @@ public T get() {
...
+        ImmutableList.Builder<RelDataType> typesBuilder = ImmutableList.builder();
+        List<RelDataType> types = RelOptUtil.getFieldTypeList(rowType);
+        int i = 0;
+        for (RexNode node : lowerInit) {
+            if (node != null) {
+                typesBuilder.add(types.get(i));
+                lowerBuilder.add(node);
+            }
+            ++i;
+        }
```
xtern, line 750:
```suggestion
        ImmutableList.Builder<RexNode> lowerBuilder = ImmutableList.builder();
        ImmutableList.Builder<RelDataType> typesBuilder = ImmutableList.builder();
        List<RelDataType> types = RelOptUtil.getFieldTypeList(rowType);
        for (int i = 0; i < bound.size(); i++) {
            RexNode node = bound.get(i);

            if (node != null) {
                typesBuilder.add(types.get(i));
                lowerBuilder.add(node);
            }
        }
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
```diff
@@ -746,6 +735,25 @@ public T get() {
         }
     }
 
+    private static Map.Entry<List<RexNode>, RelDataType> shrinkBounds(IgniteTypeFactory factory, List<RexNode> bound, RelDataType rowType) {
```
korlov42, line 738: TBH, I don't fully understand this `shrinkBounds` thing... Let's add tests for `ExpressionFactoryImpl#ranges`. That way it will be easier to verify and debug. Besides, it would be nice to describe how expand bound is supposed work, what invariants it relies on, etc

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
```diff
@@ -765,36 +795,44 @@ private class RangeConditionImpl implements RangeCondition<RowT> {
...
+            this.lowerFactory = lowerFactory;
+            this.upperFactory = upperFactory;
         }
 
         /** {@inheritDoc} */
         @Override
-        public RowT lower() {
-            return lowerRow != null ? lowerRow : (lowerRow = getRow(lowerBound));
+        public @Nullable RowT lower() {
+            return lowerRow != null ? lowerRow : getRow(lowerBound, lowerFactory);
```
xtern, line 829: As I mentioned earlier, with this change, lowerRow/upperRow is always null now.

zstan: got it, thnks !

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
```diff
@@ -487,18 +534,13 @@ private Scalar compile(List<RexNode> nodes, RelDataType type, boolean biInParams
...
             RexNode node = nodes.get(i);
 
             if (node != null) {
                 programBuilder.addProject(node, null);
             } else {
-                unspecifiedValues.set(i);
-
-                programBuilder.addProject(rexBuilder.makeNullLiteral(type == emptyType
-                        ? nullType : type.getFieldList().get(i).getType()), null);
+                assert false : "unexpected nullable node";
```
xtern, line 543: This conditional branching is redundant now. I suggest to remove it
```
assert node != null : "unexpected nullable node";

programBuilder.addProject(node, null);
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImplTest.java
```diff
@@ -176,6 +189,60 @@ public void testRangeConditionWithNullIsIgnored() {
...
+
+        // assembly bounds
+        List<SearchBounds> boundsList = List.of(
+                new RangeBounds(condition, val1, val2, true, true)
+        );
+
+        ranges = expFactory.ranges(boundsList, rowType, null);
+        assertEquals(1, ranges.iterator().next().lower().length);
+        assertEquals(1, ranges.iterator().next().upper().length);
+
```
xtern, line 243:
```suggestion
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImplTest.java
```diff
@@ -176,6 +189,60 @@ public void testRangeConditionWithNullIsIgnored() {
         assertEquals(List.of(new TestRange(new Object[]{1})), list);
     }
 
+    @ParameterizedTest(name = "condition satisfies: [{0}] the index")
```
xtern, line 192: `condition satisfies: [true] the index`
:thinking: Is this an intentional message format?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImplTest.java
```diff
@@ -176,6 +189,60 @@ public void testRangeConditionWithNullIsIgnored() {
...
+    @ValueSource(booleans = {true, false})
```
xtern, line 193: I am not fully understand the idea behind `conditionSatisfyIdx = false` test case.
AFAIK the handling of this case was not changed by this patch. But maybe you know better why this test case is necessary.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
zstan: this method is called from many places, not only from bound and there is no need to process null`s here, so i prefer to stay with this check.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImplTest.java
zstan: this is not message format and as i know we have no strong suggestions here, if you don`t like it - i can change for more readable - suggest ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
xtern: >  there is no need to process null`s here, 

Why, then, are you processing them in the conditional branch?
```
            if (node != null) {
                programBuilder.addProject(node, null);
            } else {
                assert false : "unexpected nullable node";
            }
```
> so I prefer to stay with this check

With what check - `else assert false`?

xtern: My proposal about code cleanup is to replace for-loop with the following:
```
        for (RexNode node : nodes) {
            assert node != null : "unexpected nullable node";
            
            programBuilder.addProject(node, null);
        }
```

xtern: Perhaps I wasn't clear enough.
I'm just suggesting to replace this code block
```
        for (int i = 0; i < nodes.size(); i++) {
            RexNode node = nodes.get(i);

            if (node != null) {
                programBuilder.addProject(node, null);
            } else {
                assert false : "unexpected nullable node";
            }
        }
```
with this one
```
        for (RexNode node : nodes) {
            assert node != null : "unexpected nullable node";
            
            programBuilder.addProject(node, null);
        }
```

Why you don't like it?)
What fundamental differences do you see with the current behavior?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
```diff
@@ -376,17 +388,45 @@ private void expandBounds(
             List<RexNode> curLower,
             List<RexNode> curUpper,
             boolean lowerInclude,
-            boolean upperInclude
+            boolean upperInclude,
+            boolean containNull
```
korlov42, line 392: as far as I an see, `containNull` is never `false`. Thus, why do we need this flag?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
```diff
@@ -20,18 +20,19 @@
 import static org.apache.ignite.internal.util.CollectionUtils.nullOrEmpty;
 
 import com.github.benmanes.caffeine.cache.Caffeine;
+import com.google.common.collect.ImmutableList;
```
korlov42, line 23: we should avoid using guava library

zstan: miss it ( done

zstan:
@korlov42 @xtern all done guys

korlov42 (approved):

xtern (approved):

PR #2676, opened 2023-10-11
IGNITE-19276 Implement a mechanism to build indices distributively

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-19276

### The Review Checklist

### Notes


modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
```diff
@@ -0,0 +1,327 @@
...
+            @Override
+            public void onError(Throwable e) {
+                LOG.error("Error on handle partition build index key", e);
+            }
+        });
+
+        indexBuilder.listen((indexId, tableId, partitionId) -> onIndexBuildForPartition(indexId, partitionId));
+    }
+
+    private CompletableFuture<?> onIndexCreate(CreateIndexEventParameters parameters) {
```
rpuch, line 162: Do I understand correctly that all nodes of the cluster will react on creation/removal/made-available/build-finished events, and we'll rely on MetaStorage to only allow one of the attempts to succeed?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
```diff
@@ -0,0 +1,327 @@
...
+        return inBusyLockAsync(busyLock, () -> {
+            int indexId = parameters.indexDescriptor().id();
+
+            int partitions = getPartitionCountFromCatalog(indexId, parameters.catalogVersion());
+
+            ByteArray startBuildIndexKey = startBuildIndexKey(indexId);
```
rpuch, line 168: Looks like this key is not about 'index build start', but about 'index is being built' because the index is built strictly between the key is created and removed. So I suggest to rename the key accordingly.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
```diff
@@ -0,0 +1,327 @@
...
+            return completedFuture(null);
+        });
+    }
+
+    private void onIndexBuildForPartition(int indexId, int partitionId) {
+        inBusyLock(busyLock, () -> {
+            ByteArray partitionBuildIndexKey = partitionBuildIndexKey(indexId, partitionId);
+
+            // We don't need to wait for the operation to complete.
+            metaStorageManager.invoke(exists(partitionBuildIndexKey), remove(partitionBuildIndexKey), noop());
```
rpuch, line 251: What if the invocation fails (synchronously or asynchronously, inside a future)? Will partition 'under build' status remain forever (or till a restart)?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
```diff
@@ -0,0 +1,327 @@
...
+            // to free the metastore thread.
+            catalogManager.execute(buildMakeIndexAvailableCommand(indexDescriptor));
+
+            return completedFuture(null);
+        });
+    }
+
+    private void onIndexBuildForPartition(int indexId, int partitionId) {
+        inBusyLock(busyLock, () -> {
+            ByteArray partitionBuildIndexKey = partitionBuildIndexKey(indexId, partitionId);
```
rpuch, line 248: Does a node know somehow that it finished building an index for a partition, is this persisted somewhere else? Imagine a scenario when an index gets built for partition 1, but before the key is removed from the MetaStorage the server gets restarted. Will it be able to see on recovery that the index has already been built for partition 1? I realize this should not be implemented in this PR (it's not about recovery), but I'm still curious.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
```diff
@@ -0,0 +1,327 @@
...
+                    || isMetastoreKeyAbsent(startBuildIndexKey, metastoreRevision)) {
+                return completedFuture(null);
+            }
+
+            // We can use the latest version of the catalog since we are on the metastore thread.
+            CatalogIndexDescriptor indexDescriptor = getIndexDescriptorStrict(indexId, catalogManager.latestCatalogVersion());
+
+            // We will not wait for the command to be executed, since we will then find ourselves in a dead lock since we will not be able
+            // to free the metastore thread.
+            catalogManager.execute(buildMakeIndexAvailableCommand(indexDescriptor));
```
rpuch, line 240: What if the catalog operation fails (due to some unexpected runtime exception in the MetaStorage), and this happens for all nodes processing this event, but this happens inside the future that we don't wait. Won't we lose the update?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
```diff
@@ -0,0 +1,327 @@
...
+
+            return completedFuture(null);
+        });
+    }
+
+    private void onIndexBuildForPartition(int indexId, int partitionId) {
+        inBusyLock(busyLock, () -> {
+            ByteArray partitionBuildIndexKey = partitionBuildIndexKey(indexId, partitionId);
+
+            // We don't need to wait for the operation to complete.
```
rpuch, line 250: Why shouldn't we wait?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
```diff
@@ -0,0 +1,327 @@
...
+ * @see CatalogIndexDescriptor#writeOnly()
+ */
+// TODO: IGNITE-20637 Recovery needs to be implemented
+// TODO: IGNITE-20637 Need integration with the IgniteImpl
+public class IndexAvailabilityController implements ManuallyCloseable {
+    private static final IgniteLogger LOG = Loggers.forClass(IndexAvailabilityController.class);
+
+    private static final String START_BUILD_INDEX_KEY_PREFIX = "startBuildIndex";
+
+    private static final String PARTITION_BUILD_INDEX_KEY_PREFIX = "partitionBuildIndex";
```
rpuch, line 95: It seems that if we include the period '.' in the prefix, the code will become simpler both for building and parsion the keys

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
```diff
@@ -0,0 +1,327 @@
...
+ * </ul>
+ *
+ * <p>Notes: At {@link CatalogEvent#INDEX_DROP}, the keys in the metastore are deleted: {@code startBuildIndex.<indexId>} and
+ * {@code partitionBuildIndex.<indexId>.<partitionId_0>}...{@code partitionBuildIndex.<indexId>.<partitionId_N>}.</p>
+ *
+ * @see CatalogIndexDescriptor#writeOnly()
+ */
+// TODO: IGNITE-20637 Recovery needs to be implemented
+// TODO: IGNITE-20637 Need integration with the IgniteImpl
+public class IndexAvailabilityController implements ManuallyCloseable {
```
rpuch, line 90: I think it makes sense to clarify in the documentation (this javadoc or an implementation comment below) that handling of create/drop/made-available/finished-building is made by the whole cluster (and only one node makes a write to the MetaStorage) as these events are global, but only one node (a primary replica owning a partition) handles 'index-finished-for-partition' event.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
rpuch: Also, if it fails inside a future, we'll not see an exception. `whenComplete()` might be useful to log it.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
rpuch: Also, if it fails inside a future, we'll not see an exception. `whenComplete()` might be useful to log it.

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
```diff
@@ -0,0 +1,405 @@
...
+
+    private CatalogManager catalogManager;
+
+    private IndexBuilder indexBuilder;
+
+    private IndexAvailabilityController indexAvailabilityController;
+
+    @BeforeEach
+    void setUp() {
+        vaultManager = new VaultManager(new InMemoryVaultService());
```
rpuch, line 94: Let's make final all fields that allow this. For instance, `vaultManager`, `metaStorageManager`, `catalogManager`, `indexBuilder`, `indexAvailabilityController` can be made final and initialized via field initializers, if I'm not mistaken.

When a field is final, you don't need to bother for searching whether its value is redefined somewhere, so it becomes easier to reason about the code.

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
```diff
@@ -0,0 +1,405 @@
...
+
+        for (int partitionId = 1; partitionId < partitions; partitionId++) {
+            assertPartitionBuildIndexKeyExists(indexId, partitionId);
+        }
+
+        assertTrue(indexDescriptor(INDEX_NAME).writeOnly());
+    }
+
+    @Test
+    void testMetastoreKeysAfterFinishBuildIndexForAllPartition() throws Exception {
```
rpuch, line 181:
```suggestion
    void testMetastoreKeysAfterFinishBuildIndexForAllPartitions() throws Exception {
```

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
```diff
@@ -0,0 +1,405 @@
...
+
+        assertStartBuildIndexKeyAbsent(indexId);
+
+        for (int partitionId = 0; partitionId < partitions; partitionId++) {
+            assertPartitionBuildIndexKeyAbsent(indexId, partitionId);
+        }
+    }
+
+    @Test
+    void testMetastoreKeysAfterDropIndexWithTwoRemainingPartition() throws Exception {
```
rpuch, line 269:
```suggestion
    void testMetastoreKeysAfterDropIndexWithTwoRemainingPartitions() throws Exception {
```

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
```diff
@@ -0,0 +1,405 @@
...
+        awaitActualMetastoreRevision();
+
+        assertStartBuildIndexKeyAbsent(indexId);
+
+        for (int partitionId = 0; partitionId < partitions; partitionId++) {
+            assertPartitionBuildIndexKeyAbsent(indexId, partitionId);
+        }
+    }
+
+    private void awaitActualMetastoreRevision() throws Exception {
```
rpuch, line 289: Should the name say something about applied revision? As the method seems to wait till applied revision reaches the 'global' revision.

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
```diff
@@ -0,0 +1,405 @@
...
+
+        assertStartBuildIndexKeyExists(indexId);
+
+        assertPartitionBuildIndexKeyAbsent(indexId, 0);
+
+        for (int partitionId = 1; partitionId < partitions; partitionId++) {
+            assertPartitionBuildIndexKeyExists(indexId, partitionId);
+        }
+
+        assertTrue(indexDescriptor(INDEX_NAME).writeOnly());
```
rpuch, line 177: `awaitActualMetastoreRevision()` seems to wait till the local applied revision reaches the current global MetaStorage revision. But switching the index state in the controller happens asynchronously with respect to the watch thread. Is there a guarantee that at line 177 we'll see the new index descriptor version? Is it possible that we only wait for some revision N in `awaitActualMetastoreRevision`, while the index will be updated in revision N+1?

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
```diff
@@ -0,0 +1,405 @@
...
+
+        awaitActualMetastoreRevision();
+
+        assertStartBuildIndexKeyAbsent(indexId);
+
+        for (int partitionId = 0; partitionId < partitions; partitionId++) {
+            assertPartitionBuildIndexKeyAbsent(indexId, partitionId);
+        }
+    }
+
```
rpuch, line 288: How about a test for dropping an index after it has been fully built, but has more than one partition?

modules/table/src/testFixtures/java/org/apache/ignite/internal/table/TableTestUtils.java
```diff
@@ -198,10 +198,33 @@ public static int getTableIdStrict(CatalogService catalogService, String tableNa
      * @throws AssertionError If table is absent.
      */
     public static int getIndexIdStrict(CatalogService catalogService, String indexName, long timestamp) {
-        Integer indexId = getIndexId(catalogService, indexName, timestamp);
+        return getIndexStrict(catalogService, indexName, timestamp).id();
+    }
+
+    /**
+     * Returns index descriptor form catalog, {@code null} if table is absent.
```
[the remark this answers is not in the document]
tkalkirill: fix it

modules/table/src/testFixtures/java/org/apache/ignite/internal/table/TableTestUtils.java
```diff
@@ -198,10 +198,33 @@ public static int getTableIdStrict(CatalogService catalogService, String tableNa
...
+     * @param catalogService Catalog service.
+     * @param indexName Index name.
+     * @param timestamp Timestamp.
+     */
+    public static @Nullable CatalogIndexDescriptor getIndex(CatalogService catalogService, String indexName, long timestamp) {
+        return catalogService.index(indexName, timestamp);
+    }
+
+    /**
+     * Returns index descriptor form catalog.
```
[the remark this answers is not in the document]
tkalkirill: fix it

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
```diff
@@ -0,0 +1,327 @@
...
+import org.apache.ignite.internal.metastorage.WatchListener;
+import org.apache.ignite.internal.metastorage.dsl.Operation;
+import org.apache.ignite.internal.metastorage.dsl.Operations;
+import org.apache.ignite.internal.table.distributed.index.IndexBuildCompletionListener;
+import org.apache.ignite.internal.table.distributed.index.IndexBuilder;
+import org.apache.ignite.internal.util.Cursor;
+import org.apache.ignite.internal.util.IgniteSpinBusyLock;
+
+/**
+ * Component is responsible for ensuring that the index, upon completion of the distributed index building for all partitions, becomes
```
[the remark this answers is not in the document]
tkalkirill: fix it

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
```diff
@@ -0,0 +1,327 @@
...
+
+            // We will not wait for the command to be executed, since we will then find ourselves in a dead lock since we will not be able
+            // to free the metastore thread.
+            catalogManager.execute(buildMakeIndexAvailableCommand(indexDescriptor));
+
+            return completedFuture(null);
+        });
+    }
+
+    private void onIndexBuildForPartition(int indexId, int partitionId) {
```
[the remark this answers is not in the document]
tkalkirill: It seems that we now use **index build** and not **index built** everywhere, I’ll leave it as it is.

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
tkalkirill: fix it

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
tkalkirill: fix it

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
tkalkirill: What name do you think would be suitable?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
tkalkirill: Yes.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    private boolean isRemainingPartitionBuildIndexKeys(int indexId, long metastoreRevision) {
+        try (Cursor<Entry> cursor = metaStorageManager.prefixLocally(partitionBuildIndexKeyPrefix(indexId), metastoreRevision)) {
+            return cursor.stream().anyMatch(not(Entry::tombstone));
+        }
+    }
+
+    private boolean isMetastoreKeyAbsent(ByteArray key, long metastoreRevision) {
+        return metaStorageManager.getLocally(key, metastoreRevision).tombstone();
```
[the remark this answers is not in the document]
tkalkirill: thanks. fix it

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
tkalkirill: > What if the invocation fails (synchronously or asynchronously, inside a future)? Will partition 'under build' status remain forever (or till a restart)?

If I understand you correctly, then such a problem will have to be solved and taken into account in recovery ticket https://issues.apache.org/jira/browse/IGNITE-20637

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
tkalkirill: Why not

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
tkalkirill: As far as I understand, if we made N global changes and waited for them, and in parallel did not change anything else, then we are guaranteed to wait for all the changes to be applied in the metastore.

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
tkalkirill: I didn't understand the test scenario a bit.
If the index is built, then all partition keys must be deleted.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
tkalkirill: try fix it

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
tkalkirill: I tried to rephrase.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
tkalkirill: I do not think so =)

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
tkalkirill: > What if the catalog operation fails (due to some unexpected runtime exception in the MetaStorage), and this happens for all nodes processing this event, but this happens inside the future that we don't wait. Won't we lose the update?

I think that if the operation cannot be performed for reasons other than: lack of index, already available, or the node is stopped, then this is a problem and the error handler will have to fail the node from the cluster.

> Also, if it fails inside a future, we'll not see an exception. whenComplete() might be useful to log it.

I'll add logging.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
tkalkirill: I indicated this problem in the recovery ticket [IGNITE-20637](https://issues.apache.org/jira/browse/IGNITE-20637) and we will need to carefully solve it.
At the moment, in my understanding, during recovery we will need to wait until the replication groups with the replication log applied are raised, and only after that will it be possible to look into the index storage and find out whether the index has been built, if so, then it will be possible to delete the index construction key for the partition.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
tkalkirill: How exactly to rename?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
```diff
@@ -0,0 +1,327 @@
...
+ * {@code partitionBuildIndex.<indexId>.<partitionId_0>}...{@code partitionBuildIndex.<indexId>.<partitionId_N>}.</p>
+ *
+ * @see CatalogIndexDescriptor#writeOnly()
+ */
+// TODO: IGNITE-20637 Recovery needs to be implemented
+// TODO: IGNITE-20637 Need integration with the IgniteImpl
+public class IndexAvailabilityController implements ManuallyCloseable {
+    private static final IgniteLogger LOG = Loggers.forClass(IndexAvailabilityController.class);
+
+    private static final String START_BUILD_INDEX_KEY_PREFIX = "startBuildIndex";
```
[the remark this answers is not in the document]
tkalkirill: why not

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
rpuch: Like 'indexBuild.inProgress.'

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
rpuch: While creating the full key, you will need to write `PREFIX + indexId` instead of `PREFIX + '.' + indexId`. This is shorter and simpler.

While parsing the full key, you will be able to omit the `indexOf('.', ...)` operation, so the code, again, becomes simpler.

And while registering watches on the prefix, the prefix with dot on the end will work the same.

2 of 3 operations will be made a bit simpler, for one of them nothing will change. There is a clear profit.

What are the benefits of the original code? Please elaborate.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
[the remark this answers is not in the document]
rpuch: I'm not an expert in English, but 'On index build' seems to mean that the index has started being built, or that it's being built right now. But this method is about the 'finish' event. So the method has to either have 'finish' (or its analogue) in its name, or it should mention the 'build' in the past ('built'), otherwise it confuses a reader.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
rpuch: Then I suggest adding a TODO about Failure Handler here (or will it catch the exception in another place?) as we rely on Failure Handler in many places, it will be really easy to forget some of them :)

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
rpuch: `awaitTillGlobalMetastoreRevisionIsApplied()`

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
rpuch: But doesn't the code under test makes a parallel change?

rpuch (commented):
Great job!

modules/index/src/test/java/org/apache/ignite/internal/index/IndexAvailabilityControllerTest.java
tkalkirill: Nope.

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
tkalkirill: Why not

modules/index/src/main/java/org/apache/ignite/internal/index/IndexAvailabilityController.java
tkalkirill: You're right, I'll fix it.

rpuch (approved):

PR #2684, opened 2023-10-12
IGNITE-20522 Create the default user 'Ignite'

Flaugh24:
https://issues.apache.org/jira/browse/IGNITE-20522

### The Review Checklist

### Notes


modules/cli/src/integrationTest/java/org/apache/ignite/internal/rest/ItGeneratedRestClientTest.java
```diff
@@ -286,12 +286,26 @@ void updateTheSameNodeConfiguration() {
     void updateNodeConfigurationWithInvalidParam() throws JsonProcessingException {
         ApiException thrown = assertThrows(
                 ApiException.class,
-                () -> clusterConfigurationApi.updateClusterConfiguration("security.authentication.enabled=true")
+                () -> clusterConfigurationApi.updateClusterConfiguration("{\n"
```
ibessonov, line 289: Why so many quotes? It's HOCON, you don't need them

modules/runner/src/test/java/org/apache/ignite/internal/configuration/ClusterConfigurationDefaultsSetterTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+        Config config = ConfigFactory.parseString(defaults);
+
+        List<? extends Config> providers = config.getConfigList("security.authentication.providers");
+
+        assertThat(providers.size(), equalTo(1));
+
+        Config defaultProvider = providers.get(0);
+
+        assertAll(
```
ibessonov, line 92: Can it be 4 different lines? What's the reason for having the "assertAll", looks inappropriate.
There are other instances of the same pattern as well, I wonder why

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterInitializer.java
```diff
@@ -50,11 +55,20 @@ public class ClusterInitializer {
 
     private final ClusterService clusterService;
 
+    private final ConfigurationDefaultsSetter configurationDefaultsSetter;
+
+    private final ConfigurationValidator clusterConfigurationValidator;
     private final CmgMessagesFactory msgFactory = new CmgMessagesFactory();
```
ibessonov, line 61: Formatting is broken.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterInitializer.java
```diff
@@ -50,11 +55,20 @@ public class ClusterInitializer {
...
 
     /** Constructor. */
-    public ClusterInitializer(ClusterService clusterService) {
+    public ClusterInitializer(
+            ClusterService clusterService,
+            ConfigurationDefaultsSetter configurationDefaultsSetter,
+            ConfigurationValidator clusterConfigurationValidator
```
ibessonov, line 67: I don't see reasons to validate configuration here one more time. An explanation in the comment somewhere would be helpful. Maybe we should remove it

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterInitializer.java
```diff
@@ -122,44 +136,57 @@ public CompletableFuture<Void> initCluster(
...
-                            );
-
-                            return CompletableFuture.<Void>completedFuture(null);
-                        } else {
-                            if (e instanceof CompletionException) {
-                                e = e.getCause();
-                            }
+            String initialClusterConfiguration = clusterConfiguration == null
+                    ? ""
+                    : clusterConfiguration;
```
ibessonov, line 141: I find it strange that you must convert null to empty string, just to call a private method, that does the same null-check. Something's wrong here, please simplify your code.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterInitializer.java
```diff
@@ -235,4 +262,20 @@ private static List<ClusterNode> resolveNodes(ClusterService clusterService, Col
                 })
                 .collect(Collectors.toList());
     }
+
+
+    private CompletableFuture<String> setDefaultsToClusterConfiguration(@Nullable String hocon) {
```
ibessonov, line 267: Does it really have to be a future? I don't think that there are any asynchronous operations inside of the setter.

modules/configuration-api/src/main/java/org/apache/ignite/configuration/ConfigurationDefaultsSetter.java
```diff
@@ -0,0 +1,34 @@
...
+ * in the configuration itself.
+ */
+public interface ConfigurationDefaultsSetter {
+    /*
+     * Sets the default values for the cluster configuration. This method is called when the cluster is started for the first time.
+     *
+     * @param hocon HOCON representation of the cluster configuration.
+     * @return Future that is completed when the default values are set.
+     */
+    CompletableFuture<String> setDefaults(String hocon);
```
ibessonov, line 33: I would prefer another name. `set` usually means that we mutate the state of the `this` instance, using argument as a source of changes. Here it's the opposite, we update the argument using `this` as a source of changes.
It's also hard to understand what I said from the documentation, which is confusing. Please rename the method and write better documentation.

modules/rest/src/main/java/org/apache/ignite/internal/rest/cluster/ClusterManagementController.java
```diff
@@ -84,27 +74,15 @@ public CompletableFuture<Void> init(@Body InitCommand initCommand) {
...
-    private CompletableFuture<Void> validateConfiguration(@Nullable String configuration) {
-        if (configuration != null) {
-            List<ValidationIssue> validationIssues = clusterCfgValidator.validateHocon(configuration);
-            if (!validationIssues.isEmpty()) {
-                return CompletableFuture.failedFuture(new ConfigurationValidationException(validationIssues));
-            }
-        }
-
-        return CompletableFuture.completedFuture(null);
+        //return validateConfiguration(initCommand.clusterConfiguration())
```
ibessonov, line 77: Please remove this line

modules/rest/src/main/java/org/apache/ignite/internal/rest/cluster/ClusterManagementController.java
```diff
@@ -84,27 +74,15 @@ public CompletableFuture<Void> init(@Body InitCommand initCommand) {
...
+        return clusterInitializer.initCluster(
+                initCommand.metaStorageNodes(),
+                initCommand.cmgNodes(),
+                initCommand.clusterName(),
+                initCommand.clusterConfiguration()
+        ).exceptionally(ex -> {
+            throw mapException(ex);
```
ibessonov, line 84: A note for the future: I don't think that we have an adequate mapping for configuration exceptions. Are there any tests for that?

modules/rest/src/main/java/org/apache/ignite/internal/rest/cluster/ClusterManagementController.java
```diff
@@ -133,7 +111,7 @@ private static RuntimeException mapException(Throwable ex) {
         }
     }
 
-    private static Throwable unwrapExceptionCompletionException(Throwable ex) {
+    private static Throwable unwrapException(Throwable ex) {
```
ibessonov, line 114: The same method exists in IgniteUtils I believe

modules/runner/src/integrationTest/java/org/apache/ignite/internal/rest/ItNotInitializedClusterRestTest.java
```diff
@@ -210,7 +210,40 @@ void nodeVersion() throws IOException, InterruptedException {
     }
 
     @Test
-    @DisplayName("Cluster is not initialized, if config is invalid")
+    @DisplayName("Cluster is not initialized, if config has a syntax error")
+    void initClusterWithInvalidHoconConfig() throws IOException, InterruptedException {
+        // When POST /management/v1/cluster/init with invalid config
+        String requestBody = "{\n"
+                + "    \"metaStorageNodes\": [\n"
```
ibessonov, line 217: Same comment about quotes. Let's make these tests more readable

modules/runner/src/integrationTest/java/org/apache/ignite/internal/rest/ItNotInitializedClusterRestTest.java
```diff
@@ -210,7 +210,40 @@ void nodeVersion() throws IOException, InterruptedException {
     }
 
     @Test
-    @DisplayName("Cluster is not initialized, if config is invalid")
+    @DisplayName("Cluster is not initialized, if config has a syntax error")
+    void initClusterWithInvalidHoconConfig() throws IOException, InterruptedException {
+        // When POST /management/v1/cluster/init with invalid config
```
ibessonov, line 215:
```suggestion
        // When POST /management/v1/cluster/init with invalid config.
```

ibessonov: According to code style, we put a dot at the end.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/rest/authentication/ItAuthenticationTest.java
```diff
@@ -107,7 +107,38 @@ public void disabledAuthentication(TestInfo testInfo) {
     }
 
     @Test
-    public void changeCredentials(TestInfo testInfo) throws InterruptedException {
+    public void defaultUser(TestInfo testInfo) throws InterruptedException, IOException {
+        // when
```
ibessonov, line 111: Here you use absolutely different notation. Should be `// When.`, right? The goal is to have uniformed approach

modules/runner/src/main/java/org/apache/ignite/internal/configuration/ClusterConfigurationDefaultsSetter.java
```diff
@@ -0,0 +1,146 @@
...
+                rootKeys,
+                new RuntimeConfigurationStorage(ConfigurationType.DISTRIBUTED),
+                generator,
+                configurationValidator
+        );
+    }
+
+    /**
+     * {@inheritDoc}
+     */
```
ibessonov, line 78: Please remove it, we don't use these anymore

modules/runner/src/main/java/org/apache/ignite/internal/configuration/ClusterConfigurationDefaultsSetter.java
```diff
@@ -0,0 +1,146 @@
...
+     * Constructor.
+     *
+     * @param rootKeys Root keys.
+     * @param generator Configuration tree generator.
+     */
+    public ClusterConfigurationDefaultsSetter(
+            Collection<RootKey<?, ?>> rootKeys,
+            ConfigurationTreeGenerator generator
+    ) {
+        configurationValidator = new ConfigurationValidatorImpl(generator, Set.of());
```
ibessonov, line 67: Another validation? Why?
Defaults must be covered with tests, there's no point in validating them on every step. One time, before application, is enough

modules/runner/src/main/java/org/apache/ignite/internal/configuration/ClusterConfigurationDefaultsSetter.java
```diff
@@ -0,0 +1,146 @@
...
+        configurationManager = new ConfigurationManager(
```
ibessonov, line 68: Configuration registry is the only reason to create manager right?

ibessonov: What was the problem with using existing configuration manager?

modules/runner/src/main/java/org/apache/ignite/internal/configuration/ClusterConfigurationDefaultsSetter.java
```diff
@@ -0,0 +1,146 @@
...
+                );
+    }
+
+    private CompletableFuture<Void> validateCfg(String hocon) {
+        try {
+            List<ValidationIssue> validationIssues = configurationValidator.validateHocon(hocon);
+            if (validationIssues.isEmpty()) {
+                return CompletableFuture.completedFuture(null);
+            } else {
+                return failedFuture(new ConfigurationValidationException(validationIssues));
```
ibessonov, line 104: Any specific reason for using futures here?

modules/runner/src/main/java/org/apache/ignite/internal/configuration/ClusterConfigurationDefaultsSetter.java
```diff
@@ -0,0 +1,146 @@
...
+                .includeInternal(true)
+                .maskSecretValues(false)
+                .build();
+
+        ConfigurationRegistry configurationRegistry = configurationManager.configurationRegistry();
+
+        return validateCfg(hocon)
+                .thenCompose(v -> parseHocon(hocon))
+                .thenCompose(
+                        v -> setSecurityDefaults(configurationRegistry.getConfiguration(SecurityConfiguration.KEY))
```
ibessonov, line 91: Instead of a hardcode, we may have asked corresponding configuration modules to patch the tree.
Why have you decided to put these defaults into a modules that doesn't own them?

modules/runner/src/main/java/org/apache/ignite/internal/configuration/ClusterConfigurationDefaultsSetter.java
```diff
@@ -0,0 +1,146 @@
...
+            ConfigurationRegistry configurationRegistry = configurationManager.configurationRegistry();
+
+            return configurationRegistry.change(hoconSource);
+        } catch (Exception e) {
+            return failedFuture(e);
+        }
+    }
+
+    private static CompletableFuture<Void> setSecurityDefaults(SecurityConfiguration securityCfg) {
+        if (securityCfg.authentication().providers().value().size() == 0) {
```
ibessonov, line 124: Instead of updating `*Configuration` instance, you could update `*View` instance. It's much simpler and doesn't require the storage.
Let's re-write this class a bit, I don't like current approach

modules/runner/src/main/java/org/apache/ignite/internal/configuration/storage/RuntimeConfigurationStorage.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicLong;
+import org.apache.ignite.configuration.annotation.ConfigurationType;
+
+/**
+ * Test configuration storage.
+ */
+public class RuntimeConfigurationStorage implements ConfigurationStorage {
```
ibessonov, line 35: This class should not exist. Please delete it.

ibessonov (changes requested):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/rest/ItNotInitializedClusterRestTest.java
ibessonov: Why do we force users to use json in REST API? Is this the same in CLI?

modules/security/src/test/java/org/apache/ignite/internal/security/authentication/SecurityConfigurationModuleTest.java
```diff
@@ -24,24 +24,57 @@
...
+
+        SuperRoot superRoot = generator.createSuperRoot();
+
+        rootChange = new SuperRootChangeImpl(superRoot);
+    }
+
+    @AfterEach
+    void tearDown() {
+        generator.close();
+        generator = null;
```
ibessonov, line 67: This is not necessary, but I'm OK with it

ibessonov (approved):

modules/configuration/src/test/java/org/apache/ignite/internal/configuration/ConfigurationDefaultsPatcherImplTest.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @AfterAll
+    static void afterAll() {
+        generator.close();
+    }
+
+    @Test
+    void patchedHoconDoesNotContainNullValues() {
+        String patchedHocon = patcher.patchWithDefaults("");
```
PakhomovAlexander, line 54: I expect an empty hocon being patched to... empty hocon. That is absolutely implicit that you pass an empty string and get some hocon configuration. The name of the method `patchWithDefaults` and javadocs says nothing about it.

PR #2689, opened 2023-10-14
IGNITE-20116 Splitting operationTimestamp and safeTimeTimestamp

sanpwc:
https://issues.apache.org/jira/browse/IGNITE-20116

- operationTimestamp and safeTimestamp are splitted.
- linearization for commands processing added.


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1776,8 +1788,19 @@ private CompletableFuture<ReplicaResult> processMultiEntryAction(ReadWriteMultiR
...
                     }
 
-                    return validateAtTimestampAndBuildUpdateAllCommand(request, rowIdsToDelete, lastCommitTimes, txCoordinatorId)
-                            .thenCompose(cmd -> applyUpdateAllCommand(cmd, request.skipDelayedAck()))
+                    return validateOperationAgainstSchemaAtTimestamp(request.transactionId())
+                            .thenCompose(
+                                    catalogVersion -> applyUpdateAllCommand(
+                                            rowIdsToDelete,
+                                            lastCommitTimes,
+                                            request.commitPartitionId(),
```
rpuch, line 1796: Would it make sense to create an overload of `applyUpdateAllCommand` that would accept a `request` and take out commitPartitionId, transactionId, full and skipDelayedAck from it? As this seems to always repeat. This would reduce duplication and also make it easier to handle new message fields in the future (they seem to be added again and again).

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2038,133 +2094,195 @@ private CompletableFuture<Object> applyCmdWithExceptionHandling(Command cmd) {
...
-        } else {
-            return applyCmdWithExceptionHandling(cmd).thenApply(res -> {
-                // This check guaranties the result will never be lost. Currently always null.
-                assert res == null : "Replication result is lost";
+                    // Set context for delayed response.
+                    return cmd.txId();
+                });
 
-                // Try to avoid double write if an entry is already replicated.
+                // TODO: https://issues.apache.org/jira/browse/IGNITE-20124 tmp
```
rpuch, line 2139: `tmp`?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2038,133 +2094,195 @@ private CompletableFuture<Object> applyCmdWithExceptionHandling(Command cmd) {
...
 
-                return null;
-            });
+                return completedFuture(fut);
+            } else {
+                return applyCmdWithExceptionHandling(cmd).thenApply(res -> {
+                    // This check guaranties the result will never be lost. Currently always null.
+                    assert res == null : "Replication result is lost";
+
+                    // TODO: https://issues.apache.org/jira/browse/IGNITE-20124 tmp
```
rpuch, line 2162: tmp?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2213,14 +2331,19 @@ private CompletableFuture<ReplicaResult> processSingleEntryAction(ReadWriteSingl
...
-                                        validatedRowId.uuid(),
-                                        null,
-                                        lastCommitTime,
-                                        txCoordinatorId
-                                )
-                                        .thenCompose(this::applyUpdateCommand)
+                                return validateOperationAgainstSchemaAtTimestamp(request.transactionId())
+                                        .thenCompose(
+                                                catalogVersion -> applyUpdateCommand(
+                                                        request.commitPartitionId().asTablePartitionId(),
```
rpuch, line 2337: Same suggestion about an overload

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2860,96 +3046,19 @@ private CompletableFuture<Void> validateAtTimestamp(UUID txId) {
...
-                lastCommitTimestamp,
-                request.transactionId(),
-                request.full(),
-                txCoordinatorId
-        );
-    }
-
-    /**
-     * Chooses operation timestamp, makes validations that require it and constructs an {@link UpdateCommand} object.
+     * Chooses operation timestamp and makes schema related validations.
```
rpuch, line 3049:
```suggestion
     * Chooses operation timestamp and makes schema related validations. The operation timestamp is only used for validation, it is NOT sent as safeTime timestamp.
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2860,96 +3046,19 @@ private CompletableFuture<Void> validateAtTimestamp(UUID txId) {
...
-    private CompletableFuture<UpdateCommand> validateAtTimestampAndBuildUpdateCommand(
-            TablePartitionId tablePartId,
-            UUID rowUuid,
-            @Nullable BinaryRow row,
-            @Nullable HybridTimestamp lastCommitTimestamp,
-            UUID txId,
-            boolean full,
-            String txCoordinatorId
-    ) {
+    private CompletableFuture<Integer> validateOperationAgainstSchemaAtTimestamp(UUID txId) {
```
rpuch, line 3054: It seems that now it makes no sense to have `AtTimestamp` in the name as we don't pass any timestamp as a parameter. How about `validateOperationAgainstSchemaNow()` (or just `validateOperationAgainstSchema()`)?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: tmp here means that code below is temporary and will be adjusted within IGNITE-20124. Will rewrite the comment.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: tmp here means that code below is temporary and will be adjusted within IGNITE-20124. Will rewrite the comment.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: I've added an overloaded version with request and removed initial one because occurred that there are no usages any longer.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: Fixed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: Agree. I've renamed it to `validateOperationAgainstSchema()` as you've suggested.

rpuch (approved):

PR #2700, opened 2023-10-17
IGNITE-20659 Placement driver do not create a lease

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-20659

### DIFF
--- a/modules/placement-driver/src/test/java/org/apache/ignite/internal/placementdriver/LeaseUpdaterTest.java
+++ b/modules/placement-driver/src/test/java/org/apache/ignite/internal/placementdriver/LeaseUpdaterTest.java
@@ -1,3 +1,20 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


modules/placement-driver/src/test/java/org/apache/ignite/internal/placementdriver/LeaseUpdaterTest.java
```diff
@@ -0,0 +1,274 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+/**
+ * The class contains unit tests for {@link LeaseUpdater}.
+ */
+@ExtendWith({MockitoExtension.class})
+public class LeaseUpdaterTest extends BaseIgniteAbstractTest {
```
denis-chudov, line 83: These tests still pass without the fix. Maybe they become flaky, but with very low failure rate. I tried to run them multiple times without changes from `LeaseUpdater` and got 1 failure after ~600 runs. There is no test that is specific for this fix, let's try to add it.

vldpyatkov: Unfortunately, I cannot bring anything except the test, which increases the likelihood of the race. The issue has never been reproduced on my computer; even for debugging, I asked a colleague.

modules/placement-driver/src/test/java/org/apache/ignite/internal/placementdriver/LeaseUpdaterTest.java
```diff
@@ -0,0 +1,274 @@
...
+            t.join();
+        }
+
+        if (leaseUpdater.active()) {
+            assertTrue(IgniteTestUtils.waitForCondition(() -> getUpdaterThread() != null, 10_000));
+
+            awaitForLease();
+
+            leaseUpdater.deactivate();
+        }
```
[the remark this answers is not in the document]
vldpyatkov: Fixed.

denis-chudov (approved):

PR #2711, opened 2023-10-17
IGNITE-20359 Expose storage profiles as a node attribute

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-20359

### The Review Checklist

### Notes


modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/configuration/StorageProfileConfigurationSchema.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.apache.ignite.configuration.annotation.Config;
+import org.apache.ignite.configuration.annotation.InjectedName;
+import org.apache.ignite.configuration.annotation.Value;
+
+/**
+ * TODO: add java doc https://issues.apache.org/jira/browse/IGNITE-20564.
+ */
+@Config
+public class StorageProfileConfigurationSchema {
```
kgusakov, line 28: I would prefer to use the current *StorageEngineConfigurationSchema roots + *DataRegionConfigurationSchema pair instead of introducing this the the next one configurations. Because my current issue 20564 is on pause, we can live with this meaningless stubs in configuration some period of time :(

kgusakov (changes requested):

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/configuration/StorageProfileConfigurationSchema.java
alievmirza: As far as I understand, we will create separate schema for storage profiles. Why do we need then do double work using `StorageEngineConfigurationSchema`  and then implementing again separate schema?

alievmirza: I hope that your current issue 20564  won't be on stop a long period of time, so I don't see a big problem in this stub being in the code base, taking into account that you will reuse it soon

kgusakov: Ok, I see

kgusakov (approved):

PR #2722, opened 2023-10-19
IGNITE-20599 Implement 'NOT_TOMBSTONE' operation in the meta storage dsl.

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-20599

### The Review Checklist

### Notes


sk0x50 (approved):

modules/metastorage-api/src/main/java/org/apache/ignite/internal/metastorage/dsl/Conditions.java
```diff
@@ -256,6 +256,19 @@ public static SimpleCondition tombstone(ByteArray key) {
                 .build();
     }
 
+    /**
+     * Produces the condition of type {@link ConditionType#NOT_TOMBSTONE}. This condition tests that an entry's value, identified by the
+     * given key, is not tombstone.
```
sashapolo, line 261:
```suggestion
     * given key, is not a tombstone.
```

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/TombstoneCondition.java
```diff
@@ -23,18 +23,51 @@
  * Condition tests an entry's value is tombstone in meta storage. Entry is tombstone if it is not empty and doesn't exists.
  */
 public class TombstoneCondition extends AbstractSimpleCondition {
+    /** Condition type. */
+    private final TombstoneCondition.Type type;
```
sashapolo, line 27: Why do we need this enum, can we have a separate class, like `NotTombstoneCondition`? If this is needed to make it similar to `ExistenceCondition`, then I would propose to make a separate class for it as well

alievmirza: Yes, this was done to follow the general pattern that is implemented in `ExistenceCondition`, `RevisionCondition` or `ValueCondition`. For your proposed change it would be reasonable to separate all types from `org.apache.ignite.internal.metastorage.server.RevisionCondition.Type` and `ValueCondition` as well, but I don't see the justification for such a big refactoring, so I would stick with the proposed solution.

sashapolo: Regarding `RevisionCondition` and `ValueCondition`, having a separate enum there is somewhat justified, because its has so many implementations which share the same set of fields. In case of `TombstoneCondition` it looks like an overcomplication, becuase it only has two types

sashapolo: I'm going to approve the PR, because it's a kind of a matter of taste, even if I don't like it.

sashapolo (approved):

PR #2726, opened 2023-10-19
IGNITE-20304 Sql. Documentation for system views module.

xtern:


modules/system-view-api/README.md
```diff
@@ -0,0 +1,127 @@
...
+* `SystemViewProvider` denotes a component that provides system views.
+* `SystemViewManager` is responsible for views registration in the cluster.
+
+## Basic concepts
+
+* Component that want to expose system views must extend `SystemViewProvider` interface.
+* All system views must be registered during node startup.
+* All system views are reside in `SYSTEM` schema.
+* From SQL perspective, a system view is a virtual immutable table.
+* System views doesn't support transactions.
```
ygerzhedovich, line 28: let's clarify what does it mean.

modules/system-view-api/README.md
```diff
@@ -0,0 +1,127 @@
...
+| isvt_n_0         | isvt_n_2  | 192.168.0.103 | 3346 |
+| isvt_n_2         | isvt_n_0  | 192.168.0.103 | 3344 |
+| isvt_n_2         | isvt_n_1  | 192.168.0.103 | 3345 |
+| isvt_n_2         | isvt_n_2  | 192.168.0.103 | 3346 |
+| isvt_n_1         | isvt_n_0  | 192.168.0.103 | 3344 |
+| isvt_n_1         | isvt_n_1  | 192.168.0.103 | 3345 |
+| isvt_n_1         | isvt_n_2  | 192.168.0.103 | 3346 |
+
+Since the local cluster includes 3 nodes, the data is duplicated three times.
+Because for a local view, data is retrieved on **each node** where the view is registered.
```
ygerzhedovich, line 127: Should we add additional sections with some rocomndations and best approaches with working with the system views?
Just my thoughts:
1) Use concrete class in generic instead of `?`, like a SystemViews.<ParentIdAwareDescriptor<CatalogTableColumnDescriptor>>clusterViewBuilder()
2) Use upper case for name of view and their columns + may be add information about the no ability to use case sensitive mode
3) where we usaly register new views
......

modules/system-view/README.md
```diff
@@ -1,3 +1,37 @@
...
+   * Executes the `CreateSystemViewCommand` catalog command. This command adds non-existent system views into the catalog.
+   * Collects a list of names of registered system views as a system node attribute.
+3. Raft group service completes joining a node to the cluster by sending a `ClusterNodeMessage`. 
+   with system attributes containing the names of locally residing system views.
+4. Each node updates logical nodes in logical topology snapshot with received system attributes.  
+   This way, any node in the cluster knows which nodes have which system views.
+
+### Execution
+
+From the query engine perspective the system view is an unmodifiable table that doesn't support transactions.
```
ygerzhedovich, line 32: let's explain the meaning of "does not support transactions"

modules/system-view/README.md
ygerzhedovich, line 1: not sure, but maybe need to say where we keep information about existing views, I mean node attributes

modules/system-view-api/README.md
xtern: I did my best, but I'm bad at explaining things :sweat_smile:

ygerzhedovich: Thanks, now it looks much better and clearer

modules/system-view-api/README.md
xtern: > Use upper case for name of view

Added clarification about case insensitivity names in the "Basic concepts" section.

> Use concrete class in generic instead of ?

To be honest, I was unable to specify the provider and data provider without specifying a specific class.
The only thing you can do is not specify generic (that is, keep Object), but in this case, firstly, a warning is highlighted and secondly, you will not be able to extract data for the column.

> where we usaly register new views

I'm not sure I understand this :thinking:

ygerzhedovich: ok, thanks for the explanation, you understand me correct.

modules/system-view/README.md
xtern: Node attributes are used to understand on which node which system views are available - this information is present in the document.

```
4. Each node updates logical nodes in logical topology snapshot with received system attributes.  
   This way, any node in the cluster knows which nodes have which system views.
```

modules/system-view/README.md
xtern: Explanation added.

modules/system-view-api/README.md
xtern: Added note about that
```
* All system views must be registered during node startup. The `IgniteImpl` constructor is currently used as a starting point to do this.
```
:thinking: :thinking:

ygerzhedovich (approved):

modules/system-view-api/README.md
korlov42: it worth to mention though that it safer to explicitly specify type of the column:

```
SystemViews.clusterViewBuilder()
                .<Integer>addColumn("VIEW_ID", NativeTypes.INT32, entry -> entry.id)
```

in the example above, there is no way the type of `entry.id` will be changed silently

xtern: The example has been updated with column type parameterization and a note about this has been added.

korlov42 (approved):

PR #2728, opened 2023-10-19
IGNITE-20618 Sql. Degradation of SELECT operations performance over time

korlov42:

### The Review Checklist

### Notes


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ScannableTableImpl.java
```diff
@@ -148,24 +147,31 @@ public <RowT> Publisher<RowT> indexLookup(
...
+
             pub = internalTable.lookup(
                     partWithTerm.partId(),
-                    txAttributes.time(),
+                    readTime,
                     ctx.localNode(),
                     indexId,
                     keyTuple,
-                    requiredColumns
+                    null
```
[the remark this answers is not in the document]
korlov42: > do we need them ?

I don't know 🤷‍♂️ It's part of internal table API which is not implemented. From sql point of view it's safer to pass 'null' explicitly, otherwise we will have problems if someone will decide to implement missing functionality

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/ArrayRowHandler.java
```diff
@@ -108,17 +129,137 @@ public Object[] create(Object... fields) {
...
+            builder.appendNull();
+
+            return;
+        }
+
+        NativeType nativeType = NativeTypes.fromObject(value);
+
+        assert nativeType != null;
+
+        value = TypeUtils.fromInternal(value, NativeTypeSpec.toClass(nativeType.spec(), true));
```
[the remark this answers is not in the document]
korlov42: I'm wondering which tests do you expect to fail?

Currently, ArrayRowHandler is used only in tests which primary goal is to verify behaviour of particular execution nodes. These tests are usually using primitive rows with integers or strings in it, so conversion is not required indeed.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/rel/ExchangeExecutionTest.java
```diff
@@ -495,8 +497,13 @@ private RewindableAsyncRoot<Object[], Object[]> createRootFragment(
         ExchangeService exchangeService = exchangeServices.computeIfAbsent(localNode.name(), name ->
                 createExchangeService(taskExecutor, serviceFactory.forNode(localNode.name()), mailboxRegistry));
 
+        RowSchema schema = RowSchema.builder()
```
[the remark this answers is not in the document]
korlov42: I would say, TestDataProvider depends on that schema. Thus, I moved schema definition to a static field and left a comment that all data providers created within this test class must conform to that schema

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/BinaryTupleBuilder.java
```diff
@@ -82,11 +82,24 @@ public BinaryTupleBuilder(int numElements) {
...
     public BinaryTupleBuilder(int numElements, int totalValueSize) {
+        this(numElements, totalValueSize, true);
+    }
+
+    /**
+     * Creates a builder.
+     *
+     * @param numElements Number of tuple elements.
+     * @param totalValueSize Total estimated length of non-NULL values, -1 if not known.
+     * @param exactEstimate Whether the total size is exact estimate or approximate. The
```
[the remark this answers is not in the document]
korlov42: the idea of this flag is to introduce two different modes to BinaryTupleBuilder: the one where we want to specify maximum possible size of the value and another where we can only specify lower bound. I'll appreciate any suggestion on how to do it better

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/SqlRowHandler.java
```diff
@@ -226,24 +219,61 @@ void set(int field, @Nullable Object value) {
...
+                        estimatedSize += ((ByteString) value).length();
+                    } else {
+                        assert (value instanceof BigDecimal) || (value instanceof BigInteger) : "unexpected value " + value.getClass();
+
+                        exactEstimate = false;
+                    }
+                }
+            }
+
+            BinaryTupleBuilder tupleBuilder = new BinaryTupleBuilder(row.length, estimatedSize, exactEstimate);
```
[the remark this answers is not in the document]
korlov42: whole point of this patch is to avoid using this `new BinaryTupleBuilder(row.length)` version of constructor. The problem here is without any estimate it preallocates buffer with size ~4kb (see `BinaryTupleBuilder#DEFAULT_BUFFER_SIZE`).

If we will follow your suggestion, we fix issue for all types but string, decimal, and number. I'm not sure if anyone is indexing decimals, but I definitely sure string is a popular choice for PK. Thus, I would sacrifice a little code clarity in favour of performance

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/BinaryTupleBuilder.java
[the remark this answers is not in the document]
xtern: Sorry, I missed this. I thought `BinaryTupleBuilder(1, 1, false)` is equal to `BinaryTupleBuilder(1, -1)`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/SqlRowHandler.java
```diff
@@ -55,12 +55,11 @@
...
- * using the {@link #toByteBuffer(RowWrapper) toByteBuffer} method.
+ * using the {@link #toBinaryTuple(RowWrapper)} method.
  *
- * <p>Factory methods {@link RowFactory#create(InternalTuple) wrap(InternalTuple)} and
- * {@link RowFactory#create(ByteBuffer) create(ByteBuffer)} allow create rows without
- * any additional conversions. But the fields in binary tuple must match the
- * factory {@link RowSchema row schema}.
+ * <p>Factory method {@link RowFactory#create(InternalTuple) wrap(InternalTuple)}
+ * allow create rows without any additional conversions. But the fields in binary
+ * tuple must match the factory {@link RowSchema row schema}.
```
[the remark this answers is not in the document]
korlov42: done

zstan (approved):

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
```diff
@@ -73,6 +97,102 @@ public static BinaryTuplePrefix fromBinaryTuple(BinaryTuple tuple) {
...
+                    dataBeginOffset = begin;
+                }
+
+                if (index < size) {
+                    dataEndOffset = end;
+                }
+            }
+        };
+
+        tuple.parse(stats);
```
sashapolo, line 177: This doesn't look optimal to me. Why do we have to iterate over the whole tuple just to get the start and end of the data range? I think we can use a different API (this is pseudocode):
```
tuple.fetch(0, (index, begin, end) -> dataBeginOffset = begin);
tuple.fetch(size - 1, (index, begin, end) -> dataBeginOffset = end);
```

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
```diff
@@ -73,6 +97,102 @@ public static BinaryTuplePrefix fromBinaryTuple(BinaryTuple tuple) {
...
+
+        BinaryTuplePrefixBuilder builder = new BinaryTuplePrefixBuilder(size, size, stats.dataEndOffset - stats.dataBeginOffset);
+
+        tuple.parse((index, begin, end) -> {
```
sashapolo, line 181: Same here:
```
    byte[] valueBytes = tuple.bytesValue(i);
    
    builder.appendBytes(valueBytes);
}
```

sashapolo: BTW, maybe we could introduce a more effective API to the BinaryTuple, because in this approach we do a lot of array copying

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
sashapolo: Or maybe a better approach is to introduce a new API to the BinaryTuple that will allow to get a "sub-tuple" by providing two indexes

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
```diff
@@ -51,13 +52,36 @@ public BinaryTuplePrefix(int elementCount, ByteBuffer buffer) {
...
+     * excess columns will be truncated.
+     *
+     * @param size The size of the complete tuple.
+     * @param tuple Tuple to create a prefix from.
+     * @return Prefix, created from provided tuple with regards to desired size.
+     */
+    public static BinaryTuplePrefix fromBinaryTuple(int size, BinaryTuple tuple) {
+        if (size == tuple.elementCount()) {
+            return entireTuple(tuple);
+        } else if (size > tuple.elementCount()) {
```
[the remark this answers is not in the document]
korlov42: I'm not sure what exactly you are asking...

as utility method, it must cover all possible inputs. In that particular case, size of the given BinaryTuple may be greater, lesser, or equal to desired number of elements in prefix. NB: I refer to `number of elements` as to size of the full schema of search bound, not the actual number of elements in prefix

as user of this method, I want to have convenient utility method to convert given BinaryTuple which contains only relevant fields to a proper prefix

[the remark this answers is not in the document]
sashapolo: My question is, why would a _prefix_ of a tuple have more elements than a tuple?

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/BinaryTupleBuilder.java
```diff
@@ -82,11 +82,24 @@ public BinaryTupleBuilder(int numElements) {
...
+     *      difference here is with exact estimate allocation will be optimal, while with
+     *      approximate estimate some excess allocation is possible.
+     */
+    public BinaryTupleBuilder(int numElements, int totalValueSize, boolean exactEstimate) {
```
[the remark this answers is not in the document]
korlov42: Please read [these](https://github.com/apache/ignite-3/pull/2728#discussion_r1369686009) [two](https://github.com/apache/ignite-3/pull/2728#discussion_r1369698209) replies. If the reason of adding this flag still remains not clear, please don't hesitate to ask follow up questions

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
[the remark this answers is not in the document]
korlov42: because, although they have pretty similar name, it's a complete different object. Prefix is a part of index api, it represents partially filled search bound of a particular index. BinaryTuple, on the other hand, is a general purpose container.

So we have general purpose container that we want to reuse as search bound, but we got only relevant fields in it. In that case, size of BinaryTuple will be less than full size of index-related BinaryTuplePrefix

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
```diff
@@ -51,13 +52,36 @@ public BinaryTuplePrefix(int elementCount, ByteBuffer buffer) {
         super(elementCount, buffer);
     }
 
+    /**
+     * Creates a prefix from provided {@link BinaryTuple}. If given tuple has lesser or equal number of
+     * columns, then all elements will be used in resulting prefix. If given tuple has more columns, then
+     * excess columns will be truncated.
+     *
+     * @param size The size of the complete tuple.
```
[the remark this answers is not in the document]
korlov42: renamed `size` to `numElements`.  But renaming of entire method doesn't seem correct since it not only truncates given tuple, but may expand it as well

modules/schema/src/test/java/org/apache/ignite/internal/schema/BinaryTuplePrefixTest.java
```diff
@@ -76,6 +78,101 @@ public void testInternalBufferReallocation() {
         assertThat(prefix.intValue(0), is(Integer.MAX_VALUE));
     }
 
+    @Test
+    public void testCreatePrefixFromBinaryTuple1() {
```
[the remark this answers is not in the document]
korlov42: I've changed names of introduced tests. Hope new names better reflects purpose of tests

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/BinaryTupleBuilder.java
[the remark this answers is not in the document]
sashapolo: Ok, I missed the same thing)

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
korlov42: great advice with `fetch`, thanks!

about new API: I agree that this looks expensive to create all these byte arrays, but I doubt new api will improve performance in any way. All created objects have very short lifetime and never escape the thread, thus, I believe, they will be created in TLAB, which is relatively cheap

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
korlov42: answered [here](https://github.com/apache/ignite-3/pull/2728#discussion_r1370086037)

sashapolo: You've answered my comment about the new API, but kept the same API, why?

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
sashapolo: About new API: I meant this not only in terms of performance but also in terms of readability. Using `fetch` to find the boundaries of a subarray looks like a battle against the current API. A more straightforward API can be both faster and easier to use

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
korlov42: sorry, I thought it all about array copying...

if it's about iterating value by value, then it's because we can just copy prefix of header as we do in `expandTuple` method. Due to truncation, the size of element in offset map may be adjusted as well, I don't want do reimplement that logic once again.

A new API (like `BinTuple.project(int... fields):ByteBuffer`) will definitely help to make this case a lot cleaner, but the truth is, `truncateTuple` is used only in tests for now. I'm not sure if anyone will use it future, thus don't pay too much attention to this method. I believe we can revise it once it will start cause any problem

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
korlov42: > A more straightforward API can be both faster and easier to use

I agree with this, but not sure how such API may look like. Primary cases I interested in are `entireTuple` and `expandTuple`.  I think, former is already good enough. But how to handle `expandTuple`? We basically need to have an ability to append arbitrary amount of null elements to the existing tuple. Beside, we have to reserve additional 4 bytes to append actual number of elements

modules/schema/src/test/java/org/apache/ignite/internal/schema/BinaryTuplePrefixTest.java
```diff
@@ -76,6 +78,81 @@ public void testInternalBufferReallocation() {
         assertThat(prefix.intValue(0), is(Integer.MAX_VALUE));
     }
 
+    @Test
+    public void testCreatePrefixFromBinaryTupleWhichSizeIsLesserThanRequired() {
```
sashapolo, line 82:
```suggestion
    public void testCreatePrefixFromBinaryTupleWhichSizeIsLessThanRequired() {
```

modules/schema/src/test/java/org/apache/ignite/internal/schema/BinaryTuplePrefixTest.java
```diff
@@ -76,6 +78,81 @@ public void testInternalBufferReallocation() {
...
+
+        assertThat(prefix.elementCount(), equalTo(sourceTupleSize));
+        assertThat(prefix.intValue(0), equalTo(10));
+        assertThat(prefix.hasNullValue(1), equalTo(true));
+        assertThat(prefix.hasNullValue(2), equalTo(true));
+        assertThat(prefix.hasNullValue(3), equalTo(true));
+    }
+
+    @Test
+    public void testCreatePrefixFromBinaryTupleWhichSizeIsEqualsToRequired() {
```
sashapolo, line 99:
```suggestion
    public void testCreatePrefixFromBinaryTupleWhichSizeIsEqualToRequired() {
```

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
sashapolo: Yeah, looking at `expandTuple` we need the offset values, this sucks. Are you sure that `expandTuple` will be used anywhere?

modules/schema/src/test/java/org/apache/ignite/internal/schema/BinaryTuplePrefixTest.java
korlov42: fixed, thanks!

modules/schema/src/test/java/org/apache/ignite/internal/schema/BinaryTuplePrefixTest.java
korlov42: fixed, thanks!

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
```diff
@@ -73,6 +97,88 @@ public static BinaryTuplePrefix fromBinaryTuple(BinaryTuple tuple) {
...
+
+        byte flags = tupleBuffer.get(0);
+        int entrySize = BinaryTupleCommon.flagsToEntrySize(flags);
+
+        ByteBuffer prefixBuffer = ByteBuffer.allocate(
+                        tupleBuffer.remaining()
+                                + (entrySize * (numElements - tuple.elementCount()))
+                                + Integer.BYTES)
+                .order(ORDER)
+                .put(tupleBuffer.slice().limit(dataBeginOffsetHolder[0])); // header
```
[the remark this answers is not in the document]
korlov42: replaced

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
```diff
@@ -73,6 +97,88 @@ public static BinaryTuplePrefix fromBinaryTuple(BinaryTuple tuple) {
...
+        tuple.fetch(0, (index, begin, end) -> dataBeginOffsetHolder[0] = begin);
+        tuple.fetch(tuple.elementCount() - 1, (index, begin, end) -> dataEndOffsetHolder[0] = end);
+
+        ByteBuffer tupleBuffer = tuple.byteBuffer();
+
+        byte flags = tupleBuffer.get(0);
+        int entrySize = BinaryTupleCommon.flagsToEntrySize(flags);
+
+        ByteBuffer prefixBuffer = ByteBuffer.allocate(
+                        tupleBuffer.remaining()
```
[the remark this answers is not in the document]
korlov42: done

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
korlov42: make sense, replaced

modules/schema/src/main/java/org/apache/ignite/internal/schema/BinaryTuplePrefix.java
korlov42: it's used in this very patch. We have BinaryTuple that represents search row. This tuple contains only meaningful values, thus its number of elements may be less than number of indexed fields. 

If you think this method is not very useful as general purpose utility for tuple prefix, then I will move it to sql domain

sashapolo: ok, makes sense, let's leave it as-is

sashapolo (approved):

PR #2751, opened 2023-10-25
IGNITE-20700 Implement durable transaction coordinator finish

Cyrill:

https://issues.apache.org/jira/browse/IGNITE-20700

### The Review Checklist

### Notes


modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxManager.java
```diff
@@ -116,22 +115,27 @@ public interface TxManager extends IgniteComponent {
...
     );
 
+
+    /**
+     * Make sure the state of the provided transaction is updated correctly.
+     *
+     * @param commit {@code true} if a commit requested.
+     * @param txId Transaction id.
+     */
+    CompletableFuture<Void> finish(boolean commit, UUID txId);
```
vldpyatkov, line 137: It is not an overloded finish method. It would be better to rename it (_finishEmpty_, for example).

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -295,139 +312,244 @@ public void finishFull(HybridTimestampTracker timestampTracker, UUID txId, boole
...
+                        if (ex.getCause() != null && ex.getCause() instanceof TransactionException) {
+                            TransactionException cause = (TransactionException) ex.getCause();
+
+                            if (cause.code() == TX_WAS_ABORTED_ERR) {
+                                updateTxMeta(txId, old -> new TxStateMeta(ABORTED, old.txCoordinatorId(), null));
+
+                                return CompletableFuture.<Void>failedFuture(cause);
+                            }
+                        }
+                        LOG.warn("Failed to finish Tx {}. The operation will be retried.", ex, txId);
```
vldpyatkov, line 473: Do not forget about the log message rules.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1437,6 +1450,47 @@ private CompletableFuture<Void> finishAndCleanup(
...
+        // Check locksReleased flag. If it is already set, do nothing and return a successful result.
+        // Even if the outcome is different (the transaction was aborted, but we want to commit it),
+        // we return 'success' to be in alignment with common transaction handling.
+        if (transactionAlreadyFinished) {
+            if (txMeta.locksReleased()) {
+                return completedFuture(null);
+            }
+            // If the locks were not released, we are likely to be in a recovery mode and retrying the finish request.
+            // In this case we want to check the expected outcome and the actual one.
+            if (commit && txMeta.txState() == ABORTED) {
```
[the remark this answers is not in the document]
vldpyatkov: Do you mean we will send a cleanup request with the commit flag == false, although the transaction state is committed?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -295,139 +319,252 @@ public void finishFull(HybridTimestampTracker timestampTracker, UUID txId, boole
...
+        AtomicBoolean performingFinish = new AtomicBoolean();
         TxContext tuple = txCtxMap.compute(txId, (uuid, tuple0) -> {
             if (tuple0 == null) {
                 tuple0 = new TxContext(); // No writes enlisted.
             }
 
-            if (tuple0.finishFut == null) {
-                tuple0.finishFut = new CompletableFuture<>();
-                ref.set(tuple0.finishFut);
+            if (!tuple0.isTxFinishing()) {
```
vldpyatkov, line 365: Currently, it is impossible to double-finish one transaction. I would replace the `if` clause with `assert !tuple0.isTxFinishing()`.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -295,139 +319,252 @@ public void finishFull(HybridTimestampTracker timestampTracker, UUID txId, boole
...
+        HybridTimestamp commitTimestamp = getCommitTimestamp(commit);
+        // In case of commit it's required to check whether current primaries are still the same that were enlisted and whether
+        // given primaries are not expired or, in other words, whether commitTimestamp is less or equal to the enlisted primaries
+        // expiration timestamps.
+        CompletableFuture<Void> verificationFuture =
+                commit ? verifyCommitTimestamp(enlistedGroups, commitTimestamp) : completedFuture(null);
+
+        return verificationFuture.handle(
+                        (unused, throwable) -> {
+                            boolean verifiedCommit = throwable == null && commit;
```
vldpyatkov, line 426: Here we may get an axception close to that you declared above: TX_WAS_ABORTED_ERR.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -295,139 +319,252 @@ public void finishFull(HybridTimestampTracker timestampTracker, UUID txId, boole
...
 
             return tuple0;
         });
 
-        if (ref.get() != null) { // This is a finishing thread.
-            if (!commit) {
-                clo.apply(null).handle((ignored, err) -> {
+        // This is a finishing thread.
+        if (performingFinish.get()) {
+            Function<Void, CompletableFuture<Void>> finishAction = ignored ->
```
denis-chudov, line 376: What purpose does this function serve? It is just passed to the method only once, and called inside only once. Does it just pick some local context for `runFinish` to reduce the number of parameters?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1437,6 +1450,49 @@ private CompletableFuture<Void> finishAndCleanup(
...
+
+        // Check locksReleased flag. If it is already set, do nothing and return a successful result.
+        // Even if the outcome is different (the transaction was aborted, but we want to commit it),
+        // we return 'success' to be in alignment with common transaction handling.
+        if (transactionAlreadyFinished) {
+            if (txMeta.locksReleased()) {
+                return completedFuture(null);
+            }
+
+            assert !(txMeta.txState() == COMMITED && !commit) : "Not allowed to abort an already committed transaction.";
```
denis-chudov, line 1466:
```suggestion
            assert commit || txMeta.txState() == ABORTED : "Not allowed to abort an already committed transaction [txState=" + txMeta.txState() + "].";
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1437,6 +1450,49 @@ private CompletableFuture<Void> finishAndCleanup(
...
+            // If the locks were not released, we are likely to be in a recovery mode and retrying the finish request.
+            // In this case we want to check the expected outcome and the actual one.
+            if (commit && txMeta.txState() == ABORTED) {
+                LOG.error("Failed to commit a transaction that is already aborted [txId={}].", txId);
+
+                throw new TransactionException(TX_WAS_ABORTED_ERR,
```
denis-chudov, line 1472: Why assertion error in case of `!commit && txState == COMMITED`
but transaction exception in case of `commit && txState == ABORTED`?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -295,139 +319,252 @@ public void finishFull(HybridTimestampTracker timestampTracker, UUID txId, boole
...
+                    );
+
+            runFinish(commit, tuple, finishAction);
+        }
+
+        // The method `runFinish` has a side effect on `finishInProgressFuture` future, it kicks off another future that will complete it.
+        return tuple.finishInProgressFuture;
+    }
+
+    private static void runFinish(boolean commit, TxContext tuple, Function<Void, CompletableFuture<Void>> finishAction) {
```
denis-chudov, line 393: I like the idea of splitting the spaghetti code to the smaller methods, but their names should make sense in order to allow guessing which logic they contain. Now in this PR we have 5 "finish" methods:
- finish
- finishEmpty
- finishFull
- finishOnPrimary
- runFinish

and I am already lost...

Same with `makeFinishRequest` (for which making the request is not even its main job) and `makeDurableFinishRequest`.

I suggest to 
- remove `finishEmpty`, `runFinish` and `makeFinishRequest`;
- `finishOnPrimary` should be renamed to `makeFinishRequest`;
- move the logic related to computing of `performingFinish` to a method inside of TxContext;
- rename `makeDurableFinishRequest` to `durableFinish`

WDYT?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -295,139 +319,252 @@ public void finishFull(HybridTimestampTracker timestampTracker, UUID txId, boole
...
-                }
+    private CompletableFuture<Void> makeFinishRequest(
+            HybridTimestampTracker observableTimestampTracker,
+            TablePartitionId commitPartition,
+            boolean commit,
+            Map<TablePartitionId, Long> enlistedGroups,
+            UUID txId,
+            CompletableFuture<TransactionMeta> txFinishFuture
+    ) {
+        HybridTimestamp commitTimestamp = getCommitTimestamp(commit);
```
denis-chudov, line 417: I would suggest to leave the generation of commit timestamp in `finish` method to emphasize that it must be generated only once. It's not obvious in this method and can be lost during some refactoring. But I also suggest to refactor it, as I said in another comment.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -295,139 +319,252 @@ public void finishFull(HybridTimestampTracker timestampTracker, UUID txId, boole
...
+                        ))
+                .handle((res, ex) -> {
+                    if (ex != null) {
+                        Throwable cause = ExceptionUtils.unwrapCause(ex);
+
+                        if (cause instanceof TransactionException) {
+                            TransactionException transactionException = (TransactionException) cause;
+
+                            if (transactionException.code() == TX_WAS_ABORTED_ERR) {
+                                updateTxMeta(txId, old -> new TxStateMeta(ABORTED, old.txCoordinatorId(), null));
```
denis-chudov, line 477: There should be a completion of future on `TxStateMetaFinishing`, you pass it here as `txFinishFuture`. I think it requires some dedicated method as `finalStateTransition(txId, commit)` as it happens in multiple places now.

denis-chudov (changes requested):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/NodeUtils.java
```diff
@@ -0,0 +1,142 @@
...
+import org.apache.ignite.internal.raft.service.RaftGroupService;
+import org.apache.ignite.internal.replicator.TablePartitionId;
+import org.apache.ignite.internal.testframework.IgniteTestUtils;
+import org.apache.ignite.table.Tuple;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A helper class to manipulate Ignite nodes in tests.
+ */
+public class NodeUtils {
```
sanpwc, line 43: Class name seems to be confusing. In any case, because we know that transferPrimary will be added as @TestOnly one to PlacementDriver, I'd rather add current static helper method to the TestPlacementDriver. Minor though, because anyway it'll be refactored soon.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -572,7 +573,8 @@ public class IgniteImpl implements Ignite {
                 new TransactionIdGenerator(() -> clusterSvc.nodeName().hashCode()),
                 () -> clusterSvc.topologyService().localMember().id(),
                 placementDriverMgr.placementDriver(),
-                partitionIdleSafeTimePropagationPeriodMsSupplier
+                partitionIdleSafeTimePropagationPeriodMsSupplier,
+                new SimpleFailHandler()
```
sanpwc, line 577: We don't even know who will actually handle unrecoverable finish exceptions: tx coordinator, commit partition or someone else? We(at least I) also don't know how it'll be propagated to classes, e.g. it might be some static helper class. For know it's better just to log the exception without pretending that we will somehow handle it fail handler. In any case it should be covered within separate ticket.

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestSingleNode.java
```diff
@@ -238,6 +238,7 @@ protected boolean assertPartitionsSame(TableImpl table, int partId) {
     protected void injectFailureOnNextOperation(TableImpl accounts) {
         InternalTable internalTable = accounts.internalTable();
         ReplicaService replicaService = IgniteTestUtils.getFieldValue(internalTable, "replicaSvc");
+        Mockito.doReturn(CompletableFuture.failedFuture(new Exception())).when(replicaService).invoke((String) any(), any());
         Mockito.doReturn(CompletableFuture.failedFuture(new Exception())).when(replicaService).invoke((ClusterNode) any(), any());
```
sanpwc, line 242: Do we still have ClusterNode parametrized call?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -233,6 +234,8 @@ public class PartitionReplicaListener implements ReplicaListener {
 
     private final Supplier<Map<Integer, IndexLocker>> indexesLockers;
 
+    private final ConcurrentMap<UUID, CompletableFuture<Void>> txDurableFinishFutures = new ConcurrentHashMap<>();
```
sanpwc, line 237: Please add an explanation why we need given futures. And by the way (I didn't check it yet) why we really need them? 
Finish operation is idempotent, we will retry it rather rare.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1437,6 +1450,49 @@ private CompletableFuture<Void> finishAndCleanup(
             UUID txId,
             String txCoordinatorId
     ) {
+        TxMeta txMeta = txStateStorage.get(txId);
+
+        // Check that a transaction has already been finished.
```
sanpwc, line 1455: Minor. "Check that" -> "Check whether"

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: Original option seems more clear. And I'd also add an explanation why it's not valid to have abort request over already committed transaction, while it's fine to have e.g. commit over aborted.

sanpwc (changes requested):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: Because it's not possible from our logics point of view to send rollback over finished transaction.

- First of all txCoordinator calls use same tx state over retries, both abort and commit are possible.
- Server side recovery (which is not implemented yet) may only change tx state to aborted.
- TxCoordinator itself should prevent user calls with different proposed state to the one, that was already triggered. Here I'm talking about client -> txCoordinator.commitAsync(); txCoordinator.rollbackAsync()

All in all, that means that with commit it's valid to see:

- null (if it's the first change state attempt)
- commit (if it was already updated by the prior iteration)
- abort (if it was rolled back by initiate recover logic, this is very unlikely case because initiate recover will only roll back the tx if coordinator is dead)

And within roll back it's possible to see: 

- null,
- abort

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1437,6 +1450,47 @@ private CompletableFuture<Void> finishAndCleanup(
...
+            if (commit && txMeta.txState() == ABORTED) {
+                LOG.error("Failed to commit a transaction that is already aborted [txId={}].", txId);
+
+                throw new TransactionException(TX_WAS_ABORTED_ERR,
+                        "Failed to change the outcome of a finished transaction"
+                                + " [txId=" + txId + ", txState=" + txMeta.txState() + "].");
+            }
+            // The transaction has already been finished, but the locks are not released.
+            // Waiting for the cleanup to do this.
+            // TODO: There is a risk that nobody is cleaning up this transaction.
```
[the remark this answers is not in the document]
sanpwc: Let's discuss the question. Generally speaking it should not be part of tx coordinator recovery but a commit partition one that should be already implemented. What I mean here is that, when tx state is updated to either committed or aborted tx coordinator should believe that his work is done and commit partition itself should retry cleanup until locks release.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1437,6 +1450,49 @@ private CompletableFuture<Void> finishAndCleanup(
...
+            // If the locks were not released, we are likely to be in a recovery mode and retrying the finish request.
+            // In this case we want to check the expected outcome and the actual one.
+            if (commit && txMeta.txState() == ABORTED) {
+                LOG.error("Failed to commit a transaction that is already aborted [txId={}].", txId);
+
+                throw new TransactionException(TX_WAS_ABORTED_ERR,
+                        "Failed to change the outcome of a finished transaction"
+                                + " [txId=" + txId + ", txState=" + txMeta.txState() + "].");
+            }
+            // The transaction has already been finished, but the locks are not released.
```
sanpwc, line 1476: As far as I remember, you've told that some additional FINISHING handling logic is required. Which one?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxManager.java
```diff
@@ -116,22 +115,27 @@ public interface TxManager extends IgniteComponent {
...
     );
 
+
+    /**
+     * Make sure the state of the provided transaction is updated correctly.
+     *
+     * @param commit {@code true} if a commit requested.
+     * @param txId Transaction id.
+     */
+    CompletableFuture<Void> finishEmpty(boolean commit, UUID txId);
```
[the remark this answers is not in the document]
sanpwc: Agree.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/FailHandler.java
```diff
@@ -0,0 +1,31 @@
...
+ * A transaction fail handler.
+ */
+public interface FailHandler {
+    /**
+     * Check if the provided exception is recoverable.
+     *
+     * @param throwable Exception to test.
+     * @return {@code true} if recoverable, {@code false} otherwise.
+     */
+    boolean checkRecoverable(Throwable throwable);
```
sanpwc, line 30: Just remove the FailHandler for now. And move isRecoverable to where it's currently needed. BTW static?

denis-chudov (approved):

sanpwc (approved):

PR #2761, opened 2023-10-27
IGNITE-20715 Check that versions of tuples sent to PartitionReplicaListener match tx-bound schema version

rpuch:
https://issues.apache.org/jira/browse/IGNITE-20715

### The Review Checklist

### Notes

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
@@ -19,9 +19,11 @@ package org.apache.ignite.internal.table;
 
 import static java.util.concurrent.CompletableFuture.completedFuture;
 import static java.util.concurrent.CompletableFuture.failedFuture;
+import static java.util.function.Function.identity;
 import static org.apache.ignite.internal.lang.IgniteExceptionMapperUtil.convertToPublicFuture;
 import static org.apache.ignite.internal.util.ExceptionUtils.sneakyThrow;
 
+import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.function.BiFunction;
@@ -99,24 +101,36 @@ abstract class AbstractTableView {
      * @return Whatever the action returns.
      */
     protected final <T> CompletableFuture<T> withSchemaSync(@Nullable Transaction tx, KvAction<T> action) {
+        return withSchemaSync(tx, null, action);
+    }
+
+    private <T> CompletableFuture<T> withSchemaSync(@Nullable Transaction tx, @Nullable Integer previousSchemaVersion, KvAction<T> action) {
         CompletableFuture<Integer> schemaVersionFuture = tx == null
                 ? schemaVersions.schemaVersionAtNow(tbl.tableId())
                 : schemaVersions.schemaVersionAt(((InternalTransaction) tx).startTimestamp(), tbl.tableId());
 
-        CompletableFuture<T> future = schemaVersionFuture.thenCompose(action::act)
-                .handle((BiFunction<T, Throwable, CompletableFuture<T>>) (res, ex) -> {
-                    if (ex != null && isOrCausedBy(InternalSchemaVersionMismatchException.class, ex)) {
-                        // Repeat.
-                        return withSchemaSync(tx, action);
-                    }
+        CompletableFuture<T> future = schemaVersionFuture
+                .thenCompose(schemaVersion -> {
+                    return action.act(schemaVersion)
+                            .handle((BiFunction<T, Throwable, CompletableFuture<T>>) (res, ex) -> {
+                                if (ex != null && isOrCausedBy(InternalSchemaVersionMismatchException.class, ex)) {
+                                    assert tx == null : "Only for implicit transactions a retry might be requested";
+                                    assert previousSchemaVersion == null || !Objects.equals(schemaVersion, previousSchemaVersion)
+                                            : "Same schema version (" + schemaVersion
+                                                    + ") on a retry: something is wrong, is this caused by the test setup?";
 
-                    if (ex != null) {
-                        return failedFuture(ex);
-                    }
+                                    // Repeat.
+                                    return withSchemaSync(tx, schemaVersion, action);
+                                }
 
-                    return completedFuture(res);
+                                if (ex != null) {
+                                    return failedFuture(ex);
+                                }
+
+                                return completedFuture(res);
+                            });
                 })
-                .thenCompose(f -> f);
+                .thenCompose(identity());
 
         return convertToPublicFuture(future);
     }
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
@@ -132,15 +132,7 @@ abstract class AbstractTableView {
     }
 
     private static boolean isOrCausedBy(Class<? extends Exception> exceptionClass, @Nullable Throwable ex) {
-        if (ex == null) {
-            return false;
-        }
-
-        if (exceptionClass.isInstance(ex)) {
-            return true;
-        }
-
-        return isOrCausedBy(exceptionClass, ex.getCause());
+        return ex != null && (exceptionClass.isInstance(ex) || isOrCausedBy(exceptionClass, ex.getCause()));
     }
 
     /**
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
@@ -18,7 +18,6 @@
 package org.apache.ignite.internal.table;
 
 import static java.util.concurrent.CompletableFuture.completedFuture;
-import static java.util.concurrent.CompletableFuture.failedFuture;
 import static java.util.function.Function.identity;
 import static org.apache.ignite.internal.lang.IgniteExceptionMapperUtil.convertToPublicFuture;
 import static org.apache.ignite.internal.util.ExceptionUtils.sneakyThrow;
@@ -26,7 +25,6 @@ import static org.apache.ignite.internal.util.ExceptionUtils.sneakyThrow;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
-import java.util.function.BiFunction;
 import org.apache.ignite.internal.lang.IgniteExceptionMapperUtil;
 import org.apache.ignite.internal.schema.SchemaRegistry;
 import org.apache.ignite.internal.table.distributed.replicator.InternalSchemaVersionMismatchException;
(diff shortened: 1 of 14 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
```diff
@@ -95,17 +101,52 @@ protected final <T> T sync(CompletableFuture<T> fut) {
...
+        return withSchemaSync(tx, null, action);
+    }
 
+    private <T> CompletableFuture<T> withSchemaSync(@Nullable Transaction tx, @Nullable Integer previousSchemaVersion, KvAction<T> action) {
         CompletableFuture<Integer> schemaVersionFuture = tx == null
                 ? schemaVersions.schemaVersionAtNow(tbl.tableId())
                 : schemaVersions.schemaVersionAt(((InternalTransaction) tx).startTimestamp(), tbl.tableId());
 
-        CompletableFuture<T> future = schemaVersionFuture.thenCompose(action::act);
+        CompletableFuture<T> future = schemaVersionFuture
```
sashapolo, line 112: May I propose the following code re-arrangement to make it shorter?

```
CompletableFuture<T> future = schemaVersionFuture
        .thenCompose(schemaVersion -> action.act(schemaVersion)
                .handle((res, ex) -> {
                    if (isOrCausedBy(InternalSchemaVersionMismatchException.class, ex)) {
                        assert tx == null : "Only for implicit transactions a retry might be requested";
                        assert previousSchemaVersion == null || !Objects.equals(schemaVersion, previousSchemaVersion)
                                : "Same schema version (" + schemaVersion
                                        + ") on a retry: something is wrong, is this caused by the test setup?";

                        // Repeat.
                        return withSchemaSync(tx, schemaVersion, action);
                    } else {
                        return ex == null ? completedFuture(res) : CompletableFuture.<T>failedFuture(ex);
                    }
                }))
        .thenCompose(identity());
```

modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
```diff
@@ -95,17 +101,52 @@ protected final <T> T sync(CompletableFuture<T> fut) {
...
+                                return completedFuture(res);
+                            });
+                })
+                .thenCompose(identity());
 
         return convertToPublicFuture(future);
     }
 
+    private static boolean isOrCausedBy(Class<? extends Exception> exceptionClass, @Nullable Throwable ex) {
+        if (ex == null) {
```
sashapolo, line 139: Can we replace this with `return ex != null && (exceptionClass.isInstance(ex) || isOrCausedBy(exceptionClass, ex.getCause()));` ?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -472,34 +472,30 @@ private CompletableFuture<?> processRequest(ReplicaRequest request, @Nullable Bo
...
     }
 
     /**
-     * Makes sure that we have schemas corresponding to the moment of tx start; this makes PK extraction safe WRT
-     * {@link org.apache.ignite.internal.schema.SchemaRegistry#schema(int)}.
+     * Validates that the table exists at a timestamp corresponding to the request operation.
      *
-     * @param request Request that's being processed.
+     * <ul>
+     *     <li>For an RW read/write, it's 'now'</li>
```
sashapolo, line 487:
```suggestion
     *     <li>For a read/write, it's 'now'</li>
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -472,34 +472,30 @@ private CompletableFuture<?> processRequest(ReplicaRequest request, @Nullable Bo
...
+     *     <li>For an RO read (with readTimestamp), it's readTimestamp (matches readTimestamp in the transaction)</li>
```
sashapolo, line 488:
```suggestion
     *     <li>For a read (with readTimestamp), it's readTimestamp (matches readTimestamp in the transaction)</li>
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -472,34 +472,30 @@ private CompletableFuture<?> processRequest(ReplicaRequest request, @Nullable Bo
...
+     *     <li>For an RO direct read, it's the timestamp chosen (as 'now') to process the request</li>
```
sashapolo, line 489:
```suggestion
     *     <li>For a direct read, it's the timestamp chosen (as 'now') to process the request</li>
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -520,18 +518,84 @@ private CompletableFuture<HybridTimestamp> validateTableExistence(ReplicaRequest
...
+     * @param request Replica request corresponding to the operation.
+     * @param opTsIfDirectRo Operation timestamp for a direct RO, {@code null} otherwise.
+     * @return Future completed when the validation is finished.
+     */
+    private CompletableFuture<Void> validateSchemaMatch(ReplicaRequest request, @Nullable HybridTimestamp opTsIfDirectRo) {
+        if (!(request instanceof SchemaVersionAwareReplicaRequest)) {
+            return completedFuture(null);
+        }
+
+        SchemaVersionAwareReplicaRequest versionAwareRequest = (SchemaVersionAwareReplicaRequest) request;
```
sashapolo, line 537: Looks like this variable is used at the very end, better move it there

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -520,18 +518,84 @@ private CompletableFuture<HybridTimestamp> validateTableExistence(ReplicaRequest
...
+
+        HybridTimestamp tsToWaitForSchema = getTxStartTimestamp(request);
```
sashapolo, line 539: It may be a good idea to pass `opTsIfDirectRo` to `getTxStartTimestamp` so that it will return it instead of `null`, this will make the code shorter

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/SchemaCompatValidator.java
```diff
@@ -221,4 +221,14 @@ void failIfTableDoesNotExistAt(HybridTimestamp operationTimestamp, int tableId)
             throw tableWasDroppedException(tableId);
         }
     }
+
+    void failIfRequestSchemaDiffersFromTxTs(HybridTimestamp txTs, int requestSchemaVersion, int tableId) {
```
sashapolo, line 225: Missing javadoc

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/replication/PartitionReplicaListenerTest.java
```diff
@@ -2256,6 +2258,129 @@ private void testCommitRequestIfTableWasDropped(
...
     }
 
+    @CartesianTest
+    @CartesianTest.MethodFactory("singleRowRwOperationTypesFactory")
+    void singleRowRwOperationsFailIfSchemaVersionMismatchesTx(RequestType requestType, boolean onExistingRow, boolean full) {
+        RwListenerInvocation invocation = null;
+
+        if (RequestTypes.isSingleRowRwPkOnly(requestType)) {
+            invocation = (targetTxId, key) -> {
+                return doSingleRowPkRequest(targetTxId, marshalKeyOrKeyValue(requestType, key), requestType, full);
```
sashapolo, line 2268: you can use expression lambda here and in similar places

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
rpuch: There are writes, RW reads and RO reads. This is about RW reads and writes, but if we change it the way you suggest it will be about any of 3

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
rpuch: It seems to be important to highlight that it's an RO read, not just any read

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
rpuch: Again, there are RW reads and RO reads which further split into RO reads with a timestamp and direct RO reads. Here, 'RO' seems to be important.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
rpuch: Well, `getTxStartTimestamp()` returns the timestamp corresponding to the start of a transaction. I'm not sure that we should stretch the direct RO operation timestamp to think about it as a transaction start ts.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
rpuch: I rephrased a bit

sashapolo: My suggestion may be wrong, but "RW read/write" makes even less sense

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sashapolo: Please specify that it is an "RO transaction", like you did in other places

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sashapolo: as you wish

modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
```diff
@@ -95,17 +101,40 @@ protected final <T> T sync(CompletableFuture<T> fut) {
...
+    private <T> CompletableFuture<T> withSchemaSync(@Nullable Transaction tx, @Nullable Integer previousSchemaVersion, KvAction<T> action) {
         CompletableFuture<Integer> schemaVersionFuture = tx == null
                 ? schemaVersions.schemaVersionAtNow(tbl.tableId())
                 : schemaVersions.schemaVersionAt(((InternalTransaction) tx).startTimestamp(), tbl.tableId());
 
-        CompletableFuture<T> future = schemaVersionFuture.thenCompose(action::act);
+        CompletableFuture<T> future = schemaVersionFuture
+                .thenCompose(schemaVersion -> {
+                    return action.act(schemaVersion)
+                            .handle((BiFunction<T, Throwable, CompletableFuture<T>>) (res, ex) -> {
```
sashapolo, line 115: In my proposal I also provided a way to get rid of the cast to `BiFunction`, shall we use it?

sashapolo (approved):

vldpyatkov (approved):

PR #2774, opened 2023-10-31
IGNITE-20764 Rework script parsing to return correct number of dynamic parameters for each statement.

xtern:
https://issues.apache.org/jira/browse/IGNITE-20764

* The results of the `IgniteSqlParser#parse` method have been reworked to preserve the number of dynamic parameters for each statement in a multi-statement query.

* Added a separate `parseScript()` method to `PrepareService` for parsing multi-statement queries.


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/sql/ScriptParserService.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.ignite.internal.sql.engine.sql.ParserServiceImpl.ParsedResultImpl;
+import org.apache.ignite.internal.sql.engine.util.Commons;
+
+/**
+ * An implementation of {@link ParserService} for parsing multi-statements, this service does not cache parsing results.
+ */
+public class ScriptParserService implements ParserService<List<ParsedResult>> {
+    /** {@inheritDoc} */
+    @Override
+    public List<ParsedResult> parse(String query) {
```
korlov42, line 33: honestly, I would prefer extend existing class with `parseScript` or `parseMulti` method

xtern: Done, `ScriptParserService` removed.

zstan (approved):

korlov42 (approved):

PR #2786, opened 2023-11-03
IGNITE-18662 Sql. Numeric to/from decimal cast with overflow does not produce an error

zstan:

### The Review Checklist

### Notes


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -479,6 +487,61 @@ private void checkTypesInteroperability(SqlValidatorScope scope, SqlNode expr) {
...
             }
+
+            literalCanFitType(expr, returnType);
+        }
+    }
+
+    /** Check literal can fit to declared exact numeric type, work only for single literal. */
+    private static void literalCanFitType(SqlNode expr, RelDataType toType) {
+        if (INT_TYPES.contains(toType.getSqlTypeName())) {
+            LiteralExtractor litExtractor = new LiteralExtractor();
```
lowka, line 498: There is no need to create a new instance for every expression, because `SqlValidator` is stateful and `LiteralExtractor` can be stored in a field.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItDynamicParameterTest.java
```diff
@@ -82,6 +82,9 @@ public void testDynamicParameters() {
         assertQuery("SELECT ? % ?").withParams(11, 10).returns(1).check();
         assertQuery("SELECT ? + ?, LOWER(?) ").withParams(2, 2, "TeSt").returns(4, "test").check();
         assertQuery("SELECT LOWER(?), ? + ? ").withParams("TeSt", 2, 2).returns("test", 4).check();
+        // TODO: returns incorrect result, need to be fixed in scope of https://issues.apache.org/jira/browse/IGNITE-20784
+        assertQuery("SELECT (2147483647 + ?)::INTEGER").withParams(1).returns(-2147483648).check();;
+        assertQuery("SELECT (? + 2147483647)::INTEGER").withParams(1).returns(-2147483648).check();;
```
AMashenkov, line 87: What is the purpose of writing invalid test?
It would be better writing a commented code with expected result.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/Commons.java
```diff
@@ -861,4 +864,46 @@ public static SqlQueryType getQueryType(SqlNode sqlNode) {
...
+    public static long getMinValue(RelDataType type) {
+        SqlTypeName typeName = type.getSqlTypeName();
+        switch (typeName) {
+            case TINYINT:
+                return Byte.MIN_VALUE;
+            case SMALLINT:
+                return Short.MIN_VALUE;
+            case INTEGER:
+                return Integer.MIN_VALUE;
+            case BIGINT:
```
AMashenkov, line 881: Is BIGINT scaled? or `Long.MIN_VALUE` is missed here?

zstan: I just copy-paste this methods from calcite, cause i found that they not used anywhere and probably can be removed .. Plz explain what i need to check here ?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/ImplicitCastsTest.java
```diff
@@ -346,18 +341,14 @@ public Stream<DynamicTest> testUpdate() {
 
                 checkStatement()
                         .table("t1", "c1", NativeTypes.INT32)
-                        .sql("UPDATE t1 SET c1 = 'abc'")
-                        .project("$t0", "CAST(_UTF-8'abc'):INTEGER NOT NULL"),
```
lowka, line 350: I think this test case is either no longer necessary or should be updated to reflect that fact that numeric literals are checked during the validation phase.

lowka: The same could be applied to the several test cases for `SELECT IN`

zstan: I remove this test and prefer to stay with 'SELECT IN' tests ,they are quite small and seems useful

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/Commons.java
AMashenkov: `type.getPrecision()` is 19 for BIGINT
`NumberUtil.getMinUnscaled(type.getPrecision())` creates `BigInteger.valueOf(Long.MIN_VALUE)` for param value of 19, then converts to long.

Seems, it is safe to return Long.MIN_VALUE here.
Same is true for max_value.

AMashenkov (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -178,6 +194,33 @@ public void validateUpdate(SqlUpdate call) {
...
+            SqlValidatorScope sourceScope,
+            SqlValidatorTable table,
+            RelDataType sourceRowType,
+            RelDataType targetRowType,
+            SqlNode query
+    ) {
+        boolean coerced = false;
+
+        if (config().typeCoercionEnabled()) {
+            if (SqlTypeUtil.equalAsStructSansNullability(typeFactory,
```
lowka, line 209: Special casing for BIGINT does not fix the problem with BIGINT literals in apache calcite (there is a bug in calcite). Add this code should be removed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlToRelConvertor.java
```diff
@@ -83,6 +87,28 @@ public class IgniteSqlToRelConvertor extends SqlToRelConverter {
         }
     }
 
+    @Override
+    protected RexNode convertExtendedExpression(
```
lowka, line 91: This won't fix problem with BIGINT literals and this code should be removed. 
See comments below.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -479,6 +522,64 @@ private void checkTypesInteroperability(SqlValidatorScope scope, SqlNode expr) {
...
+    /** Check literal can fit to declared exact numeric type, work only for single literal. */
+    private void literalCanFitType(SqlNode expr, RelDataType toType) {
+        if (INT_TYPES.contains(toType.getSqlTypeName())) {
+            SqlLiteral literal = Objects.requireNonNullElseGet(litExtractor, LiteralExtractor::new).getLiteral(expr);
+
+            if (literal == null || literal.toValue() == null) {
+                return;
+            }
+
+            long max = Commons.getMaxValue(toType);
```
lowka, line 541: I think it is better to use calcite's `SqlTypeName::getLimit` here see `RexUtils::toSaturatedValue`.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteTypeCoercion.java
```diff
@@ -271,6 +271,12 @@ protected boolean needToCast(SqlValidatorScope scope, SqlNode node, RelDataType
             if (fromType == null) {
                 return false;
             }
+
+            if (fromType.getSqlTypeName() == SqlTypeName.BIGINT && toType.getSqlTypeName() == SqlTypeName.BIGINT) {
```
lowka, line 275: As mentioned earlier this is not necessary as well because it does not fix a problem with BIGINT literals.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteTypeCoercion.java
```diff
@@ -281,6 +287,10 @@ protected boolean needToCast(SqlValidatorScope scope, SqlNode node, RelDataType
             return TypeUtils.customDataTypeNeedCast(typeFactory, fromType, toType);
         }
 
+        if (fromType.getSqlTypeName() == SqlTypeName.ROW) {
```
lowka, line 290: It seems there is a bug in handling of Scalar subqueries in DML operations and this is not the way to fix it because this makes a cast from `RecordType(some_type)` to `some_type` legal, when it is obviously not the case.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/Commons.java
```diff
@@ -861,4 +864,46 @@ public static SqlQueryType getQueryType(SqlNode sqlNode) {
...
+            default:
+                throw new AssertionError("getMinValue(" + typeName + ")");
+        }
+    }
+
+    /** Returns the maximum unscaled value of a numeric type.
+     *
+     * @param type a numeric type
+     */
+    public static long getMaxValue(RelDataType type) {
```
lowka, line 893: Both this and `getMinValue` can be removed in favour of calcite's `SqlTypeName::getLimit`

lowka (changes requested):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
lowka: Well, I am mistaken about a bug in calcite. I think but special casing of BIGINT does not belong to this PR I removed the code related to special casing of BIGINT observed no changes in code behaviour.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteTypeCoercion.java
lowka: It more or less looks like an attempt to fix BIGINTs for scalar subqueries in DML statements. But it is not correct (see the next comment on cast from `ROW` type).

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -87,6 +95,11 @@ public class IgniteSqlValidator extends SqlValidatorImpl {
 
     private static final Set<SqlKind> HUMAN_READABLE_ALIASES_FOR;
 
+    public static final String NUMERIC_FIELD_OVERFLOW_ERROR = "Numeric field overflow";
+
+    //approximate and exact numeric types
+    private static final Pattern numeric = Pattern.compile("^\\s*\\d+(\\.{1}\\d*)\\s*$");
```
lowka, line 101: Javacode style: `numeric` -> `NUMERIC`.

AMashenkov (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
zstan: removing of this part leads to testInsertValueOverflow failure, something wrong with Idea caching or rebuilding, first run of such a test with commented lines gives me correct tests passing but second is already not, the same for IgniteSqlToRelConvertor part.

lowka (approved):

PR #2789, opened 2023-11-03
IGNITE-20443 Sql. Implement script processing logic.

xtern:
https://issues.apache.org/jira/browse/IGNITE-20443

* Added `queryScriptAsync` method to `QueryProcessor`
* Extended `AsyncSqlCursor` interface with two methods `hasNextResult` and `nextResult` (returns a future for the next statement cursor).

Implementation details.
1. Parsing script.
2. Creating a list of futures (for each statement). The iterator over these futures is returned to the user.
3. After prefetch for the current statement is done, we complete current cursor future with the corresponding sql cursor.. And move on to execute the next statement.


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -535,4 +604,179 @@ private static void validateParsedStatement(
...
+                            if (ex != null) {
+                                txWrapper.rollback();
+                                cursorFuture.completeExceptionally(ex);
+                                cancelAll(ex);
+
+                                return;
+                            }
+
+                            if (parsedResult.queryType() == SqlQueryType.DML || parsedResult.queryType() == SqlQueryType.DDL) {
+                                txWrapper.commitImplicit();
```
korlov42, line 689: why do we commit transaction only in case of DML | DDL statements?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -535,4 +604,179 @@ private static void validateParsedStatement(
...
+            PrefetchCallback(CompletableFuture<Void> prefetchFuture) {
+                this.prefetchFuture = prefetchFuture;
+            }
+
+            @Override
+            public void onPrefetchComplete(@Nullable Throwable ex) {
+                if (ex == null) {
+                    prefetchFuture.complete(null);
+
+                    processNext();
```
korlov42, line 763: It's better to keep all decision in a single place: if we make decision to proceed execution in PrefetchCallback, then we should interrupt execution in case of any error also here... Or we can move invocation of `processNext()` to the MultiStatementHandler

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -535,4 +604,179 @@ private static void validateParsedStatement(
...
+            for (ScriptStatementParameters parameters : statements) {
+                CompletableFuture<AsyncSqlCursor<List<Object>>> fut = parameters.cursorFuture;
+
+                if (fut == null || fut.isDone()) {
+                    continue;
+                }
+
+                fut.completeExceptionally(new SqlException(
+                        EXECUTION_CANCELLED_ERR,
+                        "The script statement execution was canceled due to an error in the previous statement.",
```
korlov42, line 745: nitpicking: I would omit first `statement` in the message

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryProcessor.java
```diff
@@ -53,4 +54,33 @@ CompletableFuture<AsyncSqlCursor<List<Object>>> querySingleAsync(
...
+     * @param transactions Transactions facade.
+     * @param transaction A transaction to use for query execution. If null, an implicit transaction
+     *      will be started by provided transactions facade.
+     * @param qry Single statement SQL query.
+     * @param params Query parameters.
+     * @return Sql cursor.
+     *
+     * @throws IgniteException in case of an error.
+     */
+    CompletableFuture<AsyncSqlCursorIterator<List<Object>>> queryScriptAsync(
```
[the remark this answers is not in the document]
xtern: You mean
```
**CompletableFuture**<AsyncIterator<AsyncSqlCursor<List<Object>>>> queryScriptAsync(...);
```
?
is it ok? :thinking:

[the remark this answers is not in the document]
xtern: Should it prefetch the first element (current) similar to AsyncResultSet?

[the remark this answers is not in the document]
korlov42: hm, I see confusion... My suggestion was to bring `java.util.Iterator` to async world. With `java.util.Iterator` the first element is not available until you call `next`.

So, in order to preserve semantic of `java.util.Iterator`, we need only two operation: `hasNext` and `next`. Perhaps, we should proceed without `current` to avoid confusion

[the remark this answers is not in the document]
xtern: Seems current() with prefetch adds complications for the user.
We do not have a method for checking that the iterator is empty (hasRowSet), so the user must, after receiving the iterator, not only call hasNext/fetchNext but also check that "current" exists or not.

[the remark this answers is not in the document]
xtern: Sorry, I missed your last comment, don't mind my last one. Thank you.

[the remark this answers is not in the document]
xtern: Thanks, added  `AsyncIterator` interface with  basic implementation `AsyncIteratorImpl`.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -535,4 +604,179 @@ private static void validateParsedStatement(
...
+            CompletableFuture<Void> prefetchFut = new CompletableFuture<>();
+            QueryCancel cancel = new QueryCancel();
+
+            return executeParsedStatement(schemaName, parsedResult, txWrapper, cancel, dynamicParams, new PrefetchCallback(prefetchFut))
+                    // Wait for the prefetch, otherwise the user will be able to execute
+                    // a dependent statement before the current one completes.
+                    .thenCompose(cur -> prefetchFut.thenApply(ignored -> cur));
+        }
+
+        private List<ScriptStatementParameters> splitDynamicParameters(List<ParsedResult> parsedResults, Object[] params) {
```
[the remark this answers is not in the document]
xtern: Thank you, completely missed this. 
Fixed. Adjustment is done now in `ScriptParseResult`.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
xtern: :thinking:  Fixed: Implicit transaction is now committed after prefetching for all query types.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
xtern: Fixed, thanks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
xtern: Thanks, `processNext()` moved to `CompletableFuture` listener.

korlov42 (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/sql/ScriptParseResult.java
```diff
@@ -69,25 +80,33 @@ public List<StatementParseResult> results() {
...
-
-            return null;
+        public SqlNode visit(SqlDynamicParam param) {
+            return new SqlDynamicParam(counter++, param.getParserPosition());
         }
 
-        int forNode(SqlNode node) {
-            count = 0;
+        /**
+         * Rewrites dynamic parameter nodes in the provided tree.
```
AMashenkov, line 95: I guess, this javadoc describes prev. method `visit(SqlDynamicParam param)`.
This method resets state between processing different `StatementParseResult` and allow to reuse SqlDynamicParamsAdjuster, isn't it?

Can you drop this method and write a new one that will reset state explicitly?
`void reset() { counter = 0; }`
then fix SqlDynamicParamsAdjuster usage: 
```
   dynamicParamsAdjuster.reset();
   SqlNode newTree = dynamicParamsAdjuster.visitNode(node);
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -535,4 +612,147 @@ private static void validateParsedStatement(
...
+            ParsedResult parsedResult = parameters.parsedResult;
+            Object[] dynamicParams = parameters.dynamicParams;
+            CompletableFuture<AsyncSqlCursor<List<Object>>> cursorFuture = parameters.cursorFuture;
+
+            try {
+                if (cursorFuture.isDone()) {
+                    return cursorFuture;
+                }
+
+                QueryTransactionWrapper txWrapper = wrapTxOrStartImplicit(parsedResult.queryType(), transactions, explicitTransaction);
```
AMashenkov, line 651: Can we replace both agruments with a single QueryTransactionWrapper
```
IgniteTransactions transactions,
@Nullable InternalTransaction explicitTransaction,
```
Then we could make use a pattern
```
txWrapper.startTxIfNeeded(queryType);
... execute query ... 
txWrapper.commitImplicit();
```
Inside the wrapper, we can have all the tx logic in one place and ensure 
* a query type match a tx type
* no inner transactions started
* prev. tx was commited/rolled back
* ignore tx commands in case of explicit tx.

AMashenkov: WDYT?

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/jdbc/JdbcQueryCursor.java
```diff
@@ -87,4 +87,18 @@ public SqlQueryType queryType() {
...
+    public boolean hasNextResult() {
+        // TODO https://issues.apache.org/jira/browse/IGNITE-20661
+        return false;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public CompletableFuture<AsyncSqlCursor<T>> nextResult() {
+        // TODO https://issues.apache.org/jira/browse/IGNITE-20661
+        throw new UnsupportedOperationException();
```
AMashenkov, line 102: Seems, we can throw correct error here
```suggestion
       if (!hasNextResult()) {
           throw new QueryHasNoMoreResultsException();
       }
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/sql/ScriptParseResult.java
xtern: Done.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
xtern: Overall, I agree with you.

But since I remember that there were a LOT of discussions on this topic previously [1] and since this task is not directly related to current issue? I filled out another ticket to resolve this separately [2]

[1] https://github.com/apache/ignite-3/pull/2449
[2] https://issues.apache.org/jira/browse/IGNITE-20855

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -535,4 +612,156 @@ private static void validateParsedStatement(
...
+            this.schemaName = schemaName;
+            this.transactions = transactions;
+            this.explicitTransaction = explicitTransaction;
+            this.statements = prepareStatementsQueue(parsedResults, params);
+        }
+
+        CompletableFuture<AsyncSqlCursor<List<Object>>> processNext() {
+            if (statements == null) {
+                return CompletableFuture.completedFuture(null);
+            }
```
AMashenkov, line 638: Why do you complete future with `null` value? Is it a kind of special value? 
Why don't return a cursor over empty collection?

xtern: As it described in javadoc
`     * @return Sql cursor or {@code null} if specified SQL query does not create cursors
      *         (for example, if it contains only commands for managing a transaction).`

AMashenkov (approved):

PR #2792, opened 2023-11-03
IGNITE-19925 Reduce the number of NodeStoppingExceptions in logs on node stop

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-19925

### The Review Checklist

### Notes


ibessonov (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -964,23 +967,30 @@ public void stop() {
...
 
         cleanUpTablesResources(tablesToStop);
+    }
 
-        try {
-            IgniteUtils.closeAllManually(lowWatermark, mvGc);
-        } catch (Throwable t) {
-            LOG.error("Failed to close internal components", t);
+    @Override
+    public void stop() throws Exception {
```
sanpwc, line 979: I almost sure that in some tests we may just call stop instead of proper chain of beforeNodeStop + stop. I'd add corresponding assertion that beforeStopGuard is true here in order to check (and fix) the guess.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -964,23 +967,30 @@ public void stop() {
         metaStorageMgr.unregisterWatch(stableAssignmentsRebalanceListener);
```
sanpwc, line 967: beforeNodeStop was added in order to have an ability to finish some actions that require local network layer to be up and running.
Why do we call e.g. metaStorageMgr.unregisterWatch() within beforeNodeStop? Does it really need a network? Or it should go before some other action that requires network or ...? It's also about stopping tables. Does it required a network?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sashapolo: Great idea!

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sashapolo: > Why do we call e.g. metaStorageMgr.unregisterWatch() within beforeNodeStop? Does it really need a network?

I've put these calls here, so that meta storage watches would be disabled when we are calling `stop`. Does it make sense?

> It's also about stopping tables. Does it required a network?

For tables - yes, we need the network. This is what the original problem of this ticket was about: when we call `TableManager#stop` we also call `Replica#shutdown` which does a network call (which never succeeds). That's why I moved the whole `cleanupTableResources` to `beforeNodeStop`.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -989,73 +999,46 @@ public void stop() {
...
-            InternalTable internalTable = table.internalTable();
+                InternalTable internalTable = table.internalTable();
 
-            for (int p = 0; p < internalTable.partitions(); p++) {
-                int partitionId = p;
+                for (int p = 0; p < internalTable.partitions(); p++) {
+                    TablePartitionId replicationGroupId = new TablePartitionId(table.tableId(), p);
 
-                TablePartitionId replicationGroupId = new TablePartitionId(table.tableId(), p);
+                    stopping.add(() -> raftMgr.stopRaftNodes(replicationGroupId));
```
sanpwc, line 1017: The order seems strange to me. Initially it was also confusing but because we've stooped network at the very beginning we might receive NodeStoppingException when replica tried to touch already stopped raft endpoint (and having NodeStoppingException is fine). Currently we will probably fail with something like "UnknownRaftGroup" because we've originally stopped raft before stopping replica. Basically the order of stopping will should be opposite to the order of starting, right? 
And, yep, you might say that busyLocks will protect us, but they won't. Let's even put aside the fact that not all the operations are covered with busyLocks. Besides that busyLocks only covers local operations, so e.g. on message send you may exit the busylock and within given point you may start beforeNodeStop that will stop the raft endpoint.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sanpwc: > I've put these calls here, so that meta storage watches would be disabled when we are calling stop. Does it make sense?

That doesn't matter, but because ms.unregister should go before replica.stop and replica stop do require network, proposed order and localization seems reasonable.

> For tables - yes, we need the network. This is what the original problem of this ticket was about: when we call TableManager#stop we also call Replica#shutdown which does a network call (which never succeeds). That's why I moved the whole cleanupTableResources to beforeNodeStop.

Sounds reasonable.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sashapolo: > The order seems strange to me

Well, this was the original order, I didn't change anything here.

> Basically the order of stopping will should be opposite to the order of starting, right?

I agree, let's change it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1014,10 +1014,10 @@ private void cleanUpTablesResources(Map<Integer, TableImpl> tables) {
...
 
-                    stopping.add(() -> raftMgr.stopRaftNodes(replicationGroupId));
-
                     stopping.add(() -> replicaMgr.stopReplica(replicationGroupId).get(10, TimeUnit.SECONDS));
 
+                    stopping.add(() -> raftMgr.stopRaftNodes(replicationGroupId));
+
                     CompletableFuture<Void> removeFromGcFuture = mvGc.removeStorage(replicationGroupId);
 
                     stopping.add(() -> closePartitionTrackers(internalTable, replicationGroupId.partitionId()));
```
sanpwc, line 1023: I'd also close partition trackers before stopping partition.

sashapolo: okay

sanpwc (approved):

PR #2810, opened 2023-11-07
Criteria query API design.

nva:


modules/api/src/main/java/org/apache/ignite/sql/Criteria.java
```diff
@@ -0,0 +1,132 @@
...
+ */
+public interface Criteria {
+    /**
+     * Create a predicate for testing the column is equal to a given value.
+     *
+     * @param columnName Column name.
+     * @param value Column value.
+     * @return the created <b>equal</b> predicate instance.
+     */
+    public static <R> Criteria equal(String columnName, Comparable<R> value) {
```
AMashenkov, line 33:
```suggestion
    public static <R> Criteria equal(String columnName, Object value) {
```
`byte[]` is valid argument for binary column, but arrays are not Comparable in Java.

modules/api/src/main/java/org/apache/ignite/table/KeyValueView.java
```diff
@@ -491,4 +494,23 @@ public interface KeyValueView<K, V> extends DataStreamerTarget<Entry<K, V>> {
...
+    }
+
+    /**
+     * Criteria query over cache entries.
+     *
+     * @param tx Transaction or {@code null} to auto-commit.
+     * @param criteria If {@code null} then all entries will be returned.
+     * @param opts Criteria query options.
+     */
+    CriteriaQueryCursor<Entry<K, V>> criteriaQuery(@Nullable Transaction tx, @Nullable Criteria criteria, CriteriaQueryOptions opts);
```
ptupitsyn, line 515: To avoid duplicating those declarations across multiple interfaces, let's move them to `CriteriaQuerySource` interface - similar to `DataStreamerTarget`.

modules/api/src/main/java/org/apache/ignite/table/KeyValueView.java
```diff
@@ -491,4 +494,23 @@ public interface KeyValueView<K, V> extends DataStreamerTarget<Entry<K, V>> {
...
      */
     CompletableFuture<NullableValue<V>> getNullableAndReplaceAsync(@Nullable Transaction tx, K key, V val);
+
+    /**
+     * Criteria query over cache entries.
+     *
+     * @param tx Transaction or {@code null} to auto-commit.
+     * @param criteria If {@code null} then all entries will be returned.
+     */
+    default CriteriaQueryCursor<Entry<K, V>> criteriaQuery(@Nullable Transaction tx, @Nullable Criteria criteria) {
```
ptupitsyn, line 504:
```suggestion
    default CriteriaQueryCursor<Entry<K, V>> queryCriteria(@Nullable Transaction tx, @Nullable Criteria criteria) {
```

To be consistent with `queryContinuously` and `streamData`. Also makes it easier to discover with IDE autocomplete.

modules/api/src/main/java/org/apache/ignite/sql/CriteriaQueryCursor.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * Criteria query result cursor. Implements {@link Iterable} only for convenience, e.g. {@link #iterator()}
+ * can be obtained only once. Also if iteration is started then {@link #getAll()} method calls are prohibited.
+ */
+public interface CriteriaQueryCursor<T> extends Iterator<T>, Iterable<T>, AutoCloseable {
```
ptupitsyn, line 31: We should probably reuse `AsyncResultSet` and `ResultSet`.

modules/api/src/main/java/org/apache/ignite/table/KeyValueView.java
ptupitsyn, line 515: Async variants are missing.

ptupitsyn (changes requested):

PR #2818, opened 2023-11-08
IGNITE-20524 Fix nullable annotations and javadoc in KeyValueView

ptupitsyn:
`KeyValueView` allows null value in all methods when `OneColumnMapper` is used. Reflect this in Javadoc and add `@Nullable` annotations accordingly.


modules/api/src/main/java/org/apache/ignite/table/KeyValueView.java
```diff
@@ -155,21 +155,21 @@ public interface KeyValueView<K, V> extends DataStreamerTarget<Entry<K, V>> {
      *
      * @param tx Transaction or {@code null} to auto-commit.
      * @param key Key with which the specified value is to be associated. The key cannot be {@code null}.
-     * @param val Value to be associated with the specified key.
+     * @param val Value to be associated with the specified key. Can be null when mapped to a single column with a simple type.
```
AMashenkov, line 158: Here and below
```suggestion
     * @param val Value to be associated with the specified key. Can be {@code null} when mapped to a single column with a simple type.
```

AMashenkov (approved):

modules/api/src/main/java/org/apache/ignite/table/KeyValueView.java
ptupitsyn: Fixed.

PR #2820, opened 2023-11-08
IGNITE-20709

sanpwc:

### The Review Checklist

### Notes

### DIFF
--- a/modules/raft/src/main/java/org/apache/ignite/raft/jraft/error/RaftError.java
+++ b/modules/raft/src/main/java/org/apache/ignite/raft/jraft/error/RaftError.java
@@ -238,7 +238,12 @@ public enum RaftError {
     /**
      * Permission denied
      */
-    EACCES(1016);
+    EACCES(1016),
+
+    /**
+     * Command reordering detected
+     */
+    EREORDER(1017);
 
     private static final Map<Integer, RaftError> RAFT_ERROR_MAP = new HashMap<>();
 
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
@@ -157,8 +157,10 @@ public class PartitionListener implements RaftGroupListener {
                     maxObservableSafeTime = cmd.safeTimeLong();
                 } else {
                     System.out.println("!!! maxObservableSafeTime = " + maxObservableSafeTime + ", cmd.safeTimeLong() = " + cmd.safeTimeLong() + ", " + clo.command().getClass());
-                    clo.result(new SafeTimeReorderException(REPLICATION_SAFE_TIME_MISS, "errorMsg"));
-                    return;
+                    if (!(command instanceof TxCleanupCommand)) {
+                        clo.result(new SafeTimeReorderException(REPLICATION_SAFE_TIME_MISS, "errorMsg"));
+                        return;
+                    }
                 }
             }
 
@@ -224,7 +226,10 @@ public class PartitionListener implements RaftGroupListener {
 
                 assert safeTimePropagatingCommand.safeTime() != null;
 
-                updateTrackerIgnoringTrackerClosedException(safeTime, safeTimePropagatingCommand.safeTime());
+                // TODO: sanpwc txcleanup command shouldn't implement SafeTimePropagatingCommand
+                if (!(command instanceof TxCleanupCommand)) {
+                    updateTrackerIgnoringTrackerClosedException(safeTime, safeTimePropagatingCommand.safeTime());
+                }
             }
 
             updateTrackerIgnoringTrackerClosedException(storageIndexTracker, commandIndex);
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
@@ -217,7 +217,9 @@ public class PartitionListener implements RaftGroupListener, BeforeApplyHandler 
 
                 // TODO: sanpwc txcleanup command shouldn't implement SafeTimePropagatingCommand
                 if (!(command instanceof TxCleanupCommand)) {
-                    updateTrackerIgnoringTrackerClosedException(safeTime, safeTimePropagatingCommand.safeTime());
+                    synchronized (safeTime) {
+                        updateTrackerIgnoringTrackerClosedException(safeTime, safeTimePropagatingCommand.safeTime());
+                    }
                 }
             }
 
@@ -465,8 +467,10 @@ public class PartitionListener implements RaftGroupListener, BeforeApplyHandler 
 //                System.out.println("Updating maxObservableSafeTime cmd.safeTimeLong() = " + cmd.safeTimeLong() + ", " + command.getClass());
                 maxObservableSafeTime = cmd.safeTimeLong();
             } else {
-                System.out.println("!!! maxObservableSafeTime = " + maxObservableSafeTime + ", cmd.safeTimeLong() = " + cmd.safeTimeLong() + ", " + command.getClass());
-                if (!(command instanceof TxCleanupCommand)) {
+//                System.out.println(
+//                        "!!! maxObservableSafeTime = " + maxObservableSafeTime + ", cmd.safeTimeLong() = " + cmd.safeTimeLong() + ", "
+//                                + command.getClass());
+                if (cmd instanceof TxCleanupCommand) {
                     // TODO: Use proper message instead
                     throw new SafeTimeReorderException(REPLICATION_SAFE_TIME_MISS, "errorMsg");
                 }
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
@@ -47,6 +47,7 @@ import org.apache.ignite.internal.logger.Loggers;
 import org.apache.ignite.internal.raft.Command;
 import org.apache.ignite.internal.raft.ReadCommand;
 import org.apache.ignite.internal.raft.WriteCommand;
+import org.apache.ignite.internal.raft.service.BeforeApplyHandler;
 import org.apache.ignite.internal.raft.service.CommandClosure;
 import org.apache.ignite.internal.raft.service.CommittedConfiguration;
 import org.apache.ignite.internal.raft.service.RaftGroupListener;
@@ -79,7 +80,7 @@ import org.jetbrains.annotations.TestOnly;
 /**
  * Partition command handler.
  */
-public class PartitionListener implements RaftGroupListener {
+public class PartitionListener implements RaftGroupListener, BeforeApplyHandler {
     /** Transaction manager. */
     private final TxManager txManager;
 
@@ -151,18 +152,6 @@ public class PartitionListener implements RaftGroupListener {
     public void onWrite(Iterator<CommandClosure<WriteCommand>> iterator) {
         iterator.forEachRemaining((CommandClosure<? extends WriteCommand> clo) -> {
             Command command = clo.command();
-            if (command instanceof SafeTimePropagatingCommand) {
-                SafeTimePropagatingCommand cmd = (SafeTimePropagatingCommand) command;
-                if (cmd.safeTimeLong() > maxObservableSafeTime) {
-                    maxObservableSafeTime = cmd.safeTimeLong();
-                } else {
-                    System.out.println("!!! maxObservableSafeTime = " + maxObservableSafeTime + ", cmd.safeTimeLong() = " + cmd.safeTimeLong() + ", " + clo.command().getClass());
-                    if (!(command instanceof TxCleanupCommand)) {
-                        clo.result(new SafeTimeReorderException(REPLICATION_SAFE_TIME_MISS, "errorMsg"));
-                        return;
-                    }
-                }
-            }
 
             long commandIndex = clo.index();
             long commandTerm = clo.term();
@@ -464,6 +453,25 @@ public class PartitionListener implements RaftGroupListener {
     @Override
     public void onShutdown() {
         storage.close();
+    }
+
+    @Override
+    public void onBeforeApply(Command command) {
+        // TODO: comment about thread safety
+        if (command instanceof SafeTimePropagatingCommand) {
+            SafeTimePropagatingCommand cmd = (SafeTimePropagatingCommand) command;
+
+            if (cmd.safeTimeLong() > maxObservableSafeTime) {
+//                System.out.println("Updating maxObservableSafeTime cmd.safeTimeLong() = " + cmd.safeTimeLong() + ", " + command.getClass());
+                maxObservableSafeTime = cmd.safeTimeLong();
+            } else {
+                System.out.println("!!! maxObservableSafeTime = " + maxObservableSafeTime + ", cmd.safeTimeLong() = " + cmd.safeTimeLong() + ", " + command.getClass());
+                if (!(command instanceof TxCleanupCommand)) {
+                    // TODO: Use proper message instead
+                    throw new SafeTimeReorderException(REPLICATION_SAFE_TIME_MISS, "errorMsg");
+                }
+            }
+        }
     }
 
     /**
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
@@ -25,6 +25,7 @@ import static org.apache.ignite.internal.tx.TxState.ABORTED;
 import static org.apache.ignite.internal.tx.TxState.COMMITED;
 import static org.apache.ignite.internal.tx.TxState.PENDING;
 import static org.apache.ignite.internal.util.CollectionUtils.last;
+import static org.apache.ignite.lang.ErrorGroups.Replicator.REPLICATION_SAFE_TIME_MISS;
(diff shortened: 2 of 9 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2383,6 +2424,35 @@ private CompletableFuture<Object> applyCmdWithExceptionHandling(Command cmd) {
...
+        raftClient.run(cmd).whenComplete((res, ex) -> {
+            if (ex != null) {
+                if (ex instanceof SafeTimeReorderException || ex.getCause() instanceof SafeTimeReorderException) {
+                    assert cmd instanceof SafeTimePropagatingCommand;
+
+                    SafeTimePropagatingCommand safeTimePropagatingCommand = (SafeTimePropagatingCommand) cmd;
+
+                    HybridTimestamp safeTimeForRetry = hybridClock.now();
+
+                    if ((cmd instanceof UpdateCommand && !((UpdateCommand) cmd).full())
```
rpuch, line 2437: Why is this condition needed? Could you please add a comment explaining this?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2383,6 +2424,35 @@ private CompletableFuture<Object> applyCmdWithExceptionHandling(Command cmd) {
...
+                            || (cmd instanceof UpdateAllCommand && !((UpdateAllCommand) cmd).full())) {
+                        synchronized (safeTime) {
+                            updateTrackerIgnoringTrackerClosedException(safeTime, safeTimeForRetry);
+                        }
+                    }
+
+                    safeTimePropagatingCommand.safeTimeLong(safeTimeForRetry.longValue());
+
+                    applyCmdWithRetryOnSafeTimeReorderException(safeTimePropagatingCommand, resultFuture);
```
rpuch, line 2446: Is it possible to add some kind of a safeguard to check/assert that we are not retrying infinitely due to some bug?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2420,34 +2490,40 @@ private CompletableFuture<CompletableFuture<?>> applyUpdateCommand(
             );
 
             if (!cmd.full()) {
-                CompletableFuture<UUID> fut = applyCmdWithExceptionHandling(cmd).thenApply(res -> {
+                // TODO: https://issues.apache.org/jira/browse/IGNITE-20124 Temporary code below
+                synchronized (safeTime) {
+                    storageUpdateHandler.handleUpdate(
```
rpuch, line 2495: Why is it that for non-full commands, we apply the command on the Primary first, and only then send the Raft command, but for full commands, the order is opposite?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
rpuch: Also, this check seems to be fragile. What if we add an analogous request in the future, but forget to mention it here?

How about adding a marker interface common for both update commands? It might be more difficult to forget about it when adding a third 'update' command in the future.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2420,34 +2490,40 @@ private CompletableFuture<CompletableFuture<?>> applyUpdateCommand(
...
+                            null,
+                            null,
+                            null);
+
+                    updateTrackerIgnoringTrackerClosedException(safeTime, cmd.safeTime());
+                }
+
+                CompletableFuture<Object> resultFuture = new CompletableFuture<>();
+
+                applyCmdWithExceptionHandling(cmd, resultFuture);
```
rpuch, line 2510: So we just applied the command on the Primary, and then we send it to Raft with the same SafeTime? Will not this mean that the command will be rejected with probability 100%?

rpuch (changes requested):

modules/raft/src/main/java/org/apache/ignite/raft/jraft/error/RaftError.java
```diff
@@ -238,7 +238,12 @@ public enum RaftError {
     /**
      * Permission denied
      */
-    EACCES(1016);
+    EACCES(1016),
+
+    /**
+     * Command reordering detected
+     */
+    EREORDER(1017);
```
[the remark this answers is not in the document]
sanpwc: Sure. Fixed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -99,6 +101,15 @@ public class PartitionListener implements RaftGroupListener {
     /** Storage index tracker. */
     private final PendingComparableValuesTracker<Long, Void> storageIndexTracker;
 
+    // TODO: https://issues.apache.org/jira/browse/IGNITE-20826 Restore on restart
+    /** Is used in order to detect and retry safe time reordering within onBeforeApply. */
+    private long maxObservableSafeTime = -1;
+
+    // TODO: https://issues.apache.org/jira/browse/IGNITE-20826 Restore on restart
+    /** Is used in order to assert safe time reordering within onWrite. */
+    private long maxObservableSafeTimeVerificator = -1;
```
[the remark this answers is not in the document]
sanpwc: Fixed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -455,6 +478,21 @@ public void onShutdown() {
         storage.close();
     }
 
+    @Override
+    public void onBeforeApply(Command command) {
+        // Given method is is synchronized by replication group specific monitor, see ActionRequestProcessor#handleRequest.
```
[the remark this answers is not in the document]
sanpwc: Fixed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1587,34 +1594,53 @@ private CompletableFuture<Object> finishTransaction(
...
 
+    private CompletableFuture<Object> applyFinishCommand(
+            UUID transactionId,
+            boolean commit,
+            HybridTimestamp commitTimestamp,
+            String txCoordinatorId,
+            int catalogVersion,
+            List<TablePartitionIdMessage> tablePartitionIds
+    ) {
+        // TODO: synchonized should go here. Not within finishTransaction but with all apply command.
```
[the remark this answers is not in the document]
sanpwc: TODO was outdated. Removed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1677,32 +1703,45 @@ private CompletableFuture<Void> processTxCleanupAction(TxCleanupReplicaRequest r
...
+                .requiredCatalogVersion(catalogVersion)
+                .build();
+
+        storageUpdateHandler.handleTransactionCleanup(transactionId, commit, commitTimestamp);
+
+        CompletableFuture<Object> resultFuture = new CompletableFuture<>();
+
+        applyCmdWithRetryOnSafeTimeReorderException(txCleanupCmd, resultFuture);
+
+        return resultFuture.thenApply(res -> null);
```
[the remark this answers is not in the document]
sanpwc: No, it's a bug. Fixed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2371,8 +2410,10 @@ private static <T> boolean allElementsAreNull(List<T> list) {
      * @param cmd Raft command.
      * @return Raft future.
      */
-    private CompletableFuture<Object> applyCmdWithExceptionHandling(Command cmd) {
-        return raftClient.run(cmd).exceptionally(throwable -> {
+    private void applyCmdWithExceptionHandling(Command cmd, CompletableFuture<Object> resultFuture) {
+        applyCmdWithRetryOnSafeTimeReorderException(cmd, resultFuture);
+
+        resultFuture.exceptionally(throwable -> {
```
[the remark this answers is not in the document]
sanpwc: Good catch! Fixed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2383,6 +2424,35 @@ private CompletableFuture<Object> applyCmdWithExceptionHandling(Command cmd) {
         });
     }
 
+    private void applyCmdWithRetryOnSafeTimeReorderException(Command cmd, CompletableFuture<Object> resultFuture) {
+        raftClient.run(cmd).whenComplete((res, ex) -> {
+            if (ex != null) {
+                if (ex instanceof SafeTimeReorderException || ex.getCause() instanceof SafeTimeReorderException) {
+                    assert cmd instanceof SafeTimePropagatingCommand;
```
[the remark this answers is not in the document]
sanpwc: Do you mean that applyCmdWithRetryOnSafeTimeReorderException should have SafeTimePropagatingCommand instead of Command as a parameter? I don't think it's convenient because in that case we should skip calling  applyCmdWithRetryOnSafeTimeReorderException for all commands except SafeTimePropagatingCommand and skip raftClientRun directly.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: > Why is this condition needed? Could you please add a comment explaining this?

Added.
```
// Within primary replica it's required to update safe time in order to prevent double storage updates in case of !1PC.
// Otherwise, it may be possible that a newer entry will be overwritten by an older one that came as part of the raft
// replication flow:
// tx1 = transactions.begin();
// tx1.put(k1, v1) -> primary.apply(k1,v1) + asynchronous raft replication (k1,v1)
// tx1.put(k1, v2) -> primary.apply(k1,v2) + asynchronous raft replication (k1,v1)
// (k1,v1) replication overrides newer (k1, v2). Eventually (k1,v2) replication will restore proper value.
// However it's possible that tx1.get(k1) will see v1 instead of v2.
```

> 
> Also, this check seems to be fragile. What if we add an analogous request in the future, but forget to mention it here?
> 
> How about adding a marker interface common for both update commands? It might be more difficult to forget about it when adding a third 'update' command in the future.

I agree that given solution is fragile. It should be refactored with something more reliable in https://issues.apache.org/jira/browse/IGNITE-20124  TODO added. Here I'm talking not only about safeTime adjustment on retry but about general double updates prevention logic.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: Because in case of full command we apply not the write intent but write itself. Writes doesn't require txn state resolution and thus visible to other transactions directly. All in all that means that we should prevent write visibility in case of replication failure. The reliable way to do it is to put replication prior to primary replica adjustment.

Why we need direct put on primary replica?
Not only because replication is finished when it's finished on majoirty that may miss primary but also because primary replica may not be a part of replication group.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: Attempts counter along with limit added.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
rpuch: Thanks for adding the comment, but my confusion is caused by another question: why DON'T we update safe time in case of 1PC?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
rpuch: This explains why for 1PC we do 'send to RAFT first, write to the Primary storage second'. But why for !1PC do we do it in the opposite order?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: Because in case of 1PC replication goes prior to local storage update.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: It's just a statement and you do not expect any comments here, right?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -938,7 +942,8 @@ private CompletableFuture<Void> processReplicaSafeTimeSyncRequest(ReplicaSafeTim
 
         applyCmdWithRetryOnSafeTimeReorderException(
                 REPLICA_MESSAGES_FACTORY.safeTimeSyncCommand().safeTimeLong(hybridClock.nowLong()).build(),
-                resultFuture
+                resultFuture,
+                0
```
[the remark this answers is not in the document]
sanpwc: Agree. Fixed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
rpuch: Replication is async anyway (it returns a future which we don't wait). For !1PC, before your fix it was (schematically):

```
var future = replicateCommand(cmd);
applyLocally(cmd);
return future;
```

Now it is

```
applyLocally(cmd);
var future = replicateCommand(cmd);
return future;
```
It seems that we did not wait for replication in the previous version of the code as well. But the order has changed. Why? :)

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
rpuch: True

rpuch (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: Because in new version it's guaranteed that we will update storage and safeTime prior to replication. Before there was a race here.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -187,6 +190,9 @@ public class PartitionReplicaListener implements ReplicaListener {
     /** Factory for creating replica command messages. */
     private static final ReplicaMessagesFactory REPLICA_MESSAGES_FACTORY = new ReplicaMessagesFactory();
 
+    /** Replication retries limit. */
+    private static final int MAX_RETIES_ON_SAFE_TIME_REORDERING = 1000;
```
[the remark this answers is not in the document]
sanpwc: Please check prior Roma's comments.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -148,7 +159,17 @@ public void onWrite(Iterator<CommandClosure<WriteCommand>> iterator) {
...
+            if (command instanceof SafeTimePropagatingCommand) {
+                SafeTimePropagatingCommand cmd = (SafeTimePropagatingCommand) command;
+                long proposedSafeTime = cmd.safeTime().longValue();
+
+                if (proposedSafeTime > maxObservableSafeTimeVerifier) {
+                    maxObservableSafeTimeVerifier = proposedSafeTime;
+                } else {
+                    assert false : "Safe time reordering detected [current=" + maxObservableSafeTimeVerifier
+                            + ", proposed=" + proposedSafeTime + "]";
+                }
```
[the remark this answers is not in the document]
sanpwc: > it would be better to remove it altogether.

I don't think so. Given assertion will ease tests failures analysis. In case of unexpected data we will see whether there was a reordering or not.

>  assert roposedSafeTime <= maxObservableSafeTimeVerifier : "Safe time reordering detected [current=" + maxObservableSafeTimeVerifier + ", proposed=" + proposedSafeTime + "].";

maxObservableSafeTimeVerifier = proposedSafeTime;

Agreed. Fixed. Except the part that proposed should be greater than maxObservableSafeTimeVerifier

PR #2823, opened 2023-11-09
IGNITE-20819 Use shared lock on HTTP server startup

PakhomovAlexander:
It happens bacause of the race condition inside
micronaut.

https://github.com/micronaut-projects/micronaut-core/issues/10091

https://issues.apache.org/jira/browse/IGNITE-20819

### DIFF
--- a/modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
+++ b/modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
@@ -31,6 +31,8 @@ import java.net.UnknownHostException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Supplier;
 import org.apache.ignite.internal.lang.IgniteInternalException;
 import org.apache.ignite.internal.logger.IgniteLogger;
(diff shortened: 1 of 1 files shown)


modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
```diff
@@ -51,6 +53,11 @@
  * resources for the example.
  */
 public class RestComponent implements IgniteComponent {
+    /**
+     * Lock for micronaut server startup.
+     * TODO: remove when fix https://github.com/micronaut-projects/micronaut-core/issues/10091
+     */
+    private static final Lock SHARED_STARTUP_LOCK = new ReentrantLock();
```
sashapolo, line 60: Just curious: why do you use a `Lock` and not `synchronized` on a static Object?

modules/rest/src/main/java/org/apache/ignite/internal/rest/RestComponent.java
```diff
@@ -115,9 +122,12 @@ public void start() {
      * @return {@code True} if server was started successfully, {@code False} if couldn't bind one of the ports.
      */
     private synchronized boolean startServer(int httpPortCandidate, int httpsPortCandidate, boolean sslEnabled, boolean dualProtocol) {
+        // Workaround to avoid micronaut race condition on startup.
+        SHARED_STARTUP_LOCK.lock();
         try {
             httpPort = httpPortCandidate;
```
sashapolo, line 128: Not related to this PR, but both `httpPort` and `httpsPort` must be made `volatile`, there's a race condition in `httpPort()` and `httpsPort()`  methods

sashapolo (approved):

PR #2825, opened 2023-11-09
IGNITE-20701 Use primary replica for client partition awareness

ptupitsyn:
* Use primary replica instead of leader for client partition awareness - primary replica can handle requests more efficiently.
* Track primary replicas per partition per table in `ClientPrimaryReplicaTracker`. 
  * Use placement driver to request initial replica assignment
  * Then, keep the list up to date with event subscription
  * Therefore, most client requests (`ClientTablePartitionPrimaryReplicasGetRequest`) will be handled synchronously
* Send assignment timestamp (`max(leaseStartTime)`) to the client and back:
  * Client can ignore stale updates from lagging nodes
  * Server waits for an event with the given timestamp when lagging behind
* **Important**: Primary replica assignment can contain nulls now - we wait for a lease for a small amount of time, then skip the partition. In this case the client should use the default connection for this partition.
* **Client protocol changed**: 
  * Partition assignment change notification includes timestamp
  * Partition assignment request includes timestamp from the notification


isapego (approved):

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
```diff
@@ -690,21 +689,23 @@ private void processOperation(ChannelHandlerContext ctx, ClientMessageUnpacker i
...
+        boolean primaryReplicasChanged = localUpdateCount < updateCount
+                && primaryReplicaUpdateCount.compareAndSet(localUpdateCount, updateCount);
+
+        if (primaryReplicasChanged && LOG.isInfoEnabled()) {
+            LOG.info("Partition primary replica changed, notifying client [connectionId=" + connectionId + ", remoteAddress="
                     + ctx.channel().remoteAddress() + ']');
         }
 
-        var flags = ResponseFlags.getFlags(assignmentChanged);
+        var flags = ResponseFlags.getFlags(primaryReplicasChanged);
```
sanpwc, line 705: Why not to send primaryReplicaUpdateCount backwards in order to prevent schema notifications reordering?

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
```diff
@@ -0,0 +1,239 @@
...
+    private final HybridClock clock;
+
+    private final AtomicLong lruCheckTime = new AtomicLong(0);
+
+    private final EventProducer<CatalogEvent, CatalogEventParameters> catalogEventProducer;
+
+    /**
+     * Constructor.
+     *
+     * @param placementDriver Placement driver.
```
sanpwc, line 67: There are 4 parameters actually.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
```diff
@@ -0,0 +1,239 @@
...
+        placementDriver.removeListener(PrimaryReplicaEvent.PRIMARY_REPLICA_ELECTED, (EventListener) this);
+        primaryReplicas.clear();
+    }
+
+    private CompletableFuture<List<ReplicaHolder>> initReplicasForTableAsync(Integer tableId) {
+        try {
+            // Initially, request all primary replicas for the table.
+            // Then keep them updated via PRIMARY_REPLICA_ELECTED events.
+            return igniteTables
+                    .tableAsync(tableId)
```
sanpwc, line 121: Why do you need tableAsync()? TableId is available though catalog. Partitions is a zone attribute
`org.apache.ignite.internal.catalog.descriptors.CatalogZoneDescriptor#partitions` where corresponding zoneId is also available in org.apache.ignite.internal.catalog.descriptors.CatalogTableDescriptor#zoneId. Thus you probably need catalog but not table and especially InternalTable.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerModule.java
```diff
@@ -204,6 +212,7 @@ public void start() {
     @Override
     public void stop() throws Exception {
         metricManager.unregisterSource(metrics);
```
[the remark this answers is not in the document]
ptupitsyn: It's a local operation.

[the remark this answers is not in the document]
ptupitsyn: 1. All the logic in this method is idempotent, so `stopGuard` does not matter.
2. As I understand, `busyLock` won't do anything either. All external input for this module comes through Netty. We can't wait for clients to disconnect in a busy lock. We just drop them by closing the channel.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerModule.java
```diff
@@ -177,6 +182,7 @@ public ClientHandlerModule(
         this.clock = clock;
         this.schemaSyncService = schemaSyncService;
         this.catalogService = catalogService;
+        this.primaryReplicaTracker = new ClientPrimaryReplicaTracker(placementDriver, igniteTables, catalogService, clock);
```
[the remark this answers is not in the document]
ptupitsyn: Catalog does not provide the number of partitions for a given table, which is needed in `initReplicasForTableAsync`. That's the only reason to pull `igniteTables` in there. 

It would be great to get rid of `igniteTables` if we can retrieve the partition count from `PlacementDriver` or `CatalogService`, but I could not find a way to do that.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
ptupitsyn: `primaryReplicaUpdateCount` values from different nodes are not comparable.

It's a local count, just a trick to track whether a particular client connection (represented by `ClientInboundMessageHandler`)  has sent the notification flag for the recent update in the (global, one per node) `ClientPrimaryReplicaTracker`.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
ptupitsyn: Fixed.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
ptupitsyn: Thank you, good point. I've added catalog later and did not notice that `partitions` is available from there.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerModule.java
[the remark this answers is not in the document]
ptupitsyn: `igniteTables` is only used to get `partitions` - I missed the fact that `catalogService` can provide that. Will remove.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerModule.java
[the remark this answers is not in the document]
sanpwc: What if we have some pending action along with concurrent stop?

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
sanpwc: Ok, in that case, like we've discussed in person it's required to use some linearisable "counter" e.g. timestamp as you've proposed. And probably such counter should go to the client and backwards in order to handle races within multi-connections.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerModule.java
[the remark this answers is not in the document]
ptupitsyn: There is no such action. `stop` and `start` can't be concurrent (right?), and we don't have any other public methods.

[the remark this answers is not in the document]
ptupitsyn: If you insist, we can add those things just to be safe and consistent with other components.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
```diff
@@ -0,0 +1,253 @@
...
+        } catch (Throwable t) {
+            return CompletableFuture.failedFuture(t);
+        }
+    }
+
+    private CompletableFuture<List<ReplicaHolder>> primaryReplicasAsyncInternal(int tableId, int partitions) {
+        CompletableFuture<ReplicaMeta>[] futs = (CompletableFuture<ReplicaMeta>[]) new CompletableFuture[partitions];
+
+        for (int partition = 0; partition < partitions; partition++) {
+            futs[partition] = placementDriver.getPrimaryReplica(new TablePartitionId(tableId, partition), clock.now());
```
sanpwc, line 143: There's no need to call clock.now() each time. Basically even `long timestamp = clock.nowLong();` from initReplicasForTableAsync might be used.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        CompletableFuture<Void> all = CompletableFuture.allOf(futs);
+
+        return all.thenApply(v -> {
```
sanpwc, line 148: I'd rather add an exception handler with a corresponding assertion.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
ptupitsyn: Good point, fixed.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
ptupitsyn: Could you please elaborate? Not sure I understand.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
```diff
@@ -0,0 +1,253 @@
...
+                        ? new ReplicaHolder(null, HybridTimestamp.MIN_VALUE)
+                        : new ReplicaHolder(replicaMeta.getLeaseholder(), replicaMeta.getStartTime()));
+            }
+
+            return replicaHolders;
+        });
+    }
+
+    @Override
+    public CompletableFuture<Boolean> notify(EventParameters parameters, @Nullable Throwable exception) {
```
sanpwc, line 166: Instead of having one notify for both catalog and replica events I'd rather parametrize each listener with corresponding handler. It's however a matter of taste - up to you.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
```diff
@@ -0,0 +1,253 @@
...
+
+        PrimaryReplicaEventParameters primaryReplicaEvent = (PrimaryReplicaEventParameters) parameters;
+        if (!(primaryReplicaEvent.groupId() instanceof TablePartitionId)) {
+            return CompletableFuture.completedFuture(false);
+        }
+
+        TablePartitionId tablePartitionId = (TablePartitionId) primaryReplicaEvent.groupId();
+        var hld = primaryReplicas.get(tablePartitionId.tableId());
+
+        if (hld != null) {
```
sanpwc, line 192: Seems that you have a race here.
0. On start you've registered a listener on primary replica election.
1. Thread 1 at some point fixes the getPrimaryTimestamp and hangs.
2. Primary expires and new one is elected - you receive the notification and ignore it
3. Thread 1 wakes up and populates the map with obsolete primary replica.

I believe that you can use primaryReplica startTime in order to linearise all possible updates and thus understand whether you need to update corresponding partition meta or not.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerModule.java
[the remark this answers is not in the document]
sanpwc: > There is no such action. stop and start
I'm talking about concurrent stop and e.g. notification(PrimaryReplicaElection) processing. It might be covered with separate ticket in order not to mess given one.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
sanpwc: Let's say that one of futs was completed exceptionally in that case `all` will also fail and thenApply won't be called and we will basically miss map update. We either not expect futs to fail or should add corresponding exception handler. In any case silent no-op will complicate bug analysis.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    /**
+     * Gets primary replicas by partition for the table.
+     *
+     * @param tableId Table ID.
+     * @param maxStartTime Timestamp.
+     * @return Primary replicas for the table, or null when not yet known.
+     */
+    public CompletableFuture<PrimaryReplicasResult> primaryReplicasAsync(int tableId, @Nullable Long maxStartTime) {
```
sanpwc, line 103: Not sure that I completely get the idea of a method. Is it valid to get a bit outdated primaryReplicas set? Meaning the one that matches max(methodParam.maxStartTime, local.maxStartTime)? Why do you need clock.now() here?

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
```diff
@@ -0,0 +1,327 @@
...
+        HybridTimestamp timestamp = clock.now();
+
+        if (maxStartTime == null) {
+            maxStartTime = this.maxStartTime.get();
+        } else {
+            // If the client provides an old maxStartTime, ignore it and use the current one.
+            maxStartTime = Math.max(maxStartTime, this.maxStartTime.get());
+        }
+
+        // Check happy path: if we already have all replicas, and maxStartTime > timestamp, return synchronously.
```
sanpwc, line 113: `maxStartTime > timestamp` why do we need this? Seems that if methodParam.maxStartTime <= local.MaxStartTime is enough.

ptupitsyn: > `methodParam.maxStartTime <= local.MaxStartTime`

Yes, that's what we check in `primaryReplicasNoWait`. The comment was a bit misleading, I've fixed it.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerModule.java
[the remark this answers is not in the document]
ptupitsyn: Ticket created to address all client-related classes: https://issues.apache.org/jira/browse/IGNITE-20921

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java
ptupitsyn: Done - now we send `maxStartTime` timestamp to the client and back:
* Client can avoid stale updates from a lagging node
* Server will wait for a given timestamp if it lags behind

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
ptupitsyn: I'd prefer to keep it as is, since type erasure does not allow me to implement the same interface multiple times in the same class with different generic type arguments. The workaround is to put handlers into fields, but this seems like more code for no good reason.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
ptupitsyn: Good catch. Fixed.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
ptupitsyn: Fixed - `updatePrimaryReplica` now uses `leaseStartTime` to skip stale updates.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
```diff
@@ -0,0 +1,342 @@
...
+            if (existingVal != null
+                    && existingVal.leaseStartTime != null
+                    && existingVal.leaseStartTime.longValue() >= startTimeLong) {
+                return existingVal;
+            }
+
+            return new ReplicaHolder(nodeName, startTime);
+        });
+
+        while (true) {
```
sanpwc, line 293: Here and there. Seems that it's possible to use `maxStartTime.updateAndGet(value -> Math.max(value, startTimeLong));` instead.

modules/client/src/test/java/org/apache/ignite/client/fakes/FakeInternalTable.java
```diff
@@ -53,6 +53,8 @@
  * Fake internal table.
  */
 public class FakeInternalTable implements InternalTable {
+    public static final int PARTITIONS = 4;
```
sanpwc, line 56: public?

modules/table/src/main/java/org/apache/ignite/internal/table/IgniteTablesInternal.java
```diff
@@ -17,9 +17,7 @@
 
 package org.apache.ignite.internal.table;
 
-import java.util.List;
```
sanpwc, line 20: It was also expected that you'll remove `org.apache.ignite.internal.table.distributed.storage.InternalTableImpl#assignments`
Please check https://github.com/apache/ignite-3/pull/2848/files#diff-e319264cb2e94af03c766ec91510f68ba83e69cc93873b8e214e9289f52bb67a

sanpwc (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/IgniteTablesInternal.java
ptupitsyn: 1. #2848 is not yet merged
2. The method is used by `ItTableScanTest`

I'd like to avoid inflating this PR further, separate ticket created: https://issues.apache.org/jira/browse/IGNITE-20933

modules/client/src/test/java/org/apache/ignite/client/fakes/FakeInternalTable.java
ptupitsyn: Yes, this is used to initialize `FakeCatalogService` and in some other places which should have the same number of partitions to make the tests work. The idea is to express this dependency explicitly and avoid magic numbers.

modules/placement-driver-api/src/main/java/org/apache/ignite/internal/placementdriver/event/PrimaryReplicaEventParameters.java
```diff
@@ -26,18 +27,21 @@ public class PrimaryReplicaEventParameters extends CausalEventParameters {
...
     private final String leaseholder;
 
+    private final HybridTimestamp startTime;
+
     /**
      * Constructor.
      *
      * @param causalityToken Causality token.
      * @param groupId Replication group ID.
      * @param leaseholder Leaseholder node consistent ID.
```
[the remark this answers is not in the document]
ptupitsyn: Fixed.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientPrimaryReplicaTracker.java
ptupitsyn: Nice! Did not know about that. Fixed.

Should we (later) update other places like `HybridClockImpl`?

sanpwc: We either should, or maybe @denis-chudov can tell us that while(true) is faster (which is important in HybridClock and not important here).

modules/client/src/test/java/org/apache/ignite/client/fakes/FakeInternalTable.java
ptupitsyn: The getter is not `static`.

PR #2834, opened 2023-11-12
IGNITE-19218 Implement special columns query in ODBC 

isapego:
- Ported code from AI2;
- Un-commented test.


modules/platforms/cpp/ignite/odbc/query/special_columns_query.cpp
```diff
@@ -0,0 +1,91 @@
...
+sql_result special_columns_query::close()
+{
+    m_executed = false;
+
+    return sql_result::AI_SUCCESS;
+}
+
+} // namespace ignite
+
+
```
ptupitsyn, line 91: Extra space

ptupitsyn (approved):

PR #2840, opened 2023-11-15
IGNITE-20854 Configure idleSafeTimePropagationPeriod with a distributed property

rpuch:
https://issues.apache.org/jira/browse/IGNITE-20854

### The Review Checklist

### Notes

### DIFF
--- a/modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
+++ b/modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
@@ -115,6 +115,7 @@ import org.apache.ignite.internal.raft.configuration.RaftConfiguration;
 import org.apache.ignite.internal.raft.storage.impl.VolatileLogStorageFactoryCreator;
 import org.apache.ignite.internal.replicator.ReplicaManager;
 import org.apache.ignite.internal.replicator.ReplicaService;
+import org.apache.ignite.internal.replicator.configuration.ReplicationConfiguration;
 import org.apache.ignite.internal.rest.RestComponent;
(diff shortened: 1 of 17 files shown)


modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -484,7 +485,9 @@ public class IgniteImpl implements Ignite {
                 clock
         );
 
-        LongSupplier partitionIdleSafeTimePropagationPeriodMsSupplier = partitionIdleSafeTimePropagationPeriodMsSupplier();
+        ReplicationConfiguration replicationConfig = clusterConfigRegistry.getConfiguration(ReplicationConfiguration.KEY);
+
+        LongSupplier partitionIdleSafeTimePropagationPeriodMsSupplier = partitionIdleSafeTimePropagationPeriodMsSupplier(replicationConfig);
```
sashapolo, line 490: What's the point of having suppliers here?

sashapolo (approved):

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
rpuch: First it was because configuration object seems to be too complex to pass when you just need to get one property. This is arguable.

But then it turned out that it would be cool to verify from the outside (wrt configuration) that the configuration returns the same value. Also, this logic is needed in a few components, so it seems natural to have a common supplier.
We can monitor the situation and later remove the assertions, probably returning to operating vanilla configuration objects.

PR #2845, opened 2023-11-16
IGNITE-20871 Fix stack trace in client exceptions

isapego:
- Fixed stack traces;
- Added checks to existing tests to ensure we're getting proper stack traces.

### DIFF
--- a/modules/client/src/main/java/org/apache/ignite/internal/client/ClientUtils.java
+++ b/modules/client/src/main/java/org/apache/ignite/internal/client/ClientUtils.java
@@ -20,6 +20,7 @@ package org.apache.ignite.internal.client;
 import static org.apache.ignite.lang.ErrorGroups.Common.INTERNAL_ERR;
 
 import java.util.Objects;
+import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import org.apache.ignite.client.ClientOperationType;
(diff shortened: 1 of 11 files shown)


modules/client/src/main/java/org/apache/ignite/internal/client/ClientUtils.java
```diff
@@ -17,22 +17,58 @@
...
+    public static IgniteException ensurePublicException(Throwable e) {
+        Objects.requireNonNull(e);
+
+        e = ExceptionUtils.unwrapCause(e);
+
+        if (e instanceof IgniteException) {
+            IgniteException iex = (IgniteException) e;
+
+            try {
+                return ExceptionUtils.copyExceptionWithCause(e.getClass(), iex.traceId(), iex.code(), e.getMessage(), e);
```
ptupitsyn, line 54: `copyExceptionWithCause` can return null - what happens in this case?

modules/client/src/main/java/org/apache/ignite/internal/client/ClientUtils.java
```diff
@@ -17,22 +17,58 @@
...
+            } catch (Exception ex) {
+                throw new RuntimeException("IgniteException-derived class does not have required constructor: "
```
ptupitsyn, line 56: `copyExceptionWithCause` returns null when there is no constructor, or am I mistaken?

modules/client/src/main/java/org/apache/ignite/internal/client/ClientUtils.java
```diff
@@ -17,22 +17,58 @@
...
+
+        e = IgniteExceptionMapperUtil.mapToPublicException(e);
+
+        if (e instanceof IgniteCheckedException) {
+            IgniteCheckedException iex = (IgniteCheckedException) e;
+
+            return new IgniteException(iex.traceId(), iex.code(), e.getMessage(), e);
+        }
+
+        return (IgniteException) e;
```
ptupitsyn, line 69: `mapToPublicException` potentially returns `e` as is. We should always throw a new exception to ensure correct stack trace. 

By the way, do we have a test for this part of the method?

ptupitsyn (changes requested):

modules/client/src/main/java/org/apache/ignite/internal/client/ClientUtils.java
```diff
@@ -17,22 +17,61 @@
...
+            return new IgniteException(INTERNAL_ERR, "IgniteException-derived class does not have required constructor: "
+                    + e.getClass().getName(), iex);
+        }
+
+        e = IgniteExceptionMapperUtil.mapToPublicException(e);
+
+        if (e instanceof IgniteCheckedException) {
+            IgniteCheckedException iex = (IgniteCheckedException) e;
+
+            return new IgniteException(iex.traceId(), iex.code(), e.getMessage(), e);
```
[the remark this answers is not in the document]
isapego: I discussed it with @sk0x50 and we both are not sure that this should be done. Correct me if I'm wrong, @sk0x50.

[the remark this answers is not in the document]
sk0x50: I think that we need to use the same approach for handling `IgniteException` and `IgniteCheckedException` until the issue is discussed and approved on the dev list.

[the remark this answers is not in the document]
isapego: Done. Please, take another look.

ptupitsyn (approved):

modules/client/src/main/java/org/apache/ignite/internal/client/ClientUtils.java
```diff
@@ -17,22 +17,92 @@
...
+        return new IgniteException(INTERNAL_ERR, e.getMessage(), e);
+    }
+
+    /**
+     * Try to copy exception using ExceptionUtils.copyExceptionWithCause and return new exception if it was not possible.
+     *
+     * @param e Exception.
+     * @return Properly copied exception or a new error, if exception can not be copied.
+     */
+    private static Throwable copyExceptionWithCauseIfPossible(IgniteException e) {
```
sk0x50, line 72: It can be optimized a bit using the `TraceableException` interface.

sk0x50:
In general, the proposed patch is OK.

PR #2846, opened 2023-11-16
IGNITE-20463 Sql. Support for transaction control statements.

xtern:
https://issues.apache.org/jira/browse/IGNITE-20463

* Made a small refactoring based on [IGNITE-20855](https://issues.apache.org/jira/browse/IGNITE-20855) description The logic for starting transactions has been moved to so-called `QueryTransactionHandler` :thinking: .
* The type of exceptions when starting transactions has been changed from `VALIDATION` to `RUNTIME` (since everything is fine with the statement itself and this is not a validation error).
* A separate public `TxControlInsideExternalTxNotSupportedException` has been added, which should reach the jdbc user with a message about autocommit mode (something like autocommit=false doesn't support transaction control statements).
* Extracted base test class to make it easier to test queries with multiple `BaseSqlMultiStatementTest` statements.


modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
```diff
@@ -0,0 +1,219 @@
...
+        validateSingleResult(cursors.get(0));
+        validateSingleResult(cursors.get(1));
+
+        checkFinished(1);
+    }
+
+    @Test
+    void basicTxStatements() {
+        fetchAllCursors(runScript(
+                "START TRANSACTION;"
```
zstan, line 83: i`m wondering why we call InternalTableImpl#primaryReplicas for simple (check below) query 3 times ? But as i understand this need to be done only once. If this is not a part of this issue can u fill appropriate one plz ?
```
                "START TRANSACTION;"
                        + "INSERT INTO test VALUES(0);"
                        + "INSERT INTO test VALUES(1);"
                        + "COMMIT;"
```

zstan: additionally check this comment https://github.com/apache/ignite-3/pull/2848/files#r1398747886

xtern: I can see that we are calling `primaryReplicas` for each `insert` statement - two times for specified query. 
As I understand we calling it during the mapping phase.

Do you mean that we must optimize script execution to re-use somehow mapping for similar queries?

zstan: no, this optimization is beyond the scope of this issue but i see 3 times - i recheck it

xtern: I think the third one you noticed is the delete query that is called after every test, and is not part of this query.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
```diff
@@ -0,0 +1,219 @@
...
+            checkNoPendingTransactions();
+            checkFinished(1);
+
+            assertQuery("select count(id) from test")
+                    .returns(0L).check();
+        }
+    }
+
+    @Test
+    void commitWithoutOpenTransactionDoesNothing() {
```
AMashenkov, line 148: Should DDL fail transaction in the next case?
```
START TRANSACTION;
INSERT INTO test VALUES (0);
CREATE TABLE ....;
COMMIT;
```

xtern: yes

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryTransactionHandler.java
```diff
@@ -0,0 +1,192 @@
...
+        }
+    }
+
+    /**
+     * Starts an implicit transaction if there is no external transaction.
+     * Supports script transaction management using {@link SqlQueryType#TX_CONTROL} statements.
+     */
+    class ControlStatementAwareTransactionHandler extends QueryTransactionHandlerImpl {
+        private static final NoopTransactionWrapper noopTxWrapper = new NoopTransactionWrapper();
+        private volatile InternalTransaction scriptManagedTransaction;
```
AMashenkov, line 114:
```suggestion
        /** Transaction, which is managed by SQL engine via {@link SqlQueryType#TX_CONTROL} statements. */
        private volatile InternalTransaction managedTransaction;
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryTransactionHandler.java
```diff
@@ -0,0 +1,192 @@
...
+import org.apache.ignite.sql.SqlException;
+import org.apache.ignite.tx.IgniteTransactions;
+import org.apache.ignite.tx.TransactionOptions;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * The query transaction handler is responsible for running implicit or script managed transactions during query execution.
+ */
+@SuppressWarnings("InterfaceMayBeAnnotatedFunctional")
+public interface QueryTransactionHandler {
```
AMashenkov, line 38: Maybe it would better have QueryTransactionHandlerImpl and ScriptTransactionHandlerImpl,
and factory methods with similar names.
As for now, methods and classes have different names and introduces to much terms...

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/QueryTransactionWrapperSelfTest.java
```diff
@@ -91,27 +97,51 @@ public void testCommitImplicit() {
...
 
         wrapper.rollback();
 
         assertThat(tx.rollbackFuture().isDone(), equalTo(true));
         assertThat(tx.commitFuture().isDone(), equalTo(false));
     }
 
-    private QueryTransactionWrapper prepareImplicitTx() {
-        when(transactions.begin(any())).thenReturn(new NoOpTransaction("test"));
+    @Test
```
AMashenkov, line 109: Let's add a test calling `startTxIfNeeded` twice, which should be impossible, because it semantic of nested transactions.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryTransactionHandler.java
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        @Override
+        public QueryTransactionWrapper startTxIfNeeded(ParsedResult parsedResult) {
+            try {
+                SqlQueryType queryType = parsedResult.queryType();
+
+                if (queryType == SqlQueryType.TX_CONTROL) {
+                    if (externalTransaction != null) {
+                        throw new ExternalTransactionNotSupportedException();
```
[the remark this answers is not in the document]
xtern: Actually, this needs some clarification.
We don’t have a native api at all to pass an external transaction. This case is only possible in jdbc.
And the point here is not that we launched a nested transaction, but that our jdbc non-autocommit mode implementation cannot be integrated with a transaction created in the script.
For example:
```
connection.setAutoCommit(false)
...
execute("INSERT...");
execute("SELECT...");
connection.commit();
```
Jdbc handler implicitly starts `external` transaction and commits it when user call `commit`.
This is ok, but what with the following
```
connection.setAutoCommit(false)

execute("START TRANSACTION"
+ "INSERT..."
+ "COMMIT; 
+ "START TRANSACTION"
+ "INSERT...

connection.commit()
```
Our jdbc implementation is not aware at all about transaction started in script, so user cannot commit it by calling "connection.commit()" 

So for the consistency "START TRANSACTION/COMMIT" are forbidden for autocommit=false jdbc mode.

That's way we must throw this exception even for single commit statement, for example:

```
queryScript(...tx, "COMMIT;")
```
In this case, we are not talking about an nested transaction.
```

File: modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTest.java [hunk 7]
```diff
@@ -24,42 +24,30 @@
...
+public class ItSqlMultiStatementTest extends BaseSqlMultiStatementTest {
     @AfterEach
     void dropTables() {
         dropAllTables();
     }
 
-    @AfterEach
-    void checkNoPendingTransactions() {
-        assertEquals(0, txManager().pending());
-    }
```
zstan, line 61: not related to this part but all statements below need to be .ordered()

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTest.java
```diff
@@ -138,131 +126,76 @@ void queryWithIncorrectNumberOfDynamicParametersFailsWithValidationError() {
...
-        );
-        assertNotNull(cursor);
-        assertTrue(cursor.hasNextResult());
+        {
+            AsyncSqlCursor<List<Object>> cursor = runScript(
+                    "CREATE TABLE test (id INT PRIMARY KEY);"
+                            + "SELECT 2/0;" // Runtime error.
+                            + "INSERT INTO test VALUES (0)"
+            );
+            assertNotNull(cursor);
```
zstan, line 138: we already have notNull check inside BaseSqlMultiStatementTest#runScript() i suppose this check is redundant. Also do we need to check here if empty table is created ?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
```diff
@@ -0,0 +1,219 @@
...
+            fetchAllCursors(runScript("START TRANSACTION;"
+                    + "INSERT INTO test VALUES(0);"
+                    + "INSERT INTO test VALUES(1);"
+                    + "COMMIT;"
+                    + "START TRANSACTION;"
+                    + "INSERT INTO test VALUES(2);"));
+
+            checkFinished(3);
+
+            checkNoPendingTransactions();
```
zstan, line 120: seems we need documentation issue here, isn`t it ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryTransactionHandler.java
xtern: Done

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/QueryTransactionWrapperSelfTest.java
xtern: Done

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryTransactionHandler.java
xtern: Done

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -667,13 +652,15 @@ CompletableFuture<AsyncSqlCursor<List<Object>>> processNext() {
 
                             txWrapper.commitImplicit();
 
-                            if (nextCursorFuture != null) {
+                            if (parameters.nextStatementFuture == null) {
```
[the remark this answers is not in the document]
xtern: `parameters` is a local variable
Are you suggesting to add `isLastStatement` method to `ScriptStatementParameters`?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -615,49 +586,63 @@ private static void validateDynamicParameters(int expectedParamsCount, Object[]
...
+            // We fill parameters in reverse order, because each script statement
+            // requires a reference to the future of the next statement.
+            for (int i = parsedResults0.size(); i > 0; i--) {
+                ParsedResult result = parsedResults0.get(i - 1);
+
+                Object[] params0 = Arrays.copyOfRange(params, paramsCount - result.dynamicParamsCount(), paramsCount);
+                paramsCount -= result.dynamicParamsCount();
+
+                results[i - 1] = new ScriptStatementParameters(result, params0,
+                        i < parsedResults0.size() ? results[i].cursorFuture : null);
```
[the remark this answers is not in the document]
xtern: Done, thanks.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTest.java
xtern: Thanks. Removed all redundant null-checks that I found

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
xtern: `ddlInsideExplicitTransactionFails` covers this case

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/QueryTransactionWrapperSelfTest.java
xtern: Added test `throwsExceptionForTxControlStatementInsideExternalTransaction`.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
xtern: Yes, definitely :sunglasses:

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/tx/ScriptTransactionHandler.java
```diff
@@ -0,0 +1,319 @@
...
+        ManagedTransactionWrapper txWrapper = this.wrapper;
+
+        if (node instanceof IgniteSqlCommitTransaction) {
+            if (txWrapper == null) {
+                return NOOP_TX_WRAPPER;
+            }
+
+            this.wrapper = null;
+
+            return txWrapper.forCommit();
```
zstan, line 121: looking at this place i expected test will pass, but it didn\`t, i\`m wrong ? i.e. i see that RO transaction already commited but we still have partially readed cursors.
```
    @Test
    void readMoreRowsThanCanBePrefetchedReadOnlyTx() {
        String query = "START TRANSACTION READ ONLY;"
                + "SELECT id FROM big;"
                + "SELECT 1;"
                + "COMMIT;";

        List<AsyncSqlCursor<List<Object>>> cursors = fetchAllCursors(runScript(query));
        assertThat(cursors, hasSize(4));
        verifyFinishedTxCount(0);

        BatchedResult<List<Object>> res = await(cursors.get(2).requestNextAsync(1));
        assertNotNull(res);
        verifyFinishedTxCount(0);

        res = await(cursors.get(2).requestNextAsync(100));
        assertNotNull(res);
        verifyFinishedTxCount(0);

        res = await(cursors.get(1).requestNextAsync(1));
        assertNotNull(res);
        verifyFinishedTxCount(0);

        cursors.forEach(AsyncSqlCursor::closeAsync);
    }
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
zstan: plz link appropriate issue if it already done ?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/BaseSqlMultiStatementTest.java
```diff
@@ -0,0 +1,104 @@
...
+public abstract class BaseSqlMultiStatementTest extends BaseSqlIntegrationTest {
+    @AfterEach
+    protected void checkNoPendingTransactionsAndOpenedCursors() {
+        assertEquals(0, txManager().pending());
+
+        try {
+            boolean success = waitForCondition(() -> queryProcessor().openedCursors() == 0, 5_000);
+
+            if (!success) {
+                assertEquals(0, queryProcessor().openedCursors());
```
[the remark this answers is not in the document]
xtern: From my point of view it's better to show the number of open cursors in error.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTest.java
```diff
@@ -24,86 +24,77 @@
...
         assertNotNull(curItr.next()); // skip EXPLAIN.
-        validateSingleResult(curItr.next(), 0, 0);
+        validateSingleResult(curItr.next(), 0, 1);
+        validateSingleResult(curItr.next(), 1L);
 
         assertFalse(curItr.hasNext());
 
+        cursors.forEach(AsyncSqlCursor::closeAsync);
+
         // Ensures that the script is executed completely, even if the cursor data has not been read.
```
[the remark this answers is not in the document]
xtern: `fetchAllCursors` method does not fetches data from cursors.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTest.java
```diff
@@ -117,13 +108,16 @@ void queryWithDynamicParameters() {
                 + "INSERT INTO test VALUES(?, DEFAULT);"
                 + "INSERT INTO test VALUES(?, ?);";
 
-        fetchAllCursors(runScript(sql, null, 0, "1", 2, 4, "5"));
+        List<AsyncSqlCursor<List<Object>>> cursors =
```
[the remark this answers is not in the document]
xtern: Not only for jdbc, we have public API for this.
If you are asking about dynamic parameters - they are used from the public API.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
```diff
@@ -0,0 +1,301 @@
...
+    void createTable() {
+        sql("CREATE TABLE test (id INT PRIMARY KEY)");
+
+        createTable("big", 1, 1);
+        sql("INSERT INTO big (id) SELECT x FROM TABLE(SYSTEM_RANGE(1, " + BIG_TABLE_ROWS_COUNT + "));");
+    }
+
+    @AfterEach
+    void clearTestTable() {
+        sql("DELETE FROM test");
```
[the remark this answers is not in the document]
xtern: On startup we populate only table named `BIG`. This table doesn't cleared after test. But I agree it looks a bit confusing, my idea is to just separate the different test cases but keep running the tests quickly.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
```diff
@@ -0,0 +1,301 @@
...
+        assertNotNull(res);
+        assertThat(res.items(), hasSize(BIG_TABLE_ROWS_COUNT));
+
+        verifyFinishedTxCount(1);
+
+        cursors.forEach(AsyncSqlCursor::closeAsync);
+    }
+
+    @Test
+    void readMoreRowsThanCanBePrefetchedReadWriteTx() {
```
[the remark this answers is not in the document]
xtern: Done, thanks.
Actually, the initial version of this test looked exactly as you wrote) But after a while it seemed to me that the second cursor was redundant.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
```diff
@@ -0,0 +1,301 @@
...
+            verifyFinishedTxCount(1);
+        }
+
+        {
+            List<AsyncSqlCursor<List<Object>>> cursors = fetchAllCursors(
+                    runScript("START TRANSACTION;"
+                            + "INSERT INTO test VALUES(0);"
+                            + "INSERT INTO test VALUES(1);"
+                            + "COMMIT;"
+
```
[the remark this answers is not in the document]
xtern: We already have a test for this case - IgniteSqlParserTest.testScriptMode (see also testEmptyStatements). If you think this is not enough, we can expand these tests to other cases.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
```diff
@@ -0,0 +1,301 @@
...
+            verifyFinishedTxCount(3);
+
+            assertThat(cursors, hasSize(7));
+            cursors.subList(0, 4).forEach(AsyncSqlCursor::closeAsync);
+
+            // Cursors associated with a transaction must be closed when the transaction is rolled back.
+            checkNoPendingTransactionsAndOpenedCursors();
+
+            assertQuery("select count(id) from test")
+                    .returns(2L).check();
```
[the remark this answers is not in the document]
xtern: we already have in test
```
checkNoPendingTransactionsAndOpenedCursors();
```
Those. either all cursors will be closed or this check will fail. Why do we need to re-close them?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
```diff
@@ -0,0 +1,301 @@
...
+        verifyFinishedTxCount(2);
+    }
+
+    private void verifyFinishedTxCount(int expected) {
+        int expectedTotal = numberOfFinishedTransactionsOnStart + expected;
+
+        boolean success;
+
+        try {
+            success = waitForCondition(() -> expectedTotal == txManager().finished(), 2_000);
```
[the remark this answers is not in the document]
xtern: From my point of view, it is always easier to diagnose error messages containing specific numbers (expected vs actual).

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/tx/ScriptTransactionHandler.java
xtern: We have decided that in the current implementation, only the RW transaction have limitation - that all cursors must be closed.  Because our current RO transactions are just a timestamp, and it looks like we don't need to block script execution with the RO transaction.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
```diff
@@ -0,0 +1,301 @@
...
+        {
+            AsyncSqlCursor<List<Object>> cursor = runScript(
+                    "START TRANSACTION READ WRITE;"
+                    + "INSERT INTO test VALUES(2);"
+                    + "INSERT INTO test VALUES(2/0);"
+                    + "SELECT 1;"
+                    + "COMMIT;"
+            );
+
+            assertThrowsSqlException(RUNTIME_ERR, "/ by zero", () -> fetchAllCursors(cursor));
```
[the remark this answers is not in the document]
xtern: done :thinking:  
https://issues.apache.org/jira/browse/IGNITE-20997

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/tx/ImplicitTransactionWrapper.java
```diff
@@ -24,38 +24,32 @@
 /**
  * Wrapper for the transaction that encapsulates the management of an implicit transaction.
  */
-public class QueryTransactionWrapper {
+public class ImplicitTransactionWrapper implements QueryTransactionWrapper {
```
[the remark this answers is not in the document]
xtern: renamed to `QueryTransactionWrapperImpl` :thinking:

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/tx/ScriptTransactionHandler.java
```diff
@@ -0,0 +1,319 @@
...
+ * {@link SqlQueryType#TX_CONTROL} statements.
+ */
+public class ScriptTransactionHandler extends QueryTransactionHandler {
+    /** No-op transaction wrapper. */
+    private static final QueryTransactionWrapper NOOP_TX_WRAPPER = new NoopTransactionWrapper();
+
+    /** Wraps a transaction, which is managed by SQL engine via {@link SqlQueryType#TX_CONTROL} statements. */
+    private volatile @Nullable ManagedTransactionWrapper wrapper;
+
+    public ScriptTransactionHandler(IgniteTransactions transactions, @Nullable InternalTransaction externalTransaction) {
```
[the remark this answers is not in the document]
xtern: renamed to `outerTransaction` :thinking:

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
```diff
@@ -0,0 +1,301 @@
...
+        assertFalse(selectCursor.nextResult().isDone());
+
+        await(selectCursor.requestNextAsync(BIG_TABLE_ROWS_COUNT)); // Cursor must close implicitly.
+        verifyFinishedTxCount(1);
+    }
+
+    @Test
+    void openedScriptTransactionRollsBackImplicitly() {
+        {
+            fetchAllCursors(runScript("START TRANSACTION;"));
```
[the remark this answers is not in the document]
xtern: We have test `commitWithoutOpenTransactionDoesNothing` that checks commit without tx.
We have test `emptyTransactionControlStatement` that checks `start; commit;` (validates cursor data, as you suggest here).
We have this test that verifies that the transaction will be rolled back in case of user input error. 
What kind of test you want to add?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/tx/ScriptTransactionHandler.java
zstan: ok, got it

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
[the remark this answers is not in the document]
zstan: oh, got it !

modules/runner/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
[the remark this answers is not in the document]
xtern: This test has been reworked a bit and updated.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/tx/QueryTransactionHandler.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.ignite.tx.IgniteTransactions;
+import org.apache.ignite.tx.TransactionOptions;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Starts an implicit transaction if there is no external transaction.
+ */
+public class QueryTransactionHandler {
+    final IgniteTransactions transactions;
+    final @Nullable InternalTransaction outerTransaction;
```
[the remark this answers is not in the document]
xtern: According to our discussion - renamed to `externalTx`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/tx/QueryTransactionHandler.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public QueryTransactionHandler(IgniteTransactions transactions, @Nullable InternalTransaction outerTransaction) {
```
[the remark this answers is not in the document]
xtern: According to our discussion - renamed to `externalTx`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/tx/QueryTransactionHandler.java
```diff
@@ -0,0 +1,74 @@
...
+                    new TransactionOptions().readOnly(queryType != SqlQueryType.DML)), true);
+        }
+
+        validateStatement(queryType, activeTx);
+
+        return new QueryTransactionWrapperImpl(activeTx, false);
+    }
+
+    protected @Nullable InternalTransaction activeTransaction() {
+        return outerTransaction;
```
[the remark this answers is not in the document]
xtern: According to our discussion - renamed to `externalTx`

modules/sql-engine/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
```diff
@@ -0,0 +1,344 @@
...
+            // The transaction depends on the cursors of the SELECT statement,
+            // so it waits for them to close.
+            assertEquals(1, txManager().pending());
+
+            cursors.forEach(AsyncSqlCursor::closeAsync);
+            verifyFinishedTxCount(2);
+        }
+
+        {
+            AsyncSqlCursor<List<Object>> cur = runScript("START TRANSACTION READ WRITE;"
```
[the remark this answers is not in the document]
xtern: thanks, extracted this test case to test `dmlScriptRollsBackImplicitly`

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/AsyncSqlCursorImplTest.java
```diff
@@ -51,7 +52,7 @@ public class AsyncSqlCursorImplTest {
     /** Cursor should trigger commit of implicit transaction (if any) only if data is fully read. */
     @ParameterizedTest(name = "{0}")
     @MethodSource("transactions")
-    public void testTriggerCommitAfterDataIsFullyRead(boolean implicit, QueryTransactionWrapper txWrapper) {
+    public void testTriggerCommitAfterDataIsFullyRead(boolean implicit, QueryTransactionWrapperImpl txWrapper) {
```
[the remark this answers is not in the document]
xtern: Fixed, thanks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -681,66 +661,92 @@ private static void validateDynamicParameters(int expectedParamsCount, Object[]
...
         }
 
-        CompletableFuture<AsyncSqlCursor<List<Object>>> processNext() {
-            if (statements == null) {
-                // TODO https://issues.apache.org/jira/browse/IGNITE-20463 Each tx control statement must return an empty cursor.
-                return CompletableFuture.completedFuture(null);
+        /**
+         * Returns a queue. each element of which represents parameters required to execute a single statement of the script.
+         */
+        private Queue<ScriptStatementParameters> prepareStatementsQueue(List<ParsedResult> parsedResults0, Object[] params) {
```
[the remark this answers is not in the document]
xtern: Fixed, thanks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/tx/ScriptTransactionWrapper.java
```diff
@@ -0,0 +1,178 @@
...
+        return transaction;
+    }
+
+    @Override
+    public CompletableFuture<Void> commitImplicit() {
+        return Commons.completedFuture();
+    }
+
+    @Override
+    public CompletableFuture<Void> rollback(Throwable cause) {
```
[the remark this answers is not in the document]
xtern: > i see that Throwable cause is ignored in more cases, is it ok ?

In general, I added this exception only so as not to introduce another method to roll back the transaction when the script ends :pensive: .
If an exception is passed, it means there was an error and we must close the associated cursors.
If an exception is not passed, then there was no error and we are trying to roll back the transaction opened in the script, in this case we should wait until the user subtracts all the cursors before rollback transaction.

Those. it's better to add a separate method. But I’m still stuck with its name. "tryRollbackScript" maybe or just "rollbackScript"?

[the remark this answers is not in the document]
xtern: > Also i found that sql parsing errors are not informative,
> it would be hard to found a problem if script will consist from numerous of lines

I don't quite understand, the parser exception message contains the line number and position, what needs to be improved?

zstan (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/AsyncSqlCursorImpl.java
```diff
@@ -156,8 +157,10 @@ public CompletableFuture<Void> closeAsync() {
 
         dataCursor.closeAsync()
                 .thenCompose(ignored -> txWrapper.commitImplicit())
-                .thenRun(onClose)
                 .whenComplete((r, e) -> {
+                    // Anyway run close handler.
+                    onClose.run();
```
AMashenkov, line 162: Should we catch and log an exception from `onClose` ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/tx/ScriptTransactionHandler.java
```diff
@@ -0,0 +1,153 @@
...
+ * {@link SqlQueryType#TX_CONTROL} statements.
+ */
+public class ScriptTransactionHandler extends QueryTransactionHandler {
+    /** No-op transaction wrapper. */
+    private static final QueryTransactionWrapper NOOP_TX_WRAPPER = new NoopTransactionWrapper();
+
+    /** Wraps a transaction, which is managed by SQL engine via {@link SqlQueryType#TX_CONTROL} statements. */
+    private volatile @Nullable ScriptTransactionWrapper wrapper;
+
+    public ScriptTransactionHandler(IgniteTransactions transactions, @Nullable InternalTransaction externalTx) {
```
AMashenkov, line 49: ``` @Nullable InternalTransaction externalTx ```
😃

modules/sql-engine/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTest.java
```diff
@@ -138,131 +128,77 @@ void queryWithIncorrectNumberOfDynamicParametersFailsWithValidationError() {
         assertThrowsSqlException(STMT_VALIDATION_ERR, expectedMessage, () -> runScript(sql, null, 0, 1, 2, 3, 4, 5));
     }
 
-    @Test
-    void transactionControlStatementDoesNotCreateCursor() {
```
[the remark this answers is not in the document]
xtern: This is already done. This test has been moved to TX-related tests and called "emptyTransactionControlStatement(){"

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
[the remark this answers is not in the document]
AMashenkov: I'm just confused, the "parameters" has a future for the "next statement"...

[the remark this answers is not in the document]
xtern: Thanks, reworked a bit. Added `isLastStatement()`  method.

modules/api/src/main/java/org/apache/ignite/sql/ExternalTransactionNotSupportedException.java
```diff
@@ -0,0 +1,42 @@
...
+ * The exception is thrown when an attempt is made to execute a transaction control statement
+ * ({@code START TRANSACTION} or {@code COMMIT}) while an external transaction is already running.
+ */
+public class ExternalTransactionNotSupportedException extends SqlException {
+    /**
+     * Default constructor.
+     */
+    public ExternalTransactionNotSupportedException() {
+        super(Sql.TX_CONTROL_INSIDE_EXTERNAL_TX_ERR, "Transaction control statement cannot be executed within an external transaction.");
+    }
```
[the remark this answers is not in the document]
xtern: Moved to internal errors and renamed to `TxControlInsideExternalTxNotSupportedException` :thinking: .

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/AsyncSqlCursorImpl.java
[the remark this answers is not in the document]
xtern: Thanks, fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/tx/ScriptTransactionHandler.java
xtern: Thanks, renamed `externalTx` => `transaction`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/tx/ScriptTransactionWrapper.java
```diff
@@ -0,0 +1,179 @@
...
+
+    @Override
+    public CompletableFuture<Void> commitImplicit() {
+        return nullCompletedFuture();
+    }
+
+    @Override
+    public CompletableFuture<Void> rollback(Throwable cause) {
+        // Close all associated cursors on error.
+        for (CompletableFuture<? extends AsyncCursor<?>> fut : openedCursors.values()) {
```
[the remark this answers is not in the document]
xtern: Thanks, added synchronization between creating new statement and rolling back on error.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
```diff
@@ -543,55 +542,77 @@ List<AsyncSqlCursor<?>> openedCursors() {
...
+                }
+
+                if (!busyLock.enterBusy()) {
+                    completeScriptExecution(sessionIsClosedException());
+                    return;
+                }
+
+                try {
+                    if (cursor.hasNextResult()) {
+                        cursor.nextResult().whenComplete(this::processCursor);
```
AMashenkov, line 574: Let's avoid synchronous continuation execution under busylock.
```suggestion
                        cursor.nextResult().whenCompleteAsync(this::processCursor, pool);
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
```diff
@@ -543,55 +542,77 @@ List<AsyncSqlCursor<?>> openedCursors() {
...
-                        if (nextCursor != null) {
-                            int nextCursorId = registerCursor(nextCursor);
-                            processCursor(nextCursor, nextCursorId);
-                        } else {
-                            resFut.completeExceptionally(t);
-                        }
-                    });
-                } else {
-                    closeCursor(cursor, cursorId);
+            resFut.completeExceptionally(err);
```
AMashenkov, line 599: Is it ok to complete future under the busyLock?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
```diff
@@ -543,55 +542,77 @@ List<AsyncSqlCursor<?>> openedCursors() {
...
         }
 
-        void processCursor(AsyncSqlCursor<InternalSqlRow> cursor, int cursorId) {
-            if (!busyLock.enterBusy()) {
-                closeCursor(cursor, cursorId);
+        private void completeScriptExecution(@Nullable Throwable err) {
+            Throwable savedError = cursorCloseErrHolder.get();
+
+            if (err == null && savedError == null) {
+                resFut.complete(null);
```
AMashenkov, line 588: Is it ok to complete the future synchronously under the busylock?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -682,66 +661,98 @@ private static void validateDynamicParameters(int expectedParamsCount, Object[]
...
+                ParsedResult parsedResult = scriptStatement.parsedResult;
 
-                QueryCancel cancel = new QueryCancel();
+                QueryTransactionWrapper txWrapper = scriptTxCtx.getOrStartImplicit(parsedResult.queryType());
 
-                executeParsedStatement(schemaName, parsedResult, txWrapper, cancel, dynamicParams, true, nextCursorFuture)
-                        .whenComplete((res, ex) -> {
+                scriptTxCtx.registerCursor(parsedResult.queryType(), cursorFuture);
+
+                executeStatement(parsedResult, txWrapper, scriptStatement.dynamicParams, scriptStatement.nextStatementFuture)
```
AMashenkov, line 732: If an implicit transaction was started above in 728, will control statement fails here?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
xtern: Can you suggest what pool should be used? Currently I decided to keep system pool.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
xtern: Reworked a bit - future that returns to the user completes asynchronously.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
xtern: I don't quite understand the case you describe.

The implicit transaction at line 728 can only be started if this is NOT a control statement. Otherwise - even we fail on line 728 (nested tx) or we will start/use script-driven transaction.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
```diff
@@ -543,55 +548,77 @@ List<AsyncSqlCursor<?>> openedCursors() {
...
-        void processFirstResult(AsyncSqlCursor<InternalSqlRow> cursor, Throwable t) {
-            if (t != null) {
-                resFut.completeExceptionally(t);
-            } else {
-                int cursorId = registerCursor(cursor);
-                processCursor(cursor, cursorId);
+        void processCursor(AsyncSqlCursor<InternalSqlRow> cursor, Throwable scriptError) {
+            if (scriptError != null) {
+                // Stop script execution.
+                completeScriptExecution(scriptError);
```
AMashenkov, line 562:
```suggestion
                saveError(scripError)
                completeScript();
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
```diff
@@ -543,55 +548,77 @@ List<AsyncSqlCursor<?>> openedCursors() {
...
+                if (!busyLock.enterBusy()) {
+                    completeScriptExecution(sessionIsClosedException());
+                    return;
+                }
+
+                try {
+                    if (cursor.hasNextResult()) {
+                        cursor.nextResult().whenCompleteAsync(this::processCursor);
+                    } else {
+                        completeScriptExecution(null);
```
AMashenkov, line 582:
```suggestion
                        completeScript();
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
```diff
@@ -543,55 +548,77 @@ List<AsyncSqlCursor<?>> openedCursors() {
     }
 
     private class ScriptHandler {
+        private final CompletableFuture<Void> resFut = new CompletableFuture<>();
 
-        private final CompletableFuture<Void> resFut;
+        private final AtomicReference<Throwable> cursorCloseErrHolder = new AtomicReference<>();
```
AMashenkov, line 553:
```suggestion
        private final List<Throwable> cursorCloseErrHolder = Coliections.synchronizedList(new ArrayList<>());
```

xtern: Done, thanks

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
```diff
@@ -543,55 +549,60 @@ List<AsyncSqlCursor<?>> openedCursors() {
...
-                    resFut.complete(null);
+                try {
+                    if (cursor.hasNextResult()) {
+                        cursor.nextResult().whenCompleteAsync(this::processCursor);
+                    } else {
+                        onComplete();
+                    }
+                } finally {
+                    busyLock.leaveBusy();
                 }
```
AMashenkov, line 586:
```suggestion
                    if (cursor.hasNextResult()) {
                        cursor.nextResult().whenCompleteAsync(this::processCursor);
                        return;
                    }              
                } finally {
                    busyLock.leaveBusy();
                }
                
                onComplete();
```

AMashenkov (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/SessionImpl.java
xtern: Thanks, With this improvement, we don't complete the future inside busy lock, so I removed redundant moving completion of the user future to another thread.

I mean we don't need this async completion now:
```
handler.resultFuture().whenCompleteAsync((res, th) -> {
            if (th == null) {
                userFut.complete(null);
            } else {
                Throwable cause = ExceptionUtils.unwrapCause(th);

                userFut.completeExceptionally(mapToPublicSqlException(cause));
            }
        });
```

PR #2848, opened 2023-11-17
IGNITE-19619 Sql. Term-bases to lease-based switch.

zstan:

### The Review Checklist

### Notes


zstan:
@vldpyatkov can you make a partial review plz ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -317,6 +334,60 @@ public CompletableFuture<ExecutionTarget> forSystemView(ExecutionTargetFactory f
         services.forEach(LifecycleAware::start);
     }
 
+    /** Get primary replicas. */
+    private CompletableFuture<List<PrimaryReplica>> primaryReplicas(int tableId) {
```
sanpwc, line 338: Could you please clarify when given method is actually used?
Let's say that we have following flow:
```
tx1 = transactions.begin();
insert into t1 values(k1,v1);
insert into t1 values(k2, v2);
tx1.commit();
```
When will the primaryReplcias() method be called? How many times? How do you manage previously retrieved primary replicas? Especially when it was changes (hope that you won't silently overwrite it in the inner primary replicas cache)? 

What I mean here, is that from transactions point of view it's only required to await primary replicas on **first** only.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -317,6 +334,60 @@ public CompletableFuture<ExecutionTarget> forSystemView(ExecutionTargetFactory f
...
+                    partGroupId,
+                    clockNow,
+                    AWAIT_PRIMARY_REPLICA_TIMEOUT,
+                    SECONDS
+            );
+
+            result.add(f.thenApply(primaryReplica -> {
+                String holder = primaryReplica.getLeaseholder();
+
+                if (holder == null) {
```
sanpwc, line 367: I don't think it's actually possible. In case of awaitPrimaryReplica you will either get valid !null one or PrimaryReplicaAwaitException will be thrown. So, I'd rather put an assert here.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -317,6 +334,60 @@ public CompletableFuture<ExecutionTarget> forSystemView(ExecutionTargetFactory f
...
+                    throw new IgniteInternalException(Sql.MAPPING_ERR, "Unable to map query, node is lost or offline");
+                }
+
+                return new PrimaryReplica(node, primaryReplica.getStartTime().longValue());
+            }));
+        }
+
+        CompletableFuture<Void> all = CompletableFuture.allOf(result.toArray(new CompletableFuture[0]));
+
+        return all.thenApply(v -> result.stream()
```
sanpwc, line 385: How do you handle PrimaryReplicaAwaitException or any other exceptions that might be thrown from awaitPrimaryReplica?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
```diff
@@ -643,12 +644,10 @@ public void testMvScan(boolean readOnly) throws Exception {
             Publisher<BinaryRow> publisher;
 
             if (readOnly) {
-                List<String> assignments = internalTable.assignments();
-
                 // Any node from assignments will do it.
-                ClusterNode node0 = CLUSTER.aliveNode().clusterNodes().stream().filter(clusterNode -> {
-                    return assignments.contains(clusterNode.name());
-                }).findFirst().orElseThrow();
+                ClusterNode node0 = internalTable.leaderAssignment(ThreadLocalRandom.current().nextInt(internalTable.partitions()));
```
sanpwc, line 648: leaderAssignment breaks the encapsulation and should be dropped (not by you and not within given ticket of course), why do you need it?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -317,6 +334,60 @@ public CompletableFuture<ExecutionTarget> forSystemView(ExecutionTargetFactory f
...
+                }
+
+                ClusterNode node = clusterSrvc.topologyService().getByConsistentId(holder);
+
+                if (node == null) {
+                    // additional recovery logic is need to be present around here.
+                    throw new IgniteInternalException(Sql.MAPPING_ERR, "Unable to map query, node is lost or offline");
+                }
+
+                return new PrimaryReplica(node, primaryReplica.getStartTime().longValue());
```
sanpwc, line 379: Seems that there are similar classes with exact the same semantics e.g `PrimaryReplica` and `NodeWithTerm`. Do we really need all of them?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -317,6 +334,60 @@ public CompletableFuture<ExecutionTarget> forSystemView(ExecutionTargetFactory f
...
+            }));
+        }
+
+        CompletableFuture<Void> all = CompletableFuture.allOf(result.toArray(new CompletableFuture[0]));
```
sanpwc, line 383:
```
                        tx.enlist(new TablePartitionId(tableId, p),
                                new IgniteBiTuple<>(topSrvc.getByConsistentId(leaderWithTerm.name()), leaderWithTerm.term()));
```
First of all, why don't you just await primary replica here? (Or use one if it was already enlisted).
And why it's called `leaderWithTerm`?

- Please use enlistmentConsistencyToken (or just enlistmentToken) instead of term within SQL. I'll rename it in other places under https://issues.apache.org/jira/browse/IGNITE-20377
- And please don't use `leader` because it's not leader any longer.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
zstan: cause i found that InternalTable#assignments (which marked under this issue too) is used from 2 places, one is that test and other - from client code which will be fixed by another issue, if this method is deprecated too can u suggest correct approach here ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
zstan: this code will be called on request mapping phase, 2 times (for now). There is no additional nodes caching are present now. And seems described case is not  case for this issue, i fill additional comment here https://github.com/apache/ignite-3/pull/2846/files#r1399126739

sanpwc: Well, I'm not sure whether this should be fixed within given issue or another one. The main point is that, there's no sense to awaitPrimary on same partitions touch or it may even lead to inappropriate group re-enlistment.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
sanpwc: According to the comment `// Any node from assignments will do it.` there's no need to retrieve leader, just take the first one, or random one - up to you.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
zstan: I fill the issue:
https://issues.apache.org/jira/browse/IGNITE-20899

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
```diff
@@ -872,21 +871,21 @@ private static Row createKeyRow(int id) {
...
         InternalTransaction tx = (InternalTransaction) ignite.transactions().begin(new TransactionOptions().readOnly(readOnly));
 
         InternalTable table = ((TableViewInternal) ignite.tables().table(TABLE_NAME)).internalTable();
         TablePartitionId tblPartId = new TablePartitionId(table.tableId(), partId);
 
-        PlacementDriver placementDriver = ((IgniteImpl) ignite).placementDriver();
+        PlacementDriver placementDriver = (ignite).placementDriver();
         ReplicaMeta primaryReplica = IgniteTestUtils.await(
-                placementDriver.awaitPrimaryReplica(tblPartId, ((IgniteImpl) ignite).clock().now(), 30, TimeUnit.SECONDS));
+                placementDriver.awaitPrimaryReplica(tblPartId, (ignite).clock().now(), 30, TimeUnit.SECONDS));
```
lowka, line 883: > (ignite)
I think these parentheses are redundant.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
zstan: > why don't you just await primary replica here
cause **assignments** are also used for colocation detection logic (i.e. sql through tables with different affinity distributions) thus it just reused here.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
zstan: no how for now, i have a discussion and seems that issues with unstable topology is a part of another activity.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
zstan: done thanks !

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
sanpwc: Ok, in that case let's add todo in order not to forget it.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
sanpwc: Could you please elaborate the point? Would the code be simpler if instead of propagating primaries from collocation to enlistment it will request primaries twice?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItTableScanTest.java
zstan: @sanpwc ok i change it

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
zstan: i append exception handling there

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -858,10 +858,14 @@ private void enlist(int tableId, List<NodeWithTerm> assignments) {
                     tx.assignCommitPartition(new TablePartitionId(tableId, ThreadLocalRandom.current().nextInt(partsCnt)));
 
                     for (int p = 0; p < partsCnt; p++) {
-                        NodeWithTerm enlistmentToken = assignments.get(p);
+                        TablePartitionId tablePartId = new TablePartitionId(tableId, p);
 
-                        tx.enlist(new TablePartitionId(tableId, p),
-                                new IgniteBiTuple<>(topSrvc.getByConsistentId(enlistmentToken.name()), enlistmentToken.term()));
+                        if (tx.enlistedNodeAndTerm(tablePartId) == null) {
```
sanpwc, line 863: It's not that was expected.
First of all proposed solution isn't thread safe. However it doesn't matter because enlist() itself won't re-enlist the partition. The main point here was in a fact that we shouldn't re-await primary replicas if they were already enlisted.

sanpwc (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -317,6 +341,69 @@ public CompletableFuture<ExecutionTarget> forSystemView(ExecutionTargetFactory f
...
+    private CompletableFuture<List<PrimaryReplica>> primaryReplicas(int tableId) {
+        int catalogVersion = catalogManager.latestCatalogVersion();
+
+        Catalog catalog = catalogManager.catalog(catalogVersion);
+
+        CatalogTableDescriptor tblDesc = Objects.requireNonNull(catalog.table(tableId), "table");
+
+        CatalogZoneDescriptor zoneDesc = Objects.requireNonNull(catalog.zone(tblDesc.zoneId()), "zone");
+
+        int partitions = zoneDesc.partitions();
```
korlov42, line 354: does it make sense to put partition count into IgniteTable to avoid going to catalog all the time?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -1472,7 +1473,7 @@ Int2ObjectMap<RowBatch> toRowBatchByPartitionId(Collection<BinaryRowEx> rows) {
     }
 
     /** {@inheritDoc} */
-    // TODO: https://issues.apache.org/jira/browse/IGNITE-19619 The method should be removed, SQL engine should use placementDriver directly
+    // TODO: https://issues.apache.org/jira/browse/IGNITE-20701 The method should be removed
```
ptupitsyn, line 1476:
```suggestion
    // TODO: https://issues.apache.org/jira/browse/IGNITE-20933 The method should be removed
```

I can't do it in #2825 - there are some other usages. Let's remove separately.

zstan: @ptupitsyn **assignments** is used from 1 test in your #2825 PR, i fix this usage here, probably you can take this test fixture into your branch ? Or i can see - 2 approvals on your branch, if you merge it - i can easily fix this place.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
zstan: just 2 map touch here, i don`t think it very harmful

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -317,6 +341,69 @@ public CompletableFuture<ExecutionTarget> forSystemView(ExecutionTargetFactory f
...
+            CompletableFuture<ReplicaMeta> f = placementDriver.awaitPrimaryReplica(
+                    partGroupId,
+                    clockNow,
+                    AWAIT_PRIMARY_REPLICA_TIMEOUT,
+                    SECONDS
+            );
+
+            result.add(f.handle((primaryReplica, e) -> {
+                if (e != null) {
+                    LOG.error("Failed to retrieve primary replica for partition {}", e, partitionId);
```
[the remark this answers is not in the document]
zstan: why ? this is root cause why query can`t be processed so we need to log it. Plz clarify you suggest.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -317,6 +341,69 @@ public CompletableFuture<ExecutionTarget> forSystemView(ExecutionTargetFactory f
...
+                        assert holder != null : "Unable to map query, nothing holds the lease";
+
+                        ClusterNode node = clusterSrvc.topologyService().getByConsistentId(holder);
+
+                        if (node == null) {
+                            // additional recovery logic is need to be present around here.
+                            throw new IgniteInternalException(Sql.MAPPING_ERR, "Unable to map query, node is lost or offline");
+                        }
+
+                        return new PrimaryReplica(node, primaryReplica.getStartTime().longValue());
```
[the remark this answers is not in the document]
zstan: PrimaryReplica need to be removed in scope of https://issues.apache.org/jira/browse/IGNITE-20925

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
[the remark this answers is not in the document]
korlov42: Because from sql engine's point of view it's just one reason from thousands why query may fail. If PlacementDriver went to failed state and can't provide placement for table -- it's up to PlacementDriver to report this error. Sql engine on the other hand is just a middle ware, thus we should just propagate this error up to the client. We _may_ report to the log, but only to help debugging certain problems. But it worth to remember that according to [logging guideline](https://cwiki.apache.org/confluence/display/IGNITE/Java+Logging+Guidelines#JavaLoggingGuidelines-ERROR) `ERROR is when you need a duty engineer to wake up in the middle of the night`. I see no reason to wake duty engineer if a random query got failed, this error will be propagate to caller anyway

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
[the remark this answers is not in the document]
korlov42: I mean why do we need to return PrimaryReplica on this step? On the very next one we will convert it NodeWithTerm, so why don't return NodeWithTerm here and reduce pipeline by one step?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
zstan: this check is appended just to avoid second call on already enlisted partitions. And need to be removed (as mentioned) in https://issues.apache.org/jira/browse/IGNITE-20925

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -317,6 +333,66 @@ public CompletableFuture<ExecutionTarget> forSystemView(ExecutionTargetFactory f
...
+                    String holder = primaryReplica.getLeaseholder();
+
+                    assert holder != null : "Unable to map query, nothing holds the lease";
+
+                    ClusterNode node = clusterSrvc.topologyService().getByConsistentId(holder);
+
+                    if (node == null) {
+                        // additional recovery logic is need to be present around here.
+                        throw new IgniteInternalException(Sql.MAPPING_ERR, "Unable to map query, node is lost or offline");
+                    }
```
korlov42, line 381: do we really need to resolve node?

zstan: yep, seems we can remove this part, done

korlov42 (approved):

PR #2856, opened 2023-11-21
IGNITE-20735 Implement initiate recovery handling logic

Cyrill:

https://issues.apache.org/jira/browse/IGNITE-20735

### The Review Checklist

### Notes


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -507,7 +508,12 @@ private CompletableFuture<?> processRequest(ReplicaRequest request, @Nullable Bo
      * @param request Tx recovery request.
      * @return The future is complete when the transaction state is finalized.
      */
-    private CompletableFuture<Void> processTxRecoveryAction(TxRecoveryMessage request) {
+    private CompletableFuture<Void> processTxRecoveryAction(TxRecoveryMessage request, String senderId) {
```
denis-chudov, line 511: could you please create a separate method like `private CompletableFuture<TransactionMeta> triggerTxRecovery(UUID txId)` for this, because it is needed to trigger the recovery from tx state resolution process.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -507,7 +508,12 @@ private CompletableFuture<?> processRequest(ReplicaRequest request, @Nullable Bo
...
+        // Get node id of the sender to send back cleanup requests.
+        String nodeConsistentId = transactionStateResolver.consistentId(senderId);
+
+        assert nodeConsistentId != null;
```
denis-chudov, line 515: here we can have assertion errors if the node left topology

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/TransactionStateResolver.java
```diff
@@ -129,6 +129,18 @@ public void start() {
         });
     }
 
+    /**
+     * Get consistent id by node id.
+     *
+     * @param id Node inconsistent id.
+     * @return Node consistent id.
+     */
+    public @Nullable String consistentId(String id) {
```
denis-chudov, line 138: its definitely not a task for tx state resolver, to resolve the node names by their ids

denis-chudov (changes requested):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -516,13 +522,32 @@ private CompletableFuture<Void> processTxRecoveryAction(TxRecoveryMessage reques
         boolean transactionAlreadyFinished = txMeta != null && isFinalState(txMeta.txState());
 
         if (transactionAlreadyFinished) {
-            return nullCompletedFuture();
+            if (txMeta.locksReleased()) {
+                // Locks released - nothing to do.
+                return nullCompletedFuture();
+            }
+            // Locks were not released - send cleanup.
+            if (txMeta.enlistedPartitions().isEmpty()) {
```
denis-chudov, line 530: by the way, if you call `txManager.finish` like below, you will never get empty list here: even in case of recovery you will send a tx finish request to commit partition with one partition in `enlistedPartitions` collection: the commit partition itself.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -516,13 +522,32 @@ private CompletableFuture<Void> processTxRecoveryAction(TxRecoveryMessage reques
...
         }
 
         LOG.info("Orphan transaction has to be aborted [tx={}].", txId);
 
-        // TODO: IGNITE-20735 Implement initiate recovery handling logic.
-        return nullCompletedFuture();
+        // If the transaction state is pending, then the transaction should be rolled back,
+        // meaning that the state is changed to aborted and a corresponding cleanup request
+        // is sent in a common durable manner to a partition that have initiated recovery.
+        return txManager.finish(
```
denis-chudov, line 544: now we must check on PartitionListener side that the state that is already in txStateStorage, is not final, while doing CAS. If there is already a final state, the operation should be aborted.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/rel/TableScanNodeExecutionTest.java
```diff
@@ -207,7 +208,17 @@ private static class TestInternalTableImpl extends InternalTableImpl {
...
+                        @Override
+                        public ClusterNode getById(String id) {
+                            return mock(ClusterNode.class);
+                        }
+
+                        @Override
+                        public ClusterNode getByConsistentId(String consistentId) {
+                            return mock(ClusterNode.class);
+                        }
+                    },
```
denis-chudov, line 221: I see too many of similar snippets, lets create a test implementation of ClusterNodeResolver.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -507,22 +514,59 @@ private CompletableFuture<?> processRequest(ReplicaRequest request, @Nullable Bo
...
+                        new HybridTimestampTracker(),
+                        replicationGroupId,
+                        false,
+                        Map.of(replicationGroupId, 0L), // term is not required for the rollback.
+                        txId
+                )
+                .thenCompose(unused -> runCleanupOnNode(txId, senderId));
```
denis-chudov, line 569: I think, there is a possibility of returning a failed future from `txManager.finish()` due to a race with another recovery process for the same transaction. Maybe there should be `whenComplete` here.

denis-chudov: maybe I don't get something but there is still thenCompose

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -505,46 +506,74 @@ private CompletableFuture<?> processRequest(ReplicaRequest request, @Nullable Bo
...
+     * @param txStateMeta Transaction meta.
+     * @param senderId Sender inconsistent id.
      * @return Tx recovery future, or failed future if the tx recovery is not possible.
      */
-    private CompletableFuture<TransactionMeta> triggerTxRecovery(UUID txId) {
-        // TODO: IGNITE-20735 Implement initiate recovery handling logic. This part has to be fully rewritten.
-        TxStateMeta txStateMeta = txManager.stateMeta(txId);
-
+    private CompletableFuture<TransactionMeta> triggerTxRecoveryWithState(UUID txId, @Nullable TxStateMeta txStateMeta, String senderId) {
         if (txStateMeta == null || txStateMeta.txState() == ABANDONED) {
```
denis-chudov, line 574: Why do you need this condition? this method is called when `txStateMeta == null || txStateMeta.txState() == ABANDONED || txCoordinatorIsDead()` so here you just eliminate the last condition and return the completed future when tx state is pending but coordinator is missing.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -505,46 +506,74 @@ private CompletableFuture<?> processRequest(ReplicaRequest request, @Nullable Bo
...
-                    txManager.updateTxMeta(txId, old -> {
-                        if (old == null) {
-                            return new TxStateMeta(ABORTED, null, replicationGroupId, null);
-                        } else {
-                            return new TxStateMeta(ABORTED, old.txCoordinatorId(), replicationGroupId, null);
-                        }
-                    })
-            );
+            return triggerTxRecovery(txId, senderId)
+                    .thenApply(v -> txStateStorage.get(txId));
```
denis-chudov, line 576: There are cases when we are on the commit partition primary which is not a leader, so if this state in tx state storage had been modified by the concurrent recovery process, the finish future from our recovery could fail but tx state could be not replicated yet:
- recovery A (current): start txManager.finish()
- recovery B (concurrent): start txManager.finish(), finishes successfully, but the final tx state is applied only on leader (which is some remote node)
- recovery A: txManager.finish() is failed due to recovery B, but local instance of tx state storage still doesn't have an actual tx meta.

In this case we can check the local tx state map (because the concurrent recovery happens on the same commit partition replica) and it will be either FINISHING (if the concurrent process still wasn't able to put the final state) or some final state, and in case of FINISHING we can return the finish future. What do you think?

denis-chudov: Actually, seems that it happens in another way. Any concurrent recovery may fail with assertion in txManager.finish here:
`assert !tuple0.isTxFinishing() : "Transaction is already finished [id=" + uuid + "].";`
We should return, for example, failed future with IllegalStateException instead of throwing an assertion error.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -795,28 +813,50 @@ private CompletableFuture<TransactionMeta> processTxStateCommitPartitionRequest(
...
+                        .handle((v, ex) -> {
+                            TransactionMeta transactionMeta = txManager.stateMeta(txId);
+
+                            assert transactionMeta != null;
+
+                            if (transactionMeta.txState() == FINISHING) {
+                                return ((TxStateMetaFinishing) transactionMeta).txFinishFuture();
+                            }
+
+                            assert transactionMeta.txState() == ABORTED;
```
denis-chudov, line 846: Why only aborted? There could be successful commit from coordinator.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -375,6 +398,18 @@ public CompletableFuture<Void> finish(
...
     }
 
+    private static CompletableFuture<Void> checkTxOutcome(boolean commit, UUID txId, TransactionMeta stateMeta) {
+        if ((stateMeta.txState() == COMMITTED) == commit) {
+            return nullCompletedFuture();
+        }
+
+        int result = (stateMeta.txState() == COMMITTED) && !commit ? TX_WAS_COMMITTED_ERR : TX_WAS_ABORTED_ERR;
+
+        return CompletableFuture.failedFuture(new TransactionException(result,
```
denis-chudov, line 408:
```suggestion
        int errorCode = (stateMeta.txState() == COMMITTED) && !commit ? TX_WAS_COMMITTED_ERR : TX_WAS_ABORTED_ERR;

        return CompletableFuture.failedFuture(new TransactionException(errorCode,
```

modules/api/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -354,6 +354,9 @@ public static class Transactions {
 
         /** Coordinator tries to commit a transaction that has already been aborted. */
         public static final int TX_WAS_ABORTED_ERR = TX_ERR_GROUP.registerErrorCode((short) 15);
+
+        /** Coordinator tries to abort a transaction that has already been committed. */
+        public static final int TX_WAS_COMMITTED_ERR = TX_ERR_GROUP.registerErrorCode((short) 16);
```
denis-chudov, line 359: You should also check this code in `TxManagerImpl#durableFinish` to stop durable finish. By the way, seems we dont return the correct error code from `PartitionListener#onTxStateStorageCasFail` - there is `TX_UNEXPECTED_STATE_ERR` in any case, should be either `TX_WAS_ABORTED_ERR` or `TX_WAS_COMMITTED_ERR`.

denis-chudov: I am not sure, maybe it still can be useful for some cases of parallel recovery to see the actual state in FSM

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -576,7 +583,11 @@ private static void onTxStateStorageCasFail(UUID txId, TxMeta txMetaBeforeCas, T
                 txMetaToSet
         );
 
-        IgniteInternalException stateChangeException = new IgniteInternalException(TX_UNEXPECTED_STATE_ERR, errorMsg);
+        IgniteInternalException stateChangeException =
+                new UnexpectedTransactionStateException(
+                        errorMsg,
+                        new TransactionResult(txMetaBeforeCas.txState(), txMetaBeforeCas.commitTimestamp())
```
denis-chudov, line 589: txMetaBeforeCas can be null, still CAS can fail.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TransactionResult.java
```diff
@@ -0,0 +1,51 @@
...
+
+import static org.apache.ignite.internal.hlc.HybridTimestamp.hybridTimestampToLong;
+import static org.apache.ignite.internal.hlc.HybridTimestamp.nullableHybridTimestamp;
+
+import java.io.Serializable;
+import org.apache.ignite.internal.hlc.HybridTimestamp;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Represents transaction result - the state and the commit timestamp.
```
denis-chudov, line 28:
```suggestion
 * Represents transaction final result - the state and the commit timestamp.
```

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -375,6 +397,17 @@ public CompletableFuture<Void> finish(
                 ));
     }
 
+    private static CompletableFuture<Void> checkTxOutcome(boolean commit, UUID txId, TransactionMeta stateMeta) {
+        if ((stateMeta.txState() == COMMITTED) == commit) {
+            return nullCompletedFuture();
+        }
+
+        return CompletableFuture.failedFuture(new TransactionAlreadyFinishedException(
+                "Failed to change the outcome of a finished transaction [txId=" + txId + "].",
```
denis-chudov, line 406: Don't you think that the exception message should include the transaction result?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1575,33 +1608,38 @@ private CompletableFuture<Void> finishAndCleanup(
...
+                        txMeta.txState()
+                );
 
-                throw new TransactionException(TX_WAS_ABORTED_ERR,
-                        "Failed to change the outcome of a finished transaction"
-                                + " [txId=" + txId + ", txState=" + txMeta.txState() + "].");
+                throw new TransactionAlreadyFinishedException(
+                        "Failed to change the outcome of a finished transaction [txId=" + txId + "].",
+                        new TransactionResult(txMeta.txState(), txMeta.commitTimestamp())
+                );
```
denis-chudov, line 1642: Don't you think that the exception message should include the transaction result?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -439,20 +472,26 @@ private CompletableFuture<Void> durableFinish(
...
+                                TxStateMeta txStateMeta =
+                                        new TxStateMeta(
+                                                result.transactionState(),
+                                                old.txCoordinatorId(),
+                                                commitPartition,
+                                                result.commitTimestamp()
+                                        );
 
-                                    txFinishFuture.complete(txStateMeta);
+                                txFinishFuture.complete(txStateMeta);
```
denis-chudov, line 489: By the way, let's better move this future completion out of the meta updater (and hence from the map.compute closure)

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
denis-chudov: sorry, we are in FSM here. Should be ok.

denis-chudov (approved):

vldpyatkov (approved):

PR #2857, opened 2023-11-21
IGNITE-20778: Provide internal API to get a list of parameters for non-executed query

lowka:
Adds type inference for unspecified dynamic parameters via `SqlValidator` / `TypeCoercion`. Implementation is based on the following idea:

- We can use type information provided to `TypeCoercion` to infer type of unspecified dynamic parameters. Example: for `+` operator when both sides are unspecified dynamic parameters - this is a type inference error, if one side is specified (and it does not matter whether it is a dynamic parameter or not, because the type is known), assume that another side has the same type.

- cast operator accepts any value and relies on additional help at runtime - `ExecutionContext::get(?x)` should be modified to perform conversion at runtime.

**Limitations**

- Passing type information to nested statements is not supported. E.g. Both `SELECT 1 UNION SELECT ?` and  `INSERT INTO t VALUES (1), ((SELECT ?))`, `SELECT * FROM t WHERE col = SOME(SELECT ? ...)` are rejected, since no type information is provided from the outer scope. 

- Functions require additional handling because of type families and due to the fact that type inference for some operators does not use type coercion (see `COUNT`).

- `COALESCE`, `NULLIF` - due to rewrites into `CASE` expressions, errors for these operators are rather confusing.

### What this PR covers

- Should work for all statement types (queries, DML, etc).
- Binary operators, binary comparison,  predicate operators (IS NULL, IS NOT NULL, etc).
- `CASE` expression , `NULLIF`, `COALESCE`.
- Functions: since function definitions in calcite are defined in terms of type families (and most of type families have more than one type) dynamic parameters with no value in function arguments require explicit casts. There are about 50+ functions and they require case by case review/fixes with possible changes of their definitions. This can be improved subsequent PRs.
- Aggregates require extra handling to support correct error messages as well (see `AVG` vs `MAX` vs `COUNT`) - the first one uses type inference for builtin functions, the second one uses type coercion for binary operators, the third one uses neither.

### Implementation details

- Uses `SqlValidator::getParameterRowType` to retrieve parameter types.
- All dynamic parameters types are inferred as nullable types.
- This patch reveals a problem in type inference for `IN` operator (`InferTypes::FIRST_KNOWN` traverses into SqlNodeList which has no implementation).

Error reporting:

Unspecified parameters:
Adds error messages for type validation of unspecified dynamic parameters. E.g. `SELECT ? + ?` now returns `Ambiguous operator <UNKNOWN> + <UNKNOWN>". Dynamic parameter requires adding explicit type cast`.

Specified parameters:
This PR does not updates an error message for this case, and it remains the same (fixing it requires propagating derived operand types into `TypeCoercion::coerceOperandType`. 

### Other databases

It seems like both Cockroach and PostgreSQL are able to infer type in case of both `SELECT 1 UNION SELECT ?` and `SELECT ? UNION SELECT 1` but both unable to do so for `INSERT INTO t VALUES (1), ((SELECT ?))`.

https://issues.apache.org/jira/browse/IGNITE-20778

---

### The Review Checklist

### Notes


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -772,91 +858,189 @@ private boolean isSystemFieldName(String alias) {
...
+         * <ul>
+         *    <li>{@code null} - parameter has not been checked - this is a bug.</li>
+         *    <li>{@code unknownType} means the type of this parameter has not been resolved due to ambiguity.</li>
+         *    <li>Otherwise contains a type of a parameter.</li>
+         * </ul>
+         */
+        RelDataType resolvedType;
+    }
+
+    /** Unspecified value. */
```
lowka, line 1042: This part is debatable.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -516,12 +522,16 @@ private List<AbstractNode<?>> localFragments() {
         private CompletableFuture<Void> sendFragment(
                 String targetNodeName, String serialisedFragment, FragmentDescription desc, TxAttributes txAttributes
         ) {
+            Object[] parameterValues = Arrays.stream(ctx.parameters())
+                    .map((v) -> v.value())
+                    .toArray();
```
AMashenkov, line 527: Let's rewrite this avoiding streams and extract this to a method of ArrayUtils class.

AMashenkov: To reduce number of objects... maybe it will be better to wrap a whole Object[] into DynamicParametersHolder instead of wrapping individual parameters?
```
DynamicParametersHolder {
    Objects[] params;
    BitSet hasValueFlags
}
```

lowka: I agree, but at the moment there is no separation between prepare and execution and because of that they share the same context class , when in fact for execution we only need exact values ((param_indes, value) ) and for preparation we need (param_index, value/no_value). 

Maybe it would be better to fix this when we introduce a separate preparation phase.

lowka: Removed java stream usage.

lowka: I moved code that creates an array of dynamic parameter values from array to `DynamicParameterValue`. Because creating a generic array requires use of reflection to create an array  + such method is very niche.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/DynamicParametersTest.java
```diff
@@ -283,12 +636,62 @@ public Stream<DynamicTest> testMergeFull() {
...
+
     /**
      * Function calls.
      */
     @TestFactory
     public Stream<DynamicTest> testFunction() {
         return Stream.of(
+                checkStatement()
+                        .sql("SELECT SUBSTRING(?, 1)", Unspecified.UNKNOWN)
+                        .fails("Ambiguous operator SUBSTRING(<UNKNOWN>, <INTEGER>)"),
```
AMashenkov, line 689: Why does it fail, but a "nested call" is ok?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/DynamicParametersTest.java
```diff
@@ -301,6 +704,29 @@ public Stream<DynamicTest> testFunction() {
...
+
+                sql("SELECT ? UNION SELECT 1", 1)
+                        .parameterTypes(nullable(NativeTypes.INT32))
+                        .ok(),
+
+                sql("SELECT ? UNION SELECT 1", Unspecified.UNKNOWN)
+                        .fails("Unable to determine type of a dynamic parameter#0."),
+
+                sql("SELECT ? UNION SELECT ?", 1, Unspecified.UNKNOWN)
+                        .fails("Unable to determine type of a dynamic parameter#1.")
```
AMashenkov, line 726: Does it make sens to add smth like that
```
sql("SELECT 1 UNION SELECT ?", 1L)
sql("SELECT ? UNION SELECT ?", 1L, 2)
```

lowka:
@korlov42 The number of validation parameters check should stay. Because some parameters may be left unspecified, and some are not.  e.g. for a query SELECT *, c = 1 +? FROM t = ?  it possible to specified value of the second parameter, but left the first one unspecified.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -440,49 +531,41 @@ public void validateAggregateParams(SqlCall aggCall,
...
-        checkTypesInteroperability(scope, expr);
-
-        RelDataType dataType = super.deriveType(scope, expr);
-
-        // If type of dynamic parameter has not been inferred, use a type of its value.
-        RelDataType paramType = expr instanceof SqlDynamicParam
-                ? getDynamicParamType((SqlDynamicParam) expr) : null;
+        if (expr instanceof SqlDynamicParam) {
+            return deriveDynamicParamType((SqlDynamicParam) expr);
+        } else {
```
AMashenkov, line 536:
```suggestion
        }
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -440,49 +531,41 @@ public void validateAggregateParams(SqlCall aggCall,
...
+            if (!SqlKind.BINARY_COMPARISON.contains(sqlKind)) {
+                return dataType;
+            }
 
-        // Comparison and arithmetic operators are SqlCalls.
-        SqlCall sqlCall = (SqlCall) expr;
-        var lhs = getValidatedNodeType(sqlCall.operand(0));
-        var rhs = getValidatedNodeType(sqlCall.operand(1));
+            // Comparison and arithmetic operators are SqlCalls.
+            SqlCall sqlCall = (SqlCall) expr;
```
AMashenkov, line 548: Can we get class cast exception here?
Why do you drop the branch?
```
if (!(expr instanceof SqlCall)) {		
             return dataType;		
         }
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -772,91 +857,200 @@ private boolean isSystemFieldName(String alias) {
...
 
-                throw new AssertionError(message);
-            }
+    /** Returns {@code true} if the given node is dynamic parameter that has no value set. */
+    public boolean isUnspecifiedDynamicParam(SqlNode node) {
+        if (node.getKind() != SqlKind.DYNAMIC_PARAM) {
+            return false;
+        } else {
+            return isUnspecified((SqlDynamicParam) node);
+        }
```
AMashenkov, line 1026:
```suggestion
       return node.getKind() == SqlKind.DYNAMIC_PARAM && isUnspecified((SqlDynamicParam) node);
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/DynamicParametersTest.java
AMashenkov: You have a test that expects INT32 type, but here is a Long (INT64).
Will and should this work?

lowka: Yes for both questions.

lowka: >You have a test that expects INT32 type, but here is a Long (INT64).
Well, The correct answer to this question is the following:
- The test expect that the statement `SELECT ? UNION SELECT ?` works. if of one parameter is an integer and another is a bigint. Nothing more, nothing less.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/metrics/PlanningCacheMetricsTest.java
```diff
@@ -56,20 +56,20 @@ public void plannerCacheStatisticsTest() throws Exception {
         MetricSet metricSet = metricManager.enable(SqlPlanCacheMetricSource.NAME);
 
         try {
```
lowka, line 58: Now we have 1 additional read be prepare with no dynamic params or with all parameters set.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/DynamicParameterValue.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+package org.apache.ignite.internal.sql.engine.prepare;
+
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Dynamic parameter value.
+ */
+public final class DynamicParameterValue {
```
korlov42, line 25: why do we need this wrapper?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ExplainPlan.java
```diff
@@ -36,10 +36,12 @@ public class ExplainPlan implements QueryPlan {
 
     private final PlanId id;
     private final MultiStepPlan plan;
+    private final ParameterMetadata parameterMetadata;
 
-    ExplainPlan(PlanId id, MultiStepPlan plan) {
+    ExplainPlan(PlanId id, MultiStepPlan plan, ParameterMetadata parameterMetadata) {
```
korlov42, line 41: when `parameterMetadata` for explain may be different from one returned by the plan it's meant to explain?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ParameterMetadata.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /** Return metadata for each parameter. */
+    public List<ParameterType> parameterTypes() {
+        return parameterTypes;
+    }
+
+    @Override
+    public String toString() {
+        return S.toString(ParameterMetadata.class, this, "parametersTypes", parameterTypes);
```
korlov42, line 42: why do you need additional parameters here?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/util/QueryCheckerTest.java
```diff
@@ -247,6 +248,11 @@ void testMetadata() {
                 .check();
     }
 
+    @Test
+    public void testParameterMetadata() {
+
```
korlov42, line 253: empty test?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImpl.java
```diff
@@ -292,30 +303,53 @@ private CompletableFuture<QueryPlan> prepareQuery(ParsedResult parsedResult, Pla
...
 
     private PlanId nextPlanId() {
         return new PlanId(prepareServiceId, planIdGen.getAndIncrement());
     }
 
     private CompletableFuture<QueryPlan> prepareDml(ParsedResult parsedResult, PlanningContext ctx) {
-        var key = createCacheKey(parsedResult, ctx);
+        // If a caller passes all the parameters, then get parameter types and check to see whether a plan future already exists.
+        CompletableFuture<QueryPlan> f = getPlanIfParameterHaveValues(parsedResult,
+                ctx);
```
korlov42, line 347: kinda strange formatting

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ParameterMetadata.java
lowka: Because `S::toString(this)` always produces `ParameterMetadata []` which does not include string representations of parameters.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ExplainPlan.java
lowka: Thanks. It never does.

korlov42 (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -513,6 +570,27 @@ private CompletableFuture<AsyncSqlCursor<List<Object>>> queryScript0(
...
 
+    private CompletableFuture<QueryPlan> prepareParsedStatement(String schemaName,
+            ParsedResult parsedResult,
+            HybridTimestamp timestamp,
+            QueryCancel queryCancel,
+            Object[] params) {
+
+        return waitForActualSchema(schemaName, timestamp)
+                .thenCompose(schema -> {
+                    Map<Integer, Object> dynamicParams = Commons.arrayToMap(params);
```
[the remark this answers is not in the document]
lowka: Fixed

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -138,11 +149,14 @@ public class IgniteSqlValidator extends SqlValidatorImpl {
      * @param parameters    Dynamic parameters
      */
     public IgniteSqlValidator(SqlOperatorTable opTab, CalciteCatalogReader catalogReader,
-            IgniteTypeFactory typeFactory, SqlValidator.Config config, Object[] parameters) {
+            IgniteTypeFactory typeFactory, SqlValidator.Config config, Map<Integer, Object> parameters) {
```
AMashenkov, line 152: Can IntMap be used here?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -119,14 +126,18 @@ public class IgniteSqlValidator extends SqlValidatorImpl {
...
+    private final Map<Integer, DynamicParamState> dynamicParameters;
 
-    /** Dynamic parameters SQL AST nodes for invariant checks - see {@link #validateInferredDynamicParameters()}. */
-    private final SqlDynamicParam[] dynamicParamNodes;
+    /**
+     * The same dynamic parameter can be used in the same SQL tree multiple types after a rewrite.
+     * (E.g. COALESCE(?0, ?1) is rewritten into CASE WHEN ?0 IS NOT NULL THEN ?0 ELSE ?1 END)
+     * We store them to check that every i-th parameter has the same type.
+     */
+    private final IdentityHashMap<SqlDynamicParam, SqlDynamicParam> dynamicParamNodes = new IdentityHashMap<>();
```
AMashenkov, line 137: Can IntMap be used here?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
```diff
@@ -485,6 +580,37 @@ public RelDataType deriveType(SqlValidatorScope scope, SqlNode expr) {
...
     }
 
+    /** {@inheritDoc} */
+    @Override public RelDataType getParameterRowType(SqlNode sqlQuery) {
+        // We do not use calcite' version since it is contains a bug,
+        // alreadyVisited visited uses object identity, but rewrites of NULLIF, COALESCE
+        // into dynamic parameters may place the same parameter into multiple positions
+        // in SQL tree.
+        List<RelDataType> types = new ArrayList<>();
+        Set<Integer> alreadyVisited = new HashSet<>();
```
AMashenkov, line 590: Can IntSet be used here?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ParameterMetadata.java
AMashenkov: you can mark `parameterTypes` field with `@IgniteToStringInclude` annotation

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImplTest.java
```diff
@@ -103,18 +118,158 @@ void prepareServiceCachesPlanCreatedForExplain() {
...
+        // Parameter types do not match
+        QueryPlan queryPlan3 = await(service.prepareAsync(
+                parse("UPDATE t SET a = ? WHERE c = ?"),
+                createContext(1, 1L)
+        ));
+        assertNotSame(queryPlan1, queryPlan3);
+    }
+
+    @Test
+    void preparePropagatesValidationError() {
```
AMashenkov, line 184:
```suggestion
    pubic void preparePropagatesValidationError() {
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
lowka: No we, do not. As using int -> sql node here defeats the purpose of this map. (see the comment)
Because Validator rewrites COALESCE into a case expressions that uses the same parameter multiple times.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImplTest.java
```diff
@@ -103,18 +118,158 @@ void prepareServiceCachesPlanCreatedForExplain() {
...
+        // Parameter types do not match
+        QueryPlan queryPlan3 = await(service.prepareAsync(
+                parse("SELECT * FROM t WHERE a = ? and c = ?"),
+                createContext(1, 1L)
+        ));
+        assertNotSame(queryPlan1, queryPlan3);
+    }
+
+    @Test
+    void prepareReturnsDmlPlanThatDependsOnParameterTypeMatchInferred() {
```
AMashenkov, line 153:
```suggestion
    public void prepareReturnsDmlPlanThatDependsOnParameterTypeMatchInferred() {
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImplTest.java
```diff
@@ -103,18 +118,158 @@ void prepareServiceCachesPlanCreatedForExplain() {
         assertThat(plan.plan(), sameInstance(queryPlan));
     }
 
+    @Test
+    void prepareReturnsQueryPlanThatDependsOnParameterTypeMatchInferred() {
```
AMashenkov, line 122:
```suggestion
    public void prepareReturnsQueryPlanThatDependsOnParameterTypeMatchInferred() {
```

AMashenkov (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
lowka: Sure. Done.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlValidator.java
lowka: Absolutely.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImplTest.java
lowka: For some reason all test cases in this class are package private .

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImplTest.java
lowka: Fixed.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImplTest.java
lowka: Fixed.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImplTest.java
lowka: Fixed.

PR #2867, opened 2023-11-23
Start partitions on corresponding assignments.stable, calculate if missing, cleanup obsolete resources (part 1)

denis-chudov:
https://issues.apache.org/jira/browse/IGNITE-20210

### The Review Checklist

### Notes


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1057,9 +1079,8 @@ private CompletableFuture<?> createTableLocally(long causalityToken, int catalog
 
             // Check if the table already has assignments in the vault.
             // So, it means, that it is a recovery process and we should use the vault assignments instead of calculation for the new ones.
-            // TODO: IGNITE-20210 Fix it
-            if (partitionAssignments(vaultManager, tableId, 0) != null) {
-                assignmentsFuture = completedFuture(tableAssignments(vaultManager, tableId, zoneDescriptor.partitions()));
+            if (partitionAssignments(metaStorageMgr, tableId, 0, causalityToken) != null) {
```
vldpyatkov, line 1082: Why do you use the call for 0 participation only?

denis-chudov: It's used only for condition to peek whether there are any assignments in meta storage. If yes, we read the assignments themselves below

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -629,201 +630,222 @@ private CompletableFuture<?> createTablePartitionsLocally(
...
+                return ByteUtils.toBytes(assignments);
+            }
 
-                var safeTimeTracker = new PendingComparableValuesTracker<HybridTimestamp, Void>(
-                        new HybridTimestamp(1, 0)
-                );
-                var storageIndexTracker = new PendingComparableValuesTracker<Long, Void>(0L);
+            @Override
+            public long revision() {
+                return 1;
```
kgusakov, line 686: I would prefer to throw `java.lang.UnsupportedOperationException`here and in other unused methods - to simplify the debugging of any connected issues in future.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/RebalanceUtil.java
```diff
@@ -517,23 +516,26 @@ static CompletableFuture<List<Set<Assignment>>> tableAssignments(
     /**
      * Returns table assignments for all table partitions from vault.
```
kgusakov, line 517: "from vault" the same as previous

kgusakov:
As I understand: we are not cleaning resources from the obsolete partitions on recovery. So, if there were some updates for `stable` assignments during the node unavailability - we will have some resources leak. Is this issue out of ticket scope? Because now it is part of it's title

kgusakov (changes requested):

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/RebalanceUtil.java
```diff
@@ -466,17 +465,17 @@ public static CompletableFuture<Set<Assignment>> partitionAssignments(
     /**
      * Returns partition assignments from vault.
      *
-     * @param vaultManager Vault manager.
+     * @param metaStorageManager Meta storage manager.
      * @param tableId Table id.
      * @param partitionNumber Partition number.
+     * @param revision Revision.
      * @return Returns partition assignments from vault or {@code null} if assignments is absent.
```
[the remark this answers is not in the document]
denis-chudov: fixed

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/RebalanceUtil.java
denis-chudov: fixed

modules/table/src/integrationTest/java/org/apache/ignite/internal/rebalance/ItRebalanceDistributedTest.java
```diff
@@ -493,7 +493,7 @@ void testDestroyPartitionStoragesOnEvictNode() throws Exception {
     @Test
     @UseTestTxStateStorage
     @UseRocksMetaStorage
-    @Disabled("https://issues.apache.org/jira/browse/IGNITE-20210")
+    @Disabled("https://issues.apache.org/jira/browse/IGNITE-21463")
```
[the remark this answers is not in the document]
denis-chudov: Accidently put IGN ticket number, fixed

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
denis-chudov: Fixed.

modules/table/src/integrationTest/java/org/apache/ignite/internal/rebalance/ItRebalanceDistributedTest.java
```diff
@@ -690,11 +690,19 @@ private void directUpdateMetastoreRebalanceAssignmentKeys() throws Exception {
...
-        for (int i = 0; i < NODE_COUNT; i++) {
-            verify(getNode(i).raftManager, times(1))
-                    .startRaftGroupNode(any(), any(), any(), any(), any(RaftGroupOptions.class));
-            verify(getNode(i).replicaManager, times(1)).startReplica(any(), any(), any(), any(), any());
-        }
+        waitForCondition(() -> {
+            try {
+                for (int i = 0; i < NODE_COUNT; i++) {
+                    verify(getNode(i).raftManager, times(1))
+                            .startRaftGroupNode(any(), any(), any(), any(), any(RaftGroupOptions.class));
```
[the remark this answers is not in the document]
denis-chudov: Reworked.

kgusakov (approved):

modules/table/src/integrationTest/java/org/apache/ignite/internal/rebalance/ItRebalanceDistributedTest.java
```diff
@@ -500,7 +500,7 @@ void testDestroyPartitionStoragesOnEvictNode() throws Exception {
     @Test
     @UseTestTxStateStorage
     @UseRocksMetaStorage
-    @Disabled("https://issues.apache.org/jira/browse/IGNITE-20210")
+    @Disabled("https://issues.apache.org/jira/browse/IGNITE-19170")
     void testDestroyPartitionStoragesOnRestartEvictedNode(TestInfo testInfo) throws Exception {
```
sanpwc, line 504: Should we write some tests that will ensure that proposed implementation is correct?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -523,38 +520,57 @@ public void start() {
     }
 
     private void performRebalanceOnRecovery(long recoveryRevision) {
-        CompletableFuture<Void> pendingAssignmentsRecoveryFuture;
-
-        var prefix = new ByteArray(PENDING_ASSIGNMENTS_PREFIX);
+        var pendingAssignmentsPrefix = new ByteArray(PENDING_ASSIGNMENTS_PREFIX);
```
sanpwc, line 523: Why did you change performRebalanceOnRecovery? rebalance is about handling pending keys. I expect it to be properly implemented in IGNITE-20344. Did you find bugs there?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -523,38 +520,57 @@ public void start() {
...
+        startVv.update(recoveryRevision, (v, e) -> handleAssignmentsEventsOnRecovery(
+                pendingAssignmentsPrefix,
+                recoveryRevision,
+                (entry, rev) -> handleChangePendingAssignmentEvent(entry, localPartsByTableIdVv.get(recoveryRevision), rev, true),
+                "pending"
+        ));
+    }
 
-        try (Cursor<Entry> cursor = metaStorageMgr.prefixLocally(prefix, recoveryRevision)) {
+    private CompletableFuture<Void> handleAssignmentsEventsOnRecovery(
```
sanpwc, line 540: What do you mean? On recovery there are no events, right? We just scan catalog, retrive corresponding stable and pending assignments, etc? What kind of events do you handle on recovery?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -630,201 +646,222 @@ private CompletableFuture<?> createTablePartitionsLocally(
...
-
             for (int i = 0; i < partitions; i++) {
                 int partId = i;
 
                 Set<Assignment> newPartAssignment = newAssignments.get(partId);
 
-                InternalTable internalTbl = table.internalTable();
+                // TODO https://issues.apache.org/jira/browse/IGNITE-19170 #handleChangePendingAssignmentEvent should be called on
+                // TODO actual event, the method #createTablePartitionsLocally should be removed.
+                handleChangePendingAssignmentEvent(
```
sanpwc, line 665: What's the point in considering initial assignments as pending one? Generally it sounds reasonable but only if you took into consideration the fact that assignments.stable might be empty and in this case there's no sense in calling changePeersAsync, did you?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1272,6 +1346,8 @@ private void dropTableLocally(long causalityToken, CatalogTableDescriptor tableD
 
             CompletableFuture<?>[] stopReplicaFutures = new CompletableFuture<?>[partitions];
```
sanpwc, line 1347: Seems that proposed changes are too complicated. Let's step back for a while and discuss (and design) an expected flow.

- Should we treat initial table creation as a rebalance from null assignments to pending-initial one? Or we should consider it as direct assignments.stable management? 
- Do we treat assignments.stable handling differently while it's the initial set or assignments.switch? Is it reasonable to split given handling into two methods or merge them within one?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/RebalanceUtil.java
```diff
@@ -465,19 +464,19 @@ public static CompletableFuture<Set<Assignment>> partitionAssignments(
...
      *
-     * @param vaultManager Vault manager.
+     * @param metaStorageManager Meta storage manager.
      * @param tableId Table id.
      * @param partitionNumber Partition number.
-     * @return Returns partition assignments from vault or {@code null} if assignments is absent.
+     * @param revision Revision.
+     * @return Returns partition assignments from meta storage locally or {@code null} if assignments is absent.
      */
     public static Set<Assignment> partitionAssignments(
```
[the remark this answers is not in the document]
denis-chudov: fixed

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/RebalanceUtil.java
```diff
@@ -465,19 +464,19 @@ public static CompletableFuture<Set<Assignment>> partitionAssignments(
...
-            VaultManager vaultManager, int tableId, int partitionNumber) {
-        VaultEntry entry =
-                vaultManager.get(stablePartAssignmentsKey(new TablePartitionId(tableId, partitionNumber))).join();
+            MetaStorageManager metaStorageManager, int tableId, int partitionNumber, long revision) {
```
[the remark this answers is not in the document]
denis-chudov: Otherwise it would break string length limitation. I separated them into different lines.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/RebalanceUtil.java
```diff
@@ -465,19 +464,19 @@ public static CompletableFuture<Set<Assignment>> partitionAssignments(
...
+        Entry entry = metaStorageManager.getLocally(stablePartAssignmentsKey(new TablePartitionId(tableId, partitionNumber)), revision);
 
-        return (entry == null) ? null : ByteUtils.fromBytes(entry.value());
+        return (entry == null || entry.empty() || entry.tombstone()) ? null : ByteUtils.fromBytes(entry.value());
```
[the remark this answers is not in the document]
denis-chudov: fixed

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/RebalanceUtil.java
```diff
@@ -516,25 +515,28 @@ static CompletableFuture<List<Set<Assignment>>> tableAssignments(
...
-                .mapToObj(i ->
-                        (Set<Assignment>) ByteUtils.fromBytes(
-                                vaultManager.get(
-                                        stablePartAssignmentsKey(new TablePartitionId(tableId, i))
-                                ).join().value())
-                )
+                .mapToObj(p -> {
+                    Entry e = metaStorageManager.getLocally(stablePartAssignmentsKey(new TablePartitionId(tableId, p)), revision);
+
+                    assert !e.empty() && !e.tombstone() : e;
```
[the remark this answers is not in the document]
denis-chudov: Because we first check the presence of assignments for one partition, and then presume that others are also present. If some partitions missing, this is some kind of corruption and we throw assertion.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
denis-chudov: Probably it should be renamed to `processAssignmentsOnRecovery`. We will need to process both kinds of assignments anyway.

modules/table/src/integrationTest/java/org/apache/ignite/internal/rebalance/ItRebalanceDistributedTest.java
denis-chudov: On this step I mostly care not to break something, but probably there should be some tests for storages cleanup. I expect more tests to be added in [IGNITE-20993](https://issues.apache.org/jira/browse/IGNITE-20993)

modules/table/src/main/java/org/apache/ignite/internal/table/InternalTable.java
```diff
@@ -451,6 +451,14 @@ Publisher<BinaryRow> lookup(
      */
     RaftGroupService partitionRaftGroupService(int partition);
 
+    /**
+     * Returns raft group client for corresponding partition.
+     *
+     * @param partition Partition number.
+     * @return Whether raft group client for corresponding partition is started.
+     */
+    boolean partitionRaftGroupServiceStarted(int partition);
```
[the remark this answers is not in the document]
denis-chudov: It is needed because `partitionRaftGroupService` doesn't return null if service is not started, it throws exception.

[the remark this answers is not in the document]
sanpwc: Yep, but why we need it? I mean it's used in order to check whether it's required to start raft server on handling changePendingAssignmentEvent: if there's server up and running - no need to start it. But why not to compare assignments.stable and assignments.pending and start (assignments.pending - assignments.stable)?

As far as I remember we've discussed similar topic several months ago with @kgusakov thus he may have some concerns here.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/RebalanceUtil.java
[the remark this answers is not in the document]
sanpwc: I didn't get the point. Why we treat assignments presence differently in `tableAssignmentsGetLocally` and `tableAssignments`?

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/verification/WaitForTimes.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private WaitForTimes(long wantedTimes, long timeoutMillis) {
+        this.wantedTimes = wantedTimes;
+        this.timeoutMillis = timeoutMillis;
+    }
+
+    @Override
+    public void verify(VerificationData data) {
+        try {
+            waitForCondition(() -> {
```
sanpwc, line 43: Well, let's skip this for now, but proposed approach doesn't look reliable enough. We may be lucky enough to match an expected number within specific iteration. In that case we will stop the check, despite the fact that the number might have increased with further waiting.

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/verification/WaitForTimes.java
```diff
@@ -0,0 +1,69 @@
...
+                MatchableInvocation wanted = data.getTarget();
+                long actual = data.getAllInvocations().stream().filter(wanted::matches).count();
+
+                if (actual > wantedTimes) {
+                    throw new AssertionError("Called more times than expected: " + actual);
```
sanpwc, line 48: Should we also print an expected number?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sanpwc: Seems that it's not true for now. On node restart we handle stable assignments directly with `startLocalPartitionsAndUpdateClients` without calling `handleAssignmentsOnRecovery`.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -526,39 +524,51 @@ public void start() {
...
+        startVv.update(recoveryRevision, (v, e) -> handleAssignmentsOnRecovery(
+                pendingAssignmentsPrefix,
+                recoveryRevision,
+                (entry, rev) -> handleChangePendingAssignmentEvent(entry, rev, true),
+                "pending"
+        ));
+    }
 
-        try (Cursor<Entry> cursor = metaStorageMgr.prefixLocally(prefix, recoveryRevision)) {
+    private CompletableFuture<Void> handleAssignmentsOnRecovery(
```
sanpwc, line 538: As mentioned above `processAssignmentsOnRecovery` is only used to handle pending assignments and not the stable one. Thus either approach or naming is misleading.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -612,223 +622,191 @@ private CompletableFuture<Void> onTableDelete(DropTableEventParameters parameter
...
      * Updates or creates partition raft groups and storages.
      *
-     * @param causalityToken Causality token.
      * @param assignmentsFuture Table assignments.
      * @param table Initialized table entity.
      * @return future, which will be completed when the partitions creations done.
      */
-    private CompletableFuture<?> createTablePartitionsLocally(
-            long causalityToken,
+    private CompletableFuture<Void> startLocalPartitionsAndUpdateClients(
```
sanpwc, line 629: Where do we actually update clients? And where do we actually use that updated clients?

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/verification/WaitForTimes.java
denis-chudov: well, after thh las rework this change is no longer needed, so I remove this class completely

modules/core/src/testFixtures/java/org/apache/ignite/internal/testframework/verification/WaitForTimes.java
denis-chudov: removed

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
denis-chudov: Inside of startPartitionAndStartClient, in thenComposeAsync for startGroupFut. After that they are put into InternalTableImpl
`((InternalTableImpl) internalTbl).updateInternalTableRaftGroupService(partId, updatedRaftGroupService);`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
denis-chudov: I think it will be reasonable after stable assignments will be processed on recovery as well, so I decided to leave it as is

sanpwc: Ok

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sanpwc: NOTSUREIF

sanpwc: Makes sense.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
```diff
@@ -1191,6 +1207,90 @@ public void updateClusterCfgWithDefaultValue() {
...
+            Set<Assignment> assignment = ByteUtils.fromBytes(e.value());
+
+            boolean shouldBe = assignment.stream().anyMatch(n -> n.consistentId().equals(restartedNode.name()));
+
+            PeersAndLearners configuration = PeersAndLearners.fromConsistentIds(assignment.stream().map(a -> a.consistentId()).collect(
+                    Collectors.toSet()), Set.of());
+
+            Peer peer = configuration.peer(restartedNode.name());
+
+            boolean isStarted = peer == null ? false : restartedNode.raftManager().isStarted(new RaftNodeId(tablePartitionId, peer));
```
kgusakov, line 1288: Could you explain this check? As I understand - we received assignments from local metastore and check double times (with the different ways, but anyway), if this assignments has the restarted node...

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1988,51 +1981,94 @@ protected CompletableFuture<Void> handleChangeStableAssignmentEvent(WatchEvent e
...
+                            pendingAssignments,
+                            isRecovery,
+                            revision
+                    );
+                }, ioExecutor);
+    }
 
-                    boolean shouldStopLocalServices = Stream.concat(stableAssignments.stream(), pendingAssignments.stream())
-                            .noneMatch(assignment -> assignment.consistentId().equals(localMemberName));
+    private CompletableFuture<Void> updateClientsOnPartitionStop(
```
kgusakov, line 2032: It is hard to understand, why we need to update the clients on partition stop. Why do you want to move it from the handleChangeStableAssignmentEvent to stopAndDestroyPartitionAndUpdateClients?

kgusakov (changes requested):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
denis-chudov: Because it appears that we should only stop partitions on handling of stable assignments. I decided to move this part to separate method because it improved the readability of code, as for me.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
denis-chudov: The reason is that I've actually taken this part from table manager, and I didn't like this check and thought it would be nice to create some fake configuration in test that includes all nodes, to be able to create raft node id. I even thought about it but forgot in final version.

Could you check the current version?

modules/table/src/main/java/org/apache/ignite/internal/table/InternalTable.java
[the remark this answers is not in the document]
sanpwc: Any updates?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
kgusakov: Ok, but the method itself is not connected with stop, it is just updating the clients. maybe the "updatePartitoinClients" will be better?

denis-chudov: ok will rename it

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
kgusakov: Yep, thanks!

modules/table/src/main/java/org/apache/ignite/internal/table/InternalTable.java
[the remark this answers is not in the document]
denis-chudov: I think I can't remember the exact issue which was the reason why this method was needed, but it's gone after the latest rework. I removed this method.

sanpwc (approved):

kgusakov (approved):

PR #2873, opened 2023-11-24
IGNITE-20604 Refactor CausalityDataNodesEngine

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-20604

### The Review Checklist

### Notes


modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/causalitydatanodes/CausalityDataNodesEngine.java
```diff
@@ -147,41 +148,32 @@ public CompletableFuture<Set<String>> dataNodes(long causalityToken, int zoneId)
...
+            long lastScaleDownRevision = getRevisionsOfLastScaleDownEvent(causalityToken, catalogVersion, zoneId);
 
-            if (lastCfgRevision == versionedCfg.firstKey()
-                    && lastCfgRevision >= lastScaleUpRevision
-                    && lastCfgRevision >= lastScaleDownRevision
+            if (createRevision != null && createRevision.equals(descLastUpdateRevision)
+                    && descLastUpdateRevision >= lastScaleUpRevision
+                    && descLastUpdateRevision >= lastScaleDownRevision
             ) {
                 // It means that the zone was created but the data nodes value had not updated yet.
```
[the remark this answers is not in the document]
sergeyuttsel: I think we don't need all code related to the augmentation map in the CausalityDataNodesEngine#dataNodes method. This code is needed for zones with immediate timers. But now the initDataNodesAndTriggerKeysInMetaStorage() watch listener updates augmentation and returns the future which is completed when the data nodes is updated in MS. So we don't need to add/remove nodes from the augmentation map to data nodes from MS.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/causalitydatanodes/CausalityDataNodesEngine.java
```diff
@@ -147,41 +148,32 @@ public CompletableFuture<Set<String>> dataNodes(long causalityToken, int zoneId)
...
-                // So the data nodes value will be equals to the logical topology on the lastCfgRevision.
```
sergeyuttsel, line 182: I think we don't need all code related to the augmentation map in the CausalityDataNodesEngine#dataNodes method. This code is needed for zones with immediate timers. But now the initDataNodesAndTriggerKeysInMetaStorage() watch listener updates augmentation and returns the future which is completed when the data nodes is updated in MS. So we don't need to add/remove nodes from the augmentation map to data nodes from MS.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/causalitydatanodes/CausalityDataNodesEngine.java
```diff
@@ -310,32 +305,28 @@ private long getLastScaleDownConfigRevision(
      */
     private long getLastConfigRevision(
             long causalityToken,
```
sergeyuttsel, line 307: causalityToken is unused in this and other methods.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/causalitydatanodes/CausalityDataNodesEngine.java
```diff
@@ -98,39 +93,45 @@ public class CausalityDataNodesEngine {
...
      * {@link DistributionZoneNotFoundException} if the zone with the provided zoneId does not exist.
      *
      * @param causalityToken Causality token.
+     * @param catalogVersion Catalog version.
      * @param zoneId Zone id.
      * @return The future with data nodes for the zoneId.
      */
-    public CompletableFuture<Set<String>> dataNodes(long causalityToken, int zoneId) {
+    public CompletableFuture<Set<String>> dataNodes(long causalityToken, int catalogVersion, int zoneId) {
         if (causalityToken < 1) {
```
sergeyuttsel, line 135: I think we need to add similar validation for catalogVersion.

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/causalitydatanodes/CausalityDataNodesEngine.java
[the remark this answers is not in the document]
alievmirza: From my point of view, it is still needed to have such case, because some other component can listen to zone creation event and ask for dataNodes with the causality token of the creation of a zone, and because the order of listener notifications is undefined, we can be in a situation, when data nodes is not initialised, but component asked for dataNodes

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/causalitydatanodes/CausalityDataNodesEngine.java
alievmirza: I think that this case is similar to this answer https://github.com/apache/ignite-3/pull/2873#discussion_r1410523946

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1070,7 +1070,7 @@ private CompletableFuture<?> createTableLocally(
             if (partitionAssignments(vaultManager, tableId, 0) != null) {
                 assignmentsFuture = completedFuture(tableAssignments(vaultManager, tableId, zoneDescriptor.partitions()));
             } else {
-                assignmentsFuture = distributionZoneManager.dataNodes(causalityToken, zoneId)
+                assignmentsFuture = distributionZoneManager.dataNodes(causalityToken, catalogVersion, zoneId)
```
sanpwc, line 1073: It might be a bit fragile. Let's assume that tableManager.start() is partially asynchronous. In that case we may start new events processing while recovery on restart isn't fully finished. Generally speaking it'll be a disaster. But within given context we may finish with non-matching token and catalogVersion.

sanpwc (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
alievmirza: From my point of view, we must forbid such situations

PR #2883, opened 2023-11-27
IGNITE-20875 Add enlistment consistency token to PrimaryReplicaReques…

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-20875


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -3355,32 +3356,20 @@ private CompletableFuture<IgniteBiTuple<RowId, Collection<Lock>>> takeLocksForRe
...
 
-        // TODO: IGNITE-20875 Add enlistment consistency token to PrimaryReplicaTestRequest interface.
-        if (request instanceof ReadWriteReplicaRequest) {
-            expectedTerm = ((ReadWriteReplicaRequest) request).term();
-
-            assert expectedTerm != null;
-        } else if (request instanceof TxFinishReplicaRequest) {
-            expectedTerm = ((TxFinishReplicaRequest) request).term();
+        if (request instanceof PrimaryReplicaRequest) {
+            expectedTerm = ((PrimaryReplicaRequest) request).enlistmentConsistencyToken();
```
sanpwc, line 3362: It's not longer expected**Term** but rather an expectedEnlistmentConsistencyToken, right?
Same for variable declaration.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -3355,32 +3356,20 @@ private CompletableFuture<IgniteBiTuple<RowId, Collection<Lock>>> takeLocksForRe
...
-            assert expectedTerm != null;
-        } else if (request instanceof ReadOnlyDirectReplicaRequest) {
-            expectedTerm = ((ReadOnlyDirectReplicaRequest) request).enlistmentConsistencyToken();
-
-            assert expectedTerm != null;
-        } else if (request instanceof BuildIndexReplicaRequest) {
-            expectedTerm = ((BuildIndexReplicaRequest) request).enlistmentConsistencyToken();
-        } else if (request instanceof TxRecoveryMessage) {
-            expectedTerm = ((TxRecoveryMessage) request).enlistmentConsistencyToken();
+            assert expectedTerm != 0 :
```
sanpwc, line 3364: It's tricky. Why not to use Long instead of long and thus check whether ist's null or not? I mean that it's never specified that 0 is a special value for the token.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/message/PrimaryReplicaRequest.java
```diff
@@ -21,4 +21,11 @@
  * This is an interface for requests are designed for primary replica.
  */
 public interface PrimaryReplicaRequest extends ReplicaRequest {
+    /**
+     * Gets an enlistment consistency token.
+     * The token is used to check that the lease is still actual while the message goes to the replica.
+     *
+     * @return Enlistment consistency token.
+     */
+    long enlistmentConsistencyToken();
```
sanpwc, line 30: Why it's long and not Long? Originally it was marked as @Nullable and by the fact that you've added an assert within `ensureReplicaIsPrimary` I believe that it's no longer true. Thus I'd rather have it Long without @Nullable - in that case we will check whether proper value is propagated to a message or not.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/message/TxRecoveryMessage.java
```diff
@@ -19,26 +19,17 @@
...
 import org.apache.ignite.internal.replicator.message.PrimaryReplicaRequest;
-import org.apache.ignite.internal.replicator.message.ReplicaRequest;
 import org.apache.ignite.network.annotations.Transferable;
 
 /**
  * Transaction recovery message.
  */
 @Transferable(TxMessageGroup.TX_RECOVERY_MSG)
-public interface TxRecoveryMessage extends PrimaryReplicaRequest, ReplicaRequest {
+public interface TxRecoveryMessage extends PrimaryReplicaRequest {
```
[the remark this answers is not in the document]
vldpyatkov: Because each request requires a response in the sense of the name. TxRecoveryMessage does not have a response. This is just a network message.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
vldpyatkov: I prefer to use long instead of Long for two reasons:
1) long is significantly less in serializable representation;
2) I don't like to rely on automated checkers.
But because you think using Long is a more pritty way, I changed it.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/message/PrimaryReplicaRequest.java
vldpyatkov: Look at the comment above.

PR #2900, opened 2023-11-29
IGNITE-20900 .NET: Handle replica assignment timestamp

ptupitsyn:


isapego (approved):
I like it how the code and the logic became more clear and less hacky, actually.

PR #2906, opened 2023-11-30
IGNITE-20661 JDBC. Support multi statement queries by JDBC

zstan:

### The Review Checklist

### Notes


zstan:
@korlov42 can u make a review plz ?

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
```diff
@@ -396,7 +398,7 @@ public ResultSet getResultSet() throws SQLException {
 
         JdbcResultSet rs = resSets.get(curRes);
 
-        if (!rs.isQuery()) {
+        if (rs == null || !rs.isQuery()) {
```
korlov42, line 401: if null is considering as legit element of `resSets` collection, then we need to mark generic type as `@Nullable`

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
```diff
@@ -414,6 +416,10 @@ public int getUpdateCount() throws SQLException {
 
         JdbcResultSet rs = resSets.get(curRes);
 
+        if (rs == null) {
+            rs = resSets.get(curRes - 1);
+        }
```
korlov42, line 421: why do we get the previous result set here?

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
```diff
@@ -432,10 +438,14 @@ public boolean getMoreResults() throws SQLException {
     public boolean getMoreResults(int curr) throws SQLException {
```
korlov42, line 438: don't know where to put this, so let it be here:

for sequence of DML, expression `!stmt.getMoreResults() && stmt.getUpdateCount() == -1` never returns `true` (that is now we determine there are no more results, see javadoc for `getMoreResults`)

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
```diff
@@ -454,11 +464,11 @@ public boolean getMoreResults(int curr) throws SQLException {
                     throw new SQLFeatureNotSupportedException("Multiple open results is not supported.");
```
korlov42, line 464: whe don't we support this?

korlov42 (changes requested):

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -363,29 +375,31 @@ private CompletionStage<JdbcQuerySingleResult> createJdbcResult(AsyncSqlCursor<L
         return cur.requestNextAsync(req.pageSize()).thenApply(batch -> {
             boolean hasNext = batch.hasMore();
 
+            long cursorId;
```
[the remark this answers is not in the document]
zstan: reflection usage code is appended

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcResultSet.java
```diff
@@ -201,6 +205,37 @@ public JdbcResultSet(List<List<Object>> rows, List<JdbcColumnMeta> meta) throws
...
+            long updCount = 0;
+
+            if (res.isDml()) {
+                CompletableFuture<JdbcQueryFetchResult> fetch = cursorHandler.fetchAsync(
+                        new JdbcQueryFetchRequest(cursorId0, 1));
+                JdbcQueryFetchResult fetchRes = fetch.get();
+                updCount = (long) fetchRes.items().get(0).get(0);
+            }
+
+            return new JdbcResultSet(cursorHandler, stmt, cursorId0, getFetchSize(), false, Collections.emptyList(), res.isQuery(),
```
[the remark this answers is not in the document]
zstan: I misunderstand here, if you suggest to call fetchAsync inside  of getMoreResultsAsync then we obtain blocking code, otherwise i miss your suggestion, can you show where i can obtain first page results plz ?

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
zstan: reworked

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
zstan: fixed

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcResultSet.java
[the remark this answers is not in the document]
zstan: oh, catch it, still in progress here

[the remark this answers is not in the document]
zstan: but .. in case if some one not planned to use this prefetched results - we will make an extra work\send.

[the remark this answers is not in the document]
korlov42: well, you've done this already for DML (by the cost of extra round trip though)... And regarding just a query: why one might execute SELECT statement if they don't need a result? I believe the common case here is that every result will be consumed, thus prefetching the first page seems like a legit optimisation of the common case.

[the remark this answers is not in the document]
zstan: Amen, i fix it )

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcMultiStatementSelfTest.java
```diff
@@ -52,48 +54,229 @@ public void setupTables() throws Exception {
...
-                + "INSERT INTO public.transactions VALUES (1,2,'some', 'word') ; "
-                + "CREATE INDEX transactions_k_v_id ON public.transactions (k, v, id) INLINE_SIZE 150; "
-                + "CREATE INDEX transactions_pk_id ON public.transactions (pk, id) INLINE_SIZE 20;");
+    public void testSimpleQueryExecute() throws Exception {
+        int initial = openCursorsRegistered();
+
+        boolean res = stmt.execute("INSERT INTO TEST_TX VALUES (5, 5, '5');");
+        assertEquals(false, res);
+        stmt.getResultSet();
+        stmt.getMoreResults();
```
xtern, line 70:
```suggestion
        assertFalse(res);
        assertNull(stmt.getResultSet());
        assertFalse(stmt.getMoreResults());
```
:thinking:

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcStatementSelfTest.java
```diff
@@ -783,4 +779,33 @@ public void testStatementTypeMismatchUpdate() throws Exception {
...
+        ResultSet rs = stmt.getResultSet();
+        stmt.execute("SELECT 3;");
+        assertTrue(rs.isClosed());
+
+        assertTrue(populateStmtCnt < 100);
+        //more than one fetch request
+        for (int i = populateStmtCnt; i < stmt.getMaxRows() + 100; ++i) {
+            stmt.execute(String.format("INSERT INTO TEST VALUES (%d, '1')", i));
+        }
+        //stmt.close();
```
xtern, line 796:
```suggestion

```

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcMultiStatementSelfTest.java
```diff
@@ -17,21 +17,23 @@
...
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 /**
- * Tests for ddl queries that contain multiply sql statements, separated by ";".
+ * Tests for queries that contain multiply sql statements, separated by ";".
```
xtern, line 35: please add some test(s) for autocommit=false (without TX_CONTROL statements) seems this should work fine,

xtern, line 35: also, this is out of scope, since IGNITE-20463 delayed 😭 , but it would be great if you investigate how to handle TxControlInsideExternalTxNotSupportedException (that will be introduced in IGNITE-20463). may be you can add muted test that will show the desired behavior 😞
```
conn.setAutoCommit(false)

execute("START TRANSACTION; COMMIT")

// query processor will throw TxControlInsideExternalTxNotSupportedException, but user must see something like "transaction control statements are not supported in non autocommit mode" 🤔
```
(see https://issues.apache.org/jira/browse/IGNITE-21020 )

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcMultiStatementSelfTest.java
zstan: I understand you, but this case is not for check intermediate statement commands but about correct results after multiple calls, but ok, i fix it, it makes no harm.

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcMultiStatementSelfTest.java
zstan: i think we need to check it after IGNITE-20463

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcStatementSelfTest.java
```diff
@@ -783,4 +779,27 @@ public void testStatementTypeMismatchUpdate() throws Exception {
...
+
+        assertTrue(populateStmtCnt < 100);
+        //more than one fetch request
+        for (int i = populateStmtCnt; i < stmt.getMaxRows() + 100; ++i) {
+            stmt.execute(String.format("INSERT INTO TEST VALUES (%d, '1')", i));
+        }
+    }
+
+    @Test
+    public void testOpenCursorsWithDdl() throws Exception {
```
korlov42, line 798: why doesn't this test have a single assertion?

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcStatementSelfTest.java
```diff
@@ -783,4 +779,27 @@ public void testStatementTypeMismatchUpdate() throws Exception {
...
+        stmt.execute("SELECT 1; SELECT 2;");
+        ResultSet rs = stmt.getResultSet();
+        stmt.execute("SELECT 3;");
+        assertTrue(rs.isClosed());
+
+        assertTrue(populateStmtCnt < 100);
+        //more than one fetch request
+        for (int i = populateStmtCnt; i < stmt.getMaxRows() + 100; ++i) {
+            stmt.execute(String.format("INSERT INTO TEST VALUES (%d, '1')", i));
+        }
```
korlov42, line 794: what is the purpose of this loop?

modules/client-common/src/main/java/org/apache/ignite/internal/jdbc/proto/JdbcQueryCursorHandler.java
```diff
@@ -37,6 +39,14 @@ public interface JdbcQueryCursorHandler {
      */
     CompletableFuture<JdbcQueryFetchResult> fetchAsync(JdbcQueryFetchRequest req);
 
+    /**
+     * {@link Statement#getMoreResults()} command implementor.
+     *
+     * @param req Results request.
+     * @return Result future.
+     */
+    CompletableFuture<JdbcQuerySingleResult> getMoreResultsAsync(JdbcQueryFetchRequest req);
```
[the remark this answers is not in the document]
zstan: probably we need to rename somehow ? but hold additional equal class just different action seems not so good too, wdyt ?

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcResultSet.java
```diff
@@ -215,6 +218,37 @@ public JdbcResultSet(List<List<Object>> rows, List<JdbcColumnMeta> meta) throws
...
     }
 
+    @Nullable JdbcResultSet getNextResultSet() throws SQLException {
+        try {
+            JdbcQueryFetchRequest req = new JdbcQueryFetchRequest(cursorId, fetchSize);
+            JdbcQuerySingleResult res = cursorHandler.getMoreResultsAsync(req).get();
+
+            close0(true);
+
+            if (!res.hasResults()) {
```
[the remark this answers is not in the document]
zstan: append test

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcMultiStatementSelfTest.java
```diff
@@ -52,48 +54,231 @@ public void setupTables() throws Exception {
...
-                + "CREATE INDEX transactions_pk_id ON public.transactions (pk, id) INLINE_SIZE 20;");
+    public void testSimpleQueryExecute() throws Exception {
+        int initial = openCursorsRegistered();
+
+        boolean res = stmt.execute("INSERT INTO TEST_TX VALUES (5, 5, '5');");
+        assertEquals(false, res);
+        assertFalse(res);
+        assertNull(stmt.getResultSet());
+        assertFalse(stmt.getMoreResults());
+        assertEquals(-1, getResultSetSize());
```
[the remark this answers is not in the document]
zstan: change `getResultSetSize` javadoc

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcResultSet.java
```diff
@@ -215,6 +218,37 @@ public JdbcResultSet(List<List<Object>> rows, List<JdbcColumnMeta> meta) throws
         initColumnOrder();
     }
 
+    @Nullable JdbcResultSet getNextResultSet() throws SQLException {
+        try {
+            JdbcQueryFetchRequest req = new JdbcQueryFetchRequest(cursorId, fetchSize);
+            JdbcQuerySingleResult res = cursorHandler.getMoreResultsAsync(req).get();
+
+            close0(true);
```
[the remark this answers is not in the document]
zstan: add test

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcPreparedStatement.java
```diff
@@ -762,7 +762,7 @@ private void executeWithArguments(JdbcStatementType statementType) throws SQLExc
         Object[] args = currentArgs == null ? ArrayUtils.OBJECT_EMPTY_ARRAY :
                 currentArgs.stream().map(this::convertJdbcTypeToInternal).toArray();
 
-        execute0(statementType, sql, args);
+        execute0(statementType, sql, false, args);
```
xtern, line 765: I don't know where to put this question.

Why don't we allow script execution using PreparedStatement#execute method?

zstan: yep, u a right, fixed

modules/client-common/src/main/java/org/apache/ignite/internal/jdbc/proto/JdbcQueryCursorHandler.java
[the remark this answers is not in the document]
korlov42: I don't mind to rename it, but have no good options for name. 

btw, what is wrong with new class?

modules/client-common/src/main/java/org/apache/ignite/internal/jdbc/proto/event/JdbcQuerySingleResult.java
```diff
@@ -69,15 +69,30 @@ public JdbcQuerySingleResult(int status, String err) {
...
      * @param rowTuples Serialized SQL result rows.
+     * @param columnTypes Ordered list of types of columns in serialized rows.
+     * @param decimalScales Decimal scales in appearance order.
+     * @param isQuery {@code true} if query is SELECT/EXPLAIN.
+     * @param updateCnt Update count.
      * @param last     Flag indicates the query has no unfetched results.
      */
     public JdbcQuerySingleResult(long cursorId, List<BinaryTupleReader> rowTuples, List<ColumnType> columnTypes, int[] decimalScales,
-            boolean last) {
+            boolean last, boolean isQuery, long updateCnt) {
```
korlov42, line 95: we already have contractor for DML statements, so let's use that one

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -369,56 +380,75 @@ private CompletionStage<JdbcQuerySingleResult> createJdbcResult(AsyncSqlCursor<I
...
-                            decimalScales[countOfDecimal++] = column.scale();
-                        }
-                    }
-                    decimalScales = Arrays.copyOf(decimalScales, countOfDecimal);
-
-                    return new JdbcQuerySingleResult(cursorId, rows, schema, decimalScales, !hasNext);
+                    return buildSingleRequest(batch, columns, cursorId, !hasNext, cur.queryType());
                 }
-                case DML:
+                case DML: {
```
korlov42, line 400: if DML is handled by `buildSingleRequest` as well, then we should merge these branches

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -369,56 +380,75 @@ private CompletionStage<JdbcQuerySingleResult> createJdbcResult(AsyncSqlCursor<I
...
+    static JdbcQuerySingleResult buildSingleRequest(
+            BatchedResult<InternalSqlRow> batch,
+            List<ColumnMetadata> columns,
+            long cursorId,
+            boolean hasNext,
+            SqlQueryType queryType
+    ) {
+        List<BinaryTupleReader> rows = new ArrayList<>(batch.items().size());
+        for (InternalSqlRow item : batch.items()) {
+            rows.add(item.asBinaryTuple());
```
korlov42, line 427: we don't need to do it for DML

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcResultSet.java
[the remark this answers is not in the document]
korlov42: probably I'm missing something, but I still no sight of error handling. Could you point out to me on particular test that covers failed response?

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcResultSet.java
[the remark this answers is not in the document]
korlov42: I see you've added `org.apache.ignite.jdbc.ItJdbcMultiStatementSelfTest#testCloseOnCompletion`, but I don't quite understand how it is suppose to relate to my comment...

modules/client-common/src/main/java/org/apache/ignite/internal/jdbc/proto/event/JdbcQuerySingleResult.java
zstan: `JdbcQuerySingleResult(boolean hasNext, long updCount)` is not for DML needs for now.

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcResultSet.java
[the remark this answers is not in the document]
zstan: ItJdbcMultiStatementSelfTest#testSimpleQueryError got an error from handler, or you are talking about something more ?

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcMultiStatementSelfTest.java
```diff
@@ -17,21 +17,25 @@
...
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 /**
- * Tests for ddl queries that contain multiply sql statements, separated by ";".
+ * Tests for queries containing multiple sql statements, separated by ";".
```
xtern, line 37: It seems to me that we can unmute the `testEmptyStatements` test?

I also haven't found a test that checks the following scenario.
```
START TRANSASCTION; // RW TX
SELECT * FROM TABLE(system_range(0, 2000))  // tab size must be greater than a "fetch size"
COMMIT; // blocks until previous cursor is closed
START TRANSASCTION;
SELECT 2; 
COMMIT;
```
the user should be able to switch to SELECT statement 2 without explicit close/read the first select statement

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcMultiStatementSelfTest.java
```diff
@@ -52,48 +56,321 @@ public void setupTables() throws Exception {
...
+        stmt.getResultSet();
+        ResultSet rs = stmt.getResultSet();
+
+        res = stmt.getMoreResults();
+        assertEquals(true, res);
+
+        assertTrue(rs.isClosed());
+        assertEquals(1, getResultSetSize());
+
+        stmt.execute("SELECT 1; SELECT 2; SELECT 3;");
```
xtern, line 253: I see that the following test case hangs
```
        stmt.execute("START TRANSACTION; SELECT * FROM TABLE(system_range(0, 2000)); COMMIT;");
        stmt.getMoreResults();
        stmt.getMoreResults();
```
and the following finished with error
```
        stmt.execute("START TRANSACTION; SELECT * FROM TABLE(system_range(0, 2000)); COMMIT;");
        stmt.getMoreResults();
        stmt.getResultSet().close();
        stmt.getMoreResults(); //  Failed to find resource with id: 6
```
Can you investigate this issue?

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcMultiStatementSelfTest.java
```diff
@@ -52,48 +56,321 @@ public void setupTables() throws Exception {
...
+        assertArrayEquals(new int[]{1, 1}, arr);
+        assertEquals(-1, stmt.getUpdateCount());
+        assertFalse(stmt.getMoreResults());
     }
 
     /**
      * Sanity test for scripts, containing empty statements are handled correctly.
      */
     @Test
+    @Disabled("https://issues.apache.org/jira/browse/IGNITE-21081")
```
xtern, line 357: I think we can enable this test, we have already discussed that there is only one problematic case here ";" (empty statement), we have test for this IgniteSqlParaserTest#testEmptyStatements.
Other cases should work fine.

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcMultiStatementSelfTest.java
zstan: tx scenario is fixed, appropriate test appended.

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcMultiStatementSelfTest.java
zstan: what about : ";;;SELECT 1 + 2", seems it valid statement ?

xtern: I've updated description in the linked ticket.

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcResultSet.java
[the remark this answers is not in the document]
zstan: fix tests: noMoreResultsArePossibleAfterCloseOnCompletion, testCloseOnCompletionFirstRsClosed

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcResultSet.java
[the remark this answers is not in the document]
zstan: as for now - we stop further script execution if exception is raised, i.e. 
```
CREATE TABLE test (id INT PRIMARY KEY);
CREATE TABLE test (id INT PRIMARY KEY);
SELECT 1;
```
we can`t obtain SELECT 1 results, thus it seems enough, or you think that some mock are still necessary here ?

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryCursorHandlerImpl.java
```diff
@@ -91,6 +95,49 @@ public CompletableFuture<JdbcQueryFetchResult> fetchAsync(JdbcQueryFetchRequest
...
+    /** {@inheritDoc} */
+    @Override
+    public CompletableFuture<JdbcQuerySingleResult> getMoreResultsAsync(JdbcFetchQueryResultsRequest req) {
+        AsyncSqlCursor<InternalSqlRow> asyncSqlCursor;
+        try {
+            asyncSqlCursor = resources.get(req.cursorId()).get(AsyncSqlCursor.class);
+        } catch (IgniteInternalCheckedException e) {
+            StringWriter sw = getWriterWithStackTrace(e);
+
+            return CompletableFuture.completedFuture(new JdbcQuerySingleResult(Response.STATUS_FAILED,
```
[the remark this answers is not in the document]
zstan: yep, i see it but it not relates to this issue.

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
```diff
@@ -472,7 +462,7 @@ public boolean getMoreResults(int curr) throws SQLException {
             switch (curr) {
                 case CLOSE_CURRENT_RESULT:
                     if (curRes > 0) {
-                        resSets.get(curRes - 1).close0();
+                        resSets.get(curRes - 1).close0(false);
```
xtern, line 465: Why can't we call `close0(true)` here?

The following example demonstrate resource leakage (or may be I missed something):
```
    // test can be added to ItJdbcMultiStatementSelfTest
    @Test
    public void testX() throws Exception {
        long initial = openCursorsRegistered();
        
        stmt.executeUpdate("CREATE TABLE TESTX(ID INT PRIMARY KEY)");
        stmt.getMoreResults();
        stmt.close();

        assertEquals(0L, initial - openCursorsRegistered());
    }
```

As I see, the current combination in `close0` of `isQuery`, `finished`, etc. flags were being used because we were not registering cursors for DML, DDL (and possibly other) queries.

Since we are now registering cursors for this types of queries, I think it is worth reviewing the current conditions in `close0`, because now they look too complex and leading to errors.

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcResultSet.java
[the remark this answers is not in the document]
zstan: done, append additional handler test

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
zstan: this is connection context , append below to test and it will pass:
```
        conn.close();
        assertTrue(waitForCondition(() -> 0L == openCursorsRegistered(), 5000));
```

xtern: Yes you are right. I missed this.
But as I see, now - getNextResultSet() already closes resultset with enforce=true, why we need this second call here?

zstan: done

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
zstan: getResultSet() marked as nullable, @Nullable JdbcResultSet rs = too, did i miss smth ?

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcResultSet.java
```diff
@@ -265,7 +307,7 @@ public boolean next() throws SQLException {
     /** {@inheritDoc} */
     @Override
     public void close() throws SQLException {
-        close0();
+        closed = true;
```
korlov42, line 310: all resources must be cleaned up as soon as possible, it's not enough to just pretend that result set is closed

modules/client-handler/src/test/java/org/apache/ignite/client/handler/JdbcQueryCursorHandlerImplTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+        CompletableFuture<JdbcQuerySingleResult> fut = IgniteTestUtils.runAsync(() ->
+                await(cursorHandler.getMoreResultsAsync(new JdbcFetchQueryResultsRequest(1, 100)), 5, TimeUnit.SECONDS)
+        );
+
+        try {
+            await(fut, 5, TimeUnit.SECONDS);
+        } catch (Throwable e) {
+            fail("Unexpected exception is raised.");
+        }
```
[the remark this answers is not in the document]
zstan: I verify that JdbcQueryCursorHandlerImpl correctly handles exceptions raised from AsyncSqlCursor, do you think it`s redundant ?

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
zstan: i found that resSets has volatile notation seems our Statement implementation is not a thread safe at all.
apeend @Nullable, fix nullability checks

modules/client-common/src/main/java/org/apache/ignite/internal/jdbc/proto/event/JdbcQuerySingleResult.java
zstan: done

modules/client-common/src/main/java/org/apache/ignite/internal/jdbc/proto/event/JdbcQuerySingleResult.java
```diff
@@ -69,15 +69,30 @@ public JdbcQuerySingleResult(int status, String err) {
...
 
+    /**
+     * Constructor.
+     *
+     * @param hasNext {@code true} if more results are present.
+     * @param updCount Update counter.
+     */
+    public JdbcQuerySingleResult(boolean hasNext, long updCount) {
+        hasResults = hasNext;
+        this.updateCnt = updCount;
```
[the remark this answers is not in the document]
zstan: done

modules/client-common/src/main/java/org/apache/ignite/internal/jdbc/proto/event/JdbcQuerySingleResult.java
```diff
@@ -179,6 +197,7 @@ public long updateCount() {
     @Override
     public void writeBinary(ClientMessagePacker packer) {
         super.writeBinary(packer);
+        packer.packLong(updateCnt);
```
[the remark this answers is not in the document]
zstan: done

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryCursorHandlerImpl.java
```diff
@@ -91,6 +95,51 @@ public CompletableFuture<JdbcQueryFetchResult> fetchAsync(JdbcQueryFetchRequest
...
+                .thenCompose(cur -> cur.requestNextAsync(req.fetchSize())
+                    .thenApply(batch -> {
+                        try {
+                            SqlQueryType queryType = cur.queryType();
+
+                            long cursorId = resources.put(new ClientResource(cur, cur::closeAsync));
+
+                            List<ColumnMetadata> columns = cur.metadata().columns();
+
+                            return buildSingleRequest(batch, columns, cursorId, !batch.hasMore(), queryType);
```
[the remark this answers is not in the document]
zstan: done

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
```diff
@@ -460,33 +440,53 @@ public boolean getMoreResults() throws SQLException {
...
+
+        resSets.add(nextResultSet);
+
+        curRes++;
+
+        if (exceptionally != null) {
+            throw exceptionally;
+        }
+
+        return nextResultSet != null && nextResultSet.holdResults();
```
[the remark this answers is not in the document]
zstan: changed, not here but logic at all

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
```diff
@@ -460,33 +440,53 @@ public boolean getMoreResults() throws SQLException {
...
-
                     break;
 
                 case CLOSE_ALL_RESULTS:
                 case KEEP_CURRENT_RESULT:
                     throw new SQLFeatureNotSupportedException("Multiple open results is not supported.");
 
                 default:
-                    throw new SQLException("Invalid 'current' parameter.");
+                    throw new SQLException("Invalid 'curr' parameter.");
```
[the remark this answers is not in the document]
zstan: done

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcResultSet.java
zstan: done

zstan:
@xtern @korlov42 guys, seems i finish. There is 1 question about test from me, check above. Nullability checks .. idea due to volatile notation highlight all places there this value used no matter exist previous _not null_ check or not.

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
```diff
@@ -711,10 +708,37 @@ void ensureNotClosed() throws SQLException {
...
+                if (last != null) {
+                    while (last != null) {
+                        try {
+                            last = last.getNextResultSet();
+                        } catch (SQLException ex) {
+                            // TODO: https://issues.apache.org/jira/browse/IGNITE-21133
+                            // implicitly silent close previous result set.
+                        }
+                    }
+                }
```
xtern, line 741: As was discussed today - let's throw error in such situation (better late than never).

```suggestion
                do {
                    lastRs = lastRs.getNextResultSet();
                } while (lastRs != null);
```

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcMultiStatementSelfTest.java
```diff
@@ -52,48 +59,443 @@ public void setupTables() throws Exception {
...
+
+        // pk violation exception
+        // TODO: https://issues.apache.org/jira/browse/IGNITE-21133
+        stmt.execute("START TRANSACTION; INSERT INTO TEST_TX VALUES (1, 1, '1'); COMMIT");
+        stmt.execute("SELECT COUNT(*) FROM TEST_TX");
+        try (ResultSet rs = stmt.getResultSet()) {
+            assertTrue(rs.next());
+            assertEquals(4, rs.getInt(1));
+        }
+    }
```
xtern, line 92:
```suggestion
    @Test
    @Disabled("https://issues.apache.org/jira/browse/IGNITE-21133")
    public void testAllStatementsAppliedIfExecutedWithFailure() throws Exception {
        stmt.execute("SELECT COUNT(*) FROM TEST_TX");
        try (ResultSet rs = stmt.getResultSet()) {
            assertTrue(rs.next());
            assertEquals(4, rs.getInt(1));
        }

        // pk violation exception
        assertThrows(SQLException.class, () -> stmt.execute("START TRANSACTION; INSERT INTO TEST_TX VALUES (1, 1, '1'); COMMIT"));
        stmt.execute("SELECT COUNT(*) FROM TEST_TX");
        try (ResultSet rs = stmt.getResultSet()) {
            assertTrue(rs.next());
            assertEquals(4, rs.getInt(1));
        }
    }
```

zstan: done

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
zstan: done

modules/client-common/src/main/java/org/apache/ignite/internal/jdbc/proto/event/JdbcQuerySingleResult.java
```diff
@@ -53,10 +51,14 @@ public class JdbcQuerySingleResult extends Response {
     /** Decimal scales in appearance order. Can be empty in case no any decimal columns. */
     private int[] decimalScales;
 
+    /** {@code true} if results are available, {@code false} otherwise. */
+    private boolean hasResult = true;
```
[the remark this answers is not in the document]
zstan: changed

modules/client-common/src/main/java/org/apache/ignite/internal/jdbc/proto/event/JdbcQuerySingleResult.java
```diff
@@ -184,6 +169,11 @@ public boolean isQuery() {
         return isQuery;
     }
 
+    /** Results availability flag. */
```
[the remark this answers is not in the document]
zstan: done

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
```diff
@@ -711,10 +708,25 @@ void ensureNotClosed() throws SQLException {
...
+            boolean allFetched = lastRs == null || lastRs.isClosed();
+
+            if (allFetched) {
+                for (JdbcResultSet rs : resSets) {
+                    if (rs != null) {
+                        rs.close0(true);
+                    }
+                }
+            } else {
+                last = lastRs.getNextResultSet();
```
xtern, line 725: as I see it, we also need to reset `resSets` and `curRes` in case of error, otherwise the next stmt.close() will throw the same error.

xtern (approved):

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/JdbcStatement.java
ygerzhedovich: should be fixed

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcMultiStatementSelfTest.java
ygerzhedovich: seems need to be addressed

ygerzhedovich: need to be addressed

modules/client-common/src/main/java/org/apache/ignite/internal/jdbc/proto/event/JdbcQuerySingleResult.java
```diff
@@ -180,18 +192,22 @@ public long updateCount() {
...
-            return;
+        packer.packBoolean(resultsAvailable);
+        if (resultsAvailable) {
+            packer.packLong(updateCnt);
+
+            if (cursorId != null) {
+                packer.packLong(cursorId);
+            } else {
+                packer.packNil();
+            }
```
ygerzhedovich, line 203:
```suggestion
packer.packLongNullable(cursorId);
```

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcMultiStatementSelfTest.java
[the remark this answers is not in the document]
ygerzhedovich: Fully support aforementioned comments. Why don't check update Count explicitly?

modules/client-handler/src/test/java/org/apache/ignite/client/handler/JdbcQueryCursorHandlerImplTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+                    @Override
+                    public CompletableFuture<Void> closeAsync() {
+                        return nullCompletedFuture();
+                    }
+                };
+            }
+        });
+
+        try {
```
ygerzhedovich, line 131: what reason to have the try/catch block?

zstan: look, this test check that no exception is raised from JdbcQueryCursorHandlerImpl#getMoreResultsAsync
comment exception handling in mention method and try to run test.

ygerzhedovich (approved):

PR #2923, opened 2023-12-05
IGNITE-20771 Implement tx coordinator liveness check

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-20771

### DIFF
--- a/modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxStateMeta.java
+++ b/modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxStateMeta.java
@@ -17,10 +17,13 @@
 
 package org.apache.ignite.internal.tx;
 
+import static org.apache.ignite.internal.tx.TxState.ABANDONED;
+
 import java.util.Objects;
 import org.apache.ignite.internal.hlc.HybridTimestamp;
 import org.apache.ignite.internal.replicator.TablePartitionId;
 import org.apache.ignite.internal.tostring.S;
+import org.apache.ignite.internal.util.FastTimestamps;
 import org.jetbrains.annotations.Nullable;
 
 /**
@@ -38,6 +41,9 @@ public class TxStateMeta implements TransactionMeta {
 
     private final HybridTimestamp commitTimestamp;
 
+    /** Timestamp when the latest {@code ABANDONED} state set. If the transaction state is not {@code ABANDONED}, it is {@code 0}. */
+    private final long lastAbandonedMarkerTs;
+
     /**
      * Constructor.
      *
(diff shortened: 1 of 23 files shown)


modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/OrphanDetector.java
```diff
@@ -151,26 +174,37 @@ private CompletableFuture<Void> handleLockHolder(UUID txId) {
...
+                    "Conflict was found, and the coordinator of the transaction that holds a lock is not available "
+                            + "[txId={}, txCrd={}].",
+                    txId,
+                    txState.txCoordinatorId()
+            );
+
+            sentTxRecoveryMessage(txState.commitPartitionId(), txId);
+        }
+
+        throw new TransactionException(ACQUIRE_LOCK_ERR, "The lock is held by the abandoned transaction [tx=" + txId + "].");
```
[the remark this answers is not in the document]
vldpyatkov: Checked _ts={}_ to _abandonedTx={}_.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxStateMeta.java
```diff
@@ -51,11 +57,44 @@ public TxStateMeta(
...
         this.commitTimestamp = commitTimestamp;
+        this.lastAbandonedMarkerTs = lastAbandonedMarkerTs;
+    }
+
+    /**
+     * Creates a transaction state for same transaction, but this one is marked abandoned.
+     *
+     * @return Transaction state meta.
+     */
+    public TxStateMeta markAbandoned() {
```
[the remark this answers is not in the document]
vldpyatkov: Technically, this method creates a new object that is described in javadoc, but semantically, it is a similar object (for the same transaction) with ABANDONED state.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/OrphanDetector.java
```diff
@@ -151,26 +176,37 @@ private CompletableFuture<Void> handleLockHolder(UUID txId) {
...
+            sentTxRecoveryMessage(txState.commitPartitionId(), txId);
         }
 
-        LOG.info(
-                "Conflict was found, and the coordinator of the transaction that holds a lock is not available "
-                        + "[txId={}, txCrd={}].",
-                txId,
-                txState.txCoordinatorId()
-        );
+        throw new TransactionException(ACQUIRE_LOCK_ERR, "The lock is held by the abandoned transaction [abandonedTx=" + txId + "].");
```
denis-chudov, line 198:
```suggestion
        throw new TransactionException(ACQUIRE_LOCK_ERR, "The lock is held by the abandoned transaction [abandonedTxId=" + txId + "].");
```

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxStateMeta.java
```diff
@@ -51,13 +55,54 @@ public TxStateMeta(
...
+     * Creates a transaction state for the same transaction, but this one is marked finishing.
+     *
+     * @return Transaction state meta.
+     */
+    public TxStateMetaFinishing markFinishing() {
+        assert checkTransitionCorrectness(txState, FINISHING) : "Transaction state is incorrect [txState=" + txState + "].";
+
+        return new TxStateMetaFinishing(txCoordinatorId, commitPartitionId);
+    }
+
```
denis-chudov, line 105: I mean, we use the same naming in *Listener classes, to actually mark the transactions with some state in the local state map. Here it is not marking, it's a creation of a new meta. Let's name it simply 'abandoned` and `finishing`.

vldpyatkov: Renamed.

denis-chudov (approved):

denis-chudov (approved):

PR #2924, opened 2023-12-06
IGNITE-20845 Compute jobs cancellation

valepakh:
https://issues.apache.org/jira/browse/IGNITE-20845

### The Review Checklist

### Notes

### DIFF
--- a/modules/api/src/main/java/org/apache/ignite/compute/JobExecution.java
+++ b/modules/api/src/main/java/org/apache/ignite/compute/JobExecution.java
@@ -15,33 +15,32 @@
  * limitations under the License.
  */
 
-package org.apache.ignite.internal.compute;
+package org.apache.ignite.compute;
 
-import java.util.concurrent.atomic.AtomicBoolean;
-import org.apache.ignite.Ignite;
-import org.apache.ignite.compute.JobExecutionContext;
+import java.util.concurrent.CompletableFuture;
 
 /**
- * Implementation of {@link JobExecutionContext}.
+ * Represents job's execution.
+ *
+ * @param <R> Job result type.
  */
-public class JobExecutionContextImpl implements JobExecutionContext {
-    private final Ignite ignite;
+public interface JobExecution<R> {
+    /**
+     * Returns job's execution result.
+     *
+     * @return Job's execution result future.
+     */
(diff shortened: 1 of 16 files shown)


modules/api/src/main/java/org/apache/ignite/compute/JobExecution.java
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.compute;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Represents job's execution.
```
PakhomovAlexander, line 23: It would be awesome if you explain here what is the job execution. Not from the Java code perspective but from the conceptual perspective. What is the job execution? The control object? The snapshot of the job state? This is a part of the public API I would like to see move information and examples here.

modules/compute/src/main/java/org/apache/ignite/internal/compute/executor/QueueExecutionImpl.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Override
+    public JobState state() {
+        return stateMachine.currentState(jobId);
+    }
+
+    @Override
+    public void cancel() {
+        try {
+            executor.remove(queueEntry);
```
PakhomovAlexander, line 81: You remove the job from the executor before transferring its state. I would say it is not safe, because cancellingJob can throw an exception. Shall the `stateMachine.cancelligJob` be the first call?

modules/compute/src/main/java/org/apache/ignite/internal/compute/queue/QueueEntry.java
```diff
@@ -71,6 +96,31 @@ CompletableFuture<R> toFuture() {
         return future;
     }
 
+    /**
+     * Sets interrupt status of the worker thread.
+     */
+    public void interrupt() {
+        // Interrupt under the lock to prevent interrupting thread used by the pool for another task
+        lock.lock();
```
PakhomovAlexander, line 104: I think `AtomicReference` can be a simpler solution for the problem you are trying to solve with the lock.

PakhomovAlexander (approved):

PR #2925, opened 2023-12-06
IGNITE-21030 Dump debug info if the ExecutionService could not be stopped within a timeout

xtern:
https://issues.apache.org/jira/browse/IGNITE-21030

* added timeout for ExecutionService shutdown
* added diagnostic information output when a timeout occurs

Current timeout set as constant 60 seconds (debatable).
Configuration will be added separately.

Diagnostic info output example:

```
Debug info for query: 08034d4d-2586-474e-86dc-bd69a3a52bfd (canceled=false, stopped=false)
  Coordinator node: node_1 (current node)
  Root node state: opened

  Fragments awaiting init completion:
    id=0, node=node_1
    id=1, node=node_1
    id=1, node=node_2
    id=1, node=node_3

  Local fragments:
    id=0, state=opened, canceled=false, class=Inbox  (root)
    id=1, state=opened, canceled=false, class=Outbox
```

### DIFF
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
@@ -148,6 +148,8 @@ public class SqlQueryProcessor implements QueryProcessor {
 
     private static final CacheFactory CACHE_FACTORY = CaffeineCacheFactory.INSTANCE;
 
+    private static final long EXECUTION_SERVICE_SHUTDOWN_TIMEOUT = 60_000;
+
     private final ParserService parserService = new ParserServiceImpl(
             PARSED_RESULT_CACHE_SIZE, CACHE_FACTORY
     );
@@ -322,7 +324,8 @@ public class SqlQueryProcessor implements QueryProcessor {
                 mailboxRegistry,
                 exchangeService,
                 mappingService,
-                dependencyResolver
+                dependencyResolver,
+                EXECUTION_SERVICE_SHUTDOWN_TIMEOUT
         ));
 
         clusterSrvc.topologyService().addEventHandler(executionSrvc);
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
@@ -423,7 +423,7 @@ public class ExecutionServiceImpl<RowT> implements ExecutionService, TopologyEve
         );
 
         try {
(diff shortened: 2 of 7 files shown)


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -148,6 +148,8 @@ public class SqlQueryProcessor implements QueryProcessor {
 
     private static final CacheFactory CACHE_FACTORY = CaffeineCacheFactory.INSTANCE;
 
+    private static final long EXECUTION_SERVICE_SHUTDOWN_TIMEOUT = 60_000;
```
ygerzhedovich, line 151: just a question of separate activity - should we move it to configuration or we consider it only as internal constant?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -454,6 +475,120 @@ private DistributedQueryManager getOrCreateQueryManager(String coordinatorNodeNa
...
+    String dumpDebugInfo() {
+        IgniteStringBuilder buf = new IgniteStringBuilder();
+
+        for (Map.Entry<UUID, DistributedQueryManager> entry : queryManagerMap.entrySet()) {
+            UUID queryId = entry.getKey();
+            DistributedQueryManager mgr = entry.getValue();
+
+            Long rootFragmentId = mgr.rootFragmentId;
+
+            if (rootFragmentId == null) {
```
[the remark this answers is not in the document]
xtern: Fixed, thanks.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImplTest.java
```diff
@@ -699,6 +702,88 @@ public void testErrorIsPropagatedToPrefetchCallback() {
...
+
+        await(cursor.closeAsync());
+
+        assertTrue(waitForCondition(
+                () -> executionServices.stream().map(es -> es.localFragments(ctx.queryId()).size())
+                        .mapToInt(i -> i).sum() == 0, TIMEOUT_IN_MS));
+
+        String nl = System.lineSeparator();
+
+        String expectedOnCoordinator = format(nl
```
zstan, line 759: i think that we need additionally dump all threads stacks, wdyt?

xtern: Added threads dumping.

zstan (approved):

PR #2928, opened 2023-12-07
IGNITE-20909 Use server-to-client notifications to signal Compute job completion

ptupitsyn:
Compute jobs are long-running and may outlive client connections. New Compute API is coming that will return some "execution" object immediately, which can be used to monitor or cancel the job. Therefore, job startup and completion should be separated:
1. Client sends a request to start the job execution and gets a response immediately
2. When the job completes, server sends a notification back to the client

**Protocol Changes**

* New flags: NOTIFICATION, ERROR
* Remove messageType byte, remove null error byte (replaced by flags)

**Notification Mechanism**

* Existing client-generated `requestId` is used for notifications (unlike Ignite 2.x) - this allows the client to set up the listener before the initial response arrives, making the implementation a lot simpler
* Notification has the same format as a regular response, apart from the NOTIFICATION flag
* Notification can arrive before the response. It is not an issue because of reused `requestId`. We could fix this on the server, but this does not seem to provide any benefits.


isapego:
> @isapego could you please help me fix the C++ client?

Sue, let me trake a look

isapego (approved):

PR #2933, opened 2023-12-08
IGNITE-20879 Additional criterions for queries

nva:

### The Review Checklist

### Notes


modules/api/src/main/java/org/apache/ignite/table/criteria/Criteria.java
ygerzhedovich, line 1: Seems need to add example how user can use the new API. Right now I can't understand this

nva: Added

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
```diff
@@ -0,0 +1,218 @@
...
+
+            return this;
+        }
+
+        /**
+         * Set the given criteria.
+         *
+         * @param where where condition.
+         */
+        public SqlSerializer.Builder where(@Nullable Criteria where) {
```
ygerzhedovich, line 184: What does mean `null` ?

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
```diff
@@ -0,0 +1,218 @@
...
+import org.apache.ignite.table.criteria.CriteriaVisitor;
+import org.apache.ignite.table.criteria.Expression;
+import org.apache.ignite.table.criteria.Operator;
+import org.apache.ignite.table.criteria.Parameter;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Serializes {@link Criteria} into into SQL.
+ */
+public class SqlSerializer implements CriteriaVisitor<Void> {
```
ygerzhedovich, line 41: what does mean Void in the context? Why it so?

modules/table/src/test/java/org/apache/ignite/internal/table/criteria/SqlSerializerTest.java
```diff
@@ -0,0 +1,110 @@
...
+        assertEquals("SELECT * FROM test WHERE NOT A IN (?, ?, ?)", ser.toString());
+        assertArrayEquals(new Object[] {"a", "b", "c"}, ser.getArguments());
+    }
+
+    @Test
+    void testAnd() {
+        var ser = new SqlSerializer.Builder()
+                .tableName("test")
+                .columns(Set.of("A", "B"))
+                .where(and(columnValue("a", nullValue()), not(columnValue("b", in("a", "b", "c")))))
```
ygerzhedovich, line 80: I've tried the following condition:
`.where(columnValue("a", equalTo(in(columnValue("a", nullValue()), columnValue("b", nullValue())))))`
Unfortunately no any error and I have result as `a = ?`
It look as incorrect

modules/api/src/main/java/org/apache/ignite/table/criteria/Criteria.java
```diff
@@ -0,0 +1,228 @@
...
+     * <pre>
+     *     columnValue(&quot;category&quot;, equalTo(&quot;toys&quot;))
+     * </pre>
+     *
+     * @param columnName Column name.
+     * @param condition Target condition.
+     * @return The created expression instance.
+     */
+    static Expression columnValue(String columnName, Condition condition) {
+        var oldElements = condition.getElements();
```
ygerzhedovich, line 57: using `var` for such caseses are forbidden

modules/api/src/main/java/org/apache/ignite/table/criteria/Criteria.java
```diff
@@ -0,0 +1,228 @@
...
+ */
+public interface Criteria {
+    /**
+     * Accept the visitor with the given context.
+     *
+     * @param <C> context type.
+     * @param v visitor.
+     * @param context context of visit.
+     */
+    <C> void accept(CriteriaVisitor<C> v, @Nullable C context);
```
ygerzhedovich, line 42: what purpose if the context? I don't see any real usage of it

modules/api/src/main/java/org/apache/ignite/table/criteria/Criteria.java
```diff
@@ -0,0 +1,228 @@
...
+ * </code></pre>
+ *
+ * @see CriteriaQuerySource
+ */
+public interface Criteria {
+    /**
+     * Accept the visitor with the given context.
+     *
+     * @param <C> context type.
+     * @param v visitor.
```
ygerzhedovich, line 39:
```suggestion
     * @param v Visitor.
```
Please check all javadocs

modules/api/src/main/java/org/apache/ignite/table/criteria/Column.java
ygerzhedovich, line 1: What reason have the class in public API?  Is it require for end user? The question also relate for part other classes in the package

modules/api/src/main/java/org/apache/ignite/table/criteria/Criteria.java
nva: Fixed

modules/api/src/main/java/org/apache/ignite/table/criteria/Criteria.java
nva: Fixed. Checked all other places

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
nva: null means return all entries in record view. Fixed javadoc.

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
nva: This means that this visitor doesn't require any context

modules/table/src/test/java/org/apache/ignite/internal/table/criteria/SqlSerializerTest.java
nva: This problem occurred due to https://github.com/apache/ignite-3/pull/2810#discussion_r1386303866.
Replaced `Object` with `Comparable` in argument and append another `equalsTo` method with byte[] as argument.

modules/api/src/main/java/org/apache/ignite/table/criteria/Column.java
nva: We can't remove as`Criteria#columnValue` use Column. 
It's like `Tuple` and `TupleImpl`. User can't create an instance of `TupleImpl` directly, but we store it in public api module.
So if we want to use `Criteria` as an entry point for building criteria, we need to put all these classes in a public api module. All these classes have a package visible constructor, so it shouldn't not confuse the user

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
ygerzhedovich: can we add explanation to javadoc?

nva: Improve javadoc

modules/table/src/test/java/org/apache/ignite/internal/table/criteria/SqlSerializerTest.java
ygerzhedovich: also need to have for all supported types

modules/api/src/main/java/org/apache/ignite/table/criteria/Criteria.java
```diff
@@ -17,9 +17,226 @@
...
+     *
+     * <p>For example:
+     * <pre>
+     *     columnValue("age", lessThanOrEqualTo(35))
+     * </pre>
+     *
+     * @param <T> Value type.
+     * @param value Target value.
+     */
+    static <T> Condition lessThanOrEqualTo(Comparable<T> value) {
```
ygerzhedovich, line 196: we should have tests for all methods for all types

modules/client/src/main/java/org/apache/ignite/internal/client/table/AbstractClientView.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Construct SQL query and arguments for prepare statement.
+     *
+     * @param tableName Table name.
+     * @param columns Table columns.
+     * @param criteria The predicate to filter entries or {@code null} to return all entries from the underlying table.
+     * @return SQL query and it's arguments.
+     */
+    static SqlSerializer createSqlSerializer(String tableName, ClientColumn[] columns, @Nullable Criteria criteria) {
```
ygerzhedovich, line 60: I can't find any tests related to the class

modules/api/src/main/java/org/apache/ignite/table/criteria/Criteria.java
```diff
@@ -17,9 +17,226 @@
...
+     * Creates a condition that test the examined object is equal to the specified {@code value}.
+     *
+     * <p>For example:
+     * <pre>
+     *     columnValue(&quot;category&quot;, equalTo(&quot;toys&quot;))
+     * </pre>
+     *
+     * @param value Target value.
+     */
+    static Condition equalTo(byte[] value) {
```
ygerzhedovich, line 136: should we have similar overloaded methods for another operations?

modules/api/src/main/java/org/apache/ignite/table/criteria/Column.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package org.apache.ignite.table.criteria;
+
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Represents a column reference for criteria query.
+ */
+public class Column implements Criteria {
```
ygerzhedovich, line 25: seems all of this class should be final

modules/api/src/main/java/org/apache/ignite/table/criteria/Criteria.java
```diff
@@ -17,9 +17,226 @@
...
+    static <T> Condition equalTo(Comparable<T> value) {
+        return new Condition(Operator.EQ, new Parameter<>(value));
+    }
+
+    /**
+     * Creates a condition that test the examined object is equal to the specified {@code value}.
+     *
+     * <p>For example:
+     * <pre>
+     *     columnValue(&quot;category&quot;, equalTo(&quot;toys&quot;))
```
ygerzhedovich, line 131: example is inccorect.

modules/table/src/test/java/org/apache/ignite/internal/table/criteria/SqlSerializerTest.java
```diff
@@ -0,0 +1,110 @@
...
+                .build();
+
+        assertEquals("SELECT * FROM test WHERE (A IS NULL) AND (NOT B IN (?, ?, ?))", ser.toString());
+        assertArrayEquals(new Object[] {"a", "b", "c"}, ser.getArguments());
+    }
+
+    @Test
+    void testSqlInjection() {
```
ygerzhedovich, line 88: How it related to Sql injection?

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
```diff
@@ -0,0 +1,221 @@
...
+
+    /** {@inheritDoc} */
+    @Override
+    public <T> void visit(Expression expression, @Nullable Void context) {
+        Operator operator = expression.getOperator();
+        Criteria[] elements = expression.getElements();
+
+        String template;
+
+        if (operator == Operator.AND || operator == Operator.OR) {
```
ygerzhedovich, line 92: we shoud have more tests for the code. As I see not covered cases with one parameter, for three parameters, no test for OR, and so on....

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
```diff
@@ -0,0 +1,221 @@
...
+
+            template = IntStream.range(1, elements.length)
+                    .mapToObj(i -> String.format("{%d}", i))
+                    .collect(Collectors.joining(", ", prefix, ")"));
+        } else {
+            template = ELEMENT_TEMPLATES.get(operator);
+        }
+
+        int end = 0;
+        Matcher matcher = Pattern.compile("\\{(\\d+)\\}").matcher(template);
```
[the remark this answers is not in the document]
ygerzhedovich: or better yet, insert parameters immediately when generating strings

modules/api/src/main/java/org/apache/ignite/table/criteria/Column.java
ygerzhedovich: seems you are right here

modules/api/src/main/java/org/apache/ignite/table/criteria/Criteria.java
nva: I can't imagine any example for such a case.

modules/client/src/main/java/org/apache/ignite/internal/client/table/AbstractClientView.java
nva: That tests do you expect for an abstract class? This is the same as the `AbstractTableView` and there are no tests in it. If you speak about package-private visibility, I replaced it with protected.

modules/api/src/main/java/org/apache/ignite/table/criteria/Column.java
nva: Done

modules/api/src/main/java/org/apache/ignite/table/criteria/Criteria.java
nva: Fixed

modules/table/src/test/java/org/apache/ignite/internal/table/criteria/SqlSerializerTest.java
nva: Column name can be used for sql inject if came from rest api, for example. Replaced with column name validation.

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
nva: Extended

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
```diff
@@ -0,0 +1,221 @@
...
+        Operator operator = expression.getOperator();
+        Criteria[] elements = expression.getElements();
+
+        String template;
+
+        if (operator == Operator.AND || operator == Operator.OR) {
+            String delimiter = operator == Operator.AND ? ") AND (" : ") OR (";
+
+            template = IntStream.range(0, elements.length)
+                    .mapToObj(i -> String.format("{%d}", i))
```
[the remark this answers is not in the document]
nva: Removed

modules/api/src/main/java/org/apache/ignite/table/criteria/Criteria.java
nva: Added

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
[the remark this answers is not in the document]
nva: Move to the field

modules/api/src/main/java/org/apache/ignite/table/criteria/Criteria.java
ygerzhedovich: For example I would select rows for particular values of bynary columns, Why I can not do it?
like SELECT * FROM T WHERE bin_col in ('0xAA','0xBB')
the same for NOT IN

nva: Added for in, not in

modules/api/src/main/java/org/apache/ignite/table/criteria/Criteria.java
```diff
@@ -290,6 +290,32 @@ static <T> Condition in(Comparable<T>... values) {
...
+     *
+     * <p>For example:
+     * <pre>
+     *     columnValue(&quot;password&quot;, in(&quot;MyPassword&quot;.getBytes(), &quot;MyOtherPassword&quot;.getBytes()))
+     * </pre>
+     *
+     * @param values The collection in which matching items must be found.
+     */
+    static Condition in(byte[]... values) {
+        if (values == null || values.length == 0) {
```
ygerzhedovich, line 304: Why it can't be `null` or empty?

PR #2937, opened 2023-12-08
IGNITE-20568 .NET: Add support for IgniteCheckedException in ExceptionsGenerator

ptupitsyn:
Currently there are no public exceptions derived from `IgniteCheckedException`, but we don't want to miss them in .NET when they are added.


isapego (approved):
Looks good to me

PR #2948, opened 2023-12-12
IGNITE-20652 .NET: Add SQL script execution API

ptupitsyn:
* Change protocol: `SQL_EXEC` and `SQL_EXEC_SCRIPT` now have the same payload format - we send all session properties. This is simpler and future-proof. We don't know yet which properties will be usable in scripts - quite possibly all of them.
* Implement script API in .NET

Extras:
* Remove outdated `TODO IGNITE-20568: Cast to AuthenticationException`
* Fix `SslTests` flakiness


modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/sql/ClientSqlExecuteScriptRequest.java
```diff
@@ -17,17 +17,15 @@
 
 package org.apache.ignite.client.handler.requests.sql;
 
-import static org.apache.ignite.client.handler.requests.sql.ClientSqlExecuteRequest.readSessionProperties;
+import static org.apache.ignite.client.handler.requests.sql.ClientSqlExecuteRequest.readSession;
```
isapego, line 20: Looks a bit weird to me. Let's move it to `ClientUtils`, or something.

isapego (approved):

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/sql/ClientSqlExecuteScriptRequest.java
ptupitsyn: Moved to `ClientSqlCommon`.

PR #2958, opened 2023-12-14
IGNITE-21062 Fix safe time reordering in partitions.

sanpwc:
https://issues.apache.org/jira/browse/IGNITE-21062


modules/replicator/src/main/java/org/apache/ignite/internal/replicator/command/SafeTimePropagatingCommand.java
```diff
@@ -46,4 +46,11 @@ default void safeTimeLong(long safeTime) {
     default HybridTimestamp safeTime() {
         return hybridTimestamp(safeTimeLong());
     }
+
+    /**
+     * Public clone version that is implemented in generated *Impl class.
+     */
+    default Object clone() throws CloneNotSupportedException {
```
ibessonov, line 53: I think we could add "clone" to the "NetworkMessage" instead, avoiding handling the impossible exception as a result

ibessonov (commented):
Looks good to me!

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/command/SafeTimePropagatingCommand.java
sanpwc: Do you assume following (or similar) implementation in MessageImpl classes?
```
    @Override
    public Object clone() {
        try {
            return super.clone();
        }
        catch (CloneNotSupportedException e) {
            // Never expected to be thrown because whole message class hierarchy implements clone();
            throw new AssertionError(e);
        }
    }
```

ibessonov: Almost. I thought its enough to declare `Object clone();` in `NetworkMessage`, without exception. MEssage generator always creates implementation, that's fine. I'm only worried about the exception in the signature, that will never happen in reality

sanpwc: If Object clone() in NetworkMessage (or NetworkMessage clone() which currently doesn't matter) doesn't through an exception implementation (*Impl) also should not declare throws CloneNotSupportedException. And because every MessageImple class directly extends Object supler.clone() will through CloneNotSupportedException and thus should be handled. Am I missing something?

ibessonov: Sorry, after re-reading your last message, I say "yes", I assumed the same implementation that you provided

ibessonov (approved):

PR #2960, opened 2023-12-15
IGNITE-21052 Improved measurement of first query planning time

xtern:
https://issues.apache.org/jira/browse/IGNITE-21052

Issue:
 first query planning takes much longer, mainly due to class loading.

We found that the most time was spent scanning the class path during initialization of `RelMetadataQueryEx`.
And this is redundant now, since from Calcite 1.28.0 there is no need to register all known RelNode subtypes (see [CALCITE-4546](https://issues.apache.org/jira/browse/CALCITE-4546) for details).

Also it was found that initialization of `PlannerPhase` enum loads 600+ classes, and this loading affects planning time.

So I decided:
1. Remove redundant `RelNode` subtypes registration during initialization of `RelMetadataQueryEx` (on TeamCity it took up to 5 seconds)
2. Load `PlannerPhase` related classes during static initialization of planner (on TeamCity it took up to 2 seconds).

After these changes I run unit tests on TC 30+ times without timeout: https://ci.ignite.apache.org/viewType.html?buildTypeId=ApacheIgnite3xGradle_Test_RunUnitTests&branch_ApacheIgnite3xGradle_Test=pull%2F2960&tab=buildTypeStatusDiv

I also analyzed over 10 runs and the (first) planning time now looks more stable, between 1800-2100ms.

### DIFF
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/RelMetadataQueryEx.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/RelMetadataQueryEx.java
@@ -33,21 +33,6 @@ import org.apache.ignite.internal.sql.engine.rel.IgniteRel;
  * See {@link RelMetadataQuery}.
  */
 public class RelMetadataQueryEx extends RelMetadataQuery {
-    static {
-        try (ScanResult scanResult = new ClassGraph().acceptPackages("org.apache.ignite.internal.sql.engine.rel")
-                .addClassLoader(igniteClassLoader())
-                .enableClassInfo().scan()
-        ) {
-            //noinspection unchecked
-            List<Class<? extends RelNode>> types = scanResult.getClassesImplementing(IgniteRel.class.getName())
-                    .filter(classInfo -> !classInfo.isInterface())
-                    .filter(classInfo -> !classInfo.isAbstract())
-                    .stream().map(classInfo -> (Class<? extends RelNode>) classInfo.loadClass()).collect(toList());
-
-            JaninoRelMetadataProvider.DEFAULT.register(types);
-        }
-    }
-
     /**
      * Factory method.
      *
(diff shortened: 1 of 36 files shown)


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/RelMetadataQueryEx.java
```diff
@@ -17,37 +17,14 @@
 
 package org.apache.ignite.internal.sql.engine.metadata;
 
-import static java.util.stream.Collectors.toList;
-import static org.apache.ignite.internal.util.IgniteUtils.igniteClassLoader;
-
-import io.github.classgraph.ClassGraph;
```
ygerzhedovich, line 23: can we remove the dependency at all?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgnitePlanner.java
```diff
@@ -133,6 +133,12 @@ public class IgnitePlanner implements Planner, RelOptTable.ViewExpander {
 
     private RelOptCluster cluster;
 
+    static {
+        // Preload some classes to reduce first planning time.
```
ygerzhedovich: In the fact visible time of first planning will be the same. 
I see tow options here:
1) Move loading and initialization classes on earlier stage, during node startup. Due to java could do classpoading in parallel it shouldn't significant increase time of node start, but our very first parsing will be visible faster. For tests which could instantiate some of services himself we could keep initialization in two places.
2) Just fix a coomet to show real life )

xtern: Thanks, I expanded the comment because we have tests that use the planner directly, without starting nodes, services, etc.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/metadata/RelMetadataQueryEx.java
xtern: Thank you. It seems so. The dependency has been removed.

lowka (approved):

lowka:
I have the same concerns as Yuri (point 1 in his comment). But I think it is out of scope of this PR.

AMashenkov (approved):

ygerzhedovich (approved):

PR #2961, opened 2023-12-15
IGNITE-21083 Fix IllegalReferenceCountException in JDBC and Java client

ptupitsyn:
**Problem:** `payloadReader` can escape `asyncContinuationExecutor` thread in some cases due to `thenApply` call in `TcpClientChannel.receiveAsync`. In particular, this happened due to blocking future wait in `JdbcResultSet.close0`. In this case, `TcpClientChannel.onMessage` closes the buffer before we read from it, causing `IllegalReferenceCountException`.

**Solution:** ensure that `payloadReader` is called synchronously before we close the buffer. Store it separately and invoke directly without relying on future continuations.

Note: `supportedVers` and related handshake retry removed for simplicity - we only have one supported protocol version currently.


lowka (approved):
Looks good to me.

PR #2965, opened 2023-12-18
IGNITE-20673 Sql. Independent multi statement queries need to be executed in parallel

korlov42:
This patch introduces parallelisation for sequence of SELECT statements.

===============

### The Review Checklist

### Notes

### DIFF
--- a/modules/sql-engine/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
+++ b/modules/sql-engine/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
@@ -24,6 +24,8 @@ import static org.apache.ignite.internal.testframework.IgniteTestUtils.waitForCo
 import static org.apache.ignite.lang.ErrorGroups.Sql.RUNTIME_ERR;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
@@ -304,6 +306,46 @@ public class ItSqlMultiStatementTxTest extends BaseSqlMultiStatementTest {
                 () -> await(insCur.nextResult()));
 
         verifyFinishedTxCount(1);
+    }
+
+    @Test
+    void concurrentExecutionDoesntAffectSelectWithExplicitTx() {
+        long tableSize = BIG_TABLE_ROWS_COUNT;
+
+        @SuppressWarnings("ConcatenationWithEmptyString")
+        String script = ""
+                + "CREATE TABLE integers (i INT PRIMARY KEY);"
+                + "START TRANSACTION;"
+                + "INSERT INTO integers SELECT * FROM TABLE(system_range(1, " + tableSize + "));"
+                + "SELECT count(*) FROM integers;"
+                + "DELETE FROM integers;"
+                + "COMMIT;"
+                + "DROP TABLE integers;";
+
+        AsyncSqlCursor<InternalSqlRow> cursor = runScript(script); // CREATE TABLE...
+        cursor.closeAsync();
+
+        cursor = await(cursor.nextResult()); // BEGIN...
+
+        assertThat(cursor, notNullValue());
+        cursor.closeAsync();
+
+        cursor = await(cursor.nextResult()); // INSERT...
+
+        assertThat(cursor, notNullValue());
+        cursor.closeAsync();
+
+        cursor = await(cursor.nextResult()); // SELECT...
+
+        assertThat(cursor, notNullValue());
+
+        BatchedResult<InternalSqlRow> batch = await(cursor.requestNextAsync(1));
+
+        assertThat(batch, notNullValue());
+
+        assertThat(batch.items().get(0).get(0), is(tableSize));
+
+        iterateThroughResultsAndCloseThem(cursor);
     }
 
     @Test
--- a/modules/sql-engine/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
+++ b/modules/sql-engine/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
@@ -230,22 +230,18 @@ public class ItSqlMultiStatementTxTest extends BaseSqlMultiStatementTest {
 
     @Test
     void openedScriptTransactionRollsBackOnError() {
-        {
-            AsyncSqlCursor<InternalSqlRow> cursor = runScript(
-                    "START TRANSACTION READ WRITE;"
-                    + "INSERT INTO test VALUES(2);"
-                    + "INSERT INTO test VALUES(2/0);"
-                    + "SELECT 1;"
-                    + "COMMIT;"
-            );
+        String script = "START TRANSACTION READ WRITE;"
+                + "INSERT INTO test VALUES(2);"
+                + "INSERT INTO test VALUES(2/0);"
+                + "SELECT 1;"
+                + "COMMIT;";
 
-            assertThrowsSqlException(RUNTIME_ERR, "Division by zero", () -> fetchAllCursors(cursor));
+        assertThrowsSqlException(RUNTIME_ERR, "Division by zero", () -> executeScript(script));
 
-            verifyFinishedTxCount(1);
+        verifyFinishedTxCount(1);
 
-            assertQuery("select count(id) from test")
-                    .returns(0L).check();
-        }
+        assertQuery("select count(id) from test")
+                .returns(0L).check();
     }
 
     @Test
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/AsyncSqlCursorImpl.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/AsyncSqlCursorImpl.java
@@ -44,8 +44,8 @@ public class AsyncSqlCursorImpl<T> implements AsyncSqlCursor<T> {
     private final ResultSetMetadata meta;
     private final QueryTransactionWrapper txWrapper;
     private final AsyncCursor<T> dataCursor;
-    private final Runnable onClose;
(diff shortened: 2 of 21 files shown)


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -835,7 +851,42 @@ CompletableFuture<AsyncSqlCursor<InternalSqlRow>> processNext() {
...
+                            // tx is started already, no need to wait
+                            triggerFuture = nullCompletedFuture();
+                        } else if (txWrapper.implicit()) {
+                            if (cursor.queryType() != SqlQueryType.QUERY) {
+                                // any query apart from type of QUERY returns at most a single row, so
+                                // it should be safe to commit transaction prematurely after receiving
+                                // `firstPageReady` signal, since all sources have been processed
+                                triggerFuture = cursor.onFirstPageReady().thenCompose(none -> txWrapper.commitImplicit());
+                            } else if (nextStatement != null
+                                    && readOnlyQuery(nextStatement.parsedResult.queryType())) {
```
xtern, line 867: why do we need this condition (why do we need to wait for prefetch of SELECT if the next statement is read-write)?
Do we have a test that shows that it is necessary?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/AsyncSqlCursorImpl.java
```diff
@@ -158,17 +131,6 @@ public CompletableFuture<Void> closeAsync() {
         dataCursor.closeAsync()
                 .thenCompose(ignored -> txWrapper.commitImplicit())
                 .whenComplete((r, e) -> {
```
AMashenkov, line 133: `closeResult` future goes outside.
Should we complete it in a separate thread? I mean the case, when a continuation with a user code may be executed synchronously in SQL thread pool.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -835,7 +851,42 @@ CompletableFuture<AsyncSqlCursor<InternalSqlRow>> processNext() {
...
+                                // if next statement only reads data, no need to wait
+                                triggerFuture = nullCompletedFuture();
+                            } else {
+                                triggerFuture = cursor.onFirstPageReady();
+                            }
+                        } else {
+                            if (readOnlyQuery(cursor.queryType())
+                                    && nextStatement != null
+                                    && readOnlyQuery(nextStatement.parsedResult.queryType())) {
```
xtern, line 876: The same question - why we need to wait for the prefetch of the last SELECT?
Even if this is necessary - we can't be sure that the prefetch for the previous SELECT has already been done, so it looks strange.

for example 
```
SELECT 1
SELECT 2
INSERT ...
```

as I understand we are waiting here for prefetch of SELECT 2, but not for SELECT 1, in theory prefetch for SELECT 2 can finishes before prefetch for SELECT 1

modules/sql-engine/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
```diff
@@ -230,22 +230,18 @@ void dmlScriptRollsBackImplicitly() throws InterruptedException {
...
-            AsyncSqlCursor<InternalSqlRow> cursor = runScript(
-                    "START TRANSACTION READ WRITE;"
-                    + "INSERT INTO test VALUES(2);"
-                    + "INSERT INTO test VALUES(2/0);"
-                    + "SELECT 1;"
-                    + "COMMIT;"
-            );
+        String script = "START TRANSACTION READ WRITE;"
+                + "INSERT INTO test VALUES(2);"
+                + "INSERT INTO test VALUES(2/0);"
```
xtern, line 235: what is the expected behavior if SELECT will raise an error?
Should script be executed completely?

From my experiments (may be wrong) I see that sometime the script completes with error, but sometimes without, my be we need to not interrupt script execution in case of errors from  SELECT statements at all.

AMashenkov (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/AsyncSqlCursorImpl.java
korlov42: > Should we complete it in a separate thread?

I would say "no". Query processor is an internal API and doesn't expose any results directly to the user. Depending on a particular public API implementation, an internal results follow by different paths:

- in case of thin clients, the results are serialised and send over the network. To me, it's perfectly fine to use SQL thread pool for that
- in case of embedded sync API, user will wait till the future completion, and then will proceed on its own thread
- in case of embedded async API there is a chance that user's workload will be done on SQL thread pool, but this is a problem of a particular API and should be addressed accordingly (by introducing continuation executor, for instance, as it done in thin clients; BTW, not only sql is affected, by KV/Record views as well)

in other word, the problem exists, but it's subject to separate JIRA

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -656,9 +649,10 @@ private CompletableFuture<SchemaPlus> waitForActualSchema(String schemaName, Hyb
         }
     }
 
-    private AsyncSqlCursor<InternalSqlRow> executePlan(
+    private CompletableFuture<AsyncSqlCursor<InternalSqlRow>> executePlan(
             QueryTransactionWrapper txWrapper,
             BaseQueryContext ctx,
+            PrefetchCallback callback,
```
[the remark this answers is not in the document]
korlov42: actually, we can't. I passed it explicitly in order to preserve type `PrefetchCallback` since I need a future, not the callback to notify. Another options are to do a cast from `QueryPrefetchCallback` to `PrefetchCallback`, and to extend QueryPrefetchCallback exposing a future.

I prefer to pass callback explicitly, this approach seams to be lesser evil

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -603,12 +607,11 @@ private CompletableFuture<AsyncSqlCursor<InternalSqlRow>> executeParsedStatement
             QueryTransactionWrapper txWrapper,
             QueryCancel queryCancel,
             Object[] params,
-            boolean waitForPrefetch,
             @Nullable CompletableFuture<AsyncSqlCursor<InternalSqlRow>> nextStatement
     ) {
         return waitForActualSchema(schemaName, txWrapper.unwrap().startTimestamp())
                 .thenCompose(schema -> {
-                    PrefetchCallback callback = waitForPrefetch ? new PrefetchCallback() : null;
+                    PrefetchCallback callback = new PrefetchCallback();
```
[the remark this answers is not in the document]
korlov42: to be honest, I don't like this callback either. But at the moment I have no idea how to do it better

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -619,17 +622,7 @@ private CompletableFuture<AsyncSqlCursor<InternalSqlRow>> executeParsedStatement
                             .build();
 
                     CompletableFuture<AsyncSqlCursor<InternalSqlRow>> fut = prepareSvc.prepareAsync(parsedResult, ctx)
```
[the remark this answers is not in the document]
korlov42: fixed

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
korlov42: I gave it a second thought, and indeed looks like we don't need this check. For implicit transaction we may wait nothing, because with implicit tx we always read a consistent snapshot

thanks for noticing! fixed

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -835,7 +851,42 @@ CompletableFuture<AsyncSqlCursor<InternalSqlRow>> processNext() {
...
+
+                        if (txWrapper == null) {
+                            // tx is started already, no need to wait
+                            triggerFuture = nullCompletedFuture();
+                        } else if (txWrapper.implicit()) {
+                            if (cursor.queryType() != SqlQueryType.QUERY) {
+                                // any query apart from type of QUERY returns at most a single row, so
+                                // it should be safe to commit transaction prematurely after receiving
+                                // `firstPageReady` signal, since all sources have been processed
+                                triggerFuture = cursor.onFirstPageReady().thenCompose(none -> txWrapper.commitImplicit());
```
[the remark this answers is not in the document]
korlov42: make sense, thanks! I've also added assertion to be 100% sure a first page is ready

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
korlov42: > why we need to wait for the prefetch of the last SELECT?

with explicit tx things a bit trickier here: we have no isolation within the same transaction, meaning you can even see results from operation itself (for instance, a query `INSERT INTO t SELECT * FROM t may insert more rows then it was prior to statement execution).With that said, we have to make sure that DML is started not earlier than first pages of SELECT statements prior to that DML become ready.

I've added tracking of such in-flight selects

modules/sql-engine/src/integrationTest/java/org/apache/ignite/internal/sql/engine/ItSqlMultiStatementTxTest.java
korlov42: > what is the expected behavior if SELECT will raise an error?

as far as I see, it depends... If exception is raised during cursor preparation, then an entire script will be interrupted. But now, data fetch is not part of the cursor preparation for SELECT statement, thus, script won't be interrupted, if it works in autoCommit=true mode

xtern (approved):

zstan (approved):

PR #2972, opened 2023-12-18
IGNITE-20598 Java client: Fix incorrect error for query on closed transaction

ptupitsyn:
* Add local tx state check - disallow using finished tx
* Wrap all `ClientSession` exceptions in `SqlException` (match embedded API behavior)

Also fixes [IGNITE-20742 Thin client cannot find the resource when it is trying to execute an operation on the finalized transaction](https://issues.apache.org/jira/browse/IGNITE-20742)


modules/client/src/main/java/org/apache/ignite/internal/client/tx/ClientTransaction.java
```diff
@@ -134,6 +153,23 @@ public static ClientTransaction get(Transaction tx) {
                     + "'. Use IgniteClient.transactions() to start transactions.");
         }
 
-        return (ClientTransaction) tx;
+        ClientTransaction clientTx = (ClientTransaction) tx;
+
+        var state = clientTx.state.get();
```
ygerzhedovich, line 158: using `var` in the context is forbidden by our codestyle guid

ygerzhedovich (approved):

modules/client/src/main/java/org/apache/ignite/internal/client/tx/ClientTransaction.java
ptupitsyn: Fixed.

modules/client/src/main/java/org/apache/ignite/internal/client/tx/ClientTransaction.java
```diff
@@ -134,6 +153,23 @@ public static ClientTransaction get(Transaction tx) {
...
+
+        if (state == STATE_OPEN) {
+            return clientTx;
+        }
+
+        // Match org.apache.ignite.internal.tx.TxState enum:
+        String stateStr = state == STATE_COMMITTED ? "COMMITTED" : "ABORTED";
+
+        throw new TransactionException(
```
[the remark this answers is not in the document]
ptupitsyn: We have tests that run against client and embedded APIs and ensure consistent behavior (ItSqlSynchronousApiTest & ItSqlClientSynchronousApiTest, etc).

It would be great to rely on the server only for this, but when the tx is finished, the server removes it from `ClientResourceRegistry` and we can't use it anymore from the client.

PR #2979, opened 2023-12-19
IGNITE-21124 Add ability to log long message processing

sanpwc:
https://issues.apache.org/jira/browse/IGNITE-21124

### DIFF
--- a/modules/network-api/src/main/java/org/apache/ignite/network/TrackableNetworkMessageHandler.java
+++ b/modules/network-api/src/main/java/org/apache/ignite/network/TrackableNetworkMessageHandler.java
@@ -24,7 +24,7 @@ import org.jetbrains.annotations.Nullable;
 /**
  * Trackable message handler that will log long-running messages.
  */
-public class TrackableNetworkMessageHandler implements NetworkMessageHandler{
+public class TrackableNetworkMessageHandler implements NetworkMessageHandler {
     private static final IgniteLogger LOG = Loggers.forClass(TrackableNetworkMessageHandler.class);
 
     /**
--- a/modules/network-api/src/main/java/org/apache/ignite/network/TrackableNetworkMessageHandler.java
+++ b/modules/network-api/src/main/java/org/apache/ignite/network/TrackableNetworkMessageHandler.java
@@ -17,6 +17,8 @@
 
 package org.apache.ignite.network;
 
+import static org.apache.ignite.internal.tostring.IgniteToStringBuilder.includeSensitive;
+
 import org.apache.ignite.internal.logger.IgniteLogger;
 import org.apache.ignite.internal.logger.Loggers;
 import org.jetbrains.annotations.Nullable;
@@ -57,7 +59,8 @@ public class TrackableNetworkMessageHandler implements NetworkMessageHandler {
             LOG.warn(
                     "Message handling has been too long [duration={}ns, message=[{}]]",
                     durationNanos,
-                    message
+                    // Message may include sensitive data, however it seems useful to print full message content while testing.
+                    includeSensitive() ? message : message.getClass()
             );
         }
     }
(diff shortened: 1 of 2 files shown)


modules/network-api/src/main/java/org/apache/ignite/network/TrackableNetworkMessageHandler.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private final NetworkMessageHandler targetHandler;
+
+    TrackableNetworkMessageHandler(NetworkMessageHandler targetHandler) {
+        this.targetHandler = targetHandler;
+    }
+
+    @Override
+    public void onReceived(NetworkMessage message, String senderConsistentId, @Nullable Long correlationId) {
+        long startTimeNanos = System.nanoTime();
```
vldpyatkov, line 48: I am sure you need to use `FastTimestamps#startUpdater` here because the accuracy of 10 ms is enough to estimate the latency.

modules/network-api/src/main/java/org/apache/ignite/network/TrackableNetworkMessageHandler.java
```diff
@@ -0,0 +1,67 @@
...
+public class TrackableNetworkMessageHandler implements NetworkMessageHandler {
+    private static final IgniteLogger LOG = Loggers.forClass(TrackableNetworkMessageHandler.class);
+
+    /**
+     * If message handling takes more time, than this constant, we will log warning message with some information.
+     */
+    // TODO: I don't know which value to use here, let's discuss in the PR. On the one hand every IO should be highlighted and thus
+    // TODO: the value should be rather small, on the other hand, currently we have lots of slow messages and we will just flood the log
+    // TODO: will log running ones. So for now, I believe we should set rather big value here in order to ease test failure analysis.
+    private static final int WATCH_EVENT_PROCESSING_LOG_THRESHOLD_NANOS = 1_000_000;
```
ibessonov, line 38: Please consider renaming the constant. It has nothing to do with watch processing

ibessonov: 1 million nanos is a millisecond. It would have been nice to process everything within such time, but I suggest being more pessimistic just in case

modules/network-api/src/main/java/org/apache/ignite/network/TrackableNetworkMessageHandler.java
ibessonov: What are the advantages? Measuring time with `nanoTime()` is a standard approach after all

sanpwc: What do you mean? I won't to catch messages that took 1 ms. Thus 10 ms frequency won't be enough.

modules/network-api/src/main/java/org/apache/ignite/network/TrackableNetworkMessageHandler.java
sanpwc: > but I suggest being more pessimistic just in case

Any suggestions?

ibessonov: We could experiment with it. The goal is to see problems, but not to pollute the log. I don't have specific numbers, unfortunately. We would need to run some suites with different values (ideally) and see the real output

sanpwc: Adjusted to 5 ms for know. Suggestions still welcome.

modules/network-api/src/main/java/org/apache/ignite/network/TrackableNetworkMessageHandler.java
sanpwc: Discussed in person. `FastTimestamps#startUpdater` is not applicable here.

ibessonov (approved):

vldpyatkov (approved):

PR #2982, opened 2023-12-21
IGNITE-21018 Sql. Introduce cache for plan mapping

xtern:
https://issues.apache.org/jira/browse/IGNITE-21018

Each cache entry includes the following:
1. participating table identifiers
2. topology version
3. future that will return plan mapping.

Cache entry may be invalidated in the following cases:
1. The primary replica has expired - we invalidate the records that contain the identity table for which the event was published (current implementation scans whole cache entries to do this).
* If cached mapping include system view and topology version changes, the next access to this cache entry will cause the plan fragments to be remapped and the cache entry to be replaced.

NOTE: original idea to cache minimum lease timestamp and invalidate the cache entry after the lease expired was rejected because the lease time is just around 5 seconds and we have no information about the prolongation status.

NOTE 2: Invalidation of cache entries when the primary replica expires can be optimized by using an additional CHM-like cache containing all cached table identifiers. Implementation of such cache strategy is attached to the [ticket](https://issues.apache.org/jira/browse/IGNITE-21018) (see [ReferencesCache.java](https://issues.apache.org/jira/secure/attachment/13065664/ReferencesCache.java)). During the discussion, it was decided that for now we can leave the trivial implementation with a whole cache scan, since the proposed algorithm needs to be more carefully analyzed and benchmarked.

Benchmark results.

`SelectBenchmark.sqlGet` (nodes=1, partitions=1, tableSize=1000, fsync=false) shows very little performance improvement.

Benchmark result without patch:
```
Benchmark               (clusterSize)  (fsync)  Mode  Cnt    Score   Error  Units
SelectBenchmark.sqlGet              1    false  avgt   20  188.619 ± 6.569  us/op
```

Benchmark result with patch:
```
Benchmark               (clusterSize)  (fsync)  Mode  Cnt    Score   Error  Units
SelectBenchmark.sqlGet              1    false  avgt   20  177.816 ± 7.506  us/op
```

May be something wrong with my hardware (replacing `compute` with `computeIfAbsent` when accessing the cache gives very similar results).


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/mapping/MappingServiceImpl.java
```diff
@@ -96,12 +108,54 @@ public CompletableFuture<List<MappedFragment>> map(MultiStepPlan multiStepPlan)
...
 
+    /** Called when the primary replica has expired. */
+    public CompletableFuture<Boolean> onPrimaryReplicaExpired(PrimaryReplicaEventParameters parameters) {
+        assert parameters != null;
+        assert parameters.groupId() instanceof TablePartitionId;
+
+        int tabId = ((TablePartitionId) parameters.groupId()).tableId();
+
+        return CompletableFuture.supplyAsync(() -> {
+            mappingsCache.values().removeIf(value -> value.tableIds.contains(tabId));
```
xtern, line 119: It might make sense to wrap this in a synchronized block to reduce contention if there are will be a lot of simultaneous events about the same table (for each partition for example).

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/mapping/MappingServiceImpl.java
```diff
@@ -96,12 +108,54 @@ public CompletableFuture<List<MappedFragment>> map(MultiStepPlan multiStepPlan)
...
+
+            return false;
+        }, ForkJoinPool.commonPool());
```
xtern, line 122: It looks like the full scan should be done asynchronously, at least to avoid blocking other listeners. But I'm not sure which pool is best to use for this.

I prefer not to use the query execution pool for this. Because firstly, there may be quite long tasks in the queue, and secondly, we can slow it down with idle cache scanning tasks.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/mapping/MappingServiceImpl.java
```diff
@@ -96,12 +108,54 @@ public CompletableFuture<List<MappedFragment>> map(MultiStepPlan multiStepPlan)
...
+                        tableIds.add(source.id());
+                    }
+                }
+
+                long topVer = topAware ? topologyHolder.ver : Long.MAX_VALUE;
+
+                return new MappingsCacheValue(topVer, tableIds, mapFragments(context, template));
+            }
+
+            if (val.topVer < topologyHolder.ver) {
```
ygerzhedovich, line 148: `<=` ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/mapping/MappingServiceImpl.java
```diff
@@ -169,6 +223,8 @@ private CompletableFuture<List<MappedFragment>> map0(MultiStepPlan multiStepPlan
                             }
 
                             lastAttemptSucceed = true;
+
+                            break;
```
ygerzhedovich, line 227: good catch. I think potentially more obvious will be add the condition to for loop instead of `break` statement here

xtern: Done, thanks

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/mapping/MappingServiceImpl.java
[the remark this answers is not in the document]
xtern: Done, thanks

xtern: Result of benchmark for 10 times scan with random `removeIf(...)


| Benchmark   |    (cacheSize) | (tablesInPlan) |  (totalTables) |     Score   |    Error |  Units |
| ------------|----------------|----------------|----------------|-------------|----------|------- |
| 1 thread               |1024              |10          |10000  |    317.112 ±    |29.731  |us/op
| 8 threads              |1024              |10          |10000  |    286.909 ±     |5.741  |us/op
| 8 threads + sync       |1024              |10          |10000  |    294.775 ±     |4.226  |us/op
| ------------|----------------|----------------|----------------|-------------------------------- |
|1 thread               |1024             |100          |10000   |   340.265 ±    |15.269  |us/op
|8 threads              |1024             |100          |10000   |   287.221 ±    | 3.058  |us/op
|8 threads + sync       |1024             |100          |10000   |   349.131 ±    |49.229  |us/op
| ------------|----------------|----------------|----------------|-------------------------------- |
| 1 thread              |10000              |10          |10000  |   3136.671 ±   |596.822  |us/op
| 8 threads             |10000              |10          |10000  |   2822.520 ±   |280.104  |us/op
| 8 threads + sync      |10000              |10          |10000  |   3252.009 ±   |205.003  |us/op
| ------------|----------------|----------------|----------------|-------------------------------- |

| 1 thread              |10000             |100          |10000  |   3586.382 ±   |159.551  |us/op
| 8 threads             |10000             |100          |10000  |   3453.345 ±   |104.515  |us/op
| 8 threads + sync      |10000             |100          |10000  |   3641.002 ±   |192.040  |us/op
| ------------|----------------|----------------|----------------|-------------------------------- |

| 1 thread             |100000              |10          |10000  |  40390.912 ±  |5187.521  |us/op
| 8 threads            |100000              |10          |10000  |  38434.787 ±  |8073.539  |us/op
| 8 threads + sync     |100000              |10          |10000  |  41262.388 ±  |8081.194  |us/op
| ------------|----------------|----------------|----------------|-------------------------------- |
| 1 thread             |100000             |100          |10000  |  63233.467 ± |12400.052  |us/op
| 8 threads            |100000             |100          |10000  |  64562.198 ±  |7600.780  |us/op
| 8 threads + sync     |100000             |100          |10000  |  63495.429 ±  |6680.330  |us/op
| ---------------------|----------------|----------------|-------|---------------|----------|---------|

xtern: Benchmark scenario:
* create cache + fill cache
* measure the time of 10 calls `mappingsCache.values().removeIf(value -> value.tableIds.contains(RANDOM_TABLE_ID));`

Measuring time for single thread and for multiple threads.
`8 threads + sync` means that the `mappingsCache.values().removeIf(value -> value.tableIds.contains(RANDOM_TABLE_ID));` is enclosed in a synchronized block (to reduce possible contention when scanning).

`tablesInPlan` - size of set `tableIds`.
`totalTables` - total number of table IDs (the lower this number, the more cache elements will be deleted in one scan)

```
Benchmark     (cacheSize)  (tablesInPlan)  (totalTables)  Mode  Cnt       Score       Error  Units
1 thread             1024              10          10000  avgt    8     298.156 ±    16.015  us/op
8 threads            1024              10          10000  avgt    8     401.057 ±    14.550  us/op
8 threads + sync     1024              10          10000  avgt    8    1577.090 ±    77.315  us/op

1 thread             1024             100          10000  avgt    8     318.541 ±    12.416  us/op
8 threads            1024             100          10000  avgt    8     418.963 ±    30.124  us/op
8 threads + sync     1024             100          10000  avgt    8     878.342 ±    92.320  us/op

1 thread            10000              10          10000  avgt    8    2974.037 ±   115.337  us/op
8 threads           10000              10          10000  avgt    8    4595.802 ±   282.642  us/op
8 threads + sync    10000              10          10000  avgt    8   15558.993 ±  1107.365  us/op

1 thread            10000             100          10000  avgt    8    3481.707 ±    37.420  us/op
8 threads           10000             100          10000  avgt    8    4652.641 ±   124.261  us/op
8 threads + sync    10000             100          10000  avgt    8    8283.897 ±   588.364  us/op

1 thread           100000              10          10000  avgt    8   37814.228 ±  3901.386  us/op
8 threads          100000              10          10000  avgt    8   56042.789 ±  2641.133  us/op
8 threads + sync   100000              10          10000  avgt    8  170200.745 ± 50732.738  us/op

1 thread           100000             100          10000  avgt    8   62899.692 ± 11203.015  us/op
8 threads          100000             100          10000  avgt    8   66636.729 ±  6559.783  us/op
8 threads + sync   100000             100          10000  avgt    8  159911.761 ± 54040.121  us/op
```


The results obtained on my machine show that there are no contention between threads during this operation (or it is not significant).

ygerzhedovich (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/mapping/MappingServiceImpl.java
```diff
@@ -96,12 +109,53 @@ public CompletableFuture<List<MappedFragment>> map(MultiStepPlan multiStepPlan)
...
-        MappingContext context = new MappingContext(localNodeName, nodes);
+        TopSnapshot topology = topologyHolder.topology();
+        MappingContext context = new MappingContext(localNodeName, topology.nodes());
 
         FragmentsTemplate template = getOrCreateTemplate(multiStepPlan, context);
 
+        MappingsCacheValue cacheValue = mappingsCache.compute(multiStepPlan.id(), (key, val) -> {
+            if (val == null) {
+                IntSet tableIds = new IntOpenHashSet();
+                boolean topAware = false;
```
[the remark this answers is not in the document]
xtern: Done

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/mapping/MappingServiceImplTest.java
```diff
@@ -134,6 +164,66 @@ public void lateServiceInitializationOnNodeJoin() {
...
+
+        // Simulate expiration of the primary replica for non-mapped table - the cache entry should not be invalidated.
+        await(mappingService.onPrimaryReplicaExpired(prepareEvtParams.apply("T2")));
+        assertSame(mappedFragments, await(mappingService.map(PLAN_WITH_SYSTEM_VIEW)));
+
+        // Simulate expiration of the primary replica for mapped table - the cache entry should be invalidated.
+        await(mappingService.onPrimaryReplicaExpired(prepareEvtParams.apply("T1")));
+        assertNotSame(mappedFragments, await(mappingService.map(PLAN_WITH_SYSTEM_VIEW)));
+    }
+
```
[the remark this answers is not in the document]
xtern: Thanks, added additional checks to verify `ExecutionTargetProvider` calls.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/mapping/MappingServiceImpl.java
```diff
@@ -85,6 +94,10 @@ public MappingServiceImpl(
         this.targetProvider = targetProvider;
         this.templatesCache = cacheFactory.create(cacheSize);
         this.taskExecutor = taskExecutor;
+        this.mappingsCache = Caffeine.newBuilder()
```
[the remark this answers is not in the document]
xtern: Added two additional methods to internal cache interface.
`compute`
`removeIfValue` (this is debatable - instead of this weird method we can add a `values()` method). :thinking:

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/framework/TestNode.java
```diff
@@ -225,6 +229,10 @@ public void initSchema(String script) {
         }
     }
 
+    public CatalogService catalogService() {
+        return catalogService;
```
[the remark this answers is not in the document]
xtern: Done, refactored a bit (`ComponentToLifecycleAwareAdaptor` moved to `TestNode`).

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/cache/Cache.java
```diff
@@ -60,4 +62,21 @@ public interface Cache<K, V> {
...
     /** Clears the given cache. That is, remove all keys and associated values. */
     void clear();
+
+    /**
+     * Attempts to compute a mapping for the specified key and its current
+     * mapped value (or {@code null} if there is no current mapping).
+     *
+     * @param key key with which the specified value is to be associated
+     * @param remappingFunction the remapping function to compute a value
+     * @return the new value associated with the specified key, or null if none
```
korlov42, line 72:
```suggestion
     * @param key Key with which the specified value is to be associated.
     * @param remappingFunction The remapping function to compute a value.
     * @return The new value associated with the specified key, or null if none.
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/cache/Cache.java
```diff
@@ -60,4 +62,21 @@ public interface Cache<K, V> {
...
+     */
+    V compute(K key, BiFunction<? super K, ? super V, ? extends V> remappingFunction);
+
+    /**
+     * Removes all cache entries whose values match the specified predicate.
+     *
+     * @param valueFilter a predicate which returns {@code true} for the values of entries to be removed.
```
korlov42, line 79:
```suggestion
     * @param valueFilter A predicate which returns {@code true} for the values of entries to be removed.
```

korlov42 (approved):

PR #2987, opened 2023-12-25
IGNITE-21147 Transaction finish fails with a timeout exception

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-21147

Fixed the bug with a TimeoutException on transaction finish:
See `PartitionReplicaListener.appendTxCommand`.
The issue was that an RW operation adds a cleanup future to `txOps.futures` map and the future is completed based on the result of the transaction operation future. If a transaction fails to build the latter, for example because of a lock conflict, the future in  `txOps.futures` map will never be completed. Thus a subsequent finish will wait on that future until a timeout expires.

Another bug fixed in this commit:
`TransactionStateResolver.resolveTxStateFromTxCoordinator`: If a request to a coordinator fails with an exception (see `send`), the fallback to get the state from commit partition will not be triggered.

Also added Tx state propagation from the server to the client

### DIFF
--- a/modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/HeapLockManager.java
+++ b/modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/HeapLockManager.java
@@ -424,7 +424,11 @@ public class HeapLockManager extends AbstractEventProducer<LockEvent, LockEventP
                 LockMode mode = lockedMode(tmp);
 
                 if (mode != null && !mode.isCompatible(waiter.intendedLockMode())) {
-                    conflictFound(waiter.txId(), tmp.txId());
+                    if (conflictFound(waiter.txId(), tmp.txId())) {
+                        waiter.fail(lockException(waiter, tmp));
+
+                        return true;
+                    }
 
                     if (!deadlockPreventionPolicy.usePriority() && deadlockPreventionPolicy.waitTimeout() == 0) {
                         waiter.fail(lockException(waiter, tmp));
@@ -441,7 +445,11 @@ public class HeapLockManager extends AbstractEventProducer<LockEvent, LockEventP
                 LockMode mode = lockedMode(tmp);
 
                 if (mode != null && !mode.isCompatible(waiter.intendedLockMode())) {
-                    conflictFound(waiter.txId(), tmp.txId());
+                    if (conflictFound(waiter.txId(), tmp.txId())) {
+                        waiter.fail(lockException(waiter, tmp));
+
+                        return true;
+                    }
 
                     if (skipFail) {
                         return false;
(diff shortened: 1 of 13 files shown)


modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/OrphanDetector.java
```diff
@@ -195,7 +182,8 @@ private void checkTxOrphanedInternal(UUID txId) {
             sentTxRecoveryMessage(txState.commitPartitionId(), txId);
         }
 
-        throw new TransactionException(ACQUIRE_LOCK_ERR, "The lock is held by the abandoned transaction [abandonedTxId=" + txId + "].");
+        return failedFuture(
+                new TransactionException(ACQUIRE_LOCK_ERR, "The lock is held by the abandoned transaction [abandonedTxId=" + txId + "]."));
```
vldpyatkov, line 186: I don't think it is definitely right to ignore the exception. Let's add TODO here.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -530,7 +530,7 @@ private CompletableFuture<Void> makeFinishRequest(
                 commitPartition, primaryConsistentId, term, commit, txId, replicationGroupIds);
 
         return txMessageSender.finish(primaryConsistentId, commitPartition, replicationGroupIds, txId, term, commit, commitTimestamp)
-                .thenRun(() -> {
+                .thenAccept(txResult -> {
                     updateTxMeta(txId, old -> {
```
vldpyatkov, line 534: Might we complete txFinishFuture out of the update closuer to avoid invoking that undr a lock?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/HeapLockManager.java
```diff
@@ -668,8 +684,13 @@ private void track(UUID txId) {
          * @param acquirerTx Transaction which tries to acquire the lock.
          * @param holderTx Transaction which holds the lock.
          */
-        private void conflictFound(UUID acquirerTx, UUID holderTx) {
-            fireEvent(LockEvent.LOCK_CONFLICT, new LockEventParameters(acquirerTx, holderTx));
+        private boolean conflictFound(UUID acquirerTx, UUID holderTx) {
+            CompletableFuture<Void> eventResult = fireEvent(LockEvent.LOCK_CONFLICT, new LockEventParameters(acquirerTx, holderTx));
+            // No async handling is expected.
+            // TODO: https://issues.apache.org/jira/browse/IGNITE-21153
+            assert eventResult.isDone();
```
vldpyatkov, line 691: Add a comment to the assertion.
Assert without a comment is forbidden in our code style.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -530,26 +530,18 @@ private CompletableFuture<Void> makeFinishRequest(
...
-                        }
-
-                        assert old instanceof TxStateMetaFinishing;
-
-                        TxStateMeta finalTxStateMeta = coordinatorFinalTxStateMeta(
-                                commit,
-                                old.commitPartitionId(),
-                                commitTimestamp
-                        );
+                    assert isFinalState(updatedMeta.txState());
```
vldpyatkov, line 542: Add a message for this assertion.

vldpyatkov (approved):

vldpyatkov (approved):

PR #2989, opened 2023-12-26
IGNITE-21146 Error handling in Criteria queries

nva:

### The Review Checklist

### Notes


modules/api/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -577,4 +577,17 @@ public static class PlacementDriver {
...
+    @ErrorCodeGroup
+    public static class Criteria {
+        /** Criteria error group. */
+        public static final ErrorGroup CRITERIA_ERR_GROUP = registerGroup("CRITERIA", (short) 19);
+
+        /** No more pages in the cursor error. */
+        public static final int CRITERIA_CURSOR_NO_MORE_PAGES_ERR = CRITERIA_ERR_GROUP.registerErrorCode((short) 1);
+
+        /** Cursor is already closed error. */
+        public static final int CRITERIA_CURSOR_CLOSED_ERR = CRITERIA_ERR_GROUP.registerErrorCode((short) 2);
```
AMashenkov, line 591: Can this error be in common group?

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/CursorAdapter.java
```diff
@@ -54,9 +52,9 @@ public CursorAdapter(AsyncCursor<T> ac) {
     @Override
     public void close() {
         try {
-            convertToPublicFuture(ac.closeAsync().toCompletableFuture()).join();
-        } catch (Throwable e) {
-            throw sneakyThrow(mapToPublicException(unwrapCause(e)));
+            ac.closeAsync().toCompletableFuture().join();
```
AMashenkov, line 55: AFAIK, we agree to use `future.get()` instead of `join`, because `join` is uninteruptable.

modules/api/src/main/java/org/apache/ignite/lang/ErrorGroups.java
AMashenkov: I think these codes (+ similar SQL codes) are not specific to Criteria (or SQL) modules. Moreover, Cursors now are part of Public API, and we can introduce a common exception e.g. CursorAccessException for that purpose, and use it like an IllegalArgumentException for arguments validation.

modules/api/src/main/java/org/apache/ignite/table/criteria/CriteriaQuerySource.java
```diff
@@ -36,7 +35,8 @@ public interface CriteriaQuerySource<T> {
      * @param tx Transaction to execute the query within or {@code null} to run within implicit transaction.
      * @param criteria The predicate to filter entries or {@code null} to return all entries from the underlying table.
      * @return Iterator with query results.
-     * @throws IgniteException If failed.
+     * @throws CriteriaException If failed.
+     * @throws IllegalArgumentException If the names of the provided columns are incorrect.
```
AMashenkov, line 39: If you think the change is correct, then should we do the same for all public API methods?Because all (?) of public API methods validates argument and could throw this exception.

I think the exception mentioning can be dropped as it brings nothing and in normal case user can't handle and shouldn't care about it.  

Also, we can't foresee all the cases, when IllegalArgumentException could be thrown. E.g. also, the exception could be thrown if condition has an argument of unexpected type.

modules/api/src/main/java/org/apache/ignite/table/criteria/CriteriaQuerySource.java
```diff
@@ -59,7 +60,8 @@ default Cursor<T> query(@Nullable Transaction tx, @Nullable Criteria criteria) {
      * @param tx Transaction to execute the query within or {@code null} to run within implicit transaction.
      * @param criteria The predicate to filter entries or {@code null} to return all entries from the underlying table.
      * @return Future that represents the pending completion of the operation.
-     * @throws IgniteException If failed.
+     * @throws CriteriaException If failed.
+     * @throws IllegalArgumentException If the names of the provided columns are incorrect.
```
AMashenkov, line 64: Async methods return futures. 
Imho, async method should never throw an exception. They should return failed future instead.

nva: Replaced with failed future

modules/api/src/main/java/org/apache/ignite/table/criteria/CriteriaQuerySource.java
nva: Removed

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/CursorAdapter.java
nva: Replaced with `future.get()`

modules/api/src/main/java/org/apache/ignite/lang/ErrorGroups.java
nva: Moved to common group. Moved `CursorAccessException` to lang package. Added `NoMorePagesException`

modules/api/src/main/java/org/apache/ignite/lang/NoMorePagesException.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package org.apache.ignite.lang;
+
+import static org.apache.ignite.lang.ErrorGroups.Common.CURSOR_NO_MORE_PAGES_ERR;
+
+/**
+ * Exception is thrown when there are no more pages in the cursor.
+ */
+public class NoMorePagesException extends IgniteException {
```
AMashenkov, line 25: Seems, this class is not used.

nva: Restored the use of `NoMorePagesException` since `mapToPublicSqlException` will wrap `NoSuchElementException` into `IgniteException`

AMashenkov: 1. Do we really need this exception? 
It means a cursor reach end of data and can be closed. Maybe, we could close it instantly and throw CursorClosedException. Is there any case, when user need to know the real reason?

2. "Pages" in NoMorePagesException name looks a bit inconsistent with the method signature `util/AsyncCursor`.
It returns "batch", not a page.
`CompletableFuture<BatchedResult<T>> requestNextAsync(int rows);`
 However, public interface `lang/AsyncCursor` have different method.
`CompletableFuture<? extends AsyncCursor<T>> fetchNextPage();`

AMashenkov: > 1. On the other hand NoMorePagesException is a valid case, and you can complete the processing.
Cursor has methods `hasMorePages` or there is a flag in BatchedResult. So, NoMorePagesException will be thrown only if the one violates the contract.

AMashenkov (approved):

PR #2992, opened 2023-12-27
IGNITE-19854 Parameter metadata fetching for non-executed query

isapego:
- Added `ClientSqlParameterMetadataRequest` server side class;
- Implemented request sending on ODBC side;
- Un-muted old tests;
- Added a new test.


modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/sql/ClientSqlExecuteRequest.java
```diff
@@ -177,7 +177,7 @@ private static void packMeta(ClientMessagePacker out, @Nullable ResultSetMetadat
 
             out.packString(col.name());
             out.packBoolean(col.nullable());
-            out.packInt(col.type().ordinal());
+            out.packInt(col.type().id());
```
ptupitsyn, line 180: :+1:

ptupitsyn (approved):

PR #2997, opened 2023-12-27
IGNITE-20477 Introduce async components start

alievmirza:
https://issues.apache.org/jira/browse/IGNITE-20477

### The Review Checklist

### Notes


modules/catalog/src/testFixtures/java/org/apache/ignite/internal/catalog/CatalogTestUtils.java
```diff
@@ -155,10 +161,10 @@ public static CatalogManager createCatalogManagerWithTestUpdateLog(String nodeNa
 
         return new CatalogManagerImpl(new TestUpdateLog(clock), clockWaiter) {
             @Override
-            public void start() {
-                clockWaiter.start();
+            public CompletableFuture<Void> start() {
+                CompletableFuture<Void> fut = clockWaiter.start();
 
-                super.start();
+                return allOf(new CompletableFuture[]{fut, super.start()});
```
sashapolo, line 167: no need to create an array explicitly here, `allOf` has a vararg argument

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -855,19 +860,22 @@ public CompletableFuture<Ignite> start(Path configPath) {
...
                             throw new CompletionException(e);
                         }
                     }, startupExecutor)
                     .thenComposeAsync(v -> {
                         LOG.info("Components started, performing recovery");
 
-                        return recoverComponentsStateOnStart(startupExecutor);
+                        return recoverComponentsStateOnStart(
+                                startupExecutor,
+                                CompletableFuture.allOf(startFuturesAfterJoin.toArray(CompletableFuture[]::new))
```
sashapolo, line 877: Why do we need to pass this future here?

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -785,19 +786,23 @@ public CompletableFuture<Ignite> start(Path configPath) {
         ExecutorService startupExecutor = Executors.newSingleThreadExecutor(NamedThreadFactory.create(name, "start", LOG));
 
         try {
+            List<CompletableFuture<Void>> startFuturesBeforeJoin = new ArrayList<>();
```
sashapolo, line 789: This looks suspicious. I would expect all components to start sequentially for now (i.e. by chaining the start futures), not in parallel. This should probably be encapsulated inside the `lifecycleManager`.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -1098,11 +1107,11 @@ private CompletableFuture<Void> initializeClusterConfiguration(ExecutorService s
      * Recovers components state on start by invoking configuration listeners ({@link #notifyConfigurationListeners()} and deploying watches
      * after that.
      */
-    private CompletableFuture<?> recoverComponentsStateOnStart(ExecutorService startupExecutor) {
+    private CompletableFuture<?> recoverComponentsStateOnStart(ExecutorService startupExecutor, CompletableFuture<Void> startFuture) {
```
sashapolo, line 1110: The `startFuture` parameter looks redundant to me

modules/runner/src/main/java/org/apache/ignite/internal/app/LifecycleManager.java
```diff
@@ -81,10 +84,13 @@ void startComponent(IgniteComponent component) throws NodeStoppingException {
      * @param components Ignite components to start.
      * @throws NodeStoppingException If node stopping intention was detected.
      */
-    void startComponents(IgniteComponent... components) throws NodeStoppingException {
+    CompletableFuture<Void> startComponents(IgniteComponent... components) throws NodeStoppingException {
+        List<CompletableFuture<Void>> startFutures = new ArrayList<>();
```
sashapolo, line 88: This should be an array, not an `ArrayList`. This way we can avoid copying in the end. But I also think that components should be started sequentially and not in parallel, at least for now

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
alievmirza: This was done to satisfy the third item from the ticket

> IgniteImpl startup procedure should be updated to collect the futures from all components and wait for their completion inside recoverComponentsStateOnStart, after metaStorageMgr.notifyRevisionUpdateListenerOnStart() has been called, but before Watches are deployed (metaStorageMgr.deployWatches())

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
alievmirza: Why does it look suspicious? Synchronous parts still start sequentially, as it were before.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -806,15 +811,15 @@ public CompletableFuture<Ignite> start(Path configPath) {
...
             clusterSvc.updateMetadata(new NodeMetadata(restComponent.hostName(), restComponent.httpPort(), restComponent.httpsPort()));
 
             restAddressReporter.writeReport(restHttpAddress(), restHttpsAddress());
 
             LOG.info("Components started, joining the cluster");
 
-            return cmgMgr.joinFuture()
+            CompletableFuture<Ignite> startFuture = cmgMgr.joinFuture()
                     //Disable REST component during initialization.
                     .thenAcceptAsync(unused -> restComponent.disable())
```
sashapolo, line 824: Not related to this PR, but please add `startupExecutor` here, it's clearly missing

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -855,19 +860,22 @@ public CompletableFuture<Ignite> start(Path configPath) {
                                     clientHandlerModule,
                                     deploymentManager,
                                     sql
-                            );
+                            ));
 
                             // The system view manager comes last because other components
                             // must register system views before it starts.
-                            lifecycleManager.startComponent(systemViewManager);
+                            startFuturesAfterJoin.add(lifecycleManager.startComponent(systemViewManager));
```
sashapolo, line 867: Why isn't this component simply put add the end of the list above?

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
sashapolo: Why don't we account for `startFuturesBeforeJoin` here then?

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -834,7 +839,7 @@ public CompletableFuture<Ignite> start(Path configPath) {
 
                         // Start all other components after the join request has completed and the node has been validated.
                         try {
-                            lifecycleManager.startComponents(
+                            startFuturesAfterJoin.add(lifecycleManager.startComponents(
```
sashapolo, line 842: Can we change `runAsync` to `applyAsync` here and return `startFuturesAfterJoin` as the result of this function?

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -897,6 +905,7 @@ public CompletableFuture<Ignite> start(Path configPath) {
                     .whenCompleteAsync((res, ex) -> {
                         startupExecutor.shutdownNow();
                     });
+            return CompletableFuture.allOf(startFuturesBeforeJoin.toArray(CompletableFuture[]::new)).thenCompose((ignored) -> startFuture);
```
sashapolo, line 908: Why do we need `thenCompose` here, can we pass the `startFuture` to `allOf` instead?

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
alievmirza: We do, but when we return this future from `org.apache.ignite.internal.app.IgniteImpl#start`, there is chaining 
> return CompletableFuture.allOf(startFuturesBeforeJoin.toArray(CompletableFuture[]::new)).thenCompose((ignored) -> startFuture);

sashapolo: Why do we wait for `startFuturesAfterJoin` before waiting for `startFuturesBeforeJoin` then? I'm confused.

alievmirza: Where do we do that? We have chaining `startFuturesBeforeJoin` -> `startFuturesAfterJoin`

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
alievmirza: to be honest, I didn't understand your proposal

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
sashapolo: As you said above:

```
IgniteImpl startup procedure should be updated to collect the futures from all components and wait for their completion inside recoverComponentsStateOnStart, after metaStorageMgr.notifyRevisionUpdateListenerOnStart() has been called, but before Watches are deployed (metaStorageMgr.deployWatches())
```

You current logic is: wait for `startFuturesAfterJoin` and then deploy the watches. `startFuturesBeforeJoin` are not taken into account.

> Where do we do that? We have chaining startFuturesBeforeJoin -> startFuturesAfterJoin

Yes, you've added the chaining but you start all futures in parallel, so this chaining is useless in this regard.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
alievmirza: ok, I haven't thought about the comment, so your solution confused me

modules/runner/src/main/java/org/apache/ignite/internal/app/LifecycleManager.java
alievmirza: refactored

modules/catalog/src/testFixtures/java/org/apache/ignite/internal/catalog/CatalogTestUtils.java
alievmirza: done

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
alievmirza: refactored

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
alievmirza: refactored

modules/runner/src/main/java/org/apache/ignite/internal/app/LifecycleManager.java
```diff
@@ -105,6 +110,10 @@ void onStartComplete() throws NodeStoppingException {
         }
     }
 
+    synchronized CompletableFuture<Void> allComponentsStartFuture() {
```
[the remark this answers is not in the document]
alievmirza: done

modules/runner/src/main/java/org/apache/ignite/internal/app/LifecycleManager.java
```diff
@@ -135,5 +144,7 @@ private synchronized void stopAllComponents() {
                 LOG.warn("Unable to stop component [component={}, nodeName={}]", e, component, nodeName);
             }
         });
+
+        allComponentsStartFuture.clear();
```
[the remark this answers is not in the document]
alievmirza: done

modules/runner/src/main/java/org/apache/ignite/internal/app/LifecycleManager.java
```diff
@@ -110,8 +110,13 @@ void onStartComplete() throws NodeStoppingException {
...
     }
 
+    /**
+     * Represents future that will be completed when all components start futures will be completed.
+     *
+     * @return Future that will be completed when all components start futures will be completed.
+     */
     synchronized CompletableFuture<Void> allComponentsStartFuture() {
-        return allOf(allComponentsStartFuture.toArray(CompletableFuture[]::new));
+        return allOf(allComponentsStartFuture.toArray(CompletableFuture[]::new)).whenComplete((v, e) -> allComponentsStartFuture.clear());
```
sashapolo, line 119: This is a race, you need to use `syncrhonized` inside `whenComplete`

sashapolo (approved):

modules/runner/src/main/java/org/apache/ignite/internal/app/LifecycleManager.java
```diff
@@ -105,6 +110,20 @@ void onStartComplete() throws NodeStoppingException {
         }
     }
 
+    /**
+     * Represents future that will be completed when all components start futures will be completed.
+     *
+     * @return Future that will be completed when all components start futures will be completed.
+     */
+    synchronized CompletableFuture<Void> allComponentsStartFuture() {
```
kgusakov, line 118: So, every call for this method will produce new future with new "clear" flow. Could you add a comment at least, that this method is planned to call only once by design?

alievmirza: done

modules/core/src/main/java/org/apache/ignite/internal/manager/IgniteComponent.java
```diff
@@ -17,15 +17,19 @@
...
  */
 public interface IgniteComponent {
     /**
      * Starts the component. Depending on component flow both configuration properties listeners, meta storage watch registration, starting
      * thread pools and threads goes here.
+     *
+     * @return Future that will be completed when the asynchronous part of the start is processed.
      */
-    void start();
+    CompletableFuture<Void> start();
```
sanpwc, line 33: Right after the merge, please broadcast in ai3 slack channel that you've introduced an engine that allows to have an async components start and that no *new* async logic was added within your PR.

modules/runner/src/main/java/org/apache/ignite/internal/app/LifecycleManager.java
alievmirza: It was proposed by @sashapolo and I agree with him, because it is a potential memory leak

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/DistributionZoneManager.java
```diff
@@ -271,20 +263,10 @@ public void start() {
...
-
-            futures.add(createOrRestoreZonesStates(recoveryRevision));
-
-            futures.add(restoreLogicalTopologyChangeEventAndStartTimers(recoveryRevision));
-
-            try {
-                // TODO: return this futures to start method https://issues.apache.org/jira/browse/IGNITE-20477
-                allOf(futures.toArray(CompletableFuture[]::new)).get(START_TIMEOUT, MILLISECONDS);
-            } catch (Exception e) {
-                throw new IgniteException(Common.COMPONENT_NOT_STARTED_ERR, e);
```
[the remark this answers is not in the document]
alievmirza: Yes, that's correct

modules/runner/src/main/java/org/apache/ignite/internal/app/LifecycleManager.java
sashapolo: It was proposed by me, because we don't need these futures as soon as they are complete. What's the point of keeping them in memory for the whole node lifetime?

modules/core/src/main/java/org/apache/ignite/internal/manager/IgniteComponent.java
```diff
@@ -17,15 +17,19 @@
...
```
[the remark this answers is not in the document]
alievmirza: Do we really need this method right now? I mean, it is quiet easy to add it to `IgniteComponents` as `deafult` method, but in this PR we do not override this method in any class, so it will be just unused method. I would add it when somebody really needs this method in they logic.

modules/runner/src/main/java/org/apache/ignite/internal/app/LifecycleManager.java
sanpwc: In order to check whether component is started. I assume that when we will add true async logic we will have code similar to
aComponent.start() {
  bComponent.ready().thenRun(b -> b.fooBar());
}

modules/core/src/main/java/org/apache/ignite/internal/manager/IgniteComponent.java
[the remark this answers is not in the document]
sanpwc: I believe that we need it. Because without it, general async components infrastructure isn't complete. Basically that means that it's not possible to run safe async code just because it's not possible to verify whether dependent component is ready.

modules/runner/src/main/java/org/apache/ignite/internal/app/LifecycleManager.java
alievmirza: @sanpwc I didn't fully understand how this aggregated future relates to the async logic from the components. We can check that component is started using the API that you proposed for `IgniteComponents`

modules/core/src/main/java/org/apache/ignite/internal/manager/IgniteComponent.java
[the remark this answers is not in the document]
alievmirza: ok, I'll add this default method to `IgniteComponents`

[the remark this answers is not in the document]
sanpwc: Just in case, simple default method addition won't be enough of course.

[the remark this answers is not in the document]
alievmirza: Could you please explain then, what do you expect? Implementation of this method for all components?

modules/runner/src/main/java/org/apache/ignite/internal/app/LifecycleManager.java
sanpwc: component#ready or component#isStarted future will be used at any point of time, thus we will had to preserve them.
That means that we may clear only aggregated future. Slight optimization vs code simplicity, does it worth it?

sanpwc (approved):

PR #2999, opened 2023-12-28
IGNITE-21068 Ignite node must not communicate with a node removed from the Physical Topology

rpuch:
https://issues.apache.org/jira/browse/IGNITE-21068

### The Review Checklist

### Notes

### DIFF
--- a/modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
+++ b/modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
@@ -35,7 +35,6 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.ignite.internal.future.OrderingFuture;
 import org.apache.ignite.internal.lang.IgniteInternalException;
@@ -330,6 +329,15 @@ public class ConnectionManager implements ChannelCreationListener {
         // Old channel can still be in the map, but it must be closed already by the tie breaker in the
         // handshake manager.
         assert oldChannel == null || !oldChannel.isOpen() : "Incorrect channel creation flow";
+
+        // Preventing a race between calling closeConnectionsWith() and putting a new channel that was just opened (with the node
+        // which is already stale). If it's stale, then the stale detector already knows it (and it knows it before
+        // closeConnectionsWith() gets called as it subscribes very first).
+        // This is the only place where a new sender might be added to the map.
+        if (staleIdDetector.isIdStale(channel.launchId())) {
+            channel.close();
+            channels.remove(key, channel);
+        }
     }
 
     /**
--- a/modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
+++ b/modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
@@ -18,6 +18,7 @@
 package org.apache.ignite.internal.network.netty;
 
 import static java.util.function.Function.identity;
+import static java.util.stream.Collectors.toList;
 import static org.apache.ignite.network.ChannelType.getChannel;
 
 import io.netty.bootstrap.Bootstrap;
@@ -26,6 +27,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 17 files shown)


modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
```diff
@@ -327,6 +329,15 @@ public void handshakeFinished(NettySender channel) {
         // Old channel can still be in the map, but it must be closed already by the tie breaker in the
         // handshake manager.
         assert oldChannel == null || !oldChannel.isOpen() : "Incorrect channel creation flow";
+
+        // Preventing a race between calling closeConnectionsWith() and putting a new channel that was just opened (with the node
+        // which is already stale). If it's stale, then the stale detector already knows it (and it knows it before
+        // closeConnectionsWith() gets called as it subscribes very first).
```
sashapolo, line 335:
```suggestion
        // closeConnectionsWith() gets called as it subscribes first).
```

modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
```diff
@@ -486,4 +497,22 @@ public Map<ConnectorKey<String>, NettySender> channels() {
...
     }
+
+    /**
+     * Closes physical connections with an Ignite node identified by the given ID (it's not consistentId,
+     * it's ID that gets regenerated at each node restart).
+     *
+     * @param id ID of the node.
+     */
+    public void closeConnectionsWith(String id) {
+        List<Entry<ConnectorKey<String>, NettySender>> entriesToRemove = channels.entrySet().stream()
```
sashapolo, line 508: Can we use `removeIf` instead?

modules/network/src/main/java/org/apache/ignite/internal/network/netty/MessageHandler.java
```diff
@@ -48,9 +50,15 @@ public class MessageHandler extends ChannelInboundHandlerAdapter {
      * @param consistentId Consistent id of the remote node.
      * @param serializationService Serialization service.
      */
-    public MessageHandler(Consumer<InNetworkObject> messageListener, String consistentId, short connectionIndex,
-            PerSessionSerializationService serializationService) {
+    public MessageHandler(
```
sashapolo, line 53: Please update the javadoc

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
```diff
@@ -83,16 +83,26 @@ public ClusterService createClusterService(
...
-        var messageFactory = new NetworkMessagesFactory();
-
         var topologyService = new ScaleCubeTopologyService();
 
+        // Adding this handler as the first handler to make sure that StaleIds is at least up-to-date as any
+        // other component that watches topology events.
+        topologyService.addEventHandler(new TopologyEventHandler() {
+            @Override
+            public void onDisappeared(ClusterNode member) {
+                staleIds.markAsStale(member.id());
```
sashapolo, line 93: Why don't we call `connectionManager.closeConnectionsWith(member.id());` here as well?

modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
rpuch: This would mean the following:

1. The predicate would have a side effect (which is weird; `close()` is probably idempotent, but still)
2. `close()` invoked inside a concurrent map, so probably under a lock. This is scary to do; also, channel closure might trigger some events (reacting to the closure); if they go to the same map, a deadlock might happen

I don't think that the convenience of `removeIf()` is worth it given the necessity to reason about all this.

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
rpuch: First handler is added in the very beginning of `createClusterService` (before start), to make sure that it gets notified first. On the other hand, to close connections, we need an instance of `ConnecitonManager` which is created on start (as it needs configuration). So it doesn't seem we should join them in the same handler.

I agree that it's a bit weird to add this handler in `DefaultMessagingService`, so I moved it to `ScaleCubeClusterServiceFactory` as well.

WDYT?

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
```diff
@@ -135,6 +135,15 @@ public void start() {
 
                 connectionMgr.start();
 
+                topologyService.addEventHandler(new TopologyEventHandler() {
+                    private final ConnectionManager capturedConnectionManager = connectionMgr;
```
sashapolo, line 139: Not a big deal, but maybe we should extract all volatile fields into local variables and only assign them at the end of the method? This way we can avoid declaring this field and possibly avoid volatile reads.

modules/network/src/main/java/org/apache/ignite/network/scalecube/ScaleCubeClusterServiceFactory.java
sashapolo: Not ideal, but definitely better than before.

PR #3035, opened 2024-01-11
IGNITE-21155 Adding a in-flight RW transaction counter by catalog version on a node

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-21155

### The Review Checklist

### Notes


modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -598,6 +634,7 @@ public void start() {
     public void beforeNodeStop() {
         orphanDetector.stop();
         txStateVolatileStorage.stop();
+        rwTxCounterByCatalogVersion.clear();
```
rpuch, line 637: Why do we need this? Will not this allow a false 'yes, all transactions have finished' response to be sent, even though some transactions are still pending?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -854,4 +891,66 @@ static boolean isRecoverable(Throwable throwable) {
...
+    }
+
+    /**
+     * Returns {@code true} iff the requested catalog version is registered locally and all RW transactions before it have completed,
+     * otherwise {@code false}.
+     *
+     * @param catalogVersion Catalog version to check.
+     */
+    boolean isRwTransactionsFinished(int catalogVersion) {
+        if (catalogVersion > catalogService.latestCatalogVersion()) {
```
rpuch, line 939: Here, we need to look at the 'currently active' version of the catalog (that is, the version that is active at ts=now), not just the 'latest known version of the catalog'

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -232,9 +246,11 @@ public InternalTransaction begin(HybridTimestampTracker timestampTracker) {
     public InternalTransaction begin(HybridTimestampTracker timestampTracker, boolean readOnly) {
         HybridTimestamp beginTimestamp = clock.now();
         UUID txId = transactionIdGenerator.transactionIdFor(beginTimestamp);
-        updateTxMeta(txId, old -> new TxStateMeta(PENDING, localNodeId, null, null));
+        updateTxMeta(txId, old -> new TxStateMeta(PENDING, localNodeId, null, null, readOnly));
 
         if (!readOnly) {
+            rwTxCounterByCatalogVersion.incrementTxCount(txCatalogVersion(beginTimestamp));
```
rpuch, line 252: There is a problem with this approach. As we resolve catalog version by ts without doing a proper schema sync (to avoid waiting), we might get different results when doing increment and decrement, so we might increment for one catalog version, but decrement for a later version, so the transaction might make the index await hang for a long time (till next catalog update arrives).

It seems that we can solve this by resolving the catalog version by the ts (in the same way it's done now) before doing increment and save the catalog version in `TxstateMeta`; then, when doing decrement, use this version from the catalog version.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -854,4 +891,66 @@ static boolean isRecoverable(Throwable throwable) {
...
+
+    /**
+     * Returns {@code true} iff the requested catalog version is registered locally and all RW transactions before it have completed,
+     * otherwise {@code false}.
+     *
+     * @param catalogVersion Catalog version to check.
+     */
+    boolean isRwTransactionsFinished(int catalogVersion) {
+        if (catalogVersion > catalogService.latestCatalogVersion()) {
+            // Requested catalog version has not yet registered locally.
```
rpuch, line 940: Another problem is that this check is not atomic wrt the increment, so it might happen that we respond with 'we don't have transactions started on versions before bersion 10' and then start a transaction on version 9 (or even an earlier version).

I think this can be resolved if we add a new field for tracking the most recent **activated** catalog version we know about (`lastKnownActiveCatalogVer`). We will update it here, just before doing the check. (The update can be done under a write lock). When doing an increment, we would do a loop:

1. beginTs = now
2. catalogVer  = catalogVerByTs(beginTs)
3. [under read lock] if catalogVer >= lastKnownActiveCatalogVer then do increment; else goto 1

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -854,4 +891,66 @@ static boolean isRecoverable(Throwable throwable) {
...
+            return false;
+        }
+
+        return !rwTxCounterByCatalogVersion.isExistsTxBefore(catalogVersion);
+    }
+
+    private void decrementRwTxCount(TxStateMeta txStateMeta, UUID txId) {
```
rpuch, line 947: This new logic about maintaining counters and responding to messages increased `TxManagerImpl` quite a bit. It seems that it might make sense to create another class just to encapsulate this logic and put all the new methods and fields in it, to make sure this logic does not get in the way of perceiving the main transaction-related logic. WDYT?

rpuch (changes requested):

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/replication/PartitionReplicaListenerTest.java
```diff
@@ -690,11 +691,11 @@ public void testReadOnlySingleRowReplicaRequestResolveWriteIntentCommitted() thr
         BinaryRow testBinaryKey = nextBinaryKey();
         BinaryRow testBinaryRow = binaryRow(key(testBinaryKey), new TestValue(1, "v1"));
         var rowId = new RowId(PART_ID);
-        txState = TxState.COMMITTED;
+        txState = COMMITTED;
 
         pkStorage().put(testBinaryRow, rowId);
         testMvPartitionStorage.addWrite(rowId, testBinaryRow, txId, TABLE_ID, PART_ID);
-        txManager.updateTxMeta(txId, old -> new TxStateMeta(TxState.COMMITTED, localNode.id(), commitPartitionId, clock.now()));
+        txManager.updateTxMeta(txId, old -> new TxStateMeta(COMMITTED, localNode.id(), commitPartitionId, clock.now(), true));
```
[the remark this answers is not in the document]
tkalkirill: If you look at the name of the test and what is being done below, it made me think that this test still means read-only transactions. Maybe @sanpwc will help resolve this issue/dispute.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/replication/PartitionReplicaListenerTest.java
```diff
@@ -712,7 +713,7 @@ public void testReadOnlySingleRowReplicaRequestResolveWriteIntentPending() throw
 
         pkStorage().put(testBinaryRow, rowId);
         testMvPartitionStorage.addWrite(rowId, testBinaryRow, txId, TABLE_ID, PART_ID);
-        txManager.updateTxMeta(txId, old -> new TxStateMeta(TxState.PENDING, localNode.id(), commitPartitionId, null));
+        txManager.updateTxMeta(txId, old -> new TxStateMeta(PENDING, localNode.id(), commitPartitionId, null, true));
```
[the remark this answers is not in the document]
tkalkirill: Same

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/replication/PartitionReplicaListenerTest.java
```diff
@@ -727,11 +728,11 @@ public void testReadOnlySingleRowReplicaRequestResolveWriteIntentAborted() throw
         BinaryRow testBinaryKey = nextBinaryKey();
         BinaryRow testBinaryRow = binaryRow(key(testBinaryKey), new TestValue(1, "v1"));
         var rowId = new RowId(PART_ID);
-        txState = TxState.ABORTED;
+        txState = ABORTED;
 
         pkStorage().put(testBinaryRow, rowId);
         testMvPartitionStorage.addWrite(rowId, testBinaryRow, txId, TABLE_ID, PART_ID);
-        txManager.updateTxMeta(txId, old -> new TxStateMeta(TxState.ABORTED, localNode.id(), commitPartitionId, null));
+        txManager.updateTxMeta(txId, old -> new TxStateMeta(ABORTED, localNode.id(), commitPartitionId, null, true));
```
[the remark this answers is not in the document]
tkalkirill: same

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/ReadOnlyTransactionImpl.java
```diff
@@ -116,7 +116,13 @@ public CompletableFuture<Void> finish(boolean commit, HybridTimestamp executionT
         return ((TxManagerImpl) txManager).completeReadOnlyTransactionFuture(new TxIdAndTimestamp(readTimestamp, id()))
                 .thenRun(() -> txManager.updateTxMeta(
                         id(),
-                        old -> new TxStateMeta(COMMITTED, old.txCoordinatorId(), old.commitPartitionId(), old.commitTimestamp())
+                        old -> new TxStateMeta(
+                                COMMITTED,
+                                old.txCoordinatorId(),
+                                old.commitPartitionId(),
+                                old.commitTimestamp(),
+                                old.readOnly()
```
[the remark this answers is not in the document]
tkalkirill: Why not

modules/transactions/src/main/java/org/apache/ignite/internal/tx/message/RwTransactionsFinishedRequest.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.tx.message;
+
+import org.apache.ignite.network.NetworkMessage;
+import org.apache.ignite.network.annotations.Transferable;
+
+/** Response to check that the requested catalog version has been reached and all RW transactions prior to it have completed. */
```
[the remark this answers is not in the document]
tkalkirill: fix it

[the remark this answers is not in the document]
tkalkirill: fix it

modules/transactions/src/main/java/org/apache/ignite/internal/tx/message/RwTransactionsFinishedRequest.java
```diff
@@ -0,0 +1,28 @@
...
+@Transferable(TxMessageGroup.RW_TRANSACTIONS_FINISHED_REQUEST)
+public interface RwTransactionsFinishedRequest extends NetworkMessage {
+    /** Returns the catalog version of interest. */
+    int targetSchemaVersion();
```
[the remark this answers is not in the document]
tkalkirill: I took the name from IGNITE-21112, I can rename it without any problems.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/message/RwTransactionsFinishedResponse.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.tx.message;
+
+import org.apache.ignite.network.NetworkMessage;
+import org.apache.ignite.network.annotations.Transferable;
+
+/** Request to check that the requested catalog version has been reached and all RW transactions prior to it have completed. */
```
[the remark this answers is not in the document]
tkalkirill: fix it

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
tkalkirill: I think you're right, I'll move it to TxManagerImpl#stop.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxStateMeta.java
```diff
@@ -42,21 +42,10 @@ public class TxStateMeta implements TransactionMeta {
...
      */
-    public TxStateMeta(
-            TxState txState,
-            @Nullable String txCoordinatorId,
-            @Nullable TablePartitionId commitPartitionId,
-            @Nullable HybridTimestamp commitTimestamp
-    ) {
-        this(txState, txCoordinatorId, commitPartitionId, commitTimestamp, 0);
-    }
+    private final @Nullable Boolean readOnly;
```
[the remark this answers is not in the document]
rpuch: Is it about transactions which belong to other coordinators?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
tkalkirill: Whu not

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
tkalkirill: You're right, I'll fix it as you suggested.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
tkalkirill: At the moment, the latest version of the catalog is the last active version of the catalog, you can check it in the code.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
tkalkirill: When working with a catalog, we always deal with active versions of the catalog; when we get the latest version, we get the latest active version.

We will have an additional mechanism for each partition, which will allow us to wait for all operations until the version in which the index appears, I think we can cover corner cases with it.

Otherwise, we will need to do a blocking algorithm, which we would like to avoid.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
tkalkirill: We discussed it in person and I corrected it.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -921,37 +912,17 @@ private int txCatalogVersion(HybridTimestamp beginTs) {
...
 
-    /**
-     * Returns the transaction catalog version.
-     *
-     * @param txId Transaction ID.
-     */
-    private int txCatalogVersion(UUID txId) {
-        return txCatalogVersion(TransactionIds.beginTimestamp(txId));
+    /** See {@link RwTxFinishedHandler#isRwTransactionsFinished(int)}. */
+    boolean isRwTransactionsFinished(int catalogVersion) {
```
rpuch, line 916: Do we need it here just for tests? Can't the test be rewritten to actually test `RwTxFinishedHandler`? (because now this method just delegates to the corresponding method in `RwTxFinishedHandler`)

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/RwTxFinishedHandler.java
```diff
@@ -0,0 +1,142 @@
...
+     * otherwise {@code false}.
+     *
+     * @param catalogVersion Catalog version to check.
+     */
+    boolean isRwTransactionsFinished(int catalogVersion) {
+        checkFinishedReadWriteLock.writeLock().lock();
+
+        try {
+            if (catalogVersion > catalogService.activeCatalogVersion(clock.nowLong())) {
+                // Requested catalog version has not yet registered locally.
```
rpuch, line 122:
```suggestion
                // Requested catalog version has not yet registered or activated locally.
```

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -253,12 +250,13 @@ public InternalTransaction begin(HybridTimestampTracker timestampTracker, boolea
 
     @Override
     public InternalTransaction begin(HybridTimestampTracker timestampTracker, boolean readOnly, TxPriority priority) {
-        HybridTimestamp beginTimestamp = clock.now();
-        UUID txId = transactionIdGenerator.transactionIdFor(beginTimestamp);
-        updateTxMeta(txId, old -> new TxStateMeta(PENDING, localNodeId, null, null, readOnly));
+        HybridTimestamp beginTimestamp = newBeginTimestamp(readOnly);
```
rpuch, line 253: It seems there is still a race. Let's imagine there are 2 threads: T1 starts a transaction, T2 checks whether there are not transactions started on versions strictly before 2.

1. Initially, all counters are 0
2. T1: Takes read lock, takes beginTs=now() [beginTs=10], releases the lock
3. T1: Resolves catalog version by beginTs, it's 1
4. T2: Takes write lock
5. T2: Takes now() (it's 12), resolves catalog version for 12, it's 2
6. T2: Sees that the version of interest (2) is already activated and all counters are 0, responds with TRUE
7. T1: Increments count for catalog version = 1 (which should never happen)

It seems that this can be fixed if the increment happens under read lock **together** with a check that the catalog version we just obtained is not less than some version for which we already responded 'yes' while handling a message.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/message/RwTransactionsFinishedResponse.java
```diff
@@ -20,7 +20,10 @@
 import org.apache.ignite.network.NetworkMessage;
 import org.apache.ignite.network.annotations.Transferable;
 
-/** Request to check that the requested catalog version has been reached and all RW transactions prior to it have completed. */
+/**
+ * Response to check that the requested catalog version has been reached and all RW transactions started on versions before it have
```
rpuch, line 24:
```suggestion
 * Response to check that the requested catalog version has been activated and all RW transactions started on versions before it have
```

It seems we need to be specific that is activation which is reached.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
rpuch: The test that uses is an integration test, so what I suggested is probably not too relevant.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxStateMeta.java
```diff
@@ -42,11 +42,14 @@ public class TxStateMeta implements TransactionMeta {
...
 
     /**
-     * {@code True} for a read-only transaction, {@code false} for a read-write transaction and {@code null} if unknown, for example, if
+     * {@code True} for a read-only transaction, {@code false} for a read-write transaction. {@code null} if unknown, for example, if
      * there is no previous meta.
      */
     private final @Nullable Boolean readOnly;
 
+    /** Catalog version of transaction at its beginning timestamp. {@code null} if unknown, for example, if there is no previous meta. */
+    private final @Nullable Integer catalogVersion;
```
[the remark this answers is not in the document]
tkalkirill: I suggest leaving the name as is, it will be possible to reflect in the documentation everything that you highlighted.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/RwTxFinishedHandler.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    /** Starts the handler. */
+    void start() {
+        messagingService.addMessageHandler(TxMessageGroup.class, this::onReceiveTxNetworkMessage);
+    }
+
+    /** Stops the handler. */
+    void stop() {
+        rwTxCounterByCatalogVersion.clear();
```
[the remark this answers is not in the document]
tkalkirill: I don’t think it’s needed, this is an auxiliary handler that is used inside the `TxManagerImpl`.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
tkalkirill: Yes, this method is now used only in integration tests; unit tests do not conveniently cover the logic described in the ticket. I don’t know the best way to test the logic from the ticket; making the handler external doesn’t seem like the right approach. If you have a suggestion, then welcome.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/RwTxFinishedHandler.java
tkalkirill: fix it

modules/transactions/src/main/java/org/apache/ignite/internal/tx/message/RwTransactionsFinishedResponse.java
tkalkirill: fix it

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
tkalkirill: Fix it

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -629,14 +636,22 @@ public CompletableFuture<Void> start() {
 
         messagingService.addMessageHandler(ReplicaMessageGroup.class, this);
 
+        messagingService.addMessageHandler(TxMessageGroup.class, (message, senderConsistentId, correlationId) -> {
+            rwTxFinishedHandlerLock.writeLock().lock();
```
rpuch, line 640: It looks like we can move `rwTxFinishedHandlerLock`, handler subscription, this method and `newBeginTimestamp()` back to `RwTimestampFinishedHandler`; it seems natural to have it there and not clutter `TxManagerImpl`.

Also, the write lock now covers too much, even sending response is included. It seems that it's enough to only do `isRwTransactionsFinished()` under the lock.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/RwTxFinishedHandler.java
```diff
@@ -0,0 +1,115 @@
...
+import org.apache.ignite.internal.tx.message.RwTransactionsFinishedRequest;
+import org.apache.ignite.internal.tx.message.TxMessageGroup;
+import org.apache.ignite.internal.tx.message.TxMessagesFactory;
+import org.apache.ignite.network.MessagingService;
+import org.apache.ignite.network.NetworkMessage;
+import org.apache.ignite.network.NetworkMessageHandler;
+import org.jetbrains.annotations.Nullable;
+
+/** Auxiliary {@link RwTransactionsFinishedRequest} handler to encapsulate the logic associated with this handling. */
+class RwTxFinishedHandler {
```
sanpwc, line 32: Naming is confusing. Same as for `RwTransactionsFinishedRequest` TxFinish request has another meaning in tx flow. I also doubt that given requests should be in tx module because it's not a part of tx protocol, should we move it to index module or something?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -858,4 +917,41 @@ static boolean isRecoverable(Throwable throwable) {
             return false;
         }
     }
+
+    private void decrementRwTxCount(TxStateMeta txStateMeta) {
```
sanpwc, line 921: As 've mentioned I'd consider moving it to index related modules. Within transactions scope it should be required to have extended txnStateMeta with tx type and catalog version.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxStateMeta.java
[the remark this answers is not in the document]
rpuch: Making a schema sync might add a wait, and it returns a future, so we would have to make `TxManager#begin()` asynchronous. It seems to work without making a schema sync, so why add it :)

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
tkalkirill: Fix it

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -903,11 +918,34 @@ static boolean isRecoverable(Throwable throwable) {
         }
     }
 
-    /**
-     * Returns the transaction catalog version.
-     *
-     * @param beginTs Transaction begin timestamp.
-     */
+    private void decrementRwTxCount(TxStateMeta txStateMeta) {
+        rwTxFinishedHandlerLock.readLock().lock();
```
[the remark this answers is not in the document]
tkalkirill: Disscuss it

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
tkalkirill: fix it

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -576,14 +576,16 @@ public class IgniteImpl implements Ignite {
 
         // TODO: IGNITE-19344 - use nodeId that is validated on join (and probably generated differently).
         txManager = new TxManagerImpl(
+                name,
```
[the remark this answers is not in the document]
tkalkirill: It could be used, but there is a nuance: `clusterSvc.topologyService().localMember().name()` can only be used during the start of components and not in the constructor, for now I will leave the constructor parameter.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
tkalkirill: Fix it

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/RwTxFinishedHandler.java
tkalkirill: Yes, now this is used by indexes, but for example, we can later easily adapt the logic for other algorithms, for example, for collecting statistics about transactions.
We could somehow move this into the index module, but it is important for us to ensure atomicity and consistency when processing messages. I don’t know how convenient and beautiful it is to provide this during the creation of a transaction and it looks cumbersome in reality.

PR #3040, opened 2024-01-12
IGNITE-20381 Prevent modification operations call after CatalogManagerImpl is closed

zstan:
…rImpl is closed

### The Review Checklist

### Notes

### DIFF
--- a/modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogManagerImpl.java
+++ b/modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogManagerImpl.java
@@ -23,7 +23,6 @@ import static java.util.concurrent.CompletableFuture.failedFuture;
 import static org.apache.ignite.internal.catalog.commands.CatalogUtils.fromParams;
 import static org.apache.ignite.internal.util.CollectionUtils.nullOrEmpty;
 import static org.apache.ignite.internal.util.CompletableFutures.nullCompletedFuture;
-import static org.apache.ignite.internal.util.IgniteUtils.inBusyLock;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -177,6 +176,7 @@ public class CatalogManagerImpl extends AbstractEventProducer<CatalogEvent, Cata
     @Override
     public void stop() throws Exception {
         busyLock.block();
+        versionTracker.close();
         updateLog.stop();
     }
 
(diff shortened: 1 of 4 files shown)


modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogManagerImpl.java
```diff
@@ -321,8 +328,10 @@ public CompletableFuture<Void> execute(List<CatalogCommand> commands) {
     }
 
     private void registerCatalog(Catalog newCatalog) {
-        catalogByVer.put(newCatalog.version(), newCatalog);
-        catalogByTs.put(newCatalog.time(), newCatalog);
+        inBusyLock(busyLock, () -> {
+            catalogByVer.put(newCatalog.version(), newCatalog);
+            catalogByTs.put(newCatalog.time(), newCatalog);
+        });
```
AMashenkov, line 334: This method is called in `OnUpdateHandlerImpl.handle(...)`.
` handleFutures.add(onUpdateHandler.handle(update, event.timestamp(), event.revision()));`
It make sense to guard the `UpdateListener.onUpdate(WatchEvent)` with busylock instead.

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogManagerImpl.java
```diff
@@ -170,6 +176,7 @@ public void start() {
 
     @Override
     public void stop() throws Exception {
+        busyLock.block();
```
AMashenkov, line 179: Should we close versionTracker to unblock all the waiters?

AMashenkov: And is it ok to allow getting latest catalog version after CatalogManager was stopped?

We may violate invariant and the one will wait for a new catalog version, but the new version event will never be accepted due to stopping.

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogManagerImpl.java
zstan: seems no need to use busyLock here it correctly stops before.

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogManagerImpl.java
zstan: done, tests fixed

AMashenkov (approved):

PR #3061, opened 2024-01-17
IGNITE-21155 Adding a in-flight RW transaction counter by catalog version on a node

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-21155

### The Review Checklist

### Notes


modules/index/src/main/java/org/apache/ignite/internal/index/message/IsNodeReadyToStartBuildingIndexRequest.java
```diff
@@ -0,0 +1,36 @@
...
+ * Request to check whether a node is ready to start building an index.
+ *
+ * <p>To be able to safely start building an index on a node, the following conditions must be met:</p>
+ * <ul>
+ *     <li>Catalog version in which the index appeared must be active on the node.</li>
+ *     <li>All read-write transactions up to the catalog version in which the index appeared must be completed.</li>
+ * </ul>
+ */
+@Transferable(IndexMessageGroup.IS_NODE_READY_TO_START_BUILDING_INDEX_REQUEST)
+public interface IsNodeReadyToStartBuildingIndexRequest extends NetworkMessage {
```
rpuch, line 33: We'll need same waiting logic when dropping indexes, see https://issues.apache.org/jira/browse/IGNITE-21117

So it makes sense to generalize the name a little bit. `IsNodeReadyToStartBuildingIndex` can be renamed to `IsNodeFinishedRwTransactionsStartedBefore`. This is a bit too long, probably it can be shortened. The same rename should probably be made in other places.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -185,16 +185,19 @@ public class TxManagerImpl implements TxManager, NetworkMessageHandler {
      * @param transactionIdGenerator Used to generate transaction IDs.
      * @param placementDriver Placement driver.
      * @param idleSafeTimePropagationPeriodMsSupplier Used to get idle safe time propagation period in ms.
+     * @param localRwTxCounter Counter of read-write transactions that were created and completed locally on the node.
      */
     public TxManagerImpl(
+            String nodeName,
```
rpuch, line 191: `nodeName` can be obtained from `ClusterService`, so it's redundant here probably

tkalkirill: You're right, I'll fix it.

modules/index/src/main/java/org/apache/ignite/internal/index/message/IsNodeReadyToStartBuildingIndexRequest.java
tkalkirill: Why not

modules/index/src/main/java/org/apache/ignite/internal/index/IndexNodeFinishedRwTransactionsChecker.java
```diff
@@ -139,37 +140,37 @@ public <T> T inUpdateRwTxCountLock(Supplier<T> supplier) {
...
+                    FACTORY.isNodeFinishedRwTransactionsStartedBeforeResponse().finished(finished).build(),
                     correlationId
             );
         });
     }
 
     /**
-     * Returns {@code true} iff the catalog version in which the index appeared is active and all read-write transactions up to the
-     * requested catalog version have been completed on the node.
+     * Returns {@code true} iff the requested catalog version is active and all RW transactions up to that version have finished on the
```
rpuch, line 168: It seems that 'up to' means that the upper bound is included, but in our case it's not. Please rephrase, like 'and all RW transactions started on versions strictly before that version'...

rpuch (approved):

modules/index/src/main/java/org/apache/ignite/internal/index/IndexNodeFinishedRwTransactionsChecker.java
tkalkirill: why not

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -342,6 +344,8 @@ public CompletableFuture<Void> finish(
             // If there are no enlisted groups, just update local state - we already marked the tx as finished.
             updateTxMeta(txId, old -> new TxStateMeta(commit ? COMMITTED : ABORTED, localNodeId, commitPartition, commitTimestamp(commit)));
 
+            decrementRwTxCount(txId);
```
sanpwc, line 347: Finish might be called multiple times for the same transaction, thus I'd say that it would be better to handle decrement in an idempotent manner.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -395,7 +399,12 @@ public CompletableFuture<Void> finish(
                         enlistedGroups,
                         txId,
                         finishingStateMeta.txFinishFuture()
-                ));
+                )
+        ).thenAccept(unused -> {
+            if (localNodeId.equals(finishingStateMeta.txCoordinatorId())) {
```
sanpwc, line 404: I believe that it'll work. However in case of using txId as in inner counter identifier instead of beginTs you may just skip this check. Decrementing the counter wihtin recovery on non coordinator would just have no effect because there won't be any transactions with given id. Currently while using beginTs it won't work, because beginTs is unique only within node. Up to you.

modules/index/src/integrationTest/java/org/apache/ignite/internal/index/ItIndexNodeFinishedRwTransactionsCheckerTest.java
```diff
@@ -0,0 +1,230 @@
...
+
+    private static int latestCatalogVersion() {
+        return node().catalogManager().latestCatalogVersion();
+    }
+
+    private static IgniteImpl node() {
+        return CLUSTER.node(0);
+    }
+
+    private static boolean isNodeFinishedRwTransactionsStartedBeforeFromNetwork(int catalogVersion) {
```
sanpwc, line 219: Not sure. Do you have tests, that verify that in case of 1PC transaction we will increment and decrement the counter.
I mean simple KV.put(null, 1, "val1");  It's important to verify that increment and decrement were called.

tkalkirill: Added a test to check 1PC `org.apache.ignite.internal.index.ItIndexNodeFinishedRwTransactionsCheckerTest#testOnePhaseCommitViaKeyValue`

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
tkalkirill: fix it

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
tkalkirill: I made changes about the independence of the decrement, I think this should be enough.

modules/index/src/integrationTest/java/org/apache/ignite/internal/index/ItIndexNodeFinishedRwTransactionsCheckerTest.java
sanpwc: It's not 1PC. Please do not create an explicit transaction, let it be an implicit one. I mean, literally do KV.put() without any runInTx() or similar.

tkalkirill: Fix it

sanpwc (approved):

PR #3068, opened 2024-01-20
IGNITE-21227 Monitor network critical threads for being blocked

rpuch:
https://issues.apache.org/jira/browse/IGNITE-21227

### The Review Checklist

### Notes

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/worker/CriticalSingleThreadExecutor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/worker/CriticalSingleThreadExecutor.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.worker;
+
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Single thread executor instrumented to be used as a {@link CriticalWorker} and being monitorable by the {@link CriticalWorkerWatchdog}.
+ * Must be registered with the watchdog explicitly.
+ */
+public class CriticalSingleThreadExecutor extends ThreadPoolExecutor implements CriticalWorker {
+
+    private Thread lastSeenThread;
+    private volatile long heartbeatNanos = NOT_MONITORED;
+
+    public CriticalSingleThreadExecutor(ThreadFactory threadFactory) {
+        super(1, 1, 0, TimeUnit.SECONDS, new LinkedBlockingQueue<>(), threadFactory);
+    }
+
+    @Override
+    protected void beforeExecute(Thread t, Runnable r) {
+        try {
+            lastSeenThread = t;
+            heartbeatNanos = System.nanoTime();
+        } finally {
+            super.beforeExecute(t, r);
+        }
+    }
+
+    @Override
+    protected void afterExecute(Runnable r, Throwable t) {
+        try {
+            super.afterExecute(r, t);
+        } finally {
+            heartbeatNanos = NOT_MONITORED;
+        }
+    }
+
+    @Override
+    public long threadId() {
+        Thread thread = lastSeenThread;
+
+        assert thread != null;
+
+        return thread.getId();
+    }
+
+    @Override
+    public long heartbeatNanos() {
+        return heartbeatNanos;
+    }
+}
--- a/modules/core/src/main/java/org/apache/ignite/internal/worker/CriticalWorkerWatchdog.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/worker/CriticalWorkerWatchdog.java
@@ -1,0 +1,114 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.worker;
+
+import static org.apache.ignite.internal.util.CompletableFutures.nullCompletedFuture;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadInfo;
+import java.lang.management.ThreadMXBean;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import org.apache.ignite.internal.logger.IgniteLogger;
+import org.apache.ignite.internal.logger.Loggers;
+import org.apache.ignite.internal.manager.IgniteComponent;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A watchdog that monitors liveness of the registered workers and, if a worker is found to be hung, logs the corresopnding
+ * information (including the stack trace corresponding to the woeker's thread) and then calls the failure handler
+ * to fail the node.
+ *
+ * <p>Each worker is expected to maintain its {@link CriticalWorker#heartbeatNanos()} growing while the worker executes some computations.
+ * If the worker does not do any computations (it is blocked on an I/O operation, waits for a lock, or has no work to do),
+ * it must set its {@link CriticalWorker#heartbeatNanos()} to {@link CriticalWorker#NOT_MONITORED}.
+ *
+ * <p>The watchdog periodically does its check; if it finds a worker that lags more than allowed and it is not in
+ * NOT_MONITORED state, then logging and failure handling notification is triggered.
+ */
+public class CriticalWorkerWatchdog implements CriticalWorkerRegistry, IgniteComponent {
+    private static final IgniteLogger LOG = Loggers.forClass(CriticalWorkerWatchdog.class);
+
+    private static final long LIVENESS_CHECK_INTERVAL_MS = 500;
+    private static final long MAX_ALLOWED_LAG_MS = 500;
+
+    private final ScheduledExecutorService scheduler;
+
+    private final Set<CriticalWorker> registeredWorkers = ConcurrentHashMap.newKeySet();
+
+    @Nullable
+    private volatile ScheduledFuture<?> livenessProbeTaskFuture;
+
+    public CriticalWorkerWatchdog(ScheduledExecutorService scheduler) {
+        this.scheduler = scheduler;
+    }
+
+    @Override
+    public void register(CriticalWorker worker) {
+        registeredWorkers.add(worker);
+    }
+
+    @Override
+    public void unregister(CriticalWorker worker) {
+        registeredWorkers.remove(worker);
+    }
+
+    @Override
+    public CompletableFuture<Void> start() {
+        livenessProbeTaskFuture = scheduler.scheduleAtFixedRate(
+                this::probeLiveness,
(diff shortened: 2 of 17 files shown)


modules/network/src/main/java/org/apache/ignite/network/DefaultMessagingService.java
```diff
@@ -110,18 +115,20 @@ public DefaultMessagingService(
...
     ) {
         this.factory = factory;
         this.topologyService = topologyService;
         this.staleIdDetector = staleIdDetector;
         this.classDescriptorRegistry = classDescriptorRegistry;
         this.marshaller = marshaller;
+        this.criticalWorkerRegistry = criticalWorkerRegistry;
 
-        this.outboundExecutor = Executors.newSingleThreadExecutor(NamedThreadFactory.create(nodeName, "MessagingService-outbound-", LOG));
+        this.outboundExecutor = new CriticalSingleThreadExecutor(NamedThreadFactory.create(nodeName, "MessagingService-outbound", LOG));
```
sashapolo, line 128: Why did you remove the "-" at the end?

modules/network/src/main/java/org/apache/ignite/network/DefaultMessagingService.java
```diff
@@ -531,4 +547,36 @@ public void stopDroppingMessages() {
...
+
+            return executor;
+        }
+
+        @Override
+        protected void onStoppingInitiated() {
+            super.onStoppingInitiated();
+
+            for (CriticalWorker worker : registeredWorkers) {
+                workerRegistry.unregister(worker);
```
sashapolo, line 578: Maybe we should have a method like `close` or `unregisterAll` on the registry instead?

modules/network/src/main/java/org/apache/ignite/network/NettyWorkersRegistrar.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.ignite.internal.worker.CriticalWorkerRegistry;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * This component is responsible for registering Netty workers with the {@link CriticalWorkerRegistry} and for updating their heartbeats.
+ */
+public class NettyWorkersRegistrar implements IgniteComponent {
+    /*
+     * It seems impossible to instrument tasks executed by Netty event loops, so the strategy we use is to
+     * send 'heartbeat' tasks to each of the event loops periodically; this tasks just update the heartbeat timestamp
```
sashapolo, line 44:
```suggestion
     * send 'heartbeat' tasks to each of the event loops periodically; these tasks just update the heartbeat timestamp
```

modules/network/src/main/java/org/apache/ignite/network/NettyWorkersRegistrar.java
```diff
@@ -0,0 +1,147 @@
...
+     * of the worker corresponding to an event loop. If an event loop's thread is hung, the hearbeat will not
```
sashapolo, line 45:
```suggestion
     * of the worker corresponding to an event loop. If an event loop's thread is blocked, the hearbeat will not
```

modules/network/src/main/java/org/apache/ignite/network/NettyWorkersRegistrar.java
```diff
@@ -0,0 +1,147 @@
...
+        this.bootstrapFactory = bootstrapFactory;
+    }
+
+    @Override
+    public CompletableFuture<Void> start() {
+        for (NioEventLoopGroup group : bootstrapFactory.eventLoopGroups()) {
+            registerWorkersFor(group);
+        }
+
+        sendHearbeatsTaskFuture = scheduler.scheduleAtFixedRate(this::sendHearbeats, 100, 100, MILLISECONDS);
```
sashapolo, line 84: Should we introduce some constants?

modules/network/src/main/java/org/apache/ignite/network/NettyWorkersRegistrar.java
```diff
@@ -0,0 +1,147 @@
...
+            criticalWorkerRegistry.unregister(worker);
+        }
+    }
+
+    private static class NettyWorker implements CriticalWorker {
+        private final SingleThreadEventExecutor eventExecutor;
+
+        private volatile long heartbeatNanos = System.nanoTime();
+
+        private final Runnable sendHeartbeatTask = () -> heartbeatNanos = System.nanoTime();
```
sashapolo, line 127: Can it simply be a method?

modules/network/src/main/java/org/apache/ignite/network/NettyWorkersRegistrar.java
```diff
@@ -0,0 +1,147 @@
...
+     * be updated, and the worker watchdog will treat the event loop worker as blocked.
+     */
+
+    private final CriticalWorkerRegistry criticalWorkerRegistry;
+
+    private final ScheduledExecutorService scheduler;
+
+    private final NettyBootstrapFactory bootstrapFactory;
+
+    private final List<NettyWorker> workers = new CopyOnWriteArrayList<>();
```
sashapolo, line 55: Looks like this thing can simply be a volatile immutable list

sashapolo (commented):
I don't see any tests, why?

modules/core/src/main/java/org/apache/ignite/internal/worker/CriticalSingleThreadExecutor.java
```diff
@@ -0,0 +1,70 @@
...
+    private Thread lastSeenThread;
+    private volatile long heartbeatNanos = NOT_MONITORED;
+
+    public CriticalSingleThreadExecutor(ThreadFactory threadFactory) {
+        super(1, 1, 0, TimeUnit.SECONDS, new LinkedBlockingQueue<>(), threadFactory);
+    }
+
+    @Override
+    protected void beforeExecute(Thread t, Runnable r) {
+        try {
```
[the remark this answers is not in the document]
rpuch: To make sure that `super.beforeExecute()` gets called

modules/core/src/main/java/org/apache/ignite/internal/worker/CriticalSingleThreadExecutor.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Single thread executor instrumented to be used as a {@link CriticalWorker} and being monitorable by the {@link CriticalWorkerWatchdog}.
+ * Must be registered with the watchdog explicitly.
+ */
+public class CriticalSingleThreadExecutor extends ThreadPoolExecutor implements CriticalWorker {
+
+    private Thread lastSeenThread;
```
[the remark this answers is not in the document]
rpuch: We write to this field before writing to volatile`heartbeatNanos`, and we read from `heartbeatNanos` before reading from `lastSeenThread`, so happens-before is guaranteed. It's a bit fragile because we have to rely on the watchdog doing reads in this order.

I made this field non-volatile to try reducing the number of volatile writes on each job execution (from 2 to 1).

Do you think it's not worth it?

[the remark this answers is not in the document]
sashapolo: I really don't like the fact that we rely on the external party to always call `heartbeatNanos` before `threadId`. I prefer such happens-before relationships to be encapsulated or at least explicitly specified in the javadoc.

modules/core/src/main/java/org/apache/ignite/internal/worker/CriticalWorkerWatchdog.java
```diff
@@ -0,0 +1,121 @@
...
+
+            if (heartbeatNanos == CriticalWorker.NOT_MONITORED) {
+                continue;
+            }
+
+            long delayMillis = TimeUnit.NANOSECONDS.toMillis(nowNanos - heartbeatNanos);
+            if (delayMillis > MAX_ALLOWED_LAG_MS) {
+                ThreadMXBean bean = ManagementFactory.getThreadMXBean();
+
+                ThreadInfo[] threadInfos = bean.getThreadInfo(new long[]{worker.threadId()}, true, true, Integer.MAX_VALUE);
```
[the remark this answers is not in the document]
rpuch: Is this about optimization? Finding a blocked thread will lead to a node being halted, so do we need to optimize this so hard?

modules/network/src/main/java/org/apache/ignite/network/DefaultMessagingService.java
rpuch: Because `NamedThreadFactory#threadPrefix()` adds the dash itself

modules/network/src/main/java/org/apache/ignite/network/DefaultMessagingService.java
rpuch: Components that register their workers stop before the registry is stopped. So, if they don't de-register their workers, the registry might think it still should monitor them, and we might get a false positive detection: a worker is not updated anymore

modules/network/src/main/java/org/apache/ignite/network/DefaultMessagingService.java
sashapolo: Oh, looks like I missed that

modules/network/src/main/java/org/apache/ignite/network/NettyWorkersRegistrar.java
rpuch: The idea was to spare an allocation of a lambda/method reference

modules/network/src/main/java/org/apache/ignite/network/DefaultMessagingService.java
sashapolo: My point is: it is inconvenient that all components need to maintain the list of their workers themselves. Maybe we can alter the inner structure of the Registry so that it would have something like a `Map<Class, List>` so that we can de-register a bunch of workers using a component's class?

modules/network/src/main/java/org/apache/ignite/network/NettyWorkersRegistrar.java
sashapolo: Is there a difference if you allocate it here or once inside `sendHeartbeat`?

modules/core/src/main/java/org/apache/ignite/internal/worker/CriticalSingleThreadExecutor.java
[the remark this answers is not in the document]
sashapolo: I understand that, just looks redundant here

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/worker/ItCriticalWorkerMonitoringTest.java
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.ignite.internal.table.distributed.raft.snapshot.message.SnapshotMetaResponse;
+import org.apache.ignite.internal.testframework.log4j2.LogInspector;
+import org.apache.ignite.internal.worker.CriticalWorkerWatchdog;
+import org.apache.ignite.raft.jraft.RaftMessagesFactory;
+import org.apache.logging.log4j.core.LogEvent;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+
+@SuppressWarnings("resource")
+class ItCriticalWorkerMonitoringTest extends ClusterPerTestIntegrationTest {
```
sashapolo, line 38: Do we really need a full blown integration test here? My idea was that we can write something lightweight and check that the Registry works.

rpuch: I think that in this case yes, we need a full integration test, as this functionality is very important to have it working, it can be easily broken and it may be broken in a few places. It's scary to just replicate 'the way IgniteImpl doing this' *in vitro*.

modules/network/src/main/java/org/apache/ignite/network/NettyWorkersRegistrar.java
rpuch: I suppose it will be allocated on each `sendHeartbeat()` call

modules/network/src/main/java/org/apache/ignite/network/DefaultMessagingService.java
rpuch: And if a component has just one worker, it will still need to register it with that grouping passing an additional parameter; or we'll need to have 2 `register()` methods, one with one parameter, another with 2 parameters.

Right now we have just 2 kinds of critical workers: Netty threads (that we must keep track of anyway because we need to send heartbeats to them, so your proposal does not work for them) and inbound/outbound threads (for which your proposal might be useful). So 'all components' *for now* boils down to just 1 component.

I think we should see what happens further. If more components are added, and most of them benefit from such batch deregistration, we'll introduce it. For now it looks like simplifying in one place while complicating in other 2 places (the watchdog and the Netty workers registrar). WDYT?

modules/core/src/main/java/org/apache/ignite/internal/worker/CriticalWorkerWatchdog.java
```diff
@@ -86,8 +90,28 @@ public CompletableFuture<Void> start() {
...
+
+                // TODO: IGNITE-16899 - invoke failure handler.
+            }
+        }
+    }
+
+    private Long2LongMap getDelayedThreadIdsAndDelays() {
         long nowNanos = System.nanoTime();
 
+        Long2LongMap delayedThreadIdsToDelays = new Long2LongOpenHashMap();
```
sashapolo, line 113: I would suggest to use lazy initialization here to avoid spamming unnecessary objects

sashapolo (approved):

modules/network/src/main/java/org/apache/ignite/network/NettyWorkersRegistrar.java
rpuch: Having read on this, I'm not sure now. But I remember that a couple of times this (cache/not cache instances) my intuition betrayed me, so I still like the most obvious approach: no doubts can arise about whether it allocates a new instance each time or not :) We lose lazyness, but we already know that this will be called more than 0 times, so it's not a problem.

PR #3077, opened 2024-01-22
IGNITE-21267 Implement async ClientHandlerModule startup

ptupitsyn:


modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerModule.java
```diff
@@ -319,32 +313,35 @@ protected void initChannel(Channel ch) {
...
-
-        if (LOG.isInfoEnabled()) {
-            LOG.info("Thin client protocol started successfully [port={}]", port);
-        }
 
-        return ch.closeFuture();
+        bootstrap.bind(port).addListener((ChannelFutureListener) bindFut -> {
+            if (bindFut.isSuccess()) {
+                if (LOG.isInfoEnabled()) {
+                    LOG.info("Thin client protocol started successfully [port={}]", port);
```
isapego, line 320: Sounds a little weird, does not it?

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerModule.java
```diff
@@ -319,32 +313,35 @@ protected void initChannel(Channel ch) {
...
+
+                result.complete(bindFut.channel());
+                return;
+            }
+
+            if (bindFut.cause() instanceof BindException) {
+                result.completeExceptionally(
+                        new IgniteException(
+                                PORT_IN_USE_ERR,
+                                "Cannot start thin client connector endpoint. Port " + port + " is in use.",
```
isapego, line 331: Here we use `connector endpoint`, not `protocol`, which sounds better to me.

isapego (approved):

modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientHandlerModule.java
ptupitsyn: Agree, fixed to be consistent with other messages below.

PR #3079, opened 2024-01-23
IGNITE-21158 ODBC: Column metadata for the non-executed query

isapego:
- Re-factored server side. Now a single request returns metadata for both parameter and result set meta;
- Re-factored client side - moved parameter metadata to `data_query` class, as this is the only type of query that can have this metadata;
- Implemented on client side;
- Un-commented related tests and added some new;
- Fixed Windows compilation in C++ project.

### DIFF
--- a/modules/platforms/cpp/ignite/odbc/query/data_query.cpp
+++ b/modules/platforms/cpp/ignite/odbc/query/data_query.cpp
@@ -122,13 +122,20 @@ sql_result data_query::execute() {
 
 const column_meta_vector *data_query::get_meta() {
     if (!m_result_meta_available) {
-        make_request_resultset_meta();
+        update_meta();
 
         if (!m_result_meta_available)
             return nullptr;
     }
 
     return &m_result_meta;
+}
+
+const sql_parameter *data_query::get_sql_param(std::int16_t idx) {
+    if (idx > 0 && static_cast<std::size_t>(idx) <= m_params_meta.size())
+        return &m_params_meta.at(idx - 1);
+
+    return nullptr;
 }
 
 sql_result data_query::fetch_next_row(column_binding_map &column_bindings) {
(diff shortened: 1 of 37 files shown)


modules/platforms/cpp/ignite/odbc/query/data_query.cpp
```diff
@@ -454,16 +441,29 @@ sql_result data_query::process_conversion_result(
...
 
     for (size_t i = 0; i < m_result_meta.size(); ++i) {
         column_meta &meta = m_result_meta.at(i);
-        LOG_MSG("\n[" << i << "] SchemaName:     " << meta.get_schema_name() << "\n[" << i << "] TypeName:       "
-                      << meta.get_table_name() << "\n[" << i << "] ColumnName:     " << meta.get_column_name() << "\n["
-                      << i << "] ColumnType:     " << static_cast<int32_t>(meta.get_data_type()));
+        LOG_MSG("[" << i << "] SchemaName: " << meta.get_schema_name());
+        LOG_MSG("[" << i << "] TableName:  " << meta.get_table_name());
+        LOG_MSG("[" << i << "] ColumnName: " << meta.get_column_name());
+        LOG_MSG("[" << i << "] ColumnType: " << static_cast<int32_t>(meta.get_data_type()));
```
ptupitsyn, line 453: Does this logging have any performance implications? `LOG_MSG` does not seem to have any checks like `TRACE_MSG` has (`is_trace_enabled`).

ptupitsyn (approved):

modules/platforms/cpp/ignite/odbc/query/data_query.cpp
isapego: Logging in ODBC only enabled in debug builds, it's removed altogether in release.

PR #3094, opened 2024-01-25
IGNITE-21252 start clients for stable + pending when rebalancing

alievmirza:

### The Review Checklist

### Notes


modules/index/src/integrationTest/java/org/apache/ignite/internal/index/ItBuildIndexTest.java
```diff
@@ -339,10 +339,10 @@ private static Map<Integer, List<Ignite>> waitForIndexBuild(String tableName, St
                 for (int partitionId = 0; partitionId < internalTable.partitions(); partitionId++) {
                     RaftGroupService raftGroupService = internalTable.partitionRaftGroupService(partitionId);
 
-                    Stream<Peer> allPeers = Stream.concat(Stream.of(raftGroupService.leader()), raftGroupService.peers().stream());
```
sanpwc, line 342: Do you know why originally there was a leader in concatenation?

modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
```diff
@@ -381,6 +408,27 @@ private <T extends RaftGroupService> CompletableFuture<T> startRaftGroupNodeInte
                 : raftServiceFactory.startRaftGroupService(nodeId.groupId(), configuration, raftConfiguration, executor, cmdMarshaller);
     }
 
+    private void startRaftGroupNodeInternalWithoutService(
```
sanpwc, line 411: Should we call this method inside `startRaftGroupNodeInternal` in order to remove code duplication or they are not the same?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -804,17 +807,16 @@ private CompletableFuture<Void> startPartitionAndStartClient(
 
         var raftNodeId = localMemberAssignment == null ? null : new RaftNodeId(replicaGrpId, serverPeer);
 
-        boolean shouldStartRaftListeners = localMemberAssignment != null && !((Loza) raftMgr).isStarted(raftNodeId);
+        boolean shouldStartPartitionTrackers = localMemberAssignment != null && !((Loza) raftMgr).isStarted(raftNodeId);
```
sanpwc, line 810: Who is partition tracker?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -867,6 +869,8 @@ private CompletableFuture<Void> startPartitionAndStartClient(
                     }
                 }), ioExecutor)
                 .thenAcceptAsync(updatedRaftGroupService -> inBusyLock(busyLock, () -> {
+                    updatedRaftGroupService.updateConfiguration(configurationFromAssignments(union(newPartAssignment, oldPartAssignment)));
```
sanpwc, line 872: So, you claim that we will finish here both in case of node recovery on restart and common pending assignments update handling?

modules/index/src/integrationTest/java/org/apache/ignite/internal/index/ItBuildIndexTest.java
alievmirza: No, but leader is also presented in the peers' list, so seems like this code were inefficient.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
alievmirza: I realised that the better naming is `shouldUpdatePartitionTrackers`, will update. I decided to rename it because originally the name `shouldStartRaftListeners` had no meaning  for me because I had the same question about "Raft Listeners"

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
alievmirza: I've changed nothing in this code regarding recovery. As far as I can see from the code, for the persistence storage `shouldStartGroupFut` will return true, so the flow will be the same, as for common pending assignments update handling

modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
alievmirza: Yes, thanks, will fix

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sanpwc: PartitionListerner aka RaftListener is a common one, PartitionTracker is not.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
sanpwc: I mean that we should union pending and stable both for recovery and common events processing. Please check whether it's true.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
alievmirza: reverted

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
alievmirza: I've checked, we start pending and stable clients for recovery as well

modules/table/src/integrationTest/java/org/apache/ignite/internal/rebalance/ItRebalanceDistributedTest.java
```diff
@@ -660,6 +663,63 @@ void testRaftClientsUpdatesAfterRebalance() throws Exception {
...
+                                .internalTable()
+                                .partitionRaftGroupService(0)
+                                .peers()
+                                .stream()
+                                .collect(toSet())
+                                .equals(Set.of(new Peer(newNodeNameForAssignment), new Peer(assignmentsBeforeRebalance)))),
+                (long) AWAIT_TIMEOUT_MILLIS * nodes.size()
+        ));
+
+        stopDropping.set(false);
```
sanpwc, line 720: And what about verification that will check that there are only stable assignments in client.peers() on rebalance finished?

alievmirza: well, I could add it, but in general, such logic is checked in `ItRebalanceDistributedTest#testRaftClientsUpdatesAfterRebalance`

sanpwc: No need in that case.

PR #3101, opened 2024-01-26
IGNITE-20881 Add ability to enforce an index to be used

nva:

### The Review Checklist

### Notes


modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
```diff
@@ -216,8 +230,13 @@ public SqlSerializer build() {
             }
 
             SqlSerializer ser = new SqlSerializer()
-                    .append("SELECT * ")
-                    .append("FROM ").append(quoteIfNeeded(tableName));
+                    .append("SELECT");
+
+            if (!nullOrBlank(indexName)) {
+                ser.append(" /*+ FORCE_INDEX(").append(indexName).append(") */");
```
AMashenkov, line 236: This approach allow SQL injections.

AMashenkov (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
nva: The index name is also not something an app would be getting from the UI, so there is no way for a user to inject using that vector, but to be bulletproof added forbid characters except [a-zA-Z0-9_] in the index names passed via the Critetia API.

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
```diff
@@ -232,5 +255,13 @@ public SqlSerializer build() {
 
             return ser;
         }
+
+        private static String normalizeIndexName(String name) {
+            if (!CANONICAL_NAME.matcher(name).matches()) {
+                throw new IllegalArgumentException("Index name must be alphanumeric with underscore and start with letter. Was: " + name);
+            }
```
AMashenkov, line 262: Let's 
1. split into validation and normalization functions.
2. move regex pattern and general validation method to some Utility class
3. the exception should be thrown here, in the current class, because here we have a context that we validate an index name.

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
```diff
@@ -161,12 +163,17 @@ private void append(String delimiter, String prefix, String suffix, Criteria[] e
      * Builder.
      */
     public static class Builder  {
+        private static final Pattern CANONICAL_NAME = Pattern.compile("^(?:\\p{Alpha}[\\w_]*\\.)?\\p{Alpha}[\\w_]*$");
```
AMashenkov, line 166: Seems, additional underscore is not needed.
> \w	A word character: [a-zA-Z_0-9]

Also, in case of simple name, moving the dot into a second group allows regexp matcher to not discard the first group.
```suggestion
        private static final Pattern CANONICAL_NAME = Pattern.compile("^(?:\\p{Alpha}[\\w]*)(?:\\.\\p{Alpha}[\\w]*)?$");
```

nva: Yes, you are right

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/SqlSerializer.java
nva: Done

nva: Another good candidate for such refactoring is `SystemView#normalizeIdentifier`

modules/api/src/main/java/org/apache/ignite/table/criteria/CriteriaQuerySource.java
```diff
@@ -38,19 +38,33 @@ public interface CriteriaQuerySource<T> {
...
      *
      * @param tx Transaction to execute the query within or {@code null} to run within implicit transaction.
      * @param criteria The predicate to filter entries or {@code null} to return all entries from the underlying table.
+     * @param indexName The name of the index to force usage of this index in the query.
      * @param opts Criteria query options or {@code null} to use default.
      * @return Iterator with query results.
      * @throws CriteriaException If failed.
      */
-    Cursor<T> query(@Nullable Transaction tx, @Nullable Criteria criteria, @Nullable CriteriaQueryOptions opts);
+    Cursor<T> query(@Nullable Transaction tx, @Nullable Criteria criteria, @Nullable String indexName, @Nullable CriteriaQueryOptions opts);
```
AMashenkov, line 67: All method parameters are optional.
Does the next call make sense?
"query(null, null, null, null)" ?

AMashenkov (approved):

modules/api/src/main/java/org/apache/ignite/table/criteria/CriteriaQuerySource.java
nva: Yes, this query will be run within implicit transaction and return all entries from the underlying table.

PR #3104, opened 2024-01-26
IGNITE-21250 Client compute should pass all candidate nodes

valepakh:
https://issues.apache.org/jira/browse/IGNITE-21250

We need to pass all of the candidate nodes to the client handler so that the worker node failover works


modules/compute/src/integrationTest/java/org/apache/ignite/internal/compute/ItEmbeddedWorkerShutdownTest.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.compute;
+
+import org.apache.ignite.compute.IgniteCompute;
+import org.apache.ignite.internal.app.IgniteImpl;
+
+class ItEmbeddedWorkerShutdownTest extends ItWorkerShutdownTest {
+
```
ptupitsyn, line 24: Redundant blank like

modules/compute/src/integrationTest/java/org/apache/ignite/internal/compute/ItThinClientWorkerShutdownTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @AfterEach
+    void cleanup() throws Exception {
+        client.close();
+    }
+
+    @Override
+    IgniteCompute compute(IgniteImpl entryNode) {
+        String address = "127.0.0.1:" + entryNode.clientAddress().port();
+        client = IgniteClient.builder().addresses(address).build();
```
ptupitsyn, line 37: 1. A new client is started on every call to this method
2. The clients are never closed, causing extra resource consumption and potential issues in other tests

Let's start one client before the test, save it to a field, and close after the test.

modules/client/src/main/java/org/apache/ignite/internal/client/compute/ClientCompute.java
```diff
@@ -271,16 +274,12 @@ private CompletableFuture<PayloadInputChannel> executeOnOneNode(
...
                 ClientOp.COMPUTE_EXECUTE,
                 w -> {
-                    if (w.clientChannel().protocolContext().clusterNode().name().equals(node.name())) {
-                        w.out().packNil();
-                    } else {
-                        w.out().packString(node.name());
-                    }
-
+                    String[] nodeNames = nodes.stream().map(ClusterNode::name).toArray(String[]::new);
+                    w.out().packObjectArrayAsBinaryTuple(nodeNames);
```
ptupitsyn, line 278: 1. Java streams are unnecessary here (and inefficient)
2. `packObjectArrayAsBinaryTuple` is excessive here, it includes type code for every element

Instead, let's do `packInt` for size and then `packString` in a loop.

modules/client/src/main/java/org/apache/ignite/internal/client/compute/ClientCompute.java
```diff
@@ -261,8 +263,9 @@ public <R> Map<ClusterNode, JobExecution<R>> broadcastAsync(
         return map;
     }
 
-    private CompletableFuture<PayloadInputChannel> executeOnOneNode(
-            ClusterNode node,
+    private CompletableFuture<PayloadInputChannel> executeOnNodesAsync(
+            ClusterNode coordinatorNode,
```
ptupitsyn, line 267: `Coordinator` does not make sense.
* For `execute` it is not needed, we can use any node from the set (and we should pick one of the nodes where a direct client connection exists, if possible)
* For `broadcast` we only have one node in the set anyway

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/compute/ClientComputeExecuteRequest.java
```diff
@@ -50,29 +56,82 @@ public class ClientComputeExecuteRequest {
...
+            ClusterNode node = cluster.topologyService().getByConsistentId(id);
+            if (node == null) {
+                notFoundNodes.add(id);
+            } else {
+                candidates.add(node);
+            }
+        });
+
+        if (!notFoundNodes.isEmpty()) {
+            throw new IgniteException("Specified nodes are not present in the cluster: " + notFoundNodes);
```
ptupitsyn, line 97: Deprecated constructor

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/compute/ClientComputeExecuteRequest.java
```diff
@@ -50,29 +56,82 @@ public class ClientComputeExecuteRequest {
...
+        }
+        return candidates;
```
ptupitsyn, line 99:
```suggestion
    int size = in.unpackInt();
    List<ClusterNode> nodes = new ArrayList<>(size);
    for (int i = 0; i < size; i++) {
        String nodeName = in.unpackString();
        ClusterNode node = cluster.topologyService().getByConsistentId(nodeName);
        
        if (node == null) {
            // Throw on first issue
        }
        
        nodes.Add(node);
    }
```

Keep it simple.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/compute/ClientComputeExecuteRequest.java
```diff
@@ -50,29 +56,82 @@ public class ClientComputeExecuteRequest {
...
+    }
+
+    /**
+     * Selects a random node from the set of candidates, preferably not a local node.
```
ptupitsyn, line 103: Why do we prefer a non-local node?

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/compute/ClientComputeExecuteRequest.java
```diff
@@ -50,29 +56,82 @@ public class ClientComputeExecuteRequest {
...
+     *
+     * @param candidates Set of candidate nodes.
+     * @param localNode Local node.
+     *
+     * @return Target node to run a job on.
+     */
+    private static ClusterNode selectTargetNode(Set<ClusterNode> candidates, ClusterNode localNode) {
```
ptupitsyn, line 110: I don't think that node selection logic should be in the client request handler. Can we move this logic to the Compute component?

modules/compute/src/main/java/org/apache/ignite/internal/compute/IgniteComputeInternal.java
```diff
@@ -17,16 +17,45 @@
...
+     * @param candidates Candidate nodes; In case target node left the cluster, the job will be restarted on one of them.
+     * @param units Deployment units. Can be empty.
+     * @param jobClassName Name of the job class to execute.
+     * @param options Job execution options.
+     * @param args Arguments of the job.
+     * @return CompletableFuture Job result.
+     */
+    <R> JobExecution<R> executeAsyncWithFailover(
+            ClusterNode targetNode,
+            Set<ClusterNode> candidates,
```
ptupitsyn, line 52: Do we really need to pass instances of `Set` around in the internal API? `List` is more efficient and easier.

ptupitsyn (changes requested):

ptupitsyn:
@valepakh I'll help with the .NET part once we finalize the protocol changes.

modules/compute/src/integrationTest/java/org/apache/ignite/internal/compute/ItThinClientWorkerShutdownTest.java
ptupitsyn: We can have a map from address to client, and clean it up after the test. But we can't leave the clients running in the background.

modules/client/src/main/java/org/apache/ignite/internal/client/compute/ClientCompute.java
ptupitsyn: The best approach would be to pass only a collection of nodes to this method, then pick `preferredNode` for the `serviceAsync` according to the existing connections, or `null` if there is no intersection of passed nodes and connections.

modules/compute/src/main/java/org/apache/ignite/internal/compute/IgniteComputeInternal.java
ptupitsyn: Ok.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/compute/ClientComputeExecuteRequest.java
ptupitsyn: Sorry, I still don't understand. All nodes have a similar chance of failing, how is the local one different?

modules/client/src/main/java/org/apache/ignite/internal/client/compute/ClientCompute.java
ptupitsyn: > Do you think it's worth randomizing the selection
Yes, for load balancing.

modules/api/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -203,6 +203,9 @@ public static class Client {
 
         /** Client handshake header error. */
         public static final int HANDSHAKE_HEADER_ERR = CLIENT_ERR_GROUP.registerErrorCode((short) 10);
+
+        /** Specified node is not found in the cluster. */
+        public static final int NODE_NOT_FOUND_ERR = CLIENT_ERR_GROUP.registerErrorCode((short) 11);
```
ptupitsyn, line 208: This error is not specific to the client. The same issue can occur when using the embedded API, right?

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/compute/ClientComputeExecuteRequest.java
```diff
@@ -50,29 +53,39 @@ public class ClientComputeExecuteRequest {
...
 
-        JobExecution<Object> execution = compute.executeAsync(Set.of(node), deploymentUnits, jobClassName, options, args);
+        JobExecution<Object> execution = compute.executeAsyncWithFailover(candidates, deploymentUnits, jobClassName, options, args);
         sendResultAndStatus(execution, notificationSender);
         return execution.idAsync().thenAccept(out::packUuid);
     }
 
+    private static Set<ClusterNode> unpackCandidateNodes(ClientMessageUnpacker in, ClusterService cluster) {
+        int size = in.unpackInt();
+        Set<ClusterNode> nodes = new HashSet<>();
```
ptupitsyn, line 74:
```suggestion
        Set<ClusterNode> nodes = new HashSet<>(size);
        
```

modules/client/src/main/java/org/apache/ignite/internal/client/compute/ClientCompute.java
```diff
@@ -271,28 +274,36 @@ private CompletableFuture<PayloadInputChannel> executeOnOneNode(
...
+        // Select the preferred node from the intersection of the candidate nodes and existing connections
+        preferredNodeNames.retainAll(connections);
+
+        if (preferredNodeNames.isEmpty()) {
+            return null;
+        }
+        return randomNode(preferredNodeNames);
+    }
+
+    private static String randomNode(Set<String> nodes) {
```
ptupitsyn, line 299:
```suggestion
    private @Nullable String selectPreferredNodeName(Set<ClusterNode> nodes) {
        List<String> candidateNodeNames = new ArrayList<>(nodes.size());
        Set<String> connectedNodeNames = ch.connectedNodeNames();

        for (ClusterNode node : nodes) {
            if (connectedNodeNames.contains(node.name())) {
                candidateNodeNames.add(node.name());
            }
        }

        if (candidateNodeNames.isEmpty()) {
            return null;
        }

        if (candidateNodeNames.size() == 1) {
            return candidateNodeNames.get(0);
        }

        int randomIdx = ThreadLocalRandom.current().nextInt(candidateNodeNames.size());
        return candidateNodeNames.get(randomIdx);
    }
```

* Avoid streams
* Initialize collections with capacity


`ReliableChannel.connectedNodeNames`:
```java
    /**
     * Gets connected node names.
     *
     * @return Set of connected node names.
     */
    public Set<String> connectedNodeNames() {
        Set<String> res = new HashSet<>(channels.size());

        for (var holder : nodeChannelsByName.values()) {
            var chFut = holder.chFut;

            if (chFut != null) {
                var ch = ClientFutureUtils.getNowSafe(chFut);

                if (ch != null && !ch.closed()) {
                    res.add(ch.protocolContext().clusterNode().name());
                }
            }
        }

        return res;
    }
```

modules/api/src/main/java/org/apache/ignite/lang/ErrorGroups.java
ptupitsyn: The API (interface) is the same for client and embedded. Exceptions and error codes must be the same.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/compute/ClientComputeExecuteRequest.java
ptupitsyn: * It is faster to run the job on the current node (no need for a network hop)
* Client logic specifically tries to send the job to the target node directly and avoid a network hop, but this logic does the opposite and tries to add a network hop back - does not make sense
* Node failure is relatively rare
* Client has retry mechanism to deal with failures

Let's invert the logic and invoke locally when possible.

modules/api/src/main/java/org/apache/ignite/lang/ErrorGroups.java
ptupitsyn: Ok, let's handle this separately: https://issues.apache.org/jira/browse/IGNITE-21390 (I've checked what happens in embedded mode, and it does not look good at all).

Here let's revert related changes (remove error code, keep existing exception with deprecated ctor).

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/compute/ClientComputeExecuteRequest.java
ptupitsyn: @Pochatkin Sending the job to a different node does not affect the chance of coordinator failure. 

If the coordinator fails, the job will fail for the client anyway (due to connection loss). And the client will retry it.

ptupitsyn: In other words:
* Let's say node A is _destined_ to fail very soon
* Client sends the job to the node A. 

Variant 1 (your logic): 
* Node A sends the job to node B
* Node A fails, client sees broken connection, and considers the job failed

Variant 2:
* Node A executes the job locally
* Node A fails, client sees broken connection, and considers the job failed

The outcome is identical.

ptupitsyn: Agree with this, client retry is different and can even be disabled. 

Still, using a remote node instead of a local one for job execution does not make sense. It degrades performance and does not improve anything.


I think we should design client-side compute job failover. The design goals can't be achieved with "coordinator node" approach, because coordinator can fail, or the connection to coordinator can be briefly lost (and re-established). The client should be the coordinator.

ptupitsyn: @Pochatkin please read my previous replies. The single point of failure is not eliminated by sending the job to another node.

modules/client/src/main/java/org/apache/ignite/internal/client/compute/ClientCompute.java
```diff
@@ -271,35 +269,15 @@ private CompletableFuture<PayloadInputChannel> executeOnOneNode(
...
-                    } else {
-                        w.out().packString(node.name());
-                    }
-
+                    packNodeNames(w.out(), nodes);
                     packJob(w.out(), units, jobClassName, options, args);
                 },
                 ch -> ch,
-                node.name(),
+                null,
```
ptupitsyn, line 276: Don't pass null as `preferredNode`. Let's pass random node from `nodes` to increase the chance of local execution.

This is especially important when `nodes` has only one node and a direct connection exists - guaranteed single network call.

modules/platforms/dotnet/Apache.Ignite/Internal/Compute/Compute.cs
```diff
@@ -69,7 +69,7 @@ public Compute(ClientFailoverSocket socket, Tables tables)
             IgniteArgumentCheck.NotNull(nodes);
             IgniteArgumentCheck.NotNull(jobClassName);
 
-            return await ExecuteOnOneNode<T>(GetRandomNode(nodes), units, jobClassName, args).ConfigureAwait(false);
+            return await ExecuteOnNodes<T>(GetRandomNode(nodes), nodes, units, jobClassName, args).ConfigureAwait(false);
```
ptupitsyn, line 72: [Multiple enumeration of IEnumerable](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1851).

Let's do the following:
* Change `GetRandomNode` and `ExecuteOnNodes` to require `ICollection<IClusterNode>`
* Convert `IEnumerable<IClusterNode>` to `ICollection<IClusterNode>` with `GetNodesCollection` once on the previous line

modules/platforms/dotnet/Apache.Ignite/Internal/Compute/Compute.cs
```diff
@@ -171,19 +167,41 @@ private static void WriteUnits(IEnumerable<DeploymentUnit> units, PooledArrayBuf
...
             }
 
             countSpan[0] = MsgPackCode.Array32;
             BinaryPrimitives.WriteInt32BigEndian(countSpan[1..], count);
         }
 
-        private async Task<T> ExecuteOnOneNode<T>(
+        private static void WriteUnits(IEnumerable<DeploymentUnit> units, PooledArrayBuffer buf)
+        {
+            WriteEnumerable(units, buf, writerFunc: (unit, buf) =>
```
ptupitsyn, line 182: `buf` here hides parent method parameter with the same name - please rename.

ptupitsyn:
> I found the now unused `get_random_element` in the `compute.cpp` and realized that apparently C++ client always selects a random channel?

@isapego could you please have a look?

ptupitsyn (approved):
:+1:

isapego (approved):

isapego:
> I found the now unused `get_random_element` in the `compute.cpp` and realized that apparently C++ client always selects a random channel?

Yeah, channel is chosen randomly. Why? That's how all the clients work AFAIK.

isapego:
> > > I found the now unused `get_random_element` in the `compute.cpp` and realized that apparently C++ client always selects a random channel?
> > 
> > 
> > Yeah, channel is chosen randomly. Why? That's how all the clients work AFAIK.
> 
> Other clients are choosing the node from the set of nodes passed to `executeAsync` to increase the chance of local job execution.

Oh, you mean that. Yeah, partition awareness in not implemented in C++ client yet.

isapego:
> > Oh, you mean that. Yeah, partition awareness in not implemented in C++ client yet.
> 
> No, it's not even partition awareness. The `executeAsync` takes a set of nodes so the job is executed on a random node from this set. So when the client sends a request, it selects a random node from this set and sends a request to this node. The node then also picks a random node to execute a job on. In the extreme case suppose we have a client with 3 direct connections but the `executeAsync` is called with a single candidate node. All of the clients except C++ will send the request to this particular node, while the C++ client will pick a random client from the list of connections.

Yeah, this kind of optimization is not implemented in C++ yet.

PakhomovAlexander (approved):

PR #3105, opened 2024-01-26
IGNITE-21277: Sql. Partition pruning. Port partitionExtractor from AI2.

lowka:
Implementation of extract partition values from a predicate.

Basic idea - collect `values` (literal and dynamic parameter nodes) from a predicate into disjoint sets of `(colocation, key value)`, where `AND` operation adds a column to a set,  and `OR` operation creates a new set:

> c1 = 1 => [c1 = 1]
> c1 = 1 OR c1 = 2 => [c1 = 1], [c1 = 2]

When predicate expression traversal completes, traverse collected sets and if any of them does not contain all columns, then return none:

> colocation key: [c1, c2]
> predicate: c1 = 2 AND c2 = 3 
> result: [c1 = 2, c2 = 3]

> colocation key: [c1, c2]
> predicate: c1 = 2 
> result: []

Additionally adds test cases for `CASE` expressions (e.g. `CASE col1 WHEN 42 THEN true ELSE false END`) because such expressions are rewritten into combinations `OR`s with optional `IS NOT TRUE`s. 

https://issues.apache.org/jira/browse/IGNITE-21277

---

### The Review Checklist

### Notes

### DIFF
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/pruning/PartitionPruningMetadataExtractor.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/pruning/PartitionPruningMetadataExtractor.java
@@ -546,7 +546,8 @@ public class PartitionPruningMetadataExtractor extends IgniteRelShuttle {
                             }
                         }
 
-                        Int2ObjectMap<RexNode> newValue = new Int2ObjectArrayMap<>(val.columns);
+                        Int2ObjectMap<RexNode> newValue = new Int2ObjectArrayMap<>(val.columns.size() + candidate.columns.size());
+                        newValue.putAll(val.columns);
                         newValue.putAll(candidate.columns);
 
                         PruningColumnSet newColSet = new PruningColumnSet(newValue);


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/pruning/PartitionPruningColumns.java
```diff
@@ -0,0 +1,71 @@
...
+
+    /** {@inheritDoc} */
+    @Override
+    public String toString() {
+        return S.toString(this);
+    }
+
+    /** Returns column values in canonical form. E.g. {@code [1=2, 0=3]} becomes {@code [0=3, 1=2]} */
+    @TestOnly
+    public List<List<Map.Entry<Integer, RexNode>>> canonicalForm() {
```
AMashenkov, line 64: This can be a static method.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/pruning/PartitionPruningMetadataExtractor.java
```diff
@@ -0,0 +1,560 @@
...
+                    return Result.UNKNOWN;
+                }
+            }
+        }
+    }
+
+    private static boolean isColocationKey(RexNode node, List<Integer> keys) {
+        if (node instanceof RexLocalRef) {
+            RexLocalRef localRef = (RexLocalRef) node;
+            return keys.contains(localRef.getIndex());
```
AMashenkov, line 398: `contains` use unboxed value.
Maybe it make sense to use `ImmutableIntList` interface and `ImmutableIntList.indexOf(int)` instead. Or use fastutils`IntList` and rely data  from Calcite immutable list, right after getting from `distribution` via `IntArrayList.of(ImmutableIntList.toArray())`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/pruning/PartitionPruningMetadataExtractor.java
```diff
@@ -0,0 +1,560 @@
...
+        if (res instanceof PruningColumnSet) {
+            PruningColumnSet columnSet = (PruningColumnSet) res;
+            columnSets = new PruningColumnSets();
+            columnSets.candidates.add(columnSet);
+        } else {
+            columnSets = (PruningColumnSets) res;
+        }
+
+        for (PruningColumnSet columnSet : columnSets.candidates) {
+            if (!columnSet.columns.keySet().containsAll(keys)) {
```
AMashenkov, line 161: Seems, `containsAll` uses boxed collections, because `keys` kind is of Calcite `ImmutableIntList`.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/pruning/PartitionPruningMetadataExtractor.java
```diff
@@ -0,0 +1,560 @@
...
+                            if (!Objects.equals(existing, ckv.getValue())) {
+                                conflict = true;
+                                return;
+                            }
+                        }
+
+                        Int2ObjectMap<RexNode> newValue = new Int2ObjectArrayMap<>(val.columns);
+                        newValue.putAll(candidate.columns);
+
+                        PruningColumnSet newColSet = new PruningColumnSet(newValue);
```
AMashenkov, line 545:
```suggestion
                        Int2ObjectMap<RexNode> newValue = new Int2ObjectArrayMap<>(val.columns.size() + candidate.columns.size());
                        newValue.putAll(val.columns);
                        newValue.putAll(candidate.columns);

                        PruningColumnSet newColSet = new PruningColumnSet(newValue);
```
and maybe move to a separate method `concat(PruningColumnSet, PruningColumnSet)`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/pruning/PartitionPruningMetadataExtractor.java
lowka: Fixed automatically by using `IntList` instead of a boxed `java.util.Lis`t because `IntSet` has a method `containAll(IntCollection)`.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/pruning/PartitionPruningMetadataExtractor.java
```diff
@@ -0,0 +1,560 @@
...
+    @Override
+    public IgniteRel visit(IgniteIndexScan rel) {
+        RexNode condition = rel.condition();
+
+        IgniteTable table = rel.getTable().unwrap(IgniteTable.class);
+        assert table != null : "No table";
+
+        extractFromTable(rel.sourceId(), table, condition, rel.getCluster().getRexBuilder());
+
+        return super.visit(rel);
```
zstan, line 108:
```suggestion
        return rel;
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/pruning/PartitionPruningMetadataExtractor.java
zstan: ImmutableIntList.contains has complexity O(n) as i can see.

zstan (approved):
i fill some comments, overall looks good

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/pruning/PartitionPruningMetadataExtractor.java
AMashenkov: I believe there will be few int values (less then 10). 
So, complexity may affect less than data locality, bucket calculation and additional level of indirection.

AMashenkov (approved):

PR #3128, opened 2024-01-31
IGNITE-21371 Add tx coordinator id to TX requests

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-21371


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replication/request/CommittableTxRequest.java
```diff
@@ -32,6 +32,13 @@ public interface CommittableTxRequest extends ReplicaRequest {
      */
     boolean full();
 
+    /**
+     * Get the transaction coordinator inconsistent ID.
+     *
+     * @return the transaction coordinator inconsistent ID
+     */
+    String txCoordinatorId();
```
sanpwc, line 40: I'd say that it should be a field in ReadWriteReplicaRequest instead of CommittableTxRequest.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -248,6 +254,10 @@ public void name(String newName) {
         this.tableName = newName;
     }
 
+    private String txCoordinatorId() {
```
sanpwc, line 257: I'd rather add coordinatorId to InternalTransaction and thus remove given method from InternalTableImpl. What I mean here is that txManager should propagate coordinatorId to InternalTransaction on tx.begin().

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -170,6 +170,8 @@ public class InternalTableImpl implements InternalTable {
     /** Placement driver. */
     private final PlacementDriver placementDriver;
 
+    private final ClusterNode localNode;
```
sanpwc, line 173: Please check my comment bellow. I believe that we should retrieve coordinatorId from txManager on tx begin, thus we will not add given field here.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxMessageSender.java
```diff
@@ -48,6 +49,9 @@ public class TxMessageSender {
     /** Messaging service. */
     private final MessagingService messagingService;
 
+    /** Topology service. */
+    private final TopologyService topologyService;
```
sanpwc, line 53: I'd rather make move it to finish params list in a form of String txCoordId.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/message/TxFinishReplicaRequest.java
```diff
@@ -59,6 +59,13 @@ public interface TxFinishReplicaRequest extends PrimaryReplicaRequest, Timestamp
      */
     long commitTimestampLong();
 
+    /**
+     * Get the transaction coordinator inconsistent ID.
+     *
+     * @return the transaction coordinator inconsistent ID
+     */
+    String txCoordinatorId();
```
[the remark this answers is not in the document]
sanpwc: I mean logically. Why do we need coordinatorId on finish request? From my point of view, we use coordinatorId retrieved from txnState local map in two cases.

1. In order to check whether coordinator is dead and trigger the recovey, which definitely has no sense in case of finished transaction.
2. In order to decrement counter of an "in-flight" transactions that is used within index build procedure. In that case it's guaranteed that txnState local map will be populated with txCoordId prior to finish processing.

All in all, I believe that we may remove txCoordinatorId() from TxFinishReplicaRequest and update 
```
    private void markFinished(UUID txId, boolean commit, @Nullable HybridTimestamp commitTimestamp) {
        txManager.updateTxMeta(txId, old -> new TxStateMeta(
                commit ? COMMITTED : ABORTED,
                old == null ? null : old.txCoordinatorId(),
...
```

sanpwc (approved):
LGTM

PR #3139, opened 2024-02-01
IGNITE-21394 TimeoutException in the listener of pending assignments change shouldn't fail the watch processor

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-21394

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
@@ -97,6 +97,7 @@ import org.apache.ignite.internal.distributionzones.rebalance.RebalanceUtil;
 import org.apache.ignite.internal.hlc.HybridClock;
 import org.apache.ignite.internal.hlc.HybridTimestamp;
 import org.apache.ignite.internal.lang.ByteArray;
+import org.apache.ignite.internal.lang.IgniteInternalException;
 import org.apache.ignite.internal.lang.IgniteStringFormatter;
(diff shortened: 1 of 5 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1830,8 +1832,17 @@ private CompletableFuture<Void> changePeersOnRebalance(
         RaftGroupService partGrpSvc = table.internalTable().partitionRaftGroupService(partId);
 
         return partGrpSvc.refreshAndGetLeaderWithTerm()
+                .exceptionally(throwable -> {
+                    if (throwable instanceof TimeoutException) {
+                        LOG.info("Node couldn't get the leader within timeout so change peer is skipped [grp={}].", replicaGrpId);
```
denis-chudov, line 1837:
```suggestion
                        LOG.info("Node couldn't get the leader within timeout so the changing peers is skipped [grp={}].", replicaGrpId);
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1830,8 +1832,17 @@ private CompletableFuture<Void> changePeersOnRebalance(
...
+
+                        return LeaderWithTerm.NO_LEADER;
+                    }
+
+                    throw new IgniteInternalException(throwable);
```
denis-chudov, line 1842: its deprecated, why just not rethrow? it would make sense if you added some message

denis-chudov (approved):

PR #3148, opened 2024-02-02
IGNITE-21401 DataStreamer data removal: improve RW_UPSERT_ALL to support delete

ptupitsyn:
Add `ReadWriteMultiRowReplicaRequest#deleted` to indicate delete operations. Initially the idea was to use key-only rows to indicate delete, but it does not work for single-column tables.

https://issues.apache.org/jira/browse/IGNITE-21401


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replication/request/ReadWriteMultiRowReplicaRequest.java
```diff
@@ -31,4 +33,12 @@ public interface ReadWriteMultiRowReplicaRequest extends MultipleRowReplicaReque
...
      */
     boolean skipDelayedAck();
+
+    /**
+     * Optional operation types (one for every tuple in {@link #binaryTuples()}.
+     * See {@link RowOpType}.
+     *
+     * @return A byte for every tuple in {@link #binaryTuples()} indicating an operation type for that row.
+     */
+    byte @Nullable [] operationTypes();
```
ascherbakoff, line 43: Maybe use a bitset here ? A corresponding bit is set to 1 for removal, 0 otherwise.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/action/RowOpType.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.table.distributed.replicator.action;
+
+
+/**
+ * Row operation type.
+ */
+public enum RowOpType {
```
ascherbakoff, line 24: Enum is not required for two state op type, can just use a single bit.

ascherbakoff (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replication/request/ReadWriteMultiRowReplicaRequest.java
ptupitsyn: Good point, done.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/action/RowOpType.java
ptupitsyn: Done.

PR #3149, opened 2024-02-03
IGNITE-21438 Add thread assertions to MV partition and index storages

rpuch:
https://issues.apache.org/jira/browse/IGNITE-21438

### The Review Checklist

### Notes


modules/core/src/main/java/org/apache/ignite/internal/thread/ThreadAttributes.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Holds some thread attributes.
+ */
+@SuppressWarnings("InterfaceMayBeAnnotatedFunctional")
+public interface ThreadAttributes {
+    /**
+     * Returns all operations that this thread allows to execute.
+     */
+    Set<ThreadOperation> allowedOperations();
```
ibessonov, line 30: Would it be convenient to have a method `boolean allowsOperation(ThreadOperation op);`?

modules/core/src/main/java/org/apache/ignite/internal/util/JavaAssertions.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.util;
+
+/**
+ * Code to work with Java language assertions (those that relate to the {@code assert} keyword).
+ */
+public class JavaAssertions {
```
ibessonov, line 23: I think that this is not a good idea. Some developers might abuse methods like this. Generally speaking, all its usages could be replaced with different pattern in code. For example, extracting a method that would contain assertion logic.

modules/network/src/main/java/org/apache/ignite/internal/network/netty/NamedNioEventLoopGroup.java
```diff
@@ -59,7 +65,9 @@ protected Thread newThread(Runnable r, String unused) {
     /**
      * Marker class for network threads. Basically is just a {@link FastThreadLocalThread}.
      */
-    public static class NetworkThread extends FastThreadLocalThread {
+    public static class NetworkThread extends FastThreadLocalThread implements ThreadAttributes {
+        private static final Set<ThreadOperation> ALLOWED_OPERATIONS = unmodifiableSet(EnumSet.noneOf(ThreadOperation.class));
```
ibessonov, line 69: Why not simply assign an `emptySet()`?

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/ThreadAssertingMvPartitionStorage.java
```diff
@@ -0,0 +1,164 @@
...
+import org.apache.ignite.internal.util.Cursor;
+import org.apache.ignite.internal.worker.ThreadAssertions;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * {@link MvPartitionStorage} that performs thread assertions when doing read/write operations.
+ *
+ * @see ThreadAssertions
+ */
+public class ThreadAssertingMvPartitionStorage implements MvPartitionStorage {
```
ibessonov, line 37: Can you please explain the reason of having a wrapper class instead of putting assertions into `SnapshotAwarePartitionDataStorage`, for example? Or maybe we should put some assertions in places where we use API, not into the implementation.
Let me elaborate. For example, what if we have a cursor and somebody reads from it? Should we create wrapper for it as well? Should we create wrapper for all cursors in storages?
It makes the code too complicated in my opinion, too many classes and too many wrappers, and what are the benefits? The fact that you would place assertion into a single implementation rather than several implementations for several engines (potentially)? This is arguable. All I see is a code that re-implements a lot of methods of existing interface, forcing us to modify more classes if we modify the interface itself.

modules/workers/src/main/java/org/apache/ignite/internal/worker/ThreadAssertions.java
```diff
@@ -27,8 +26,19 @@
  * Tools to assert that the current thread allows to perform a requested operation.
  */
 public class ThreadAssertions {
+    public static final String ENABLED_PROPERTY = "thread.assertions.enabled";
```
ibessonov, line 29: I would add `ignite.` at the beginning, we should always claim our constants

ibessonov (approved):

PR #3150, opened 2024-02-03
IGNITE-21381 ActiveActorTest#testChangeLeaderForce is flaky

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-21381

### DIFF
--- a/modules/replicator/src/testFixtures/java/org/apache/ignite/internal/raft/client/AbstractTopologyAwareGroupServiceTest.java
+++ b/modules/replicator/src/testFixtures/java/org/apache/ignite/internal/raft/client/AbstractTopologyAwareGroupServiceTest.java
@@ -259,7 +259,6 @@ public abstract class AbstractTopologyAwareGroupServiceTest extends IgniteAbstra
 
         afterClusterInit(leader.name());
 
-        raftClients.add(raftClient);
         raftClients.add(raftClientNoInitialNotify);
 
         return new IgniteBiTuple<>(raftClient, raftClientNoInitialNotify);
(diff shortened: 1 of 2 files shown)


modules/raft/src/main/java/org/apache/ignite/internal/raft/Loza.java
```diff
@@ -175,6 +170,11 @@ public void actionRequestInterceptor(ActionRequestInterceptor actionRequestInter
     /** {@inheritDoc} */
     @Override
     public CompletableFuture<Void> start() {
+        executor = new ScheduledThreadPoolExecutor(
```
alievmirza, line 173: All over the codebase we create executors in constructors, why have you moved this executor to the start method?

alievmirza (changes requested):
Please take a look at tests in `org.apache.ignite.internal.raft.client.AbstractTopologyAwareGroupServiceTest`.

We still call `stopCluster` in the tests itself, that mean that if test fail, we won't stop cluster. Please move all resource cleanup to AfterEach section

modules/replicator/src/testFixtures/java/org/apache/ignite/internal/raft/client/AbstractTopologyAwareGroupServiceTest.java
```diff
@@ -261,26 +259,24 @@ private IgniteBiTuple<TopologyAwareRaftGroupService, TopologyAwareRaftGroupServi
 
         afterClusterInit(leader.name());
 
+        raftClients.add(raftClient);
```
alievmirza, line 262: I would move this to the `org.apache.ignite.internal.raft.client.AbstractTopologyAwareGroupServiceTest#startCluster`

because in the test `org.apache.ignite.internal.raft.client.AbstractTopologyAwareGroupServiceTest#testOneNodeReplicationGroup` we do not add created client to `raftClients`, hence we do not shutdown it

vldpyatkov: It was just a mistack. We can't move the cluster start before startup because the number of nodes is defined by a test.

alievmirza (approved):

PR #3210, opened 2024-02-13
IGNITE-19608 Improve the UIX of ignite3db script

PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-19608

The following changes are made:
* The output is redirected to the ignite3db-startup.log
* Useful error-handling flags are set
* Timeout is set to wait java application startup.

### DIFF
--- a/packaging/zip/ignite3db
+++ b/packaging/zip/ignite3db
@@ -50,7 +50,7 @@ start() {
   done
 
  if [ ! -f "$rest_address_file" ]; then
-     echo "Failure: Application did not start within the expected time frame. Check out ${STARTUP_LOG_FILE}." >&2
+     echo "Failure: Application did not start within the expected time frame. Check out ${STARTUP_LOG_FILE}" >&2
      exit 1
  else
      rest_address=$(cat "$rest_address_file")
--- a/packaging/zip/ignite3db
+++ b/packaging/zip/ignite3db
@@ -18,10 +18,6 @@
 #
 set -o errexit # Exit on error
 set -o nounset # Exit on unset variables
-if [ -n "$(command -v bash)" ]; then
-  set -o pipefail # Exit if any command in a pipeline fails
-fi
-
 
 SCRIPT_DIR=$(dirname "$0")
 
--- a/packaging/zip/ignite3db
+++ b/packaging/zip/ignite3db
@@ -16,6 +16,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+set -o errexit # Exit on error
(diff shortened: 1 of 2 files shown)


packaging/zip/ignite3db
```diff
@@ -16,6 +16,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+set -o errexit # Exit on error
+set -o nounset # Exit on unset variables
+set -o pipefail # Exit if any command in a pipeline fails
```
ahitrin, line 21: A situation with `pipefail` in non-bash shells is [controversal](https://unix.stackexchange.com/questions/654885/who-is-responsible-for-providing-set-o-pipefail/654932). Some shells, especially in legacy environments (even POSIX `sh`), may not support it. I'm afraid it may be risky to use `/bin/sh` as a shebang with it.

Probably, a `/bin/bash` or `/usr/bin/env bash` shebang is more preferrable now.

packaging/zip/ignite3db
```diff
@@ -34,14 +38,26 @@ cd ${IGNITE_HOME} || exit
...
+  count=0
+  max_attempts=60 # 30 seconds, considering sleep of 0.5s
+  while [ ! -f "$rest_address_file" ] && [ "$count" -lt "$max_attempts" ]; do
+    sleep 0.5
+    ((count++))
+  done
 
-  echo "Node named ${NODE_NAME} started successfully. REST addresses are [$rest_address]"
+ if [ ! -f "$rest_address_file" ]; then
+     echo "Failure: Application did not start within the expected time frame." >&2
```
ahitrin, line 55: Probably, a mention of `$STARTUP_LOG_FILE` could be useful here. WDYT?

packaging/zip/ignite3db
```diff
@@ -34,14 +36,26 @@ cd ${IGNITE_HOME} || exit
...
+  max_attempts=30
+  count=0
+  while [ ! -f "$rest_address_file" ] && [ "$count" -lt "$max_attempts" ]; do
+    sleep 1
+    count=$((count + 1))
+  done
 
-  echo "Node named ${NODE_NAME} started successfully. REST addresses are [$rest_address]"
+ if [ ! -f "$rest_address_file" ]; then
+     echo "Failure: Application did not start within the expected time frame. Check out ${STARTUP_LOG_FILE}." >&2
```
ahitrin, line 53: Optional: probably, there shouldn't be a dot after the log name (in order to simplify copy-paste)

ahitrin (approved):

PR #3219, opened 2024-02-15
IGNITE-21291 Scan cursors do not close when an RO transaction is finalized

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-21291


modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/ClosedTransactionTracker.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.ignite.internal.tx.message.TxMessagesFactory;
+import org.apache.ignite.network.ClusterNode;
+import org.apache.ignite.network.TopologyService;
+
+/**
+ * Keeps track of all closed RO transactions.
+ */
+public class ClosedTransactionTracker {
+
+    private static final int maxClosedTransactionsInBatch = 10_000;
```
denis-chudov, line 39: should be upper case

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/ClosedTransactionTracker.java
```diff
@@ -0,0 +1,93 @@
...
+    private static final long RPC_TIMEOUT = 3000;
+
+    /** Tx messages factory. */
+    private static final TxMessagesFactory FACTORY = new TxMessagesFactory();
+
+    private final Set<UUID> closedTransactions = ConcurrentHashMap.newKeySet();
+
+    /**
+     * Topology service.
+     */
```
denis-chudov, line 50:
```suggestion
    /** Topology service. */
```

modules/transactions/src/test/java/org/apache/ignite/internal/tx/impl/ReadOnlyTransactionImplTest.java
```diff
@@ -36,12 +36,22 @@ class ReadOnlyTransactionImplTest extends BaseIgniteAbstractTest {
...
+    @Mock
+    private ResourceCleanupManager resourceCleanupManager;
+
     @Test
     void effectiveSchemaTimestampIsReadTimestamp() {
         HybridTimestamp readTimestamp = new HybridClockImpl().now();
         UUID txId = TestTransactionIds.TRANSACTION_ID_GENERATOR.transactionIdFor(readTimestamp);
 
-        var tx = new ReadOnlyTransactionImpl(txManager, new HybridTimestampTracker(), txId, "localId", readTimestamp);
+        var tx = new ReadOnlyTransactionImpl(
```
denis-chudov, line 47: What do you think about adding additional tests checking that RO transaction's cursors are closed after its finish?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/CursorCleanupRequestHandler.java
```diff
@@ -0,0 +1,66 @@
...
+            }
+        });
+    }
+
+    public void stop() {
+    }
+
+    private void processTxCleanup(CloseCursorsBatchMessage closeCursorsMessage) {
+        closeCursorsMessage.transactions().forEach(resourcesRegistry::close);
+    }
```
denis-chudov, line 65: Here we can process potentially huge batch of transactions (tracker allows max size of 10k), I an not sure it's a good idea to do this in network thread. What do you think about adding it to resource cleanup manager to close resources in background? On the other side, it's additional pressure on the heap, but network delay may be more critical

modules/transactions/src/main/java/org/apache/ignite/internal/tx/message/CloseCursorsBatchMessage.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Collection;
+import java.util.UUID;
+import org.apache.ignite.internal.network.annotations.Transferable;
+import org.apache.ignite.internal.replicator.message.TimestampAware;
+
+/**
+ * Close transaction cursors message.
+ */
+@Transferable(TxMessageGroup.TX_CLOSE_CURSORS)
+public interface CloseCursorsBatchMessage extends TimestampAware {
```
denis-chudov, line 29: What about `FinishedTransactionsBatchMessage`?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -291,12 +297,13 @@ public TxManagerImpl(
...
         this.localRwTxCounter = localRwTxCounter;
         this.partitionOperationsExecutor = partitionOperationsExecutor;
+        this.resourceCleanupManager = resourceCleanupManager;
 
         placementDriverHelper = new PlacementDriverHelper(placementDriver, clock);
 
         int cpus = Runtime.getRuntime().availableProcessors();
 
-        cleanupExecutor = new ThreadPoolExecutor(
+        writeIntentSwitchPool = new ThreadPoolExecutor(
```
denis-chudov, line 306: It think it's worthy to rename as well
- `TxManagerImpl#executeCleanupAsync`
- `IgniteThreadFactory.create(nodeName, "tx-async-cleanup", LOG, STORAGE_READ, STORAGE_WRITE)` - thread name

denis-chudov (approved):

modules/table/src/integrationTest/java/org/apache/ignite/internal/table/ItTransactionRecoveryTest.java
```diff
@@ -955,6 +958,78 @@ private void scanSingleEntryAndLeaveCursorOpen(IgniteImpl targetNode, TableViewI
...
+        assertEquals(1, txExecNode.resourcesRegistry().resources().size());
+
+        roTx.commit();
+
+        // Wait for the cursor cleanup message to arrive.
+        assertThat(txRequestCaptureFut, willCompleteSuccessfully());
+        assertThat(txResponseCaptureFut, willCompleteSuccessfully());
+
+        // Now check that the cursor is closed.
+        assertEquals(0, txExecNode.resourcesRegistry().resources().size());
```
vldpyatkov, line 1030: Here, the FinishedTransactionsBatchResponse might not be handled yet, isn't it?
Why don't you wait for the condition?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/FinishedTransactionBatchRequestHandler.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public void start() {
+        messagingService.addMessageHandler(TxMessageGroup.class, (msg, sender, correlationId) -> {
+            if (msg instanceof FinishedTransactionsBatchMessage) {
+                processTxCleanup((FinishedTransactionsBatchMessage) msg, sender, correlationId);
+            }
+        });
+    }
+
+    public void stop() {
```
vldpyatkov, line 78: I do not see where this method is used.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/FinishedTransactionBatchRequestHandler.java
```diff
@@ -0,0 +1,95 @@
...
+        asyncExecutor.execute(() -> {
+            closeCursorsMessage.transactions().forEach(resourcesRegistry::close);
+
+            messagingService.respond(sender, prepareResponse(), correlationId);
+        });
+    }
+
+    private NetworkMessage prepareResponse() {
+        return FACTORY
+                .finishedTransactionsBatchResponse()
```
vldpyatkov, line 91: Why is this response needed?
Besides, it is aware of the timestamp.

vldpyatkov (approved):

PR #3220, opened 2024-02-15
IGNITE-20553 Add await self in local logical topology logic to node startup procedure

sanpwc:
https://issues.apache.org/jira/browse/IGNITE-20553

- Await self in logical topology logic added withih node start.
- Fix missing update of a CMG raft client peersAndLearners after rebalance.


modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/api/LogicalTopologyService.java
```diff
@@ -62,6 +62,11 @@ public interface LogicalTopologyService {
      */
     CompletableFuture<LogicalTopologySnapshot> logicalTopologyOnLeader();
 
+    /**
+     * Retrieves the current logical topology snapshot stored in the local storage.
+     */
+    LogicalTopologySnapshot getLogicalTopology();
```
rpuch, line 68: Does it make sense to reflect it in the name of the method that the topology is returned based on the local view of the node? Like `localLogicalTopology()`.

Also, `get` prefix does not match the style of `logicalTopologyOnLeader()`: they should both have `get` prefix, or none of them should have it.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -990,6 +994,37 @@ public CompletableFuture<Ignite> start(Path configPath) {
 
                         return cmgMgr.onJoinReady();
                     }, startupExecutor)
+                    .thenComposeAsync(ignored -> {
+                        CompletableFuture<Void> awaitSelfInLogicalTopologyFuture = new CompletableFuture<>();
+
+                        LogicalTopologyEventListener awaitSelfListener = new LogicalTopologyEventListener() {
+                            @Override
+                            public void onNodeJoined(LogicalNode joinedNode, LogicalTopologySnapshot newTopology) {
+                                if (newTopology.nodes().stream().map(LogicalNode::id).collect(Collectors.toSet()).contains(id())) {
```
rpuch, line 1003: If `toSet` is imported statically, this will read nicely in plain English: 'collect to set'

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -990,6 +994,37 @@ public CompletableFuture<Ignite> start(Path configPath) {
 
                         return cmgMgr.onJoinReady();
                     }, startupExecutor)
+                    .thenComposeAsync(ignored -> {
+                        CompletableFuture<Void> awaitSelfInLogicalTopologyFuture = new CompletableFuture<>();
```
rpuch, line 998: Is it possible to extract this as a method? The method in which this code is defined is already huge.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -990,6 +994,37 @@ public CompletableFuture<Ignite> start(Path configPath) {
...
+                            public void onNodeJoined(LogicalNode joinedNode, LogicalTopologySnapshot newTopology) {
+                                if (newTopology.nodes().stream().map(LogicalNode::id).collect(Collectors.toSet()).contains(id())) {
+                                    awaitSelfInLogicalTopologyFuture.complete(null);
+                                    logicalTopologyService.removeEventListener(this);
+                                }
+                            }
+
+                            @Override
+                            public void onTopologyLeap(LogicalTopologySnapshot newTopology) {
+                                if (newTopology.nodes().stream().map(LogicalNode::id).collect(Collectors.toSet()).contains(id())) {
```
rpuch, line 1011: It seems like there are 3 duplicates of this code, it it possible to extract them to methods?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/topology/api/LogicalTopologyService.java
sanpwc: Agee. I've used the mehod naming from internals, but you proposal looks reasonable.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
sanpwc: Sure.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
sanpwc: Ok, make sense.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
sanpwc: Don't think that it's really needed, taking into considiration that everething inside thenCompose will be extracted I'd rather left inner duplication as is. However if you will intend I'll also add a private helper for this.

sanpwc: Ok, will add inner method.

rpuch (approved):

PR #3222, opened 2024-02-15
IGNITE-20593 Sql. Fix all possible default expressions

zstan:

### The Review Checklist

### Notes

### DIFF
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ddl/DdlSqlToCommandConverter.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ddl/DdlSqlToCommandConverter.java
@@ -366,7 +366,7 @@ public class DdlSqlToCommandConverter {
 
             dedupSetPk.remove(name);
 
-            DefaultValueDefinition dflt = convertDefault(col.expression, relType);
+            DefaultValueDefinition dflt = convertDefault(col.expression, relType, name);
             if (dflt.type() == DefaultValueDefinition.Type.FUNCTION_CALL && !pkCols.contains(name)) {
                 throw new SqlException(STMT_VALIDATION_ERR,
                         "Functional defaults are not supported for non-primary key columns [col=" + name + "]");
@@ -408,9 +408,8 @@ public class DdlSqlToCommandConverter {
 
             Boolean nullable = col.dataType.getNullable();
             RelDataType relType = ctx.planner().convert(col.dataType, nullable != null ? nullable : true);
-            DefaultValueDefinition dflt = convertDefault(col.expression, relType);
-
             String name = col.name.getSimple();
+            DefaultValueDefinition dflt = convertDefault(col.expression, relType, name);
 
             cols.add(new ColumnDefinition(name, relType, dflt));
         }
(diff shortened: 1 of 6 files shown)


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ddl/DdlSqlToCommandConverter.java
```diff
@@ -824,60 +842,142 @@ private Type convertIndexType(IgniteSqlIndexType type) {
...
                 case BOOLEAN:
-                    return literal.getValueAs(Boolean.class);
+                    String val = Objects.requireNonNull(literal.toValue());
+                    val = val.trim().toLowerCase();
+
+                    if ("true".equals(val) || "yes".equals(val) || "1".equals(val)) {
+                        return true;
+                    }
+
+                    if ("false".equals(val) || "no".equals(val) || "0".equals(val)) {
```
lowka, line 964: `numeric` to `boolean` is not a legal type transformation - both standard disallows iand current implementation reject it.

> sql("SELECT 1::BOOLEAN");
>Err: Cast function cannot convert value of type INTEGER to type BOOLEAN

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ddl/DdlSqlToCommandConverter.java
```diff
@@ -824,60 +842,142 @@ private Type convertIndexType(IgniteSqlIndexType type) {
...
             return null;
         }
 
         try {
             switch (columnType) {
-                case STRING:
-                    return literal.getValueAs(String.class);
+                case PERIOD: {
+                    if (!(literal instanceof SqlCharStringLiteral) && !(literal instanceof SqlIntervalLiteral)) {
+                        throw new SqlException(STMT_VALIDATION_ERR,
```
lowka, line 854: I think we should come up with a better error message in case of this error (for both PERIOD and DURATION).

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ddl/DdlSqlToCommandConverter.java
lowka: `no` `yes` are also not considered legal.

zstan: ok i will check standard but .. i think that cast differs from default expressions and also this how it works in PG.

korlov42: it's Ignite, not PG

korlov42 (changes requested):
the ticket states, that we should make DDL path _consistent_ with regular query execution. Instead, you have introduced yet another implementation of cast rules and (even worse) aligned it with Postgres, not the Ignite.

In order, to make behaviour of the system consistent across different paths, we have to reuse the same infrastructure. It's the only way to make it consistent and be sure it will remain so

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ddl/DdlSqlToCommandConverter.java
zstan: fixed

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/ddl/DdlSqlToCommandConverter.java
AMashenkov: Why can't we apply Default expression to interval type? Because intervals are not supported?

korlov42 (approved):

PR #3227, opened 2024-02-19
IGNITE-21554 Encapsulate raft-related code of InternalTable

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-21554

- Moved raft-related methods to a separate entity - TableRaftService
- InternalTableImpl now uses the consistent way of building a replication group id

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/TableRaftService.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/TableRaftService.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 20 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/TableRaftServiceImpl.java
```diff
@@ -0,0 +1,227 @@
...
+public class TableRaftServiceImpl implements TableRaftService {
+
+    /** Mutex for the partition maps update. */
+    private final Object updatePartitionMapsMux = new Object();
+
+    /** Map update guarded by {@link #updatePartitionMapsMux}. */
+    private volatile Int2ObjectMap<RaftGroupService> raftGroupServiceByPartitionId;
+
+    /** Map update guarded by {@link #updatePartitionMapsMux}. */
+    private volatile Int2ObjectMap<PendingComparableValuesTracker<HybridTimestamp, Void>> safeTimeTrackerByPartitionId = emptyMap();
```
kgusakov, line 56: Not sure that this item and the `storageIndexTrackerByPartitionId` has the semantic connection with table *raft* service, as the class name says

modules/table/src/main/java/org/apache/ignite/internal/table/TableRaftService.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.ignite.internal.lang.IgniteInternalException;
+import org.apache.ignite.internal.raft.service.RaftGroupService;
+import org.apache.ignite.internal.util.PendingComparableValuesTracker;
+import org.apache.ignite.network.ClusterNode;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Internal facade that provides methods for table's raft operations.
+ */
+public interface TableRaftService extends ManuallyCloseable {
```
kgusakov, line 31: Do we really need the separate interface for this kind of small utility class?

kgusakov (approved):

kgusakov (changes requested):
Pls, see comments

kgusakov (approved):

PR #3236, opened 2024-02-20
IGNITE-20242 Retry outdated schema error

isapego:
- Added tests;
- Found an issue in the current implementation - there was a crash if server returns a tuple with the unknown schema version;
- Re-factored `table_impl` to allow performing multiple subsequent async calls;
- Read and store known error extensions data from server;
- Retry operation with specified schema on error.

### DIFF
--- a/modules/platforms/cpp/ignite/client/detail/cluster_connection.cpp
+++ b/modules/platforms/cpp/ignite/client/detail/cluster_connection.cpp
@@ -212,4 +212,35 @@ std::shared_ptr<node_connection> cluster_connection::get_random_channel() {
     return std::next(m_connections.begin(), idx)->second;
 }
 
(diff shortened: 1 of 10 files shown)


modules/platforms/cpp/ignite/client/detail/cluster_connection.cpp
```diff
@@ -212,4 +212,35 @@ std::shared_ptr<node_connection> cluster_connection::get_random_channel() {
...
+
+        return;
+    }
+
+    while (true) {
+        auto channel = get_random_channel();
+        if (!channel)
+            throw ignite_error("No nodes connected");
+
+        auto res = channel->perform_request(op, wr, handler);
```
ptupitsyn, line 234: This sends data to the socket, but does not wait for the result, right? So we only retry connection-related issues in this loop?

ptupitsyn (approved):

PR #3264, opened 2024-02-22
IGNITE-21452: add fail handler to query task execution

ygerzhedovich:

### The Review Checklist

### Notes


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/QueryTaskExecutorImpl.java
```diff
@@ -81,15 +81,30 @@ public void execute(UUID qryId, long fragmentId, Runnable qryTask) {
                 () -> {
                     try {
                         qryTask.run();
+                    } catch (IgniteException | IgniteInternalException iex) {
+                        //it's normal situation when we have our own exception here. Just ignore.
+                        LOG.debug("Uncaught exception", iex);
+                    } catch (Error err) {
```
lowka, line 87: I think it should catch `Throwable` here.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExchangeServiceImpl.java
```diff
@@ -114,9 +114,7 @@ public CompletableFuture<Void> sendError(String nodeName, UUID queryId, long fra
 
         if (!(traceableErr instanceof TraceableException)) {
             traceableErr = error = new IgniteInternalException(INTERNAL_ERR, error);
-        }
 
-        if (!(traceableErr instanceof QueryCancelledException)) {
```
korlov42, line 119: why did you merge these branches?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionContext.java
```diff
@@ -310,6 +311,10 @@ public void execute(RunnableX task, Consumer<Throwable> onError) {
             } catch (Throwable e) {
                 onError.accept(e);
 
+                if (e instanceof IgniteException) {
```
korlov42, line 314: perhaps, we should not re-throw exceptions here at all. WDYT?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/QueryTaskExecutorImpl.java
```diff
@@ -81,15 +81,30 @@ public void execute(UUID qryId, long fragmentId, Runnable qryTask) {
                 () -> {
                     try {
                         qryTask.run();
+                    } catch (IgniteException | IgniteInternalException iex) {
+                        //it's normal situation when we have our own exception here. Just ignore.
+                        LOG.debug("Uncaught exception", iex);
```
korlov42, line 86: instead of deciding for caller which exceptions are severe and which are not, why don't force caller to do exception handling in submitted task? In my opinion, every exception is severe on this level, thus I would just submit everything to `failureProcessor`

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExchangeServiceImpl.java
ygerzhedovich: To avoid excessive logging. For example `Division by zero` in query was here.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionContext.java
ygerzhedovich: agree, thanks for the note

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/QueryTaskExecutorImpl.java
ygerzhedovich: agree, thanks for the note

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionContext.java
```diff
@@ -310,6 +311,16 @@ public void execute(RunnableX task, Consumer<Throwable> onError) {
...
+                    return;
+                }
+
+                if (e instanceof AssertionError) {
```
zstan, line 318: still not clear to me ((( if we want to defense here from some calcite assertions probably we can implement it somehow more accurate ? What test will fail if i comment this part ? i can`t find it quick ((

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExchangeServiceImpl.java
```diff
@@ -170,6 +168,10 @@ private void onMessage(String nodeName, QueryBatchMessage msg) {
             } catch (Throwable e) {
                 inbox.onError(e);
 
+                if (e instanceof IgniteException) {
```
zstan, line 171: why do we need to play such games ? Exceptions is global system machinery thus such code will bring more harm for further debugging if someone somewhere will decide to change one exception into another ? can you point me test which fail if i comment this part plz ?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/rel/ExecutionTest.java
```diff
@@ -621,25 +621,25 @@ static class CorruptedNode<T> implements Node<T> {
         /** {@inheritDoc} */
         @Override
         public ExecutionContext<T> context() {
-            throw new AssertionError();
+            throw new IllegalAccessError();
```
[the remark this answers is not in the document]
ygerzhedovich: agree

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExchangeServiceImpl.java
ygerzhedovich: IgniteException is our own exception which we expect here and which will be processed by inbox. For any other exception we consider as unexpected error which should be processed by fail handler.
As example situation when it works - we trying insert value wider then allowed values. In this case we receive SqlException with `INTEGER out of range` (for example) and the error is expected.

To write test right now is hard unfortunately due to we can't configure fail handler yet. I will try invent how it can be tested.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionContext.java
ygerzhedovich: We known that Calcite can throw AssertionError. Such error doesn't mean that should fail handler work for the case - it just error for one concrete query which just will be failed for user, we shouldn't pass it to fail handler. In case it will be our own AssertionError - it's also a bug which should be logged to have ability to analyze it.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExchangeServiceImpl.java
ygerzhedovich: @zstan I've added test ItDmlTest#testNoFailHandlerForRuntimeSqlError

modules/failure-handler/src/main/java/org/apache/ignite/internal/failure/FailureProcessor.java
```diff
@@ -49,6 +51,8 @@ public class FailureProcessor implements IgniteComponent {
     /** Node name. */
     private final String nodeName;
 
+    private volatile @Nullable FailureHandler interceptor;
```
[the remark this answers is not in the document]
ygerzhedovich: it will be used with different threads, so it required have volatile here or synchronized settet, which you suggested below. In this case it doesn't matter which approach to use.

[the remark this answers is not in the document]
zstan: it`s matter, check: if you use volatile approach - check can passed but in next line you can obtain NPE
        if (interceptor != null) {
            interceptor.onFailure(nodeName, failureCtx);
        }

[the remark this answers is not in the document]
ygerzhedovich: agree, thanks

zstan (approved):
looks good

modules/failure-handler/src/main/java/org/apache/ignite/internal/failure/FailureProcessor.java
```diff
@@ -145,4 +154,14 @@ private static boolean failureTypeIgnored(FailureContext failureCtx, FailureHand
...
                 && ((AbstractFailureHandler) handler).ignoredFailureTypes().contains(failureCtx.type());
     }
+
+    /**
+     * Set FailHander interceptor to provide ability tщ test scenarios related to fail handler.
+     *
+     * @param interceptor Interceptor of fails.
+     */
+    @TestOnly
+    public synchronized void setInterceptor(@Nullable FailureHandler interceptor) {
```
korlov42, line 164: does it make to add TODO here to remove this method when failure handler becomes configurable?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExchangeServiceImpl.java
```diff
@@ -176,6 +174,10 @@ private void onMessage(String nodeName, QueryBatchMessage msg) {
...
+                    return;
+                }
+
                 throw new IgniteInternalException(INTERNAL_ERR, "Unexpected exception", e);
```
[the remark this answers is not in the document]
ygerzhedovich: it require to pass the error to fail handler in case it unexpected to us

korlov42 (approved):

PR #3267, opened 2024-02-22
IGNITE-21426 Create tables from Java classes basic implementation

valepakh:
https://issues.apache.org/jira/browse/IGNITE-21426


modules/api/src/main/java/org/apache/ignite/catalog/annotations/Column.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * Describes a column of the table. Type of the column is derived from the type of the annotated field or method.
+ */
+@Target({METHOD, FIELD})
+@Retention(RUNTIME)
+public @interface Column {
```
PakhomovAlexander, line 32: Why do we have Col and Column?

modules/catalog-dsl/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+apply from: "$rootDir/buildscripts/java-core.gradle"
+apply from: "$rootDir/buildscripts/publishing.gradle"
+apply from: "$rootDir/buildscripts/java-junit5.gradle"
+apply from: "$rootDir/buildscripts/java-integration-test.gradle"
+
+description = 'ignite-catalog-ddl'
```
PakhomovAlexander, line 23: dsl or ddl?

modules/catalog-dsl/src/main/java/org/apache/ignite/internal/catalog/sql/CreateFromAnnotationsImpl.java
```diff
@@ -0,0 +1,195 @@
...
+
+    private IndexType pkType;
+
+    CreateFromAnnotationsImpl(IgniteSql sql, Options options) {
+        super(sql, options);
+    }
+
+    CreateFromAnnotationsImpl keyValueView(Class<?> keyClass, Class<?> valueClass) {
+        if (keyClass.getAnnotation(Table.class) == null && valueClass.getAnnotation(Table.class) == null) {
+            throw new IllegalArgumentException("Table annotation must be present.");
```
PakhomovAlexander, line 54: I think a user-facing error message should be as clear as possible. Here and everywhere else I would expect something like: "Can not find @Table annotation on keyCless.name() or valueClass.name. One of these classes must be annotated with @Table ...".

modules/catalog-dsl/src/main/java/org/apache/ignite/internal/catalog/sql/Name.java
```diff
@@ -0,0 +1,48 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.catalog.sql;
+
+/**
+ * Qualified SQL identifier.
```
PakhomovAlexander, line 21: Quoted?

modules/catalog-dsl/src/main/java/org/apache/ignite/internal/catalog/sql/CreateFromAnnotationsImpl.java
```diff
@@ -0,0 +1,195 @@
...
+
+    private CreateTableImpl createTable;
+
+    private IndexType pkType;
+
+    CreateFromAnnotationsImpl(IgniteSql sql, Options options) {
+        super(sql, options);
+    }
+
+    CreateFromAnnotationsImpl keyValueView(Class<?> keyClass, Class<?> valueClass) {
```
PakhomovAlexander, line 52: Why is the method called ..View? It has nothing in common with a view.

modules/catalog-dsl/src/test/java/org/apache/ignite/internal/catalog/sql/CreateTableTest.java
```diff
@@ -0,0 +1,188 @@
...
+        sql = createTableQuoted().ifNotExists().name("table1").addColumn("col1", INTEGER).toSqlString();
+        assertThat(sql, is("CREATE TABLE IF NOT EXISTS \"table1\" (\"col1\" int);"));
+    }
+
+    @Test
+    void names() {
+        String sql = createTable().name("public", "table1").addColumn("col1", INTEGER).toSqlString();
+        assertThat(sql, is("CREATE TABLE public.table1 (col1 int);"));
+
+        sql = createTable().name("", "table;1--test\n\r\t;").addColumn("col1", INTEGER).toSqlString();
```
PakhomovAlexander, line 45: Thaaaat is reeeeally interesting. How do we define that "table;1--test\n\r\t;" will became "table1"???

modules/catalog-dsl/src/test/java/org/apache/ignite/internal/catalog/sql/CreateTableTest.java
```diff
@@ -0,0 +1,188 @@
...
+        assertThat(sql, is("CREATE TABLE table1 (col1 int);"));
+
+        // quote identifiers
+        sql = createTableQuoted().name("public", "table1").addColumn("col1", INTEGER).toSqlString();
+        assertThat(sql, is("CREATE TABLE \"public\".\"table1\" (\"col1\" int);"));
+
+        sql = createTableQuoted().name("", "table;1--test\n\r\t;").addColumn("col1", INTEGER).toSqlString();
```
PakhomovAlexander, line 52: And here it is even more interesting.

modules/catalog-dsl/src/test/java/org/apache/ignite/internal/catalog/sql/CreateTableTest.java
PakhomovAlexander: Good point, maybe we should fail the query in these case?

PR #3281, opened 2024-02-23
IGNITE-21600 Make partition-operations a plain executor

rpuch:
It was decided that we don't need the pool to be striped, so it should be replaced with a plain thread pool to avoid confusion.

https://issues.apache.org/jira/browse/IGNITE-21600


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1484,7 +1477,7 @@ private CompletableFuture<List<Table>> tablesAsyncInternalBusy() {
                             .toArray(CompletableFuture[]::new);
 
                     return CompletableFutures.allOf(tableImplFutures);
-                }), ioExecutor);
```
sashapolo, line 1487: Why is this change needed?

sashapolo (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
rpuch: The wrapper around SchemaSyncService switches execution to the correct thread, so this would do a second switch

PR #3290, opened 2024-02-24
IGNITE-21602 Fix direct buffers allocation on Java 21

rpuch:
* DirectByteBuffer constructor signature has changed, length is now long, not int
* MethodHandle-based mechanism used on Java 11 was broken as a wrong package was used (jdk.internal.misc instead of jdk.internal.access)

https://issues.apache.org/jira/browse/IGNITE-21602

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/GridUnsafe.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/GridUnsafe.java
@@ -107,9 +107,6 @@ public abstract class GridUnsafe {
     /** {@link java.nio.Buffer#address} field offset. */
     private static final long DIRECT_BUF_ADDR_OFF = bufferAddressOffset();
 
-    /** Null object. */
-    private static final Object NULL_OBJ = null;
-
     /**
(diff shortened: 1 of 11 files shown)


modules/core/src/main/java/org/apache/ignite/internal/util/GridUnsafe.java
```diff
@@ -2134,4 +2116,93 @@ public static boolean compare(long ptr1, long ptr2, int size) {
...
+            this.newDirectBufMh = newDirectBufMh;
+            this.javaNioAccessObj = javaNioAccessObj;
+        }
+
+        @Override
+        public ByteBuffer wrapPointer(long ptr, int len) {
+            try {
+                ByteBuffer buf = (ByteBuffer) newDirectBufMh.invokeExact(javaNioAccessObj, ptr, len, NULL_OBJ);
+
+                assert buf.isDirect();
```
tkalkirill, line 2152:
```suggestion
                assert buf.isDirect(): "ptr=" + ptr + ", len=" + len;
```

PR #3300, opened 2024-02-27
IGNITE-20373 Fix IO threading model

rpuch:
https://issues.apache.org/jira/browse/IGNITE-20373


modules/network/src/main/java/org/apache/ignite/internal/network/DefaultMessagingService.java
```diff
@@ -337,82 +337,167 @@ private List<ClassDescriptorMessage> prepareMarshal(NetworkMessage msg) throws E
...
+        // Now execute those handlers that are guaranteed to be executed on the same thread as the first one.
+        for (NetworkMessageHandler handler : handlersWantingSameThreadAsFirst) {
+            handler.onReceived(payload, senderConsistentId, correlationId);
         }
     }
 
-    private boolean senderIdIsStale(InNetworkObject obj) {
-        return staleIdDetector.isIdStale(obj.launchId());
+    private static boolean wantSamePool(HandlerContext handlerContext1, HandlerContext handlerContext2) {
+        return wantsInboundPool(handlerContext1) && wantsInboundPool(handlerContext2);
```
vldpyatkov, line 496: Why is it not calculated by Choosers comparison?
handlerContext1.executorChooser() == handlerContext2.executorChooser()

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/ItIgniteNodeRestartTest.java
```diff
@@ -351,11 +353,16 @@ private PartialNode startPartialNode(
         LongSupplier partitionIdleSafeTimePropagationPeriodMsSupplier
                 = () -> TestIgnitionManager.DEFAULT_PARTITION_IDLE_SYNC_TIME_INTERVAL_MS;
 
-        var replicaService = new ReplicaService(
+        MessagingService messagingServiceReturningToStorageOperationsPool = new JumpToExecutorByConsistentIdAfterSend(
                 clusterSvc.messagingService(),
-                hybridClock,
                 name,
-                threadPools.partitionOperationsExecutor()
+                message -> threadPoolsManager.partitionOperationsExecutor()
```
vldpyatkov, line 359: It is weird that the pool is called a partial operation.
This looks like a common system pool, based on this usage pattern.

rpuch: This pool is used for processing operations on partitions, hence the name. The wrapper around MessagingService that is constructed here is also used by the components that deal with processing operations on partitions, so it returns the control to the same partition-operations pool. But there are parts of the system that don't use it; they might use another wrapper around MessagingService to return to that another pool (maybe operations with the Metastorage should work like this). So, it's not the system-wide pool, its usage is still restricted.

modules/network/src/main/java/org/apache/ignite/internal/network/DefaultMessagingService.java
```diff
@@ -337,82 +337,167 @@ private List<ClassDescriptorMessage> prepareMarshal(NetworkMessage msg) throws E
...
+     * handlers (second and so on) on executors chosen by their choosers.
+     */
+    private void handleStartingWithFirstHandler(
+            NetworkMessage payload,
+            @Nullable Long correlationId,
+            InNetworkObject obj,
+            HandlerContext firstHandlerContext,
+            Iterator<HandlerContext> remainingContexts
+    ) {
+        if (senderIdIsStale(obj)) {
```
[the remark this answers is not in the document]
rpuch: These 2 checks are made in different threads. First one is to skip handling as early as possible. But after that a task is put to an executor, and who knows how much time it will sit there awaiting for execution. During this time it might happen that the sender has already become stale, hence second check.

Do you think second check is redundant?

modules/network/src/main/java/org/apache/ignite/internal/network/DefaultMessagingService.java
rpuch: Fixed

vldpyatkov (approved):

PR #3307, opened 2024-02-28
IGNITE-21525 Java Client: Clean up IEP-54 leftovers

ptupitsyn:
* Pass key column indices from `SchemaDescriptor` to the client
* Refactor the logic that relies on "key columns come first" design, support any key column order
* Fix colocation hash calculation to use `keyIndex`


isapego (approved):

modules/client/src/main/java/org/apache/ignite/internal/client/table/ClientSchema.java
```diff
@@ -63,34 +66,64 @@ public class ClientSchema {
...
+        this.keyColumns = keyColumnCount == 0 ? EMPTY_COLUMNS : new ClientColumn[keyColumnCount];
+        this.colocationColumns = colocationColumnCount == 0 ? keyColumns : new ClientColumn[colocationColumnCount];
+        this.valColumns = valColumnCount == 0 ? EMPTY_COLUMNS : new ClientColumn[valColumnCount];
+
+        for (var col : columns) {
+            if (col.key()) {
+                assert this.keyColumns[col.keyIndex()] == null : "Duplicate key index: name=" + col.name() + ", keyIndex=" + col.keyIndex()
+                        + ", other.name=" + this.keyColumns[col.keyIndex()].name();
+
+                assert col.keyIndex() == 0 || this.keyColumns[col.keyIndex() - 1] != null : "Key index is out of order: name="
```
korlov42: This assertion doesn't look valid. The order of key columns may not match the order of columns in the table declaration, and this is perfectly valid schema. I would say, the proper assertion here is that after we processed all the columns, neither of `keyColumns` and `valColumns` contains `null`'s

korlov42 (approved):
The patch looks good to me except for the assertion in constructor of ClientSchema. I approved this patch in advance assuming you will agree and will rework this part :) Otherwise let's sort out thing first in the corresponding thread

modules/client/src/main/java/org/apache/ignite/internal/client/table/ClientSchema.java
ptupitsyn: Fixed.

PR #3309, opened 2024-02-28
IGNITE-21331 Sql. Fix deadlock caused by sorting on distributed correlated path

korlov42:
…

### The Review Checklist

### Notes


modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/planner/ColocatedSortAggregatePlannerTest.java
```diff
@@ -293,47 +295,17 @@ public void noSortAppendingWithCorrectCollation() throws Exception {
...
-                                                .and(input(isTableScan("TEST")))
-                                        ))
-                                ))
-                        ))
-                ),
-                disableRules
-        );
+        RuntimeException e = assertThrows(RuntimeException.class,
+                () -> assertPlan(TestCase.CASE_17, isInstanceOf(IgniteRel.class), disableRules));
+        assertThat(e.getMessage(), containsString("There are not enough rules to produce a node with desired properties"));
```
AMashenkov, line 300: Maybe, it would be better to disable tests instead of rewriting their code?

korlov42: I thought about it, but decided to go with changing the test. First, I want to make sure such plan is impossible, so we need such test anyway. Second, this is not just side effect of my change, but rather intended result. When (or most probably "if") we fix problem with correlated sorted exchange, there is no guarantee the resulting plan will be the same.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/benchmark/TpchBenchmark.java
```diff
@@ -52,10 +55,26 @@
...
-            "customer", "lineitem", "nation", "orders",
-            "part", "partsupp", "region", "supplier"
-    };
+    /*
+        Minimal configuration of this benchmark requires specifying pathToDataset. Latest known location
+        of dataset is https://github.com/cmu-db/benchbase/tree/main/data/tpch-sf0.01 for scale factor 0.01
+        and https://github.com/cmu-db/benchbase/tree/main/data/tpch-sf0.1 for scale factor 0.1. Dataset
+        is set of CSV files with name `{$tableName}.tbl` per each table and character `|` as separator.
+
+        By default, cluster's work directory will be created as a temporal folder. This implies,
```
zstan, line 64:
```suggestion
        By default, cluster's work directory will be created as a temporary folder. This implies,
```

zstan (approved):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/benchmark/TpchBenchmark.java
korlov42: fixed, thanks!

AMashenkov (approved):

PR #3318, opened 2024-02-28
IGNITE-21311: Sql. Partition pruning. Introduce pruning for correlated scans

lowka:
WIP

---

### The Review Checklist

### Notes


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/externalize/RelJsonReader.java
```diff
@@ -84,6 +86,21 @@ public static <T extends RelNode> T fromJson(BaseQueryContext ctx, String json)
         return (T) reader.read(json);
     }
 
+    /** Creates a {@link RexNode} from the given json. */
+    public static RexNode fromExprJson(byte[] json) {
```
korlov42, line 90: why `byte[]` ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/pruning/PartitionPrunerImpl.java
```diff
@@ -71,6 +71,13 @@ public List<MappedFragment> apply(
                 continue;
             }
 
+            // Do not update colocation groups, in case of correlated fragment,
+            // because partitions for such fragments can be removed only at runtime.
+            if (fragment.correlated()) {
```
korlov42, line 76: this flag will be `false` in case of colocated join

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/externalize/RelJsonReader.java
lowka: `fromExprJson` uses bytes because `toJsonExpr` converts expressions to bytes[].
And `toJsonExpr` returns an array of bytes, because serialised expressions are then immediately written to output stream.

An alternative approach would be to convert expression into a string and then use methods from data stream than convert string into bytes.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/pruning/PartitionPrunerImpl.java
lowka: Thanks. Replaced with `correlated variables` check.

korlov42 (approved):

ygerzhedovich (approved):

PR #3326, opened 2024-03-01
IGNITE-21640 Exchange with ClusterNode instances on handshake

rpuch:
https://issues.apache.org/jira/browse/IGNITE-21640

### DIFF
--- a/modules/network/src/main/java/org/apache/ignite/internal/network/message/ClusterNodeMessage.java
+++ b/modules/network/src/main/java/org/apache/ignite/internal/network/message/ClusterNodeMessage.java
@@ -15,16 +15,31 @@
  * limitations under the License.
  */
 
-package org.apache.ignite.internal.network.recovery.message;
+package org.apache.ignite.internal.network.message;
 
+import static org.apache.ignite.internal.network.NetworkMessageTypes.CLUSTER_NODE_MESSAGE;
+
+import java.io.Serializable;
+import org.apache.ignite.internal.network.ClusterNodeImpl;
 import org.apache.ignite.internal.network.NetworkMessage;
+import org.apache.ignite.internal.network.annotations.Transferable;
+import org.apache.ignite.network.ClusterNode;
+import org.apache.ignite.network.NetworkAddress;
 
 /**
- * Message that does not need an acknowledgement from the remote node.
+ * {@link ClusterNode} as a network message class.
  */
-public interface InternalMessage extends NetworkMessage {
-    @Override
-    default boolean needAck() {
-        return false;
+@Transferable(CLUSTER_NODE_MESSAGE)
+public interface ClusterNodeMessage extends NetworkMessage, Serializable {
+    String id();
+
+    String name();
+
+    String host();
+
+    int port();
+
+    default ClusterNode asClusterNode() {
+        return new ClusterNodeImpl(id(), name(), new NetworkAddress(host(), port()));
     }
 }
--- a/modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
+++ b/modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
@@ -32,6 +32,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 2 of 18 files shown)


modules/network/src/main/java/org/apache/ignite/internal/network/message/ClusterNodeMessage.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.ignite.internal.network.NetworkMessage;
+import org.apache.ignite.internal.network.annotations.Transferable;
+import org.apache.ignite.network.ClusterNode;
+import org.apache.ignite.network.NetworkAddress;
+
+/**
+ * {@link ClusterNode} as a network message class.
+ */
+@Transferable(CLUSTER_NODE_MESSAGE)
+public interface ClusterNodeMessage extends NetworkMessage, Serializable {
```
sashapolo, line 33: We already have a `ClusterNodeMessage`, can we reuse it?

modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
```diff
@@ -105,8 +107,7 @@ public class ConnectionManager implements ChannelCreationListener {
     /** Node consistent id. */
     private final String consistentId;
 
-    /** Node launch id. As opposed to {@link #consistentId}, this identifier changes between restarts. */
-    private final UUID launchId;
+    private volatile @Nullable ClusterNode localNode;
```
sashapolo, line 110: Why is this field volatile now?

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
```diff
@@ -237,17 +230,16 @@ private void onHandshakeStartMessage(HandshakeStartMessage handshakeStartMessage
             return;
         }
 
-        this.remoteLaunchId = handshakeStartMessage.launchId();
-        this.remoteConsistentId = handshakeStartMessage.consistentId();
+        this.remoteNode = handshakeStartMessage.serverNode().asClusterNode();
 
-        ChannelKey channelKey = new ChannelKey(remoteConsistentId, remoteLaunchId, connectionId);
+        ChannelKey channelKey = new ChannelKey(remoteNode.name(), UUID.fromString(remoteNode.id()), connectionId);
```
sashapolo, line 235: Should we pass the cluster node to all such places?

modules/network/src/main/java/org/apache/ignite/internal/network/message/ClusterNodeMessage.java
rpuch: It's in another module and it is specific to the CMG: take a look at its 3 last methods. I also considered a possibility to create a base interface in the network module and extend it in the CMG modules with those 3 methods, but this does not seem to give a lot of gain (it allows us to reduce duplication by removing 4 lines defining id/name/host/port), but it introduces a dependency which usually bites afterwards. That's why it seemed to me to just copy the interface and remove the 3 last methods from it.

modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
rpuch: It is not final. I'm not sure we have a happens-before between the startup of `ConnectionManager` and establishment of connections. Also, this field is used rarely (each time when a connection is established), so the performance penalty should be negligible.

modules/network/src/main/java/org/apache/ignite/internal/network/recovery/RecoveryClientHandshakeManager.java
rpuch: `ChannelKey` is defined as a triple `<consistentId, launchId, channelIndex>`. According to the `ClusterNodeImpl` code, it considers 2 nodes equal if their name and address match. The ID is not even taken into consideration (which is weird, by the way). So with the current code replacement of `<node, channelIndex>` will just break everything.

Even if it did not break it (due to having a compatible equals semantics), it seems more straightforward and safer to list the components explicitly.

modules/network/src/main/java/org/apache/ignite/internal/network/message/ClusterNodeMessage.java
sashapolo: Those three last methods are nullable, can we simply pass nulls?

modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
sashapolo: My question was about why is this field no longer final)

modules/network/src/main/java/org/apache/ignite/internal/network/message/ClusterNodeMessage.java
rpuch: This looks like an abuse, to be honest. And that message is in another module.

modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
rpuch: Before building that ClusterNode, we need to start the ConnectionManager to obtain a bound server socket from which we take local address, and then apply some transformations to it. The starting is made in the `start()` method. So the field cannot be final.

PR #3336, opened 2024-03-01
IGNITE-21373 Sql. Get rid of the Ignite prefix in SQL operators

xtern:
https://issues.apache.org/jira/browse/IGNITE-21373

Overrides `getRelTypeName` for all implemented `RelNode` descendants in the project.

Unfortunately, it is useless to add a single implementation of `getRelTypeName()` to the `IgniteRel` interface, because `AbstractRelNode` (Calcite) is lower in the hierarchy and contains a different implementation.

diagram

### DIFF
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/AbstractIgniteJoin.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/AbstractIgniteJoin.java
@@ -292,4 +292,8 @@ public abstract class AbstractIgniteJoin extends Join implements TraitsAwareIgni
                 (left2Right ? right : left).getRowType().getFieldCount()
         );
     }
+
+    @Override public String getRelTypeName() {
+        return Commons.trimRelTypeName(super.getRelTypeName());
+    }
 }
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/AbstractIgniteJoin.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/AbstractIgniteJoin.java
@@ -293,6 +293,7 @@ public abstract class AbstractIgniteJoin extends Join implements TraitsAwareIgni
         );
     }
 
+    /** {@inheritDoc} */
     @Override public String getRelTypeName() {
         return Commons.makeRelTypeName(getClass());
     }
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/AbstractIgniteJoin.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/AbstractIgniteJoin.java
@@ -294,6 +294,6 @@ public abstract class AbstractIgniteJoin extends Join implements TraitsAwareIgni
     }
 
     @Override public String getRelTypeName() {
-        return Commons.trimRelTypeName(super.getRelTypeName());
+        return Commons.makeRelTypeName(getClass());
     }
 }
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/AbstractIgniteJoin.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/AbstractIgniteJoin.java
@@ -292,9 +292,4 @@ public abstract class AbstractIgniteJoin extends Join implements TraitsAwareIgni
                 (left2Right ? right : left).getRowType().getFieldCount()
         );
     }
-
-    /** {@inheritDoc} */
-    @Override public String getRelTypeName() {
-        return Commons.makeRelTypeName(getClass());
-    }
 }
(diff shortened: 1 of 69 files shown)


lowka:
Could you also fix `FragmentPrinter` used in mapping tests to use `getRelTypeName` as well? 

> LINE 256: output.writeString(rel.getClass().getSimpleName());

xtern:
> Could you also fix `FragmentPrinter` used in mapping tests to use `getRelTypeName` as well?
> 
> > LINE 256: output.writeString(rel.getClass().getSimpleName());

Very good point, thanks, fixed the fragment printer and related tests.

lowka (approved):

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/util/CommonsTest.java
```diff
@@ -86,4 +94,10 @@ private static void assertProjectionMapping(ImmutableIntList source, ImmutableIn
 
         assertEquals(expected, Mappings.apply(mapping, source));
     }
+
+    private static class IgniteN {}
+
+    private static class Ignite {}
+
+    private static class Ignit {}
```
AMashenkov, line 102:
```suggestion
    private static class Ignit {}
    
        private static class NonIgniteRel {}
```

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/util/CommonsTest.java
```diff
@@ -74,6 +74,14 @@ public void testArrayToMap() {
         assertEquals(vals, Commons.arrayToMap(new Object[]{1, null, "3"}));
     }
 
+    @Test
+    public void testMakeRelTypeName() {
+        assertEquals("N", Commons.makeRelTypeName(IgniteN.class));
+        assertEquals("Ignite", Commons.makeRelTypeName(Ignite.class));
+        assertEquals("Ignit", Commons.makeRelTypeName(Ignit.class));
```
AMashenkov, line 81:
```suggestion
        assertEquals("Ignit", Commons.makeRelTypeName(Ignit.class));
        assertEquals("NonIgniteRel", Commons.makeRelTypeName(NonIgniteRel.class));
```

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/AbstractIgniteJoin.java
```diff
@@ -292,4 +292,8 @@ protected Mappings.TargetMapping buildTransposeMapping(boolean left2Right) {
                 (left2Right ? right : left).getRowType().getFieldCount()
         );
     }
+
+    @Override public String getRelTypeName() {
```
AMashenkov, line 296: Why don't have a default implementation in `IgniteRel` interface?

xtern: Because all Ignite RelNodes inherit calcite `AbstractRelNode` (abstract class), which has a different `getRelTypeName()` implementation, therefore the default implementation from `IgniteRel` will not be used.

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/util/CommonsTest.java
xtern: Done, thanks

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/AbstractIgniteJoin.java
korlov42: Does it make sense to introduce abstract class `IgniteSingleRel` with overriden `getRelTypeName()` method?

korlov42 (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/AbstractIgniteJoin.java
AMashenkov: >   // It is not recommended to override this method, but sub-classes can do it at their own risk.
>   @Override public String getRelTypeName() {

😃

xtern: > However, the better way is to use string constants

I missed this note, personally I agree, I updated the PR to use string constants :thinking: .

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/Commons.java
```diff
@@ -831,4 +831,13 @@ public static SqlQueryType getQueryType(SqlNode sqlNode) {
                 return null;
         }
     }
+
+    /** Removes the 'Ignite' prefix from the provided string. */
+    public static String removeIgnitePrefix(String str) {
```
AMashenkov, line 836: Please, check if this method is unused.

AMashenkov (approved):

AMashenkov:
Overall PR looks good.
Let's add an ArchUnit test that checks all non-abstract classes, which are descendants of IgniteRel, overrides the `getRelTypeName` method.

PR #3348, opened 2024-03-04
IGNITE-21540 Handle lock exception for transaction operations

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-21540

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
@@ -29,6 +29,7 @@ import static org.apache.ignite.internal.table.distributed.storage.RowBatch.allR
 import static org.apache.ignite.internal.util.CompletableFutures.emptyListCompletedFuture;
 import static org.apache.ignite.internal.util.CompletableFutures.nullCompletedFuture;
 import static org.apache.ignite.internal.util.ExceptionUtils.withCause;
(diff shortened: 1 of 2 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -121,6 +123,8 @@ public class InternalTableImpl implements InternalTable {
     /** Primary replica await timeout. */
     public static final int AWAIT_PRIMARY_REPLICA_TIMEOUT = 30;
 
+    private Long implicitTransactionTimeout = IgniteSystemProperties.getLong("IGNITE_IMPLICIT_TRANSACTION_TIMEOUT", 3_000L);
```
denis-chudov, line 126: please add description what is this

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -298,21 +320,55 @@ private <R> CompletableFuture<R> enlistInTx(
...
+                    if (isRestartTransactionPossible(e) && coarseCurrentTimeMillis() - ts < implicitTransactionTimeout) {
+                        return enlistInTx(row, null, fac, noWriteChecker, ts);
+                    }
+                }
+
+                throw wrapReplicationException(e);
+            }
+
+            return completedFuture(r);
+        }).thenCompose(x -> x);
```
denis-chudov, line 349: looks like code duplication, seems it deserves a dedicated method

PR #3350, opened 2024-03-05
IGNITE-21552 Sql. Add hybrid clock propagation during query execution

xtern:
https://issues.apache.org/jira/browse/IGNITE-21552

Added hybrid timestamp propagation:
* from root node to leaf nodes (with `QueryStartRequest` message).
* from leaf nodes to root node (with `QueryBatchMessage` message).


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExchangeServiceImpl.java
```diff
@@ -162,6 +172,10 @@ private void onMessage(String nodeName, QueryBatchRequestMessage msg) {
     }
 
     private void onMessage(String nodeName, QueryBatchMessage msg) {
+        if (!nodeName.equals(localNodeName)) {
+            clock.update(msg.timestamp());
+        }
```
AMashenkov, line 177: I think this should be done at lower-level, e.g. somewhere in Messaging sub-system.
Shortly, a HLC (or similar) should register a hook in MessageService for that purpose, instead of shifting a responsibility to the other components. 

Let's add a TODO and create a ticket to fix this.

AMashenkov (approved):

korlov42 (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExchangeServiceImpl.java
xtern: Ticket created https://issues.apache.org/jira/browse/IGNITE-21709

modules/arch-test/src/test/java/org/apache/ignite/internal/SqlEngineModuleArchTest.java
```diff
@@ -48,10 +48,10 @@ public class SqlEngineModuleArchTest {
             .should(new ArchCondition<>("override org.apache.calcite.rel.RelNode.getRelTypeName") {
                 @Override
                 public void check(JavaClass javaClass, ConditionEvents conditionEvents) {
-                    boolean overrided = javaClass.getMethods().stream()
-                            .anyMatch(it -> "getRelTypeName".equals(it.getName()));
+                    boolean methodNotImplemented = javaClass.getMethods().stream()
+                            .noneMatch(it -> "getRelTypeName".equals(it.getName()));
 
-                    if (!overrided) {
+                    if (methodNotImplemented) {
```
AMashenkov, line 54: If someone will rollback current PR due to any reason, this fix will be lost as well.
You can left this code in patch/comment to the followup ticket.

xtern: Thanks, changes reverted.

korlov42 (approved):

PR #3353, opened 2024-03-05
IGNITE-21626 Make each component of the Inbound pool striped

rpuch:
https://issues.apache.org/jira/browse/IGNITE-21626


modules/workers/src/main/java/org/apache/ignite/internal/worker/CriticalStripedThreadPoolExecutor.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.ignite.internal.thread.StripedThreadPoolExecutor;
+
+/**
+ * Same as {@link StripedThreadPoolExecutor}, but each stripe is a critical worker monitored for being blocked.
+ */
+public class CriticalStripedThreadPoolExecutor extends AbstractStripedThreadPoolExecutor<ExecutorService> implements StripedExecutor {
+    /**
+     * Create a blockage-monitored striped thread pool.
+     *
+     * @param concurrencyLvl          Concurrency level.
```
sashapolo, line 40: Is this copied from somewhere? We don't use parameters alignment in javadocs

sashapolo (approved):

PR #3375, opened 2024-03-07
IGNITE-21634 Fixed NPE in HeapLockManager

denis-chudov:
https://issues.apache.org/jira/browse/IGNITE-21634


alievmirza (approved):

ascherbakoff (approved):

modules/transactions/src/integrationTest/java/org/apache/ignite/internal/tx/readonly/ItMultipleLocksTest.java
```diff
@@ -0,0 +1,54 @@
+/*
```
ascherbakoff, line 1: Wrong package.

ascherbakoff (approved):

modules/transactions/src/integrationTest/java/org/apache/ignite/internal/tx/readonly/ItMultipleLocksTest.java
denis-chudov: fixed

PR #3376, opened 2024-03-07
IGNITE-21088 Impossible to restart node with json config

valepakh:
https://issues.apache.org/jira/browse/IGNITE-21088

Force HOCON format for the node config file.


modules/runner/src/test/java/org/apache/ignite/internal/configuration/storage/LocalFileConfigurationStorageTest.java
```diff
@@ -507,6 +508,28 @@ void innerNodeWithPartialContent() throws Exception {
         assertThat(storage.readDataOnRecovery().get().values(), aMapWithSize(1));
     }
 
+    /** File content is parsed using HOCON format regardless of the file extension. */
+    @Test
+    void hoconContentInJsonFile() throws IOException {
+        // Given
+        String fileContent = "top {\n"
+                + "    namedList=[\n"
```
PakhomovAlexander, line 516: Could you please add \" to the identifiers as it is described in the ticket?

PR #3385, opened 2024-03-08
IGNITE-20870 Explain why snapshots lock is not taken in PartitionReplicaListener

rpuch:
https://issues.apache.org/jira/browse/IGNITE-20870

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
@@ -184,6 +184,8 @@ public class PartitionListener implements RaftGroupListener, BeforeApplyHandler 
                             + ", mvAppliedIndex=" + storage.lastAppliedIndex()
                             + ", txStateAppliedIndex=" + txStateStorage.lastAppliedIndex() + "]";
 
+            Serializable result = null;
+
             // NB: Make sure that ANY command we accept here updates lastAppliedIndex+term info in one of the underlying
             // storages!
             // Otherwise, a gap between lastAppliedIndex from the point of view of JRaft and our storage might appear.
@@ -194,8 +196,6 @@ public class PartitionListener implements RaftGroupListener, BeforeApplyHandler 
             // repeat same thing over and over again.
 
             storage.acquirePartitionSnapshotsReadLock();
-
-            Serializable result = null;
 
             try {
                 if (command instanceof UpdateCommand) {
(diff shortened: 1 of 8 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -231,6 +229,10 @@ public void onWrite(Iterator<CommandClosure<WriteCommand>> iterator) {
                 storage.releasePartitionSnapshotsReadLock();
             }
 
+            // Completing the closure out of the partition snapshots lock to reduce possibility of deadlocks as it might
```
sanpwc, line 232: I didn't get this point. Why it's possible to have a deadlock here?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1765,6 +1760,21 @@ private CompletableFuture<Void> applyWriteIntentSwitchCommand(
...
+
+        try {
+            storageUpdateHandler.switchWriteIntents(
+                    transactionId,
+                    commit,
+                    commitTimestamp,
+                    indexIdsAtRwTxBeginTs(transactionId)
+            );
+        } finally {
+            storageUpdateHandler.releasePartitionSnapshotsReadLock();
```
sanpwc, line 1774: I'm not sure that we should use snapshot lock within partition replica listener. I've scheduled a meeting to discuss this.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2665,37 +2686,11 @@ private CompletableFuture<CompletableFuture<?>> applyUpdateAllCommand(
...
-                                cmd.tablePartitionId().asTablePartitionId(),
-                                true,
-                                null,
-                                null,
-                                indexIdsAtRwTxBeginTs(transactionId)
-                        );
-
-                        updateTrackerIgnoringTrackerClosedException(safeTime, cmd.safeTime());
-                    }
+                    handleUpdateAllUnderLocksOnSnapshotsAndSafeTime(transactionId, cmd);
```
sanpwc, line 2689: I don't think that all that private methods actually increase readability. Matter of taste though.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
rpuch: I was able to get into a deadlock with a buggy implementation of `TestPartitionDataStorage` (it was taking an exclusive lock instead of a shared one). But even with this bug fixed, it seems that everything that can be done out of a lock should be done out of it. Especially if it can trigger some code that's hard to imagine right now.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
rpuch: They decrease duplication, hence they decrease a change for an error

sanpwc (approved):

PR #3390, opened 2024-03-09
IGNITE-21719 Add optional syntax highlight and multiline SQL.

PakhomovAlexander:
The tree-sitter parser is used for syntax highlighting for sql mode and for HOCON output (as JSON). Now only base SQL is supported. It means some Ignite-specific keywords are not highlighted (ENGINE, for example). It is planned to write our own Ignite-SQL parser in the future.

To make it event more joyful for the user to write some SQL in CLI I've also added multiline support. With this feature the "enter" does not execute the command but moves the cursor to the new line. ";" should be put at the end of the command to execute it.

Both syntax-highlighting and multiline support can be enabled or disabled (default is disabled) via cli
configuration keys.

https://issues.apache.org/jira/browse/IGNITE-21719


PakhomovAlexander:
Screenshot 2024-03-09 at 17 59 49

PakhomovAlexander:
Screenshot 2024-03-09 at 18 00 12

PR #3393, opened 2024-03-11
IGNITE-21729 Prevent threads from being hijacked via async cursors in KV/Record view APIs

rpuch:
https://issues.apache.org/jira/browse/IGNITE-21729

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/thread/PublicApiThreading.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/thread/PublicApiThreading.java
@@ -51,16 +51,16 @@ public class PublicApiThreading {
      * @return Call result.
      */
     public static <T> T doInternalCall(Supplier<T> call) {
-        boolean wasInInternalCallBefore = inInternalCall();
+        if (inInternalCall()) {
+            return call.get();
+        }
 
         startInternalCall();
 
         try {
             return call.get();
         } finally {
-            if (!wasInInternalCallBefore) {
-                endInternalCall();
-            }
+            endInternalCall();
         }
     }
 
--- a/modules/core/src/main/java/org/apache/ignite/internal/thread/PublicApiThreading.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/thread/PublicApiThreading.java
@@ -17,11 +17,14 @@
 
 package org.apache.ignite.internal.thread;
 
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ForkJoinPool;
 import java.util.function.Supplier;
 
 /**
- * Logic related to the threading concern of public APIs: it allows to raise/clear/check a thread-local flag which gives us ability to
- * see whether a public API call actually comes from an Ignite internal code and not from an actual user.
+ * Logic related to the threading concern of public APIs: it allows to prevent Ignite thread hijack by the user's code and provides
+ * mechanisms supporting this protection.
  */
 public class PublicApiThreading {
     private static final ThreadLocal<Boolean> INTERNAL_CALL = new ThreadLocal<>();
@@ -48,12 +51,16 @@ public class PublicApiThreading {
      * @return Call result.
      */
     public static <T> T doInternalCall(Supplier<T> call) {
+        boolean wasInInternalCallBefore = inInternalCall();
+
         startInternalCall();
 
         try {
             return call.get();
         } finally {
-            endInternalCall();
+            if (!wasInInternalCallBefore) {
+                endInternalCall();
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


modules/core/src/main/java/org/apache/ignite/internal/thread/PublicApiThreading.java
```diff
@@ -48,12 +51,16 @@ public static void endInternalCall() {
      * @return Call result.
      */
     public static <T> T doInternalCall(Supplier<T> call) {
+        boolean wasInInternalCallBefore = inInternalCall();
+
         startInternalCall();
```
sashapolo, line 56: Do we need to check the `wasInInternalCallBefore` flag? I understand that this method is idempotent, but it will be more consistent. Maybe we should even avoid the try-finally schenanigans if we are already in an internal call.

modules/table/src/integrationTest/java/org/apache/ignite/internal/threading/ItKvRecordApiThreadingTest.java
```diff
@@ -84,13 +104,24 @@ private static Table testTable() {
...
     @SuppressWarnings("rawtypes")
     @CartesianTest
     void keyValueViewFuturesCompleteInContinuationsPool(
             @Enum KeyValueViewAsyncOperation operation,
             @Enum KeyValueViewKind kind
     ) {
-        assumeTrue(kind.supportsGetNullable() || !operation.isGetNullable());
+        assumeTrue(
+                kind.supportsGetNullable() || !operation.isGetNullable(),
+                "getNullable() is not supported for this view, so skipping the test"
```
sashapolo, line 123:
```suggestion
                "getNullable() is not supported by views of type " + kind
```

modules/core/src/test/java/org/apache/ignite/internal/thread/PublicApiThreadingTest.java
```diff
@@ -0,0 +1,158 @@
...
+class PublicApiThreadingTest {
+    private final ExecutorService internalThreadPool = Executors.newSingleThreadExecutor(task -> new TestThread(task, true));
+    private final ExecutorService asyncContinuationExecutor = Executors.newSingleThreadExecutor(task -> new TestThread(task, false));
+
+    private static CompletableFuture<Thread> getCompletionThreadFuture(CompletableFuture<Void> publicFuture) {
+        return publicFuture.thenApply(unused -> Thread.currentThread());
+    }
+
+    private static CombinableMatcher<Object> internalPoolThread() {
+        return both(instanceOf(TestThread.class)).and(hasProperty("internal", is(true)));
```
sashapolo, line 51: Do we really need to use `hasProperty` here? Can we read the field value in a "normal" way?

modules/table/src/integrationTest/java/org/apache/ignite/internal/threading/ItKvRecordApiThreadingTest.java
```diff
@@ -207,6 +233,119 @@ private static RecordContext<Tuple> binaryRecordContext() {
         return new RecordContext<>(KEY_RECORD.toKeyTuple(), new Record(KEY, "one").toFullTuple(), new Record(KEY, "two").toFullTuple());
     }
 
+    @SuppressWarnings("rawtypes")
```
sashapolo, line 236: Why do you need to use raw types in these tests? I was able to get rid of them by simply using the wildcard type everywhere

modules/table/src/integrationTest/java/org/apache/ignite/internal/threading/ItKvRecordApiThreadingTest.java
```diff
@@ -207,6 +233,119 @@ private static RecordContext<Tuple> binaryRecordContext() {
...
+    @CartesianTest
+    void commonViewFuturesCompleteInContinuationsPool(@Enum CommonViewAsyncOperation operation, @Enum ViewKind kind) {
+        CriteriaQuerySource tableView = kind.criteriaQuerySource();
+
+        @SuppressWarnings("unchecked") CompletableFuture<Thread> completerFuture = forcingSwitchFromUserThread(
```
sashapolo, line 241: Same applies to this: if you replace some generics with a wildcard type, then no suppression is needed.

modules/core/src/test/java/org/apache/ignite/internal/thread/PublicApiThreadingTest.java
rpuch: Using a property allows to write an expression and return a matcher. Without this, I would have to deconstuct the object or write a custom matcher

modules/table/src/integrationTest/java/org/apache/ignite/internal/threading/ItKvRecordApiThreadingTest.java
rpuch: Changed it a bit differently

modules/table/src/integrationTest/java/org/apache/ignite/internal/threading/ItKvRecordApiThreadingTest.java
```diff
@@ -524,8 +511,8 @@ private enum CommonViewAsyncOperation {
             this.action = action;
         }
 
-        <K, V> CompletableFuture<?> executeOn(CriteriaQuerySource<Object> tableView) {
-            return action.apply(tableView);
+        <T> CompletableFuture<?> executeOn(CriteriaQuerySource<T> tableView) {
```
sashapolo, line 514: Looks like you don't need the generic parameter at all here and in `executeOn`

sashapolo (approved):

PR #3401, opened 2024-03-12
IGNITE-21348 Trigger the lease negotiation retry in case when the lease candidate is no more contained in assignments

denis-chudov:
https://issues.apache.org/jira/browse/IGNITE-21348


modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
```diff
@@ -368,15 +369,14 @@ private void updateLeaseBatchInternal() {
                         writeNewLease(grpId, lease, candidate, renewedLeases, toBeNegotiated);
 
                         continue;
+                    } else {
+                        agreement.checkValid(grpId, topologyTracker.currentTopologySnapshot(), assignments);
```
vldpyatkov, line 373: I would put it up just after getting an agreement with the additional conditions.
It makes logic simpler.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/negotiation/LeaseNegotiator.java
```diff
@@ -88,28 +91,37 @@ public void negotiate(Lease lease, boolean force) {
...
      *
      * @param groupId Replication group id.
      * @return Lease agreement.
      */
-    public LeaseAgreement negotiated(ReplicationGroupId groupId) {
-        LeaseAgreement agreement = leaseToNegotiate.getOrDefault(groupId, UNDEFINED_AGREEMENT);
+    public LeaseAgreement getAndRemoveIfReady(ReplicationGroupId groupId) {
+        LeaseAgreement[] res = new LeaseAgreement[1];
+
+        leaseToNegotiate.compute(groupId, (k, v) -> {
```
vldpyatkov, line 118: If a race is excluded here, I prefer to use plain read and conditinal remove.

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/negotiation/LeaseNegotiator.java
```diff
@@ -42,7 +41,11 @@ public class LeaseNegotiator {
 
     private static final PlacementDriverMessagesFactory PLACEMENT_DRIVER_MESSAGES_FACTORY = new PlacementDriverMessagesFactory();
 
-    /** Leases ready to accept. */
+    static LeaseGrantedMessageResponse NOT_ACCEPTED_RESPONSE = PLACEMENT_DRIVER_MESSAGES_FACTORY.leaseGrantedMessageResponse()
```
vldpyatkov, line 44: Why don't you use {@code null}, like it is in the undefined agreement?

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/TopologyTracker.java
```diff
@@ -142,15 +146,6 @@ public void onUpdate(LogicalTopologySnapshot topologySnap) {
...
             LOG.debug("Logical topology updated for placement driver [topologySnap={}]", topologySnap);
-
-            triggerToRenewLeases();
         }
     }
-
-    /**
-     * Triggers to renew leases forcibly. The method wakes up the monitor of {@link LeaseUpdater}.
-     */
-    private void triggerToRenewLeases() {
```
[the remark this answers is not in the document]
denis-chudov: why is it still needed?

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/LeaseUpdater.java
denis-chudov: we don't need to check the validity if the agreement is ready (accepted also means ready), that's why I put it to `else` block

modules/placement-driver/src/main/java/org/apache/ignite/internal/placementdriver/negotiation/LeaseNegotiator.java
denis-chudov: I wouldn't say so, this is a concurrent map which can be changed from different threads (lease updater and `whenComplete` of MS invoke

PR #3404, opened 2024-03-12
IGNITE-21565

sanpwc:


modules/table/src/integrationTest/java/org/apache/ignite/distributed/ReplicasSafeTimePropagationTest.java
```diff
@@ -153,6 +151,8 @@ public void testSafeTimeReorderingOnLeaderReElection() throws Exception {
 
         RaftGroupService anotherClient = aliveNode.get().raftClient;
 
+        assertThat(anotherClient.refreshLeader(), willCompleteSuccessfully());
```
alievmirza, line 154: Why do we need to add this assertion? The same question for the other rest

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
```diff
@@ -1414,6 +1414,7 @@ else if (this.leaderId.isEmpty()) {
     }
 
     private void becomeLeader() {
+        this.fsmCaller.onBeforeLeaderStart();
```
alievmirza, line 1417: This is not the only place where node could become a leader.
See `org.apache.ignite.raft.jraft.core.NodeImpl#handleTransferTimeout`

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/FSMCallerImpl.java
```diff
@@ -283,6 +283,12 @@ public boolean onLeaderStop(final Status status) {
         });
     }
 
+    @Override
+    public boolean onBeforeLeaderStart() {
+        doBeforeLeaderStart();
```
alievmirza, line 288: It seems to me that the fact, that we call `doBeforeLeaderStart()` directly, but not through the `enqueueTask` mechanism, which is done for all leader callbacks, could lead to to the linearisability problems. For example, when we transfer leadership, we call `org.apache.ignite.raft.jraft.core.NodeImpl#onLeaderStop` , but if this transferring timeouts, we return back the old leader, and we expect that `onBeforeLeaderStart` will be called, but because we call it directly, `onLeaderStop` could be invoked after `onBeforeLeaderStart`

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
alievmirza: If we want to call it here, we must call it after 
```
Requires.requireTrue(this.state == State.STATE_CANDIDATE, "Illegal state: " + this.state);
```

alievmirza (changes requested):
In general, I cannot fully agree with the proposed solution. From my point of view, adding new callback `onBeforeLeaderStart` in the way how it is done in PR could bring bugs. 

For the resolving the original problem I would try to not add new callback `onBeforeLeaderStart`, but try to fix it from the other way: try to not start processing new commands until we receive `onLeaderStart` (so we will have linearised `onLeaderStart` and `onBeforeApply` automatically). If you've already thought about such solution and found it unsuitable, please share why.

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ReplicasSafeTimePropagationTest.java
sanpwc: Because otherwise leader might not be elected -> maxObservableTimestamp == Long.MaxValue -> command will fail with SafeTimeReorderingException. In case of common logic that will trigger a retry which is expected, however within given particular test we do not test the retry, thus we should carefully prepare the preconditions, meaning leader election.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
sanpwc: > This is not the only place where node could become a leader.
See org.apache.ignite.raft.jraft.core.NodeImpl#handleTransferTimeout
Fixed.

> If we want to call it here, we must call it after
Requires.requireTrue(this.state == State.STATE_CANDIDATE, "Illegal state: " + this.state);
Nope, onBeforeLeaderStart is called when node is about to become a leader, however it's not guaranteed that it will happen, thus no preconditions expected.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/StateMachine.java
```diff
@@ -67,6 +67,12 @@ public interface StateMachine {
      */
     boolean onSnapshotLoad(final SnapshotReader reader);
 
+    /**
+     * Invoked when the belonging node is about to become the leader of the group. Default: Do nothing
+     */
+     void onBeforeLeaderStart();
+
```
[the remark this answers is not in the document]
sanpwc: Fixed.

PR #3423, opened 2024-03-15
IGNITE-21766 Avoid a race while handling messages in jraft

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-21766


modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/core/AppendEntriesRequestProcessor.java
```diff
@@ -322,41 +322,19 @@ void sendSequenceResponse(final String groupId, final PeerPair pair, final int s
...
-                    assert (node != null);
-                    peerCtx = new PeerRequestContext(groupId, pair, node.getRaftOptions().getMaxReplicatorInflightMsgs());
-
-                    peerCtx.executor = node.getOptions().getStripedExecutor().next();
-
-                    groupContexts.put(pair, peerCtx);
-                }
-            }
-        }
+            return groupContexts.computeIfAbsent(pair, peerPair -> {
```
sanpwc, line 329: In original code we've parsed the peer. Seems that given code is missing in proposed solution. Why?

sanpwc (approved):

PR #3432, opened 2024-03-18
IGNITE-21798 Add properties to configure number of RAFT threads

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-21798


modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
```diff
@@ -78,14 +79,16 @@ public class StripedDisruptor<T extends NodeIdAware> {
      * @param stripes Amount of stripes.
      * @param supportsBatches If {@code false}, this stripe will always pass {@code true} into
      *      {@link EventHandler#onEvent(Object, long, boolean)}. Otherwise, the data will be provided with batches.
+     * @param useYieldStrategy If the parameter is true, the yield strategy is to be used otherwise, the blocking strategy.
```
zstan, line 82:
```suggestion
     * @param useYieldStrategy If {@code true}, the yield strategy is to be used otherwise the blocking strategy.
```

denis-chudov (approved):

PR #3436, opened 2024-03-19
IGNITE-18617 Clear rw tx context and cleanup ready futures on tx finish

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-18617

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
@@ -1730,7 +1730,6 @@ public class PartitionReplicaListener implements ReplicaListener {
         List<CompletableFuture<?>> txUpdateFutures = new ArrayList<>();
         List<CompletableFuture<?>> txReadFutures = new ArrayList<>();
 
-        // TODO https://issues.apache.org/jira/browse/IGNITE-18617
         txCleanupReadyFutures.compute(txId, (id, txOps) -> {
             if (txOps == null) {
                 return null;
@@ -1746,7 +1745,7 @@ public class PartitionReplicaListener implements ReplicaListener {
 
             txOps.futures.clear();
 
-            return txOps;
+            return null;
         });
 
         return allOfFuturesExceptionIgnored(txUpdateFutures, commit, txId)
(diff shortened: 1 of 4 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1746,7 +1745,7 @@ private CompletableFuture<FuturesCleanupResult> awaitCleanupReadyFutures(UUID tx
 
             txOps.futures.clear();
 
-            return txOps;
+            return null;
```
sanpwc, line 1748: Do we use cleanupReadyFutures in order to handle the race between cleanup and adding new operations? Meaning that: 

- we need to await all pending operations prior to cleanup processing - all good here.
- we need to discard new operation after cleanup was done - do we use cleanupReadyFutures to handle such race?

sanpwc: Hmm, the question is still on the table.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1894,18 +1893,23 @@ private <T> CompletableFuture<T> appendTxCommand(UUID txId, RequestType cmdType,
...
+            // First check whether the transaction has already been finished.
+            // And complete cleanupReadyFut with exception if it is the case.
             TxStateMeta txStateMeta = txManager.stateMeta(txId);
 
             if (txStateMeta == null || isFinalState(txStateMeta.txState())) {
                 cleanupReadyFut.completeExceptionally(new Exception());
-            } else {
-                txOps.futures.computeIfAbsent(cmdType, type -> new ArrayList<>()).add(cleanupReadyFut);
+
+                return txOps;
```
sanpwc, line 1903: Why do we need to put txOps into map in case of exception?

sanpwc: Make sense.

sanpwc (approved):

PR #3439, opened 2024-03-19
IGNITE-21521 Fix update order in upsertAll and DataStreamer for a new key

ptupitsyn:
Use the same `RowId` when the same key occurs multiple times within a batch in `RW_UPSERT_ALL` handler in `PartitionReplicaListener` to apply the last update for a given key, not a random one.


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2250,9 +2256,12 @@ private CompletableFuture<ReplicaResult> processMultiEntryAction(ReadWriteMultiR
 
                         boolean insert = rowId == null;
 
-                        RowId rowId0 = insert ? new RowId(partId(), UUID.randomUUID()) : rowId;
+                        RowId rowId0 = insert
```
ascherbakoff, line 2259: Can we avoid taking locks for duplicate rows (in line 2275) ?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2250,9 +2256,12 @@ private CompletableFuture<ReplicaResult> processMultiEntryAction(ReadWriteMultiR
...
+                                ? newKeyMap.computeIfAbsent(pk.byteBuffer(), rowIdGen)
+                                : rowId;
 
                         if (lastCommitTime != null) {
+                            assert rowId != null : "lastCommitTime is not null, but rowId is null";
```
ascherbakoff, line 2264: This assert looks unrelated to bug. This is executed only for existing rows.

ptupitsyn: It's just to fix the IDE warning. Replaced it with `noinspection`.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
ptupitsyn: Good catch. Fixed to reuse the lock and skip previous updates for the same key.

ascherbakoff (approved):

PR #3447, opened 2024-03-20
IGNITE-20126 Handle 'index storage is under destruction' on read/write

rpuch:
https://issues.apache.org/jira/browse/IGNITE-20126


modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/AbstractTestIndexStorage.java
```diff
@@ -55,7 +56,7 @@ public int pendingCursors() {
 
     @Override
     public Cursor<RowId> get(BinaryTuple key) {
-        checkStorageClosedOrInProcessOfRebalance();
+        checkStorageClosedOrInProcessOfRebalance(true);
```
sashapolo, line 59: Shall we create a wrapper over index storages (e.g. `RebalanceAwareIndexStorage`), that will perform all these checks? It's a bit frustrating that every storage implementation has to check this

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/AbstractPageMemoryIndexStorage.java
```diff
@@ -108,7 +109,7 @@ protected AbstractPageMemoryIndexStorage(
 
     @Override
     public @Nullable RowId getNextRowIdToBuild() {
-        return busy(() -> {
+        return busyNonRead(() -> {
```
sashapolo, line 112: Why is this a `nonRead`?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -392,6 +395,23 @@ public PartitionReplicaListener(
         prepareIndexBuilderTxRwOperationTracker();
     }
 
+    private static void rethrowTranslated(Throwable ex) {
+        Throwable unwrapped = ex;
```
sashapolo, line 399: Can we use `ExceptionUtils.unwrapCause`?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -402,6 +422,12 @@ public CompletableFuture<ReplicaResult> invoke(ReplicaRequest request, String se
                     } else {
                         return new ReplicaResult(res, null);
                     }
+                })
+                .exceptionally(ex -> {
+                    rethrowTranslated(ex);
```
sashapolo, line 427: A neat approach is to actually make `rethrowTranslated` return an exception and throw it here instead. This way you won't have to write `return null` down below, for example `throw translateException(ex);`

modules/index/src/integrationTest/java/org/apache/ignite/internal/index/ItIndexAndIndexStorageDestructionTest.java
```diff
@@ -0,0 +1,157 @@
...
+    private int indexId(String indexName) {
+        CatalogIndexDescriptor indexDescriptor = node.catalogManager().aliveIndex(indexName, node.clock().nowLong());
+        assertThat(indexDescriptor, is(notNullValue()));
+
+        return indexDescriptor.id();
+    }
+
+    @Test
+    void insertTouchingDestroyedIndexStorageDoesNotFailWholeWrite() {
+        KeyValueView<Integer, String> view = node.tables().table(TABLE_NAME)
```
sashapolo, line 105: How do you check that this call touches a destroyed index?

modules/index/src/integrationTest/java/org/apache/ignite/internal/index/ItIndexAndIndexStorageDestructionTest.java
```diff
@@ -0,0 +1,157 @@
...
+        KeyValueView<Tuple, Tuple> view = node.tables().table(TABLE_NAME).keyValueView();
+
+        CriteriaException ex = assertThrows(CriteriaException.class, () -> {
+            try (Cursor<?> cursor = view.query(null, null, INDEX_NAME)) {
+                cursor.next();
+            }
+        });
+
+        assertThat(ex.code(), is(STALE_PLAN_ERR));
+        assertThat(ex.getMessage(), is("The plan is stale. Please retry."));
```
sashapolo, line 155: This error message looks strange. Who is it intended for?

modules/core/src/main/java/org/apache/ignite/internal/table/criteria/CriteriaExceptionMapperUtil.java
```diff
@@ -57,10 +57,6 @@ public static Throwable mapToPublicCriteriaException(Throwable origin) {
             return e;
         }
 
-        if (e instanceof SqlException) {
-            return new CriteriaException(INTERNAL_ERR, e);
```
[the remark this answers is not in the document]
rpuch: Yes, this breaks propagation of the real error code to the user

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/AbstractTestIndexStorage.java
rpuch: I like the idea, but right now I'm a bit time-constrained. Will it be ok with you if I create a Jira issue and we solve it later?

modules/index/src/integrationTest/java/org/apache/ignite/internal/index/ItIndexAndIndexStorageDestructionTest.java
rpuch: Line 105 doesn't touch it. Code inside `assertDoesNotThrow()` does. I wrote the test before doing the fix, the test failed due to an exception about a destroyed storage.

modules/index/src/integrationTest/java/org/apache/ignite/internal/index/ItIndexAndIndexStorageDestructionTest.java
rpuch: For the user who issued the query. What's strange in it?

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/AbstractPageMemoryIndexStorage.java
rpuch: Because it's not a read of index data, it's related to some metadata.

I renamed the methods, is this better?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/AbstractTestIndexStorage.java
sashapolo: yes

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/AbstractPageMemoryIndexStorage.java
sashapolo: Still quite confusing, please add javadocs at least

modules/index/src/integrationTest/java/org/apache/ignite/internal/index/ItIndexAndIndexStorageDestructionTest.java
sashapolo: > Code inside assertDoesNotThrow() does. I wrote the test before doing the fix, the test failed due to an exception about a destroyed storage.

I understand, but this looks like a fragile approach. I would expect that we would validate somehow, that the destroyed index is indeed accessed,

rpuch: The index is destroyed (I destroyed the storages explicitly). There are tests that make sure that a destroyed storage throws exceptions. Then I test the system via API. No exception flyies (and it flied before the fix). I don't see what else should be done.

modules/index/src/integrationTest/java/org/apache/ignite/internal/index/ItIndexAndIndexStorageDestructionTest.java
rpuch: Changed the error message. I will ask guys from the SQL team to take a look; if they are ok, I'll as the doc writers.

modules/storage-page-memory/src/main/java/org/apache/ignite/internal/storage/pagememory/index/AbstractPageMemoryIndexStorage.java
rpuch: Added javadocs

modules/index/src/integrationTest/java/org/apache/ignite/internal/index/ItIndexAndIndexStorageDestructionTest.java
rpuch: I added a unit test for `SortedIndexLocker`

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/impl/AbstractTestIndexStorage.java
rpuch: I filed a Jira issue: https://issues.apache.org/jira/browse/IGNITE-21825

sashapolo (approved):

vldpyatkov (approved):

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/ReadFromDestroyedIndexStorageException.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package org.apache.ignite.internal.storage;
+
+import static org.apache.ignite.lang.ErrorGroups.Storage.READING_FROM_ALREADY_DESTROYED_INDEX_ERR;
+
+/**
+ * Thrown if an attempt is made to read from an index storage that is under destruction or already destroyed.
+ */
+public class ReadFromDestroyedIndexStorageException extends StorageDestroyedException {
```
ygerzhedovich, line 25: What is different between this new Exception with new code and `StorageDestroyedException` with a new error code? Can we keep just single code and single Exception here?

modules/core/src/main/java/org/apache/ignite/internal/util/subscription/OrderedMergePublisher.java
```diff
@@ -208,8 +208,7 @@ private void updateError(Throwable throwable) {
                 if (current == null) {
                     next = throwable;
                 } else {
-                    next = new Throwable();
-                    next.addSuppressed(current);
+                    next = current;
```
ygerzhedovich, line 211: this will not work properly. Below you have CAS which check that currest still the same and set next as a new value. But you always use the same value for next and current ones - so, the CAS from second error will be always done even it have had concurrent modification.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/ReadFromDestroyedIndexStorageException.java
rpuch: There are 2 cases when a 'storage is destroyed' might be thrown, they need to be handled differently. That's why there are 2 exceptions with 2 code.

ygerzhedovich: Could you provide what will be different for handle these two situations? 
Also Index it is a storage also...

modules/core/src/main/java/org/apache/ignite/internal/util/subscription/OrderedMergePublisher.java
rpuch: Fixed, please take a look

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/ReadFromDestroyedIndexStorageException.java
rpuch: I removed the added codes as it really seems an overkill to have them just for this rare corner case. The use will still get an error message saying that there was a read from a destroyed index storage.

modules/core/src/main/java/org/apache/ignite/internal/util/subscription/OrderedMergePublisher.java
```diff
@@ -410,4 +405,33 @@ public void cancel() {
...
+        private final AtomicBoolean built = new AtomicBoolean(false);
+
+        private ErrorChain(Throwable error, @Nullable ErrorChain next) {
+            this.error = error;
+            this.next = next;
+        }
+
+        Throwable buildThrowable() {
+            if (!built.compareAndSet(false, true)) {
+                // Already built, so error already contains all subsequent exceptions attached.
```
ygerzhedovich, line 423: it's a false. You can return  not fully constructed `error`, due to flag amended before the `error` will be constructed.

rpuch: Fixed

ygerzhedovich (approved):

PR #3454, opened 2024-03-20
IGNITE-21795 Unconditionally update storage with proper raft index within PartitionListener

rpuch:
https://issues.apache.org/jira/browse/IGNITE-21795

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
@@ -276,9 +276,13 @@ public class PartitionListener implements RaftGroupListener, BeforeApplyHandler 
                         cmd.lastCommitTimestamp(),
                         indexIdsAtRwTxBeginTs(catalogService, txId, storage.tableId())
                 );
-            }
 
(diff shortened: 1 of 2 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -407,6 +415,10 @@ private void handleSafeTimeSyncCommand(SafeTimeSyncCommand cmd, long commandInde
 
         // We MUST bump information about last updated index+term.
         // See a comment in #onWrite() for explanation.
+        advanceLastAppliedIndexConsistently(commandIndex, commandTerm);
```
sanpwc, line 418: Seems that we also miss lastAppliedIndexUpdate in case of `handleFinishTxCommand`.

rpuch: We don't, the index gets updated by the `compareAndSet()` method internally. I wrote a test for that.

sanpwc: Yep, now I can. Missed that.

sanpwc (approved):

PR #3459, opened 2024-03-21
IGNITE-21815 Client handler: use QueryProcessor instead of IgniteSql

ptupitsyn:
Use internal `QueryProcessor` API in client handler instead of public `IgniteSql`:
* Simplify `HybridTimestamp` propagation
* Get rid of Session usage, reduce allocations and indirection
* Unblock [IGNITE-21669 Remove sessions from SQL API](https://issues.apache.org/jira/browse/IGNITE-21669).

https://issues.apache.org/jira/browse/IGNITE-21815


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/AsyncResultSetImpl.java
```diff
@@ -41,6 +41,7 @@
  * Asynchronous result set implementation.
  */
 public class AsyncResultSetImpl<T> implements AsyncResultSet<T> {
+    @Nullable
```
korlov42, line 44: according to our guidelines, [type-use annotations appear immediately before the annotated type](https://google.github.io/styleguide/javaguide.html#s4.8.5-annotations)

korlov42 (approved):
LGTM!

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/AsyncResultSetImpl.java
ptupitsyn: Fixed.

PR #3462, opened 2024-03-21
IGNITE-21540 Handle lock exception for transaction operations

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-21540


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -405,7 +470,21 @@ private <T> CompletableFuture<T> enlistInTx(
...
+
+                    if (isRestartTransactionPossible(e) && coarseCurrentTimeMillis() - ts < implicitTransactionTimeout) {
+                        return enlistInTx(keyRows, null, fac, reducer, noOpChecker, ts);
+                    }
+                }
+
+                throw wrapReplicationException(e);
+            }
+
+            return completedFuture(r);
```
denis-chudov, line 486: let's remove code duplication

denis-chudov (approved):

PR #3465, opened 2024-03-22
IGNITE-21810 Migrate cpp client/odbc build to Conan 2

dant3:
https://issues.apache.org/jira/browse/IGNITE-21810

I also updated DEVNOTES.md with instructions on how to use Conan 2, and gradle build so that packaging works with it.

### DIFF
--- a/modules/platforms/cpp/DEVNOTES.md
+++ b/modules/platforms/cpp/DEVNOTES.md
@@ -178,6 +178,22 @@ cmake -Bbuild -DCMAKE_BUILD_TYPE=RelWithDebInfo
 cmake --build build && cmake --build build --target install
 ```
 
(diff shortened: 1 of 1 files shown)


modules/platforms/cpp/CMakeLists.txt
```diff
@@ -23,7 +23,6 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 set(CMAKE_PROJECT_VERSION ${PROJECT_VERSION})
 
-option(ENABLE_CONAN "Use Conan package manager to get dependencies" ON)
```
ademakov, line 26: I don't understand, can we still build it without conan?

ademakov (approved):

PR #3486, opened 2024-03-26
IGNITE-21527 Table schema re-factored

isapego:
Refactor the logic that relies on "key columns come first" design, support any key column order (use `key_index` from the server schema).

https://issues.apache.org/jira/browse/IGNITE-21527


ptupitsyn (changes requested):
Let's add an integration test that creates a table with custom key column order and performs various operation on it:
```sql
create table test_key_col_order (val1 varchar, key1 integer, val2 bigin, key2 varchar, primary key(key2, key1)) 
```

ptupitsyn (approved):

PR #3493, opened 2024-03-26
IGNITE-21745 Optimize BigDecimal representation in BinaryTuple

ptupitsyn:
Add scale prefix (2 bytes) to decimal values in `BinaryTuple`.

**Problem**
Small values require a lot of bytes to store when `scale` is large:

```java
ByteBuffer bytes = new BinaryTupleBuilder(1).appendDecimal(new BigDecimal(1), Short.MAX_VALUE).build();
assertEquals(13610, bytes.limit());
```

* Such a value can't be read in .NET (and potentially other clients) because of type system limitation, even though it represents `1`
* SQL returns values with `MAX_DECIMAL_SCALE` as a result of formulas, for example `select (cast(10 as decimal(20, 10)) / ?)`

**Fix**
Strip trailing zeros and store value scale with a smaller unscaled value.


modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/BinaryTupleBuilder.java
```diff
@@ -315,8 +320,34 @@ public BinaryTupleBuilder appendNumber(BigInteger value) {
...
     public BinaryTupleBuilder appendDecimalNotNull(BigDecimal value, int scale) {
-        // TODO IGNITE-21745: Inefficient serialization, many bytes wasted to store small values when scale is big.
-        putBytes(value.setScale(scale, RoundingMode.HALF_UP).unscaledValue().toByteArray());
+        value = value.stripTrailingZeros();
+        int valScale = value.scale();
+
+        if (valScale < 0) {
+            value = value.setScale(0, RoundingMode.UNNECESSARY);
+            valScale = 0;
+        }
```
isapego, line 329: Why? And what about very big numbers? Like, `1E10000`?

ptupitsyn: Yes, this is wrong, not sure what I was thinking. As discussed in private, I've removed varlen logic and now we store signed scale as 2 bytes (see https://github.com/apache/ignite-3/blob/a7188adc70d6f3bf9189feba3b5c6878bd3a27ec/modules/catalog/src/main/java/org/apache/ignite/internal/catalog/commands/CatalogUtils.java#L113).

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/BinaryTupleBuilder.java
```diff
@@ -315,8 +315,19 @@ public BinaryTupleBuilder appendNumber(BigInteger value) {
...
-        putBytes(value.setScale(scale, RoundingMode.HALF_UP).unscaledValue().toByteArray());
+        if (value.scale() > scale) {
+            value = value.setScale(scale, RoundingMode.HALF_UP);
+        }
+
+        value = value.stripTrailingZeros();
+
+        // See CatalogUtils.MAX_DECIMAL_SCALE = Short.MAX_VALUE
+        assert value.scale() <= Short.MAX_VALUE : "Decimal scale is too large: " + value.scale() + " > " + Short.MAX_VALUE;
+        assert value.scale() >= Short.MIN_VALUE : "Decimal scale is too small: " + value.scale() + " < " + Short.MIN_VALUE;
```
isapego, line 326: Can this actually happen if user uses such values? Maybe, we should use exceptions here instead? Let's maybe add tests with such a values.

ptupitsyn: Agree. Replaced with exceptions, added test.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/BinaryTupleBuilder.java
```diff
@@ -315,8 +315,24 @@ public BinaryTupleBuilder appendNumber(BigInteger value) {
...
+        value = value.stripTrailingZeros();
+
+        // See CatalogUtils.MAX_DECIMAL_SCALE = Short.MAX_VALUE
+        if (value.scale() > Short.MAX_VALUE) {
+            throw new BinaryTupleFormatException("Decimal scale is too large: " + value.scale() + " > " + Short.MAX_VALUE);
+        }
+
+        if (value.scale() < Short.MIN_VALUE) {
+            throw new BinaryTupleFormatException("Decimal scale is too small: " + value.scale() + " < " + Short.MIN_VALUE);
+        }
```
ademakov, line 331: Can there be a situation when `scale()` fits the range before `stripTrailingZeros()` and doesn't after it?

ptupitsyn: Good catch, you are right. As the javadoc says, stripping the trailing zeros from `600.0 [value=6000, scale=1]` yields `6E2  [value=6, scale=-2]`.

We don't want to remove `stripTrailingZeros` completely - it produces a more compact representation. So I've added a scale check.

ptupitsyn: Added a test too (`BinaryTupleTest.maxDecimalScaleTest`) - it failed before the fix and now works.

ademakov (approved):

PR #3496, opened 2024-03-26
IGNITE-21761 Within commitPartition mark txnState with cleanup finished timestamp

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-21761


modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxCleanupRequestHandler.java
```diff
@@ -156,11 +170,98 @@ private NetworkMessage prepareResponse() {
...
+    private void processWriteIntentSwitchResponse(ReplicaResponse response) {
+        if (response == null) {
+            return;
+        }
+
+        Object result = response.result();
+
+        if (result instanceof WriteIntentSwitchReplicatedInfo) {
+            writeIntentSwitchReplicated((WriteIntentSwitchReplicatedInfo) result);
+        }
```
denis-chudov, line 219: assertion better suits here

denis-chudov (approved):

PR #3505, opened 2024-03-28
IGNITE-21287 Sql. LogicalOrToUnionRule may hang

zstan:
Benchmark                          (queryId)  Mode  Cnt   Score   Error  Units
TpchPrepareBenchmark.prepareQuery         19  avgt   10  29.258 ± 6.672  ms/op

### DIFF
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/RexUtils.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/RexUtils.java
@@ -182,8 +182,25 @@ public class RexUtils {
         return true;
     }
 
+    /** Try to transform expression into DNF form.
+     *
+     * @param rexBuilder Expression builder.
+     * @param node Expression to process.
+     * @param maxOrNodes Max OR nodes in output.
+     * @return DNF expression representation or {@code null} if limit is reached.
+     */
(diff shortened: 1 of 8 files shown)


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/RexUtils.java
```diff
@@ -176,6 +182,104 @@ public static boolean isIdentity(List<? extends RexNode> projects, RelDataType i
...
+
+        private RexNode and(Iterable<? extends RexNode> nodes) {
+            return composeConjunction(rexBuilder, nodes);
+        }
+
+        private RexNode or(Iterable<? extends RexNode> nodes) {
+            RexNode res = composeDisjunction(rexBuilder, nodes);
+            if (res instanceof RexCall) {
+                RexCall res0 = (RexCall) res;
+                if (res0.getOperands().size() > maxOrNodes) {
```
lowka, line 275: Is this condition the only difference between this code and RexUtils::DnfHelper?

zstan: yes, on planning phase with huge amount of params it takes more time than planning is available, thus planning timeout will raised

korlov42 (approved):

lowka (approved):

PR #3506, opened 2024-03-28
IGNITE-21853 Prevent thread hijacking via IgniteTransactions

rpuch:
https://issues.apache.org/jira/browse/IGNITE-21853


modules/transactions/src/integrationTest/java/org/apache/ignite/internal/tx/ItTransactionsApiThreadingTest.java
```diff
@@ -0,0 +1,165 @@
...
+    @Override
+    protected int initialNodes() {
+        return 1;
+    }
+
+    @BeforeAll
+    void createTable() {
+        sql("CREATE TABLE " + TABLE_NAME + " (id INT PRIMARY KEY, val VARCHAR)");
+
+        //noinspection AssignmentToStaticFieldFromInstanceMethod
```
ibessonov, line 61: I believe that we should add an extra space after `//`.
For some reason, I have a feeling that this particular check stopped working on TC, we should find out why (if it did).
Or we added `noinspection` into a white list?

ibessonov (approved):

modules/transactions/src/integrationTest/java/org/apache/ignite/internal/tx/ItTransactionsApiThreadingTest.java
rpuch: checkstyle-rules.xml:

```
    <!-- Comments should start with space -->
    <module name="TodoComment">
      <property name="id" value="CommentStartWithSpace"/>
      <property name="format" value="^(?!CHECKSTYLE:ON)(?!noinspection)(?!---)(?!===)([^\s\/*])"/>
      <message key="todo.match" value="Comment text should start with space."/>
    </module>
```

It seems that this '//noinspection' was explicitly whitelisted on March, 26.

PR #3513, opened 2024-03-28
IGNITE-21567 Sql. Fixed precision when converting from TIME/TIMESTAMP to VARCHAR

xtern:
https://issues.apache.org/jira/browse/IGNITE-15622

1. Fixed conversion of TIME and TIMESTAMP to VARCHAR to maintain precision (using built-in `DateTimeUtils` method which accepts precision).

https://issues.apache.org/jira/browse/IGNITE-21567

2. The same way fixed conversion from TIMESTAMP to TIMESTAMP_WITH_LOCAL_TIME_ZONE.

3. Some Sql Logic tests have been reworked a bit because the default precision for TIME is zero (according to the SQL spec).


AMashenkov (approved):

modules/sql-engine/src/integrationTest/sql/types/date/date_parsing.test_ignore
```diff
@@ -1,84 +1,35 @@
...
 PRAGMA enable_verification
 
 query I
-SELECT '1992-01-01'::DATE::VARCHAR == '1992-01-01'
+SELECT '30000-01-01'::DATE::VARCHAR = '30000-01-01'
 ----
 1
 
 query I
-SELECT '1992-09-20'::DATE::VARCHAR == '1992-09-20'
```
korlov42, line 15: it's better not to remove test cases from '_ignored' tests. The idea is to have single file with all the test cases. If you want to see only cases which failed, you can use any diff tool

modules/sql-engine/src/integrationTest/sql/types/time/test_time.test
```diff
@@ -1,32 +1,35 @@
...
-INSERT INTO times VALUES ('00:01:20'), ('20:08:10.998'), ('20:08:10.33'), ('20:08:10.001'), (NULL)
+INSERT INTO times VALUES ('00:01:20'), ('20:08:10.998'), ('20:08:10.33'), ('20:08:10.001')
+
+# TODO https://issues.apache.org/jira/browse/IGNITE-15123 Multi-tuple insert fails on validation
+statement ok
+INSERT INTO times VALUES (NULL)
 
 query T
-SELECT * FROM times
+SELECT * FROM times order by i
```
korlov42, line 19: instead of adding `ORDER BY` clause, it's better to use `rowsort` option to do comparison. Please refer to javadoc of `ItSqlLogicTest` for details

korlov42 (approved):

modules/sql-engine/src/integrationTest/sql/types/time/test_time.test
xtern: Fixed, thanks.

modules/sql-engine/src/integrationTest/sql/types/date/date_parsing.test_ignore
xtern: Fixed, thanks.

AMashenkov (approved):

modules/sql-engine/src/integrationTest/sql/types/date/test_date.test
```diff
@@ -0,0 +1,38 @@
...
+statement ok
+INSERT INTO dates VALUES ('1993-08-14')
+
+# TODO https://issues.apache.org/jira/browse/IGNITE-15123 Multi-tuple insert fails on validation
+statement ok
+INSERT INTO dates VALUES (NULL)
+
+# check that we can select dates
+query T rowsort
+SELECT * FROM dates
```
zstan, line 21: seems we need ordering definition if we wait defined ordered results ?

xtern: We have ordering on the test framework side here (`rowsort`), is that not enough?

modules/sql-engine/src/integrationTest/sql/types/time/time_parsing.test
```diff
@@ -0,0 +1,40 @@
...
+14:42:04.999
+
+query I
+SELECT '14:42:04.999'::TIME(6)::VARCHAR
+----
+14:42:04.999000
+
+# microseconds are not supported yet, result is rounded-up
+query I
+SELECT '14:42:04.9995'::TIME(6)
```
zstan, line 32: seems you need to remove this test too ?

modules/sql-engine/src/integrationTest/sql/types/time/time_parsing.test
```diff
@@ -0,0 +1,40 @@
...
+----
+14:42:05
+
+# zeros are not trimmed
+query I
+SELECT '14:42:04.9995'::TIME(6)::VARCHAR
```
zstan, line 38: and probably this ?

xtern: Yes, I missed those tests, fixed, thanks.

zstan (approved):

PR #3514, opened 2024-03-29
IGNITE-21875 SQL API cleanup - remove properties and reactive methods

ptupitsyn:
* Statement properties are not used by the engine - remove
* Reactive methods are not implemented - remove

https://issues.apache.org/jira/browse/IGNITE-21875


ygerzhedovich (approved):

modules/client/src/main/java/org/apache/ignite/internal/client/sql/ClientSql.java
```diff
@@ -253,20 +251,6 @@ public <T> CompletableFuture<AsyncResultSet<T>> executeAsync(
...
-    /** {@inheritDoc} */
-    @Override
-    public ReactiveResultSet executeReactive(@Nullable Transaction transaction, String query, @Nullable Object... arguments) {
-        // TODO IGNITE-17058.
-        throw new UnsupportedOperationException("Not implemented yet.");
-    }
-
-    /** {@inheritDoc} */
-    @Override
-    public ReactiveResultSet executeReactive(@Nullable Transaction transaction, Statement statement, @Nullable Object... arguments) {
```
korlov42, line 265: does it make sense to remove `ReactiveResultSet` interface as well?

korlov42 (approved):

isapego (approved):

modules/client/src/main/java/org/apache/ignite/internal/client/sql/ClientSql.java
ptupitsyn: Good catch, done.

PR #3516, opened 2024-03-29
IGNITE-21288 C++: Implement job execution interface

isapego:
- Implemented `job_execution` class and re-named compute methods to `submit...`;
- Added tests;
- Fixed some bugs.

https://issues.apache.org/jira/browse/IGNITE-21288


modules/platforms/cpp/ignite/client/compute/compute.h
```diff
@@ -48,63 +49,62 @@ class compute {
     compute() = delete;
 
     /**
-     * Submits a compute job represented by the given class for an execution on one of the specified nodes.
+     * Executes a compute job represented by the given class for an execution on one of the specified nodes.
```
ptupitsyn, line 52: > Executes ... for an execution

Why not keep `Submits`?

modules/platforms/cpp/ignite/client/detail/compute/compute_impl.cpp
```diff
@@ -148,24 +291,59 @@ void compute_impl::submit_colocated(const std::string &table_name, const ignite_
...
 
+void compute_impl::get_status_async(uuid id, ignite_callback<std::optional<job_status>> callback) {
+    auto writer_func = [id](protocol::writer &writer) { writer.write(id); };
+
+    auto reader_func = [](protocol::reader &reader) -> std::optional<job_status> {
+        return read_job_status_opt(reader);
+    };
+
+    m_connection->perform_request<std::optional<job_status>>(
+        protocol::client_operation::COMPUTE_GET_STATUS, writer_func, std::move(reader_func), std::move(callback));
```
ptupitsyn, line 312: Why don't we `std::move(writer_func)`?

modules/platforms/cpp/ignite/client/detail/compute/job_execution_impl.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+        callback({std::move(copy)});
+    } else if (m_error) {
+        auto copy{*m_error};
+        guard.unlock();
+
+        callback({std::move(copy)});
+    } else {
+        if (m_result_callback)
+            throw ignite_error("A callback for this result was already submitted");
```
ptupitsyn, line 38: Please explain - when does this happen? As a user, can I call `get_result_async` multiple times?

ptupitsyn (approved):

modules/platforms/cpp/ignite/client/compute/compute.h
isapego: Agree. rollbacked.

modules/platforms/cpp/ignite/client/detail/compute/compute_impl.cpp
isapego: `writer_func` here is passed as a reference, because it's always called before we interrupt execution.

modules/platforms/cpp/ignite/client/detail/compute/job_execution_impl.cpp
isapego: If it's called from multiple threads in parallel.

isapego: This is a limitation, but I can not easily get rid of it  with the current approach to async execution.

ptupitsyn (approved):

PR #3522, opened 2024-03-29
IGNITE-21665 Implement Event log

PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-21528


modules/eventlog/src/test/java/org/apache/ignite/internal/eventlog/impl/ConfigurationBasedChannelRegistryTest.java
```diff
@@ -0,0 +1,179 @@
...
+    @Test
+    void addNewConfigurationEntry() {
+        // Given configuration with a channel.
+        cfg.channels().change(c -> c.create(TEST_CHANNEL, s -> {
+            s.changeEnabled(true);
+            s.changeEvents(IgniteEventType.USER_AUTHENTICATED.name());
+        }));
+
+        // When get channel from registry.
+        EventChannel channel = await().until(
```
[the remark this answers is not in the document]
PakhomovAlexander: I don't think we should rely on it here because we use listeners to update caches and it might be an async process.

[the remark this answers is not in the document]
PakhomovAlexander: Makes sence.

modules/eventlog/src/main/java/org/apache/ignite/internal/eventlog/api/EventLog.java
```diff
@@ -20,9 +20,31 @@
...
+ * <pre>
+ *     eventLog.log(() -> IgniteEvents.USER_AUTHENTICATED.create(EventUser.of("user1", "basicAuthenticationProvider"));
+ * </pre>
  */
 public interface EventLog {
+    /**
+     * Writes event into every channel this event relates to.
+     *
+     * @param eventProvider Event provider.
+     */
```
AMashenkov, line 48: What are the guarantees for Supplier? When and in which thread it will be called?
The javavoc should claims either "the implementation must call the lamda in the currrent thread", or notify the lamda must be thread safe.

Why do we pass a lamda instead of Event object? To do object construction lazily?
The implementation doesn't allow lazy construction, because no event filter is implied.

modules/eventlog/src/main/java/org/apache/ignite/internal/eventlog/impl/ConfigurationBasedChannelRegistry.java
```diff
@@ -0,0 +1,115 @@
...
+    @Override
+    public Set<EventChannel> findAllChannelsByEventType(String igniteEventType) {
+        guard.readLock().lock();
+        try {
+            Set<EventChannel> result = typeCache.get(igniteEventType);
+            return result == null ? Set.of() : new HashSet<>(result);
+        } finally {
+            guard.readLock().unlock();
+        }
+    }
```
AMashenkov, line 75: Why do we need a copy, if we already updated cache under lock?
This brings great performance hit.

Please, do copy-on-write in proper way instead of this.

modules/eventlog/src/main/java/org/apache/ignite/internal/eventlog/impl/ConfigurationBasedChannelRegistry.java
```diff
@@ -0,0 +1,115 @@
...
+                                    t -> new HashSet<>()
+                            ).add(channel);
+                        }
+                    }
+                });
+
+                return completedFuture(null);
+            } finally {
+                guard.writeLock().unlock();
+            }
```
AMashenkov, line 104: It will be better to create a new collection outside the lock, then just update the references under the lock.

modules/eventlog/src/main/java/org/apache/ignite/internal/eventlog/impl/ConfigurationBasedSinkRegistry.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import org.apache.ignite.configuration.NamedListView;
+import org.apache.ignite.configuration.notifications.ConfigurationListener;
+import org.apache.ignite.configuration.notifications.ConfigurationNotificationEvent;
+import org.apache.ignite.internal.eventlog.api.Sink;
+import org.apache.ignite.internal.eventlog.config.schema.EventLogConfiguration;
+import org.apache.ignite.internal.eventlog.config.schema.SinkView;
+
+class ConfigurationBasedSinkRegistry implements SinkRegistry {
```
AMashenkov, line 36: This class has the same issues as ConfigurationBasedChannelRegistry

modules/eventlog/src/main/java/org/apache/ignite/internal/eventlog/impl/ConfigurationBasedChannelRegistry.java
AMashenkov: BTW, if you cared about collection modification after returning, then you can just hold ImmutableSet.

modules/eventlog/src/main/java/org/apache/ignite/internal/eventlog/impl/EventLogImpl.java
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * @param cfg the configuration.
+     */
+    public EventLogImpl(EventLogConfiguration cfg) {
+        this(new ConfigurationBasedChannelRegistry(cfg, new ConfigurationBasedSinkRegistry(cfg)));
+    }
+
+    @Override
+    public void log(Supplier<Event> eventProvider) {
+        Event event = eventProvider.get();
```
AMashenkov, line 53: The Supplier is totally useless, because you can't get and filter by EventType without constructing an event object.
To avoid ambiguity, the method should either accepts Event object or EventType+lamda.

PR #3529, opened 2024-03-29
IGNITE-21876 Prevent thread hijacking via IgniteCompute

rpuch:
https://issues.apache.org/jira/browse/IGNITE-21876


modules/compute/src/main/java/org/apache/ignite/internal/compute/IgniteComputeImpl.java
```diff
@@ -75,8 +75,6 @@ public class IgniteComputeImpl implements IgniteComputeInternal {
 
     private final ComputeComponent computeComponent;
 
-    private final ThreadLocalRandom random = ThreadLocalRandom.current();
```
sashapolo, line 78: omg

sashapolo (approved):

PR #3536, opened 2024-04-01
IGNITE-21094 Sql. Combine single/script query execution methods into a single one.

xtern:
https://issues.apache.org/jira/browse/IGNITE-21094


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/Commons.java
```diff
@@ -896,4 +899,11 @@ public static List<RelNode> castInputsToLeastRestrictiveTypeIfNeeded(List<RelNod
 
         return actualInputs;
     }
+
+    /** Returns {@code true} if the specified properties allow multi-statement query execution. */
+    public static boolean isMultiStatementQuery(SqlProperties properties) {
```
ygerzhedovich, line 904: maybe it is better to rename it to reflect the real purpose of the method, e.g. `isMultiStatementQueryAllow'

ygerzhedovich (approved):

korlov42 (approved):

PR #3537, opened 2024-04-01
IGNITE-21897 Sql. Invalid collation passthrough for sorted aggregates

AMashenkov:
https://issues.apache.org/jira/browse/IGNITE-21897

### The Review Checklist

### Notes

### DIFF
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/agg/IgniteSortAggregateBase.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/agg/IgniteSortAggregateBase.java
@@ -17,6 +17,7 @@
 
 package org.apache.ignite.internal.sql.engine.rel.agg;
 
+import static java.util.function.Predicate.not;
 import static org.apache.ignite.internal.sql.engine.util.Commons.maxPrefix;
 
 import it.unimi.dsi.fastutil.ints.IntList;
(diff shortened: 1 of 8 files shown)


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/rel/agg/IgniteSortAggregateBase.java
```diff
@@ -77,22 +80,42 @@ default Pair<RelTraitSet, List<RelTraitSet>> passThroughCollation(
...
+    }
+
+    private static Pair<RelTraitSet, List<RelTraitSet>> passThroughCollation(
+            RelTraitSet nodeTraits,
+            List<RelTraitSet> inputTraits,
+            RelCollation collation
+    ) {
         return Pair.of(nodeTraits.replace(collation),
                 List.of(inputTraits.get(0).replace(collation)));
     }
```
korlov42, line 121: in general, such propagation (when you use the same collation for input and output) is not correct. Here is an example:

assume, I have table `T` with tuple `(a, b, c, d)`. In case of query like `GROUP BY b, d ORDER BY b, d`, the input collation should be `[1, 3]`, while the node (or, in other words, output) collation will be `[0, 1]`. Such transformation caused by the fact, that aggregate node always puts the grouping column in the very beginning of the resulting tuple

AMashenkov: Got it. Fixed.

Also, I've found I used default order (asc/desc) for prefix columns instead of preserving desired collation.
Seems, we need more tests on collation propagation.

AMashenkov: I've added more tests.
Unfortunately, in case, when there is Sort over SortAggregate, it is impossible to supply partial sort from the aggregate node to help upper Sort node. Because Sort is enforcer and erase collations on pass-through.

korlov42 (approved):

zstan (approved):

PR #3540, opened 2024-04-02
IGNITE-21762 Run cleanup on primary reelection for all finished txns.

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-21762


modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -666,7 +673,7 @@ private CompletableFuture<Void> makeFinishRequest(
                             new TxStateMeta(
                                     txResult.transactionState(),
                                     localNodeId,
-                                    old.commitPartitionId(),
+                                    old == null ? null : old.commitPartitionId(),
```
zstan, line 676: seems you need change your interface :
<T extends TxStateMeta> T updateTxMeta(UUID txId, Function<@Nullable TxStateMeta, TxStateMeta> updater);

zstan: image
and voila )

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -377,6 +379,10 @@ private UpdateCommandResult handleUpdateAllCommand(UpdateAllCommand cmd, long co
 
         TxMeta txMetaToSet = new TxMeta(
                 stateToSet,
+                cmd.tablePartitionIds()
```
sanpwc, line 382: We have a rule of thumb not to use stream API on the hot path if it's possible.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -394,6 +396,52 @@ public PartitionReplicaListener(
...
 
+    @Override
+    public CompletableFuture<Boolean> onPrimaryElected(PrimaryReplicaEventParameters evt) {
+        assert replicationGroupId.equals(evt.groupId()) : format(
+                "The replication group listener does not match the event [grp={}, eventGrp={}]",
+                replicationGroupId,
+                evt.groupId()
+        );
+
+        if (!localNode.id().equals(evt.leaseholderId())) {
```
sanpwc, line 407: Is it really possible? Primary replica elected is a local event.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -394,6 +396,52 @@ public PartitionReplicaListener(
...
+        schedulePersistentStorageScan();
+
+        // The future returned by this event handler can't wait for all cleanups because it's not necessary and it can block
+        // meta storage notification thread for a while, preventing it from delivering further updates (including leases) and therefore
+        // causing deadlock on primary replica waiting.
+        return falseCompletedFuture();
+    }
+
+    private void schedulePersistentStorageScan() {
+        txManager.executeWriteIntentSwitchAsync(this::runPersistentStorageScan);
```
sanpwc, line 420: Why we need to executeWriteIntentSwitchAync()?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -394,6 +396,52 @@ public PartitionReplicaListener(
...
+    }
+
+    private void runPersistentStorageScan() {
+        try (Cursor<IgniteBiTuple<UUID, TxMeta>> txs = txStateStorage.scan()) {
+            for (IgniteBiTuple<UUID, TxMeta> tx : txs) {
+                UUID txId = tx.getKey();
+                TxMeta txMeta = tx.getValue();
+
+                assert !txMeta.enlistedPartitions().isEmpty();
```
sanpwc, line 429: What about ABANDONED -> ABORTED, do we adjust enlistedPartitions with the one that triggered the recovery?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1665,7 +1714,10 @@ private CompletableFuture<TransactionResult> finishTransaction(
                                 txId,
                                 commit,
                                 commitTimestamp,
-                                catalogVersion
+                                catalogVersion,
+                                aggregatedGroupIds.stream()
+                                        .map(PartitionReplicaListener::tablePartitionId)
```
sanpwc, line 1719: Here and there: I'd rather skip using table in tablePartitionId because later it'll be zonePartitionId, so that, simple partitionId seems enough to me.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/TransactionStateResolver.java
```diff
@@ -241,7 +242,7 @@ private void resolveTxStateFromCommitPartition(
      * @param txId Transaction id.
      */
     private void markAbandoned(UUID txId) {
-        txManager.updateTxMeta(txId, TxStateMeta::abandoned);
+        txManager.updateTxMeta(txId, stateMeta -> stateMeta != null ? stateMeta.abandoned() : null);
```
sanpwc, line 245: It's the one related to NPE and not the ticket itself, right?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TxManager.java
```diff
@@ -155,6 +156,22 @@ CompletableFuture<Void> cleanup(
             UUID txId
     );
 
+    /**
+     * Sends cleanup request to the cluster nodes that hosts primary replicas for the enlisted partitions.
+     *
+     * @param enlistedPartitions Enlisted partition groups.
```
sanpwc, line 162: It would be nice to explain through javadoc, why sometimes it's just TableParitionIds and sometimes Map<TablePartitionId, String>.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: So, you are using `executeWriteIntentSwitchAync` only in order to have a thread switching side-effect, aren't you?
If true, there should be a better option.

modules/core/src/main/java/org/apache/ignite/internal/thread/ThreadOperation.java
```diff
@@ -25,8 +25,8 @@ public enum ThreadOperation {
     STORAGE_READ,
     /** Storage write. */
     STORAGE_WRITE,
-    /** Take replication protocol snapshot meta (used when initiating a full state transfer of a partition). */
-    TAKE_SNAPSHOT_META,
```
sanpwc, line 29: Do you mean that TAKE_SNAPHOT_META is the same as TX_METASTORAGE_READ?

modules/core/src/main/java/org/apache/ignite/internal/thread/ThreadOperation.java
```diff
@@ -25,8 +25,8 @@ public enum ThreadOperation {
...
+    /** Read TX State storage. */
+    TX_METASTORAGE_READ,
```
sanpwc, line 29: TX_METASTORAGE isn't a canonical name, I believe it'll be confused with common meta storage. TX_STATE_STRORAGE fits better from my point of view.
Anyway, do re really need special operation to read from tx state storage? If we do we probably should also have TX_STATE_STORAGE_WRITE. All in all, I believe that from the treads perspective it doesn't really matter which storage it is, it's just an ability to perform long running storage related actions.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -401,6 +404,15 @@ private UpdateCommandResult handleUpdateAllCommand(UpdateAllCommand cmd, long co
         return new TransactionResult(stateToSet, cmd.commitTimestamp());
     }
 
+    private static List<TablePartitionId> fromPartitionIdMessage(List<TablePartitionIdMessage> partitionIds) {
+        List<TablePartitionId> list = new ArrayList<>();
```
sanpwc, line 408: Minor: you know the size, let's propagate it to the constructor.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -394,6 +395,28 @@ public PartitionReplicaListener(
...
+    private void runPersistentStorageScan() {
+        try (Cursor<IgniteBiTuple<UUID, TxMeta>> txs = txStateStorage.scan()) {
+            for (IgniteBiTuple<UUID, TxMeta> tx : txs) {
+                UUID txId = tx.getKey();
+                TxMeta txMeta = tx.getValue();
+
+                assert !txMeta.enlistedPartitions().isEmpty();
+
+                if (isFinalState(txMeta.txState())) {
+                    txManager.cleanup(
```
sanpwc, line 407: Let's handle the future in order to log the expcetion if any. I understand that we really do not expect ones because of durable cleanup, however simple no-op may hide the problem.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -394,6 +395,28 @@ public PartitionReplicaListener(
         prepareIndexBuilderTxRwOperationTracker();
     }
 
+    private void runPersistentStorageScan() {
+        try (Cursor<IgniteBiTuple<UUID, TxMeta>> txs = txStateStorage.scan()) {
+            for (IgniteBiTuple<UUID, TxMeta> tx : txs) {
```
sanpwc, line 400: Could you please also add [DEBUG] logging on each primary replica start, that will print the number of transactions grouped per each state.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TxManagerImpl.java
```diff
@@ -329,20 +334,35 @@ public TxManagerImpl(
...
 
-            transactionInflights.cancelWaitingInflights(groupId);
+            action.accept(groupId);
 
             return falseCompletedFuture();
         });
     }
 
+    private CompletableFuture<Boolean> primaryReplicaElectedListener(PrimaryReplicaEventParameters eventParameters) {
+        return primaryReplicaEventListener(eventParameters, groupId -> {
```
sanpwc, line 355: Busy lock is missing.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -440,6 +467,12 @@ private CompletableFuture<?> processRequest(ReplicaRequest request, @Nullable Bo
                         processOperationRequestWithTxRwCounter(senderId, request, isPrimary, opTsIfDirectRo, leaseStartTime));
     }
 
+    private CompletableFuture<Void> processCleanupRecoveryMessage(TxCleanupRecoveryRequest request) {
+        runPersistentStorageScan();
```
sanpwc, line 471: What thread are we going to use here?

sanpwc: No, it's ok for now.

sanpwc (approved):

PR #3542, opened 2024-04-02
IGNITE-21907 Change thread count for RAFT disruptors to improve perfo…

vldpyatkov:
…rmance

https://issues.apache.org/jira/browse/IGNITE-21907


modules/raft/src/main/java/org/apache/ignite/raft/jraft/option/NodeOptions.java
```diff
@@ -49,7 +49,10 @@
  */
 public class NodeOptions extends RpcOptions implements Copiable<NodeOptions> {
     /** This value is used by default to determine the count of stripes in the striped queue. */
```
denis-chudov, line 51:
```suggestion
    /** This value is used by default to determine the count of stripes in the striped disruptors, excluding log manager disruptor. */
```

denis-chudov (approved):

PR #3547, opened 2024-04-03
IGNITE-21924 Improve user threads handling

rpuch:
https://issues.apache.org/jira/browse/IGNITE-21924


modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -241,19 +242,29 @@ private void onReplicaMessageReceived(NetworkMessage message, ClusterNode sender
...
+            if (PublicApiThreading.executingSyncPublicApi()) {
+                // It's a user thread, it executes a sync public API call, so it can do anything, no switch is needed.
+                return false;
+            }
+            if (PublicApiThreading.executingAsyncPublicApi()) {
+                // It's a user thread, it executes an async public API call, so it cannot do anything, a switch is needed.
+                return true;
+            }
+
+            // It's something else: either a JRE thread or an Ignite thread not marked with ThreadAttributes. In any case,
```
sashapolo, line 266: Should we throw an assertion here immediately?

modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
```diff
@@ -102,13 +101,13 @@ abstract class AbstractTableView<R> implements CriteriaQuerySource<R> {
...
-     * @param fut Future to wait to.
+     * @param future Future to wait to.
      * @param <T> Future result type.
      * @return Future result.
      */
-    protected final <T> T sync(Supplier<CompletableFuture<T>> fut) {
+    protected static <T> T sync(CompletableFuture<T> future) {
         try {
-            return fut.get().get();
+            return future.get();
```
sashapolo, line 110: I think the idea was that exceptions thrown by the method that creates the future will be handled in a similar way. Are you sure you don't break anything with this change?

modules/table/src/main/java/org/apache/ignite/internal/table/PublicApiThreadingRecordView.java
```diff
@@ -28,11 +28,13 @@
 import org.jetbrains.annotations.Nullable;
 
 /**
- * Wrapper around {@link RecordView} that adds anti-thread-hijacking protection to methods returning {@link CompletableFuture}s.
+ * Wrapper around {@link RecordView} that maintains public API invariants relating to threading.
+ * That is, it adds protection against thread hijacking by users and also marks threads as 'executing a sync user operation' or
```
sashapolo, line 32: Let's add a `<p>` tag here

modules/workers/src/main/java/org/apache/ignite/internal/worker/ThreadAssertions.java
```diff
@@ -64,20 +59,22 @@ public static void assertThreadAllowsToRead() {
...
-        }
-
         if (!(currentThread instanceof ThreadAttributes)) {
+            if (PublicApiThreading.executingSyncPublicApi()) {
+                // Allow everything if we ride a user thread while executing a public API call.
+
+                return;
+            }
+
             LOG.warn("Thread {} does not have allowed operations", trackerException(), currentThread);
```
sashapolo, line 69: Why do we need both a warning and an exception?

modules/core/src/main/java/org/apache/ignite/internal/lang/IgniteSystemProperties.java
```diff
@@ -70,6 +70,9 @@ public final class IgniteSystemProperties {
     /** Name of the property controlling whether thread assertions are enabled. */
     public static final String THREAD_ASSERTIONS_ENABLED = "IGNITE_THREAD_ASSERTIONS_ENABLED";
 
+    /** Name of the property controlling whether thread whitelisting is enabled for thread assertions. */
+    public static final String THREAD_ASSERTIONS_THREAD_WHITELISTING_ENABLED = "THREAD_ASSERTIONS_THREAD_WHITELISTING_ENABLED";
```
sashapolo, line 74: Is this property used anywhere? I may have missed it...

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
rpuch: Why? If thread assertions are enabled and the thread touches the storage, we'll get the same assertion, just with a deeper stack trace.

modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
rpuch: No, the idea was to have ability to do something before the operation starts being executed; I know this because I introduced that indirection :) This change just reverts it back.

modules/workers/src/main/java/org/apache/ignite/internal/worker/ThreadAssertions.java
rpuch: Some errors/exceptions get swallowed, so we can't rely on the `AssertionError` to track the problem that surfaced. Having it logged explicitly gives this guarantee.

I removed the second (tracking) exception though.

sashapolo (approved):

PR #3558, opened 2024-04-04
IGNITE-19274 Sql. Jdbc client time zone propagation

xtern:
https://issues.apache.org/jira/browse/IGNITE-19274

1. Added jdbc connection property to set "session" time zone ('connectionTimeZone`)
2. Added client time zone propagation to sql-engine in `JdbcQueryEventHandler#connect(ZoneId)`
3. Fixed the conversion from TIMESTAMP to TIMESTAMP TZ (when milliseconds=0).
```
SELECT '2008-01-01 00:00:01'::TIMESTAMP::TIMESTAMP WITH LOCAL TIME ZONE

The result was:
Caused by: java.lang.IllegalArgumentException: 2008-01-01 00:00:01.
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:143)
	at org.apache.calcite.util.TimestampString.normalize(TimestampString.java:129)
	at org.apache.calcite.util.TimestampString.<init>(TimestampString.java:64)
	at org.apache.calcite.util.TimestampWithTimeZoneString.<init>(TimestampWithTimeZoneString.java:54)
	at org.apache.calcite.runtime.SqlFunctions.toTimestampWithLocalTimeZone(SqlFunctions.java:4446)
```


modules/sql-engine/src/integrationTest/sql/types/time/time_parsing.test
```diff
@@ -25,5 +25,5 @@ SELECT '14:42:04.999'::TIME(3)::VARCHAR
 query I
 SELECT '14:42:04.999'::TIME(6)::VARCHAR
 ----
-14:42:04.999000
+14:42:04.999
```
AMashenkov, line 28: Why does precision is ignored?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/IgniteSqlFunctions.java
```diff
@@ -574,6 +575,75 @@ public static Long subtractTimeZoneOffset(long timestamp, TimeZone timeZone) {
...
+    }
+
+    /**
+     * Helper for CAST({time} AS VARCHAR(n)).
+     *
+     * <p>Note: this method is a copy of the avatica {@link DateTimeUtils#unixTimestampToString(long, int)} method,
+     *          with the only difference being that it does not add trailing zeros.
+     */
+    public static String unixTimeToString(int time, int precision) {
+        IgniteStringBuilder buf = new IgniteStringBuilder(8);
```
AMashenkov, line 609: I guess default buffer size is '8'  for precision '0'.
Should it be `8+ (precision>0) ? 1 + precision: 0 `?

modules/sql-engine/src/integrationTest/sql/types/time/time_parsing.test
xtern: Precision isn't ignored, we just don't add trailing zeros now.
I copied the original test to make it more clear.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/IgniteSqlFunctions.java
xtern: Done, thanks.

modules/jdbc/src/main/java/org/apache/ignite/internal/jdbc/ConnectionPropertiesImpl.java
```diff
@@ -124,11 +126,15 @@ private static String[] clientAuthValues() {
     private final StringProperty password = new StringProperty("password",
             "Password", null, null, false, null);
 
+    /** Client connection time zone ID. This property can be used by the client to change the time zone of the "session" on the server. */
+    private final StringProperty connectionTimeZone = new StringProperty("connectionTimeZone",
```
korlov42, line 130: why do we need this `connection` prefix?

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcClientTimeZoneTest.java
```diff
@@ -0,0 +1,226 @@
...
+                validateSingleRow("SELECT ts::VARCHAR, ts_tz::VARCHAR FROM test", stmt,
+                        "1970-01-01 00:00:00", "1970-01-01 01:00:00 " + timeZone);
+            });
+        }
+
+        {
+            String timeZone = "invalid/timezone";
+
+            assertThrows(
+                    ZoneRulesException.class,
```
zstan, line 137: do we need to wrap such exception into appropriate sqlexception ?

modules/sql-engine/src/integrationTest/sql/types/time/time_parsing.test
```diff
@@ -22,8 +22,8 @@ SELECT '14:42:04.999'::TIME(3)::VARCHAR
 ----
 14:42:04.999
 
+# trailing zeros get truncated
 query I
-SELECT '14:42:04.999'::TIME(6)::VARCHAR
+SELECT '14:42:04.000000'::TIME(6)::VARCHAR
```
zstan, line 27: i don`t understand here, you rewrite valid test into different one ?
i suppose you just need to append something like : SELECT '14:42:04.999000'::TIME(6)::VARCHAR, wdyt ?

xtern: After Andrew comment I decided to copy original test from `time_parsing_ignore.test` :thinking:

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/IgniteSqlFunctions.java
```diff
@@ -574,6 +575,75 @@ public static Long subtractTimeZoneOffset(long timestamp, TimeZone timeZone) {
         return timestamp - offset;
     }
 
+    /**
+     * Helper for CAST({timestamp} AS VARCHAR(n)).
+     *
+     * <p>Note: this method is a copy of the avatica {@link DateTimeUtils#unixTimestampToString(long, int)} method,
+     *          with the only difference being that it does not add trailing zeros.
+     */
+    public static String unixTimestampToString(long timestamp, int precision) {
```
zstan, line 584: is it critical that we append trailing zeros ? or it made for consistency between different temporal tests conversion ? what if we are using not modifying functions ?

xtern: > is it critical that we append trailing zeros ?

Without removing trailing zeros current conversion 
from TIMESTAMP to TIMESTAMP TZ fails in `TimestampString#normalize` method.
For example for string  `2008-01-01 00:00:01.000000`

It removes all zeros and got `2008-01-01 00:00:01.` and then expects that after '.' will be at least one digit.

I don't think this is critical since we can implement our own TIMESTAMP to TIMESTAMP_TZ conversion.

> it made for consistency between different temporal tests conversion

From my point of view, yes, it makes the current behavior more consistent.

The following queries:
```
SELECT (TIMESTAMP '2008-01-01 00:00:01.001')::VARCHAR
SELECT '2008-01-01 00:00:01.001'::TIMESTAMP::VARCHAR
SELECT '2008-01-01 00:00:01.001'::TIMESTAMP WITH LOCAL TIME ZONE::VARCHAR
SELECT (TIMESTAMP WITH LOCAL TIME ZONE '2008-01-01 00:00:01.001')::VARCHAR
```

Output the following result before the change:
```
2008-01-01 00:00:01.001
2008-01-01 00:00:01.001000
2008-01-01 00:00:01.001 UTC
2008-01-01 00:00:01.001 UTC
```

After the change:
```
2008-01-01 00:00:01.001
2008-01-01 00:00:01.001
2008-01-01 00:00:01.001 UTC
2008-01-01 00:00:01.001 UTC
```

modules/sql-engine/src/integrationTest/sql/types/time/time_parsing.test
xtern: But in general I agree, I think the comment is enough, so I did as you suggested.

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcClientTimeZoneTest.java
xtern: Thanks, I missed this one, wrapped in a standard jdbc SQLException (like other jdbc property validation errors).

korlov42 (approved):

zstan (approved):

PR #3562, opened 2024-04-05
IGNITE-21494 Data Streamer: use one executor per node for flushTimer

ptupitsyn:
Use one shared lazily-initialized scheduled executor for data streamer auto flush functionality per node / client (instead of new executor per streamer).


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -2542,4 +2553,13 @@ private void destroyTableStorageOnRecoveryBusy(CatalogTableDescriptor tableDescr
 
         engine.dropMvTable(tableDescriptor.id());
     }
+
+    private synchronized ScheduledExecutorService streamerFlushExecutor() {
+        if (streamerFlushExecutor == null) {
+            streamerFlushExecutor = Executors.newSingleThreadScheduledExecutor(
```
rpuch, line 2559: Are we protected from first call to this method arriving after `close()` is closed (leaving the executor never shut down)?

modules/client/src/test/java/org/apache/ignite/client/fakes/FakeInternalTable.java
```diff
@@ -492,4 +493,10 @@ private void onDataAccess(String operation, Object arg) {
     public @Nullable PendingComparableValuesTracker<Long, Void> getPartitionStorageIndexTracker(int partitionId) {
         return null;
     }
+
+    @Override
+    public ScheduledExecutorService streamerFlushExecutor() {
+        //noinspection DataFlowIssue
+        return null;
```
rpuch, line 500: Would it make sense to throw an `UnsupportedOperationException` instead of returning a `null` (that might make something else explode way later)?

modules/core/src/main/java/org/apache/ignite/internal/streamer/StreamerSubscriber.java
```diff
@@ -258,11 +254,7 @@ private void initFlushTimer() {
             return;
         }
 
-        String threadPrefix = "client-data-streamer-flush-" + hashCode();
-
-        flushTimer = Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory(threadPrefix, log));
-
-        flushTask = flushTimer.scheduleAtFixedRate(this::flushBuffers, interval, interval, TimeUnit.MILLISECONDS);
+        flushTask = flushExecutor.scheduleAtFixedRate(this::flushBuffers, interval, interval, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
ptupitsyn: Good catch, fixed this and some other potential race conditions.

modules/core/src/test/java/org/apache/ignite/internal/streamer/StreamerSubscriberTest.java
```diff
@@ -23,16 +23,35 @@
...
+
+    @BeforeAll
+    public static void flushExecutorInit() {
+        flushExecutor = Executors.newSingleThreadScheduledExecutor(
+                new NamedThreadFactory("flushExecutor", Loggers.forClass(StreamerSubscriberTest.class)));
+    }
+
+    @AfterAll
+    public static void flushExecutorShutdown() {
+        flushExecutor.shutdown();
```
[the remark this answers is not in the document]
ptupitsyn: Done.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
ptupitsyn: Added busy lock check

modules/client/src/test/java/org/apache/ignite/client/fakes/FakeInternalTable.java
ptupitsyn: Done.

rpuch (approved):

PR #3563, opened 2024-04-05
IGNITE-21904 Provide extension point to register EventSerializer

dant3:


modules/eventlog/src/main/java/org/apache/ignite/internal/eventlog/ser/JacksonBasedJsonSerializer.java
```diff
@@ -18,26 +18,26 @@
...
     private final ObjectMapper mapper;
 
     /** Default constructor. */
-    public JsonEventSerializer() {
+    public JacksonBasedJsonSerializer(Class<T> eventClass, JsonSerializer<T> eventJsonSerializer) {
         mapper = new ObjectMapper();
         SimpleModule module = new SimpleModule();
-        module.addSerializer(EventImpl.class, new EventImplJacksonSerializer());
+        module.addSerializer(eventClass, eventJsonSerializer);
         module.addSerializer(EventUser.class, new EventUserJacksonSerializer());
```
PakhomovAlexander, line 41: Can you please define such a registry that can be used to register a serializer for any type, including EventUser? I don't like that we can register the serializer for the Event but can not register a serializer for EventUser with the same approach.

modules/eventlog/src/test/java/org/apache/ignite/internal/eventlog/ser/RegistryBackedEventSerializerTest.java
```diff
@@ -0,0 +1,87 @@
...
+                                + "\"message\":{\"version\":1,\"body\":\"test\"},"
+                                + "\"fields\":{\"hasMessage\":true}"
+                                + "}"
+                ),
+                Arguments.of(
+                        IgniteEvents.USER_AUTHENTICATED.builder()
+                                .productVersion("3.0.0")
+                                .timestamp(1234567890)
+                                .user(EventUser.of("test_user", "test_provider"))
+                                .fields(Map.of("ip", "127.0.0.1", "id", "123"))
```
PakhomovAlexander, line 57: could you please put message to the fields also?

PakhomovAlexander: Thanks for the informative answer. You made me think about the following approach: we can introduce the constraint to all Event children -- they should be a Pojo classes. This constraint seems to be ok when we speak about Events and its fields. 

In this case the architecture is going to be as stupid as possible. No extension points, no way to adjust the serialization mechanism. Just give us POJO. 

WDYT?

PakhomovAlexander (approved):

PR #3565, opened 2024-04-05
IGNITE-21995 Do not disable thread assertions in tests

rpuch:
https://issues.apache.org/jira/browse/IGNITE-21995


modules/core/src/main/java/org/apache/ignite/internal/wrapper/Wrappers.java
```diff
@@ -34,4 +36,19 @@ public static <T> T unwrap(Object object, Class<T> classToUnwrap) {
...
     }
+
+    /**
+     * Unwraps an object or returns {@code null} if it's {@code null}.
+     *
+     * @param object Object to unwrap.
+     * @param classToUnwrap Class which is to be unwrapped.
+     * @see #unwrap(Object, Class)
+     */
+    public static @Nullable <T> T unwrapNullable(@Nullable Object object, Class<T> classToUnwrap) {
```
tkalkirill, line 47: I saw that method `java.lang.Class#cast` will return `null` if `null` is passed to it, I think we can change the signature of `org.apache.ignite.internal.wrapper.Wrappers#unwrap` or add a check to it.

tkalkirill (approved):

PR #3575, opened 2024-04-08
IGNITE-22001 Throw specific exception if during writeTableAssignmentsToMetastore process was interrupted

JAkutenshi:
Jira's ticket link: https://issues.apache.org/jira/browse/IGNITE-22001

The problem was in returning null in cases of `TableManger#writeTableAssignmentsToMetastore` interruption while internal future proceeds. It leads to not caught issues after the method callings.

The solution is to throw a specific exception to detect and easily fix any problem during attempt to write assignments into a metastore.

In the result:
- `WritingAssignmentsToMetastoreException` class is implemented. 
- In `.exceptionally(...)` case of `TableManger#writeTableAssignmentsToMetastore` now the new exception is thrown.
- `TableManagerTest#testWriteTableAssignmentsToMetastoreExceptionally` is introduced with 2 main cases of the method interruption: after and before metastore's invocation completion.

There is an issue that in `.exceptionally(...)` for main assignments future we can't got access to completed list, then exception class has additional `isAssignmentsWereAvailable()` method to check if the list wasn't available in a scope while exception throwing. And in `TableManager.java:755` made a check for the nested exception duplicity evasion.

---

### The Review Checklist
- [x] **Formal criteria:** TC status, codestyle, mandatory documentation. Also make sure to complete the following:  

### Notes


alievmirza:
Let's try to do the task without introducing a new exception. With the new proposed exception, we have to throw it every time we change assignments in the metastore. Let's try to not increase a lot the scope of the task.

JAkutenshi:
Fix done!

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -752,10 +762,14 @@ private CompletableFuture<List<Assignments>> writeTableAssignmentsToMetastore(
                             });
                         }
                     })
-                    .exceptionally(e -> {
-                        LOG.error("Couldn't write assignments to metastore", e);
-
-                        return null;
+                    .handle((realAssignments, e) -> {
+                        if (e != null) {
+                            LOG.error("Couldn't write assignments {} to metastore during invoke",
```
alievmirza, line 767: from my point of view, the message is a bit incorrect, exception could caught here not only from invoke, let's rephrase it

alievmirza (approved):

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -135,4 +136,14 @@ public static Assignments fromBytes(byte @Nullable [] bytes) {
...
         return S.toString(this);
     }
+
+    /**
+     * Creates a string representation of the given assignments list for logging usage purpose mostly.
+     *
+     * @param assignments List of assignments to present as string.
+     * @return String representation of the given assignments list.
+     */
+    public static String assignmentListToString(List<Assignments> assignments) {
```
sanpwc, line 146: Why we do not use toString for that purpose?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -707,20 +707,37 @@ private CompletableFuture<List<Assignments>> writeTableAssignmentsToMetastore(
...
             }
 
             Condition condition = notExists(new ByteArray(partitionAssignments.get(0).key()));
 
             return metaStorageMgr
                     .invoke(condition, partitionAssignments, Collections.emptyList())
+                    .handle((invokeResult, e) -> {
+                        if (e != null) {
+                            LOG.error(
+                                    "Couldn't write assignments {} to metastore during invoke",
```
sanpwc, line 723: Should we also wrap assignments with square braces?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -738,24 +755,31 @@ private CompletableFuture<List<Assignments>> writeTableAssignmentsToMetastore(
                                     Entry assignmentsEntry = metaStorageAssignments.get(stablePartAssignmentsKey(partId));
 
                                     assert assignmentsEntry != null && !assignmentsEntry.empty() && !assignmentsEntry.tombstone()
-                                            : "Unexpected assignments for partition [" + partId + ", entry=" + assignmentsEntry + "].";
```
sanpwc, line 741: Why?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1240,32 +1264,16 @@ private CompletableFuture<?> createTableLocally(
...
-                                zoneDescriptor.partitions(),
-                                zoneDescriptor.replicas()
-                        ).stream().map(Assignments::of).collect(toList()));
-
-                assignmentsFuture.thenAccept(assignmentsList -> {
-                    LOG.info(IgniteStringFormatter.format("Assignments calculated from data nodes [table={}, tableId={}, assignments={}, "
-                            + "revision={}]", tableDescriptor.name(), tableId, assignmentListToString(assignmentsList), causalityToken));
-                });
-            }
+            CompletableFuture<List<Assignments>> assignmentsFuture = createOrGetAssignments(
```
sanpwc, line 1271: getOrCreate

sanpwc (approved):

PR #3584, opened 2024-04-10
IGNITE-20416 Retry when schema change is detected during implicit transaction

rpuch:
https://issues.apache.org/jira/browse/IGNITE-20416

### DIFF
--- a/modules/platforms/dotnet/Apache.Ignite.Tests/Table/SchemaSynchronizationTest.cs
+++ b/modules/platforms/dotnet/Apache.Ignite.Tests/Table/SchemaSynchronizationTest.cs
@@ -339,7 +339,6 @@ public class SchemaSynchronizationTest : IgniteTestsBase
         [Values(true, false)] bool insertNewColumn,
         [Values(true, false)] bool withRemove)
     {
-        using var metricListener = new MetricsTests.Listener();
         await Client.Sql.ExecuteAsync(null, $"CREATE TABLE {TestTableName} (KEY bigint PRIMARY KEY)");
 
         var table = await Client.Tables.GetTableAsync(TestTableName);
@@ -368,10 +367,6 @@ public class SchemaSynchronizationTest : IgniteTestsBase
 
                 yield return DataStreamerItem.Create(GetTuple(i));
             }
-
-            // Wait for background streaming to complete.
-            // TODO: Remove this workaround when IGNITE-20416 is fixed.
-            metricListener.AssertMetricGreaterOrEqual("streamer-items-sent", 6, 3000);
 
             // Update schema.
             // New schema has a new column with a default value, so it is not required to provide it in the streamed data.
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
@@ -167,13 +167,20 @@ abstract class AbstractTableView<R> implements CriteriaQuerySource<R> {
                             }
 
                             if (isOrCausedBy(InternalSchemaVersionMismatchException.class, ex)) {
+                                // There is no transaction, and table version was changed between taking the table version (that was used
+                                // to marshal inputs and would be used to unmarshal outputs) and starting an implicit transaction
+                                // in InternalTable. A transaction must always work with binary rows of the same table version matching the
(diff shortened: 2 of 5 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
```diff
@@ -163,11 +164,14 @@ private <T> CompletableFuture<T> withSchemaSync(@Nullable Transaction tx, @Nulla
...
 
                                 // Repeat.
+                                return withSchemaSync(tx, schemaVersion, action);
+                            } else if (tx == null && isOrCausedBy(IncompatibleSchemaException.class, ex)) {
+                                // Schema has changed while we were executing an implicit transaction: let's retry.
+                                assertSchemaVersionIncreased(previousSchemaVersion, schemaVersion);
+
                                 return withSchemaSync(tx, schemaVersion, action);
                             } else {
                                 return ex == null ? completedFuture(res) : CompletableFuture.<T>failedFuture(ex);
```
sashapolo, line 177: I would move the null check at the very beginning, it's a bit more clear: we first check the normal outcome and then start checking the exceptions

modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
```diff
@@ -163,11 +164,14 @@ private <T> CompletableFuture<T> withSchemaSync(@Nullable Transaction tx, @Nulla
...
                             if (isOrCausedBy(InternalSchemaVersionMismatchException.class, ex)) {
                                 assert tx == null : "Only for implicit transactions a retry might be requested";
-                                assert previousSchemaVersion == null || !Objects.equals(schemaVersion, previousSchemaVersion)
-                                        : "Same schema version (" + schemaVersion
-                                        + ") on a retry: something is wrong, is this caused by the test setup?";
+                                assertSchemaVersionIncreased(previousSchemaVersion, schemaVersion);
 
                                 // Repeat.
+                                return withSchemaSync(tx, schemaVersion, action);
+                            } else if (tx == null && isOrCausedBy(IncompatibleSchemaException.class, ex)) {
```
sashapolo, line 171: If this represents a schema change, then was does `InternalSchemaVersionMismatchException` do?

modules/platforms/dotnet/Apache.Ignite.Tests/Table/SchemaSynchronizationTest.cs
```diff
@@ -339,7 +339,6 @@ public async Task TestClientUsesLatestSchemaOnReadKv()
         [Values(true, false)] bool insertNewColumn,
         [Values(true, false)] bool withRemove)
     {
-        using var metricListener = new MetricsTests.Listener();
```
sashapolo, line 342: What's this about?

rpuch: This was added to work around a transaction abort that could happen due to a concurrent schema change and an implicit transaction (generated by the streamer)

modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
rpuch: When doing a KV operation and we DON'T have an explicit transaction, this is what happens:

1. We take the 'current' table version at 'now' and marshal keys/values using this version (this is moment T1)
2. Then we call a method of `InternalTable`, it creates a new transaction at moment T2
3. Then an operation is executed at moment T3

`InternalSchemaVersionMismatchException` is for a case when table version at T2 is different from table version at T1. This is needed to close that gap between marshalling and starting a transaction: to make sure that transaction works with rows in the version it expects, but avoid creating a transaction in the KV view (we tried this, it causes troubles)

And `IncompatibleSchemaException` is for a case when table version at T3 is different from table version at T2.

sashapolo: Can we add corresponding comments inside the branches describing these situations?

sashapolo (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/AbstractTableView.java
rpuch: Added comments

PR #3589, opened 2024-04-11
IGNITE-22021 Set null into IndexStorage#getNextRowIdToBuild for the primary index

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-22021

### The Review Checklist

### Notes


modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageHashIndexDescriptor.java
```diff
@@ -33,9 +33,7 @@
  * @see HashIndexStorage
  */
 public class StorageHashIndexDescriptor implements StorageIndexDescriptor {
-    /**
-     * Descriptor of a Hash Index column.
-     */
+    /** Descriptor of a Hash Index column. */
```
ibessonov, line 36: Please revert all these useless javadoc updates

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageHashIndexDescriptor.java
```diff
@@ -81,25 +79,29 @@ public String toString() {
...
 
     /**
      * Creates an Index Descriptor from a given set of columns.
      *
      * @param indexId Index id.
      * @param columns Columns descriptors.
+     * @param pk Primary index flag.
      */
-    public StorageHashIndexDescriptor(int indexId, List<StorageHashIndexColumnDescriptor> columns) {
+    public StorageHashIndexDescriptor(int indexId, List<StorageHashIndexColumnDescriptor> columns, boolean pk) {
```
ibessonov, line 101: Please mark this constructor as `@TestOnly`

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageHashIndexDescriptor.java
```diff
@@ -112,17 +114,22 @@ public List<StorageHashIndexColumnDescriptor> columns() {
         return columns;
     }
 
+    @Override
+    public boolean isPk() {
```
ibessonov, line 118: Why does it have to be an internal flag instead of the external comparison, like you do in the constructor? Seems excessive to me

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageIndexDescriptor.java
```diff
@@ -26,40 +26,29 @@
 import org.apache.ignite.internal.type.NativeType;
 import org.apache.ignite.internal.type.NativeTypes;
 
-/**
- * Index descriptor.
- */
+/** Index descriptor. */
```
ibessonov, line 29: Please revert all these useless javadoc updates

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageSortedIndexDescriptor.java
```diff
@@ -36,9 +36,7 @@
  * @see SortedIndexStorage
  */
 public class StorageSortedIndexDescriptor implements StorageIndexDescriptor {
-    /**
-     * Descriptor of a Sorted Index column (column name and column sort order).
-     */
+    /** Descriptor of a Sorted Index column (column name and column sort order). */
```
ibessonov, line 39: Please revert all these useless javadoc updates

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageSortedIndexDescriptor.java
```diff
@@ -99,34 +95,30 @@ public String toString() {
...
 
     /**
      * Creates an Index Descriptor from a given set of column descriptors.
      *
      * @param indexId Index ID.
      * @param columnDescriptors Column descriptors.
+     * @param pk Primary index flag.
      */
-    public StorageSortedIndexDescriptor(int indexId, List<StorageSortedIndexColumnDescriptor> columnDescriptors) {
+    public StorageSortedIndexDescriptor(int indexId, List<StorageSortedIndexColumnDescriptor> columnDescriptors, boolean pk) {
```
ibessonov, line 117: I have the same list of questions:
- if this constructor has to be public for some reason, then please make it `@TestOnly`
- why does it have to be an internal state rather than an external check? This is a calculable property, after all

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageSortedIndexDescriptor.java
```diff
@@ -150,20 +145,28 @@ private static List<StorageSortedIndexColumnDescriptor> extractIndexColumnsConfi
...
+                    assert column != null : "indexId=" + index.id() + ", columnName=" + columnName;
 
                     CatalogColumnCollation collation = columnDescriptor.collation();
 
                     return new StorageSortedIndexColumnDescriptor(columnName, getNativeType(column), column.nullable(), collation.asc());
                 })
                 .collect(toList());
     }
+
+    private static BinaryTupleSchema createSchema(List<StorageSortedIndexColumnDescriptor> columns) {
```
ibessonov, line 165: You moved this method to another place. Was there any specific reason for that? It bloats the PR for no good reason

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
```diff
@@ -123,25 +126,24 @@ public abstract class AbstractMvTableStorageTest extends BaseMvStoragesTest {
...
 
+    protected StorageIndexDescriptor pkIdx;
+
     private final CatalogService catalogService = mock(CatalogService.class);
 
-    protected final StorageIndexDescriptorSupplier indexDescriptorSupplier = new StorageIndexDescriptorSupplier() {
-        @Override
-        public @Nullable StorageIndexDescriptor get(int indexId) {
-            int catalogVersion = catalogService.latestCatalogVersion();
+    protected final StorageIndexDescriptorSupplier indexDescriptorSupplier = indexId -> {
```
ibessonov, line 133: Another useless code change. What was your motivation behind it?

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractIndexStorageTest.java
```diff
@@ -79,9 +85,11 @@ public abstract class AbstractIndexStorageTest<S extends IndexStorage, D extends
 
     protected static final int TEST_PARTITION = 12;
 
-    protected static final String INDEX_NAME = "TEST_IDX";
+    protected static final String TABLE_NAME = "FOO";
+
+    protected static final String PK_INDEX_NAME = pkIndexName(TABLE_NAME);
 
-    protected static final String TABLE_NAME = "foo";
+    protected static final String INDEX_NAME = "TEST_IDX";
```
ibessonov, line 92: Did you change the order of constants manually, or it's been done automatically by IDEA somehow?

ibessonov (changes requested):
I'm not convinced that you chose the best approach to `pk` flag, let's discuss why you implemented it this exact way

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageHashIndexDescriptor.java
tkalkirill: I didn't see anywhere that this was prohibited.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageHashIndexDescriptor.java
tkalkirill: It is also used in product code.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageHashIndexDescriptor.java
tkalkirill: Please give an example of how you could implement your idea in those places where I use this internal flag.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageIndexDescriptor.java
tkalkirill: I didn't see anywhere that this was prohibited.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageSortedIndexDescriptor.java
tkalkirill: I didn't see anywhere that this was prohibited.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageSortedIndexDescriptor.java
tkalkirill: It is also used in product code, please see usages.
Please give an example of how you could implement your idea in those places where I use this internal flag.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/index/AbstractIndexStorageTest.java
tkalkirill: I don't see anything wrong with this and it's not prohibited.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageSortedIndexDescriptor.java
tkalkirill: For me it looks more convenient that the internal methods are located below, I can return this.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageHashIndexDescriptor.java
ibessonov: Excuse me? Please don't introduce useless changes that only please you, you make it harder to review your PRs.
I see that you are very motivated to update javadocs to a uniform style. That's a good thing. But please do it once in a separate PR instead of polluting every other PR. Thank you!

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageHashIndexDescriptor.java
ibessonov: Why do we use it in production code, if this flag can be calculated from previous constructor parameters? This doesn't look like constructor that should be used in production code, I need a justification of its necessity, please provide it

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageHashIndexDescriptor.java
ibessonov: What do you mean? Many of those places I wouldn't even touch, code would be the same. My proposal - set nextRowIdToBuild to null when we create / recreate index that is PK. Explicitly and in the outside code

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageHashIndexDescriptor.java
tkalkirill: fix it

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageIndexDescriptor.java
tkalkirill: > Should I start a discussion on dev-list for you?
Only if you yourself are interested in it.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageHashIndexDescriptor.java
tkalkirill: What problems do I see in your campaign:
1) We need to remember that for the primary index we need to reset field `nextRowIdToBuild`, which may be forgotten after some time.
2) Performing these two operations is not atomic, it can lead to unpleasant consequences in the product code for users; the first thing that comes to mind is that the server will be pulled out of the socket immediately after the creation of the index storage is completed. Then we will need to add the correct recovery procedure somewhere else.

I don’t claim that my solution is ideal, I don’t really like it myself, but it looks more correct at the moment, I suggest leaving it.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageSortedIndexDescriptor.java
tkalkirill: Fix it

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageSortedIndexDescriptor.java
tkalkirill: fix it

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
tkalkirill: Lambda takes fewer lines of code, looks more modern and elegant, are you against it?

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageHashIndexDescriptor.java
ibessonov: Let's leave it, with hopes that we will change it in the future.
> Performing these two operations is not atomic

We can make them atomic, what's the issue?

> which may be forgotten after some time.

same applies to your code

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
ibessonov: Generally speaking, I'm against changing the code that's already good enough. And we we really want to do so, it's better to be done centralized.
That anonymous class wasn't bad. Few lines of extra code are not a real issue. So I conclude that you just wanted to change something because it didn't look like you wanted it to look, instead of thinking "does it really look inappropriate to the point of re-writing it?". I don't like useless changes in code. They bloat your pull requests, artificially increasing their sizes. They don't really contribute to the quality of the product - code doesn't become better.
_We should concentrate our efforts on real issues, instead of changing the code that's already good._

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageIndexDescriptor.java
tkalkirill: This task is very big, I don’t want to waste my mental energy on such a large non-functional change.

I have slightly revised my approach and will make such changes only where I actually change the code and around it only if this one needs to be changed according to the meaning of the task.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
tkalkirill: Answered a little higher.

modules/storage-api/src/main/java/org/apache/ignite/internal/storage/index/StorageHashIndexDescriptor.java
tkalkirill: > Let's leave it, with hopes that we will change it in the future.

Nice.

> We can make them atomic, what's the issue?

Yes, in general, we can, but so far in my head it looks better, plus writing tests in different places is not centralized.

> same applies to your code

Maybe.

modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
tkalkirill: Reverts changes.

ibessonov (approved):

PR #3602, opened 2024-04-12
IGNITE-22038 Make idle safe time sending more robust

rpuch:
https://issues.apache.org/jira/browse/IGNITE-22038

### DIFF
--- a/modules/network/src/main/java/org/apache/ignite/internal/network/NettyWorkersRegistrar.java
+++ b/modules/network/src/main/java/org/apache/ignite/internal/network/NettyWorkersRegistrar.java
@@ -109,8 +109,12 @@ public class NettyWorkersRegistrar implements IgniteComponent {
     }
 
     private void sendHearbeats() {
-        for (NettyWorker worker : workers) {
-            worker.sendHeartbeat();
+        try {
+            for (NettyWorker worker : workers) {
+                worker.sendHeartbeat();
+            }
+        } catch (Exception | AssertionError ignored) {
+            // Ignore.
         }
     }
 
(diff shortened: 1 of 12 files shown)


modules/network/src/main/java/org/apache/ignite/internal/network/NettyWorkersRegistrar.java
```diff
@@ -109,8 +109,12 @@ private void registerWorkersFor(EventLoopGroup group, List<NettyWorker> nettyWor
...
 
     private void sendHearbeats() {
-        for (NettyWorker worker : workers) {
-            worker.sendHeartbeat();
+        try {
+            for (NettyWorker worker : workers) {
+                worker.sendHeartbeat();
+            }
+        } catch (Exception | AssertionError ignored) {
+            // Ignore.
```
sanpwc, line 117: Should we log it here?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -764,15 +773,29 @@ private NetworkMessage prepareReplicaErrorResponse(boolean sendTimestamp, Throwa
...
-        replicas.values().forEach(r -> {
-            if (r.isDone()) {
-                ReplicaSafeTimeSyncRequest req = REPLICA_MESSAGES_FACTORY.replicaSafeTimeSyncRequest()
-                        .groupId(r.join().groupId())
-                        .build();
+        for (Entry<ReplicationGroupId, CompletableFuture<Replica>> entry : replicas.entrySet()) {
+            try {
+                sendSafeTimeSyncIfReplicaReady(entry.getValue());
+            } catch (Exception | AssertionError e) {
+                LOG.warn("Error while trying to send a safe time sync request to {}", e, entry.getKey());
```
sanpwc, line 780: I'd also mention that it's groupId in order not to confuse it with node name/id. Both here and below.

sanpwc (approved):

PR #3604, opened 2024-04-15
IGNITE-22042 Do not enlist operations in a FINISHING transaction

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-22042

### DIFF
--- a/modules/table/src/test/java/org/apache/ignite/internal/table/distributed/replication/PartitionReplicaListenerTest.java
+++ b/modules/table/src/test/java/org/apache/ignite/internal/table/distributed/replication/PartitionReplicaListenerTest.java
@@ -36,6 +36,7 @@ import static org.apache.ignite.internal.testframework.matchers.CompletableFutur
 import static org.apache.ignite.internal.tx.TransactionIds.beginTimestamp;
 import static org.apache.ignite.internal.tx.TxState.ABORTED;
 import static org.apache.ignite.internal.tx.TxState.COMMITTED;
+import static org.apache.ignite.internal.tx.TxState.FINISHING;
 import static org.apache.ignite.internal.tx.TxState.checkTransitionCorrectness;
 import static org.apache.ignite.internal.util.ArrayUtils.asList;
 import static org.apache.ignite.internal.util.CompletableFutures.nullCompletedFuture;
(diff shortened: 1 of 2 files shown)


modules/table/src/test/java/org/apache/ignite/internal/table/distributed/replication/PartitionReplicaListenerTest.java
```diff
@@ -704,6 +705,30 @@ public void testTxStateReplicaRequestCommitState() throws Exception {
         assertTrue(readTimestamp.compareTo(txMeta.commitTimestamp()) > 0);
     }
 
+    @Test
+    public void testExecuteRequestOnFinishedTx() {
```
sanpwc, line 709: - I'd rather convert it to parametrized test with current state as a parameter [COMMITTED, ABORTED, FIINISHING].
- I'd also iterate through all possible RequestType options.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/replication/PartitionReplicaListenerTest.java
```diff
@@ -704,6 +705,30 @@ public void testTxStateReplicaRequestCommitState() throws Exception {
...
+        UUID txId = newTxId();
+
+        txStateStorage.put(txId, new TxMeta(ABORTED, singletonList(grpId), null));
+        txManager.updateTxMeta(txId, old -> new TxStateMeta(ABORTED, null, null, null));
+
+        BinaryRow testRow = binaryRow(0);
+
+        assertThat(doSingleRowRequest(txId, testRow, RequestType.RW_INSERT), willThrowFast(TransactionException.class));
```
sanpwc, line 717: I'd also check that the message contains "Transaction is already finished".

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/replication/PartitionReplicaListenerTest.java
```diff
@@ -704,6 +705,30 @@ public void testTxStateReplicaRequestCommitState() throws Exception {
...
+    }
```
sanpwc, line 718: BTW I believe that it's fine to have finished even it's finishing within given context, what do you think?

sanpwc (approved):

PR #3609, opened 2024-04-16
IGNITE-22047 Ignore index destruction attempt if table storage is closed/destroyed

rpuch:
https://issues.apache.org/jira/browse/IGNITE-22047

### DIFF
--- a/modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
+++ b/modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
@@ -325,6 +325,33 @@ public abstract class AbstractMvTableStorageTest extends BaseMvStoragesTest {
     }
 
     /**
+     * Tests that an attempt to destroy an index in a table storage that is already destroyed does not
+     * cause an exception.
+     */
+    @ParameterizedTest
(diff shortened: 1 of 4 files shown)


modules/storage-api/src/testFixtures/java/org/apache/ignite/internal/storage/AbstractMvTableStorageTest.java
```diff
@@ -324,6 +324,33 @@ public void testDestroyIndex(boolean waitForDestroyFuture) throws Exception {
...
+
+        assertThat(partitionStorage.flush(), willCompleteSuccessfully());
+
+        CompletableFuture<Void> destroyTableStorageFuture = tableStorage.destroy();
+
+        if (waitForDestroyFuture) {
+            assertThat(destroyTableStorageFuture, willCompleteSuccessfully());
+        }
+
+        assertDoesNotThrow(() -> tableStorage.destroyIndex(sortedIdx.id()).get(10, SECONDS));
```
ibessonov, line 350: Isn't 10 seconds too much? How long is this test?

rpuch: It passes basically immediately. I'm just very cautious to not fail the test due to a random GC pause, for example.

ibessonov (approved):

PR #3633, opened 2024-04-18
IGNITE-21805 Refactor TableManager and move all RAFT related pieces to Replica

JAkutenshi:
Apache JIRA ticket: [IGNITE-21805 | Refactor TableManager and move all RAFT related pieces to Replica](https://issues.apache.org/jira/browse/IGNITE-21805)

## The goal

The main goal of this PR is to remove as more RAFT-related entities as we can from `TableManager` to the replication layer and classes like `ReplicaManager`, `Replica`. Related subtasks:

- Move `RaftManager` into `ReplicaManager` completely.
- All calls from `RaftManager` should be internal `ReplicaManager`'s calls or a restricted list of public methods with tickets on future removal of them.
- Create `ReplicaManager#replica()` method that returns future replica by given replication group ID; the method allows to get access to raft-clients outside replication layer, but through it.
- Hide other RAFT-related entities as much as possible: `RaftGroupOptions`, `RaftGroupEventsListener` (`RebalanceRaftGroupEventsListener`), `PartitionMover`, `TopologyAwareRaftGroupServiceFactory`, `LogStorageFactoryCreator` etc.
- Localize creation of raft-client in `TableManager` through `ReplicaManager#startRaftClient`, it's a workaround for now.

## The reason

In the future there are plans to try to implement another consensus protocol beside of RAFT, but now parts of the implementation sprouted outside replication layer and then it should be localized.

## Current issues 

1. There are several methods that stick out `ReplicaManager` now, but later should be removed:
  - `startRaftClient` there is only because now there are several assumptions in code that every node should have raft-client. It's wrong and will be fixed in [IGNITE-22315 | Make raft-client starting only once and only with raft-client and replica together](https://issues.apache.org/jira/browse/IGNITE-22315).
  - `getLogSyncer` should be removed by injecting it directly to engines as an `IgniteComponent` in [IGNITE-22292 | Moving LogSyncer from Loza and RaftServer and pass it to usages as IgniteComponent](https://issues.apache.org/jira/browse/IGNITE-22292).
2. Then, there is `TableRaftService` and it's calls. The class should be removed and all it's calls for (or on) raft-clients should be made with `replica(replicaGrpId).thenApply(Replica::raftClient).thenCompose(raftClient -> ... )` in [IGNITE-22218 | Remove TableRaftService in favor of using RaftGroupService from Replica instances](https://issues.apache.org/jira/browse/IGNITE-22218)

---

### The Review Checklist

### Notes


modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -493,45 +612,92 @@ public CompletableFuture<Replica> startReplica(
...
-                raftClient,
-                localNode,
-                executor,
-                placementDriver,
-                clockService
-        );
+        CompletableFuture<Replica> newReplicaFut = raftManager
+                // TODO IGNITE-19614 This procedure takes 10 seconds if there's no majority online.
+                .startRaftGroupService(replicaGrpId, newConfiguration, raftGroupServiceFactory, raftCommandsMarshaller)
+                .thenApply(createListener)
```
alievmirza, line 680: let's name it so it could reflect the function entity of this param

JAkutenshi:
A comment about [a test's fix there](https://github.com/apache/ignite-3/pull/3633/commits/9aa02ec66f3b4f007be62079fcc32c1160cb59d4#diff-41a813f5e33a0cf366f3366994eb3ff5beabd7c9590a4f04b2c05903ac356586R279): before the ticket there wasn't `.join()` on [TableManager](https://github.com/apache/ignite-3/pull/3633/commits/9aa02ec66f3b4f007be62079fcc32c1160cb59d4#diff-8539b597e51096e7007250e746e696d4e16351d09c3756b4989a1c331f788eadR993-R994), but now there is and if `startReplica()` returns null -- it fails with `NPE`. In context of the test, `RepelicaManager` is mocked and then, e.g. `busyLock` is null and so on. Without the method mocking the result of `startReplica()` is null and then `.join()` faces `NPE` that leads to `TimeoutException` on the top of stacktrace. As a solution I just mocking `startReplica()` that returns completed with null-value future instead just `null`.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/Replica.java
```diff
@@ -119,7 +117,7 @@ public Replica(
         this.replicaGrpId = replicaGrpId;
         this.listener = listener;
         this.storageIndexTracker = storageIndexTracker;
-        this.raftClient = raftClient;
+        this.raftClient = raftClient();
```
kgusakov, line 120: Why not simple listener.raftClient()?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -465,29 +461,101 @@ private void stopLeaseProlongation(ReplicationGroupId groupId, @Nullable String
         });
     }
 
+    public CompletableFuture<Replica> getReplica(ReplicationGroupId replicationGroupId) {
```
kgusakov, line 464: No usages?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -465,29 +461,101 @@ private void stopLeaseProlongation(ReplicationGroupId groupId, @Nullable String
...
+     * TODO.
+     *
+     * @param groupOptions TODO
+     * @param raftGrpLsnr TODO
+     * @param raftGrpEvtsLsnr TODO
+     * @param raftNodeId TODO
+     * @param stableConfiguration TODO
+     * @throws NodeStoppingException TODO
+     */
+    public void startPartitionRaftGroupNode(
```
kgusakov, line 509: Why we need this direct method? Can we do it internally on startReplica instead? It was the main idea - incapsulate as many RAFT logic as we can to the Replica(Manager) layer

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -497,41 +565,76 @@ public CompletableFuture<Replica> startReplica(
...
                 assert existingReplicaFuture == null || isCompletedSuccessfully(existingReplicaFuture);
                 LOG.info("Replica is started [replicationGroupId={}].", replicaGrpId);
 
-                return completedFuture(newReplica);
+                return newReplicaFut;
             } else {
-                existingReplicaFuture.complete(newReplica);
                 LOG.info("Replica is started, existing replica waiter was completed [replicationGroupId={}].", replicaGrpId);
 
+                existingReplicaFuture.complete(newReplicaFut.join());
```
kgusakov, line 636: Joint must be replaced by future chaining

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -611,6 +613,7 @@ public class IgniteImpl implements Ignite {
 
         LongSupplier partitionIdleSafeTimePropagationPeriodMsSupplier = partitionIdleSafeTimePropagationPeriodMsSupplier(replicationConfig);
 
+        Marshaller raftMarshaller = new ThreadLocalPartitionCommandsMarshaller(clusterSvc.serializationRegistry());
```
kgusakov, line 616: I think it will be better to encapsulate this logic in ReplicaManager itself (as it was earlier in TableManager)

modules/raft/src/main/java/org/apache/ignite/internal/raft/ExecutorInclinedRaftCommandRunner.java
```diff
@@ -46,4 +46,8 @@ public <R> CompletableFuture<R> run(Command cmd) {
 
         return future.thenApplyAsync(identity(), completionExecutor);
     }
+
+    public RaftCommandRunner decoratedCommandRunner() {
```
kgusakov, line 50: What's the purpose of this decorator?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -949,9 +934,9 @@ private CompletableFuture<Void> startPartitionAndStartClient(
                     return false;
                 }
 
-                if (((Loza) raftMgr).isStarted(raftNodeId)) {
+                if (replicaMgr.isRaftClientStarted(raftNodeId)) {
```
kgusakov, line 937: I think the whole if can be replaced by the method like replicaMgr.forcePeersUpdateIfNeeded(...) - again, it will encapsulate some raft logic to Replica layer.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
JAkutenshi: All joins are removed now

modules/raft/src/main/java/org/apache/ignite/internal/raft/ExecutorInclinedRaftCommandRunner.java
JAkutenshi: ✅ A good question to revise for me. So:
1. The one of primary goals was to make a hole for `RaftGroupManager` from `ReplicaListener` through `Replica` and `ReplicaManager` to `TableManager`.
2. Then, I should have a chain of `.raftClien()` among that classes.
3. There is a problem: `RaftClient` has raft client as `RaftCommandRunner` instead `RaftGroupManager`. And the problem is that command runner interface is appliable for both raft client and for the discussed command runner that is a decorator. But the last class isn't exactly raft client. Then, there should be a workaround for `raftClient` implementation in `ReplicaListener`:
```java
    @Override
    public RaftCommandRunner raftClient() {
        if (raftClient instanceof ExecutorInclinedRaftCommandRunner) {
            return ((ExecutorInclinedRaftCommandRunner) raftClient).decoratedCommandRunner();
        }
        return raftClient;
    }
```

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/Replica.java
```diff
@@ -128,6 +126,11 @@ public Replica(
         raftClient.subscribeLeader(this::onLeaderElected);
     }
 
+    public final TopologyAwareRaftGroupService raftClient() {
```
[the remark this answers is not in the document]
JAkutenshi: This is a workpiece for the next work

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/Replica.java
JAkutenshi: ✅ Just because of class casting: inside the method I cast client to the specific `Replica`'s raft client type. And a better way to localize such too specific (workaround? see TODO there) things in one certain method.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
JAkutenshi: ✅ Just yet

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
JAkutenshi: The next 3 arguments are being created in `TableManager`:
```
RaftGroupListener raftGrpLsnr
RaftGroupEventsListener raftGrpEvtsLsnr
RaftGroupOptions groupOptions
```
Some requires storages etc.. I can pass them through another lambda -- will it better solution for the ticket purposes?

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
JAkutenshi: Create a ticket for this activity https://issues.apache.org/jira/browse/IGNITE-22222

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
JAkutenshi: Will be implemented in IGNITE-22218

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1936,6 +1854,7 @@ private CompletableFuture<Void> handleChangePendingAssignmentEvent(
                     ? pendingAssignmentsNodes
                     : union(pendingAssignmentsNodes, stableAssignments);
 
+            // var isReplicaWasStarted = replicaMgr.isReplicaStarted(new TablePartitionId(tableId, partitionId));
```
kgusakov, line 1857: Pls remove

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -466,75 +490,167 @@ private void stopLeaseProlongation(ReplicationGroupId groupId, @Nullable String
...
             busyLock.leaveBusy();
         }
     }
 
     /**
-     * Internal method for starting a replica.
+     * Starts a raft-client and pass it to a replica creation if the replica should be started too. If a replica with the same partition id
+     * already exists, the method throws an exception.
      *
      * @param replicaGrpId Replication group id.
```
kgusakov, line 587: Javadoc params is not consistent with real aruments

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -548,6 +664,100 @@ private CompletableFuture<Replica> startReplicaInternal(
                 .thenCompose(v -> replicaFuture);
     }
 
+    /**
+     * TODO: will be private after https://issues.apache.org/jira/browse/IGNITE-22315
+     * Temporary public method for RAFT-client starting.
+     *
+     * @param replicaGrpId Replication Group ID.
+     * @param newConfiguration Peers and learners nodes for a raft group.
+     * @param raftClientCache Temporal supplier that returns RAFT-client from TableRaftService if it's already exists and was put into the
```
kgusakov, line 673: Pls, add TODO with ticket for removal

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -620,7 +830,15 @@ private CompletableFuture<Boolean> stopReplicaInternal(ReplicationGroupId replic
             }
         });
 
-        return isRemovedFuture;
+        return isRemovedFuture
+                .thenApply(v -> {
+                    try {
+                        raftManager.stopRaftNodes(replicaGrpId);
```
kgusakov, line 836: Shouldn't it be a part of concrete Replica stop?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/listener/ReplicaListener.java
```diff
@@ -18,11 +18,12 @@
 package org.apache.ignite.internal.replicator.listener;
 
 import java.util.concurrent.CompletableFuture;
+import org.apache.ignite.internal.raft.service.RaftCommandRunner;
 import org.apache.ignite.internal.replicator.ReplicaResult;
 import org.apache.ignite.internal.replicator.message.ReplicaRequest;
 
 /** Replica listener. */
-@FunctionalInterface
+// @FunctionalInterface
```
kgusakov, line 26: Pls remove

modules/replicator/build.gradle
```diff
@@ -29,6 +29,9 @@ dependencies {
     implementation project(':ignite-core')
     implementation project(':ignite-raft')
     implementation project(':ignite-raft-api')
+    implementation project(':ignite-metastorage-api')
```
kgusakov, line 32: Does really needed?

modules/replicator/src/integrationTest/java/org/apache/ignite/internal/replicator/ItPlacementDriverReplicaSideTest.java
```diff
@@ -477,34 +499,45 @@ private CompletableFuture<TopologyAwareRaftGroupService> createReplicationGroup(
...
-                                            } else {
-                                                return replicaListener.apply(request, senderId);
-                                            }
-                                        });
-                            },
-                            raftClient,
-                            new PendingComparableValuesTracker<>(Long.MAX_VALUE));
+                            newConfiguration,
+                            (unused) -> { },
+                            (unused) -> listener, // TODO
```
kgusakov, line 538: TODO?

kgusakov (changes requested):

modules/replicator/build.gradle
JAkutenshi: Without this I can't pass `MetaStorageService` into `startReplica` :< It's required by `RebalanceRaftGroupEventsListener`'s constructor

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
JAkutenshi: But there is above: 
>TODO: will be private after https://issues.apache.org/jira/browse/IGNITE-22315

Maybe better to change `private` to `removed`?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
```diff
@@ -1857,9 +1776,8 @@ private CompletableFuture<Void> handleChangePendingAssignmentEvent(
             boolean isRecovery
     ) {
         ClusterNode localMember = localNode();
-        RaftNodeId raftNodeId = new RaftNodeId(replicaGrpId, new Peer(localNode().name()));
 
-        boolean pendingAssignmentsAreForced = pendingAssignments.force();
+        boolean pendingAssignmentsAreForced = pendingAssignments != null && pendingAssignments.force();
```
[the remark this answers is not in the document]
JAkutenshi: Looks like I'm wrong here or it's last from my attempts to generalize this:
```java
if (pendingAssignmentsAreForced && replicaMgr.isReplicaStarted(replicaGrpId)) {
    replicaMgr.resetPeers(replicaGrpId, configurationFromAssignments(nonStableNodeAssignmentsFinal.nodes()));
}
```
btw, I'm removing it, thank you!

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/TableManager.java
JAkutenshi: (for a while.. :D)
Thank you!

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
JAkutenshi: This method `stopRaftNodes` is `raftManager` specific. `RaftManager` is responsible for nodes' starting and stopping => we're starting them in `ReplicaManager` => we should stop them in `ReplicaManager` too with `RaftManager` as well.

kgusakov: Ok, let's do it in this way now. In future, start and stop should be placed under Replica itself

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
JAkutenshi: Added missed params javadoc and added todos with https://issues.apache.org/jira/browse/IGNITE-22373 and https://issues.apache.org/jira/browse/IGNITE-22218

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
JAkutenshi: Changed to removed

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
JAkutenshi: Added TODO with https://issues.apache.org/jira/browse/IGNITE-22373 in `replicaMgr` usages, start/stop are included

modules/replicator/build.gradle
sanpwc, line 32: That seems incorrect. Replicator module is a generic abstraction that might be implemented by partition replicas or  meta storage ones, thus it should not depend on neither metastorage-api nor distribution-zones nor storage-api. Not sure about latter one though.

sanpwc (approved):

PR #3640, opened 2024-04-19
IGNITE-19681 Java thin: Add tx partition awareness (lazy tx start)

ptupitsyn:
**Problem**
Client starts transaction on a random node. We don't know which keys will participate at the moment of `begin()` call. Chosen target node (coordinator) might have none of the participating keys, which affects performance due to extra network calls.

**Fix**
* Start client transaction lazily, together with the first enlisted operation
* Choose coordinator using partition awareness for the first operation

```java
var tx = proxy.transactions().begin;  // Allocate lazy tx, do not start yet
tableView.put(tx, rec);  // Compute hash and primary replica for `rec`, start tx there, then perform put()
tableView.put(tx, rec2);  // rec2 might be on a different node, but at least one key is on coordinator
```

https://issues.apache.org/jira/browse/IGNITE-19681


modules/client/src/main/java/org/apache/ignite/internal/client/table/ClientTable.java
```diff
@@ -419,15 +435,21 @@ private <T> CompletableFuture<T> doSchemaOutInOpAsync(
...
-                    // Perform the operation.
-                    return ch.serviceAsync(opCode,
-                            w -> writer.accept(schema, w),
-                            r -> readSchemaAndReadData(schema, r, reader, defaultValue, responseSchemaRequired),
-                            preferredNodeName,
-                            retryPolicyOverride,
-                            expectNotifications);
+                    String txPreferredNodeName = getPreferredNodeName(provider, partitionsFut.getNow(null), schema);
+
+                    return ClientLazyTransaction.ensureStarted(tx, ch, txPreferredNodeName).thenCompose(unused -> {
```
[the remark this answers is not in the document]
ptupitsyn: Yes, I had this in mind too, but decided to keep the protocol simple. 

Ticket created: https://issues.apache.org/jira/browse/IGNITE-22090

modules/client/src/main/java/org/apache/ignite/internal/client/sql/ClientSql.java
```diff
@@ -242,7 +243,17 @@ public <T> CompletableFuture<AsyncResultSet<T>> executeAsync(
...
             try {
                 //noinspection resource
-                return ClientTransaction.get(transaction).channel().serviceAsync(ClientOp.SQL_EXEC, payloadWriter, payloadReader);
+                return ClientLazyTransaction.ensureStarted(transaction, ch, null)
+                        .thenCompose(tx -> tx.channel().serviceAsync(ClientOp.SQL_EXEC, payloadWriter, payloadReader))
+                        .exceptionally(e -> {
+                            Throwable ex = unwrapCause(e);
+                            if (ex instanceof TransactionException) {
+                                var te = (TransactionException) ex;
+                                throw new SqlException(te.traceId(), te.code(), te.getMessage(), te);
```
ascherbakoff, line 252: Here you create sql exception with txn error code, which is incorrect.
I don't think any unwapping is needed on client side. It's already done on server.
Same for line 258

ascherbakoff (approved):

modules/client/src/main/java/org/apache/ignite/internal/client/sql/ClientSql.java
ptupitsyn: This was added specifically to match the embedded behavior and fix `ItSqlApiBaseTest#checkTransactionsWithDml` (this test runs against client and embedded APIs to ensure consistency):

https://github.com/apache/ignite-3/blob/0f9f1df95e0006edd9c0b0595a7c2082f2d5cb29/modules/sql-engine/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlApiBaseTest.java#L249


Server-side mapping:

https://github.com/apache/ignite-3/blob/ee00a7c028ef6c8c1d8cdab0169fdd36deb1c5fe/modules/sql-engine/src/main/java/org/apache/ignite/internal/lang/SqlExceptionMapperUtil.java#L61

ptupitsyn: > It's already done on server.

`Transaction is already finished` exception is thrown by the client in this case:

https://github.com/apache/ignite-3/blob/a70aef41a3dc56da9437acead50e76d17259967f/modules/client/src/main/java/org/apache/ignite/internal/client/tx/ClientTransaction.java#L167

PR #3645, opened 2024-04-22
IGNITE-21908 Add metrics of distribution among stripes in disruptor

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-21908


modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
```diff
@@ -59,18 +60,34 @@ public class DistributionMetric extends AbstractMetric implements CompositeMetri
     /** Bounds of measurements. */
     private final long[] bounds;
 
+    /** Value string formatter or {@code null} to use default formatter. */
```
sk0x50, line 63: It seems to me, that this change should be applied to all metric classes: `IntMetric`, `IntGaguge` etc

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
```diff
@@ -59,18 +60,34 @@ public class DistributionMetric extends AbstractMetric implements CompositeMetri
...
+    private final Function<long[], String> stringFormatter;
+
     /** List of scalar metrics. */
     private volatile List<Metric> scalarMetrics = null;
 
     /**
-     * The constructor.
+     * Constructor.
```
sk0x50, line 70: You removed the article but added it a few lines below :)

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MovingAverageMetric.java
```diff
@@ -0,0 +1,93 @@
...
+
+import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.function.DoubleFunction;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * The metric calculates the average value for the last several operations.
+ */
+public class MovingAverageMetric extends AbstractMetric implements DoubleMetric {
+    /** Default rate. */
```
sk0x50, line 28: I think this is not a rate. `window size`?

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MovingAverageMetric.java
```diff
@@ -0,0 +1,93 @@
...
+package org.apache.ignite.internal.metrics;
+
+import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.function.DoubleFunction;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * The metric calculates the average value for the last several operations.
+ */
+public class MovingAverageMetric extends AbstractMetric implements DoubleMetric {
```
sk0x50, line 27: Let's rename this class as `SimpleMovingAverage`

modules/raft/src/main/java/org/apache/ignite/raft/jraft/core/NodeImpl.java
```diff
@@ -1247,17 +1251,20 @@ private void initPools(final NodeOptions opts) {
...
-                "JRaft-FSMCaller-Disruptor-" + ownFsmCallerExecutorDisruptorConfig.getThreadPostfix(),
+                "JRaft-FSMCaller-Disruptor" + ownFsmCallerExecutorDisruptorConfig.getThreadPostfix(),
                 opts.getRaftOptions().getDisruptorBufferSize(),
                 () -> new FSMCallerImpl.ApplyTask(),
                 ownFsmCallerExecutorDisruptorConfig.getStripes(),
                 false,
-                false
+                false,
+                null
+                //opts.getRaftMetrics().disruptorMetrics("JRaft-FSMCaller-Disruptor")
```
sk0x50, line 1267: should it be removed?

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/MovingAverageMetric.java
vldpyatkov: Renamed

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
vldpyatkov: This one is just a call to another constructor.
I am not sure about the usage of article hire, but I think the only constructor can be marked with the artical 'the'.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
vldpyatkov: Maybe we will do it when we start to use them.
Currently, it just adds code to review here.

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItMetaStorageManagerImplTest.java
```diff
@@ -110,7 +111,8 @@ void setUp(
 
         var raftGroupEventsClientListener = new RaftGroupEventsClientListener();
 
-        raftManager = new Loza(clusterService, raftConfiguration, workDir.resolve("loza"), clock, raftGroupEventsClientListener);
+        raftManager = new Loza(clusterService, new MetricManager(), raftConfiguration, workDir.resolve("loza"), clock,
```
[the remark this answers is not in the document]
vldpyatkov: Added a 'noOp' implementation to use in tests.

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/DistributionMetric.java
sk0x50: Honestly, I don't see any difference between `IngGaguge` and `LongGaguge`, and therefore this improvement should be added to both, I think. Please create a ticket to address this.

vldpyatkov: I removed all the string formatters in metriics.
We will add them in the future after discussion.

sk0x50 (approved):

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/SimpleMovingAverage.java
```diff
@@ -0,0 +1,93 @@
...
+    ConcurrentLinkedDeque<Double> queue = new ConcurrentLinkedDeque<>();
+
+    /**
+     * The constructor.
+     *
+     * @param name Name.
+     * @param desc Description.
+     * @param stringFormatter String formatter to get a readable value.
+     */
+    public SimpleMovingAverage(String name, @Nullable String desc, @Nullable DoubleFunction<String> stringFormatter) {
```
kgusakov, line 47: As I understand we agreed to remove formatter for now?

kgusakov (changes requested):

modules/metrics/src/main/java/org/apache/ignite/internal/metrics/SimpleMovingAverage.java
vldpyatkov: Fixed.

modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/JraftServerImpl.java
```diff
@@ -328,7 +333,8 @@ public CompletableFuture<Void> startAsync() {
                     ApplyTask::new,
                     opts.getStripes(),
                     false,
-                    false
+                    false,
+                    opts.getRaftMetrics().disruptorMetrics("JRaft-FSMCaller-Disruptor")
```
kgusakov, line 337: Not sure, that I understand the idea - why we set here "JRaft-FSMCaller-Disruptor" name and then convert it to "jraft.fsmcaller.disruptor" immediately? Maybe use the "jraft.fsmcaller.disruptor" here directly?

kgusakov (changes requested):

kgusakov (approved):

PR #3660, opened 2024-04-25
IGNITE-21953: Cover SQL E021-01(Character string types. CHARACTER data type) feature by tests

ygerzhedovich:
https://issues.apache.org/jira/browse/IGNITE-21953

### DIFF
--- a/modules/sql-engine/src/integrationTest/sql/types/char/test_char_character_set.test
+++ b/modules/sql-engine/src/integrationTest/sql/types/char/test_char_character_set.test
@@ -1,0 +1,60 @@
+# name: test/sql/types/char/test_char_character_set.test
+# description: E021-01 Character string types. CHARACTER SET tests
+# group: [char]
+
+# Incorrect name of charset
+statement error
+CREATE TABLE t_incorrect(C1 CHAR CHARACTER SET UTF_8);
+
+statement ok
+CREATE TABLE t_default (c1 CHAR, c2 CHARACTER, c3 CHAR(3), c4 CHARACTER(3));
+
+statement ok
+CREATE TABLE t_utf8 (c1 CHAR CHARACTER SET UTF8, c2 CHARACTER CHARACTER SET UTF8, c3 CHAR(3) CHARACTER SET UTF8, c4 CHARACTER(3) CHARACTER SET UTF8);
+
+statement ok
+CREATE TABLE t_latin1 (c1 CHAR CHARACTER SET LATIN1, c2 CHARACTER CHARACTER SET LATIN1, c3 CHAR(3) CHARACTER SET LATIN1, c4 CHARACTER(3) CHARACTER SET LATIN1);
+
+statement ok
+INSERT INTO t_default VALUES('1', '2', '🍞', 'Ю44');
+
+statement ok
+INSERT INTO t_utf8 VALUES('1', '2', '🍞', 'Ю44');
+
+# Must fail due to use symbol doesn't belong to charset of column
+skipif ignite3
+# https://issues.apache.org/jira/browse/IGNITE-22100
+statement error
+INSERT INTO t_latin1 VALUES('1', '2', '🍞', 'Ю44');
+
+# Must fail due to different character set
+skipif ignite3
+# https://issues.apache.org/jira/browse/IGNITE-22100
+statement error
+INSERT INTO t_latin1 SELECT * FROM t_utf8;
+
+# Must fail due to different character set
+skipif ignite3
+# https://issues.apache.org/jira/browse/IGNITE-22100
+statement error
+INSERT INTO t_utf8 SELECT * FROM t_latin1;
+
+query I
+SELECT count(*) from t_utf8 t1, t_default t2 where t1.c1=t2.c1;
+----
+1
+
+query I
+SELECT count(*) from t_utf8 t1, t_default t2 where t1.c2=t2.c2;
+----
+1
+
+query T
+SELECT count(*) from t_utf8 t1 join t_default t2 on (t1.c3=t2.c3);
+----
+1
+
+query T
+SELECT count(*) from t_utf8 t1 join t_default t2 on (t1.c4=t2.c4);
+----
+1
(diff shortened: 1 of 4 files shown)


modules/sql-engine/src/integrationTest/sql/types/char/test_char_length.test
```diff
@@ -0,0 +1,65 @@
+# name: test/sql/types/char/test_char_length.test
+# description: E021-01 Character string types. CHARACTER data type. Check length of the CHAR datatype.
+# group: [char]
+
+# length shouldn't be negative;
+statement error
+CREATE TABLE t_negative(c1 CHAR(-1));
+
+# length shouldn't be zero;
+skipif ignite3
```
zstan, line 10: skip without issue ?

modules/sql-engine/src/integrationTest/sql/types/char/test_char_length.test
```diff
@@ -0,0 +1,65 @@
...
+
+
+statement ok
+CREATE TABLE t_three (c1 CHAR(3), c2 CHARACTER(3));
+
+# MUST FAIL due to length
+skipif ignite3
+# https://issues.apache.org/jira/browse/IGNITE-18492
+statement error
+INSERT INTO t_three VALUES('1', '4444');
```
zstan, line 38: what about insertion with cast to char(3) ?

modules/sql-engine/src/integrationTest/sql/types/char/test_char_character_set.test
```diff
@@ -0,0 +1,60 @@
...
+
+# Incorrect name of charset
+statement error
+CREATE TABLE t_incorrect(C1 CHAR CHARACTER SET UTF_8);
+
+statement ok
+CREATE TABLE t_default (c1 CHAR, c2 CHARACTER, c3 CHAR(3), c4 CHARACTER(3));
+
+statement ok
+CREATE TABLE t_utf8 (c1 CHAR CHARACTER SET UTF8, c2 CHARACTER CHARACTER SET UTF8, c3 CHAR(3) CHARACTER SET UTF8, c4 CHARACTER(3) CHARACTER SET UTF8);
```
zstan, line 13: you cover CHARACTER SET but not cover COLLATION should it be done under different activity ?

modules/sql-engine/src/integrationTest/sql/types/char/test_char_length.test
ygerzhedovich: yep, thanks for the notice.

modules/sql-engine/src/integrationTest/sql/types/char/test_char_character_set.test
ygerzhedovich: yes, it's separate F690 feature

modules/sql-engine/src/integrationTest/sql/types/char/test_char_character_set.test
```diff
@@ -0,0 +1,60 @@
...
+INSERT INTO t_default VALUES('1', '2', '🍞', 'Ю44');
+
+statement ok
+INSERT INTO t_utf8 VALUES('1', '2', '🍞', 'Ю44');
+
+# Must fail due to use symbol doesn't belong to charset of column
+skipif ignite3
+# https://issues.apache.org/jira/browse/IGNITE-22100
+statement error
+INSERT INTO t_latin1 VALUES('1', '2', '🍞', 'Ю44');
```
zstan, line 28: seems we also need to check equality of returning types for char char(1) and character(1) ?
SELECT TYPEOF(SELECT '1'::char), TYPEOF(SELECT '1'::char(1))

modules/sql-engine/src/integrationTest/sql/types/char/test_char_length.test
ygerzhedovich: added

modules/sql-engine/src/integrationTest/sql/types/char/test_char_character_set.test
ygerzhedovich: `typeof` it's not a standard function and yes, our implementation returns type together with length, but I think it's not require to check it here.

modules/sql-engine/src/integrationTest/sql/types/char/test_char_length.test
```diff
@@ -0,0 +1,65 @@
...
+
+# MUST FAIL due to length
+skipif ignite3
+# https://issues.apache.org/jira/browse/IGNITE-18492
+statement error
+INSERT INTO t_three VALUES('1234', '444');
+
+statement ok
+INSERT INTO t_three VALUES('123', '456'), ('1  ', '1'), (' 2 ', '2'), ('  3','3');
```
zstan, line 54: did we have appropriate issue if we change this test for insertion not 3 but more white space consistent strings ?  seems we need such a case to be appended here ?
for example :
INSERT INTO t_three VALUES('     3     ', '     4          ');
SELECT LENGTH(c1), LENGTH(c2) FROM t_three;
----
3	1

modules/sql-engine/src/integrationTest/sql/types/char/test_char_character_set.test
zstan: yes i understand that this is not standard function i mean use this func to check corresponding equality

modules/sql-engine/src/integrationTest/sql/types/char/test_char_length.test
ygerzhedovich: No, it does not require the additional test. It is not actual, due to we have 
 https://issues.apache.org/jira/browse/IGNITE-18492

modules/sql-engine/src/integrationTest/sql/types/char/test_char_character_set.test
ygerzhedovich: Do you have an example in real life when it will be useful?

zstan (approved):

PR #3679, opened 2024-04-29
IGNITE-22086 Thin client: initialize observableTimestamp in handshake

ptupitsyn:
Before this fix we propagated `observableTimestamp` to the client with every response, but not on handshake. As a result, the very first operation from the client has `observableTimestamp = 0`, which can lead to causality issues.

### DIFF
--- a/modules/platforms/cpp/ignite/client/detail/node_connection.h
+++ b/modules/platforms/cpp/ignite/client/detail/node_connection.h
@@ -261,6 +261,7 @@ private:
 
     /** Configuration. */
     const ignite_client_configuration &m_configuration;
+    void on_observable_timestamp_changed(int64_t observable_timestamp) const;
 };
 
 } // namespace ignite::detail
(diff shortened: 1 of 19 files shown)


modules/platforms/cpp/ignite/client/detail/node_connection.h
```diff
@@ -261,6 +261,7 @@ class node_connection : public std::enable_shared_from_this<node_connection> {
 
     /** Configuration. */
     const ignite_client_configuration &m_configuration;
+    void on_observable_timestamp_changed(int64_t observable_timestamp) const;
```
isapego, line 264: This has to be moved upper, to methods section, and given some simple description.

isapego (changes requested):

modules/platforms/cpp/ignite/client/detail/node_connection.h
ptupitsyn: Fixed.

isapego (approved):

PR #3691, opened 2024-05-01
IGNITE-22000 Get rid of unnecessary conversion of DDL commands

xtern:
https://issues.apache.org/jira/browse/IGNITE-22000

* `DdlSqlToCommandConverter` now converts AST nodes not to `DDLCommand` but to `CatalogCommand` 
* Removed `DdlToCatalogCommandConverter` and all related to `DDLCommand` classes.
* Reworked related tests and removed `DistributionZoneDdlCommandHandlerTest` because it is now looks useless

Separate issue about handling `ifExists` flag inside command will be created, because it requires a lot of additional changes and current PR is big enough


modules/catalog/src/main/java/org/apache/ignite/internal/catalog/commands/AlterTableAddColumnCommand.java
```diff
@@ -53,15 +53,17 @@ public static AlterTableAddColumnCommandBuilder builder() {
...
      * @throws CatalogValidationException if any of restrictions above is violated.
      */
     private AlterTableAddColumnCommand(
             String tableName,
             String schemaName,
+            boolean ifTableExists,
             List<ColumnParams> columns
     ) throws CatalogValidationException {
-        super(schemaName, tableName);
+        super(schemaName, tableName, ifTableExists);
```
zstan, line 66: it`s strange to see such a flag here.

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/commands/AlterTableDropColumnCommand.java
```diff
@@ -146,6 +148,8 @@ private static class Builder implements AlterTableDropColumnCommandBuilder {
 
         private String tableName;
 
+        private boolean ifExists;
```
zstan, line 151: if column exists or table exists ?

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/commands/AlterZoneSetDefaultCommand.java
```diff
@@ -31,20 +31,24 @@
...
+    private final boolean ifExists;
+
     /**
      * Constructor.
      *
      * @param zoneName Name of the zone.
      * @throws CatalogValidationException if any of restrictions above is violated.
      */
-    private AlterZoneSetDefaultCatalogCommand(String zoneName) throws CatalogValidationException {
+    private AlterZoneSetDefaultCommand(String zoneName, boolean ifExists) throws CatalogValidationException {
```
zstan, line 48: fix javadoc

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/commands/AlterTableAddColumnCommand.java
xtern: I don't fully understand you suggestion.

It's valid statement
```
alter table if exists NON_EXISTING_TABLE add SOMECOLUMN int
```
should not produce updates and errors (in case NON_EXISTING_TABLE does not exist)

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/commands/AlterZoneSetDefaultCommand.java
xtern: Fixed, thanks.

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/commands/AlterTableAddColumnCommand.java
zstan: i didn`t see if exists in standard for such a case : 11.10 <alter table statement>
i miss smth ?

xtern: I think this is out of scope of this PR. At least PG accepts such syntax.

Btw, do you see `if exists` for `DROP TABLE` for example or `if not exists` for `CREATE TABLE` in the specification?

ygerzhedovich (approved):

AMashenkov (approved):

PR #3694, opened 2024-05-02
When handling AppendEntriesRequest, check that sender is in the quorum

rpuch:


rpuch:
The fix breaks Raft protocol

PR #3700, opened 2024-05-05
IGNITE-21167 Sql. Jdbc. Raise a clearer error message when the tx control statement is used in non-autocommit mode

xtern:
https://issues.apache.org/jira/browse/IGNITE-21167

In jdbc handler exception message has changed from
`Transaction control statement cannot be executed within an external transaction`
to more understandable
`Transaction control statements are not supported when autocommit mode is disabled`

`JdbcHandlerBase` abstract class was introduced to handle such errors uniformly.

### DIFF
--- a/modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcHandlerBase.java
+++ b/modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcHandlerBase.java
@@ -1,0 +1,175 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.client.handler;
+
+import static org.apache.ignite.internal.jdbc.proto.IgniteQueryErrorCode.UNSUPPORTED_OPERATION;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.CompletionStage;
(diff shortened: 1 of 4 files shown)


korlov42 (approved):

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcMultiStatementSelfTest.java
```diff
@@ -422,11 +408,18 @@ public void testAutoCommitFalse() throws Exception {
...
+    @SuppressWarnings("ThrowableNotThrown")
+    public void testAutoCommitFalseTxControlStatementsNotSupported() throws Exception {
+        String txErrMsg = "Transaction control statements are not supported when autocommit mode is disabled";
         conn.setAutoCommit(false);
-        assertThrowsSqlException(txErrMsg, () -> stmt.execute("START TRANSACTION; SELECT 1; COMMIT;"));
+        assertThrowsSqlException(txErrMsg, () -> stmt.execute("START TRANSACTION; SELECT 1; COMMIT"));
+        assertThrowsSqlException(txErrMsg, () -> stmt.execute("COMMIT"));
+        assertThrowsSqlException(txErrMsg, () -> stmt.execute("START TRANSACTION"));
+
+        boolean res = stmt.execute("SELECT 1;COMMIT");
```
zstan, line 419: seems we already have such a tests, it need to be merged or compound somehow 
testAutoCommitFalseNonCompleted

xtern: Yep,  I've just renamed it and refactored a bit

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcHandlerBase.java
```diff
@@ -0,0 +1,175 @@
...
+import org.apache.ignite.internal.util.ExceptionUtils;
+import org.apache.ignite.sql.ColumnMetadata;
+import org.apache.ignite.sql.ColumnType;
+import org.apache.ignite.sql.ResultSetMetadata;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Contains common methods used to process jdbc requests.
+ */
+abstract class JdbcHandlerBase {
```
zstan, line 45: abstract class with no abstract methods ?

zstan (approved):
overall looks good, besides one issue, a bit was discussed in private

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcHandlerBase.java
xtern: I don't see anything special here) For now abstract is used only to mark that the class cannot be instantiated on its own.

PR #3704, opened 2024-05-06
IGNITE-22130 Fix retries.

ascherbakoff:
[This patch fixes retry logic according to IGNITE-22130](https://issues.apache.org/jira/browse/IGNITE-22130)

### DIFF
--- a/modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaService.java
+++ b/modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaService.java
@@ -17,17 +17,22 @@
 
 package org.apache.ignite.internal.replicator;
 
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static org.apache.ignite.internal.lang.IgniteStringFormatter.format;
+import static org.apache.ignite.internal.util.ExceptionUtils.matchAny;
 import static org.apache.ignite.internal.util.ExceptionUtils.unwrapCause;
 import static org.apache.ignite.internal.util.ExceptionUtils.withCause;
 import static org.apache.ignite.lang.ErrorGroups.Replicator.REPLICA_COMMON_ERR;
+import static org.apache.ignite.lang.ErrorGroups.Replicator.REPLICA_MISS_ERR;
 import static org.apache.ignite.lang.ErrorGroups.Replicator.REPLICA_TIMEOUT_ERR;
+import static org.apache.ignite.lang.ErrorGroups.Transactions.ACQUIRE_LOCK_ERR;
 
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeoutException;
 import org.apache.ignite.internal.hlc.HybridClock;
 import org.apache.ignite.internal.lang.NodeStoppingException;
@@ -45,10 +50,14 @@ import org.apache.ignite.internal.replicator.message.ReplicaRequest;
 import org.apache.ignite.internal.replicator.message.ReplicaResponse;
 import org.apache.ignite.internal.replicator.message.TimestampAware;
 import org.apache.ignite.network.ClusterNode;
+import org.jetbrains.annotations.Nullable;
 import org.jetbrains.annotations.TestOnly;
 
 /** The service is intended to execute requests on replicas. */
 public class ReplicaService {
+    /** Retry timeout. */
+    private static final int RETRY_TIMEOUT_MILLIS = 10;
+
     /** Message service. */
     private final MessagingService messagingService;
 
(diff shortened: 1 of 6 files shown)


modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaService.java
```diff
@@ -210,7 +225,14 @@ private <R> CompletableFuture<R> sendToReplica(String targetNodeConsistentId, Re
                             return null;
                         });
                     } else {
-                        res.completeExceptionally(errResp.throwable());
+                        if (retryExecutor != null && matchAny(unwrapCause(errResp.throwable()), ACQUIRE_LOCK_ERR, REPLICA_MISS_ERR)) {
```
sk0x50, line 228: I still think that using exception classes is better than error codes, in general.
By the way, should `ACQUIRE_LOCK_TIMEOUT_ERR` be taken into account as well?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/TransactionExceptionMapperProvider.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.List;
+import org.apache.ignite.internal.lang.IgniteExceptionMapper;
+import org.apache.ignite.internal.lang.IgniteExceptionMappersProvider;
+import org.apache.ignite.internal.replicator.exception.ReplicationException;
+import org.apache.ignite.tx.TransactionException;
+
+/**
+ * Transaction module exception mapper.
+ */
+@AutoService(IgniteExceptionMappersProvider.class)
```
sk0x50, line 34: :+1:

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaService.java
ascherbakoff: >I still think that using exception classes is better than error codes, in general.
I disagree. The whole exception design is based on error codes. Checking error codes is more clean then compartion exception classes.
> By the way, should ACQUIRE_LOCK_TIMEOUT_ERR be taken into account as well?
No it should not. This error code and related functionality should be removed, because we got retries from client side.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaService.java
```diff
@@ -59,6 +68,8 @@ public class ReplicaService {
 
     private final ReplicationConfiguration replicationConfiguration;
 
+    private final ScheduledExecutorService retryExecutor;
```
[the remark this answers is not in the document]
ascherbakoff: 🆗

denis-chudov (approved):

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaService.java
sk0x50: > I disagree. The whole exception design is based on error codes. Checking error codes is more clean then comparing exception classes.

I disagree, using Java classes that represent an exception is a widespread practice. Error codes are a way to provide the user with an additional clue on critical situations, especially in the case of thin clients that are not supported exceptions.

ascherbakoff: > I disagree, using Java classes that represent an exception is a widespread practice. Error codes are a way to provide the user with an additional clue on critical situations, especially in the case of thin clients that are not supported exceptions.

Looks like we have not agreement here, because changing error code checking to `instanceof ` doesn't make much sense to me. This however have should not block PR changes. Or is this a merge blocker ?

sk0x50: This is definitely not a blocker.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -626,50 +621,71 @@ private <R> CompletableFuture<R> trackingInvoke(
...
-                                "Transaction is already finished [tableName={}, partId={}, txState={}].",
-                                tableName,
-                                partId,
-                                tx.state()
-                        )));
-            }
+        if (full) { // Full transaction retries are handled in postEnlist.
+            return replicaSvc.invoke(primaryReplicaAndConsistencyToken.get1(), request);
+        } else {
+            if (write) { // Track only write requests from explicit transactions.
```
vldpyatkov, line 627: `} else if (write) {`
This code format is acceptable here.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -1022,6 +1028,10 @@ private static boolean allSchemaVersionsSame(Collection<? extends BinaryRow> row
         boolean first = true;
 
         for (BinaryRow row : rows) {
+            if (row == null) {
```
vldpyatkov, line 1031: It looks like a bug, how is it happening?

PR #3713, opened 2024-05-07
IGNITE-21992 Fix remove-after-insert in Data Streamer within the same batch

ptupitsyn:
Within the same streamer batch, skip previous insert operations when a delete operation is encountered.

https://issues.apache.org/jira/browse/IGNITE-21992


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2374,7 +2374,13 @@ private CompletableFuture<ReplicaResult> processMultiEntryAction(ReadWriteMultiR
 
                     rowIdFuts[i] = resolveRowByPk(pk, txId, (rowId, row, lastCommitTime) -> {
                         if (isDelete && rowId == null) {
-                            // Does not exist, nothing to delete.
+                            // Does not exist in storage, nothing to delete.
+                            // If there was an insert in this batch before, we need to skip it.
+                            Integer prevRowIdx = newKeyMap.remove(pk.byteBuffer());
```
ascherbakoff, line 2379: The future can be completed on another thread, but newKeyMap is not thread safe.
Same issue for insert. If this code safe ?

ptupitsyn: You are right, it is not safe, and also not efficient.

Reworked the code to detect duplicate keys in a separate loop before taking any locks or doing any async work.

ascherbakoff (approved):

PR #3737, opened 2024-05-10
IGNITE-21568 Java thin: Pass client time zone to server

ptupitsyn:


modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/sql/ClientSqlProperties.java
```diff
@@ -68,6 +75,10 @@ SqlProperties toSqlProps() {
             builder.set(QueryProperty.DEFAULT_SCHEMA, schema);
         }
 
+        if (timeZoneId != null) {
+            builder.set(QueryProperty.TIME_ZONE_ID, ZoneId.of(timeZoneId));
+        }
```
isapego, line 80: How are we going to specify or reference the timezone format in other clients? Just give a reference to `java.time.ZoneId` javadocs?

isapego (approved):
Looks good overall.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/sql/ClientSqlProperties.java
ptupitsyn: Yes, the javadoc explains it in detail: [ZoneId#of](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/ZoneId.html#of(java.lang.String))

It supports standard [IANA region-based zone IDs](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones), so I expect every client to have something suitable in the standard library or a popular package.

modules/client/src/main/java/org/apache/ignite/internal/client/sql/ClientStatement.java
```diff
@@ -31,14 +31,20 @@ public class ClientStatement implements Statement {
     private final String query;
 
     /** Default schema. */
+    @Nullable
```
korlov42, line 34: according to our [coding guidelines](https://google.github.io/styleguide/javaguide.html#s4.8.5-annotations), type-use annotations appear immediately before the annotated type. Please fix it in other places too

korlov42 (approved):
the patch looks good in general. I left minor comment regarding styles

modules/client/src/main/java/org/apache/ignite/internal/client/sql/ClientStatement.java
ptupitsyn: Fixed.

I've used `javax.annotation.Nullable` instead of `org.jetbrains.annotations.Nullable` by mistake, and the IDE displays a warning about incorrect order in that case.

PR #3740, opened 2024-05-10
IGNITE-21801 Add missing Schema Sync integration tests

rpuch:
https://issues.apache.org/jira/browse/IGNITE-21801

### DIFF
--- a/modules/runner/src/integrationTest/java/org/apache/ignite/internal/schemasync/ItSchemaSyncConfigTest.java
+++ b/modules/runner/src/integrationTest/java/org/apache/ignite/internal/schemasync/ItSchemaSyncConfigTest.java
@@ -1,0 +1,79 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.schemasync;
+
+import static org.apache.ignite.internal.testframework.asserts.CompletableFutureAssert.assertWillThrowFast;
+import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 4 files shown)


modules/runner/src/integrationTest/java/org/apache/ignite/internal/schemasync/ItSchemaSyncConfigTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+import org.apache.ignite.configuration.ConfigurationChangeException;
+import org.apache.ignite.configuration.validation.ConfigurationValidationException;
+import org.apache.ignite.internal.ClusterPerClassIntegrationTest;
+import org.apache.ignite.internal.app.IgniteImpl;
+import org.apache.ignite.internal.catalog.configuration.SchemaSynchronizationConfiguration;
+import org.junit.jupiter.api.Test;
+
+@SuppressWarnings("resource")
+class ItSchemaSyncConfigTest extends ClusterPerClassIntegrationTest {
```
tkalkirill, line 34: I think the test class is redundant, the configuration framework should test this behavior and there is no need to double-check it in other modules.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1013,22 +1013,27 @@ private CompletableFuture<List<BinaryRow>> retrieveExactEntriesUntilCursorEmpty(
...
 
             return allOf(futs).thenApply((unused) -> rows);
         });
     }
 
+    private CompletableFuture<Void> validateBackwardCompatibility(BinaryRow row, UUID txId) {
+        return schemaCompatValidator.validateBackwards(row.schemaVersion(), tableId(), txId)
+                .thenAccept(validationResult -> {
+                    if (!validationResult.isSuccessful()) {
+                        throw new IncompatibleSchemaException(String.format(
```
tkalkirill, line 1028: As a suggestion, we could add a constructor that would accept an error message pattern and varags, inside use `org.apache.ignite.internal.lang.IgniteStringFormatter#format`.

rpuch: I'm not sure such a change makes the situation so much better. The code at the exception creation site becomes just a bit shorter, but the constructor becomes more complex and now has an additional responsability. I would prefer to leave it in its current state.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/schemasync/ItSchemaSyncConfigTest.java
rpuch: This is an integration test, and it seems usual to test something at a higher level, even if you already tested it using unit tests. Also, this property ('that delayDuration can never be changed') is very important, it's currently a cornerstone for the Schema Sync to work, so a little bit of redundancy here seems to be worth having it.

tkalkirill (approved):

PR #3742, opened 2024-05-13
IGNITE-21604 .NET: Pass client time zone to server

ptupitsyn:
https://issues.apache.org/jira/browse/IGNITE-21604


isapego (approved):
Looks good to me.

PR #3753, opened 2024-05-14
IGNITE-21605 C++ Client: Pass client time zone to server

isapego:
- Added `sql_statement::timezone_id()`, so that user can set a TIMEZONE when executing SQL statements using C++ Client;
- Added `timezone` connection string argument for ODBC for the same purpose;
- By default, a server timezone is used, as there is no simple way to get a local Timezone ID in C++;
- Added tests.


modules/platforms/cpp/ignite/client/sql/sql_statement.h
```diff
@@ -132,6 +134,26 @@ class sql_statement {
...
+    [[nodiscard]] const std::string &timezone_id() const { return m_timezone_id; }
+
+    /**
+     * Sets the Timezone ID.
+     *
+     * Examples: "America/New_York", "UTC+3".
+     * Affects time-related SQL functions (e.g. @c CURRENT_TIME) and string literal conversions
+     * (e.g. <tt>TIMESTAMP WITH LOCAL TIME ZONE '1992-01-18 02:30:00.123'</tt>).
+     *
+     * If left empty, defaults to server time zone.
```
ptupitsyn, line 151: Empty or null?

ptupitsyn (approved):

modules/platforms/cpp/ignite/client/sql/sql_statement.h
isapego: Empty, only pointers in C++ can be `null`.

PR #3756, opened 2024-05-14
IGNITE-22241 Make ClusterManagementGroupManager extensible

tmgodinho:

### The Review Checklist
- [x] **Formal criteria:** TC status, codestyle, mandatory documentation. Also make sure to complete the following:  
- [x] **Design:** new code conforms with the design principles of the components it is added to.
- [x] **Patch quality:** patch cannot be split into smaller pieces, its size must be reasonable.
- [x] **Code quality:** code is clean and readable, necessary developer documentation is added if needed.
- [x] **Tests code quality:** test set covers positive/negative scenarios, happy/edge cases. Tests are effective in terms of execution time and resources.

### Notes

### DIFF
--- a/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
+++ b/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
@@ -23,6 +23,7 @@ import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 import static java.util.stream.Collectors.toUnmodifiableSet;
 import static org.apache.ignite.internal.cluster.management.ClusterTag.clusterTag;
+import static org.apache.ignite.internal.failure.FailureType.CRITICAL_ERROR;
 import static org.apache.ignite.internal.util.CompletableFutures.nullCompletedFuture;
 import static org.apache.ignite.internal.util.ExceptionUtils.unwrapCause;
 import static org.apache.ignite.internal.util.IgniteUtils.cancelOrConsume;
@@ -62,6 +63,8 @@ import org.apache.ignite.internal.cluster.management.topology.LogicalTopologyImp
 import org.apache.ignite.internal.cluster.management.topology.api.LogicalTopologySnapshot;
 import org.apache.ignite.internal.event.AbstractEventProducer;
 import org.apache.ignite.internal.event.EventParameters;
+import org.apache.ignite.internal.failure.FailureContext;
+import org.apache.ignite.internal.failure.FailureProcessor;
 import org.apache.ignite.internal.lang.IgniteInternalException;
 import org.apache.ignite.internal.lang.NodeStoppingException;
 import org.apache.ignite.internal.logger.IgniteLogger;
@@ -141,6 +144,9 @@ public class ClusterManagementGroupManager extends AbstractEventProducer<Cluster
     /** Future that resolves into the initial cluster configuration in HOCON format. */
     private final CompletableFuture<String> initialClusterConfigurationFuture = new CompletableFuture<>();
 
+    /** Failure processor that is used to handle critical errors. */
+    private final FailureProcessor failureProcessor;
+
     /** Constructor. */
     public ClusterManagementGroupManager(
             VaultManager vault,
@@ -150,7 +156,8 @@ public class ClusterManagementGroupManager extends AbstractEventProducer<Cluster
             ClusterStateStorage clusterStateStorage,
             LogicalTopology logicalTopology,
             ClusterManagementConfiguration configuration,
-            NodeAttributes nodeAttributes
+            NodeAttributes nodeAttributes,
+            FailureProcessor failureProcessor
     ) {
         this.clusterService = clusterService;
         this.clusterInitializer = clusterInitializer;
@@ -160,6 +167,7 @@ public class ClusterManagementGroupManager extends AbstractEventProducer<Cluster
         this.configuration = configuration;
         this.localStateStorage = new LocalStateStorage(vault);
         this.nodeAttributes = nodeAttributes;
+        this.failureProcessor = failureProcessor;
 
         scheduledExecutor = Executors.newSingleThreadScheduledExecutor(
                 NamedThreadFactory.create(clusterService.nodeName(), "cmg-manager", LOG)
@@ -442,13 +450,11 @@ public class ClusterManagementGroupManager extends AbstractEventProducer<Cluster
         LOG.info("CMG cancellation procedure started");
         return inBusyLockAsync(busyLock,
                 () -> fireEvent(ClusterManagerGroupEvent.BEFORE_DESTROY_RAFT_GROUP, EmptyEventParameters.INSTANCE)
-                    .whenCompleteAsync((none, err) -> {
-                        if (err != null) {
-                            LOG.error("Error while waiting for destroy listeners", err);
-                        }
-
-                        destroyCmg();
-                    }, this.scheduledExecutor)
+                    .thenRunAsync(this::destroyCmg, this.scheduledExecutor)
+                    .exceptionally(err -> {
+                        failureProcessor.process(new FailureContext(CRITICAL_ERROR, err));
+                        throw (err instanceof RuntimeException) ? (RuntimeException) err : new CompletionException(err);
+                    })
         );
     }
 
--- a/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
+++ b/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
@@ -726,9 +726,7 @@ public class ClusterManagementGroupManager extends AbstractEventProducer<Cluster
 
         initialClusterConfigurationFuture.completeExceptionally(new NodeStoppingException());
 
-        fireEvent(ClusterManagerGroupEvent.AFTER_STOP_RAFT_GROUP, EmptyEventParameters.INSTANCE);
-
-        return nullCompletedFuture();
+        return fireEvent(ClusterManagerGroupEvent.AFTER_STOP_RAFT_GROUP, EmptyEventParameters.INSTANCE);
     }
 
     /**
--- a/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
+++ b/modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
@@ -27,6 +27,7 @@ import static org.apache.ignite.internal.util.CompletableFutures.nullCompletedFu
 import static org.apache.ignite.internal.util.ExceptionUtils.unwrapCause;
 import static org.apache.ignite.internal.util.IgniteUtils.cancelOrConsume;
 import static org.apache.ignite.internal.util.IgniteUtils.inBusyLock;
+import static org.apache.ignite.internal.util.IgniteUtils.inBusyLockAsync;
 
 import java.util.Collection;
 import java.util.List;
@@ -38,6 +39,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
 import java.util.function.Function;
 import org.apache.ignite.internal.cluster.management.LocalStateStorage.LocalState;
 import org.apache.ignite.internal.cluster.management.configuration.ClusterManagementConfiguration;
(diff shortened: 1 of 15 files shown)


tmgodinho:
Hey @rpuch and @sashapolo, would you mind having a look at these changes?
Do you think it's necessary to introduce some tests to check if the events are fired? Didn't seem very error-prone to me.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -434,6 +441,12 @@ private void handleCancelInit(CancelInitMessage msg) {
      * Completely destroys the local CMG Raft service.
      */
     private void destroyCmg() {
+        try {
+            fireEvent(ClusterManagerGroupEvent.BEFORE_RAFT_GROUP_DESTROY, EmptyEventParams.INSTANCE).join();
```
sashapolo, line 445: It's better to use `get` with a timeout. Also, this method is called inside a network thread, we should probably dispatch it onto a thread pool or make it asynchronous.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/events/ClusterManagerGroupEvent.java
```diff
@@ -0,0 +1,32 @@
...
+import org.apache.ignite.internal.cluster.management.ClusterManagementGroupManager;
+import org.apache.ignite.internal.event.Event;
+
+/** Enum with events for the {@link ClusterManagementGroupManager}. */
+
+public enum ClusterManagerGroupEvent implements Event {
+    /** Fired before starting the cmg raft group. */
+    BEFORE_START_RAFT_GROUP,
+    /** Fired before cleaning the local state. */
+    BEFORE_RAFT_GROUP_DESTROY,
```
sashapolo, line 29: It would be more consistent to call this event `BEFORE_DESTROY_RAFT_GROUP`. Or you can call the other two `BEFORE_RAFT_GROUP_START` and `AFTER_RAFT_GROUP_STOP`.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/events/ClusterManagerGroupEvent.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.apache.ignite.internal.cluster.management.ClusterManagementGroupManager;
+import org.apache.ignite.internal.event.Event;
+
+/** Enum with events for the {@link ClusterManagementGroupManager}. */
+
+public enum ClusterManagerGroupEvent implements Event {
+    /** Fired before starting the cmg raft group. */
+    BEFORE_START_RAFT_GROUP,
+    /** Fired before cleaning the local state. */
```
sashapolo, line 28: This javadoc is misleading.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/events/ClusterManagerGroupEvent.java
```diff
@@ -0,0 +1,32 @@
...
+    BEFORE_RAFT_GROUP_DESTROY,
+    /** Fired after the service is stopped. */
```
sashapolo, line 30: This javadoc should also be updated, I guess.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/events/EmptyEventParams.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.cluster.management.events;
+
+import org.apache.ignite.internal.event.EventParameters;
+
+/** Empty Event Paramter Singleton. */
```
sashapolo, line 22:
```suggestion
/** Empty Event Parameters Singleton. */
```

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/events/EmptyEventParams.java
```diff
@@ -0,0 +1,29 @@
...
+public class EmptyEventParams implements EventParameters {
```
sashapolo, line 23: I would prefer this class to be called `EmptyEventParameters`, because all other similar classes do not use `Params` in their names.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/events/BeforeStartRaftGroupEventParameters.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Set;
+import org.apache.ignite.internal.event.EventParameters;
+import org.jetbrains.annotations.Nullable;
+
+/** Transparent data container for the {@link ClusterManagerGroupEvent#BEFORE_START_RAFT_GROUP}. */
+public class BeforeStartRaftGroupEventParameters implements EventParameters {
+    private final Set<String> nodeNames;
+    private final @Nullable String initialClusterConfig;
+
+    /** Constructor. */
```
sashapolo, line 30: Can you please expand this javadoc a little bit? For example, what does "initialClusterConfig" mean and why can it be `null`?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/events/BeforeStartRaftGroupEventParameters.java
```diff
@@ -0,0 +1,44 @@
...
+    public BeforeStartRaftGroupEventParameters(Set<String> nodeNames, @Nullable String initialClusterConfig) {
+        this.nodeNames = new HashSet<>(nodeNames);
```
sashapolo, line 32: Probably `Set.copyOf` would be better here, but that's not a big deal

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
rpuch: How about not using get/join at all? Just collect the listeners' futures to one using `CompletableFuture#allOf()` and than on that aggregated future make `.thenRun()` doing the rest of the method; `destroyCmg()` would return a `CompletableFuture` as well; the code that calls it and needs to execute some logic strictly after the CMG has been destroyed can just chain on the future it returns.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -700,6 +726,8 @@ public CompletableFuture<Void> stopAsync() {
 
         initialClusterConfigurationFuture.completeExceptionally(new NodeStoppingException());
 
+        fireEvent(ClusterManagerGroupEvent.AFTER_STOP_RAFT_GROUP, EmptyEventParameters.INSTANCE);
```
sashapolo, line 729: Do we need to return this future from the method? So that the component would not stop until all listeners finish their work.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
tmgodinho: Currently, there's no code after `destoryCmg` just the error handling message and the busyLock unlock.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
tmgodinho: Yeah this was a bug

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
rpuch: In `handleClusterState()`, there is a call to `destroyCmg()`, and there is a `initCmgRaftService(state)` call that probably relies on the CMG being destroyed at the moment when it's made.

tmgodinho: Yeah, I saw that afterwards. Missed that one on the first iteration.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -221,7 +229,8 @@ public CompletableFuture<Void> startAsync() {
 
                         handleClusterState((ClusterStateMessage) message, sender, correlationId);
                     } else if (message instanceof CancelInitMessage) {
-                        handleCancelInit((CancelInitMessage) message);
+                        // I think the none of the cancel iniØt messages will have correlation id but...
```
sashapolo, line 232: I think we don't need to report any errors back here. You are right, none of these message have a correlation ID.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -424,10 +434,34 @@ private CompletableFuture<Void> updateLogicalTopology(CmgRaftService service) {
...
+            return null;
+        };
+    }
+
+    /** Delegates call to {@link #destroyCmg()} but fires the associated events. */
+    private CompletableFuture<Void> destroyCmgWithEvents() {
+        LOG.info("CMG cancellation procedure started");
+        return inBusyLockAsync(busyLock,
+                () -> fireEvent(ClusterManagerGroupEvent.BEFORE_DESTROY_RAFT_GROUP, EmptyEventParameters.INSTANCE)
+                    .handle((none, err) -> {
```
sashapolo, line 455: we should use `handleAsync` here, just in case

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -424,10 +434,34 @@ private CompletableFuture<Void> updateLogicalTopology(CmgRaftService service) {
...
     }
 
-    private void handleCancelInit(CancelInitMessage msg) {
+    private void handleCancelInit(CancelInitMessage msg, Consumer<Throwable> onError) {
         LOG.info("CMG initialization cancelled [reason={}]", msg.reason());
+        this.scheduledExecutor.execute(() -> this.destroyCmgWithEvents().exceptionally(toFunction(onError)));
+    }
+
+    /** Converts the consumer to a function that returns null. */
+    private static <T> Function<T, Void> toFunction(Consumer<T> consumer) {
```
sashapolo, line 443: I think this is a bit of an overkill, this method is only used in one place

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
```diff
@@ -424,10 +434,34 @@ private CompletableFuture<Void> updateLogicalTopology(CmgRaftService service) {
                 }));
     }
 
-    private void handleCancelInit(CancelInitMessage msg) {
+    private void handleCancelInit(CancelInitMessage msg, Consumer<Throwable> onError) {
         LOG.info("CMG initialization cancelled [reason={}]", msg.reason());
+        this.scheduledExecutor.execute(() -> this.destroyCmgWithEvents().exceptionally(toFunction(onError)));
```
sashapolo, line 439: Semantically speaking, `exceptionally` is used to translate exceptions, not to consume them. `whenComplete` would be a better choice here.

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
tmgodinho: Do you think its worth to keep the error handling code just to be complete or should I remove it??

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
tmgodinho: Hey @sashapolo sorry I didn't understand your idea.
Do you mean to use `handleAsync` with the default java threadpool or with `this.scheduledExecutor` just to bump the callback to the back of the executor's queue??

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
sashapolo: I would prefer to remove it, it's unnecessary

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterManagementGroupManager.java
sashapolo: I mean the latter, we try not to use the default thread pool in the server code. The idea is that event listeners return futures which may be executed on pools, unknown to the current component, so it's better to re-dispatch the `handle` method to be sure that we will not block listener's threads on accident.

rpuch (approved):

sashapolo (approved):

PR #3763, opened 2024-05-15
IGNITE-19234 Enable and fix group reentry logic for volatile storages

rpuch:
https://issues.apache.org/jira/browse/IGNITE-19234

The reentry logic (see IGNITE-16668) was erroneously disabled. This commit enables it back and contains fixes required to make it work (due to changes made to the common rebalancing code since it was disabled).

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/PartitionReplicatorNodeRecovery.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/PartitionReplicatorNodeRecovery.java
@@ -194,13 +194,13 @@ class PartitionReplicatorNodeRecovery {
         // with current partition.
         return waitForPeersAndQueryDataNodesCounts(tableId, partId, newConfiguration.peers())
                 .thenApply(dataNodesCounts -> {
-                    boolean fullPartitionRestart = dataNodesCounts.nodesSurelyEmpty == newConfiguration.peers().size();
+                    boolean fullPartitionRestart = dataNodesCounts.emptyNodes == newConfiguration.peers().size();
 
                     if (fullPartitionRestart) {
                         return true;
                     }
 
-                    boolean majorityAvailable = dataNodesCounts.nodesSurelyHavingData >= (newConfiguration.peers().size() / 2) + 1;
+                    boolean majorityAvailable = dataNodesCounts.nonEmptyNodes >= (newConfiguration.peers().size() / 2) + 1;
 
                     if (majorityAvailable) {
                         RebalanceUtilEx.startPeerRemoval(tablePartitionId, localMemberAssignment, metaStorageManager);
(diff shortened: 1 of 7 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/message/DataPresence.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.table.distributed.message;
+
+/**
+ * Whether a node has data or not (or it's not known).
+ */
+public enum DataPresence {
```
sashapolo, line 23: Can we serialize it into an int, not a string?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/message/DataPresence.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.table.distributed.message;
+
+/**
+ * Whether a node has data or not (or it's not known).
```
sashapolo, line 21: Can you please clarify a bit, what does "not known" mean?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/PartitionReplicatorNodeRecovery.java
```diff
@@ -285,12 +314,30 @@ private CompletableFuture<Long> queryDataNodesCount(Collection<Peer> peers, HasD
...
+                            .collect(toList());
+
+                    long nodesSurelyHavingData = hasDataFlags.stream().filter(presence -> presence == DataPresence.HAS_DATA).count();
+                    long nodesSurelyEmpty = hasDataFlags.stream().filter(presence -> presence == DataPresence.EMPTY).count();
+                    return new DataNodesCounts(nodesSurelyHavingData, nodesSurelyEmpty);
+                });
+    }
+
+    private static class DataNodesCounts {
+        private final long nodesSurelyHavingData;
```
sashapolo, line 335: These fields' names are a bit of a mouthful. Can we rename them to something like "dataNodesCnt" and "emptyNodesCnt"?

rpuch: Is it a problem? As a result, the code that uses them is self-explanatory. On the other hand, `dataNodesCnt` does not tell you that we actually checked that there is the given count of replicas *available*.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/message/DataPresence.java
sashapolo: I don't mean to use ordinals, I mean to use an explicit int value. Don't we have a similar approach in other enums? Anyway, that's up to you

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/PartitionReplicatorNodeRecovery.java
sashapolo: For me, actually, the current naming was very confusing for some reason. I had to re-read them multiple times. If you don't like my suggestion, maybe you could provide something simpler?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/message/DataPresence.java
rpuch: Added clarifications

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/message/DataPresence.java
rpuch: Maybe I just missed this, but I did not see this approach in use for network messages. I still prefer strings for the reason I described earlier when the performance penalty is small (like in this case).

sashapolo (approved):

PR #3765, opened 2024-05-15
IGNITE-22245 Force switch write intents

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-22245

### DIFF
--- a/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
+++ b/modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
@@ -1837,7 +1837,7 @@ public class PartitionReplicaListener implements ReplicaListener {
                 return null;
             }
 
-            // Update futures are empty in two cases:
+            // Cleanup futures (both read and update) are empty in two cases:
             // - there were no actions in the transaction
             // - write intent switch is being executed on the new primary (the primary has changed after write intent appeared)
             // Both cases are expected to happen extremely rarely so we are fine to force the write intent switch.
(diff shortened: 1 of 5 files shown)


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1830,11 +1830,22 @@ private CompletableFuture<FuturesCleanupResult> awaitCleanupReadyFutures(UUID tx
...
+        AtomicBoolean forceCleanup = new AtomicBoolean(true);
+
         txCleanupReadyFutures.compute(txId, (id, txOps) -> {
             if (txOps == null) {
                 return null;
             }
 
+            // Update futures are empty in two cases:
+            // - there were no actions in the transaction
+            // - write intent switch is being executed on the new primary (the primary has changed after the transaction started)
```
denis-chudov, line 1842:
```suggestion
            // - write intent switch is being executed on the new primary (the primary has changed after write intent appeared)
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1830,11 +1830,22 @@ private CompletableFuture<FuturesCleanupResult> awaitCleanupReadyFutures(UUID tx
...
 
+        AtomicBoolean forceCleanup = new AtomicBoolean(true);
+
         txCleanupReadyFutures.compute(txId, (id, txOps) -> {
             if (txOps == null) {
                 return null;
             }
 
+            // Update futures are empty in two cases:
+            // - there were no actions in the transaction
```
sanpwc, line 1841: Formally it's not true, update futures will be empty if  "there were no **write** actions in the transaction". However, the logic itself is correct because 'txOps.futures' is used instead of updateFutures. So, please, adjust the javadoc.

sanpwc (approved):

PR #3772, opened 2024-05-15
IGNITE-15568 Striped Disruptor doesn't work with JRaft event handlers…

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-15568

### DIFF
--- a/modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
+++ b/modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.apache.ignite.raft.jraft.disruptor;
 
 import static org.apache.ignite.internal.lang.IgniteStringFormatter.format;
@@ -27,6 +28,8 @@ import com.lmax.disruptor.YieldingWaitStrategy;
 import com.lmax.disruptor.dsl.Disruptor;
 import com.lmax.disruptor.dsl.ProducerType;
 import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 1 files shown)


modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
```diff
@@ -149,12 +152,12 @@ public StripedDisruptor(
             String stripeName = format("{}_stripe_{}", poolName, i);
 
             Disruptor<T> disruptor = DisruptorBuilder.<T>newInstance()
-                .setRingBufferSize(bufferSize)
-                .setEventFactory(eventFactory)
-                .setThreadFactory(threadFactorySupplier.apply(nodeName, stripeName))
-                .setProducerType(ProducerType.MULTI)
-                .setWaitStrategy(useYieldStrategy ? new YieldingWaitStrategy() : new BlockingWaitStrategy())
-                .build();
+                    .setRingBufferSize(bufferSize)
```
ascherbakoff, line 155: Formatting seems incorrect

modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
```diff
@@ -121,7 +124,7 @@ public StripedDisruptor(
      * @param eventFactory Event factory for the Striped disruptor.
      * @param stripes Amount of stripes.
      * @param supportsBatches If {@code false}, this stripe will always pass {@code true} into
-     *      {@link EventHandler#onEvent(Object, long, boolean)}. Otherwise, the data will be provided with batches.
+     *         {@link EventHandler#onEvent(Object, long, boolean)}. Otherwise, the data will be provided with batches.
```
ascherbakoff, line 127: Formatting seems incorrect

modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
```diff
@@ -289,10 +301,14 @@ private class StripeEntryHandler implements EventHandler<T> {
          */
         StripeEntryHandler(int stripeId) {
             this.stripeId = stripeId;
+
+            eventCache = supportsBatches ? null : new HashMap<>();
```
ascherbakoff, line 305: Can we have different implementations of `EventHandler `for manual and build-in batching ?

ascherbakoff (changes requested):

modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
vldpyatkov: I applied fromat that we use for all cod in our project, except the ignite-raft.
So you think this is because our format never applied to the code before, but it is right.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
vldpyatkov: The same as above. I have checked the results again.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/disruptor/StripedDisruptor.java
vldpyatkov: We don't have a different implementation.
The flag is used in the only scenario where the log storage is shared. For the shared log storage, we can handle the last messages (for the current batch) at the storage level. Althougр, this message might be intended for another handler in the same stripe.

PR #3779, opened 2024-05-16
IGNITE-22094 Add removeAll method to tx state storage

denis-chudov:
https://issues.apache.org/jira/browse/IGNITE-22094


modules/transactions/src/testFixtures/java/org/apache/ignite/internal/tx/storage/state/AbstractTxStateStorageTest.java
```diff
@@ -135,6 +158,45 @@ private HybridTimestamp generateTimestamp(UUID uuid) {
         return hybridTimestamp(time);
     }
 
+    @Test
+    public void testRemoveAll() {
```
sanpwc, line 162: Do we check that lastAppliedIndex and lastAppliedTerm were properly updated in case of removeAll?

sanpwc (approved):

PR #3795, opened 2024-05-20
IGNITE-22164 Sql. Fixed conversion from SqlMerge AST to relational expression when columns defined in different order.

xtern:
https://issues.apache.org/jira/browse/IGNITE-22164

Issue description.

Schema with 2 tables.
```
TABLE TEST1(C1 INT, C2 INT)

TABLE TEST2(C1 INT, C2 INT)
```

`SqlToRelConverter` for the following insert statement
```
INSERT INTO TEST1 (C1, C2) SELECT C1, C2 FROM TEST2
```

produces the following rel node.
```
LogicalTableModify(table=[[PUBLIC, TEST1]], operation=[INSERT], flattened=[false])
  LogicalProject(C1=[$0], C2=[$1])
    LogicalTableScan(table=[[PUBLIC, TEST2]])
```

But if we swap columns C1 and C2.
```
INSERT INTO TEST1 (C2, C1) SELECT C2, C1 FROM TEST2
```
the result will be
```
LogicalTableModify(table=[[PUBLIC, TEST1]], operation=[INSERT], flattened=[false])
  LogicalTableScan(table=[[PUBLIC, TEST2]])
```

`convertMerge()` expects projection in the input of table modify node and fails with the following
```
Caused by: java.lang.ClassCastException: class org.apache.ignite.internal.sql.engine.rel.logical.IgniteLogicalTableScan cannot be cast to class org.apache.calcite.rel.logical.LogicalProject
```

Calcite issue: https://issues.apache.org/jira/browse/CALCITE-6412

As a naive workaround, an additional assembly of column expressions has been added to `convertMerge()` :thinking:

### DIFF
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlToRelConvertor.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlToRelConvertor.java
@@ -270,15 +270,17 @@ public class IgniteSqlToRelConvertor extends SqlToRelConverter {
             if (input instanceof LogicalProject) {
                 level1InsertExprs = ((LogicalProject) input).getProjects();
             } else {
-                // TODO calcite issue
+                // TODO https://issues.apache.org/jira/browse/IGNITE-22293
+                // convertInsert() may return LogicalTableModify without projection in the input.
+                // As a workaround for this case, we additionally build required column expressions.
                 RelDataType rowType = input.getRowType();
                 RexNode sourceRef = rexBuilder.makeRangeReference(rowType, 0, false);
                 ArrayList<String> targetColumnNames = new ArrayList<>(rowType.getFieldCount());
-                ArrayList<RexNode> projects = new ArrayList<>(rowType.getFieldCount());
+                ArrayList<RexNode> columnExprs = new ArrayList<>(rowType.getFieldCount());
 
-                collectInsertTargets(insertCall, sourceRef, targetColumnNames, projects);
+                collectInsertTargets(insertCall, sourceRef, targetColumnNames, columnExprs);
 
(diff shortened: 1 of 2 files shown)


korlov42 (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/IgniteSqlToRelConvertor.java
```diff
@@ -265,14 +265,29 @@ private RelNode convertMerge(SqlMerge call) {
...
+            } else {
+                // TODO https://issues.apache.org/jira/browse/IGNITE-22293
+                // convertInsert() may return LogicalTableModify without projection in the input.
+                // As a workaround for this case, we additionally build required column expressions.
+                RelDataType rowType = input.getRowType();
+                RexNode sourceRef = rexBuilder.makeRangeReference(rowType, 0, false);
+                ArrayList<String> targetColumnNames = new ArrayList<>(rowType.getFieldCount());
+                ArrayList<RexNode> columnExprs = new ArrayList<>(rowType.getFieldCount());
+
+                collectInsertTargets(insertCall, sourceRef, targetColumnNames, columnExprs);
```
zstan, line 281: seems all this stuff can be changed for more lightweight:
```
                ArrayList<RexNode> columnExprs = new ArrayList<>(rowType.getFieldCount());
                int pos = 0;
                for (RelDataTypeField type : rowType.getFieldList()) {
                    columnExprs.add(rexBuilder.makeInputRef(type.getType(), pos++));
                }
```

xtern: Fixed, thank you.

zstan (approved):

PR #3799, opened 2024-05-21
IGNITE-22288 Fixed the vacuum of persistent tx state before tx cleanup completion

denis-chudov:
https://issues.apache.org/jira/browse/IGNITE-22288


modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/PersistentTxStateVacuumizer.java
```diff
@@ -95,15 +100,29 @@ public CompletableFuture<Set<UUID>> vacuumPersistentTxStates(Map<TablePartitionI
                         // timestamp) would be updated there, and then this operation would be called from there.
                         // Also, we are going to send the vacuum request only to the local node.
                         if (replicaMeta != null && localNode.id().equals(replicaMeta.getLeaseholderId())) {
+                            // Filter txns to define those which persistent states can be vacuumized.
+                            Set<UUID> filteredTxIds = new HashSet<>();
+
+                            for (VacuumizableTx v : txs) {
+                                if (v.cleanupCompletionTimestamp == null) {
```
sanpwc, line 107: That look suspicious, we should remove all records from volatile with expired ttl. For those that have !null cleanupCompletionTimestamp we should remove persistent state prior to the removing volatile one. Removing state from the volatile map should be atomic in a way that we should only remove the records if corresponding cleanupCompletionTimestamp has the same value (including null) as it was when we defined whether we need to remove the entry from the persistent storage.

I see the union `successful.addAll(filteredTxIds)` but didn't manage to find conditional removing? Do we have it?

denis-chudov: This condition and all the checks are located in `VolatileTxStateMetaStorage#vacuum` (see `persistentVacuumOp.apply(txIds).thenAccept` )
It removes all tx states successfully processed by persistent vacuumizer, taking the cleanup completion timestamp into account.

sanpwc (approved):

PR #3800, opened 2024-05-21
IGNITE-22214 Add persistence and recovery for an idempotent command cache

sanpwc:
https://issues.apache.org/jira/browse/IGNITE-22214

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/ArrayUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/ArrayUtils.java
@@ -338,6 +338,25 @@ public final class ArrayUtils {
     }
 
     /**
+     * Concatenates an elements to an array.
+     *
+     * @param arr Array.
+     * @param bytes One or more elements.
+     * @return Concatenated array.
+     */
+    public static byte[] concat(@Nullable byte[] arr, byte... bytes) {
+        if (nullOrEmpty(arr)) {
+            return bytes;
+        }
+
+        byte[] newArr = Arrays.copyOf(arr, arr.length + bytes.length);
+
+        System.arraycopy(bytes, 0, newArr, arr.length, bytes.length);
+
+        return newArr;
+    }
+
+    /**
      * Removes an element from an array with decrementing the array itself.
      *
      * @param arr Array.
--- a/modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItIdempotentCommandCacheTest.java
+++ b/modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItIdempotentCommandCacheTest.java
@@ -19,7 +19,9 @@ package org.apache.ignite.internal.metastorage.impl;
 
 import static java.util.stream.Collectors.toSet;
 import static org.apache.ignite.internal.metastorage.dsl.Conditions.notExists;
+import static org.apache.ignite.internal.metastorage.dsl.Operations.ops;
 import static org.apache.ignite.internal.metastorage.dsl.Operations.put;
+import static org.apache.ignite.internal.metastorage.dsl.Statements.iif;
 import static org.apache.ignite.internal.network.utils.ClusterServiceTestUtils.clusterService;
 import static org.apache.ignite.internal.testframework.IgniteTestUtils.waitForCondition;
 import static org.apache.ignite.internal.testframework.matchers.CompletableFutureMatcher.willCompleteSuccessfully;
@@ -55,10 +57,13 @@ import org.apache.ignite.internal.lang.NodeStoppingException;
 import org.apache.ignite.internal.manager.IgniteComponent;
 import org.apache.ignite.internal.metastorage.Entry;
 import org.apache.ignite.internal.metastorage.MetaStorageManager;
+import org.apache.ignite.internal.metastorage.command.IdempotentCommand;
 import org.apache.ignite.internal.metastorage.command.InvokeCommand;
 import org.apache.ignite.internal.metastorage.command.MetaStorageCommandsFactory;
 import org.apache.ignite.internal.metastorage.command.SyncTimeCommand;
 import org.apache.ignite.internal.metastorage.configuration.MetaStorageConfiguration;
+import org.apache.ignite.internal.metastorage.dsl.Iif;
+import org.apache.ignite.internal.metastorage.dsl.StatementResult;
 import org.apache.ignite.internal.metastorage.server.KeyValueStorage;
 import org.apache.ignite.internal.metastorage.server.persistence.RocksDbKeyValueStorage;
 import org.apache.ignite.internal.metastorage.server.raft.MetastorageGroupId;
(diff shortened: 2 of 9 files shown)


modules/core/src/main/java/org/apache/ignite/internal/util/ArrayUtils.java
```diff
@@ -337,6 +337,25 @@ public static long[] concat(@Nullable long[] arr, long... longs) {
         return newArr;
     }
 
+    /**
+     * Concatenates an elements to an array.
```
denis-chudov, line 341:
```suggestion
     * Concatenates elements to an array.
```

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItIdempotentCommandCacheTest.java
```diff
@@ -334,8 +325,45 @@ public void testIdempotentInvokeAfterLeaderChange() {
...
+
+        // Run same idempotent command one more time and check that condition wasn't re-evaluated, but was retrieved from the cache instead.
+        CompletableFuture<Object> commandProcessingResultFuture2 = raftClient().run(idempotentCommand);
+        assertThat(commandProcessingResultFuture2, willCompleteSuccessfully());
+        Object commandProcessingResult2 = commandProcessingResultFuture2.get();
+        if (commandProcessingResult2 instanceof Boolean) {
+            assertTrue((Boolean) commandProcessingResult2);
+        } else {
+            assertTrue(((StatementResult) commandProcessingResult2).getAsBoolean());
+        }
```
denis-chudov, line 366: What about checking the actual values in the storage?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/raft/MetaStorageWriteHandler.java
```diff
@@ -310,6 +319,28 @@ boolean beforeApply(Command command) {
...
 
+    /**
+     * The callback that is called right after storage is updated with a snapshot.
+     */
+    void onSnapshotLoad() {
+        byte[] keyFrom = IDEMPOTENT_COMMAND_PREFIX_BYTES;
+        byte[] keyTo = storage.nextKey(IDEMPOTENT_COMMAND_PREFIX_BYTES);
+
+        Cursor<Entry> cursor = storage.range(keyFrom, keyTo);
+        // It's fine to loose original command start time - in that case we will store the entry a little bit longer that necessary.
```
denis-chudov, line 330:
```suggestion
        // It's fine to lose original command start time - in that case we will store the entry a little bit longer that necessary.
```

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/WatchProcessor.java
```diff
@@ -171,8 +173,18 @@ public CompletableFuture<Void> notifyWatches(List<Entry> updatedEntries, HybridT
...
 
+                    List<Entry> filteredUpdatedEntries = updatedEntries.stream()
+                            .filter(entry ->
+                                    entry.key().length <= IDEMPOTENT_COMMAND_PREFIX_BYTES.length
+                                            ||
+                                            entry.key().length > IDEMPOTENT_COMMAND_PREFIX_BYTES.length
+                                                    && !ByteBuffer.wrap(entry.key(), 0, IDEMPOTENT_COMMAND_PREFIX_BYTES.length)
+                                                            .equals(ByteBuffer.wrap(IDEMPOTENT_COMMAND_PREFIX_BYTES)))
+                            .collect(Collectors.toList());
+
```
denis-chudov, line 184: Is this necessary? I mean, if there are no subscriptions on IDEMPOTENT_COMMAND_PREFIX_BYTES these entries would be just ignored, right?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
```diff
@@ -703,7 +719,14 @@ public StatementResult invoke(If iif, HybridTimestamp opTs) {
                 if (branch.isTerminal()) {
                     Update update = branch.update();
 
-                    applyOperations(update.operations(), opTs);
+                    Collection<Operation> ops = new ArrayList<>(update.operations());
+
+                    ops.add(Operations.put(
+                            new ByteArray(ArrayUtils.concat(IDEMPOTENT_COMMAND_PREFIX_BYTES, ByteUtils.toBytes(commandId))),
+                            update.result().result())
```
denis-chudov, line 726: While restoring from snapshot, you assume that `update.result().result()` would always be 1-byte boolean. Lets add the assertion here.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/raft/MetaStorageWriteHandler.java
```diff
@@ -310,6 +319,28 @@ boolean beforeApply(Command command) {
...
+        HybridTimestamp now = clusterTime.now();
+
+        try (cursor) {
+            for (Entry entry : cursor) {
+                if (!entry.tombstone()) {
+                    byte[] commandIdBytes = copyOfRange(entry.key(), IDEMPOTENT_COMMAND_PREFIX_BYTES.length, entry.key().length);
+                    CommandId commandId = ByteUtils.fromBytes(commandIdBytes);
+                    idempotentCommandCache.put(commandId, new IdempotentCommandCachedResult(byteToBoolean(entry.value()[0]), now));
```
denis-chudov, line 338: Let's add an assertion that the length of `entry.value()` is always 1.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/WatchProcessor.java
sanpwc: Yep, there are cases when we subscribe to all keys, definitely in some tests.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/raft/MetaStorageWriteHandler.java
sanpwc: It's possible to have both boolean and StatementResult there. Reworked the code.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/persistence/RocksDbKeyValueStorage.java
sanpwc: Please check my comment below.

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/impl/ItIdempotentCommandCacheTest.java
sanpwc: Added.

PR #3812, opened 2024-05-23
IGNITE-19544 Java thin: Add Data Streamer with Receiver

ptupitsyn:
Implement streamer with receiver according to [IEP-121](https://cwiki.apache.org/confluence/display/IGNITE/IEP-121%3A+Data+Streamer+with+Receiver).

* New client request `STREAMER_WITH_RECEIVER_BATCH_SEND`
* Use internal Compute job to run the receiver on target node for partition, reuse existing class loading and job scheduling
* Serialize receiver payload into a `BinaryTuple` so it can be passed from client handler to compute job without deserialization
* Require all streamed items to have the same type (so we can write the type once per batch)
* Result handling to be implemented separately: [IGNITE-22302](https://issues.apache.org/jira/browse/IGNITE-22302)

https://issues.apache.org/jira/browse/IGNITE-19544


modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/table/ClientStreamerWithReceiverBatchSendRequest.java
```diff
@@ -0,0 +1,129 @@
...
+
+            return table.partitionManager().primaryReplicaAsync(new HashPartition(partition)).thenCompose(primaryReplica -> {
+                // Use Compute to execute receiver on the target node with failover, class loading, scheduling.
+                JobExecution<List<Object>> jobExecution = compute.executeAsyncWithFailover(
+                        Set.of(primaryReplica),
+                        deploymentUnits,
+                        ReceiverRunnerJob.class.getName(),
+                        JobExecutionOptions.DEFAULT,
+                        payloadElementCount,
+                        payload);
```
isapego, line 81: This will result in the same processer class constructed and destructed for every batch, probably resulting in poorer performance and additional garbage. We should probably construct and deploy it once, and keep it alive while the streamer is working, if it's possible. We can still re-use compute API for that, probably.

isapego (approved):

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/table/ClientStreamerWithReceiverBatchSendRequest.java
ptupitsyn: Ticket created: https://issues.apache.org/jira/browse/IGNITE-22318

PR #3813, opened 2024-05-23
IGNITE-21982 Sql. Extend test coverage for SQL T021(BINARY and VARBINARY data types).

zstan:

### The Review Checklist

### Notes

### DIFF
--- a/modules/sql-engine/src/integrationTest/sql/types/blob/test_blob.test
+++ b/modules/sql-engine/src/integrationTest/sql/types/blob/test_blob.test
@@ -1,5 +1,6 @@
 # name: test/sql/types/blob/test_blob.test
 # description: BLOB tests
+# feature: T021(BINARY and VARBINARY data types)
 # group: [blob]
 
 statement ok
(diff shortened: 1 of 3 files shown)


modules/sql-engine/src/integrationTest/sql/types/blob/test_blob_cast.test
```diff
@@ -63,3 +63,5 @@ SELECT 1::bigint::VARBINARY
 statement error
 SELECT 1::decimal::VARBINARY
 
+statement error
+select CAST(x'Aa' AS integer);
```
AMashenkov, line 67: BTW, this valid valid in 2023 SQL standard.
ISO/IEC 9075-2:2023 as the optional feature T661, “Non-decimal integer literals”

modules/sql-engine/src/integrationTest/sql/types/blob/test_blob.test
```diff
@@ -87,3 +88,55 @@ SELECT CAST(x'0101' AS BINARY(0))
...
+
+query T
+SELECT POSITION(x'b0' IN x'0B03B0');
+----
+3
+
+query T
+SELECT POSITION(x'b0' IN x'0B03B0B0');
+----
+3
```
AMashenkov, line 142: Lets move length/octet_length test from `ItFunctionsTest` to SQL logic tests.

I see POSITION function tests for strings in sql/function/string/test_position.test.
Maybe we should go similar way and move function's tests into sql/function/blob

AMashenkov (approved):

modules/sql-engine/src/integrationTest/sql/types/blob/test_blob.test
zstan: position - done into new file,
octet_length - the only thing i can do here - is to remove 1 line with **octet_length** from ItFunctionsTest cause this one already tests in test_blob_function.test. But seems we need to talk here i think that to have some duplicates in It* tests and SqlLogic*tests it`s ok just because It* tests can be run locally and passed fast while SqlLogic are very time consumable

PR #3830, opened 2024-05-28
IGNITE-22214 Meta storage idempotent invokes: implement idempotent cache cleanup logic

sanpwc:
https://issues.apache.org/jira/browse/IGNITE-22217


modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/raft/MetaStorageWriteHandler.java
```diff
@@ -95,7 +112,14 @@ void handleWriteCommand(CommandClosure<WriteCommand> clo) {
         CommandClosure<WriteCommand> resultClosure;
 
         if (command instanceof IdempotentCommand) {
-            CommandId commandId = ((IdempotentCommand) command).id();
+            IdempotentCommand idempotentCommand = ((IdempotentCommand) command);
+            CommandId commandId = idempotentCommand.id();
+
+            // TODO: https://issues.apache.org/jira/browse/IGNITE-19417 Remove.
+            if (idempotentCommand.safeTime().getPhysical() % 100 == 0) {
```
denis-chudov, line 119: Is it some kind of randomization for eviction? Maybe 100 is too big number and eviction will happen rarely? Or we will rework it to gc trigger soon and it doesnt matter?

sanpwc: > Maybe 100 is too big number and eviction will happen rarely?

denis-chudov (approved):

PR #3831, opened 2024-05-28
IGNITE-21418 Add more logs to debug this issue

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-21418

### DIFF
--- a/modules/table/src/testFixtures/java/org/apache/ignite/internal/table/TxAbstractTest.java
+++ b/modules/table/src/testFixtures/java/org/apache/ignite/internal/table/TxAbstractTest.java
@@ -112,6 +112,9 @@ import org.apache.ignite.internal.tx.TxStateMeta;
 import org.apache.ignite.internal.tx.configuration.TransactionConfiguration;
 import org.apache.ignite.internal.tx.impl.IgniteTransactionsImpl;
 import org.apache.ignite.internal.tx.impl.ReadWriteTransactionImpl;
+import org.apache.ignite.internal.tx.impl.TxManagerImpl;
+import org.apache.ignite.internal.tx.impl.VolatileTxStateMetaStorage;
+import org.apache.ignite.internal.tx.storage.state.TxStateStorage;
 import org.apache.ignite.internal.type.NativeTypes;
 import org.apache.ignite.internal.util.CollectionUtils;
 import org.apache.ignite.internal.util.Lazy;
@@ -291,7 +294,11 @@ public abstract class TxAbstractTest extends IgniteAbstractTest {
 
     /** {@inheritDoc} */
     protected TxManager txManager(TableViewInternal t) {
-        TxManager manager = txTestCluster.txManagers().get(primaryNode(t));
+        String leaseHolder = primaryNode(t);
+
+        assertNotNull(leaseHolder, "Table primary node should not be null");
+
+        TxManager manager = txTestCluster.txManagers().get(leaseHolder);
 
         assertNotNull(manager);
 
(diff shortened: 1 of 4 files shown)


modules/table/src/testFixtures/java/org/apache/ignite/internal/table/TxAbstractTest.java
```diff
@@ -276,17 +291,21 @@ protected TxManager clientTxManager() {
 
     /** {@inheritDoc} */
     protected TxManager txManager(TableViewInternal t) {
+        TxManager manager = txTestCluster.txManagers().get(primaryNode(t));
```
sanpwc, line 294: Please add assertNotNull for primaryNode(t) result. Formally it's valid to have null as getPrimaryReplica result, but likely it's not the case within given tests.

modules/table/src/testFixtures/java/org/apache/ignite/internal/table/TxAbstractTest.java
```diff
@@ -477,6 +503,70 @@ private InternalTransaction deleteUpsertAll() {
...
+    private void printDebugInfo() {
+        logger().info("Primary node for accounts table is [nodeName={}]", primaryNode(accounts));
+        logger().info("Cluster times:");
+        long latestTime = IgniteTestUtils.getFieldValue(txTestCluster.clientClock(), HybridClockImpl.class, "latestTime");
+        logger().info("Client clock [time={}]", HybridTimestamp.hybridTimestamp(latestTime));
+
+        txTestCluster.clocks().forEach((name, hybridClock) -> {
+                    long time = IgniteTestUtils.getFieldValue(hybridClock, HybridClockImpl.class, "latestTime");
+                    logger().info(
+                            "Cluster clock [cluster_name={}, time={}]",
```
sanpwc, line 531: It's rather node_name and not the cluster one. Besides that we usually do not use underscore in variable names in logs. Anyway, who cares, it's a tmp log that'll be removed as soon as we will fix the problem.

modules/table/src/testFixtures/java/org/apache/ignite/internal/table/TxAbstractTest.java
```diff
@@ -477,6 +503,70 @@ private InternalTransaction deleteUpsertAll() {
...
+    private void printReplicaInfo(String name, ReplicationGroupId replicationGroupId, CompletableFuture<Replica> replicaCompletableFuture) {
+        Replica replica;
+        try {
+            replica = replicaCompletableFuture.get();
+        } catch (InterruptedException | ExecutionException e) {
+            throw new RuntimeException(e);
+        }
+
+        PartitionReplicaListener listener = IgniteTestUtils.getFieldValue(replica, Replica.class, "listener");
+        TestMvPartitionStorage storage = IgniteTestUtils.getFieldValue(listener, PartitionReplicaListener.class, "mvDataStorage");
```
sanpwc, line 557: Let's also print txnStateStorage records. It'll be useful in order to understand how writeIntent was resolved if any.

modules/table/src/testFixtures/java/org/apache/ignite/internal/table/TxAbstractTest.java
```diff
@@ -383,7 +402,7 @@ public void testRepeatedCommitRollbackAfterUpdateWithException() throws Exceptio
     }
 
     @Test
-    public void testRepeatedCommitRollbackAfterRollbackWithException() throws Exception {
+    public void testRepeatedCommitRollbackAfterRollbackWithException() {
```
sanpwc, line 405: Let's also cover testDeleteUpsertAllRollback. It's the test with highest failure rate.

sanpwc (commented):
Could you please share an example of logs output?

sanpwc (approved):

PR #3835, opened 2024-05-29
IGNITE-21466 Add metrics for partition states

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-21466

### The Review Checklist

### Notes


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/DisasterRecoveryManager.java
```diff
@@ -575,10 +595,8 @@ private static <T> boolean containsOrEmpty(T item, Collection<T> collection) {
         return collection.isEmpty() || collection.contains(item);
     }
 
-    /**
-     * Converts internal raft node state into public local partition state.
-     */
-    private static LocalPartitionStateEnum convertState(State nodeState) {
+    /** Converts internal raft node state into public local partition state. */
```
ibessonov, line 598: Why do you keep fixing formatting for comments that you have not changed? Text is the same, literally every letter is the same, I don't get it. We discussed it, like, 10 times already. Is this the impulse that you can't control? In that case I already offered you fixing the entire project, you rejected.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/PartitionStatesMetricSource.java
```diff
@@ -0,0 +1,152 @@
...
+
+    @Override
+    public synchronized boolean enabled() {
+        return enabled;
+    }
+
+    private long calculatePartitionCountByLocalState(LocalPartitionStateEnum state) {
+        long[] count = {0};
+
+        disasterRecoveryManager.raftManager.forEach((raftNodeId, raftGroupService) -> {
```
ibessonov, line 116: I think that this algorithm will be extremely inefficient. Loading metrics for all tables will be `O(n^2)`, which is pretty bad in general. I don't know if I should ask you to fix it now, but you should keep in mind that ideally we should do better than this. If you have better ideas about improving the approach, please reply here, we can discuss them.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/PartitionStatesMetricSource.java
```diff
@@ -0,0 +1,152 @@
...
+        return count[0];
+    }
+
+    private static LocalPartitionStateEnum localPartitionState(RaftGroupService raftGroupService) {
+        Node raftNode = raftGroupService.getRaftNode();
+
+        LocalPartitionStateEnum localState = convertState(raftNode.getNodeState());
+        long lastLogIndex = raftNode.lastLogIndex();
+
+        if (localState == HEALTHY) {
```
ibessonov, line 139: Looks like a copy-paste, maybe you should re-use existing code instead. If not, then please explain you reasoning, thank you

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/DisasterRecoveryManager.java
tkalkirill: I ask you to be more polite, and to be less emotional and more professional.
I said that I would stop formatting the text and making changes that did not affect my changes.
I changed this particular method a little for the task and therefore formatted the documentation, purely for beauty.
I’ll try to answer your question again in advance, no, I won’t make any rules or big changes for the entire project.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/PartitionStatesMetricSource.java
tkalkirill: I agree that it looks like a copy-paste, I’ll try to look at it again or comment.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/PartitionStatesMetricSource.java
tkalkirill: Today I thought that it would be great to have some kind of listener that would trigger a callback when the status of the node changes (To do this you will need to update the code to jraft), but then we would need to somehow manage this and keep in memory some amount of data that we don’t really need.

I looked more closely at making some kind of `Loza#raftGroupService` to make a loop through partitions, but we will either need to have 2 maps, or it will be more difficult to create `org.apache.ignite.internal.raft.RaftNodeId`.

On the other hand, we need such a metric only for disaster recovery and we may not worry about its performance for now, since we will go through cycles through the entities in memory, what do you think?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/PartitionStatesMetricSource.java
tkalkirill: try fix it

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/PartitionStatesMetricSource.java
ibessonov: > I don’t understand where the O(n^2), can you clearly indicate?

Never thought that I would have to do this, but here we go. Suppose you have `N` tables, each of them has an associated metrics source. For each source you have `5` metrics that show partition states. For each state you scan all `N` tables while looking for the one that you need (let's ignore partitions for simplicity). That's called a linear search and it's `O(N)`. If you multiply it by the number of metrics sources and the number of metrics, you will get `5*N*N` read operations per metrics accumulation, which is `O(N^2)`. This is what I meant.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/DisasterRecoveryManager.java
ibessonov: > purely for beauty.

We discussed this many times, and I thought that we had an agreement. Apparently we don't. And I feel stupid, because I have to repeat the same comment over and over again, without any success.
> I changed this particular method a little for the task and therefore formatted the documentation

Did you change the documentation?
Let me rephrase the question. Let's assume that you changed line number 15 in the code. What's the range of lines in the code that allow for changes with _very_ subjective value, the ones that I would call _useless_? Lines 14-16, or maybe 12-18? How do you decide that?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/DisasterRecoveryManager.java
```diff
@@ -556,35 +551,19 @@ private void handleLocalPartitionStatesRequest(LocalPartitionStatesRequest reque
...
-                        // Node without log didn't process anything yet, it's not really "healthy" before it accepts leader's configuration.
-                        if (lastLogIndex == 0) {
-                            localState = INITIALIZING;
-                        }
-
-                        if (raftNode.isInstallingSnapshot()) {
-                            localState = INSTALLING_SNAPSHOT;
-                        }
-                    }
+                    LocalPartitionStateEnumWithLogIndex localPartitionStateWithLogIndex = of(raftGroupService.getRaftNode());
```
ibessonov, line 558: I'm not a fan of static imports for methods called `of`, or `getInstance`, or something else non-descriptive. Please chose another name or don't use static import in this particular case, thank you

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/PartitionStatesMetricSource.java
tkalkirill: Let's figure it out.
N - the number of tables, which is equal to the number of metrics source.
I suggest adding 5 metrics.

For each metric, O(N) will be performed.
Those. for each metric source of the total will be O(5*N), where it will be N^2 I don’t understand.

ibessonov: > what do you think?

I think that eventually it will bite us in the rear. Avoiding listeners, the only good solution would be to ask a `TableManager` for all replicas of a given table, and to have it there already grouped by `tableId`. Then we should be able to access the raft group from the replica, that would be optimal and relatively simple.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/DisasterRecoveryManager.java
tkalkirill: I was wondering which name would be more appropriate, `convert` doesn’t seem to fit, maybe `toLocalPartitionStateWithLogIndex` or `getLocalPartitionStateWithLogIndex`, what do you think?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/DisasterRecoveryManager.java
tkalkirill: Yes, we discussed this more than once, for my part I feel violence and intolerance for a different view of things.

> Did you change the documentation?

No, I just changed the number of lines.

> What's the range of lines in the code that allow for changes with very subjective value, the ones that I would call useless?

In this question you mention about yourself, I cannot answer it.

> Lines 14-16, or maybe 12-18?

Don't know.

> How do you decide that?

For me, if a multi-line comment can fit into one, then it looks more expressive, but not 100%, it depends on the case and environment.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/PartitionStatesMetricSource.java
tkalkirill: Yes, but on the other hand, in some future we will switch to partition colocation, and then it will be possible to optimize this, what do you think?

ibessonov: > what do you think?

Same problem will be there for large numbers of zones, but I think that this situation would be rare. Originally I said I don't know if I should ask you to fix it now_ for this exact reason by the way

> for each metric source of the total will be O(5*N), where it will be N^2 I don’t understand.

What exactly was confusing in my explanation? I thought it was as details as possible. If you have `O(N)` metrics, and all of them require `O(N)` operations to be computed, then computing all of them would be `O(N*N)`. What else should I add?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/DisasterRecoveryManager.java
ibessonov: I think we can avoid using static import here and that would be fine. We don't use it for `List.of` and `Set.of`, and nobody is upset about it, right?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/DisasterRecoveryManager.java
ibessonov: > In this question you mention about yourself, I cannot answer it.

That was a question directed to you, maybe I phrased it poorly. But you already replied that you don't know the answer, so that's fine

> For me, if a multi-line comment can fit into one, then it looks more expressive, but not 100%, it depends on the case and environment.

This is what we discussed. That you won't be doing such changes, because it's all too subjective. It you have a desire to fix formatting of comments in the project - then do it in the entire project. Otherwise I'm against these changes in your PRs, and you know that.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/DisasterRecoveryManager.java
tkalkirill: Reasonable, I agree.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/PartitionStatesMetricSource.java
tkalkirill: > What exactly was confusing in my explanation?

I think that it is more correct to estimate the complexity of O(MxN), where N is the number of tables and M is the number of metrics?
O(N*N) is obtained if we load all metrics for all tables at once.
If we load metrics only for a specific one or for the entire table, then this will not happen.

I looked in the `TableManager` and did not find a suitable API for getting the state of the raft node; in `org.apache.ignite.internal.table.TableImpl` we store `org.apache.ignite.internal.raft.service.RaftGroupService`, which is different from the `org.apache.ignite.raft.jraft.RaftGroupService` that is currently in use.

To avoid O(N*N) as I described above, we can make metrics common to all tables (in the future zones) and then there will always be O(M*N). I don’t really like adding `org.apache.ignite.raft.jraft.RaftGroupService` to `org.apache.ignite.internal.table.TableImpl`, we can add `org.apache.ignite.internal.raft.RaftNodeId` to `TableImpl`.

Well, or leave it as is, what do you think?

ibessonov: > O(N*N) is obtained if we load all metrics for all tables at once.

Yes, that's precisely what I said. Quote: _"Loading metrics for all tables will be O(n^2)"_

> I looked in the TableManager and did not find a suitable API for getting the state of the raft node

That's right, it doesn't exist right now, but it could if we really needed it.

> Well, or leave it as is, what do you think?

We can leave it as is. My main point was that you should be aware of the complexity of your algorithms, and be sure that there's a good reason and it's safe to have O(N*N) code in some places. Ideally, such a code should be commented with explanation why it's OK. Otherwise readers will have same questions as I do

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/disaster/DisasterRecoveryManager.java
tkalkirill: fix it

ibessonov (approved):

PR #3837, opened 2024-05-29
IGNITE-22061 Use constant value when estimating decimal size.

xtern:
https://issues.apache.org/jira/browse/IGNITE-22061

### DIFF
--- a/modules/schema/src/main/java/org/apache/ignite/internal/schema/marshaller/reflection/ObjectStatistics.java
+++ b/modules/schema/src/main/java/org/apache/ignite/internal/schema/marshaller/reflection/ObjectStatistics.java
@@ -19,12 +19,15 @@ package org.apache.ignite.internal.schema.marshaller.reflection;
 
 import static org.apache.ignite.internal.schema.marshaller.MarshallerUtil.getValueSize;
 
+import java.math.BigDecimal;
 import java.util.List;
+import org.apache.ignite.internal.binarytuple.BinaryTupleCommon;
 import org.apache.ignite.internal.marshaller.Marshaller;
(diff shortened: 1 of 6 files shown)


modules/schema/src/main/java/org/apache/ignite/internal/schema/marshaller/reflection/ObjectStatistics.java
```diff
@@ -19,33 +19,51 @@
...
     /** Cached zero statistics. */
-    private static final ObjectStatistics ZERO_STATISTICS = new ObjectStatistics(-1);
+    private static final ObjectStatistics ZERO_STATISTICS = new ObjectStatistics(-1, false);
+
+    /**
+     * The average size of a decimal number is taken from the air. This value is used to avoid unnecessary
+     * buffer reallocation when building a binary row. The exact size of the decimal value can be determined
+     * after compaction (see {@link BinaryTupleCommon#shrinkDecimal(BigDecimal, int)}.
+     */
+    private static final int DECIMAL_VALUE_SIZE = 32;
```
AMashenkov, line 45:
```suggestion
    /**
     * The value is used to avoid unnecessary
     * buffer reallocation when building a binary row. The exact size of the decimal value can be determined
     * after compaction (see {@link BinaryTupleCommon#shrinkDecimal(BigDecimal, int)}.
     */
    private static final int DECIMAL_VALUE_ESTIMATED_SIZE = 32;
```

xtern: Done, thanks

lowka (approved):

PR #3841, opened 2024-05-29
IGNITE-22371 Increase Netty IO buffer size

rpuch:
https://issues.apache.org/jira/browse/IGNITE-22371

### DIFF
--- a/modules/network/src/main/java/org/apache/ignite/internal/network/netty/OutboundEncoder.java
+++ b/modules/network/src/main/java/org/apache/ignite/internal/network/netty/OutboundEncoder.java
@@ -41,6 +41,8 @@ public class OutboundEncoder extends MessageToMessageEncoder<OutNetworkObject> {
     /** Handler name. */
     public static final String NAME = "outbound-encoder";
 
+    private static final int IO_BUFFER_CAPACITY = 16 * 1024;
+
     private static final NetworkMessagesFactory MSG_FACTORY = new NetworkMessagesFactory();
 
     /** Serialization registry. */
@@ -136,8 +138,8 @@ public class OutboundEncoder extends MessageToMessageEncoder<OutNetworkObject> {
 
         /** {@inheritDoc} */
         @Override
-        public ByteBuf readChunk(ByteBufAllocator allocator) throws Exception {
-            ByteBuf buffer = allocator.ioBuffer();
+        public ByteBuf readChunk(ByteBufAllocator allocator) {
+            ByteBuf buffer = allocator.ioBuffer(IO_BUFFER_CAPACITY);
             int capacity = buffer.capacity();
 
             ByteBuffer byteBuffer = buffer.internalNioBuffer(0, capacity);


Phillippko (approved):

modules/network/src/main/java/org/apache/ignite/internal/network/netty/OutboundEncoder.java
```diff
@@ -136,8 +138,8 @@ public ByteBuf readChunk(ChannelHandlerContext ctx) throws Exception {
 
         /** {@inheritDoc} */
         @Override
-        public ByteBuf readChunk(ByteBufAllocator allocator) throws Exception {
-            ByteBuf buffer = allocator.ioBuffer();
+        public ByteBuf readChunk(ByteBufAllocator allocator) {
+            ByteBuf buffer = allocator.ioBuffer(IO_BUFFER_CAPACITY);
             int capacity = buffer.capacity();
 
             ByteBuffer byteBuffer = buffer.internalNioBuffer(0, capacity);
```
Phillippko, line 145: Interestingly, I see "internal use only" in documentation to this method))

modules/network/src/main/java/org/apache/ignite/internal/network/netty/OutboundEncoder.java
```diff
@@ -41,6 +41,8 @@ public class OutboundEncoder extends MessageToMessageEncoder<OutNetworkObject> {
     /** Handler name. */
     public static final String NAME = "outbound-encoder";
 
+    private static final int IO_BUFFER_CAPACITY = 16 * 1024;
```
Phillippko, line 44: I assume, this number is an optimal size? Or just a guess?

modules/network/src/main/java/org/apache/ignite/internal/network/netty/OutboundEncoder.java
rpuch: The method javadoc follows:

```
    /**
     * Allocate a {@link ByteBuf}, preferably a direct buffer which is suitable for I/O.
     */
```

Where do you see that 'internal use only'?

rpuch: Oh, it's about `internalNioBuffer()`. We don't use the resizability of `ByteBuf`; instead, we use the original NIO buffer, for speed.

modules/network/src/main/java/org/apache/ignite/internal/network/netty/OutboundEncoder.java
rpuch: A ran a few benchmarks. 16kb buffer allows to reach the same network utilization for big messages as 100kb buffer, also in benchmarks using our KV/SQL APIs it behaves on par (giving the same improvement for the scenario of PUTs containing 10k entries each). On the other hand, 16kb is substantially smaller than 100kb.

So I cannot claim that 16kb is an optimal value, but it works 'pretty well'. We can change it in the future if we realize that it's not optimal (this may require a lot of benchmarking).

PR #3846, opened 2024-05-30
IGNITE-20503 Sql. Support big clusters by mapping service

zstan:
SqlBenchmark:
big cluster mapping implementation
```
Benchmark                      Mode  Cnt    Score   Error  Units
SqlBenchmark.selectAllSimple  thrpt   20  205.076 ± 3.504  ops/s
```

small cluster mapping implementation
```
Benchmark                      Mode  Cnt    Score   Error  Units
SqlBenchmark.selectAllSimple  thrpt   20  205.251 ± 9.775  ops/s
```
---
```
Benchmark                     Mode  Cnt        Score       Error  Units
Benchmark                                    Mode  Cnt         Score         Error  Units
MappingBenchmark.benchBig                   thrpt   10   4398392.466 ±  340668.001  ops/s
MappingBenchmark.benchBigPartitionedOnly    thrpt   10  46567579.309 ± 2731802.736  ops/s
MappingBenchmark.benchSmall                 thrpt   10   4829595.861 ±   16053.724  ops/s
MappingBenchmark.benchSmallPartitionedOnly  thrpt   10  17689787.947 ±  445053.549  ops/s
```

### DIFF
--- a/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/mapping/ExecutionTarget.java
+++ b/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/mapping/ExecutionTarget.java
@@ -44,15 +44,4 @@ public interface ExecutionTarget {
      * @throws ColocationMappingException In case these targets can't be colocated.
      */
     ExecutionTarget colocateWith(ExecutionTarget other) throws ColocationMappingException;
-
-    /**
-     * Finalises target by choosing exactly one node for targets with multiple options.
-     *
-     * <p>Some targets may have several options, so we have to pick one in order to get
-     * correct results. Call to this methods resolves this ambiguity by truncating all
-     * but one option. Which exactly option will be left is implementation defined.
-     *
-     * @return Finalised target.
-     */
-    ExecutionTarget finalise();
 }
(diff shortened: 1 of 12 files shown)


zstan:
@korlov42 can you make initial review plz ? probably we can store only one implementation here ?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/mapping/smallcluster/AbstractTarget.java
```diff
@@ -34,7 +34,7 @@
  * <p>Children of this class are primary used for dispatching an execution to a corresponding
  * colocation method.
  */
-abstract class AbstractTarget implements ExecutionTarget {
+public abstract class AbstractTarget implements ExecutionTarget {
```
korlov42, line 37: it's not supposed to be public. If you don't like `finalise()` on ExecutionTarget, then it's better to do finalization inside of Factory whenever it required

korlov42 (approved):
I'm a bit concerned that we don't have tests for a new factory. Does it make sense to at least introduce system property, so we can switch between two factories in test and run FragmentMappingTest in both configuration? WDYT?

PR #3847, opened 2024-05-30
IGNITE-22287 Remove decimal scale from packObjectAsBinaryTuple

ptupitsyn:
[IGNITE-21745](https://issues.apache.org/jira/browse/IGNITE-21745) #3493 adds decimal scale as part of the binary tuple payload. We no longed need to send it separately in `ClientBinaryTupleUtils#appendTypeAndScale`.


ptupitsyn:
Turns out we need the scale to preserve original `BigDecimal` representation in Java. If we remove it, tests fail with errors like:

```
Unexpected string representation of value ==> expected: <10> but was: <1E+1>
in org.apache.ignite.internal.runner.app.client.ItThinClientComputeTest.testAllSupportedArgTypes)
```

and

```
org.opentest4j.AssertionFailedError: expected: <112000> but was: <1.12E+5>
in org.apache.ignite.internal.runner.app.client.ItThinClientComputeTest.testBigDecimalPropagation([3] 1.12E5, 0)
```

PR #3852, opened 2024-05-31
IGNITE-22386 Fix revision serialization where it is required to be properly comparable

ibessonov:
https://issues.apache.org/jira/browse/IGNITE-22386

### The Review Checklist

### Notes


modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
```diff
@@ -64,6 +65,17 @@ public static long bytesToLong(byte[] bytes) {
...
     }
 
+
+    /**
+     * Constructs {@code long} from byte array created with {@link #longToComparableBytes(long)} with offset equal to 0.
+     *
+     * @param bytes Array of bytes.
+     * @return Long value.
+     */
+    public static long comparableBytesToLong(byte[] bytes) {
```
sanpwc, line 74: What's the penalty of such serialisation/deserialisation in comparison to `longToBytes` that forces you to have both? I mean that there's a chance that somebody will miss the fact the there's an extra comparable version of the method and will use common one where he actually need comparison.

modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
```diff
@@ -74,6 +86,17 @@ public static byte[] longToBytes(long l) {
         return putLongToBytes(l, new byte[8], 0);
     }
 
+    /**
```
sanpwc, line 88: What about ints?

modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
```diff
@@ -74,6 +85,17 @@ public static byte[] longToBytes(long l) {
...
+     * Converts a primitive {@code long} value to a byte array than can be compared with {@link ByteBuffer#compareTo(ByteBuffer)} with the
```
rpuch, line 89:
```suggestion
     * Converts a primitive {@code long} value to a byte array that can be compared with {@link ByteBuffer#compareTo(ByteBuffer)} with the
```

modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
```diff
@@ -74,6 +85,17 @@ public static byte[] longToBytes(long l) {
...
+     * same order as the original {@code long}.
+     *
+     * @param l Long value.
+     * @return Array of bytes.
+     */
+    public static byte[] longToComparableBytes(long l) {
```
rpuch, line 95: `Comparable` in the name is a bit confusing, it's not clear, comparable to what and why the alternative is not comparable (as it is comparable, but *differently*).

How about `longToBytesKeepingComparability()`? Or `longToBytesKeepingOrder()`

modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
ibessonov: No penalty at all. If developers won't try deserializing data by hand and will always use these methods, then there would be no difference at all

modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
ibessonov: No one needs it right now

modules/core/src/test/java/org/apache/ignite/internal/util/ByteUtilsTest.java
```diff
@@ -43,4 +48,19 @@ void testStringFromBytes() {
         assertEquals("", stringFromBytes(stringToBytes("")));
         assertEquals("abc", stringFromBytes(stringToBytes("abc")));
     }
+
+    @Test
+    void testComparableLong() {
```
rpuch, line 53: The method is pretty subtle, so we should probably have more (separate) tests. I would suggest to have the following:

1. A test that makes sure that forward-then-backward conversion yields the same result we had in the beginning (this is checked by the first line of the test)
2. For each of 8 bytes, a test that makes sure that, if just Nth byte differs, and in both cases the value of the byte has the same sign, the order is as expected
3. For each of 8 bytes, a test that makes sure that, if just Nth byte differs, and the values of the byte has different signs, the order is as expected

Parameterization could be used for 2 and 3.

modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
sanpwc: So, my question is actually following. Should we include xoring into common longToBytes, intToBytes, etc in order to preserve comparison in lex-order and remove comparable<> counterparts that you've added?

modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
sanpwc: So, someone else will have some fun while debugging similar issue in a year with ints, right? ;)

modules/core/src/test/java/org/apache/ignite/internal/util/ByteUtilsTest.java
ibessonov: Isn't that an overkill? It's really hard to implement this method wrongly

modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
rpuch: What if someone wants to get a BE representation specifically?

modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
ibessonov: Maybe

modules/core/src/test/java/org/apache/ignite/internal/util/ByteUtilsTest.java
ibessonov: I also have check where different bytes differ (1 vs 256), it validates that we use BE

modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
ibessonov: We already have instances where these methods are used exclusively as a conversion to BE. What do we do with them?

sanpwc: Discussed in private. Up to you.

modules/core/src/main/java/org/apache/ignite/internal/util/ByteUtils.java
sanpwc: Well, I strongly suggest to add comparable counterparts to int.

modules/core/src/test/java/org/apache/ignite/internal/util/ByteUtilsTest.java
rpuch: There is a constant with 8 and 0 intermingled. It's easy to mess them up in this pattern, and not so easy to spot...

rpuch (approved):

sanpwc (approved):

PR #3863, opened 2024-06-03
IGNITE-22272 Rework Catalog API

Pochatkin:
* Remove `Options` class 
* Remove `Query` interface from public API
* Remove `toSqlString` method from Query interface 
* Only default sql render options are supported for now 
* All create table methods from `IgniteCatalog` now return `Table` instance

https://issues.apache.org/jira/browse/IGNITE-22272

### DIFF
--- a/modules/api/src/main/java/org/apache/ignite/Ignite.java
+++ b/modules/api/src/main/java/org/apache/ignite/Ignite.java
@@ -85,7 +85,7 @@ public interface Ignite extends AutoCloseable {
     CompletableFuture<Collection<ClusterNode>> clusterNodesAsync();
 
     /**
-     * Returns {@link IgniteCatalog} with default options, which can be used to create and execute SQL DDL queries from annotated classes or
+     * Returns {@link IgniteCatalog} which can be used to create and execute SQL DDL queries from annotated classes or
      * from fluent-style builders.
      *
      * @return Catalog object.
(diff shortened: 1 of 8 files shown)


modules/api/src/main/java/org/apache/ignite/Ignite.java
```diff
@@ -85,22 +84,11 @@ public interface Ignite extends AutoCloseable {
...
-     * Returns {@link IgniteCatalog}, which can be used to create and execute SQL DDL queries from annotated classes or from fluent-style
-     * builders.
-     *
-     * @param options Query options.
-     * @return Catalog object.
-     */
-    IgniteCatalog catalog(Options options);
-
     /**
      * Returns {@link IgniteCatalog} with default options, which can be used to create and execute SQL DDL queries from annotated classes or
```
ptupitsyn, line 88: Let's remove "default options" part since we don't have options here anymore?

modules/api/src/main/java/org/apache/ignite/catalog/Query.java
```diff
@@ -17,19 +17,14 @@
 
 package org.apache.ignite.catalog;
 
+import java.util.concurrent.CompletableFuture;
+
 /**
  * Represents the query generated from the definition or annotation. Can be executed directly or converted to the string representation.
  */
-public interface Query {
+public interface Query<T> {
```
ptupitsyn, line 25: Looks like this interface is not used in any public API, can we remove it?

ptupitsyn (changes requested):

ptupitsyn:
@Pochatkin could you please list all the changes in the PR description?

ptupitsyn (approved):

PR #3865, opened 2024-06-03
IGNITE-22399 Improve handling of small network messages on send

rpuch:
https://issues.apache.org/jira/browse/IGNITE-22399

### DIFF
--- a/modules/network/src/main/java/org/apache/ignite/internal/network/netty/PipelineUtils.java
+++ b/modules/network/src/main/java/org/apache/ignite/internal/network/netty/PipelineUtils.java
@@ -18,6 +18,7 @@
 package org.apache.ignite.internal.network.netty;
 
 import io.netty.channel.ChannelPipeline;
+import io.netty.handler.flush.FlushConsolidationHandler;
 import io.netty.handler.ssl.SslContext;
 import io.netty.handler.stream.ChunkedWriteHandler;
 import java.util.function.Consumer;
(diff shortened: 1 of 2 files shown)


modules/network/src/main/java/org/apache/ignite/internal/network/netty/PipelineUtils.java
```diff
@@ -57,6 +58,9 @@ public static void setup(ChannelPipeline pipeline, PerSessionSerializationServic
      */
     public static void setup(ChannelPipeline pipeline, PerSessionSerializationService serializationService,
             HandshakeManager handshakeManager, Consumer<InNetworkObject> messageListener) {
+        // Consolidate flushes to bigger ones (improves throughput with smaller messages at the price of the latency).
```
tkalkirill, line 61: And if there are big messages and small ones, will they also be collected in batch or will there be a flush immediately if there are big ones?

tkalkirill (approved):

modules/network/src/main/java/org/apache/ignite/internal/network/netty/PipelineUtils.java
rpuch: Big messages are split to fragments up to 16kb each, and those fragments will be collected to batches along with small messages. Batch size is 256 at max, so this will take up to 4Mb per event loop (and number of event loops is cpus*2).

PR #3881, opened 2024-06-05
IGNITE-22202 Primary replica reservation for nodes that are excluded from assignments

denis-chudov:
https://issues.apache.org/jira/browse/IGNITE-22202


modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1028,4 +1087,357 @@ public Set<ReplicationGroupId> startedGroups() {
                 .map(Entry::getKey)
                 .collect(toSet());
     }
+
+    @TestOnly
+    public boolean isReplicaPrimaryOnly(ReplicationGroupId groupId) {
```
sanpwc, line 1092: Do we really need it? Why not to user. PD.getPrimary() check whether replica in assignments or not?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1005,6 +1021,49 @@ public boolean isReplicaStarted(ReplicationGroupId replicaGrpId) {
...
     }
 
+    /**
+     * Can possibly start replica if it's not running or is stopping.
+     *
+     * @param groupId Group id.
+     * @param startOperation Replica start operation.
+     * @param forcedAssignments Assignments to reset forcibly, if needed. Assignments reset is only available when replica is started.
+     */
+    public CompletableFuture<Boolean> weakReplicaStart(
```
sanpwc, line 1031: Didn't check that part with my own eyes. Do we update client peers and learners outside the closure? I mean that we should.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1028,4 +1087,357 @@ public Set<ReplicationGroupId> startedGroups() {
                 .map(Entry::getKey)
                 .collect(toSet());
     }
+
+    @TestOnly
+    public boolean isReplicaPrimaryOnly(ReplicationGroupId groupId) {
+        return replicaLifecycle.isReplicaPrimaryOnly(groupId);
+    }
+
+    private static class ReplicaLifecycle {
```
sanpwc, line 1096: The naming clashes with PartitionReplicaLifecycleManager. Any other ideas for the naming?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1028,4 +1087,357 @@ public Set<ReplicationGroupId> startedGroups() {
...
+
+        final Executor replicaStartStopPool;
+
+        final ClockService clockService;
+
+        final PlacementDriver placementDriver;
+
+        final ReplicaManager replicaManager;
+
+        String localNodeId;
```
sanpwc, line 1109: volatile?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1028,4 +1087,357 @@ public Set<ReplicationGroupId> startedGroups() {
...
+
+        ReplicaLifecycle(Executor replicaStartStopPool, ClockService clockService, PlacementDriver placementDriver,
```
sanpwc, line 1111: One at a row please.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1028,4 +1087,357 @@ public Set<ReplicationGroupId> startedGroups() {
...
+
+        private CompletableFuture<Boolean> onPrimaryElected(PrimaryReplicaEventParameters parameters) {
+            ReplicaLifecycleContext context = getContext(parameters.groupId());
+
+            synchronized (context) {
+                if (localNodeId.equals(parameters.leaseholderId())) {
+                    assert context.replicaState != ReplicaState.STOPPED : "Unexpected primary replica state STOPPED [groupId="
+                            + parameters.groupId() + "].";
+
+                    context.reservedForPrimary = true;
```
sanpwc, line 1133: I didn't get that part. Should it already be reserved because of reservation within waitForActualState?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1028,4 +1087,357 @@ public Set<ReplicationGroupId> startedGroups() {
...
+                } else {
+                    context.reservedForPrimary = false;
```
sanpwc, line 1135: BTW, why reserved for primary is not the state?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
sanpwc, line 1133: Hmm, is it possible to get onPrimaryElected if we are not in assignments? If it is, I'd say that we should update the state to either PRIMARY_ONLY or ASSIGNED.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1028,4 +1087,357 @@ public Set<ReplicationGroupId> startedGroups() {
...
+
+                    if (context.replicaState == ReplicaState.PRIMARY_ONLY) {
+                        stopReplica(parameters.groupId(), context, context.deferredStopOperation);
```
sanpwc, line 1138: Why stopping on PRIMARY_REPLICA_EXIRED is not enough?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1028,4 +1087,357 @@ public Set<ReplicationGroupId> startedGroups() {
...
+                    }
+                }
+            }
+
+            return falseCompletedFuture();
```
sanpwc, line 1143: Why it's boolean? Where the boolean result is used?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1028,4 +1087,357 @@ public Set<ReplicationGroupId> startedGroups() {
...
+        }
+
+        private CompletableFuture<Boolean> onPrimaryExpired(PrimaryReplicaEventParameters parameters) {
+            if (localNodeId.equals(parameters.leaseholderId())) {
```
sanpwc, line 1147: As mentioned above, I'd expect to see status update here.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1028,4 +1087,357 @@ public Set<ReplicationGroupId> startedGroups() {
...
+                } else if (state == ReplicaState.ASSIGNED) {
+                    if (forcedAssignments != null) {
+                        assert forcedAssignments.force() :
+                                format("Unexpected assignments to force [assignments={}, groupId={}].", forcedAssignments, groupId);
+
+                        replicaManager.resetPeers(groupId, fromAssignments(forcedAssignments.nodes()));
+                    }
+
+                    // Telling the caller that the replica is started.
+                    return trueCompletedFuture();
```
sanpwc, line 1196: And what if we are not in force assignments and should the replica?

modules/index/src/main/java/org/apache/ignite/internal/index/IndexBuildController.java
```diff
@@ -306,10 +306,7 @@ private void scheduleBuildIndex(
     ) {
         MvPartitionStorage mvPartition = mvPartitionStorage(mvTableStorage, replicaId);
 
-        // TODO: IGNITE-22202 Deal with this situation
-        if (mvPartition == null) {
-            return;
-        }
+        assert mvPartition != null;
```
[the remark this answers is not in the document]
denis-chudov: added the message

modules/replicator/src/integrationTest/java/org/apache/ignite/internal/replicator/ItPlacementDriverReplicaSideTest.java
```diff
@@ -208,7 +209,8 @@ public void beforeTest(TestInfo testInfo) {
                     null,
                     topologyAwareRaftGroupServiceFactory,
                     raftManager,
-                    new VolatileLogStorageFactoryCreator(nodeName, workDir.resolve("volatile-log-spillout"))
+                    new VolatileLogStorageFactoryCreator(nodeName, workDir.resolve("volatile-log-spillout")),
+                    Executors.newSingleThreadExecutor()
```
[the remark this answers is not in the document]
denis-chudov: replaced with ForkJoinPool.commonPool()

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1005,6 +1021,49 @@ public boolean isReplicaStarted(ReplicationGroupId replicaGrpId) {
         return replicaFuture != null && isCompletedSuccessfully(replicaFuture);
     }
 
+    /**
+     * Can possibly start replica if it's not running or is stopping.
```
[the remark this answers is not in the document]
denis-chudov: done

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1028,4 +1087,357 @@ public Set<ReplicationGroupId> startedGroups() {
...
+                throw new AssertionError("Replica start cannot begin while the replica is being started [groupId=" + groupId + "].");
+            }
+        }
+
+        private CompletableFuture<Boolean> startReplica(
+                ReplicationGroupId groupId,
+                ReplicaLifecycleContext context,
+                Supplier<CompletableFuture<Boolean>> startOperation
+        ) {
+            context.replicaState = ReplicaState.STARTING;
```
sanpwc, line 1212: Why it's not implemented as a state machine? It might be simper to have ReplicaLifecycleStateMachine instead  ReplicaLifecycle with single public method changeState(groupId, newState), internally it will conditionally change the state and call the closure if needed. Some states should await another some will be applied instantly, some are considered as impossible and will through an exception.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/Replica.java
```diff
@@ -94,6 +94,8 @@ public class Replica {
 
     private final ClockService clockService;
 
+    private final ReplicaManager replicaManager;
```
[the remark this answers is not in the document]
denis-chudov: replaced with closure

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1005,6 +1021,49 @@ public boolean isReplicaStarted(ReplicationGroupId replicaGrpId) {
         return replicaFuture != null && isCompletedSuccessfully(replicaFuture);
     }
 
+    /**
+     * Can possibly start replica if it's not running or is stopping.
+     *
+     * @param groupId Group id.
+     * @param startOperation Replica start operation.
```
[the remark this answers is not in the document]
denis-chudov: yes, added this to javadoc

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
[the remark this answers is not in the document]
denis-chudov: it's not related to rework of `ReplicaManager#startReplica` to CompletableFuture<Replica> it's different future.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1005,6 +1021,49 @@ public boolean isReplicaStarted(ReplicationGroupId replicaGrpId) {
...
+     * @param forcedAssignments Assignments to reset forcibly, if needed. Assignments reset is only available when replica is started.
+     */
+    public CompletableFuture<Boolean> weakReplicaStart(
+            ReplicationGroupId groupId,
+            Supplier<CompletableFuture<Boolean>> startOperation,
+            @Nullable Assignments forcedAssignments
```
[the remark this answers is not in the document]
denis-chudov: it can be added in future when it's needed. Now this would be a method that is never used.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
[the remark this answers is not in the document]
denis-chudov: done

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1028,4 +1087,357 @@ public Set<ReplicationGroupId> startedGroups() {
...
+            }
+        }
+
+        private CompletableFuture<Boolean> startReplica(
+                ReplicationGroupId groupId,
+                ReplicaLifecycleContext context,
+                Supplier<CompletableFuture<Boolean>> startOperation
+        ) {
+            context.replicaState = ReplicaState.STARTING;
+            context.previousOperationFuture = context.previousOperationFuture
```
sanpwc, line 1213: I also believe that instead of exclusive states ASSIGNED and PRIMARY_ONLY that should be STARTED, STARTED_RESERVED, STARTED_PRIMARY or we may hide RESERVED/PRIMARY inside started.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1005,6 +1021,49 @@ public boolean isReplicaStarted(ReplicationGroupId replicaGrpId) {
...
+            ReplicationGroupId groupId,
+            Supplier<CompletableFuture<Boolean>> startOperation,
+            @Nullable Assignments forcedAssignments
+    ) {
+        return replicaLifecycle.weakReplicaStart(groupId, startOperation, forcedAssignments);
+    }
+
+    /**
+     * Can possibly stop replica if it is running or starting, and is not a primary replica. Relies on the given reason. If
+     * the reason is {@link WeakReplicaStopReason#EXCLUDED_FROM_ASSIGNMENTS} then the replica can be not stopped if it is still
```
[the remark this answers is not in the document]
denis-chudov: It's correct. "Can be not stopped" means that the replica probably will be not stopped; "can not be stopped" means it won't be stopped regardless of circumstances

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1005,6 +1021,49 @@ public boolean isReplicaStarted(ReplicationGroupId replicaGrpId) {
...
+    public CompletableFuture<Boolean> weakReplicaStart(
+            ReplicationGroupId groupId,
+            Supplier<CompletableFuture<Boolean>> startOperation,
+            @Nullable Assignments forcedAssignments
+    ) {
+        return replicaLifecycle.weakReplicaStart(groupId, startOperation, forcedAssignments);
+    }
+
+    /**
+     * Can possibly stop replica if it is running or starting, and is not a primary replica. Relies on the given reason. If
```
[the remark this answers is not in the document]
denis-chudov: extended the javadoc

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1005,6 +1021,49 @@ public boolean isReplicaStarted(ReplicationGroupId replicaGrpId) {
...
+        return replicaLifecycle.weakReplicaStop(groupId, reason, stopOperation);
+    }
+
+    /**
+     * Reserve replica as primary.
+     *
+     * @param groupId Group id.
+     * @return Whether the replica was successfully reserved.
+     */
+    boolean reserveReplica(ReplicationGroupId groupId) {
```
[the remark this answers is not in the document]
denis-chudov: removed the method

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
denis-chudov: seems it's different, it tells nothing about replica state in terms of lifecycle

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
denis-chudov: fixed

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
denis-chudov: fixed

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
denis-chudov: It should. I replaced it with assertion (hope it will not fail :) )

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
denis-chudov: I've thought about it. It would require the split of ASSIGNED state to something like ASSIGNED and ASSIGNED_PRIMARY, this would complicate the transition between states and adding the state transition logic to #onPrimaryElected, #onPrimaryExpired and #reserveReplica method. Now the replica states reflect it's lifecycle, and `reservedForPrimary` exists in parallel with that, it seems to be more simple and intuitive.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
denis-chudov: It's required by `EventListener` interface, false means the listener shouldn't be removed

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
denis-chudov: I'm not sure I understand the question correctly, but both ASSIGNED and PRIMARY_ONLY states assume that the replica is already started, so we return true here (the same as startOperation was called and returned completable future with `true` result)

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1005,6 +1021,49 @@ public boolean isReplicaStarted(ReplicationGroupId replicaGrpId) {
         return replicaFuture != null && isCompletedSuccessfully(replicaFuture);
     }
 
+    /**
+     * Can possibly start replica if it's not running or is stopping.
+     *
+     * @param groupId Group id.
+     * @param startOperation Replica start operation.
+     * @param forcedAssignments Assignments to reset forcibly, if needed. Assignments reset is only available when replica is started.
```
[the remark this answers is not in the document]
denis-chudov: fixed

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -1005,6 +1021,49 @@ public boolean isReplicaStarted(ReplicationGroupId replicaGrpId) {
...
+
+    /**
+     * Can possibly stop replica if it is running or starting, and is not a primary replica. Relies on the given reason. If
+     * the reason is {@link WeakReplicaStopReason#EXCLUDED_FROM_ASSIGNMENTS} then the replica can be not stopped if it is still
+     * a primary. If the reason is {@link WeakReplicaStopReason#PRIMARY_EXPIRED} then the replica is stopped only if its state
+     * is {@link ReplicaState#PRIMARY_ONLY}, because this assumes that it was excluded from assignments before.
+     *
+     * @param groupId Group id.
+     * @param reason Reason to stop replica.
+     * @param stopOperation Replica stop operation.
```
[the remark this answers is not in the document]
denis-chudov: fixed

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
denis-chudov: actually all the logic is in 2 methods now (weakStartReplica and weakStopReplica), maybe one single method would be too big; also it would have to accept the type of action (start or stop), because the transitions depend on it, so I am not sure the conditional logic will look more simple, just sheet of `if`-s in the number equal to cartesian product size

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
denis-chudov: what about ReplicaStateManager?

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
denis-chudov: because there can be a scenario:
- lease negotiation begins
- replica is reserved
- replica is excluded from assignments and state is transfered to PRIMARY_ONLY
- negotiation fails and another node is elected as leaseholder, here we get a notification through PRIMARY_ELECTED event. Replica is neither in assignments nor primary nor going to become primary and needs to be stopped

sanpwc (commented):
Approved.

PR #3883, opened 2024-06-06
IGNITE-21958: Extend test coverage for SQL E021-11(POSITION function)

lowka:
Updates test cases for POSITION function.

https://issues.apache.org/jira/browse/IGNITE-21958

---

### The Review Checklist

### Notes


modules/sql-engine/src/integrationTest/sql/function/blob/test_position.test
```diff
@@ -46,3 +47,13 @@ null
 
 statement error
 SELECT POSITION('a' IN x'0B03B0B0');
+
+#Update error message
```
zstan, line 51: Only character position expression expect USING usage.
POSITION <left paren> <binary value expression> IN <binary value expression> <right paren>

lowka: Given 

> SELECT POSITION(x IN somestring USING CHARACTERS) 

Parser alone is incapable of deciding whether this position is valid or not, because validation of this expression requires type information. Only the validator can reject or accept this call, so the error should indicate that there is no POSITION (<binary> IN <BINARY> USING CHARACERS)

PR #3908, opened 2024-06-12
IGNITE-22224 Rework embedded mode API

valepakh:
https://issues.apache.org/jira/browse/IGNITE-22224

This PR changes the embedded API so that starting a node, initializing a cluster and getting the Ignite API facade are separate. Also it moves the embedded API interfaces from `ignite-api` module to the `ignite-runner` module.

The basic start of the embedded node, cluster initialization and working with the API could look like this:
```java
IgniteServer node = IgniteServer.start("node", configFilePath, workDir);
InitParameters initParameters = InitParameters.builder()
        .metaStorageNodeNames("node")
        .clusterName("cluster")
        .build();
node.initClusterAsync(initParameters).thenRun(() -> node.api().tables()...);
```
If cluster initialization is done externally (via REST or from another node), we need to wait for it to complete before using the API:
```java
IgniteServer node = IgniteServer.start("node", configFilePath, workDir);
node.waitForinitAsync(initParameters).thenRun(() -> node.api().tables()...);
```

### DIFF
--- a/modules/runner/src/main/java/org/apache/ignite/IgniteServer.java
+++ b/modules/runner/src/main/java/org/apache/ignite/IgniteServer.java
@@ -25,6 +25,7 @@ import org.jetbrains.annotations.Nullable;
 
 /**
  * Embedded Ignite node. Manages node's lifecycle, provides Ignite API and allows cluster initialization.
+ *
  * <p>NOTE: Methods of this interface are not thread-safe and shouldn't be called from different threads.
  */
 public interface IgniteServer {


modules/runner/src/main/java/org/apache/ignite/EmbeddedNode.java
```diff
@@ -52,40 +57,88 @@ public interface Ignition {
...
+        embeddedNode.start(); // Ignite future is kept inside this instance and will be returned from the igniteAsync.
+        return embeddedNode;
+    }
+
+    /**
+     * Returns the Ignite API. The API is available when the node has joined an initialized cluster.
+     *
+     * @return Ignite API future.
+     */
+    CompletableFuture<Ignite> igniteAsync();
```
ptupitsyn, line 77: This method returns a future that will never complete if I don't call `initClusterAsync`, right?

modules/runner/src/main/java/org/apache/ignite/EmbeddedNode.java
```diff
@@ -52,40 +57,88 @@ public interface Ignition {
...
-    CompletableFuture<Ignite> start(
+    static EmbeddedNode create(
             String nodeName,
             Path configPath,
             Path workDir,
             @Nullable ClassLoader serviceLoaderClassLoader
-    );
+    ) {
+        EmbeddedNode embeddedNode = new EmbeddedNodeImpl(nodeName, configPath, workDir, serviceLoaderClassLoader);
+        embeddedNode.start(); // Ignite future is kept inside this instance and will be returned from the igniteAsync.
```
ptupitsyn, line 68: I don't think we should start the node as part of `create` call, the API becomes confusing.

modules/runner/src/main/java/org/apache/ignite/EmbeddedNode.java
ptupitsyn: I believe this method should be synchronous and return a result immediately in any case. If the node is not in a correct state, then Ignite API methods will throw an exception.

modules/cli/src/integrationTest/java/org/apache/ignite/internal/rest/ItGeneratedRestClientTest.java
```diff
@@ -103,7 +102,7 @@ public class ItGeneratedRestClientTest extends BaseIgniteAbstractTest {
     @WorkDirectory
     private static Path WORK_DIR;
 
-    private final List<String> clusterNodeNames = new ArrayList<>();
+    private List<IgniteServer> nodes;
```
PakhomovAlexander, line 105: Why do we introduce new term "Server"? Can't we use "Node" instead?

ptupitsyn:
@valepakh  could you please outline the API changes in the PR description?

ptupitsyn (commented):
The new `IgniteServer` API looks good to me.

modules/runner/src/main/java/org/apache/ignite/IgniteServer.java
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * Initializes the cluster that the given node is present in.
+     *
+     * <p>Cluster initialization propagates information about those nodes that will host the Meta Storage and CMG Raft groups
+     * to all nodes in the cluster. After the operation succeeds, nodes can finish the start procedure and begin accepting incoming
+     * requests.
+     *
+     * <p>Meta Storage is responsible for storing cluster-wide meta information required for internal purposes and proper functioning of
+     * the cluster.
```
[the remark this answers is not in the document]
ptupitsyn: Ok, let's keep it for now.

ptupitsyn (approved):
Public API changes look good.

PR #3914, opened 2024-06-13
IGNITE-21726 Enable all metrics by default.

xtern:
https://issues.apache.org/jira/browse/IGNITE-21726


modules/cli/src/integrationTest/java/org/apache/ignite/internal/rest/ItGeneratedRestClientTest.java
```diff
@@ -53,8 +53,10 @@
 import org.apache.ignite.Ignite;
 import org.apache.ignite.IgnitionManager;
 import org.apache.ignite.InitParameters;
+import org.apache.ignite.internal.cli.CliIntegrationTest;
 import org.apache.ignite.internal.cli.call.cluster.unit.DeployUnitClient;
 import org.apache.ignite.internal.cli.core.rest.ApiClientFactory;
+import org.apache.ignite.internal.testframework.BaseIgniteAbstractTest;
```
xtern, line 59: required because of the import 'CliIntegrationTest.ALL_METRIC_SOURCES` and `Depends on Mockito` check

modules/cli/src/integrationTest/java/org/apache/ignite/internal/rest/ItGeneratedRestClientTest.java
```diff
@@ -90,7 +91,7 @@
 @ExtendWith(WorkDirectoryExtension.class)
 @MicronautTest(rebuildContext = true)
 @TestInstance(Lifecycle.PER_CLASS)
-public class ItGeneratedRestClientTest {
+public class ItGeneratedRestClientTest extends BaseIgniteAbstractTest {
```
xtern, line 94: change required because of the import CliIntegrationTest.ALL_METRIC_SOURCES and Depends on Mockito check

modules/cli/src/integrationTest/java/org/apache/ignite/internal/cli/call/metric/ItEnabledMetricCallsTest.java
```diff
@@ -69,15 +75,8 @@ void nodeMetricSourcesList() {
...
-        MetricSource[] expectedMetricSources = {
-                new MetricSource().name("jvm").enabled(true),
-                new MetricSource().name("client.handler").enabled(false),
-                new MetricSource().name("sql.client").enabled(false),
-                new MetricSource().name("sql.plan.cache").enabled(false)
-        };
-
         // And
-        assertThat(output.body()).contains(expectedMetricSources);
+        assertThat(output.body()).contains(new MetricSource().name("jvm").enabled(true));
```
ygerzhedovich, line 79: didn't catch the change. We should check that we have the expected list of metrics, but checks only 'jvm' one

xtern: Test method description states "Should display enabled jvm node metric source when cluster is up and running" :thinking:

xtern: I think I got your point, I can leave these code unchanged :blush: 

Reverted this change, thanks

ygerzhedovich (approved):

PR #3922, opened 2024-06-14
IGNITE-22480 Sql. Avoid compiling literal-only expressions when assembling a row for insert

xtern:
https://issues.apache.org/jira/browse/IGNITE-22480


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
```diff
@@ -343,6 +345,17 @@ public Iterable<RowT> values(List<RexLiteral> values, RelDataType rowType) {
         return rows;
     }
 
+    private @Nullable Object literalValue(RexLiteral literal, Class<?> type) {
+        Object val = literal.getValueAs(type);
+
+        // Literal was parsed as UTC timestamp, now we need to adjust it to the client's time zone.
+        if (val != null && literal.getTypeName() == SqlTypeName.TIMESTAMP_WITH_LOCAL_TIME_ZONE) {
```
lowka, line 352: Should we adjust `TIME_WITH_LOCAL_TIME_ZONE` as well?

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImplTest.java
```diff
@@ -704,8 +701,36 @@ public void testRowSource() {
         RexNode val10 = rexBuilder.makeExactLiteral(new BigDecimal("1"), intType);
         RexNode val11 = rexBuilder.makeExactLiteral(new BigDecimal("2"), bigIntType);
```
lowka, line 702: 1. I think we need to ensure that DECIMAL values preserve precision/scale after this transformation.
2. Ideally, we need to check literals of all types here.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
xtern: we have separate ticket for this https://issues.apache.org/jira/browse/IGNITE-21555

korlov42 (approved):

modules/sql-engine/src/test/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImplTest.java
xtern: Thanks, the test has been improved, please take a look again.

lowka (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/exp/ExpressionFactoryImpl.java
```diff
@@ -275,8 +275,18 @@ public Function<RowT, RowT> project(List<RexNode> projects, RelDataType rowType)
     public Supplier<RowT> rowSource(List<RexNode> values) {
         List<RelDataType> typeList = Commons.transform(values, v -> v != null ? v.getType() : NULL_TYPE);
         RowSchema rowSchema = TypeUtils.rowSchemaFromRelTypes(typeList);
+        List<RexLiteral> literalValues = new ArrayList<>(values.size());
 
-        return new ValuesImpl(scalar(values, null), ctx.rowHandler().factory(rowSchema));
+        // Avoiding compilation when all expressions are constants.
+        for (int i = 0; i < values.size(); i++) {
+            if (!(values.get(i) instanceof RexLiteral)) {
+                return new ValuesImpl(scalar(values, null), ctx.rowHandler().factory(rowSchema));
```
AMashenkov, line 283: Factory  could be cached in local variable.

PR #3925, opened 2024-06-14
IGNITE-19383 Add git last commit hash to version

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-19383

### DIFF
--- a/modules/runner/src/main/java/org/apache/ignite/internal/app/IgnitionImpl.java
+++ b/modules/runner/src/main/java/org/apache/ignite/internal/app/IgnitionImpl.java
@@ -239,18 +239,18 @@ public class IgnitionImpl implements Ignition {
 
         String padding = " ".repeat(22);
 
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -73,7 +73,13 @@ ext {
...
 allprojects {
     group 'org.apache.ignite'
-    version = "3.0.0-SNAPSHOT"
+    def gitLastCommitHash = new ByteArrayOutputStream()
+    exec {
+        commandLine "git", "rev-parse", "--short", "HEAD"
+        workingDir = projectDir
+        standardOutput = gitLastCommitHash
+    }
+    version = "3.0.0-SNAPSHOT-$gitLastCommitHash"
```
iartiukhov, line 82: I think we should not change the product version. Rather store the commit hash into a separate place (like file) and let the node read it from this place.

PR #3926, opened 2024-06-14
IGNITE-22478 Use single argument in Compute API

PakhomovAlexander:
The Compute API is changed:

* Only single argument can be accepted or `null` by any compute job.
* Marshaler API is introduced as a part of public API and can be used in Compute API.
* Default marshaller is implemented but not used explicitly. By default we support only 
  primitives. If the one want to use custom marshaler they should declare it explicitly.

Marshalers allow us to define the default way to serialize objects: `java serialization`.
At the same time client can implement theirs own marshalers and use them in any
public API that supports it. This is the first implementation and it is supposed to be
improved in the future. The only thing that will not be changed -- Marshaler and 
Compute interfaces (their public parts).


modules/api/src/main/java/org/apache/ignite/compute/IgniteCompute.java
```diff
@@ -41,80 +42,85 @@ public interface IgniteCompute {
...
      * @param descriptor Job descriptor.
      * @param args Arguments of the job.
      * @return Job execution object.
      */
-    <R> JobExecution<R> submit(
+    <T, R> JobExecution<R> submit(
             JobTarget target,
             JobDescriptor descriptor,
-            Object... args
+            T args
```
ptupitsyn, line 55:
```suggestion
            T arg
```

(here and below)

modules/api/src/main/java/org/apache/ignite/compute/JobDescriptor.java
```diff
@@ -19,21 +19,34 @@
...
-public class JobDescriptor {
+public class JobDescriptor<T, R> {
     private final String jobClassName;
 
     private final List<DeploymentUnit> units;
 
     private final JobExecutionOptions options;
 
-    private JobDescriptor(String jobClassName, List<DeploymentUnit> units, JobExecutionOptions options) {
+    private final Marshaler<T, byte[]> resultMarshaller;
```
ptupitsyn, line 35:
```suggestion
    private final Marshaler<R, byte[]> resultMarshaller;
```

modules/api/src/main/java/org/apache/ignite/compute/JobDescriptor.java
```diff
@@ -19,21 +19,34 @@
...
+
+    private final Marshaler<T, byte[]> argumentMarshaler;
+
+    private JobDescriptor(
+            String jobClassName,
+            List<DeploymentUnit> units,
+            JobExecutionOptions options,
+            Marshaler<T, byte[]> resultMarshaller,
```
ptupitsyn, line 43:
```suggestion
            Marshaler<R, byte[]> resultMarshaller,
```

modules/api/src/main/java/org/apache/ignite/marshaling/ByteArrayMarshaler.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+public interface ByteArrayMarshaler<T> extends Marshaler<T, byte[]> {
+    @Override
+    default byte @Nullable [] marshal(@Nullable T object) {
+        if (object == null) {
+            return null;
+        }
+
+        if (object instanceof Serializable) {
+            return JavaSerializationByteArrayMarshallilng.marshal((Serializable) object);
```
ptupitsyn, line 37: Looks like this marshaller is the default, and it does not work with simple types (e.g. String), so it can't be used from non-Java clients. Is that correct?

modules/api/src/main/java/org/apache/ignite/compute/IgniteCompute.java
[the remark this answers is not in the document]
PakhomovAlexander: Good point.  Probably it's better to overload methods but I think we can do it later.

modules/api/src/main/java/org/apache/ignite/marshaling/ByteArrayMarshaler.java
```diff
@@ -0,0 +1,51 @@
...
+import java.io.Serializable;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * The ByteArrayMarshaler interface is designed to marshal and unmarshal objects to and from byte arrays. If the byte[] is your preferred
+ * way of marshaling objects, you can implement this interface to provide custom marshaling logic.
+ *
+ * <p>The default implementation of the marshal method is plain java serialization.
+ */
+public interface ByteArrayMarshaler<T> extends Marshaler<T, byte[]> {
```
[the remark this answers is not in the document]
PakhomovAlexander: I've added a factory method. Is it ok?

modules/api/src/main/java/org/apache/ignite/compute/JobDescriptor.java
```diff
@@ -79,19 +92,29 @@ public static Builder builder(String jobClassName) {
      *
      * @return Job descriptor builder.
      */
-    public static Builder builder(Class<? extends ComputeJob<?>> jobClass) {
+    public static Builder builder(Class<? extends ComputeJob<?, ?>> jobClass) {
```
ptupitsyn, line 95: Builder class should have the same generic type arguments as the parent class to avoid wildcards.

modules/client/src/main/java/org/apache/ignite/internal/client/compute/ClientCompute.java
```diff
@@ -90,108 +92,128 @@ public ClientCompute(ReliableChannel ch, ClientTables tables) {
     }
 
     @Override
-    public <R> JobExecution<R> submit(JobTarget target, JobDescriptor descriptor, Object... args) {
+    public <T, R> JobExecution<R> submit(JobTarget target, JobDescriptor<T, R> descriptor, T args) {
```
isapego, line 95: Here and in other places
```suggestion
    public <T, R> JobExecution<R> submit(JobTarget target, JobDescriptor<T, R> descriptor, T arg) {
```

PR #3937, opened 2024-06-17
IGNITE-22427 Fix JobExecution.cancelAsync semantics

valepakh:
https://issues.apache.org/jira/browse/IGNITE-22427

### DIFF
--- a/modules/api/src/main/java/org/apache/ignite/compute/JobExecution.java
+++ b/modules/api/src/main/java/org/apache/ignite/compute/JobExecution.java
@@ -56,8 +56,8 @@ public interface JobExecution<R> {
      * Cancels the job.
      *
      * @return The future which will be completed with {@code true} when the job is cancelled, {@code false} when the job couldn't be
-     *         cancelled (if it's not queued or executing), or {@code null} if the job no longer exists due to exceeding the retention time
-     *         limit.
+     *         cancelled (if it's already completed or in the process of cancelling), or {@code null} if the job no longer exists due to
+     *         exceeding the retention time limit.
      */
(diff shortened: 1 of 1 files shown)


modules/api/src/main/java/org/apache/ignite/compute/JobExecution.java
```diff
@@ -56,8 +56,8 @@ public interface JobExecution<R> {
      * Cancels the job.
      *
      * @return The future which will be completed with {@code true} when the job is cancelled, {@code false} when the job couldn't be
-     *         cancelled (either it's not yet started, or it's already completed), or {@code null} if the job no longer exists due to
-     *         exceeding the retention time limit.
+     *         cancelled (if it's not queued or executing), or {@code null} if the job no longer exists due to exceeding the retention time
```
ptupitsyn, line 59:
```suggestion
     *         cancelled (it is already completed), or {@code null} if the job no longer exists due to exceeding the retention time
```

ptupitsyn (approved):

modules/api/src/main/java/org/apache/ignite/compute/JobExecution.java
ptupitsyn: Let's say `it is already completed or cancelled` then?

ptupitsyn: Or `if it's not queued or executing, meaning already completed or cancelled` to be very explicit.

ptupitsyn: The last one looks good to me.

PR #3950, opened 2024-06-19
IGNITE-22435 Add JobTarget interface

ptupitsyn:
Reduce the number of overloads in `IgniteCompute` and make the API more readable with new `JobTarget` interface:

```java
public interface JobTarget {
    static JobTarget node(ClusterNode node);
    static JobTarget anyNode(Collection<ClusterNode> nodes);
    static JobTarget colocated(String tableName, Tuple key);
    static <K> JobTarget colocated(String tableName, K key, Mapper<K> keyMapper);
}
```

**Usage Example**

```java
String result = entryNode.compute().execute(
        JobTarget.anyNode(node1, node2),
        JobDescriptor.builder(MyJob.class).units(units()).build(),
        "arg");

JobExecution<String> execution = entryNode.compute().submit(
        JobTarget.colocated("My-table", Tuple.create(Map.of("k", 1))),
        JobDescriptor.builder(MyColocatedJob.class).build());
```

* Broadcast methods are not affected because of different return type


modules/api/src/main/java/org/apache/ignite/compute/JobTarget.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Creates a job target for any node from the provided collection.
+     *
+     * @param nodes Collection of nodes.
+     * @return Job target.
+     */
+    static JobTarget anyNode(Collection<ClusterNode> nodes) {
+        return new NodesJobTarget(new HashSet<>(nodes));
+    }
```
isapego, line 59: Should we mention that the collection of nodes is treated as a Set (duplicates are ignored)?

isapego (approved):

modules/client/src/main/java/org/apache/ignite/internal/client/compute/ClientCompute.java
```diff
@@ -87,57 +90,53 @@ public ClientCompute(ReliableChannel ch, ClientTables tables) {
...
-    ) {
-        Objects.requireNonNull(tableName);
-        Objects.requireNonNull(key);
-        Objects.requireNonNull(descriptor);
+        if (target instanceof ColocatedExecutionTarget) {
+            ColocatedExecutionTarget colocatedTarget = (ColocatedExecutionTarget) target;
+            var mapper = (Mapper<? super Object>) colocatedTarget.keyMapper();
+
+            if (mapper != null) {
+                return new ClientJobExecution<>(ch, doExecuteColocatedAsync(
```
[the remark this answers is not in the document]
ptupitsyn: Agree, fixed.

modules/api/src/main/java/org/apache/ignite/compute/JobTarget.java
ptupitsyn: I don't think this matters with "any node" semantics because we don't explain how a node is picked from the provided collection. There is no guarantee of fairness or randomness on the API level.

modules/api/src/main/java/org/apache/ignite/compute/NodesJobTarget.java
```diff
@@ -0,0 +1,43 @@
...
+package org.apache.ignite.compute;
+
+import java.util.Objects;
+import java.util.Set;
+import org.apache.ignite.network.ClusterNode;
+
+/**
+ * Nodes-based job execution target.
+ */
+public class NodesJobTarget implements JobTarget {
```
korlov42, line 27: the name seems to vague... What if we decide to introduce new target with semantic "all nodes", how we will name the container? Does it make sense to rename this to `AnyNodeTarget` to better reflect semantic? This it part of public API, thus we should choose naming carefully

modules/api/src/main/java/org/apache/ignite/compute/JobTarget.java
```diff
@@ -0,0 +1,92 @@
...
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import org.apache.ignite.network.ClusterNode;
+import org.apache.ignite.table.Tuple;
+import org.apache.ignite.table.mapper.Mapper;
+
+/**
+ * Job execution target.
```
korlov42, line 28:
```suggestion
 * Job execution target.
 *
 * <p>Determines the rules for selecting nodes to execute a job.
```

modules/api/src/main/java/org/apache/ignite/compute/JobTarget.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+public interface JobTarget {
+    /**
+     * Creates a job target for a specific node.
+     *
+     * @param node Node.
```
korlov42, line 34:
```suggestion
     * Creates a job target for a specific node.
     *
     * <p>This target determines that a job should be executed on a given node.
     *
     * @param node Node.
```

modules/api/src/main/java/org/apache/ignite/compute/JobTarget.java
```diff
@@ -0,0 +1,92 @@
...
+     * @return Job target.
+     */
+    static JobTarget node(ClusterNode node) {
+        return new NodesJobTarget(Set.of(node));
+    }
+
+    /**
+     * Creates a job target for any node from the provided collection.
+     *
+     * @param nodes Collection of nodes.
```
korlov42, line 44:
```suggestion
     * Creates a job target for any node from the provided collection.
     *
     * <p>This target determines that a job can be executed on any node in a given collection, but only one of them.
     * Which node is chosen is implementation defined.
     *
     * @param nodes Collection of nodes.
```

modules/api/src/main/java/org/apache/ignite/compute/JobTarget.java
```diff
@@ -0,0 +1,92 @@
...
+     * @return Job target.
+     */
+    static JobTarget anyNode(ClusterNode... nodes) {
+        return new NodesJobTarget(Set.of(nodes));
+    }
+
+    /**
+     * Creates a job target for any node from the provided collection.
+     *
+     * @param nodes Collection of nodes.
```
korlov42, line 54:
```suggestion
     * Creates a job target for any node from the provided collection.
     *
     * <p>This target determines that a job can be executed on any node in a given collection, but only one of them.
     * Which node is chosen is implementation defined.
     *
     * @param nodes Collection of nodes.
```

modules/api/src/main/java/org/apache/ignite/compute/JobTarget.java
```diff
@@ -0,0 +1,92 @@
...
+     * @return Job target.
+     */
+    static JobTarget anyNode(Collection<ClusterNode> nodes) {
+        return new NodesJobTarget(new HashSet<>(nodes));
+    }
+
+    /**
+     * Creates a job target for any node from the provided collection.
+     *
+     * @param nodes Collection of nodes.
```
korlov42, line 64:
```suggestion
     * Creates a job target for any node from the provided collection.
     *
     * <p>This target determines that a job can be executed on any node in a given collection, but only one of them.
     * Which node is chosen is implementation defined.
     *
     * @param nodes Collection of nodes.
```

modules/api/src/main/java/org/apache/ignite/compute/JobTarget.java
```diff
@@ -0,0 +1,92 @@
...
+     * @return Job target.
+     */
+    static JobTarget anyNode(Set<ClusterNode> nodes) {
+        return new NodesJobTarget(nodes);
+    }
+
+    /**
+     * Creates a colocated job target for a specific table and key.
+     *
+     * @param tableName Table name.
```
korlov42, line 74:
```suggestion
     * Creates a colocated job target for a specific table and key.
     *
     * <p>This target determines that a job should be executed on the same node that hosts the data for a given key of provided table.
     *
     * @param tableName Table name.
```

modules/api/src/main/java/org/apache/ignite/compute/JobTarget.java
```diff
@@ -0,0 +1,92 @@
...
+     * @return Job target.
+     */
+    static JobTarget colocated(String tableName, Tuple key) {
+        return new ColocatedExecutionTarget(tableName, key, null);
+    }
+
+    /**
+     * Creates a colocated job target for a specific table and key with mapper.
+     *
+     * @param tableName Table name.
```
korlov42, line 85:
```suggestion
     * Creates a colocated job target for a specific table and key with mapper.
     *
     * <p>This target determines that a job should be executed on the same node that hosts the data for a given key of provided table.
     *
     * @param tableName Table name.
```

korlov42 (approved):
the changes in API module looks good in general. I left a few suggestions on how to improve javadoc though

modules/api/src/main/java/org/apache/ignite/compute/NodesJobTarget.java
ptupitsyn: Good point, renamed to `AnyNodeJobTarget`. Also renamed the other one to `ColocatedJobTarget`.

PR #3953, opened 2024-06-19
IGNITE-22073: Sql. Add processing of the provided query timeout

lowka:
Adds query timeouts implemented via `QueryCancel `callbacks.

- To distinguish between ordinary cancellations and timeouts, `QueryCancel` now has internal cancellation reason to distinguish between timeouts and other cancellations.

- To support timeouts `QueryCancel` now includes a future that completes when a query time outs.

- `QueryCancel::add` acts as a guard for an operation, because it throws `QueryCancelledException`, when a query has already been cancelled. E.g. If parsing took too long, then `QueryCancel::add` in `ExecutionService ` throws a `QueryCancelledException` exception, and the operation terminates. 

- Update Cache interface into include a method that invalidates a cache entry by key + a method to check cache size (used it tests).

# SqlQueryProcessor

Timeouts are set by `SqlQueryProcessor`. 

If a timeout is present `SqlQueryProcessor` then:
-  schedules a timeout (submits a task to the common scheduling pool that calls `QueryCancell::timeout`)  that marks the operation as canceled, triggering the necessary callbacks. This is done via `QueryCancel::addTimeout`. When a timeout occurs, a `QueryCancel::timeout` triggers cancellation callbacks.

`ExecutionService` uses `SqlOperationContext::timeoutFuture` to add callbacks that cancel query execution, if timeout occurs.

# Prepare phase

**If prepare step time outs, a cache entry for that plan is removed from the planner cache.**

# Execution

- For standard plans: Timeout simply calls `DistributedQueryManager::close`.
- For K/V plans: Timeout updates a future returned by an operation (terminating it with `QueryCancelledException`).
- DDL plans: Timeout updates a future returned by an operation (terminating it with `QueryCancelledException`).

# Scripts/Multistatements

Rely on the same machinery as single statements but use the same deadline for all statements.

# JDBC

Updates JDBC statements to support timeouts.

---

### The Review Checklist

### Notes


modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -596,14 +624,22 @@ private CompletableFuture<AsyncSqlCursor<InternalSqlRow>> queryScript(
     ) {
         String schemaName = properties.get(QueryProperty.DEFAULT_SCHEMA);
         ZoneId timeZoneId = properties.get(QueryProperty.TIME_ZONE_ID);
+        Long queryTimeout = properties.getOrDefault(QueryProperty.QUERY_TIMEOUT, 0L);
```
ygerzhedovich, line 627: the same about default and Integer.MAX_VALUE below

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -596,14 +624,22 @@ private CompletableFuture<AsyncSqlCursor<InternalSqlRow>> queryScript(
...
+
+        Instant deadline;
+        if (queryTimeout != 0) {
+            deadline = Instant.now().plusMillis(queryTimeout);
```
ygerzhedovich, line 631: what reason to use Instant instead of simple long and System.currentTimeMillis (in fact it should be FastTimestamps.coarseCurrentTimeMillis() )
The comment is common for the approach in the patch

ygerzhedovich (changes requested):

modules/sql-engine/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlSynchronousApiTest.java
```diff
@@ -131,4 +139,99 @@ public void process(ResultSet<SqlRow> resultSet) {
...
+        // Read data until exception occurs.
+        assertThrowsSqlException(Sql.EXECUTION_CANCELLED_ERR, "Query timeout", () -> {
+            while (resultSet.hasNext()) {
+                resultSet.next();
+            }
+        });
+    }
+
+    @Test
+    public void testQueryTimeoutIsPropagatedFromTheServer() throws Exception {
```
xtern, line 187: I see that this test sometimes produces NPE during cancellation
```
[2024-06-24T10:38:03,199][WARN ][%issat_n_0%sql-planning-pool-1][task] Volcano planning times out, cancels the subsequent optimization.
[2024-06-24T10:38:03,446][ERROR][%issat_n_0%sql-execution-pool-1][FailureProcessor] Critical system error detected. Will be handled accordingly to configured handler [hnd=NoOpFailureHandler [], failureCtx=FailureContext [type=CRITICAL_ERROR, err=org.apache.ignite.lang.IgniteException: IGN-CMN-65535 TraceId:05602438-1bba-4f45-94e9-e63ab8c6d21f Unexpected error during execute fragment 145 of query 6fe450b1-b2fb-45a8-9774-c1af41e7e82c]]
 org.apache.ignite.lang.IgniteException: Unexpected error during execute fragment 145 of query 6fe450b1-b2fb-45a8-9774-c1af41e7e82c
	at org.apache.ignite.internal.sql.engine.exec.QueryTaskExecutorImpl.lambda$execute$0(QueryTaskExecutorImpl.java:96) ~[main/:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
	at java.base/java.lang.Thread.run(Thread.java:834) [?:?]
Caused by: java.lang.NullPointerException
	at org.apache.ignite.internal.sql.engine.exec.ExecutionServiceImpl$DistributedQueryManager.acknowledgeFragment(ExecutionServiceImpl.java:839) ~[main/:?]
	at org.apache.ignite.internal.sql.engine.exec.ExecutionServiceImpl.onMessage(ExecutionServiceImpl.java:523) ~[main/:?]
	at org.apache.ignite.internal.sql.engine.exec.ExecutionServiceImpl.lambda$start$2(ExecutionServiceImpl.java:277) ~[main/:?]
	at org.apache.ignite.internal.sql.engine.message.MessageServiceImpl.onMessageInternal(MessageServiceImpl.java:151) ~[main/:?]
	at org.apache.ignite.internal.sql.engine.message.MessageServiceImpl.lambda$onMessage$1(MessageServiceImpl.java:117) ~[main/:?]
	at org.apache.ignite.internal.sql.engine.exec.QueryTaskExecutorImpl.lambda$execute$0(QueryTaskExecutorImpl.java:83) ~[main/:?]
	... 3 more
```
The test doesn't fail but, as I understand, there is a NPE-detector on TeamCity which may not like this.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryCancel.java
```diff
@@ -32,21 +36,66 @@ public class QueryCancel {
...
+     */
+    public CompletableFuture<Void> addTimeout(ScheduledExecutorService scheduler, long timeoutMillis) {
+        CompletableFuture<Void> fut = new CompletableFuture<>();
+        fut.thenAccept((r) -> timeout());
+
+        ScheduledFuture<?> f = scheduler.schedule(() -> {
+            fut.complete(null);
+        }, timeoutMillis, MILLISECONDS);
+
+        add((timeout) -> f.cancel(false));
```
xtern, line 94: Do I understand correctly that this code is needed to cancel a future when the request has completed without a timeout?

Then it might be worth adding 
```
if (!timeout) f.cancel(false)
```
because the logic for canceling a future may not be so trivial :thinking:

lowka: > if (!timeout) f.cancel(false)
This is not correct. When a statement is cancelled its timeout should be cancelled as well, otherwise a ScheduledFuture is still alive but a query does not exist anymore.

xtern: > When a statement is cancelled its timeout should be cancelled as well,

Yes, but when timeout occurred then this future `f` must be already completed and we don't need to cancel it again?
```
add((timeout) -> {

assert timeout == false || f.isDone(); // <-- is this correct?

f.cancel(false);

}
```

lowka: > assert timeout == false || f.isDone(); // <-- is this correct?

These assertions are not necessary, since Future::cancel is idempotent.

`QueryCancel` expects its callbacks to be idempotent as well. It is a in multithreaded environment after all, if multiple actors want to terminate a query, only one of them can succeed (other actors should just accept the fact that someone else has already done the job they tried to do).

modules/sql-engine/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlSynchronousApiTest.java
lowka: IntelliJ's static analyser does not like that line of code either. It expects an NPE here.
I guess it is time to fix it by adding some guard condition.

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcStatementSelfTest.java
```diff
@@ -837,4 +839,49 @@ public void testOpenCursorsWithDdl() throws Exception {
...
+        }
+
+        // Rise timeout
+
+        {
+            int timeoutMillis = ThreadLocalRandom.current().nextInt(10, 200);
+            igniteStmt.timeout(timeoutMillis);
+
+            assertThrowsSqlException(SQLException.class,
+                    "Query timeout", () -> {
```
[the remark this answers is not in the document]
lowka: Yes. We need to decide want to do with a planner timeout error (do we even need it?).

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
xtern: Personally, I would prefer to keep Instant. :thinking: 
From my point of view, the code is cleaner this way. But let's discuss this privately :sunglasses:

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryCancel.java
lowka: Just to make this clear. For every statement we create a QueryCancel thing that stores actions to be called when a statement is cancelled.  

A statement can is cancelled when on initiator in case of an error. DistributedQueryManage calls query cancel : cancel method. QueryCancel cancel calls all cancellation callbacks.

This commit introduced a timeout method on QueryCancel to distinguish between a cancellation due to some exception and a time out.

- QueryProcessor creates an instance of QueryTimeout q.
- it schedules a timeout action for a query that calls q::timeout (to indicate that timeout occurred and a statement should be cancelled)
- when a statement is cancelled due to some exception by its initiator a scheduled future should be cancelled as well.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImpl.java
```diff
@@ -226,26 +232,58 @@ public CompletableFuture<QueryPlan> prepareAsync(
...
+        // Or trigger timeout immediately if operation has already timed out. 
+        QueryCancel cancelHandler = operationContext.cancel();
+        assert cancelHandler != null;
+
+        Cancellable callback = (timeout) -> {
+            if (timeout) {
+                planningContext.abortByTimeout();
+            }
+        };
+        cancelHandler.add(callback);
```
xtern, line 263: Honestly I don't quite understand why we need this.
Do we have a test that will fail without this change?

lowka: Please elaborate. I think I explained it in the PR description.

> QueryCancel::add acts as a guard for an operation, because it throws QueryCancelledException, when a query has already been cancelled. E.g. If parsing took too long, then QueryCancel::add in PrepareService throws a QueryCancelledException exception, and the operation terminates. The same happens in ExecutionService.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/JdbcQueryEventHandlerImpl.java
```diff
@@ -156,7 +162,12 @@ public CompletableFuture<? extends Response> queryAsync(long connectionId, JdbcQ
                 .exceptionally(t -> createErrorResult("Exception while executing query [query=" + req.sqlQuery() + "]", t, null));
     }
 
-    private static SqlProperties createProperties(JdbcStatementType stmtType, boolean multiStatement, ZoneId timeZoneId) {
+    private static SqlProperties createProperties(
+            JdbcStatementType stmtType, 
+            boolean multiStatement, 
+            ZoneId timeZoneId,
+            long timeoutMillis
```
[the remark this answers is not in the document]
lowka: Done.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImpl.java
```diff
@@ -226,26 +232,58 @@ public CompletableFuture<QueryPlan> prepareAsync(
...
+
+        return f;
+    }
+
+    /**
+     * Sets prepare callback.
+     *
+     * @param callback Callback.
+     */
+    public void setCallback(PrepareCallback callback) {
```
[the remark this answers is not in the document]
lowka: Replaced with a Runnable.

modules/sql-engine/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlSynchronousApiTest.java
```diff
@@ -131,4 +140,101 @@ public void process(ResultSet<SqlRow> resultSet) {
             resultSet.forEachRemaining(res::add);
         }
     }
+
+    @Test
+    public void testEarlyQueryTimeout() {
```
ygerzhedovich, line 145: these tests should be added and reworked accordingly to base ItSqlApiBaseTest class to test different APIs

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/IgniteSqlImpl.java
```diff
@@ -536,14 +536,18 @@ public CompletableFuture<Void> executeScriptAsync(String query, @Nullable Object
         }
 
         try {
+            SqlProperties properties = SqlPropertiesHelper.newBuilder()
```
ygerzhedovich, line 539: it looks better to add default query timeout to org.apache.ignite.internal.sql.engine.SqlQueryProcessor#DEFAULT_PROPERTIES instead of add the timeout to all places

modules/jdbc/src/integrationTest/java/org/apache/ignite/jdbc/ItJdbcStatementSelfTest.java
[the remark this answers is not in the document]
lowka: Fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareCallback.java
```diff
@@ -0,0 +1,57 @@
...
+     * Called for a valid SQL tree after optimization step completes. When optimization step completes successfully called with
+     * {@code optimizedRel} not null and {@code err} null. When optimization step completes with an error called with {@code optimizedRel}
+     * null and {@code err} not null.
+     *
+     * @param ctx Planning context.
+     * @param node Validated SQL tree.
+     * @param optimizedRel Optimized relational tree, if optimization was successful.
+     * @param err Error that occurred during optimization, if any.
+     */
+    default void afterOptimization(PlanningContext ctx, SqlNode node, @Nullable IgniteRel optimizedRel, @Nullable Throwable err) {
```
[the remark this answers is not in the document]
lowka: Fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
lowka: Fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -530,8 +540,12 @@ private CompletableFuture<QueryMetadata> prepareSingleAsync0(
     ) {
         SqlProperties properties0 = SqlPropertiesHelper.chain(properties, DEFAULT_PROPERTIES);
         String schemaName = properties0.get(QueryProperty.DEFAULT_SCHEMA);
+        Long queryTimeout = properties0.getOrDefault(QueryProperty.QUERY_TIMEOUT, 0L);
 
         QueryCancel queryCancel = new QueryCancel();
+        if (queryTimeout != 0) {
```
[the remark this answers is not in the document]
lowka: We decided to use 0 in case of no timeout.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/SqlQueryProcessor.java
```diff
@@ -558,9 +572,21 @@ private CompletableFuture<AsyncSqlCursor<InternalSqlRow>> querySingle(
     ) {
         String schemaName = properties.get(QueryProperty.DEFAULT_SCHEMA);
         ZoneId timeZoneId = properties.get(QueryProperty.TIME_ZONE_ID);
+        Long queryTimeout = properties.getOrDefault(QueryProperty.QUERY_TIMEOUT, 0L);
```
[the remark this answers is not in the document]
lowka: Fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/api/IgniteSqlImpl.java
lowka: Thanks. Done.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -1151,7 +1186,11 @@ private void enlist(SourceAwareIgniteRel rel) {
             }.visit(mappedFragment.fragment().root());
         }
 
-        private CompletableFuture<Void> close(boolean cancel) {
+        private CompletableFuture<Void> close() {
+            return close(null);
```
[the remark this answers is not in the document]
ygerzhedovich: Plus 100500 and remove @Nullable

modules/sql-engine/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlSynchronousApiTest.java
lowka: Fixed.

lowka:
Renamed `QueryCancel::addTimeout` to `QueryCancel::setTimeout`
Moved deadline and timeout future from `SqlOperationContext` into `QueryCancel`.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryCancel.java
```diff
@@ -17,45 +17,134 @@
...
+
+        CompletableFuture<Void> fut = new CompletableFuture<>();
+        fut.thenAccept((r) -> doCancel(Reason.TIMEOUT));
+
+        ScheduledFuture<?> f = scheduler.schedule(() -> {
+            fut.complete(null);
+        }, timeoutMillis, MILLISECONDS);
+
+        add((timeout) -> {
+            // Cancel the future if we didn't timeout.
```
xtern, line 105:
```suggestion
            // Cancel the future if we didn't timeout,
            // since in the case of a timeout it is already completed.
```
:thinking:

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryCancel.java
```diff
@@ -17,45 +17,134 @@
...
+            if (!timeout) {
+                f.cancel(false);
+            }
+        });
+
+        this.timeoutFut = fut;
+        this.deadline = Instant.now().plusMillis(timeoutMillis);
```
xtern, line 112:
```suggestion
        this.deadline = Instant.ofEpochMilli(FastTimestamps.coarseCurrentTimeMillis() + timeoutMillis);
```
:thinking:

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImpl.java
```diff
@@ -226,26 +233,59 @@ public CompletableFuture<QueryPlan> prepareAsync(
...
+
+        long actualPlannerTimeout;
+        Instant deadline = cancelHandler.deadline();
+        if (deadline == null) {
+            actualPlannerTimeout = plannerTimeout;
+        } else {
+            Instant now = Instant.now();
+            long timeout = Duration.between(now, deadline).toMillis();
+
+            actualPlannerTimeout = Math.min(timeout, plannerTimeout);
```
xtern, line 249: I have doubts about this.
We have different timeouts
1. query timeout
2. planner timeout

And at this point we already scheduled query timeout. When query timeout will happen we will abort planner and we will raise error about query timeout.

Why do we also need to reduce the planner timeout?
It looks like we may be throwing the wrong error, that a  planner timeout occurred, but the user needs to increase the request timeout.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
```diff
@@ -801,11 +846,14 @@ private void acknowledgeFragment(String nodeName, long fragmentId, @Nullable Thr
 
                 if (rootFragmentId0 != null && fragmentId == rootFragmentId0) {
                     root.completeExceptionally(ex);
+                } else if (root == null) {
```
xtern, line 849: This doesn't help me a lot, because in this case we fail on the last line of this method
```
remoteFragmentInitCompletion.get(new RemoteFragmentKey(nodeName, fragmentId)).complete(null);
```

I have almost 100% NPE when run `ItSqlAsynchronousApiTest.testQueryTimeout()`, and I'm sure that the root issue is not here (at least I seee that first NPE happen in executeFragment method) .

Since we don't observe any issues when running tests on TeamCity - let's just revert this change (if (root == null)) and investigate this separately?

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PrepareServiceImpl.java
lowka: Fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryCancel.java
lowka: Done.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryCancel.java
lowka: Done.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/exec/ExecutionServiceImpl.java
lowka: Fixed NPE.

lowka: Filed an issue 
https://issues.apache.org/jira/browse/IGNITE-22585

xtern (approved):

modules/sql-engine/src/integrationTest/java/org/apache/ignite/internal/sql/api/ItSqlApiBaseTest.java
```diff
@@ -894,6 +897,102 @@ public void testTimeZoneId(String timeZoneId) {
         assertEquals(nowMillis, tsMillis, deltaMillis);
     }
 
+    @Test
+    public void testEarlyQueryTimeout() {
```
ygerzhedovich, line 901: all new test work in synchronious manner, but  API should depend on subclasses. Please see another tests how it works

lowka: Fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PlanningContext.java
```diff
@@ -151,7 +151,7 @@ public List<MetadataHandler<?>> handlers(Class<? extends MetadataHandler<?>> hnd
     private final long plannerTimeout;
 
     /** Flag indicated if planning has been canceled due to timeout. */
-    private boolean timeouted = false;
+    private final AtomicBoolean timeouted = new AtomicBoolean();
```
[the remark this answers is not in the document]
lowka: Fixed.

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/QueryCancel.java
```diff
@@ -17,45 +17,136 @@
...
+    public @Nullable CompletableFuture<Void> timeoutFuture() {
+        return timeoutFut;
+    }
+
+    /**
+     * Returns the deadline of the operation.
+     *
+     * <p>Can be null if a query has no timeout.
+     */
+    public @Nullable Instant deadline() {
```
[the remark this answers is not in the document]
lowka: Fixed.

ygerzhedovich (approved):

PR #3964, opened 2024-06-20
IGNITE-17059 Thin 3.0: Implement batch SQL API for java thin client

xtern:
https://issues.apache.org/jira/browse/IGNITE-17059

* Added support for `SQL_EXEC_BATCH` operation on java thin side.
* `BatchedArguments` has been reworked to prevent possible issue with its serialization.


modules/api/src/main/java/org/apache/ignite/sql/BatchedArguments.java
```diff
@@ -18,24 +18,25 @@
...
 
 /**
  * Arguments for batch query execution.
- *
- * <p>TODO: replace inheritance with delegation.
- * TODO: add arguments length validation.
- * TODO: add named arguments support.
  */
-public class BatchedArguments extends ArrayList<List<Object>> implements List<List<Object>> {
+public final class BatchedArguments implements Iterable<List<Object>> {
```
ptupitsyn, line 29: We have size and positional access, let's implement `List` interface.

modules/api/src/main/java/org/apache/ignite/sql/BatchedArguments.java
```diff
@@ -73,8 +72,132 @@ public static BatchedArguments of(List<List<Object>> args) {
...
+     * Appends arguments to a batch.
+     *
+     * @param argsList Arguments list.
+     * @return {@code this} for chaining.
+     */
+    public BatchedArguments add(List<Object> argsList) {
+        Objects.requireNonNull(argsList, "argsList");
+
+        if (argsList.isEmpty()) {
+            throwEmptyArgumentsException();
```
ptupitsyn, line 90: The user can modify the list after this method call, do we handle this?

modules/client/src/main/java/org/apache/ignite/internal/client/sql/ClientSql.java
```diff
@@ -264,15 +258,57 @@ public <T> CompletableFuture<AsyncResultSet<T>> executeAsync(
...
+
+            w.out().packString(statement.query());
+            w.out().packBatchedArgumentsAsBinaryTupleArray(batch);
+            w.out().packLong(ch.observableTimestamp());
+        };
+
+        PayloadReader<BatchResultInternal> payloadReader = r -> {
+            ClientMessageUnpacker unpacker = r.in();
+
+            unpacker.skipValues(4); // skipping values that are not currently in use.
```
ptupitsyn, line 286: Please list every skipped value name here.

modules/client/src/main/java/org/apache/ignite/internal/client/sql/ClientSql.java
```diff
@@ -264,15 +258,57 @@ public <T> CompletableFuture<AsyncResultSet<T>> executeAsync(
...
+
+            long[] updateCounters = unpacker.unpackLongArray();
+
+            if (unpacker.tryUnpackNil()) {
+                unpacker.skipValues(2);
```
ptupitsyn, line 291: Please add a comment explaining what is skipped an why.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/sql/ClientSqlExecuteBatchRequest.java
```diff
@@ -102,16 +103,18 @@ public static CompletableFuture<Void> process(
...
+            String errorMessage,
+            UUID traceId) {
         out.packNil(); // resourceId
 
         out.packBoolean(false); // has row set
         out.packBoolean(false); // has more pages
         out.packBoolean(false); // was applied
         out.packLongArray(affectedRows); // affected rows
-        out.packShort(errorCode); // error code
+        out.packInt(errorCode); // error code
```
ptupitsyn, line 115: We have standardized error serialization in https://github.com/apache/ignite-3/blob/d3ab33124127dd18f36727148100d632fffbd02e/modules/client-handler/src/main/java/org/apache/ignite/client/handler/ClientInboundMessageHandler.java#L498 .

Let's remove error details from this response, and return failed future instead.

ptupitsyn (changes requested):

AMashenkov (approved):

modules/api/src/main/java/org/apache/ignite/sql/BatchedArguments.java
xtern: I'm not sure about this. This class is a part of public API.
java.util.List contains ~30 methods, that are not required by this container.
I think it's better to narrow down the functionality of this class and extend it later if needed.

modules/api/src/main/java/org/apache/ignite/sql/BatchedArguments.java
xtern: Thanks, I believe I fixed this.

modules/client/src/main/java/org/apache/ignite/internal/client/sql/ClientSql.java
xtern: Done.

modules/client/src/main/java/org/apache/ignite/internal/client/sql/ClientSql.java
xtern: Done.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/sql/ClientSqlExecuteBatchRequest.java
xtern: For SqlBatchException this doesn't work very well right now.
This exception includes an "update counter" which is lost during serialization.
This counter can be used by the user to understand how many updates were completed successfully (we have several tests for this).

modules/api/src/main/java/org/apache/ignite/sql/BatchedArguments.java
ptupitsyn: Ok.

modules/client-handler/src/main/java/org/apache/ignite/client/handler/requests/sql/ClientSqlExecuteBatchRequest.java
ptupitsyn: Ok, let's handle this as a separate ticket: https://issues.apache.org/jira/browse/IGNITE-22569

PR #3971, opened 2024-06-21
IGNITE-22441 Add parameter object to Data Streamer

ptupitsyn:
* Add `ReceiverDescriptor` parameter object to Data Streamer similar to `JobDescriptor` in Compute API.
* Unlike Compute, `DataStreamerOptions` are not included in `ReceiverDescriptor`, because semantically they are not related: options affect client-side streaming logic, and do not affect the receiver.

https://issues.apache.org/jira/browse/IGNITE-22441


modules/api/src/main/java/org/apache/ignite/table/ReceiverDescriptor.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        /**
+         * Sets the deployment units.
+         *
+         * @param units Deployment units.
+         * @return This builder.
+         */
+        public Builder units(List<DeploymentUnit> units) {
+            this.units = units;
```
PakhomovAlexander, line 100: Shall we copy List?

PakhomovAlexander (approved):

modules/api/src/main/java/org/apache/ignite/table/ReceiverDescriptor.java
ptupitsyn: We don't copy in `JobDescriptor`. Also, `units()` property exposes a mutable `List`, so we would need to copy there as well for full protection. I don't think it's worth it, let's keep as is for now.

PR #3980, opened 2024-06-24
IGNITE-22351 Add Doxyfile to C++ project

isapego:
https://issues.apache.org/jira/browse/IGNITE-22351

- Moved private headers from common to common/detail;
- Added Doxygen file to generate documentation;
- Updated DEVNOTES.md accordingly.


modules/platforms/cpp/cmake/dependencies.cmake
```diff
@@ -20,7 +20,7 @@ set(MBEDTLS_AS_SUBPROJECT ON)
 
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
-add_compile_definitions(MBEDTLS_USER_CONFIG_FILE="${CMAKE_CURRENT_SOURCE_DIR}/ignite/common/ignite_mbedtls_config.h")
+add_compile_definitions(MBEDTLS_USER_CONFIG_FILE="${CMAKE_CURRENT_SOURCE_DIR}/ignite/common/detail/ignite_mbedtls_config.h")
```
ptupitsyn, line 23: Here and below: unrelated changes? Do we need this for Doxyfile?

isapego: Well, it's not directly related, but there were files in public space which should not be public, and Doxygen would generate documentation using them. I could exclude them manually, but the right way looked like moving them to `detail` folder (ie `impl`)

ptupitsyn (approved):

arcolight (approved):
LGTM

PR #4007, opened 2024-06-28
IGNITE-22434 Do not inherit TaskExecution from JobExecution

Pochatkin:
https://issues.apache.org/jira/browse/IGNITE-22434

### DIFF
--- a/modules/client/src/test/java/org/apache/ignite/client/fakes/FakeCompute.java
+++ b/modules/client/src/test/java/org/apache/ignite/client/fakes/FakeCompute.java
@@ -50,7 +50,6 @@ import org.apache.ignite.compute.JobState;
 import org.apache.ignite.compute.JobStatus;
 import org.apache.ignite.compute.JobTarget;
 import org.apache.ignite.compute.TaskState;
-import org.apache.ignite.compute.TaskStatus;
 import org.apache.ignite.compute.task.TaskExecution;
 import org.apache.ignite.deployment.DeploymentUnit;
 import org.apache.ignite.internal.compute.ComputeUtils;


modules/api/src/main/java/org/apache/ignite/compute/task/TaskExecution.java
```diff
@@ -50,4 +50,48 @@ public interface TaskExecution<R> extends JobExecution<R> {
                 .map(state -> state != null ? state.id() : null)
                 .collect(Collectors.toList()));
     }
+
+    /**
+     * Returns task's execution result.
+     *
+     * @return Job's execution result future.
```
ptupitsyn, line 57: Here and below `job` -> `task`.

modules/api/src/main/java/org/apache/ignite/compute/task/TaskExecution.java
```diff
@@ -50,4 +50,48 @@ public interface TaskExecution<R> extends JobExecution<R> {
...
+     */
+    CompletableFuture<R> resultAsync();
+
+    /**
+     * Returns the current state of the task. The task state may be deleted and thus return {@code null} if the time for retaining task
+     * state has been exceeded.
+     *
+     * @return The current state of the task, or {@code null} if the task state no longer exists due to exceeding the retention time limit.
+     */
+    CompletableFuture<@Nullable JobState> stateAsync();
```
ptupitsyn, line 67: I think we also need `TaskState` and `TaskStatus`.

ptupitsyn (changes requested):

ptupitsyn (approved):

PR #4011, opened 2024-06-28
IGNITE-22603 Add InetSocketAddress caching for DefaultMessagingService

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-22603

### The Review Checklist

### Notes

### DIFF
--- a/modules/network/src/main/java/org/apache/ignite/internal/network/DefaultMessagingService.java
+++ b/modules/network/src/main/java/org/apache/ignite/internal/network/DefaultMessagingService.java
@@ -31,6 +31,7 @@ import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
@@ -115,13 +116,22 @@ public class DefaultMessagingService extends AbstractMessagingService {
     private volatile BiPredicate<String, NetworkMessage> dropMessagesPredicate;
 
     /**
+     * Cache of {@link InetSocketAddress} of recipient nodes ({@link ClusterNode}) that are in the topology and not stale.
+     *
+     * <p>Introduced for optimization - reducing the number of address resolving for the same nodes.</p>
+     */
+    private final Map<String, InetSocketAddress> recipientInetAddrByNodeId = new ConcurrentHashMap<>();
+
+    /**
      * Constructor.
      *
+     * @param nodeName Consistent ID (aka name) of the local node associated with the service to create.
      * @param factory Network messages factory.
      * @param topologyService Topology service.
      * @param staleIdDetector Used to detect stale node IDs.
      * @param classDescriptorRegistry Descriptor registry.
      * @param marshaller Marshaller.
+     * @param criticalWorkerRegistry Used to register critical threads managed by the new service and its components.
      */
     public DefaultMessagingService(
             String nodeName,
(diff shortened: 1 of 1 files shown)


modules/network/src/main/java/org/apache/ignite/internal/network/DefaultMessagingService.java
```diff
@@ -674,15 +693,15 @@ protected void onStoppingInitiated() {
      * @param recipientNode Target cluster node.
      */
     private @Nullable InetSocketAddress resolveRecipientAddress(ClusterNode recipientNode) {
-        NetworkAddress recipientAddress = recipientNode.address();
-
         // Node name is {@code null} if the node has not been added to the topology.
         if (recipientNode.name() != null) {
-            return connectionManager.consistentId().equals(recipientNode.name()) ? null : createResolved(recipientAddress);
+            return connectionManager.consistentId().equals(recipientNode.name()) ? null : getFromCacheOrCreateResolved(recipientNode);
```
ibessonov, line 698: Should we do the same on line 714?

modules/network/src/main/java/org/apache/ignite/internal/network/DefaultMessagingService.java
```diff
@@ -705,4 +724,18 @@ protected void onStoppingInitiated() {
     private static InetSocketAddress createResolved(NetworkAddress address) {
         return new InetSocketAddress(address.host(), address.port());
     }
+
+    private InetSocketAddress getFromCacheOrCreateResolved(ClusterNode recipientNode) {
+        assert recipientNode.name() != null : "Node has not been added to the topology: " + recipientNode.id();
+
+        InetSocketAddress address = recipientInetAddrByNodeId.compute(recipientNode.id(), (nodeId, inetSocketAddress) -> {
```
ibessonov, line 731: We can theoretically resurrect the cache if we send a message to offline node, how are we avoiding it?
Why don't we cache address instance inside of `ClusterNode` like we did in Ignite 2, why map instead?

modules/network/src/main/java/org/apache/ignite/internal/network/DefaultMessagingService.java
tkalkirill: I think not, the node is not in the topology.
I thought about it, we may not delete this address later and it will hang in the air.

modules/network/src/main/java/org/apache/ignite/internal/network/DefaultMessagingService.java
tkalkirill: > We can theoretically resurrect the cache if we send a message to offline node, how are we avoiding it?

If I understand correctly, when a node becomes offline, it ends up in the “list” of stale nodes, by performing the check below we will avoid being filled with such nodes.
```
if (staleIdDetector.isIdStale(nodeId)) {
    return null;
}
```

> Why don't we cache address instance inside of ClusterNode like we did in Ignite 2, why map instead?

`org.apache.ignite.network.ClusterNode` is a public API, I don’t think the user needs to give the `InetSocketAddress`, so an internal map is used. What do you think?

ibessonov: > by performing the check below we will avoid being filled with such nodes.

This is cool, I completely forgot about it!
> is a public API

Ok, I see, although it's an interface and we could hide resolved address in the implementation. Anyway, we shouldn't use the same classes for private and public API, makes no sense to me, but it's not related to your changes.

> What do you think?
I agree with you, map will do it for now, thank you!

ibessonov (approved):

PR #4013, opened 2024-06-28
IGNITE-22527 C++ Add parameter object to Compute API

isapego:
https://issues.apache.org/jira/browse/IGNITE-22527


arcolight (approved):
LGTM

PR #4056, opened 2024-07-08
IGNITE-22661 Add catalog version into Assignments

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-22661


modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -53,32 +53,38 @@ public class Assignments implements Serializable {
...
     }
 
     /**
      * Creates a new instance.
      *
      * @param nodes Set of nodes.
      */
-    public static Assignments of(Set<Assignment> nodes) {
-        return new Assignments(nodes, false);
+    public static Assignments of(int catalogVersion, Set<Assignment> nodes) {
```
JAkutenshi, line 77: Isn't better to move `catalogVersion` to the end as in other occurrences?

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -53,32 +53,38 @@ public class Assignments implements Serializable {
...
     }
 
     /**
      * Creates a new instance.
      *
      * @param nodes Array of nodes.
      */
-    public static Assignments of(Assignment... nodes) {
-        return new Assignments(Arrays.asList(nodes), false);
+    public static Assignments of(int catalogVersion, Assignment... nodes) {
```
JAkutenshi, line 86: This looks like an exceptional case due to above comment..

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/RebalanceRaftGroupEventsListener.java
```diff
@@ -429,14 +430,22 @@ static void doStableKeySwitch(
...
                     ? 0L : bytesToLongKeepingOrder(stableChangeTriggerEntry.value());
 
             if (!retrievedPending.equals(stableFromRaft)) {
                 return;
             }
 
+            int catalogVersion = pendingAssignments.catalogVersion();
+
+            Set<Assignment> calculatedAssignments =
+                    calculateAssignments(tablePartitionId, catalogService, distributionZoneManager, catalogVersion).get();
```
JAkutenshi, line 447: Isn't a timeout should be placed there?

modules/distribution-zones/src/test/java/org/apache/ignite/internal/distributionzones/rebalance/DistributionZoneRebalanceEngineTest.java
```diff
@@ -435,7 +435,7 @@ void replicasTriggersAssignmentsChangingOnNonDefaultZones() throws Exception {
 
         when(distributionZoneManager.dataNodes(anyLong(), anyInt(), anyInt())).thenReturn(completedFuture(Set.of("node0")));
 
-        byte[] assignmentsBytes = Assignments.of(Assignment.forPeer("node0")).toBytes();
+        byte[] assignmentsBytes = Assignments.of(1, Assignment.forPeer("node0")).toBytes();
```
JAkutenshi, line 438: Why `1`? There and everywhere.

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -53,32 +53,38 @@ public class Assignments implements Serializable {
      */
     private final boolean force;
 
+    /**
+     * Version of Catalog that matches current assignments.
+     */
+    private final int catalogVersion;
```
sanpwc, line 59: I don't like how it looks. Formally Assignments is a part of affinity module that should not depend on catalog. Let's revise it one more time. Let's revise an idea of using a timestamp here.

modules/affinity/src/main/java/org/apache/ignite/internal/affinity/Assignments.java
```diff
@@ -53,32 +55,36 @@ public class Assignments implements Serializable {
      */
     private final boolean force;
 
+    /** Time when the catalog version that the assignments were calculated against becomes active (i. e. available for use). */
+    private final HybridTimestamp timestamp;
```
sanpwc, line 59: Curios how it'll be serialized, as an object or as long? I'm talking about writing to the MS and not messages here, because in messaging it should be optimised. Basically I's asking whether we should use long instead of HybridTimestamp here?

modules/distribution-zones/src/main/java/org/apache/ignite/internal/distributionzones/rebalance/DistributionZoneRebalanceEngine.java
```diff
@@ -204,10 +206,14 @@ public CompletableFuture<Void> onUpdate(WatchEvent evt) {
...
 
                     // It is safe to get the latest version of the catalog as we are in the metastore thread.
-                    // TODO: IGNITE-22661 Potentially unsafe to use the latest catalog version, as the tables might not already present
+                    // TODO: IGNITE-22723 Potentially unsafe to use the latest catalog version, as the tables might not already present
                     //  in the catalog. Better to store this version when writing datanodes.
                     int catalogVersion = catalogService.latestCatalogVersion();
 
+                    Catalog catalog = catalogService.catalog(catalogVersion);
+
+                    HybridTimestamp assignmentsTimestamp = HybridTimestamp.hybridTimestamp(catalog.time());
```
sanpwc, line 215: I assume that given code will be adjusted in IGNITE-22723, right?

sanpwc (approved):

PR #4061, opened 2024-07-09
IGNITE-22693 Add protection against internal thread hijacking to PartitionManager

rpuch:
https://issues.apache.org/jira/browse/IGNITE-22693

### DIFF
--- a/modules/table/src/integrationTest/java/org/apache/ignite/internal/threading/ItPartitionManagerApiThreadingTest.java
+++ b/modules/table/src/integrationTest/java/org/apache/ignite/internal/threading/ItPartitionManagerApiThreadingTest.java
@@ -1,0 +1,120 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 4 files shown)


modules/table/src/integrationTest/java/org/apache/ignite/internal/threading/ItPartitionManagerApiThreadingTest.java
```diff
@@ -0,0 +1,120 @@
...
+    private static final Tuple KEY_TUPLE = Tuple.create().set("id", 1);
+
+    @Override
+    protected int initialNodes() {
+        return 1;
+    }
+
+    @BeforeEach
+    void createTable() {
+        sql("CREATE TABLE " + TABLE_NAME + " (id INT PRIMARY KEY, val VARCHAR)");
```
sashapolo, line 57: Why do we need to create a drop a table for every test? Can we move it to `BeforeAll`?

sashapolo (approved):

modules/table/src/integrationTest/java/org/apache/ignite/internal/threading/ItPartitionManagerApiThreadingTest.java
rpuch: 'Table per test' allows to catch test failures; otherwise they were failing really rarely even without the fix

PR #4062, opened 2024-07-09
IGNITE-22150 .NET: Add PartitionManager API

ptupitsyn:
* Add `ITable.PartitionManager` API (`GetPrimaryReplicas`, `GetPartition`)
* Cache results based on known `PartitionAssignmentTimestamp`


modules/platforms/dotnet/Apache.Ignite/Table/IPartition.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System.Diagnostics.CodeAnalysis;
+
+/// <summary>
+/// Table partition.
+/// </summary>
+[SuppressMessage("Design", "CA1040:Avoid empty interfaces", Justification = "Reviewed.")]
+public interface IPartition
+{
+}
```
isapego, line 28: Hmm, should not we at least support comparison operations?

isapego (approved):

modules/platforms/dotnet/Apache.Ignite/Table/IPartition.cs
ptupitsyn: Do you mean relative comparison, or just equality?

Value-based equality is supported by the `HashPartition` implementation (two instances with the same id are considered equal).

ptupitsyn: After some thinking I agree. Added [IEquatable<T>](https://learn.microsoft.com/en-us/dotnet/api/system.iequatable-1?view=net-8.0) to the interface to make things clearer.

PR #4065, opened 2024-07-09
IGNITE-22619 Add `InternalTable#estimatedSize` method

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-22619

### The Review Checklist

### Notes


modules/partition-replicator/src/main/java/org/apache/ignite/internal/partition/replicator/network/replication/GetEstimatedSizeRequest.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.apache.ignite.internal.network.annotations.Transferable;
+import org.apache.ignite.internal.partition.replicator.network.PartitionReplicationMessageGroup;
+import org.apache.ignite.internal.replicator.message.ReplicaRequest;
+
+/**
+ * Request for getting an estimated size of a partition.
+ */
+@Transferable(PartitionReplicationMessageGroup.GET_ESTIMATED_SIZE_MESSAGE)
+public interface GetEstimatedSizeRequest extends ReplicaRequest {
```
sanpwc, line 28: Do you expect to process given request on primary only?

sashapolo: yes (I think). Does it make sense to retrieve the size of the partition from a non-primary replica? Will it bring any performance improvements?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -2231,6 +2232,27 @@ public ScheduledExecutorService streamerFlushExecutor() {
...
     }
 
+    @Override
+    public CompletableFuture<Long> estimatedSize() {
+        var invokeFutures = new CompletableFuture<?>[partitions];
+
+        for (int partId = 0; partId < partitions; partId++) {
+            var replicaGroupId = new TablePartitionId(tableId, partId);
+
+            invokeFutures[partId] = partitionLocation(replicaGroupId)
```
sanpwc, line 2242: I'd rather implement it in a bit different manner. For each partition I'd call awaitPrimaryReplica() in order to define the location. Just like we do for any other request.

modules/partition-replicator/src/main/java/org/apache/ignite/internal/partition/replicator/network/replication/GetEstimatedSizeRequest.java
sanpwc: It might be useful in some cases: e.g. in order to gather statistics for sql and to ease failure analysis. However within given ticket I'd expect give request to be primary only and thus I'd rather add `extends PrimaryReplicaRequest`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -517,6 +522,10 @@ private CompletableFuture<?> processRequest(ReplicaRequest request, @Nullable Bo
                         processOperationRequestWithTxRwCounter(senderId, request, isPrimary, opTsIfDirectRo, leaseStartTime));
     }
 
+    private CompletableFuture<Long> processGetEstimatedSizeRequest() {
+        return completedFuture(mvDataStorage.estimatedSize());
```
sanpwc, line 526: Curious whether estimatedSize evaluates writeIntent as +1?

modules/table/src/integrationTest/java/org/apache/ignite/internal/table/ItEstimatedSizeTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+    @Test
+    void testEstimatedSizeAfterScaleUp() throws InterruptedException {
+        for (String profile : ALL_STORAGE_PROFILES) {
+            String tableName = createTableWithData(profile);
+
+            assertThat(tableSize(tableName), willBe(NUM_ROWS));
+        }
+
+        cluster.startNode(initialNodes());
```
sanpwc, line 107: In case of primary-only logic that test is almost meaningless.

modules/table/src/integrationTest/java/org/apache/ignite/internal/table/ItEstimatedSizeTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+        for (String profile : ALL_STORAGE_PROFILES) {
+            waitForRebalance(initialNodes() + 1);
+
+            assertThat(tableSize(tableName(profile)), willBe(NUM_ROWS));
+        }
+    }
+
+    @Test
+    void testEstimatedAfterScaleDown() throws InterruptedException {
```
sanpwc, line 117: Same as above.

modules/table/src/integrationTest/java/org/apache/ignite/internal/table/ItEstimatedSizeTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+        cluster.stopNode(initialNodes() - 1);
+
+        for (String profile : ALL_STORAGE_PROFILES) {
+            waitForRebalance(initialNodes() - 1);
+
+            assertThat(tableSize(tableName(profile)), willBe(NUM_ROWS));
+        }
+    }
+
```
sanpwc, line 132: - I'd rather add test that will check whether pending transactions (that not yet committed) effect the size.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -2231,6 +2232,27 @@ public ScheduledExecutorService streamerFlushExecutor() {
         return streamerFlushExecutor.get();
     }
 
+    @Override
+    public CompletableFuture<Long> estimatedSize() {
+        var invokeFutures = new CompletableFuture<?>[partitions];
+
+        for (int partId = 0; partId < partitions; partId++) {
```
sanpwc, line 2239: What if some partitions are unavailable? Is it intended to await all of them or we should also have best effort solution?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sashapolo: No, Write Intents are going to be omitted, only committed entries change the size

modules/table/src/integrationTest/java/org/apache/ignite/internal/table/ItEstimatedSizeTest.java
sashapolo: I already have such tests on the Partition Storage level, I think writing full integration tests for that is not needed

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
sashapolo: How do we usually handle this? What will happen to the `SELECT *` query in this case? I think we should mimic the full scan behaviour here

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
sanpwc: Sounds reasonable while we retrieving data from primary.

modules/table/src/integrationTest/java/org/apache/ignite/internal/table/ItEstimatedSizeTest.java
sanpwc: Yep, makes sense.

modules/table/src/integrationTest/java/org/apache/ignite/internal/table/ItEstimatedSizeTest.java
sashapolo: These are integration tests for the logic of applying snapshots to partition storages. So here I try to test a lower-level issue: that `MvPartitionStorage#size` returns a correct value after a partition was rebalanced.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
sanpwc: Implicit `Select *` is an RO wrapped operation, thus it might be executed on backup even without majority - in that case, behavior will differ. All in all that sounds like a follow up ticket that will add an ability to retrieve size from backups.

sashapolo: Discussed this in person: there's no need for a best-effort approach, we will wait for the partition to appear

modules/table/build.gradle
```diff
@@ -64,6 +64,7 @@ dependencies {
     testImplementation project(':ignite-page-memory')
     testImplementation project(':ignite-storage-rocksdb')
     testImplementation project(':ignite-placement-driver-api')
+    testImplementation project(':ignite-placement-driver')
```
sanpwc, line 67: Curious why placement-driver-api is not enough?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -1918,43 +1913,39 @@ protected CompletableFuture<IgniteBiTuple<ClusterNode, Long>> enlist(int partId,
...
+                            "Failed to get the primary replica [tablePartitionId=" + tablePartitionId + ", awaitTimestamp=" + now + ']',
+                            e
+                    );
+                });
     }
 
-    private ClusterNode getClusterNode(@Nullable String leaserHolder) {
-        ClusterNode node = clusterNodeResolver.getByConsistentId(leaserHolder);
+    private ClusterNode getClusterNode(ReplicaMeta replicaMeta) {
+        String leaseHolder = replicaMeta.getLeaseholder();
```
sanpwc, line 1942: Seems that using getLeaseholderId will be more robust.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/storage/InternalTableEstimatedSizeTest.java
```diff
@@ -0,0 +1,471 @@
...
+                            .build()
+                    );
+        }).when(messagingService).invoke(eq(nodeName), any(ReplicaRequest.class), anyLong());
+    }
+
+    @AfterEach
+    void tearDown() {
+        Collections.reverse(components);
+
+        assertThat(stopAsync(componentContext, components), willCompleteSuccessfully());
```
sanpwc, line 250: beforeNodeStop is missing.

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/storage/InternalTableEstimatedSizeTest.java
```diff
@@ -0,0 +1,471 @@
...
+    private final HybridClockImpl clock = new HybridClockImpl();
+
+    private final ComponentContext componentContext = new ComponentContext();
+
+    private final List<MvPartitionStorage> partitionStorages = new ArrayList<>(PARTITIONS_NUM);
+
+    private final List<IgniteComponent> components = new ArrayList<>();
+
+    @BeforeEach
+    void setUp(
```
sanpwc, line 142: Great work first of all. However given test seems to be very fragile because it mainly tests mocks. Did you consider using real ignite nodes with some message blocking and transferPrimary/killing nodes? Such tests will be longer but more robust, WDYT?

sanpwc (approved):

modules/table/build.gradle
sashapolo: Because I need the `Lease` class

modules/table/src/test/java/org/apache/ignite/internal/table/distributed/storage/InternalTableEstimatedSizeTest.java
sashapolo: > Did you consider using real ignite nodes with some message blocking

I did, bu writing such tests and validating exact conditions that I want is a nightmare. I don't fully agree that it mainly tests mocks. My main goal was to see what exceptions are thrown by the `ReplicaService` in different conditions in order to check the exception types and these tests helped me achieve that

PR #4067, opened 2024-07-10
IGNITE-22682 Use IndexMeta in FullStateTransferIndexChooser

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-22682

### The Review Checklist

### Notes


modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -1246,7 +1246,7 @@ public static CompletableFuture<Void> stopAsync(Supplier<CompletableFuture<Void>
      * @param components Array of ignite components to stop.
      * @return CompletableFuture that will be completed when all components are stopped.
      */
-    public static CompletableFuture<Void> stopAsync(ComponentContext componentContext, IgniteComponent... components) {
+    public static CompletableFuture<Void> stopAsync(ComponentContext componentContext, @Nullable IgniteComponent... components) {
```
Phillippko, line 1249: IMHO nullable is confusing there - it feels like components can be nullable, and I think we don't have to annotate that varargs can be empty

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/FullStateTransferIndexChooser.java
```diff
@@ -273,137 +261,37 @@ private void onIndexRemoved(RemoveIndexEventParameters parameters) {
...
-            int finalCatalogVersion = catalogVersion;
-
-            var indexIds = new HashSet<Integer>();
-
-            catalogService.indexes(finalCatalogVersion).forEach(index -> {
-                tableVersionByIndexId.computeIfAbsent(index.id(), i -> tableVersionBusy(index, finalCatalogVersion));
-
-                if (index.status() == STOPPING) {
-                    stoppingActivationTsByIndexId.computeIfAbsent(index.id(), i -> catalogActivationTimestampBusy(finalCatalogVersion));
+                if (indexMeta.status() == MetaIndexStatus.READ_ONLY) {
```
Phillippko, line 271: I think we could move checking READ_ONLY status to the top, it looks like the simplest condition to check and the most important

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/FullStateTransferIndexChooser.java
```diff
@@ -413,17 +301,30 @@ private void onLwmChanged(ChangeLowWatermarkEventParameters parameters) {
...
-                    tableVersionByIndexId.remove(readOnlyIndexInfo.indexId());
-                }
-            }
-        });
+        long activationTs;
+
+        if (indexMeta.statusChanges().containsKey(MetaIndexStatus.STOPPING)) {
+            activationTs = activationTs(indexMeta, MetaIndexStatus.STOPPING);
+        } else {
+            activationTs = activationTs(indexMeta, MetaIndexStatus.READ_ONLY);
```
Phillippko, line 320: Let's add Nullable to indexMeta.statusChange()? javadoc - "Returns the index status change, {@code null} if absent"

Phillippko (approved):
LGTM, with small comments

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/FullStateTransferIndexChooser.java
tkalkirill: No need, there is method **org.apache.ignite.internal.table.distributed.index.IndexMeta#statusChangeNullable** for this.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/FullStateTransferIndexChooser.java
tkalkirill: We cannot, because it is in this area that we are protected from changing the low watermark and we are guaranteed not to lose the `IndexMeta`.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/snapshot/FullStateTransferIndexChooser.java
Phillippko: Didn't notice the exception - then javadoc for statusChange() is incorrect, about null if absent))

PR #4068, opened 2024-07-10
IGNITE-20415 Internal IncompatibleSchemaException is thrown from public API

rpuch:
https://issues.apache.org/jira/browse/IGNITE-20415


modules/api/src/main/java/org/apache/ignite/tx/StaleSchemaRollbackException.java
```diff
@@ -0,0 +1,39 @@
...
+package org.apache.ignite.tx;
+
+import java.util.UUID;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Thrown when, during an attempt to execute a commit, it turns out that the schema at the moment of commit is not compatible
+ * with the schema at the moment when the transaction was started and, hence, the transaction was aborted instead.
+ */
+public class StaleSchemaRollbackException extends RollbackException {
```
ptupitsyn, line 27: This looks very similar to `IncompatibleSchemaException`, is there any significant difference? Can we keep only one of those exceptions?

modules/api/src/main/java/org/apache/ignite/tx/IncompatibleSchemaException.java
```diff
@@ -0,0 +1,39 @@
...
+package org.apache.ignite.tx;
+
+import java.util.UUID;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Thrown when, during an attempt to execute a transactional operation, it turns out that the operation cannot be executed
+ * because an incompatible schema change has happened.
+ */
+public class IncompatibleSchemaException extends TransactionException {
```
ptupitsyn, line 27: Not sure if we should inherit from `TransactionException` here. 

All table operations are transactional in Ignite 3, so technically any schema incompatibility issue is related to some transaction, but it does not mean we should inherit all table update exceptions from `TransactionException`.

modules/api/src/main/java/org/apache/ignite/tx/StaleSchemaRollbackException.java
rpuch: `StaleSchemaRollbackException` means that an attempt to commit failed and the transaction was actually aborted. This seems to be an important piece of information. The user could do `catch (RollbackException)` to catch any 'could not commit, instead aborted', including the case when it happened due to an incompatible schema change.

On the other hand, `IncompatibleSchemaException` is not about a commit attempt.

modules/api/src/main/java/org/apache/ignite/tx/IncompatibleSchemaException.java
rpuch: Made `IncompatibleSchemaException` extend `IgniteException` and moved it to `org.apache.ignite`.

rpuch: After thinking about this a bit more, I've reverted this change, so it's again a `TransactionException` defined in `org.apache.ignite.tx`. The reasoning is that this type of exceptions can only be thrown by transactional operations (like `put()` or `commit()`), and yet not all API operations in Ignite 3 are transactional: for example, there is Compute API which isn't.

sk0x50 (approved):

PR #4070, opened 2024-07-11
IGNITE-22710 Add exception translation to Transaction implementations

rpuch:
https://issues.apache.org/jira/browse/IGNITE-22710


modules/core/src/main/java/org/apache/ignite/internal/lang/IgniteExceptionMapperUtil.java
```diff
@@ -115,7 +136,7 @@ public static Throwable mapToPublicException(Throwable origin) {
         }
 
         // There are no exception mappings for the given exception. This case should be considered as internal error.
-        return new IgniteException(INTERNAL_ERR, origin);
+        return unknownProblemMapper.apply(origin);
```
sk0x50, line 139: Should we check that `unknownProblemMapper` returns an instance of the `public` exception?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TransactionsExceptionMapperUtil.java
```diff
@@ -0,0 +1,85 @@
...
+                    if (err != null) {
+                        throw new CompletionException(mapToPublicTransactionException(unwrapCause(err), defaultCode));
+                    }
+
+                    return res;
+                });
+    }
+
+    private static Throwable mapToPublicTransactionException(Throwable origin, int defaultCode) {
+        if (origin instanceof PrimaryReplicaExpiredException) {
```
sk0x50, line 59: You have already added `PrimaryReplicaExpiredException` to the `TransactionExceptionMapperProvider`. So, what is the reason for mapping it here?

modules/transactions/src/test/java/org/apache/ignite/internal/tx/TxManagerTest.java
```diff
@@ -343,7 +343,10 @@ void testRepeatedCommitRollbackAfterCommitWithException() throws Exception {
 
         TransactionException transactionException = assertThrows(TransactionException.class, tx::commit);
 
-        assertInstanceOf(MismatchingTransactionOutcomeException.class, ExceptionUtils.unwrapCause(transactionException.getCause()));
+        assertInstanceOf(
+                MismatchingTransactionOutcomeException.class,
+                ExceptionUtils.unwrapCause(transactionException.getCause().getCause().getCause())
```
sk0x50, line 348: Perhaps, using `IgniteTestUtils.hasCause` would be more readable.

modules/core/src/main/java/org/apache/ignite/internal/lang/IgniteExceptionMapperUtil.java
rpuch: Added an assertion

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/TransactionsExceptionMapperUtil.java
rpuch: Good catch, thanks! Removed it from from `TransactionExceptionMapperProvider`

modules/transactions/src/test/java/org/apache/ignite/internal/tx/TxManagerTest.java
rpuch: Changed to `IgniteTestUtils.hasCause()`

sk0x50 (approved):

modules/core/src/main/java/org/apache/ignite/internal/lang/IgniteExceptionMapperUtil.java
```diff
@@ -115,7 +134,20 @@ public static Throwable mapToPublicException(Throwable origin) {
...
+    private static Throwable mapCheckingResultIsPublic(Throwable origin, Function<Throwable, Throwable> unknownProblemMapper) {
+        Throwable result = unknownProblemMapper.apply(origin);
+
+        checkResultIsPublic(result, origin);
+
+        return result;
+    }
+
+    private static void checkResultIsPublic(Throwable mappingResult, Throwable origin) {
+        assert mappingResult instanceof IgniteException || mappingResult instanceof IgniteCheckedException :
```
sanpwc, line 149: I'd also add an assertion that mappingResult is not part of any internal package.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/StaleTransactionOperationException.java
```diff
@@ -21,10 +21,10 @@
 import static org.apache.ignite.lang.ErrorGroups.Transactions.TX_STALE_OPERATION_ERR;
 
 import java.util.UUID;
-import org.apache.ignite.tx.TransactionException;
+import org.apache.ignite.internal.tx.TransactionInternalException;
 
 /** Error that occurs when a stale operation of a completed transaction is detected. */
-public class StaleTransactionOperationException extends TransactionException {
+public class StaleTransactionOperationException extends TransactionInternalException {
```
sanpwc, line 27: Why it's internal? I'd rather move it to public API. Generally I believe that we should go through all our exceptions case by case and define whether they are semantically internal or not.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/MismatchingTransactionOutcomeException.java
```diff
@@ -19,15 +19,13 @@
...
-import org.apache.ignite.tx.TransactionException;
-
 /**
  * The exception is thrown when the transaction result differs from the intended one.
  *
  * <p>For example, {@code tx.commit()} is called for a transaction, but the verification logic decided to abort it instead. The transaction
  * will be finished with {@link TxState#ABORTED} and the call to {@code tx.commit()} will throw this exception.
  */
-public class MismatchingTransactionOutcomeException extends TransactionException {
+public class MismatchingTransactionOutcomeException extends TransactionInternalException {
```
sanpwc, line 28: Semantically it's not internal.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/table/ItDurableFinishTest.java
```diff
@@ -310,7 +310,7 @@ void testCommitAlreadyAbortedTx() throws ExecutionException, InterruptedExceptio
 
         Throwable cause = ExceptionUtils.unwrapCause(transactionException.getCause());
 
-        assertInstanceOf(MismatchingTransactionOutcomeException.class, cause);
+        assertInstanceOf(MismatchingTransactionOutcomeException.class, cause.getCause());
```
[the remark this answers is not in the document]
rpuch: It cannot be public as it brings some internal state. I was going to add some kind of a public counterpart for it, like `RollbackException`, in IGNITE-20415 (which is the root cause for the current issue)

modules/core/src/main/java/org/apache/ignite/internal/lang/IgniteExceptionMapperUtil.java
```diff
@@ -84,9 +85,29 @@ static void registerMapping(
...
+    }
+
+    /**
+     * This method provides a mapping from internal exception to Ignite public ones.
+     *
+     * <p>The rules of mapping are the following:</p>
+     * <ul>
+     *     <li>any instance of {@link Error} is returned as is, except {@link AssertionError}
+     *     that will always be mapped to {@link IgniteException} with the {@link Common#INTERNAL_ERR} error code.</li>
+     *     <li>any instance of {@link IgniteException} or {@link IgniteCheckedException} is returned as is.</li>
```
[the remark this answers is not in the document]
rpuch: I think that the fact that our exceptions extend public 'root' exceptions, but get placed in internal packages, should be fixed, but not in this PR.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/MismatchingTransactionOutcomeException.java
rpuch: Please take a look at https://github.com/apache/ignite-3/pull/4070#discussion_r1678848166

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/StaleTransactionOperationException.java
rpuch: To change public API, we need some discussion. I was planning to do it in IGNITE-20415. I added a TODO for now.

modules/core/src/main/java/org/apache/ignite/internal/lang/IgniteExceptionMapperUtil.java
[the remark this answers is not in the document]
rpuch: I filed an issue for this https://issues.apache.org/jira/browse/IGNITE-22748

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/StaleTransactionOperationException.java
sanpwc: Not sure about discussion within context of such changes. Anyway ignite-20415 it is.

sanpwc (approved):

PR #4072, opened 2024-07-11
IGNITE-22709 Use IndexMeta in ChangeIndexStatusTask

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-22709

### The Review Checklist

### Notes


modules/index/src/main/java/org/apache/ignite/internal/index/ChangeIndexStatusTask.java
```diff
@@ -333,8 +335,26 @@ private <T> CompletableFuture<T> inBusyLocks(Supplier<CompletableFuture<T>> supp
 
         try {
             return supplier.get();
+        } catch (Throwable t) {
```
rpuch, line 338: We should not swallow `Error`s (which are not `AssertionError`). For them, we should probably log them and rethrow them after this.

modules/index/src/main/java/org/apache/ignite/internal/index/ChangeIndexStatusTask.java
```diff
@@ -333,8 +335,26 @@ private <T> CompletableFuture<T> inBusyLocks(Supplier<CompletableFuture<T>> supp
...
         } finally {
             leaveBusy();
         }
     }
+
+    private MetaIndexStatusChange statusChange() {
+        IndexMeta indexMeta = indexMetaStorage.indexMeta(indexDescriptor.id());
+
+        if (indexMeta == null) {
+            // Index was destroyed under a low watermark, well, we need to build it.
```
rpuch, line 349: The comment says that the index was already destroyed, but it also says that it needs to build it. Is this a typo? Should it say that it does NOT need to build it anymore?

modules/index/src/main/java/org/apache/ignite/internal/index/ChangeIndexStatusTask.java
```diff
@@ -333,8 +335,26 @@ private <T> CompletableFuture<T> inBusyLocks(Supplier<CompletableFuture<T>> supp
...
+            throw new IndexTaskStoppingException();
+        }
+
+        MetaIndexStatusChange statusChange = indexMeta.statusChangeNullable(MetaIndexStatus.convert(indexDescriptor.status()));
+
+        assert statusChange != null : IgniteStringFormatter.format("Missing index status change: [indexId={}, catalogStatus={}]",
```
rpuch, line 355: What if we run without assertions? We'll just return null and get an NPE further. How about using `Objects#requireNonNull()` instead of asserting? This doesn't seem to be a hot code.

modules/index/src/main/java/org/apache/ignite/internal/index/ChangeIndexStatusTask.java
tkalkirill: I made a mistake, I fixed it.

modules/index/src/main/java/org/apache/ignite/internal/index/ChangeIndexStatusTask.java
tkalkirill: Yes, this is not a hot code and it shouldn’t be, so let’s leave it as an assertion for now.

modules/index/src/main/java/org/apache/ignite/internal/index/ChangeIndexStatusTask.java
tkalkirill: I didn't fully understand you.
This method will return a fallen future with an error that it looks more correct, and processing this error, including logging, should not happen here. There is already processing in **org.apache.ignite.internal.index.ChangeIndexStatusTask#start**.

modules/index/src/main/java/org/apache/ignite/internal/index/ChangeIndexStatusTask.java
rpuch: This is not a hot code, so we can check for null properly at each invocation and not use assertion for this.

It's much better to see a `null` right away than deal with an NPE later on.

modules/index/src/main/java/org/apache/ignite/internal/index/ChangeIndexStatusTask.java
```diff
@@ -245,7 +245,8 @@ private CompletableFuture<ReplicaMeta> awaitPrimaryReplica() {
                             Throwable cause = unwrapCause(throwable);
 
                             if (cause instanceof PrimaryReplicaAwaitTimeoutException) {
-                                return awaitPrimaryReplica();
+                                return
+                                        awaitPrimaryReplica();
```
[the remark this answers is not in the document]
tkalkirill: Yep, thanks!

PR #4075, opened 2024-07-11
IGNITE-22353 Basic Python DB API Driver

isapego:
https://issues.apache.org/jira/browse/IGNITE-22353

Implemented:
- Basic project layout;
- Tests stub;
- Integration with C++ project;
- Connection to a cluster (re-used ODBC code);


modules/platforms/python/cpp_module/py_connection.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+    Py_TYPE(self)->tp_free(self);
+}
+
+static PyObject* py_connection_close(py_connection* self, PyObject*)
+{
+    if (self->m_conn) {
+        self->m_conn->release();
+        if (!check_errors(*self->m_conn))
+            return nullptr;
```
arcolight, line 55: Should we delete `self->m_env` in this case?

arcolight (approved):
LGTM

modules/platforms/python/cpp_module/py_connection.cpp
isapego: No, in this case neither `self->m_conn`, nor `self->m_env` are deleted until the connection is released.

ptupitsyn (approved):

PR #4090, opened 2024-07-15
IGNITE-22691 Delete startReplica(ReplicationGroupId, PendingComparableValuesTracker, CompletableFuture)

JAkutenshi:
JIRA Ticket: [IGNITE-22691 | Delete `startReplica(ReplicationGroupId, PendingComparableValuesTracker, CompletableFuture)`](https://issues.apache.org/jira/browse/IGNITE-22691)

## The goal

The goal is to replace all usages of `startReplica(ReplicationGroupId, PendingComparableValuesTracker, CompletableFuture)` in favor `startReplica` that is called in `TableManager`.

## The reason

There should the only one public `Replica#startReplica` method for replication group creation.

## The solution

1. General `TableManager`'s `startReplica` is used everywhere instead.
2. In `ItRebalanceDistributedTest#verifyThatRaftNodesAndReplicasWereStartedOnlyOnce` arguments list was increased according to used `startReplica`.
3. In case `ItTxTestCluster` there were almost all components but `SnapshotStorageFactory`. We couldn't add honest implementation of the interface because there metastore manager dependency is emerged. Then we just leave there a factory that produces "no-op" `SnapshotStorage` anonymous implementations.
4. Also there were leaved leaders' raft-clients collection in `#startTable` method, but it will be removed in [IGNITE-22774](https://issues.apache.org/jira/browse/IGNITE-22774).

---

### The Review Checklist

### Notes


modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -762,16 +762,13 @@ public CompletableFuture<Replica> startReplica(
...
      *
      * @param replicaGrpId Replication group id.
      * @param storageIndexTracker Storage index tracker.
      * @param newReplicaListenerFut Future that returns ready ReplicaListener for replica creation.
      * @return Future that promises ready new replica when done.
      */
-    @VisibleForTesting
-    @Deprecated
-    public CompletableFuture<Replica> startReplica(
+    private CompletableFuture<Replica> startReplica(
```
sanpwc, line 771: It's still @Deprecated @VisibleForTesting and should be removed in https://issues.apache.org/jira/browse/IGNITE-22373. The only change for now is public -> private.

modules/table/src/testFixtures/java/org/apache/ignite/distributed/ItTxTestCluster.java
```diff
@@ -673,112 +696,72 @@ public TableViewInternal startTable(String tableName, SchemaDescriptor schemaDes
...
+                                RaftGroupEventsListener.noopLsnr,
+                                partitionListener,
+                                false,
+                                null,
+                                createReplicaListener,
+                                storageIndexTracker,
+                                grpId,
+                                configuration
+                        )
+                        .thenAccept(unused -> { });
```
sanpwc, line 735: I'd rather make partitionReadyFutures `List<CompletableFuture<?>>`

modules/table/src/testFixtures/java/org/apache/ignite/distributed/ItTxTestCluster.java
```diff
@@ -673,112 +696,72 @@ public TableViewInternal startTable(String tableName, SchemaDescriptor schemaDes
...
             }
 
-            PeersAndLearners membersConf = PeersAndLearners.fromConsistentIds(partAssignments);
-
-            if (startClient) {
-                RaftGroupService service = RaftGroupServiceImpl
-                        .start(grpId, client, FACTORY, raftConfig, membersConf, true, executor, commandsMarshaller)
-                        .get(5, TimeUnit.SECONDS);
+            // waiting for started replicas otherwise we would have NPE on {@link Replica#replica} call below
+            allOf(partitionReadyFutures.toArray(new CompletableFuture[0])).join();
```
sanpwc, line 741: It's strange to have same line
`allOf(partitionReadyFutures.toArray(new CompletableFuture[0])).join();` twice. 741 and 756

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
JAkutenshi: Yes, it will be removed an the current state of 22373's patch [reflects it](https://github.com/apache/ignite-3/pull/4144/files#diff-69e9a603ac4198f678f9d8168888bfa70292521abcc8a22373e3114062e4fc2cL734-L820).

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
```diff
@@ -295,7 +295,7 @@ public ReplicaManager(
             Executor requestsExecutor,
             LongSupplier idleSafeTimePropagationPeriodMsSupplier,
             FailureProcessor failureProcessor,
-            Marshaller raftCommandsMarshaller,
+            @Nullable Marshaller raftCommandsMarshaller,
```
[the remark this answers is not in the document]
JAkutenshi: Previously there were empty snapshot storage implementation, but later I found that there are places where `RaftGroupOptions#defaults` is used and all discussed fields are handled with null-checks like in `org.apache.ignite.internal.raft.server.RaftGroupOptions#snapshotStorageFactory()` lines 480-482 for marshaller and 493-495 for storage factory. There is a point why I had been decided to use already existed mechanisms.

On the other hand in tests there several (mostly dependencies) issues with especially the factory and marshaller, on of the most known is [IGNITE-22222 with thread local marshaller in a test](https://issues.apache.org/jira/browse/IGNITE-22222). With the current solution we're reusing internal raft code and rely on it and get rid of dependency issues in tests.

>only commandsMarshaller is marked as https://github.com/nullable, but snapshotStorageFactory doesn't.

About that, we can look inside `RaftGroupOptions` and see that fields inconsistently marked as `@Nullable`, but they aren't initializing in the constructor and the only way to set the value is setters that may not be called. The most fair way is to mark the snapshots factory as nullable too and remove null checks in `ReplicaManager`, relying totally on raft internals.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
sanpwc: I mean that formally speaking this method is still @Deprecared and @VisibleForTesting, thus it's incorrect to remove given annotations.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/ReplicaManager.java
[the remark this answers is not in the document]
sanpwc: All I'm talking about is the fact that if you mark both fields as @Nullable in ReplicaManager constructor you should also mark the fields both in ReplicaManager and RaftGroupOptions.

sanpwc (approved):

PR #4100, opened 2024-07-17
IGNITE-22646 Support marshallers in the Java Client for Compute API

PakhomovAlexander:
https://issues.apache.org/jira/browse/IGNITE-22646

Now a user can define marshalers for input and for result
types for the Compute job. It might be java serialization,
or Jackson, or anything. 

All typos with marhal(l)ing spelling are fixed.


ptupitsyn (changes requested):

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/client/ItEmbeddedClientMarshallingTest.java
```diff
@@ -0,0 +1,139 @@
...
+import org.apache.ignite.internal.runner.app.client.Jobs.PojoArg;
+import org.apache.ignite.internal.runner.app.client.Jobs.PojoJob;
+import org.apache.ignite.internal.runner.app.client.Jobs.PojoResult;
+import org.apache.ignite.internal.runner.app.client.Jobs.ResultStringUnMarshaller;
+import org.apache.ignite.network.ClusterNode;
+import org.apache.ignite.table.Tuple;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test for embedded client marshalers for Compute API.
```
ptupitsyn, line 45: Not sure about "embedded client", what does it mean? We have "embedded API" and "client API", right?

modules/compute/src/main/java/org/apache/ignite/internal/compute/executor/ComputeExecutorImpl.java
```diff
@@ -86,27 +90,67 @@ public <T, R> JobExecutionInternal<R> executeJob(
...
             return (T) input;
         }
 
         if (input instanceof byte[]) {
-            return marshaller.unmarshal((byte[]) input);
+            try {
+                return marshaller.unmarshal((byte[]) input);
+            } catch (Exception ex) {
+                throw new ComputeException(Compute.TYPE_CHECK_MARSHALLING_ERR,
+                        "Job argument can not be unmarshalled. Both client-side and server-side marshaller should be compatible.",
```
ptupitsyn, line 140:
```suggestion
                        "Exception in user-defined marshaller: " + ex.getMessage(),
```

modules/compute/src/main/java/org/apache/ignite/internal/compute/executor/ComputeExecutorImpl.java
```diff
@@ -86,27 +90,67 @@ public <T, R> JobExecutionInternal<R> executeJob(
...
+                        ex
+                );
+            }
         }
 
-        return (T) input;
+        throw new ComputeException(
+                Compute.TYPE_CHECK_MARSHALLING_ERR,
+                "Argument must be a `byte[]` because marshaller is defined. "
```
ptupitsyn, line 148:
```suggestion
                "Marshaller is defined, expected argument type: `byte[]`, actual: `...`."
```

modules/api/src/main/java/org/apache/ignite/lang/ErrorGroups.java
```diff
@@ -577,6 +577,13 @@ public static class Compute {
 
         /** Specified node is not found in the cluster. */
         public static final int NODE_NOT_FOUND_ERR = COMPUTE_ERR_GROUP.registerErrorCode((short) 14);
+
+        /**
+         * Incompatible types for argument/result in compute job.
+         * For example, the one has defined a marshaller for Type A in the compute job
+         * but on the client side they have passed Type B.
+         */
+        public static final int TYPE_CHECK_MARSHALLING_ERR = COMPUTE_ERR_GROUP.registerErrorCode((short) 15);
```
ptupitsyn, line 586:
```suggestion
        public static final int MARSHALLING_TYPE_MISMATCH_ERR = COMPUTE_ERR_GROUP.registerErrorCode((short) 15);
```

(subject comes first in the error name)

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/client/Jobs.java
```diff
@@ -0,0 +1,291 @@
...
+import org.apache.ignite.compute.task.MapReduceJob;
+import org.apache.ignite.compute.task.MapReduceTask;
+import org.apache.ignite.compute.task.TaskExecutionContext;
+import org.apache.ignite.marshalling.ByteArrayMarshaller;
+import org.apache.ignite.marshalling.Marshaller;
+import org.apache.ignite.network.ClusterNode;
+import org.jetbrains.annotations.Nullable;
+
+class Jobs {
+
```
ptupitsyn, line 42:
```suggestion
```

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/client/ItEmbeddedMarshallingTest.java
```diff
@@ -0,0 +1,139 @@
...
+import org.apache.ignite.internal.runner.app.client.Jobs.ResultStringUnMarshaller;
+import org.apache.ignite.network.ClusterNode;
+import org.apache.ignite.table.Tuple;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test for embedded API marshallers for Compute API.
+ */
+@SuppressWarnings("resource")
+public class ItEmbeddedMarshallingTest extends ItAbstractThinClientTest {
```
ptupitsyn, line 48: We already have `ItThinClientMarshallingEmbeddedTest` (derived from `ItThinClientMarshallingTest`), do we need this class? I'd prefer to follow the pattern with a pair of client/embedded classes that test the same thing and ensure consistent behavior across both API implementations.

PakhomovAlexander: I understand what you mean but I would not like to follow this pattern in Compute API. I think that the inheritance for tests in this case might introduce pitfalls in the future. I can see us having different contracts for thin/embedded clients when it comes to Compute. For example, deployment unit might not be used at all in the embedded mode. 

If you want me to extend some test cases and double-check the contracts I can do that.

ptupitsyn: Then I don't understand why this test class is in the `client` package?

modules/platforms/cpp/ignite/odbc/common_types.cpp
```diff
@@ -312,9 +312,10 @@ sql_state error_code_to_sql_state(error::code code) {
         case error::code::NODE_START:
             return sql_state::SHY000_GENERAL_ERROR;
 
-        // Marshalling group. Group code: 22
+        // Marshalling group. Group code: асV 22
```
ptupitsyn, line 315: Where does `асV` come from? Is it a bug in the code generator?

ptupitsyn:
Looks good to me in general, still not sure about that `ItEmbeddedMarshallingTest.java` placement in `client` package.

modules/platforms/cpp/ignite/odbc/common_types.cpp
PakhomovAlexander: Good catch. `acV` comes from vim, its a typo.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/runner/app/client/ItEmbeddedMarshallingTest.java
PakhomovAlexander: makes sense, will move this test into `compute` package.

PR #4111, opened 2024-07-18
IGNITE-22759 Do not do partition SafeTime sync if previous attempt is not finished

rpuch:
https://issues.apache.org/jira/browse/IGNITE-22759


rpuch:
Superceded by #4684

PR #4113, opened 2024-07-19
IGNITE-22761 Disable WAL in RocksDB used to spill volatile Raft logs on disk

rpuch:
https://issues.apache.org/jira/browse/IGNITE-22761

### DIFF
--- a/modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/impl/RocksDbSpillout.java
+++ b/modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/impl/RocksDbSpillout.java
@@ -144,6 +144,7 @@ public class RocksDbSpillout implements Logs {
         this.groupEndBound = new Slice(groupEndPrefix);
 
         this.writeOptions = new WriteOptions();
+        this.writeOptions.setDisableWAL(true);
         this.writeOptions.setSync(false);
     }
 


modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/impl/RocksDbSpillout.java
```diff
@@ -144,6 +144,7 @@ public RocksDbSpillout(
         this.groupEndBound = new Slice(groupEndPrefix);
 
         this.writeOptions = new WriteOptions();
+        this.writeOptions.setDisableWAL(true);
```
tkalkirill, line 147: How will we maintain consistency in saving commands?

rpuch: It is related to volatile storage. When starting a node, we recreate this RocksDB effectively destroying its state.

tkalkirill: It was simply not obvious to me from the name of the class that it was for volatile tables and it was said in the documentation at the very end.
Maybe I would rename the class to something like `RocksDbLogsNotSpillout`. 
Or at least indicate in the title of the problem that it is about volatile raft groups.

sashapolo (approved):

modules/raft/src/main/java/org/apache/ignite/raft/jraft/storage/impl/RocksDbSpillout.java
rpuch: Included 'volatile' in the title of the Jira issue and PR

tkalkirill (approved):

PR #4119, opened 2024-07-22
IGNITE-22771 Create new System configuration root

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-22771


modules/core/src/main/java/org/apache/ignite/internal/util/LazyPath.java
```diff
@@ -0,0 +1,145 @@
...
+            return ensureCreated ? ensureCreated(resolved) : resolved;
+        });
+    }
+
+    @Override
+    public Path get() {
+        return super.get();
+    }
+
+    private static Path ensureCreated(Path dir) {
```
alievmirza, line 123: Why do we have two the same methods here and in `org.apache.ignite.internal.configuration.IgnitePaths#ensureCreated`

modules/core/src/main/java/org/apache/ignite/internal/util/LazyPath.java
```diff
@@ -0,0 +1,145 @@
...
+     * Resolve the other path against this one.
+     */
+    public LazyPath resolveLazy(Path other) {
+        return resolveLazy(other, false);
+    }
+
+    /**
+     * Resolve the other path against this one.
+     */
+    public LazyPath resolveLazy(String other) {
```
alievmirza, line 92: these two methods are just the same

modules/core/src/main/java/org/apache/ignite/internal/util/LazyPath.java
```diff
@@ -0,0 +1,145 @@
...
+        return new LazyPath(() -> pathOrDefault(pathSupplier.get(), defaultPath));
+    }
+
+    /**
+     * Create a new instance from {@code pathSupplier} if it returns a nonempty value, otherwise use {@code defaultPathSupplier}.
+     *
+     * @param pathSupplier Path supplier.
+     * @param defaultPathSupplier Default path supplier.
+     */
+    public static LazyPath create(Supplier<String> pathSupplier, Supplier<Path> defaultPathSupplier) {
```
alievmirza, line 62: this method is not used

modules/core/src/main/java/org/apache/ignite/internal/util/LazyPath.java
```diff
@@ -0,0 +1,145 @@
...
+        return new LazyPath(() -> defaultPath);
+    }
+
+    /**
+     * Create a new instance from {@code pathSupplier} if it returns a nonempty value, otherwise use {@code defaultPath}.
+     *
+     * @param pathSupplier Path supplier.
+     * @param defaultPath Default path.
+     */
+    public static LazyPath create(Supplier<String> pathSupplier, Path defaultPath) {
```
alievmirza, line 52: this method is not used

modules/core/src/main/java/org/apache/ignite/internal/util/LazyPath.java
alievmirza, line 123: I don't understand scenarios when we nedd ensured or not ensured paths

modules/system-configuration/src/main/java/org/apache/ignite/internal/configuration/SystemConfigurationSchema.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.apache.ignite.configuration.annotation.ConfigurationRoot;
+import org.apache.ignite.configuration.annotation.ConfigurationType;
+import org.apache.ignite.configuration.annotation.Value;
+
+/**
+ * System Configuration schema.
+ */
+@ConfigurationRoot(rootName = "system", type = ConfigurationType.LOCAL)
+public class SystemConfigurationSchema {
```
alievmirza, line 28: I'm not sure that this is a good naming. I would name it like 'PathsConfigurationSchema`. As I understood, this class will have other paths, not only `partitionsBasePath`. The same for root, I would name it `paths`

modules/core/src/main/java/org/apache/ignite/internal/util/LazyPath.java
```diff
@@ -0,0 +1,145 @@
...
+    @Override
+    public Path get() {
+        return super.get();
+    }
+
+    private static Path ensureCreated(Path dir) {
+        try {
+            Files.createDirectories(dir);
+        } catch (IOException e) {
+            throw new IgniteInternalException("Failed to create directory: " + dir + ": " + e.getMessage(), e);
```
alievmirza, line 127: deprecated constructor usage

alievmirza (changes requested):

modules/system-configuration/src/main/java/org/apache/ignite/internal/configuration/SystemConfigurationSchema.java
sashapolo: I agree with Mirza on this, "system" sounds too vague in this context

modules/system-configuration/src/main/java/org/apache/ignite/internal/configuration/SystemConfigurationSchema.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Directory where partition data is stored. By default "partitions" subfolder of data storage path is used.
+     *
+     * <pre>
+     * The internal directory structure contains:
```
sashapolo, line 34: I've heard that we wanted to be able to have raft log and partition storages on separate devices, for example. How will this configuration allow that?

sashapolo: This ticket is about decoupling MS and CMG Raft logs from partitions. I'm talking about separating partitions raft logs and partitions data

sashapolo: The current javadoc is misleading now. We have `partitionsBasePath` that is responsible for the log, meta and db files, while we also have `partitionsLogPath` which is also responsible for the log files? Why do we have this overlapping? Can we have three properties each responsible for its own folder?

modules/system-configuration/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+    annotationProcessor project(':ignite-configuration-annotation-processor')
+    annotationProcessor libs.auto.service
+
+    implementation project(':ignite-core')
+    implementation project(':ignite-configuration-api')
+    implementation libs.jetbrains.annotations
+    implementation libs.auto.service.annotations
+}
+
+description = 'ignite-system-configuration'
```
ptupitsyn, line 33: Do we really need a separate module just for configuration?

ptupitsyn: Seems weird to have a module just for that, can we put it into an existing module maybe?

ptupitsyn: Can it be `ignite-runner` or `ignite-configuration`?

modules/partition-replicator/build.gradle
```diff
@@ -74,6 +74,7 @@ dependencies {
     integrationTestImplementation project(':ignite-page-memory')
     integrationTestImplementation project(':ignite-placement-driver')
     integrationTestImplementation project(':ignite-affinity')
+    integrationTestImplementation project(':ignite-runner')
```
ascherbakoff, line 77: What is the purpose of this dependency ?

sashapolo (approved):

ascherbakoff (approved):

ptupitsyn (approved):

PR #4129, opened 2024-07-24
IGNITE-22801 Add term param to changePeers

sanpwc:
https://issues.apache.org/jira/browse/IGNITE-22801

org.apache.ignite.internal.raft.service.RaftGroupService#changePeers adjustments

- peers -> peersAndLearns
- term parameter added in order skip configuration change proposal if current term doesn't match the proposed one.


modules/raft/src/integrationTest/java/org/apache/ignite/raft/jraft/core/ItNodeTest.java
```diff
@@ -3200,28 +3202,52 @@ public void testChangePeersAddMultiNodes() throws Exception {
...
         cluster.ensureSame();
         assertEquals(3, cluster.getFsms().size());
+
         for (MockStateMachine fsm : cluster.getFsms())
             assertEquals(10, fsm.getLogs().size());
+
+        for (NodeImpl node: cluster.getNodes())
+            assertEquals(node.getConf().getConf().getPeers().size(), 3);
+
+        // another attempt to change peers with unmatched term
```
sashapolo, line 3236: Shall we have a separate test for the new logic? It's strange that we check everything in one place

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/CliRequests.java
```diff
@@ -69,6 +69,11 @@ public interface ChangePeersRequest extends Message {
         String leaderId();
 
         Collection<String> newPeersList();
+
+        Collection<String> newLearnersList();
+
+        // term is intentionally Long and not long in order to perform nullable (not initialized) check.
+        Long term();
```
sashapolo, line 76: Why isn't this field marked as `@Nullable`?

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/CliRequests.java
```diff
@@ -89,7 +99,8 @@ public interface ChangePeersAsyncRequest extends Message {
 
         Collection<String> newLearnersList();
 
-        long term();
+        // term is intentionally Long and not long in order to perform nullable (not initialized) check.
+        Long term();
```
sashapolo, line 103: Same question

modules/raft/src/test/java/org/apache/ignite/raft/jraft/rpc/impl/cli/ChangePeersRequestProcessorTest.java
```diff
@@ -32,13 +32,15 @@
 import static org.mockito.ArgumentMatchers.eq;
 
 public class ChangePeersRequestProcessorTest extends AbstractCliRequestProcessorTest<ChangePeersRequest> {
-
+    private static final long CURRENT_TERM = 1L;
     @Override
```
sashapolo, line 36: missing empty line

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/CliRequests.java
sanpwc: Because it's mandatory.

modules/raft/src/integrationTest/java/org/apache/ignite/raft/jraft/core/ItNodeTest.java
sanpwc: I don't think that it's really needed. ItNodeTest are long-running, thus I've extend already existing one with few more checks just in order not to startup cluster one more time. Besides that newly added logic is quite simple.

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/CliRequests.java
sanpwc: Same answer. Term is mandatory.

modules/raft/src/test/java/org/apache/ignite/raft/jraft/rpc/impl/cli/ChangePeersRequestProcessorTest.java
sanpwc: Fixed.

rpuch (approved):

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/CliRequests.java
sashapolo: ok, I didn't understand the original intention, sorry

sashapolo (approved):

modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
```diff
@@ -314,21 +314,29 @@ public CompletableFuture<Void> removePeer(Peer peer) {
     }
 
     @Override
-    public CompletableFuture<Void> changePeers(Collection<Peer> peers) {
+    public CompletableFuture<Void> changePeers(PeersAndLearners peersAndLearners, long term) {
```
kgusakov, line 317: So, changePeers now change peers and learners actually, it's confusing a little bit

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/CliRequests.java
```diff
@@ -89,7 +99,8 @@ public interface ChangePeersAsyncRequest extends Message {
 
         Collection<String> newLearnersList();
 
-        long term();
+        // term is intentionally Long and not long in order to perform nullable (not initialized) check.
```
kgusakov, line 102: I'm sorry, didn't understand - it is not mandatory (not @Nullable), but can be null at the same time?

modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
sanpwc: Yep, just like changePeersAsync that also changes both peers and learners. Do you believe that I should rename both?

modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/CliRequests.java
sanpwc: It's mandatory because it's not marked as @Nullable. If someone will forget to specify term() while building the request AssertionError will be thrown, something like "request field not marked with @nullable has null value".

modules/raft/src/main/java/org/apache/ignite/raft/jraft/Node.java
```diff
@@ -178,18 +178,19 @@ public interface Node extends Lifecycle<NodeOptions>, Describer {
      * Change the configuration of the raft group to |newPeers| , done.run() would be invoked after this operation
      * finishes, describing the detailed result.
      *
-     * @param newPeers new peers to change
+     * @param newPeersAndLearners new peers and learners to change
+     * @param term term on which this method was called. If real raft group term will be different - changePeers will be skipped.
      * @param done callback
      */
-    void changePeers(final Configuration newPeers, final Closure done);
+    void changePeers(final Configuration newPeersAndLearners, long term, final Closure done);
```
alievmirza, line 185: Params differ from `void changePeersAsync(final Configuration newConf,`. I would expect the same naming for both two methods.

modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
kgusakov: After discussion - changed to changePeersAndLearners, thanks!

kgusakov (approved):

modules/raft/src/main/java/org/apache/ignite/raft/jraft/Node.java
sanpwc: Fixed.

PR #4130, opened 2024-07-24
IGNITE-22695 .NET: Add host name support to ClusterNode.Address

ptupitsyn:
Previously .NET client assumed that node address is always an IP. Fix the API and parsing to support host names as well.


modules/platforms/dotnet/Apache.Ignite.Tests/FakeServer.cs
```diff
@@ -76,12 +78,15 @@ public FakeServer(bool disableOpsTracking, string nodeName = "fake-server")
...
             string nodeName = "fake-server",
-            bool disableOpsTracking = false)
+            bool disableOpsTracking = false,
+            int port = 0)
+            : base(port)
         {
             _shouldDropConnection = shouldDropConnection ?? (_ => false);
 
-            Node = new ClusterNode("id-" + nodeName, nodeName, (IPEndPoint)Listener.LocalEndPoint!);
+            Node = new ClusterNode("id-" + nodeName, nodeName, IPEndPoint.Parse("127.0.0.1:" + Port));
```
isapego, line 87: What's wrong with `Listener.LocalEndPoint`?

isapego (approved):

modules/platforms/dotnet/Apache.Ignite.Tests/FakeServer.cs
ptupitsyn: Listener listens on `0.0.0.0` (all interfaces), but client connects to `127.0.0.1`, this breaks assumptions in some tests. Easier to fix here once than to fix multiple tests.

PR #4137, opened 2024-07-25
IGNITE-22641 Introduce UpdateMinimumActiveTxStartTime RAFT command

xtern:
https://issues.apache.org/jira/browse/IGNITE-22641

Added routine to collect and store in all replicas minimum tx starting time among all started transactions in the cluster

* Added `ActiveTxMinimumBeginTimeProvider` to obtain minimum starting time among all local transactions
* Added `UpdateMinimumActiveTxBeginTimeReplicaRequest` (into `PartitionReplicationMessageGroup`, otherwise it will require the `ReplicaManager` to listen for `CatalogCompactionMessageGroup` messages)
* Added corresponding `UpdateMinimumActiveTxBeginTimeCommand` (in `PartitionReplicationMessageGroup`) to store this timestamp into replication groups.

Overall, the routine (`CatalogCompactionRunner`) currently looks like this:

```
trigger compaction
   |
   |   +--------------------------------------------+
   |-> | 1. collect [minimumRequiredTime] and       |
       |   [minimumActiveTxTime] from remote nodes  |
       | 2. compute cluster wide minimums           |
       +--------------------------------------------+
               /                 \
              /                   \
             /                     \
     +-----------------+     +--------------------------+
     | preform catalog |     | update all raft groups   |
     |   compaction    |     | with minimumActiveTxTime |
     +-----------------+     +--------------------------+
```


modules/catalog-compaction/src/main/java/org/apache/ignite/internal/catalog/compaction/CatalogCompactionRunner.java
```diff
@@ -198,115 +225,127 @@ public CompletableFuture<Boolean> onLowWatermarkChanged(HybridTimestamp newLowWa
...
-        return inBusyLock(busyLock, () -> {
-            CompletableFuture<Boolean> fut = lastRunFuture;
+        inBusyLock(busyLock, () -> {
+            synchronized (this) {
+                CompletableFuture<Void> fut = lastRunFuture;
 
-            if (!fut.isDone()) {
-                LOG.info("Catalog compaction is already in progress, skipping (timestamp={})", lwm.longValue());
+                if (!fut.isDone()) {
+                    LOG.info("Catalog compaction is already in progress, skipping (timestamp={})", lwm.longValue());
```
korlov42, line 243: looks like `return` statement is missing

modules/catalog-compaction/src/main/java/org/apache/ignite/internal/catalog/compaction/CatalogCompactionRunner.java
```diff
@@ -198,115 +225,127 @@ public CompletableFuture<Boolean> onLowWatermarkChanged(HybridTimestamp newLowWa
...
+                                }
+                            }
+                        });
+                    }
 
-                                if (!missingNodes.isEmpty()) {
-                                    if (LOG.isDebugEnabled()) {
-                                        LOG.debug("Catalog compaction aborted due to missing cluster members (nodes={})", missingNodes);
-                                    }
+                    if (minActiveTxStartTime == null) {
```
korlov42, line 290: I think, we always need to propagate some time. Given that initially partition time is set to min possible clock value, we still must be able to do compaction, even if there is no transactional load. The safest possible option here is, if current node doesn't have active transactions, send current value of low watermark acquired just _before_ collecting time of active transactions

modules/catalog-compaction/src/main/java/org/apache/ignite/internal/catalog/compaction/CatalogCompactionRunner.java
```diff
@@ -376,10 +414,101 @@ private static List<String> missingNodes(Set<String> requiredNodes, Collection<L
...
+
+    private CompletableFuture<Boolean> tryCompactCatalog(Catalog catalog, LogicalTopologySnapshot topologySnapshot) {
+        return requiredNodes(catalog)
+                .thenCompose(requiredNodes -> {
+                    List<String> missingNodes = missingNodes(requiredNodes, topologySnapshot.nodes());
+
+                    if (!missingNodes.isEmpty()) {
+                        if (LOG.isDebugEnabled()) {
+                            LOG.debug("Catalog compaction aborted due to missing cluster members (nodes={})", missingNodes);
+                        }
```
korlov42, line 484: first, you don't need to wrap such methods in `if` statement. Second, compaction seems to be really infrequent process, does it make sense to log main stages of compaction with INFO severity?

modules/catalog-compaction/src/main/java/org/apache/ignite/internal/catalog/compaction/CatalogManagerCompactionHelper.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.ignite.internal.catalog.CatalogManagerImpl;
+import org.apache.ignite.internal.catalog.descriptors.CatalogTableDescriptor;
+import org.apache.ignite.internal.catalog.descriptors.CatalogZoneDescriptor;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Class contains utility methods for interacting with the catalog manager.
+ * These methods are only needed for catalog compaction routine.
+ */
+class CatalogManagerCompactionHelper {
```
korlov42, line 33: nitpicking: it's, probably, better to use `Facade` suffix rather than `Helper`. Helper is usually utility class/set of utility methods

modules/catalog-compaction/src/main/java/org/apache/ignite/internal/catalog/compaction/CatalogCompactionRunner.java
xtern: Thanks, fixed.
Added test case for this .

modules/catalog-compaction/src/main/java/org/apache/ignite/internal/catalog/compaction/CatalogCompactionRunner.java
xtern: Done

modules/catalog-compaction/src/main/java/org/apache/ignite/internal/catalog/compaction/CatalogManagerCompactionHelper.java
xtern: Done, thanks

modules/catalog/src/main/java/org/apache/ignite/internal/catalog/CatalogManagerImpl.java
```diff
@@ -342,17 +342,26 @@ public CompletableFuture<Integer> execute(List<CatalogCommand> commands) {
...
     }
 
+    /**
+     * Trim all catalog versions up to the given catalog (exclusively).
+     *
+     * @param catalog Earliest observable catalog.
+     * @return Operation future, which is completing with {@code true} if a new snapshot has been successfully written, {@code false}
+     *         otherwise if a snapshot with the same or greater version already exists.
+     */
+    public CompletableFuture<Boolean> compactCatalog(Catalog catalog) {
```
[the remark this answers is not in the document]
xtern: Done, replaced two methods `compactCatlog(long)` and `compactCatalog(Catalog)` with single `compactCatalog(int)`.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/ActiveTxMinimumStartTimeProvider.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.tx;
+
+import org.apache.ignite.internal.hlc.HybridTimestamp;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Provides the minimum starting time among all locally started active RW transactions.
```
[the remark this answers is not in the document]
xtern: Done, thanks for the suggestion
(also renamed this class similarly)

korlov42 (approved):

modules/catalog-compaction/src/main/java/org/apache/ignite/internal/catalog/compaction/CatalogCompactionRunner.java
```diff
@@ -376,10 +406,110 @@ private static List<String> missingNodes(Set<String> requiredNodes, Collection<L
         return requiredNodes.stream().filter(not(logicalNodeIds::contains)).collect(Collectors.toList());
     }
 
+    CompletableFuture<Void> propagateTimeToReplicas(long minimumRequiredTime) {
+        Map<Integer, Integer> tablesWithPartitions = catalogManagerFacade.collectTablesWithPartitionsBetween(
```
sanpwc, line 410: Could you please elaborate what partition between (minimumRequiredTime, now()) means?

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -4136,6 +4139,23 @@ private CompletableFuture<?> processVacuumTxStateReplicaRequest(VacuumTxStateRep
...
     }
 
+    private CompletableFuture<?> processMinimumActiveTxTimeReplicaRequest(UpdateMinimumActiveTxBeginTimeReplicaRequest request) {
+        Command cmd = PARTITION_REPLICATION_MESSAGES_FACTORY.updateMinimumActiveTxBeginTimeCommand()
+                .timestamp(request.timestamp())
+                .build();
+
+        CompletableFuture<Object> resultFuture = new CompletableFuture<>();
+
+        // The timestamp must increase monotonically, otherwise it will have to be
```
sanpwc, line 4149: In that case you will need to extend SafeTimePropagatingCommand instead of simple WriteCommand and specify safe time of course.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -4136,6 +4139,23 @@ private CompletableFuture<?> processVacuumTxStateReplicaRequest(VacuumTxStateRep
...
+        // stored on disk so that reordering does not occur after the node is restarted.
+        applyCmdWithRetryOnSafeTimeReorderException(
+                cmd,
+                resultFuture
+        );
+
+        return resultFuture.thenApply(res -> null);
```
sanpwc, line 4156: `.thenApply(res -> null);` seems useless here.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/ActiveTxMinimumBeginTimeProvider.java
```diff
@@ -0,0 +1,30 @@
...
+package org.apache.ignite.internal.tx;
+
+import org.apache.ignite.internal.hlc.HybridTimestamp;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Provides the minimum begin time among all active RW transactions started locally.
+ */
+@SuppressWarnings("InterfaceMayBeAnnotatedFunctional")
+public interface ActiveTxMinimumBeginTimeProvider {
```
sanpwc, line 27: I'd rather add Local in the class name, something like `ActiveLocalTxMinimumBeginTimeProvider`.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/ActiveTxMinimumBeginTimeProvider.java
```diff
@@ -0,0 +1,30 @@
...
+    /** Returns the minimum begin time among all active RW transactions started locally. */
```
sanpwc, line 28: ... or null if there are no pending transactions, right?

modules/transactions/src/main/java/org/apache/ignite/internal/tx/ActiveTxMinimumBeginTimeProvider.java
```diff
@@ -0,0 +1,30 @@
...
+    @Nullable HybridTimestamp minimumBeginTime();
```
sanpwc, line 29: BTW, curious what time are you going to use as global minimum in case of zero pending transactions cluster wide?

modules/catalog-compaction/src/integrationTest/java/org/apache/ignite/internal/catalog/compaction/ItCatalogCompactionTest.java
```diff
@@ -0,0 +1,226 @@
...
+
+            assertThat(timeHolder.minActiveTxBeginTime, greaterThanOrEqualTo(minTime));
+            assertThat(timeHolder.minActiveTxBeginTime, lessThanOrEqualTo(maxTime));
+        });
+
+        readonlyTx.rollback();
+    }
+
+    private static void ensureTimestampStoredInAllReplicas(HybridTimestamp expTime, int expTablesCount) throws InterruptedException {
+        Int2IntMap tablesWithPartitions = catalogManagerFacade().collectTablesWithPartitionsBetween(
```
sanpwc, line 189: Generally it's not a good idea to test something that uses method X with method X itself. I mean that 
`propagateTimeToReplicas` uses `collectTablesWithPartitionsBetween`.

modules/catalog-compaction/src/integrationTest/java/org/apache/ignite/internal/catalog/compaction/ItCatalogCompactionTest.java
```diff
@@ -0,0 +1,226 @@
...
+        {
+            HybridTimestamp expectedTime = HybridTimestamp.hybridTimestamp(requiredTime - 1);
+
+            CompletableFuture<Void> fut = ignite.catalogCompactionRunner()
+                    .propagateTimeToReplicas(expectedTime.longValue());
+
+            assertThat(fut, willCompleteSuccessfully());
+
+            ensureTimestampStoredInAllReplicas(HybridTimestamp.hybridTimestamp(requiredTime), 2);
+        }
```
sanpwc, line 127: 1. Infinitely dead assignment case is not covered. I mean the one when `Assignment assignment = tokenizedAssignments.nodes().iterator().next();` is not present in topology and never will be.
2. Table adjustments while invokeOnReplicas recursive calls is not covered.
3. Failover cases are not covered. E.g. temporary assignment death.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -127,6 +128,12 @@ public class PartitionListener implements RaftGroupListener, BeforeApplyHandler
 
     private final IndexMetaStorage indexMetaStorage;
 
+    /**
+     * Timestamp with minimum starting time among all pending RW transactions in the cluster.
```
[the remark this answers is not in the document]
xtern: I believe I fixed this.

modules/catalog-compaction/src/main/java/org/apache/ignite/internal/catalog/compaction/CatalogCompactionRunner.java
```diff
@@ -376,10 +406,110 @@ private static List<String> missingNodes(Set<String> requiredNodes, Collection<L
...
+        );
+
+        return invokeOnReplicas(
+                tablesWithPartitions.entrySet().iterator(),
+                minimumRequiredTime,
+                clockService.now()
+        );
+    }
+
+    private CompletableFuture<Void> invokeOnReplicas(
```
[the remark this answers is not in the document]
xtern: Thanks, I believe that these optimizations will be done in https://issues.apache.org/jira/browse/IGNITE-22951

modules/catalog-compaction/src/main/java/org/apache/ignite/internal/catalog/compaction/CatalogCompactionRunner.java
```diff
@@ -376,10 +406,110 @@ private static List<String> missingNodes(Set<String> requiredNodes, Collection<L
...
+                                replicationGroupId
+                        );
+
+                        UpdateMinimumActiveTxBeginTimeReplicaRequest msg = REPLICATION_MESSAGES_FACTORY
+                                .updateMinimumActiveTxBeginTimeReplicaRequest()
+                                .groupId(partIdMessage)
+                                .timestamp(txBeginTime)
+                                .build();
+
+                        return replicaService.invoke(assignment.consistentId(), msg);
```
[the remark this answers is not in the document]
xtern: Fixed, now sending message to node in logical topology (if local node exists in assignments - sending to local)

modules/catalog-compaction/src/main/java/org/apache/ignite/internal/catalog/compaction/CatalogCompactionRunner.java
```diff
@@ -376,10 +406,110 @@ private static List<String> missingNodes(Set<String> requiredNodes, Collection<L
...
+                    });
+
+            partitionFutures.add(fut);
+        }
+
+        return CompletableFutures.allOf(partitionFutures)
+                .thenCompose(ignore -> invokeOnReplicas(tabItr, txBeginTime, nowTs));
```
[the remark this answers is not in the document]
xtern: > How often are you going to trigger minBeginTime move?

I suppose that mostly when LWM changes (every 5 min?)

> What if compactionCoordinatorNodeName has changed? What will prevent further recursive calls?

Nothing. We decided that this is ok.

modules/catalog-compaction/src/main/java/org/apache/ignite/internal/catalog/compaction/CatalogCompactionRunner.java
```diff
@@ -376,10 +406,110 @@ private static List<String> missingNodes(Set<String> requiredNodes, Collection<L
...
+                                .timestamp(txBeginTime)
+                                .build();
+
+                        return replicaService.invoke(assignment.consistentId(), msg);
+                    });
+
+            partitionFutures.add(fut);
+        }
+
+        return CompletableFutures.allOf(partitionFutures)
```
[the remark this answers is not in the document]
xtern: > why it's safe to silently ignore all types of exceptions

We don't ignore exceptions here. Exception will be logged.

```
                    CompletableFuture<Void> propagateToReplicasFut = propagateTimeToReplicas(minActiveTxBeginTime)
                            .whenComplete((res, ex) -> {
                                if (ex != null) {
                                    LOG.warn("Failed to propagate minimum active tx begin time to replicas", ex);
                                }
                            });
```

> why recursive call is considered as a proper failover

We don't have any failover :thinking:  process will be aborted in case of any exception with exception logging

modules/catalog-compaction/src/main/java/org/apache/ignite/internal/catalog/compaction/CatalogCompactionRunner.java
```diff
@@ -376,10 +406,110 @@ private static List<String> missingNodes(Set<String> requiredNodes, Collection<L
...
+        for (int p = 0; p < parts; p++) {
+            TablePartitionId replicationGroupId = new TablePartitionId(tableWithPartsCount.getKey(), p);
+
+            CompletableFuture<Object> fut = placementDriver.getAssignments(replicationGroupId, nowTs)
+                    .thenCompose(tokenizedAssignments -> {
+                        if (tokenizedAssignments == null) {
+                            throwAssignmentsNotReadyException(replicationGroupId);
+                        }
+
+                        Assignment assignment = tokenizedAssignments.nodes().iterator().next();
```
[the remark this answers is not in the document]
xtern: This code has been changed a bit. But I believe that the main optimization will be done separately in https://issues.apache.org/jira/browse/IGNITE-22951

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
xtern: Fixed, thanks.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/ActiveTxMinimumBeginTimeProvider.java
xtern: Done.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/ActiveTxMinimumBeginTimeProvider.java
xtern: Fixed, thanks.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/ActiveTxMinimumBeginTimeProvider.java
xtern: we use timestamp obtained just BEFORE computing minimum tx time.

```java
    private long computeLocalMinTxBeginTime() {
        // To help compaction progress when there are no active RW transactions,
        // it is safe to return the current time as minimum tx begin time.
        HybridTimestamp beforeDetermineTxTime = clockService.now();

        return Objects.requireNonNullElse(
                localActiveRwTxMinBeginTimeProvider.minimumBeginTime(),
                beforeDetermineTxTime
        ).longValue();
    }
```

modules/catalog-compaction/src/integrationTest/java/org/apache/ignite/internal/catalog/compaction/ItCatalogCompactionTest.java
xtern: > 1. Infinitely dead assignment case is not covered. I mean the one when Assignment assignment = tokenizedAssignments.nodes().iterator().next(); is not present in topology and never will be.

I added some tests on mocks for check this (`CatalogCompactionRunnerSelfTest`):
* minTxTimePropagationSucceedWhenSomeAssignmentIsMissing
* minTxTimePropagation
* minTxTimePropagationAbortedIfNoAssignmentsPresentInTopology

> 2. Table adjustments while invokeOnReplicas recursive calls is not covered.

This test case needs more clarification, I don't quite understand what this test might look like. :thinking: 

> 3.     Failover cases are not covered. E.g. temporary assignment death.

Currently we don't have any failover :sunglasses:  We just aborting the process (see `minTxTimePropagationAbortedIfNoAssignmentsPresentInTopology`).

modules/catalog-compaction/src/integrationTest/java/org/apache/ignite/internal/catalog/compaction/ItCatalogCompactionTest.java
xtern: Thanks, removed this dependency

modules/catalog-compaction/src/main/java/org/apache/ignite/internal/catalog/compaction/CatalogCompactionRunner.java
xtern: Added javadoc for `collectTablesWithPartitionsBetween`

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/raft/PartitionListener.java
```diff
@@ -666,6 +683,16 @@ private void handleVacuumTxStatesCommand(VacuumTxStatesCommand cmd, long command
...
     }
 
+    private void handleUpdateMinimalActiveTxTimeCommand(UpdateMinimumActiveTxBeginTimeCommand cmd, long commandIndex, long commandTerm) {
+        // Skips the write command because the storage has already executed it.
+        if (commandIndex <= storage.lastAppliedIndex()) {
+            return;
+        }
+
+        long minActiveTxStartTime0 = minActiveTxStartTime;
+        minActiveTxStartTime = Math.max(cmd.timestamp(), minActiveTxStartTime0);
```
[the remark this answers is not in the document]
xtern: thanks, replaced this with assertion

modules/transactions/src/main/java/org/apache/ignite/internal/tx/ActiveTxMinimumBeginTimeProvider.java
sanpwc: >         // it is safe to return the current time as minimum tx begin time.
>         HybridTimestamp beforeDetermineTxTime = clockService.now();
I'm not sure that it's safe.
I mean that there should be some gap between starting the transaction (and thus assigning some timestamp as tx.startTime) and registering it in `org.apache.ignite.internal.tx.impl.TxManagerImpl#localRwTxCounter`. 
The atomicity is achieved by using `inUpdateRwTxCountLock`
```
        return localRwTxCounter.inUpdateRwTxCountLock(() -> {
            HybridTimestamp beginTs = clockService.now();

            localRwTxCounter.incrementRwTxCount(beginTs);

            return beginTs;
        });
```
And because you don't use the lock in computeLocalMinTxBeginTime the race is possible.

sanpwc (approved):

PR #4142, opened 2024-07-25
IGNITE-22643 Support Tuple with schema in Compute API

PakhomovAlexander:
To support Tuples in compute API I've refactored some parts:

- `ClientBinaryTupleUtils` does not accept `Marshaller` when append the object.
- `ClientComputeJobPacker` and `ClientComputeJobUnpacker` are introduced.
They are supposed to encapsulate all logic about object marshalling in Compute.

https://issues.apache.org/jira/browse/IGNITE-22643


modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
```diff
@@ -0,0 +1,307 @@
...
+public final class TupleWithSchemaMarshalling {
+
+    /**
+     * Marshal tuple is the following format:
+     *
+     * <p>marshalledTuple       := | size | binaryTupleWithSchema |
+     *
+     * <p>size                  := int32
+     *
+     * <p>binaryTupleWithSchema := | column1 | ... | columnN | value1 | ... | valueN |
```
ptupitsyn, line 45: I think we should not put schema and values into the same `BinaryTuple`.

```
binaryTupleWithSchema = BinaryTuple + separate schema (most likely another BinaryTuple).
```

This will be beneficial when we need to separate the value from the schema. 

**Example 1** - `table.put`:
* If the provided schema matches the table schema, we can insert the data directly without rebuilding the BinaryTuple
* If the schema is a part of the BinaryTuple, we have to rebuild it

**Example 2** - many items with the same schema (`table.putAll`)
* Schema can be specified once for many values
* Not possible when it is a part of the BinaryTuple

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ComputeJobType.java
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * The type of the object that can be passed/returned to/from the compute job. In can be a native type that is represented by
+ * {@link ColumnType} or a marshalled object/tuple.
+ */
+public class ComputeJobType {
+    private static final int OFFSET = 1000;
+
+    static final int MARSHALLED_TUPLE_ID = OFFSET + 1;
+    static final int MARSHALLED_OBJECT_ID = OFFSET + 2;
```
ptupitsyn, line 30: I suggest using negative IDs for those special cases:
* No need to guess a safe offset
* More compact MsgPack representation

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientComputeJobPacker.java
```diff
@@ -0,0 +1,203 @@
...
+        var type = getColumnTypeFrom(obj);
+        assert type != null;
+
+        packer.packInt(type.id());
+
+        var writer = writerForType(type);
+        writer.write(obj, packer);
+    }
+
+    private static Writer writerForType(ColumnType type) {
```
ptupitsyn, line 116: Allocating a `Writer` here seems unnecessary. We can refactor the method to perform the actual write within the `switch`:

```java
       switch (type) {
            case BOOLEAN:
                packer.packBoolean(((Boolean) obj));
                return;
 ...
```

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientComputeJobUnpacker.java
```diff
@@ -0,0 +1,141 @@
...
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import org.apache.ignite.internal.binarytuple.inlineschema.TupleWithSchemaMarshalling;
+import org.apache.ignite.marshalling.Marshaller;
+import org.apache.ignite.sql.ColumnType;
+import org.jetbrains.annotations.Nullable;
+
+/** Unpacks job arguments and results. */
+public class ClientComputeJobUnpacker {
+    private final ClientMessageUnpacker unpacker;
```
ptupitsyn, line 29: I think we can remove this field, make all methods static, take `ClientMessageUnpacker` as an argument.

Static methods are easier to reason about and we get rid of another allocation.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
```diff
@@ -65,64 +71,105 @@ public final class TupleWithSchemaMarshalling {
...
+        int size = ((raw[0] & 0xFF) << 24) |
+                ((raw[1] & 0xFF) << 16) |
+                ((raw[2] & 0xFF) << 8) |
+                (raw[3] & 0xFF);
+
+        // Read second int32.
+        int valueOffset = ((raw[4] & 0xFF) << 24) |
+                ((raw[5] & 0xFF) << 16) |
+                ((raw[6] & 0xFF) << 8) |
+                (raw[7] & 0xFF);
```
ptupitsyn, line 118: Use `ByteBuffer` methods - we need it below anyway.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
```diff
@@ -65,64 +71,105 @@ public final class TupleWithSchemaMarshalling {
...
+        byte[] valueBt = valueBuilder.build().array();
+        // Size: int32 (tuple size), int32 (value offset), schema, value.
+        byte[] result = new byte[4 + 4 + schemaBt.length + valueBt.length];
 
-        byte[] arr = builder.build().array();
-        byte[] result = new byte[arr.length + 1];
-        // Put the size of the schema in the first byte.
-        result[0] = (byte) size; // todo should be int32
+        // Put the size of the schema in the first 4 bytes.
+        result[0] = (byte) (size >> 24);
```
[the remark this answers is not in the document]
ptupitsyn: Alternatively, use `ByteBuffer`.

ptupitsyn (changes requested):

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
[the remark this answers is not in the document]
PakhomovAlexander: Thanks, I've decided to copy methods from `HeapByteBufUtil`. The reason why I have not reused this class directly is that it has package-private methods and requires changes. But I do not want to modify this class since it is forked.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
```diff
@@ -0,0 +1,333 @@
...
+            values[i] = value;
+            columns[i] = tup.columnName(i);
+            types[i] = inferType(value);
+        }
+
+        BinaryTupleBuilder schemaBuilder = schemaBuilder(columns, types);
+        BinaryTupleBuilder valueBuilder = valueBuilder(values, types);
+
+        byte[] schemaBt = schemaBuilder.build().array();
+        byte[] valueBt = valueBuilder.build().array();
```
ptupitsyn, line 72: `array()` calls are redundant - unnecessary array allocation and data copying.

We can copy directly from `ByteBuffer` (result of `BinaryTupleBuilder.build()`) into the resulting array.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
```diff
@@ -0,0 +1,333 @@
...
+        }
+    }
+
+    private static <T> T checkType(Class<T> cl, Object o) {
+        if (o.getClass().isAssignableFrom(cl)) {
+            return (T) o;
+        }
+
+        throw new IllegalArgumentException("Unsupported type: " + o.getClass());
+    }
```
ptupitsyn, line 211: See `ClientBinaryTupleUtils.appendValue` - we catch `ClassCastException` and rethrow with more details, this is less code and more efficient than calling `checkType` every time.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
PakhomovAlexander: to construct the `result` array we have to know the lengths of both schema and value arrays. the array here is very handy. 

and there should not be an allocation for `array()` call because the method returns the reference to the base `byte[]`.

ptupitsyn: There is no allocation - you are right.

However, **there is a bug**. The resulting array is almost always bigger than needed, we must use `ByteBuffer.arrayOffset()` and `limit()` to get the data.

modules/api/src/main/java/org/apache/ignite/marshalling/UnmarshallingException.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package org.apache.ignite.marshalling;
+
+import java.util.UUID;
+import org.apache.ignite.lang.ErrorGroups.Marshalling;
+import org.apache.ignite.lang.IgniteException;
+
+/**
+ * Exception thrown when an object type is not supported by the marshaller.
```
ptupitsyn, line 25: Misleading javadoc?

modules/api/src/main/java/org/apache/ignite/marshalling/UnsupportedObjectTypeMarshallingException.java
```diff
@@ -27,16 +27,21 @@
...
      * Creates an exception with the given unsupported type.
      *
      * @param unsupportedType Unsupported type.
      */
     public UnsupportedObjectTypeMarshallingException(Class<?> unsupportedType) {
-        super(
-                Marshalling.UNSUPPORTED_OBJECT_TYPE_ERR,
-                "Unsupported object type: " + unsupportedType.getName() + ". Please, define the marshaller that can handle this type."
-        );
+        this("Unsupported object type: " + unsupportedType.getName() + ". Please, define the marshaller that can handle this type.");
```
ptupitsyn, line 37:
```suggestion
        this("Unsupported object type: " + unsupportedType.getName() + ". Please, define a marshaller that can handle this type.");
```

modules/binary-tuple/src/libasyncProfiler.dylib
ptupitsyn, line 1: Added by accident? Please remove.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
```diff
@@ -0,0 +1,378 @@
...
+
+        // Put the size of the schema in the first 4 bytes.
+        buff.putInt(0, size);
+
+        // Put the value offset in the second 4 bytes.
+        int offset = schemaArr.length + 8;
+        buff.putInt(4, offset);
+
+        System.arraycopy(schemaArr, 0, result, 8, schemaArr.length);
+        System.arraycopy(valueArr, 0, result, schemaArr.length + 8, valueArr.length);
```
ptupitsyn, line 98: Remove `getByteArray` method, it performs unnecessary allocations. You can copy from one `ByteBuffer` to another directly.

```suggestion
        int schemaBufLen = schemaBf.remaining();
        int valueBufLen = valueBf.remaining();

        // Size: int32 (tuple size), int32 (value offset), schema, value.
        byte[] result = new byte[4 + 4 + schemaBufLen + valueBufLen];
        ByteBuffer buff = ByteBuffer.wrap(result).order(BYTE_ORDER);

        // Put the size of the schema in the first 4 bytes.
        buff.putInt(size);

        // Put the value offset in the second 4 bytes.
        int offset = schemaBufLen + 8;
        buff.putInt(offset);

        buff.put(schemaBf);
        buff.put(valueBf);
```

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
```diff
@@ -0,0 +1,378 @@
...
+            ColumnType type = types[i];
+            Object v = values[i];
+
+            appender(type, columnNames[i]).accept(builder, v);
+        }
+
+        return builder;
+    }
+
+    private static BiConsumer<BinaryTupleBuilder, Object> appender(ColumnType type, String name) {
```
ptupitsyn, line 204: This method returns a `BiConsumer` that is immediately invoked, which causes unnecessary allocation and indirection.

We can either inline this method, or make it `void` and perform the append action immediately.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
```diff
@@ -0,0 +1,378 @@
...
+
+            columns[i] = colName;
+            types[i] = ColumnType.getById(colTypeId);
+
+            i += 1;
+        }
+
+        int k = 0;
+        while (k < size) {
+            setColumnValue(valueReader, tup, columns[k], types[k].id(), k);
```
ptupitsyn, line 173: We convert `ColumnType` to int here, then convert back inside `setColumnValue`, can we pass it as is?

ptupitsyn (changes requested):

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
```diff
@@ -0,0 +1,378 @@
...
+        // Fill in the values, column names, and types.
+        for (int i = 0; i < size; i++) {
+            var value = tuple.value(i);
+            values[i] = value;
+            columns[i] = tuple.columnName(i);
+            types[i] = inferType(value);
+        }
+
+        ByteBuffer schemaBf = schemaBuilder(columns, types).build();
+        ByteBuffer valueBf = valueBuilder(columns, types, values).build();
```
ptupitsyn, line 82: We abbreviate `buffer` as `buf`, not `bf` or `buff`. Please rename here and below.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
```diff
@@ -0,0 +1,378 @@
...
+            i += 1;
+        }
+
+        int k = 0;
+        while (k < size) {
+            setColumnValue(valueReader, tup, columns[k], types[k].id(), k);
+            k += 1;
+        }
+
+        return tup;
```
ptupitsyn, line 177:
```suggestion
        Tuple tup = Tuple.create(size);
        
        for (int i = 0; i < size; i++) {
            String colName = schemaReader.stringValue(i * 2);
            int colTypeId = schemaReader.intValue(i * 2 + 1);

            setColumnValue(valueReader, tup, colName, colTypeId, i);
        }

        return tup;
```

Unnecessary 2 loops, unnecessary arrays. Please be careful.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    private static BinaryTupleBuilder schemaBuilder(String[] columns, ColumnType[] types) {
+        BinaryTupleBuilder builder = new BinaryTupleBuilder(columns.length * 2, columns.length * 2);
```
ptupitsyn, line 181:
```suggestion
        BinaryTupleBuilder builder = new BinaryTupleBuilder(columns.length * 2);
```

Second parameter is `totalValueSize`, in bytes. We can omit it, the builder has a built-in estimate based on element count. 

The value provided here is incorrect (too small).

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
```diff
@@ -0,0 +1,378 @@
...
+        for (int i = 0; i < columns.length; i++) {
+            builder.appendString(columns[i]);
+            builder.appendInt(types[i].id());
+        }
+
+        return builder;
+    }
+
+    private static BinaryTupleBuilder valueBuilder(String[] columnNames, ColumnType[] types, Object[] values) {
+        BinaryTupleBuilder builder = new BinaryTupleBuilder(values.length, values.length);
```
ptupitsyn, line 192:
```suggestion
        BinaryTupleBuilder builder = new BinaryTupleBuilder(values.length);
```

Same as above.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
```diff
@@ -0,0 +1,378 @@
...
+                    return (b, v) -> b.appendBytes((byte[]) v);
+                case PERIOD:
+                    return (b, v) -> b.appendPeriod((Period) v);
+                case DURATION:
+                    return (b, v) -> b.appendDuration((Duration) v);
+                default:
+                    throw new IllegalArgumentException("Unsupported type: " + type);
+            }
+        } catch (ClassCastException e) {
+            throw new IgniteException(PROTOCOL_ERR, "Column's type mismatch ["
```
ptupitsyn, line 250: `PROTOCOL_ERR` is thin client specific. Is this intentional?

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientComputeJobUnpacker.java
```diff
@@ -0,0 +1,78 @@
...
+                    return tryUnmarshalOrCast(marshaller, unpacker.unpackObjectFromBinaryTuple());
+                }
+
+                return unpacker.unpackObjectFromBinaryTuple();
+            case MARSHALLED_TUPLE:
+                return TupleWithSchemaMarshalling.unmarshal(unpacker.readBinary());
+            case MARSHALLED_OBJECT:
+                return tryUnmarshalOrCast(marshaller, unpacker.readBinary());
+            default:
+                throw new IllegalArgumentException("Unsupported type id: " + typeId);
```
ptupitsyn, line 75:
```suggestion
                throw new IgniteException(PROTOCOL_ERR, "Unsupported compute job type id: " + typeId);
```

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientComputeJobPacker.java
```diff
@@ -0,0 +1,144 @@
...
+
+        if (obj == null) {
+            packer.packInt(ColumnType.NULL.id());
+            packer.packNil();
+            return;
+        }
+
+        var type = getColumnTypeFrom(obj);
+
+        packer.packInt(type.id());
```
ptupitsyn, line 100: * This type id is not used during unpack
* It is duplicated as part of the binary tuple

I suggest to make `ComputeJobType.NATIVE` just another constant.

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientComputeJobUnpacker.java
```diff
@@ -0,0 +1,78 @@
...
+
+        var type = ComputeJobType.Type.fromId(typeId);
+        switch (type) {
+            case NATIVE:
+                ColumnType columnType = ColumnType.getById(typeId);
+                if (columnType != ColumnType.BYTE_ARRAY) {
+                    return unpacker.unpackObjectFromBinaryTuple();
+                }
+
+                if (marshaller != null) {
```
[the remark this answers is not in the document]
PakhomovAlexander: Please, look at `ClientComputeJobPackerUnpackerTest.packByteArrayUnpackStringWithMarshaller_jobArgument`.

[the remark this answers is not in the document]
ptupitsyn: > pack job argument without marshaller
> unpack job argument with marshaller

This is not a valid use case. Either we use marshaller or we do not, on both sides.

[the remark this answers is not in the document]
PakhomovAlexander: Why? I can accept raw bytes in job that is marshaled on the client and vise-versa. I don't see the reason to restrict this, can you clarify your thoughts?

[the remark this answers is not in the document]
ptupitsyn: * Too complicated
* Confusing for the users
* Difficult to troubleshoot if the user makes a mistake somewhere

It is easier to declare and implement 3 specific modes
* Built-in types
* Tuples
* Custom types (marshaller required on both sides)

Let's throw an exception if we detect a mix of modes at any step.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
ptupitsyn: Still unresolved.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
ptupitsyn: Still unresolved.

modules/binary-tuple/src/main/java/org/apache/ignite/internal/binarytuple/inlineschema/TupleWithSchemaMarshalling.java
ptupitsyn: Still unresolved.

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientComputeJobUnpacker.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        int typeId = unpacker.unpackInt();
+        var type = ComputeJobType.Type.fromId(typeId);
+
+        switch (type) {
+            case NATIVE:
+                if (marshaller != null) {
+                    throw new UnmarshallingException(
+                            "Can not unpack object because the marshaller is provided but the object were packed without marshaller."
```
ptupitsyn, line 66:
```suggestion
                            "Can not unpack object because the marshaller is provided but the object was packed without marshaller."
```

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ClientComputeJobUnpacker.java
```diff
@@ -0,0 +1,107 @@
...
+                }
+
+                return unpacker.unpackObjectFromBinaryTuple();
+            case MARSHALLED_TUPLE:
+                return TupleWithSchemaMarshalling.unmarshal(unpacker.readBinary());
+
+            case MARSHALLED_OBJECT:
+                if (marshaller == null) {
+                    throw new UnmarshallingException(
+                            "Can not unpack object because the marshaller is not provided but the object were packed with marshaller."
```
ptupitsyn, line 77:
```suggestion
                            "Can not unpack object because the marshaller is not provided but the object was packed with marshaller."
```

modules/client-common/src/main/java/org/apache/ignite/internal/client/proto/ComputeJobType.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * The type of the object that can be passed/returned to/from the compute job. In can be a native type that is represented by
+ * {@link ColumnType} or a marshalled object/tuple.
+ */
+public class ComputeJobType {
+    static final int NATIVE_ID = 0;
+    static final int MARSHALLED_TUPLE_ID = 1;
+    static final int MARSHALLED_OBJECT_ID = 2;
+
```
ptupitsyn, line 30: I thin we can either use 3 constants, or a enum, no need to have both.

ptupitsyn (approved):

PR #4143, opened 2024-07-25
IGNITE-22819 Fix Metastorage revisions inconsistency

sanpwc:
https://issues.apache.org/jira/browse/IGNITE-22819


modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/command/EvictIdempotentCommandsCacheCommand.java
```diff
@@ -0,0 +1,27 @@
...
+
+package org.apache.ignite.internal.metastorage.command;
+
+import org.apache.ignite.internal.network.annotations.Transferable;
+
+/**
+ * Removes obsolete entries from both volatile and persistent idempotent command cache.
+ */
+@Transferable(MetastorageCommandsMessageGroup.EVICT_IDEMPOTENT_COMMAND_CACHE)
+public interface EvictIdempotentCommandsCacheCommand extends MetaStorageWriteCommand {
```
ibessonov, line 26: Shouldn't we pass consistent time with this command?

sanpwc: You are right, we should. Fixed.

modules/metastorage-cache/src/main/java/org/apache/ignite/internal/metastorage/cache/IdempotentCacheVacuumizer.java
```diff
@@ -0,0 +1,115 @@
...
+import org.apache.ignite.internal.hlc.ClockService;
+import org.apache.ignite.internal.hlc.HybridTimestamp;
+import org.apache.ignite.internal.logger.IgniteLogger;
+import org.apache.ignite.internal.logger.Loggers;
+import org.apache.ignite.internal.metastorage.impl.ElectionListener;
+import org.apache.ignite.network.ClusterNode;
+
+/**
+ * Scheduler wrapper that triggers idempotent cache vacuumization with an ability to suspend and resume the triggering. It is valid but not
+ * effective to have multiple vacuumizers at the same time, meaning that best-effort oneness is preferable. In order to achieve such
```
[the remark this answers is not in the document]
sanpwc: I don't think so. Not sure.

modules/metastorage-cache/src/main/java/org/apache/ignite/internal/metastorage/cache/IdempotentCacheVacuumizer.java
```diff
@@ -0,0 +1,115 @@
...
+
+        scheduler.scheduleWithFixedDelay(
+                () -> {
+                    if (triggerVacuumization.get()) {
+                        try {
+                            vacuumizationAction.accept(hybridTimestamp(clockService.nowLong()
+                                    - (idempotentCacheTtl.value() + clockService.maxClockSkewMillis())));
+                        } catch (Exception e) {
+                            LOG.warn("An exception occurred while executing idempotent cache vacuumization action."
+                                    + " Idempotent cache vacuumizer will not be stopped.", e);
```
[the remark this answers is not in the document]
sanpwc: Sure, fixed.

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageServiceImpl.java
```diff
@@ -332,4 +349,19 @@ private RemoveAllCommand removeAllCommand(MetaStorageCommandsFactory commandsFac
 
         return commandsFactory.removeAllCommand().keys(list).initiatorTime(ts).build();
     }
+
+    /**
+     * Creates evict idempotent commands cache command.
+     *
+     * @param commandsFactory Commands factory.
+     * @param evictionTimestamp Cached entries older than given timestamp will be evicted.
+     * @param ts Local time.
```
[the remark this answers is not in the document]
sanpwc: Fixed.

modules/metastorage-cache/src/main/java/org/apache/ignite/internal/metastorage/cache/IdempotentCacheVacuumizer.java
```diff
@@ -0,0 +1,115 @@
...
+    ) {
+        this.nodeName = nodeName;
+        this.triggerVacuumization = new AtomicBoolean(false);
+
+        scheduler.scheduleWithFixedDelay(
+                () -> {
+                    if (triggerVacuumization.get()) {
+                        try {
+                            vacuumizationAction.accept(hybridTimestamp(clockService.nowLong()
+                                    - (idempotentCacheTtl.value() + clockService.maxClockSkewMillis())));
```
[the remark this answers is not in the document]
sanpwc: Yep, otherwise we may evict cache prior to client operation finalization, meaning that there still a chance that client will re-send given invoke.

modules/metastorage-cache/src/main/java/org/apache/ignite/internal/metastorage/cache/IdempotentCacheVacuumizer.java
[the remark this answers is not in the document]
sanpwc: Fixed.

modules/metastorage-cache/src/main/java/org/apache/ignite/internal/metastorage/cache/IdempotentCacheVacuumizer.java
```diff
@@ -0,0 +1,115 @@
...
+            ConfigurationValue<Long> idempotentCacheTtl,
+            ClockService clockService,
+            long initialDelay,
+            long delay,
+            TimeUnit unit
+    ) {
+        this.nodeName = nodeName;
+        this.triggerVacuumization = new AtomicBoolean(false);
+
+        scheduler.scheduleWithFixedDelay(
```
ibessonov, line 73: Do we stop this future when we shutdown Ignite node? I don't see where we would do it, can you please add such code?

modules/metastorage-cache/src/main/java/org/apache/ignite/internal/metastorage/cache/IdempotentCacheVacuumizer.java
```diff
@@ -0,0 +1,115 @@
...
+     * @param idempotentCacheTtl Idempotent cache ttl.
+     * @param clockService Clock service.
+     * @param initialDelay The time to delay first execution.
+     * @param delay The delay between the termination of one execution and the commencement of the next.
+     * @param unit The time unit of the initialDelay and delay parameters.
+     */
+    public IdempotentCacheVacuumizer(
+            String nodeName,
+            ScheduledExecutorService scheduler,
+            Consumer<HybridTimestamp> vacuumizationAction,
```
ibessonov, line 63: Why do we need to abstract vacuumizer from the action that it performs?

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/raft/MetaStorageWriteHandler.java
```diff
@@ -368,56 +360,51 @@ void onSnapshotLoad() {
...
+                    .map(Entry::key)
                     .collect(toList());
 
-            if (!commandIdsToRemove.isEmpty()) {
-                List<byte[]> commandIdStorageKeys = commandIdsToRemove.stream()
-                        .map(commandId -> ArrayUtils.concat(new byte[]{}, ByteUtils.toBytes(commandId)))
-                        .collect(toList());
+            // TODO https://issues.apache.org/jira/browse/IGNITE-22828
+            evictionCandidateKeys.forEach(evictionCandidateKeyBytes -> {
+                byte[] commandIdBytes = copyOfRange(evictionCandidateKeyBytes, IDEMPOTENT_COMMAND_PREFIX_BYTES.length,
```
ibessonov, line 392: Ideally, we should be able to read data from array without cropping it, that's just a food for your thoughts for the future

ibessonov (approved):
Give you my approval in advance, looks good!

modules/metastorage-cache/src/main/java/org/apache/ignite/internal/metastorage/cache/IdempotentCacheVacuumizer.java
sanpwc: In order to ease testing. That's the only reason.

modules/metastorage-cache/src/main/java/org/apache/ignite/internal/metastorage/cache/IdempotentCacheVacuumizer.java
sanpwc: I use `org.apache.ignite.internal.app.ThreadPoolsManager#commonScheduler` in idempotent cache vacuumizer, thus pending operations effectively awaited on `org.apache.ignite.internal.app.ThreadPoolsManager#stopAsync`, precisely in `IgniteUtils.shutdownAndAwaitTermination(commonScheduler, 10, SECONDS);`.

Do you mean that it's required to explicitly await the future?
In that case I'd rather convert IdempotentCacheVacuumizer to a component with a corresponding stopAsync().

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/raft/MetaStorageWriteHandler.java
ibessonov: I meant that we can pass `(bytes, from, length)` where we need it instead of using `copyOfRange`

modules/metastorage-cache/src/main/java/org/apache/ignite/internal/metastorage/cache/IdempotentCacheVacuumizer.java
ibessonov: I think that it's a good thing to explicitly cancel all scheduled futures. Your approach does work, it's just not as clean

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/server/raft/MetaStorageWriteHandler.java
sanpwc: You mean adding
```
    /**
     * Deserializes an object from byte array using native java serialization mechanism.
     *
     * @param bytes Byte array.
     * @param from – the offset in the buffer of the first byte to read.
     * @param length – the maximum number of bytes to read from the buffer.
     * @return Object.
     */
    public static <T> T fromBytes(byte[] bytes, int from, int length) {
        try (
                var bis = new ByteArrayInputStream(bytes, from, length);
                var in = new ObjectInputStream(bis)
        ) {
            return (T) in.readObject();
        } catch (IOException | ClassNotFoundException e) {
            throw new IgniteInternalException("Could not deserialize an object", e);
        }
    }
```
?

ibessonov: Yes, that's what I meant

modules/metastorage-cache/src/main/java/org/apache/ignite/internal/metastorage/cache/IdempotentCacheVacuumizer.java
sanpwc: Okaaaaay

ibessonov (approved):

modules/metastorage/src/main/java/org/apache/ignite/internal/metastorage/impl/MetaStorageServiceImpl.java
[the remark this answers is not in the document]
denis-chudov: but you swapped the parameters in the method signature, not javadoc, and didn't change the usage...
The method is called with `evictionTimestamp` as 3rd parameter but in the method it's 2nd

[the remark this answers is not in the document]
sanpwc: Fixed.

denis-chudov (approved):

PR #4179, opened 2024-08-03
IGNITE-22635 Sql. Fix "unable to optimize plan" error for input out of range

xtern:
https://issues.apache.org/jira/browse/IGNITE-22635

Fixed the following test case

```java
BigDecimal val = new BigDecimal(Long.MAX_VALUE).add(BigDecimal.ONE);
String query = format( "SELECT {}::INT", val);
checkSqlError(Sql.RUNTIME_ERR, "out of range", igniteSql(), query);
```

previously failed 
```
org.opentest4j.AssertionFailedError: Error code does not match. Expected: "SQL-7" actual: "CMN--1"
org.apache.ignite.sql.SqlException: IGN-CMN-65535 TraceId:3c508f8c-8925-441a-a74b-7097b4e4b919 Unable to optimize plan due to internal error
```

The following issue ([IGNITE-22932](https://issues.apache.org/jira/browse/IGNITE-22932)) was created to fix an issue with the type specification in the error message:


lowka (approved):

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PlannerHelper.java
```diff
@@ -180,7 +180,7 @@ public static IgniteRel optimize(SqlNode sqlNode, IgnitePlanner planner) {
                 LOG.debug(planner.dump());
             }
 
-            if (ex instanceof CannotPlanException) {
+            if (ex instanceof CannotPlanException || ex instanceof SqlException) {
                 throw ex;
             } else if (ex.getClass() == RuntimeException.class && ex.getCause() instanceof SqlException) {
```
ygerzhedovich, line 185: but now the code is dead, isn't it?

xtern: No, it's a different case.
Test `ForceIndexHintPlannerTest.testWithMultipleIndexHints` reproduces it

modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/prepare/PlannerHelper.java
```diff
@@ -180,7 +180,7 @@ public static IgniteRel optimize(SqlNode sqlNode, IgnitePlanner planner) {
                 LOG.debug(planner.dump());
             }
 
-            if (ex instanceof CannotPlanException) {
+            if (ex instanceof CannotPlanException || ex instanceof SqlException) {
```
zstan, line 183: why do u need such a change ? it work properly without it, also as for me you need to append additional test with : @WithSystemProperty(key = "FAST_QUERY_OPTIMIZATION_ENABLED", value = "false")
and also rename initial issue

xtern: Because this change fixes described problem.
Please, try to run `ItCastToIntTest` WITHOUT this change and you'll see that the `explicitCastOfLiteralsOnSelectWithOverflow` test will fail.

xtern: >  you need to append additional test with : @WithSystemProperty(key = "FAST_QUERY_OPTIMIZATION_ENABLED", value = "false")

Done

> rename initial issue

I don't get why and how to rename, this issue is reproduced even fast query optimization is turned off

zstan (approved):

PR #4186, opened 2024-08-06
IGNITE-22803 Put cluster ID in CmgInitMessage

rpuch:
https://issues.apache.org/jira/browse/IGNITE-22803

### DIFF
--- a/modules/client-handler/src/integrationTest/java/org/apache/ignite/client/handler/TestServer.java
+++ b/modules/client-handler/src/integrationTest/java/org/apache/ignite/client/handler/TestServer.java
@@ -22,6 +22,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.mockito.Answers.RETURNS_DEEP_STUBS;
 import static org.mockito.Mockito.mock;
 
+import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import org.apache.ignite.client.handler.configuration.ClientConnectorConfiguration;
 import org.apache.ignite.internal.catalog.CatalogService;
@@ -126,7 +127,7 @@ public class TestServer {
                 mock(IgniteComputeInternal.class),
                 clusterService,
                 bootstrapFactory,
-                () -> CompletableFuture.completedFuture(ClusterTag.clusterTag(msgFactory, "Test Server")),
+                () -> CompletableFuture.completedFuture(ClusterTag.clusterTag(msgFactory, "Test Server", UUID.randomUUID())),
                 mock(MetricManagerImpl.class),
                 metrics,
                 authenticationManager,
--- a/modules/client-handler/src/integrationTest/java/org/apache/ignite/client/handler/TestServer.java
+++ b/modules/client-handler/src/integrationTest/java/org/apache/ignite/client/handler/TestServer.java
@@ -22,7 +22,6 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.mockito.Answers.RETURNS_DEEP_STUBS;
 import static org.mockito.Mockito.mock;
 
-import java.util.UUID;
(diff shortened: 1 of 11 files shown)


modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterTag.java
```diff
@@ -45,7 +45,22 @@ public interface ClusterTag extends NetworkMessage, Serializable {
...
+     *
+     * @param msgFactory Message factory to instantiate builder.
+     * @param name Cluster name.
+     * @param clusterId Cluster ID.
+     * @return Cluster tag instance.
+     */
+    static ClusterTag clusterTag(CmgMessagesFactory msgFactory, String name, UUID clusterId) {
         return msgFactory.clusterTag()
                 .clusterName(name)
                 .clusterId(UUID.randomUUID())
```
Phillippko, line 66: clusterID is ignored. Strange, that IDEA doesn't grey out unused parameters

modules/client-handler/src/integrationTest/java/org/apache/ignite/client/handler/TestServer.java
```diff
@@ -126,7 +127,7 @@ ClientHandlerModule start(TestInfo testInfo) {
                 mock(IgniteComputeInternal.class),
                 clusterService,
                 bootstrapFactory,
-                () -> CompletableFuture.completedFuture(ClusterTag.clusterTag(msgFactory, "Test Server")),
+                () -> CompletableFuture.completedFuture(ClusterTag.clusterTag(msgFactory, "Test Server", UUID.randomUUID())),
```
Phillippko, line 130: randomClusterTag?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterTag.java
```diff
@@ -45,7 +45,22 @@ public interface ClusterTag extends NetworkMessage, Serializable {
      * @param name Cluster name.
      * @return Cluster tag instance.
      */
-    static ClusterTag clusterTag(CmgMessagesFactory msgFactory, String name) {
+    static ClusterTag randomClusterTag(CmgMessagesFactory msgFactory, String name) {
```
Phillippko, line 48: I assume random tag should be used only in tests, let's annotate it as TestOnly? 

If it's really only for tests, do we really need a separate method? Name is changed, so we have to change all test classes in any case))

modules/cluster-management/src/test/java/org/apache/ignite/internal/cluster/management/raft/AbstractClusterStateStorageManagerTest.java
```diff
@@ -74,7 +73,7 @@ void tearDown() {
      */
     @Test
     void testClusterState() {
-        ClusterTag clusterTag1 = clusterTag(msgFactory, "cluster");
+        ClusterTag clusterTag1 = ClusterTag.randomClusterTag(msgFactory, "cluster");
```
Phillippko, line 76: static import?

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterTag.java
rpuch: Good catch!

modules/cluster-management/src/main/java/org/apache/ignite/internal/cluster/management/ClusterTag.java
rpuch: I added the annotation.

I think it still makes sense to use a special method, just to reduce duplication

PR #4190, opened 2024-08-06
IGNITE-22933 Replace ThreadPoolExecutors with a corePoolSize equal to 0

rpuch:
https://issues.apache.org/jira/browse/IGNITE-22933


modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
```diff
@@ -218,16 +218,19 @@ public ConnectionManager(
 
         this.clientBootstrap = bootstrapFactory.createClientBootstrap();
 
-        // We don't just use Executors#newSingleThreadExecutor() here because it defines corePoolSize=1, so the maintenance thread will
+        // We don't just use Executors#newSingleThreadExecutor() here because the maintenance thread will
         // be kept alive forever, and we only need it from time to time, so it seems a waste to keep the thread alive.
-        connectionMaintenanceExecutor = new ThreadPoolExecutor(
-                0,
+        ThreadPoolExecutor maintenanceExecutor = new ThreadPoolExecutor(
```
sashapolo, line 223: Can we use `singleThreadExecutor` now, since we are using `allowCoreThreadTimeOut`?

modules/replicator/src/integrationTest/java/org/apache/ignite/internal/replicator/ItPlacementDriverReplicaSideTest.java
```diff
@@ -152,12 +152,15 @@ public class ItPlacementDriverReplicaSideTest extends IgniteAbstractTest {
 
     @BeforeEach
     public void beforeTest(TestInfo testInfo) {
-        partitionOperationsExecutor = new ThreadPoolExecutor(
-                0, 20,
-                0, TimeUnit.SECONDS,
+        ThreadPoolExecutor threadPoolExecutor = new ThreadPoolExecutor(
+                20, 20,
```
sashapolo, line 156: Do we really need 20 threads in tests?

modules/replicator/src/integrationTest/java/org/apache/ignite/internal/replicator/ItPlacementDriverReplicaSideTest.java
```diff
@@ -152,12 +152,15 @@ public class ItPlacementDriverReplicaSideTest extends IgniteAbstractTest {
...
+                10, TimeUnit.SECONDS,
                 new LinkedBlockingQueue<>(),
                 NamedThreadFactory.create("test", "partition-operations", log)
         );
+        threadPoolExecutor.allowCoreThreadTimeOut(true);
```
sashapolo, line 161: Why do we need this here?

modules/replicator/src/test/java/org/apache/ignite/internal/replicator/ReplicaManagerTest.java
```diff
@@ -109,12 +109,15 @@ void startReplicaManager(
...
-        requestsExecutor = new ThreadPoolExecutor(
-                0, 5,
-                0, TimeUnit.SECONDS,
+        ThreadPoolExecutor threadPoolExecutor = new ThreadPoolExecutor(
+                5, 5,
+                10, TimeUnit.SECONDS,
                 new LinkedBlockingQueue<>(),
                 NamedThreadFactory.create(nodeName, "partition-operations", log)
         );
+        threadPoolExecutor.allowCoreThreadTimeOut(true);
```
sashapolo, line 118: Same question, I don't think we need to shutdown threads preemptively in tests

modules/network/src/main/java/org/apache/ignite/internal/network/netty/ConnectionManager.java
rpuch: `newSingleThreadExecutor()` sets timeout to 0, so the thread will never be let go.

sashapolo (approved):

PR #4194, opened 2024-08-07
IGNITE-22774 Remove `raftClient` map from `ItTxTestCluster`

JAkutenshi:
JIRA Ticket: [IGNITE-22774 | Remove raftClient map from ItTxTestCluster](https://issues.apache.org/jira/browse/IGNITE-22774)

## The goal

The goal is to remove `raftClient` from `ItTxTestCluster`.

## The reason

All raft-clients are created inside replicas through `ReplicaManager#startReplica`, and most of usages of `raftClient` are effectively useless

## The solution

1. The map `raftClients` is removed.
2. `getLeader` and related assertions are removed. 
3. `getLeaderId` method is rewritten as following:
```java
int partId = 0;

return replicaManagers.get(extractConsistentId(cluster.get(partId)))
                .replica(new TablePartitionId(tables.get(tableName).tableId(), partId))
                .thenApply(replica -> replica.raftClient().leader())
                .join();
```
4. Replicas are stopped instead of raft-clients from the map as following:
```java
for (String consistentId : clusterServices.keySet()) {
    ReplicaManager replicaManager = replicaManagers.get(consistentId);

    if (replicaManager == null) {
        continue;
    }

    grpIds.forEach(id -> {
        try {
            replicaManager.stopReplica(id);
        } catch (NodeStoppingException e) {
            // no-op
        }
    });
}
```
5. Extra map for tables is introduced because we need to get `tableId` in `getLeaderId`.

---

### The Review Checklist

### Notes

### DIFF
--- a/modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestSingleNodeCollocated.java
+++ b/modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestSingleNodeCollocated.java
@@ -17,8 +17,6 @@
 
 package org.apache.ignite.distributed;
 
-import static org.junit.jupiter.api.Assertions.assertSame;
-
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.TestInfo;
 
--- a/modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestSingleNodeCollocated.java
+++ b/modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestSingleNodeCollocated.java
@@ -46,15 +46,6 @@ public class ItTxDistributedTestSingleNodeCollocated extends ItTxAbstractDistrib
     @Override
     public void before() throws Exception {
         super.before();
-
-        assertSame(
-                txTestCluster.raftClients.get(ACC_TABLE_NAME).get(0).clusterService(),
-                txTestCluster.getLeader(ACC_TABLE_NAME).service()
-        );
-        assertSame(
-                txTestCluster.raftClients.get(CUST_TABLE_NAME).get(0).clusterService(),
-                txTestCluster.getLeader(CUST_TABLE_NAME).service()
-        );
     }
 }
 
--- a/modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestThreeNodesThreeReplicasCollocated.java
+++ b/modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestThreeNodesThreeReplicasCollocated.java
@@ -18,7 +18,6 @@
 package org.apache.ignite.distributed;
 
 import static org.apache.ignite.internal.testframework.IgniteTestUtils.waitForCondition;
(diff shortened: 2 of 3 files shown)


modules/table/src/testFixtures/java/org/apache/ignite/distributed/ItTxTestCluster.java
```diff
@@ -940,11 +906,25 @@ public void shutdownCluster() {
...
+            ReplicaManager replicaManager = replicaManagers.get(consistentId);
+
+            if (replicaManager == null) {
+                continue;
             }
+
+            grpIds.forEach(id -> {
+                try {
+                    replicaManager.stopReplica(id);
+                } catch (NodeStoppingException e) {
```
kgusakov, line 924: Why do we need this no-op catch? I would prefer to see this exception and not sweep it under the carpet

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestSingleNodeCollocated.java
```diff
@@ -46,15 +44,6 @@ protected boolean startClient() {
     @Override
     public void before() throws Exception {
         super.before();
-
```
kgusakov, line 49: Could you explain why these assertions were useful and why we shouldn't have the analog now?

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestThreeNodesThreeReplicasCollocated.java
```diff
@@ -50,11 +49,6 @@ public ItTxDistributedTestThreeNodesThreeReplicasCollocated(TestInfo testInfo) {
     @BeforeEach
     @Override public void before() throws Exception {
         super.before();
-
-        assertSame(
```
kgusakov, line 54: Again, could you explain?

kgusakov (changes requested):

modules/table/src/testFixtures/java/org/apache/ignite/distributed/ItTxTestCluster.java
JAkutenshi: I agree, but it's a stopping process for tests -- I'm not sure what should I do in case `NSE` while a partial node is stopping?

modules/table/src/integrationTest/java/org/apache/ignite/distributed/ItTxDistributedTestSingleNodeCollocated.java
JAkutenshi: It's explained in [the ticket's description](https://issues.apache.org/jira/browse/IGNITE-22774)

modules/table/src/testFixtures/java/org/apache/ignite/distributed/ItTxTestCluster.java
JAkutenshi: The code is deleted because it duplicates replica's stopping above the code

kgusakov (approved):

PR #4198, opened 2024-08-07
IGNITE-22955 Start local configuration before creating other components

Cyrill:
https://issues.apache.org/jira/browse/IGNITE-22955


modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteImpl.java
```diff
@@ -464,6 +463,13 @@ public class IgniteImpl implements Ignite {
                 localConfigurationValidator
         );
 
+        // Start local configuration to be able to read all local properties.
+        try {
+            lifecycleManager.startComponentsAsync(new ComponentContext(), nodeCfgMgr);
```
ibessonov, line 468: Wow, I'm amazed that local configuration manager has absolutely no dependencies, that's wild!

Could there be any issues with passing a different context instance? Maybe it would affect stopping somehow? That's just a question, I'm not aware of what this context is doing

ibessonov (approved):

PR #4209, opened 2024-08-08
IGNITE-22837 Invocation of the local raft client happens in a differe…

vldpyatkov:
…nt pool

https://issues.apache.org/jira/browse/IGNITE-22837

### DIFF
--- a/modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/JraftServerImpl.java
+++ b/modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/JraftServerImpl.java
@@ -17,6 +17,7 @@
 
 package org.apache.ignite.internal.raft.server.impl;
 
+import static java.util.concurrent.TimeUnit.SECONDS;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toUnmodifiableList;
 import static org.apache.ignite.internal.thread.ThreadOperation.STORAGE_READ;
@@ -39,6 +40,9 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 import java.util.function.BiConsumer;
 import java.util.function.BiPredicate;
 import java.util.stream.IntStream;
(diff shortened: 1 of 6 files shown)


modules/raft/src/main/java/org/apache/ignite/raft/jraft/rpc/impl/IgniteRpcServer.java
```diff
@@ -175,13 +179,37 @@ else if (prc.executor() != null) {
...
+                } else {
+                    finalPrc.handleRequest(new NetworkRpcContext(executor, sender, correlationId), message);
+                }
             } catch (RejectedExecutionException e) {
                 // The rejection is ok if an executor has been stopped, otherwise it shouldn't happen.
                 LOG.warn("A request execution was rejected [sender={} req={} reason={}]", sender, S.toString(message), e.getMessage());
             }
         }
 
+        private boolean shouldSwitchToRequestsExecutor() {
```
rpuch, line 193: This method differs from the corresponding method in ReplicaManager in a tiny way (this method does not check for 3rd permission). Would it make sense to extract a common method and parameterize it with the set of permissions to check, to avoid duplication?

modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/JraftServerImpl.java
```diff
@@ -264,7 +268,13 @@ public CompletableFuture<Void> startAsync(ComponentContext componentContext) {
             opts.setSnapshotTimer(JRaftUtils.createTimer(opts, "JRaft-SnapshotTimer"));
         }
 
-        requestExecutor = JRaftUtils.createRequestExecutor(opts);
+        requestExecutor = new ThreadPoolExecutor(
```
rpuch, line 271: It seems to be the same as `Executors.newFixedThreadPool(opts.getRaftRpcThreadPoolSize(), ... factory ...)

modules/raft/src/main/java/org/apache/ignite/internal/raft/server/impl/JraftServerImpl.java
```diff
@@ -264,7 +268,13 @@ public CompletableFuture<Void> startAsync(ComponentContext componentContext) {
...
+                opts.getRaftRpcThreadPoolSize(),
+                opts.getRaftRpcThreadPoolSize(),
+                0, SECONDS,
+                new LinkedBlockingQueue<>(),
+                IgniteThreadFactory.create(opts.getServerName(), "JRaft-Request-Processor", LOG, STORAGE_READ, STORAGE_WRITE)
```
rpuch, line 276: Why do we need to allow it STORAGE_READ and STORAGE_WRITE?

rpuch (approved):

PR #4214, opened 2024-08-09
IGNITE-22971 Reduce log pollution by RaftGroupServiceImpl

sashapolo:
https://issues.apache.org/jira/browse/IGNITE-22971

### The Review Checklist

### Notes


modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
```diff
@@ -811,9 +810,20 @@ private CompletableFuture<ClusterNode> resolvePeer(Peer peer) {
         ClusterNode node = cluster.topologyService().getByConsistentId(peer.consistentId());
 
         if (node == null) {
-            return CompletableFuture.failedFuture(new ConnectException("Peer " + peer.consistentId() + " is unavailable"));
+            return CompletableFuture.failedFuture(new PeerUnavailableException("Peer " + peer.consistentId() + " is unavailable"));
```
sanpwc, line 813: - Please fix the recoverable() javadoc. Currently it states  
`Checks if an error is recoverable, for example, {@link java.net.ConnectException}.` but ConnectException itself isn't considered as recoverable. BWT is a bit funny that ConnectionException wasn't handled.

- There are at least two placed where we still use ConnectException:

1. RaftGroupServiceTest
```
            if (peer != null && target.name().equals(peer.consistentId())) {
                return failedFuture(new ConnectException());
            }
```
2. org.apache.ignite.raft.jraft.rpc.impl.IgniteRpcClient#send
```
        if (targetNode == null) {
            // ConnectException will force a retry by the enclosing components.
            fut.completeExceptionally(new ConnectException());

            return;
        }
```

modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
```diff
@@ -598,6 +597,8 @@ private void handleThrowable(
             long stopTime,
             CompletableFuture<? extends NetworkMessage> fut
     ) {
+        err = unwrapCause(err);
```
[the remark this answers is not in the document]
sashapolo: > Am I right that you will always exclude

Yes, these exceptions are not very useful, they just always indicate that you failed inside `sendWithRetry`. 

> however it's definitely beyond the task description

This is not true to some extension, without this unwrapping, the `PeerUnavailableException` will also be wrapped in `CompeltionException`. I really don't want to add unwrapping just for this particular case.

modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
```diff
@@ -713,11 +714,9 @@ private void scheduleRetry(Runnable runnable) {
      * @return {@code True} if this is a recoverable exception.
      */
     private static boolean recoverable(Throwable t) {
-        if (t instanceof ExecutionException || t instanceof CompletionException) {
-            t = t.getCause();
-        }
+        t = unwrapCause(t);
```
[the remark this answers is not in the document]
sashapolo: yes, just in case

modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
sashapolo: > BWT is a bit funny that ConnectionException wasn't handled.

It was, `ConnectException` is an `IOException`

modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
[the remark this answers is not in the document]
sanpwc: Ok, make sense.

modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
sanpwc: > It was, ConnectException is an IOException
Yep, you are right.

What about other topics?

sashapolo: > There are at least two placed where we still use ConnectException:

The first place that you mentioned (`RaftGroupServiceTest`) is a test where we mock the `ClusterService` behavior, I think that using a `ConnectException` is the correct choice there.

sanpwc: Ok, and what about `org.apache.ignite.raft.jraft.rpc.impl.IgniteRpcClient#send`?

sashapolo: I'm trying to fix it now)

sashapolo: I'm not sure about the other place too. We can change the exception type, but then `IgniteRpcClient#send` will be able to throw both `PeerUnavailableException` (when the target is not in the physical topology) and `ConnectException` (when the target is still in the topology, but failed to respond for some reason?). This means that the code inside `AbstractClientService#invokeWithDone` will need to handle both of these exception types. I don't know why it was written like that (to handle `ConnectException` is a special way), but it is what it is

sanpwc: Well, I believe that it's better to be consistent. If we consider node == null as PeerUnavailableException we should do the same in IgniteRpcClient#send in case of targetNode == null. Up to you.

sanpwc (approved):

modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
sashapolo: Fixed

sanpwc: Thanks! Proposed changes looks good.

PR #4215, opened 2024-08-09
IGNITE-22891 Fixed ItMetaStorageRaftGroupTest#testRangeNextWorksCorrectlyAfterLeaderChange

denis-chudov:
https://issues.apache.org/jira/browse/IGNITE-22891

The reasons of test failures:
- oldLeaderId was initialized without refresh leader, sometimes wrong node was stopped
- the logic with waiting for desired count of replicators started/stopped was unstable and led to that the resultFuture was never completed


modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/server/raft/ItMetaStorageRaftGroupTest.java
```diff
@@ -224,7 +224,6 @@ public void afterTest() {
      * @throws Exception If failed.
      */
     @Test
-    @Disabled("https://issues.apache.org/jira/browse/IGNITE-22891")
```
sanpwc, line 227: Can you please explain, what was the reason for the test failure?

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/server/raft/ItMetaStorageRaftGroupTest.java
```diff
@@ -316,22 +326,24 @@ public void onNext(Entry item) {
...
+                                        .refreshAndGetLeaderWithTerm();
+                                assertThat(newLeaderWithTermFut, willCompleteSuccessfully());
+                                LeaderWithTerm newLeaderWithTerm = newLeaderWithTermFut.join();
 
-                                assertNotSame(oldLeaderId, raftGroupServiceOfLiveServer.leader().consistentId());
+                                assertNotSame(oldLeaderId, newLeaderWithTerm.leader().consistentId());
 
-                                // ensure that leader has been changed only once
-                                assertTrue(
-                                        waitForCondition(() -> replicatorStartedCounter.get() == 4, 5_000),
```
sanpwc, line 325: Is it possible to drop both replicatorStartedCounter and replcicatorStoppedCounter entirely?
Hope that it is possible to rewrite both onSubscribe and onNext in order to use terms instead of aforementioned counters.

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/server/raft/ItMetaStorageRaftGroupTest.java
denis-chudov: - oldLeaderId was initialized without refresh leader, sometimes wrong node was stopped
- the logic with waiting for desired count of replicators started/stopped was unstable and led to that the resultFuture was never completed

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/server/raft/ItMetaStorageRaftGroupTest.java
denis-chudov: As I understand, replicatorStartedCounter is used for waiting for majority in `onSubscribe`. Regarding replicatorStoppedCounter, seems it's not necessary already

sanpwc: But since we do wait for refresh leader completeness prior to range() now - explicit majority awaiting logic based on replicaStartedCounter is no longer needer, WDYT?

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/server/raft/ItMetaStorageRaftGroupTest.java
sanpwc: Yep, makes sense. Thanks!

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/server/raft/ItMetaStorageRaftGroupTest.java
denis-chudov: I agree.
Tried 500 repetitions with removed counters: 
https://ci.ignite.apache.org/buildConfiguration/ApacheIgnite3xGradle_Test_IntegrationTests_ModuleMetastorageClient/8389221?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildDeploymentsSection=false&expandBuildChangesSection=true

sanpwc (approved):

modules/metastorage/src/integrationTest/java/org/apache/ignite/internal/metastorage/server/raft/ItMetaStorageRaftGroupTest.java
```diff
@@ -316,22 +301,24 @@ public void onNext(Entry item) {
                                         .stopAsync(componentContext);
                                 assertThat(stopFuture, willCompleteSuccessfully());
 
-                                raftGroupServiceOfLiveServer.refreshLeader().get();
+                                CompletableFuture<LeaderWithTerm> newLeaderWithTermFut = raftGroupServiceOfLiveServer
+                                        .refreshAndGetLeaderWithTerm();
+                                assertThat(newLeaderWithTermFut, willCompleteSuccessfully());
+                                LeaderWithTerm newLeaderWithTerm = newLeaderWithTermFut.join();
```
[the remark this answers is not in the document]
denis-chudov: added, thanks!

PR #4218, opened 2024-08-12
IGNITE-22838 Operation returns to partition pool after it is handled …

vldpyatkov:
…in RAFT

https://issues.apache.org/jira/browse/IGNITE-22838


modules/raft/src/main/java/org/apache/ignite/internal/raft/ExecutorInclinedRaftCommandRunner.java
```diff
@@ -44,6 +46,16 @@ public <R> CompletableFuture<R> run(Command cmd) {
             return future;
         }
 
+        // We can wait for replication completion right here, because client thread waits the entire operation in synchronous API any case.
```
rpuch, line 49:
```suggestion
        // We can wait for replication completion right here, because client thread waits the entire operation in synchronous API in any case.
```

rpuch (approved):

PR #4223, opened 2024-08-13
IGNITE-22851 Move raft client's update configuration process into replicas

JAkutenshi:
JIRA Ticket: [IGNITE-22851 | Move raft client's update configuration process into replicas](https://issues.apache.org/jira/browse/IGNITE-22851)

## The goal

The main goal is to remove all calls `RaftGroupService#updateConfiguration` from `TableManager`.

## The reason

There should no any raft client usages inside `TableManager`.

## The solution

1. Method Replica#updateConfiguration is introduced.
2. All calls like `raftClient -> raftClient.updateConfiguration(...)` are replaced with `replica -> replica.updateConfiguration(...)`

---

### The Review Checklist

### Notes

### DIFF
--- a/modules/replicator/src/main/java/org/apache/ignite/internal/replicator/Replica.java
+++ b/modules/replicator/src/main/java/org/apache/ignite/internal/replicator/Replica.java
@@ -20,6 +20,7 @@ package org.apache.ignite.internal.replicator;
 import java.util.concurrent.CompletableFuture;
 import org.apache.ignite.internal.network.NetworkMessage;
 import org.apache.ignite.internal.placementdriver.message.PlacementDriverReplicaMessage;
+import org.apache.ignite.internal.raft.PeersAndLearners;
 import org.apache.ignite.internal.raft.client.TopologyAwareRaftGroupService;
 import org.apache.ignite.internal.replicator.listener.ReplicaListener;
 import org.apache.ignite.internal.replicator.message.ReplicaRequest;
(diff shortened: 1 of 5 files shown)


modules/replicator/src/main/java/org/apache/ignite/internal/replicator/Replica.java
```diff
@@ -73,4 +74,11 @@ public interface Replica {
      * Shutdowns the replica.
      */
     CompletableFuture<Void> shutdown();
+
+    /**
+     * Updates peers and learners lists in raft client.
+     *
+     * @param configuration Peers and learners configuration.
+     */
+    void updateConfiguration(PeersAndLearners configuration);
```
kgusakov, line 83: Could you rename it to something like "updateReplicationClientConfiguration"? Because in the scope of Replica the meaning of "configuration" is not obvious.

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/Replica.java
```diff
@@ -73,4 +74,11 @@ public interface Replica {
      * Shutdowns the replica.
      */
     CompletableFuture<Void> shutdown();
+
+    /**
+     * Updates peers and learners lists in raft client.
```
kgusakov, line 79: Let's use the generic "replication client" instead of "raft", because we want to be more abstract about replication protocol.

kgusakov (changes requested):

modules/replicator/src/main/java/org/apache/ignite/internal/replicator/Replica.java
JAkutenshi: Due to an onlibe discussion it is renamed to `updatePeersAndLearners`

kgusakov (approved):

PR #4226, opened 2024-08-13
IGNITE-22981 Relax IgniteClient close exception

valepakh:
https://issues.apache.org/jira/browse/IGNITE-22981


modules/client/src/main/java/org/apache/ignite/internal/client/ReliableChannel.java
```diff
@@ -143,21 +142,18 @@ public final class ReliableChannel implements AutoCloseable {
...
-                },
-                connMgr::stop,
-                () -> shutdownAndAwaitTermination(streamerFlushExecutor, 10, TimeUnit.SECONDS));
+        if (holders != null) {
+            for (ClientChannelHolder hld : holders) {
+                hld.close();
+            }
+        }
+        connMgr.stop();
+        shutdownAndAwaitTermination(streamerFlushExecutor, 10, TimeUnit.SECONDS);
```
ptupitsyn, line 156: This is not the same as the old `closeAllManually` call - we should wrap every `close()` in a try-catch so that everything is closed even if there was an exception.

We should also do this for `holders`.

ptupitsyn (changes requested):

PR #4227, opened 2024-08-13
IGNITE-22977

ascherbakoff:


modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -2795,26 +2761,28 @@ private CompletableFuture<CompletableFuture<?>> applyUpdateCommand(
                 return resultFuture.thenApply(res -> {
                     UpdateCommandResult updateCommandResult = (UpdateCommandResult) res;
 
-                    if (full && !updateCommandResult.isPrimaryReplicaMatch()) {
+                    if (full && updateCommandResult != null && !updateCommandResult.isPrimaryReplicaMatch()) {
```
vldpyatkov, line 2764: I think this is done to handle a test result (SKIP_REPLICATION_IN_BENCHMARK).
Maybe it would be better to give the correct response to avoid this checking.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -513,13 +519,38 @@ private CompletableFuture<?> processRequest(ReplicaRequest request, @Nullable Bo
             return processGetEstimatedSizeRequest();
         }
 
-        HybridTimestamp opTsIfDirectRo = (request instanceof ReadOnlyDirectReplicaRequest) ? clockService.now() : null;
+        @Nullable HybridTimestamp opTs = getTxOpTimestamp(request);
+        @Nullable HybridTimestamp opTsIfDirectRo = (request instanceof ReadOnlyDirectReplicaRequest) ? opTs : null;
+        @Nullable HybridTimestamp txTs = getTxStartTimestamp(request);
+        if (txTs == null) {
```
rpuch, line 525: It seems safer to switch to `request instanceof ReadOnlyDirectReplicaRequest` condition (or at least assert it)

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -591,96 +622,25 @@ private CompletableFuture<Void> triggerTxRecovery(UUID txId, String senderId) {
     }
 
     /**
-     * Validates that the table exists at a timestamp corresponding to the request operation.
+     * Return tx operation timestamp.
```
rpuch, line 625:
```suggestion
     * Returns tx operation timestamp.
```

rpuch: BTW, don't we need the detailed breakdown which timestamp is returned for which request type? This information was in the javadoc, but now it's removed.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -1982,28 +1942,34 @@ private <T> CompletableFuture<T> resolveRowByPk(
...
-                });
+                return resolvingFuture;
+            } finally {
+                if (!cursorClosureSetUp && cursor != null) {
+                    cursor.close();
+                }
+            }
+        };
+
+        if (lockFut.isDone() && !lockFut.isCompletedExceptionally()) {
```
rpuch, line 1968:
```suggestion
        if (CompletableFutures.isCompletedSuccessfully(lockFut)) {
```

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
ascherbakoff: This is not possible due to dependency issues. The response class is not available in raft module.

modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
```diff
@@ -171,19 +174,40 @@ public static CompletableFuture<RaftGroupService> start(
...
+                    factory,
+                    configuration,
+                    membersConfiguration,
+                    null,
+                    executor,
+                    commandsMarshaller
+            ) {
+                @Override
+                public <R> CompletableFuture<R> run(Command cmd) {
+                    return cmd.getClass().getSimpleName().contains("UpdateCommand") ? nullCompletedFuture() : super.run(cmd);
```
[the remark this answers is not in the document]
ascherbakoff: This is not possible to fix due to dependency issues. UpdateCommand is not available in raft module.

modules/core/src/main/java/org/apache/ignite/internal/lang/IgniteSystemProperties.java
```diff
@@ -73,6 +73,12 @@ public final class IgniteSystemProperties {
     /** Name of the property controlling whether, when a thread assertion is triggered, it should also be written to the log. */
     public static final String THREAD_ASSERTIONS_LOG_BEFORE_THROWING = "THREAD_ASSERTIONS_LOG_BEFORE_THROWING";
 
+    /** Skip replication in a benchmark. */
+    public static final String SKIP_REPLICATION_IN_BENCHMARK = "SKIP_REPLICATION_IN_BENCHMARK";
+
+    /** Skip storage update in a benchmark. */
+    public static final String SKIP_STORAGE_UPDATE_IN_BENCHMARK = "SKIP_STORAGE_UPDATE_IN_BENCHMARK";
```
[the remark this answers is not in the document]
ascherbakoff: 🆗

modules/raft/src/main/java/org/apache/ignite/internal/raft/RaftGroupServiceImpl.java
```diff
@@ -171,19 +174,40 @@ public static CompletableFuture<RaftGroupService> start(
...
-                groupId,
-                cluster,
-                factory,
-                configuration,
-                membersConfiguration,
-                null,
-                executor,
-                commandsMarshaller
-        );
+        boolean inBenchmark = IgniteSystemProperties.getBoolean(IgniteSystemProperties.SKIP_REPLICATION_IN_BENCHMARK, false);
```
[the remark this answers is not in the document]
ascherbakoff: 🆗

modules/runner/src/integrationTest/java/org/apache/ignite/internal/benchmark/AbstractMultiNodeBenchmark.java
```diff
@@ -68,14 +68,14 @@ public class AbstractMultiNodeBenchmark {
 
     protected static IgniteImpl clusterNode;
 
-    @Param({"false", "true"})
+    @Param({"false"})
```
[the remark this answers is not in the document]
ascherbakoff: It was, to speed up benchmarks on a local dev's hardware.
I've reverted the change for now.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/benchmark/UpsertKvBenchmark.java
```diff
@@ -0,0 +1,109 @@
...
+    public void setUp() {
+        kvView = clusterNode.tables().table(TABLE_NAME).keyValueView();
+        for (int i = 1; i < 11; i++) {
+            tuple.set("field" + i, FIELD_VAL);
+        }
+    }
+
+    @TearDown
+    public void tearDown() {
+        System.out.println("Inserted " + id + " tuples");
```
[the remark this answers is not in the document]
ascherbakoff: Currenty all benchmarks don't use logger. This can be changed separately.

modules/table/src/main/java/org/apache/ignite/internal/table/RowIdGenerator.java
```diff
@@ -0,0 +1,36 @@
...
+ * New row id allocator.
+ */
+public class RowIdGenerator {
+    /**
+     * Get next row id.
+     *
+     * @return Next row id.
+     */
+    public static UUID next() {
+        return new UUID(FastTimestamps.coarseCurrentTimeMillis(), ThreadLocalRandom.current().nextLong());
```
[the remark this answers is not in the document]
ascherbakoff: My intention to use a timestamp is to reduce duplicate value probability due to periodical nature of a generator. 
Both variants seem equal from perf point of view.
I'm ok with both, but prefer to leave as is if you have no arguments agains.

modules/transactions/src/main/java/org/apache/ignite/internal/tx/impl/ReadWriteTransactionImpl.java
```diff
@@ -101,9 +101,11 @@ public IgniteBiTuple<ClusterNode, Long> enlist(
             TablePartitionId tablePartitionId,
             IgniteBiTuple<ClusterNode, Long> nodeAndConsistencyToken
     ) {
-        checkEnlistPossibility();
-
-        enlistPartitionLock.readLock().lock();
+        // No need to wait for lock if commit is in progress.
```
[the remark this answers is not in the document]
ascherbakoff: Because if a lock is not taken immediately, it means it's held by commit execution.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
ascherbakoff: There is no need in such check or assertion.
txTs can be assigned to non null value only if request instanceof ReadOnlyDirectReplicaRequest

modules/table/src/main/java/org/apache/ignite/internal/table/RowIdGenerator.java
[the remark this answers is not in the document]
rpuch: It seems that 64 bits per millisecond is ok

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
rpuch: In the current code this is true, but if we change the code, we can inadvertently break this implicit invariant. An assertion would help to make it explicit.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
```diff
@@ -3547,8 +3554,8 @@ private CompletableFuture<IgniteBiTuple<RowId, Collection<Lock>>> takeLocksForRe
      *
      * @param request Replica request.
      * @return Future with {@link IgniteBiTuple} containing {@code boolean} (whether the replica is primary) and the start time of current
-     *     lease. The boolean is not {@code null} only for {@link ReadOnlyReplicaRequest}. If {@code true}, then replica is primary. The
-     *     lease start time is not {@code null} in case of {@link PrimaryReplicaRequest}.
+     *         lease. The boolean is not {@code null} only for {@link ReadOnlyReplicaRequest}. If {@code true}, then replica is primary. The
+     *         lease start time is not {@code null} in case of {@link PrimaryReplicaRequest}.
```
rpuch, line 3558: Are these indents intentional?

rpuch (approved):

vldpyatkov (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/replicator/PartitionReplicaListener.java
ascherbakoff: Reverted changes.

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
```diff
@@ -1579,23 +1579,23 @@ private Publisher<BinaryRow> readOnlyScan(
...
-                        .batchSize(batchSize)
-                        .indexToUse(indexId)
-                        .exactKey(binaryTupleMessage(exactKey))
-                        .lowerBoundPrefix(binaryTupleMessage(lowerBound))
-                        .upperBoundPrefix(binaryTupleMessage(upperBound))
-                        .flags(flags)
-                        .columnsToInclude(columnsToInclude)
-                        .coordinatorId(txCoordinatorId)
-                        .build();
+                    ReadOnlyScanRetrieveBatchReplicaRequest request = TABLE_MESSAGES_FACTORY.readOnlyScanRetrieveBatchReplicaRequest()
```
rpuch, line 1582: Is this too many indents again? :)

rpuch (approved):

modules/table/src/main/java/org/apache/ignite/internal/table/distributed/storage/InternalTableImpl.java
rpuch: Oops, it the code it looks right. Weird.

PR #4233, opened 2024-08-14
IGNITE-22989 Add sync/async read-write lock

rpuch:
https://issues.apache.org/jira/browse/IGNITE-22989

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/VersatileReadWriteLock.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/VersatileReadWriteLock.java
@@ -172,7 +172,7 @@ public class VersatileReadWriteLock {
 
         int newState = state(false, readLocks(curState) + 1);
 
-        return compareAndSet(STATE_VH, curState, newState);
+        return STATE_VH.compareAndSet(this, curState, newState);
     }
 
     /**
@@ -211,7 +211,7 @@ public class VersatileReadWriteLock {
                 throw new IllegalMonitorStateException();
             }
 
-            if (compareAndSet(STATE_VH, curState, state(writeLocked, readLocks - 1))) {
+            if (STATE_VH.compareAndSet(this, curState, state(writeLocked, readLocks - 1))) {
                 if (readLocks == 1) {
                     // We released the final read lock.
                     notifyWriteLockSolicitors();
@@ -251,14 +251,14 @@ public class VersatileReadWriteLock {
         while (true) {
             int curPendingWriteLocks = pendingWriteLocks;
 
-            if (compareAndSet(PENDING_WLOCKS_VH, curPendingWriteLocks, curPendingWriteLocks + 1)) {
+            if (PENDING_WLOCKS_VH.compareAndSet(this, curPendingWriteLocks, curPendingWriteLocks + 1)) {
                 break;
             }
         }
     }
 
     private boolean trySwitchStateToWriteLocked() {
-        return compareAndSet(STATE_VH, AVAILABLE, state(true, 0));
+        return STATE_VH.compareAndSet(this, AVAILABLE, state(true, 0));
     }
 
     private void decrementPendingWriteLocks() {
@@ -267,7 +267,7 @@ public class VersatileReadWriteLock {
 
             assert curPendingWriteLocks > 0;
 
-            if (compareAndSet(PENDING_WLOCKS_VH, curPendingWriteLocks, curPendingWriteLocks - 1)) {
+            if (PENDING_WLOCKS_VH.compareAndSet(this, curPendingWriteLocks, curPendingWriteLocks - 1)) {
                 break;
             }
         }
@@ -346,7 +346,7 @@ public class VersatileReadWriteLock {
             throw new IllegalMonitorStateException();
         }
 
-        boolean b = compareAndSet(STATE_VH, state, state(false, readLocks));
+        boolean b = STATE_VH.compareAndSet(this, state, state(false, readLocks));
 
         assert b;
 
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/ignite/internal/util/VersatileReadWriteLock.java
```diff
@@ -0,0 +1,498 @@
...
+
+    /**
+     * Number of pending write attempts to acquire the write lock. It is used to prioritize write lock attempts over read
+     * lock attempts when the write lock has been released (so, if both an attempt to acquire the write lock and an attempt to acquire the
+     * read lock are waiting for write lock to be released, a write lock attempt will be served first when the release happens).
+     */
+    @SuppressWarnings("unused")
+    private volatile int pendingWriteLocks;
+
+    /** Futures to be completed when read locks (one per future) are acquired after a write lock is releasaed. */
```
ibessonov, line 105:
```suggestion
    /** Futures to be completed when read locks (one per future) are acquired after a write lock is released. */
```

modules/core/src/main/java/org/apache/ignite/internal/util/VersatileReadWriteLock.java
```diff
@@ -0,0 +1,498 @@
...
+    @SuppressWarnings("unused")
+    private volatile int state;
+
+    /**
+     * Number of pending write attempts to acquire the write lock. It is used to prioritize write lock attempts over read
+     * lock attempts when the write lock has been released (so, if both an attempt to acquire the write lock and an attempt to acquire the
+     * read lock are waiting for write lock to be released, a write lock attempt will be served first when the release happens).
+     */
+    @SuppressWarnings("unused")
+    private volatile int pendingWriteLocks;
```
ibessonov, line 103: Why not making both parts a single long value?

modules/core/src/main/java/org/apache/ignite/internal/util/VersatileReadWriteLock.java
```diff
@@ -0,0 +1,498 @@
...
+     * Returns {@code true} on success.
+     *
+     * @param varHandle VarHandle.
+     * @param expect    Expected.
+     * @param update    Update.
+     * @return {@code True} on success.
+     */
+    private boolean compareAndSet(VarHandle varHandle, int expect, int update) {
+        return varHandle.compareAndSet(this, expect, update);
+    }
```
ibessonov, line 413: Why did you do this? Now you have to rely on inline if you want Java to optimize your VH. Direct access to static final constants is preferable, this way JIT knows what to optimize without additional steps

modules/core/src/test/java/org/apache/ignite/internal/util/VersatileReadWriteLockTest.java
```diff
@@ -0,0 +1,606 @@
...
+
+    @Test
+    void readLockReleasedLessTimesThanAcquiredShouldStillBeTaken() throws Exception {
+        lock.readLock();
+
+        Future<?> future = executor.submit(() -> {
+            lock.readLock();
+            lock.readUnlock();
+        });
+        future.get(10, SECONDS);
```
ibessonov, line 287: Usually we use something like `willCompleteSuccessfully`. You can get a completable future instead of a future if you call `CompletableFuture.runAsync(runnable, executor)`

modules/core/src/main/java/org/apache/ignite/internal/util/VersatileReadWriteLock.java
rpuch: Good catch. Inlined the methods

modules/core/src/main/java/org/apache/ignite/internal/util/VersatileReadWriteLock.java
rpuch: Are you asking about `state` and `pendingWriteChecks`? They don't always get CASed together, so I don't see why we would want to merge them together.

ibessonov: They're on a single cache line anyway, so there's already a naturally occurring contention

modules/core/src/test/java/org/apache/ignite/internal/util/VersatileReadWriteLockTest.java
ibessonov: I see, thank you!

modules/core/src/test/java/org/apache/ignite/internal/util/VersatileReadWriteLockTest.java
```diff
@@ -277,14 +270,14 @@ void shouldNotAllowInterleavingHoldingReadAndWriteLocks() {
...
         lock.readLock();
 
-        Future<?> future = executor.submit(() -> {
+        CompletableFuture<?> future = runAsync(() -> {
             lock.readLock();
             lock.readUnlock();
-        });
-        future.get(10, SECONDS);
+        }, executor);
+        assertThat(future, willCompleteSuccessfully());
```
ibessonov, line 280: `runWithTimeout` does the same, is that correct? That's why I don't like on-liner methods, you just forget that you could use them

rpuch: Same logic, but different type of a closure

modules/core/src/main/java/org/apache/ignite/internal/util/VersatileReadWriteLock.java
rpuch: If they are in the same long, they are guaranteed to be on the same cache line; if they aren't, maybe they will be in different cache lines (this is possible at least in theory, right?). So no gain here, maybe even a loss.

Also, squashing them together complicates the code, which is also a loss.

I see no gain from squashing, only losses.

ibessonov (approved):

PR #4239, opened 2024-08-16
IGNITE-22877 Implement basic in-process IgniteServer restart

rpuch:
https://issues.apache.org/jira/browse/IGNITE-22877

1. restartAsync() is added to IgniteServerImpl
2. RestartProof* wrappers are added around public API objects; these wrappers survive an in-process node restart
3. The wrappers only cover Ignite and IgniteTables (and below), except for cursors
4. IgniteServerImpl now returns RestartProofIgnite wrapper

Micro-benchmarks do not show degradation:

```
Benchmark                         (clusterSize)  (fsync)  (partitionCount)  Mode  Cnt   Score    Error  Units
InsertBenchmark.kvInsertInternal              1    false                25  avgt   20  94,915 ± 19,184  us/op
InsertBenchmark.kvInsertPublic                1    false                25  avgt   20  93,873 ± 20,106  us/op
```

```
Benchmark                      (clusterSize)  (fsync)  Mode  Cnt   Score   Error  Units
SelectBenchmark.kvGetInternal              1    false  avgt   20  11,830 ± 0,838  us/op
SelectBenchmark.kvGetPublic                1    false  avgt   20  11,457 ± 0,568  us/op
```

Here, 'Internal' is for objects without new wrappers and 'Public' is for objects with new wrappers.


tkalkirill:
Maybe you could post the old values of micro-benchmarks ​​too to make sure there is no degradation?

rpuch:
> Maybe you could post the old values of micro-benchmarks ​​too to make sure there is no degradation?

'Internal' shows the 'old' figures (before changes)

modules/runner/src/integrationTest/java/org/apache/ignite/internal/app/ItShutDownServerApiReferencesTest.java
```diff
@@ -0,0 +1,70 @@
...
+    @EnumSource(SyncApiOperation.class)
+    void syncOperationsThrowAfterShutdown(SyncApiOperation operation) {
+        IgniteException ex = assertThrows(IgniteException.class, () -> operation.execute(beforeShutdown));
+        assertThat(ex.getMessage(), is("The node is already shut down."));
+    }
+
+    @ParameterizedTest
+    @EnumSource(AsyncApiOperation.class)
+    void asyncOperationsWorkAfterRestart(AsyncApiOperation operation) {
+        IgniteException ex = assertWillThrow(operation.execute(beforeShutdown), IgniteException.class, 10, SECONDS);
```
tkalkirill, line 67: Maybe use `org.apache.ignite.internal.testframework.matchers.CompletableFutureExceptionMatcher#willThrow(org.hamcrest.Matcher<? extends java.lang.Exception>, int, java.util.concurrent.TimeUnit, java.lang.String)` ?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/app/ItShutDownServerApiReferencesTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+        beforeShutdown = new References(server);
+
+        assertThat(server.shutdownAsync(), willCompleteSuccessfully());
+    }
+
+    @ParameterizedTest
+    @EnumSource(SyncApiOperation.class)
+    void syncOperationsThrowAfterShutdown(SyncApiOperation operation) {
+        IgniteException ex = assertThrows(IgniteException.class, () -> operation.execute(beforeShutdown));
```
tkalkirill, line 60: Maybe use `org.apache.ignite.internal.testframework.IgniteTestUtils#assertThrows` ?

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
```diff
@@ -82,10 +86,33 @@ public class IgniteServerImpl implements IgniteServer {
...
-    private volatile @Nullable IgniteImpl instance;
+    private final Executor asyncContinuationExecutor;
+
+    /** Current Ignite instance. This field is not volatile to make hot path accesses from IgniteReference and other references
+     * faster (they always happen under a read lock, which guarantees visibility of changes to this field). So we access
+     * this field in this object under synchronization ({@link #igniteMonitor} serves as the monitor).
+     */
+    private @Nullable IgniteImpl ignite;
+
+    private final Object igniteMonitor = new Object();
```
tkalkirill, line 97: I thought the classic name for such objects was "igniteMux".

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
```diff
@@ -82,10 +86,33 @@ public class IgniteServerImpl implements IgniteServer {
 
     private final ClassLoader classLoader;
 
-    private volatile @Nullable IgniteImpl instance;
+    private final Executor asyncContinuationExecutor;
+
+    /** Current Ignite instance. This field is not volatile to make hot path accesses from IgniteReference and other references
```
tkalkirill, line 91: Please add new line.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
```diff
@@ -82,10 +86,33 @@ public class IgniteServerImpl implements IgniteServer {
...
+     * Lock used to make sure user operations don't see Ignite instances in detached state (which might occur due to a restart)
+     * and that user operations linearize wrt detach/attach pairs (due to restarts).
+     */
+    private final IgniteAttachmentLock attachmentLock;
+
+    private final Ignite publicIgnite;
 
     private volatile @Nullable CompletableFuture<Void> joinFuture;
 
+    private final Object restartOrShutdownMonitor = new Object();
```
tkalkirill, line 109: Same about mux

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
```diff
@@ -82,10 +86,33 @@ public class IgniteServerImpl implements IgniteServer {
...
+
+    /**
+     * Used to make sure restart and shutdown requests are serviced sequentially.
```
tkalkirill, line 112: Please indicate that it guarded by `restartOrShutdownMonitor`.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
```diff
@@ -281,9 +386,11 @@ private static void ackBanner() {
 
     private static void sync(CompletableFuture<Void> future) {
         try {
-            future.join();
-        } catch (CompletionException e) {
+            future.get();
```
tkalkirill, line 389: Can you explain the change?

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
```diff
@@ -188,13 +236,64 @@ public CompletableFuture<Void> waitForInitAsync() {
...
     }
 
+    /**
+     * Restarts the node asynchronously. The {@link Ignite} instance obtained via {@link #api()} and objects acquired through it remain
+     * functional, but completion of calls to them might be delayed during the restart (that is, synchronous calls might take more time,
+     * while futures from asynchronous calls might take more time to complete).
+     *
+     * @return CompletableFuture that gets completed when the node startup has completed (either successfully or with an error).
+     */
+    CompletableFuture<Void> restartAsync() {
```
tkalkirill, line 246: It looks like there might be a race here.

While we started restarting the node, it is shutdown in the process and the first checks that are indicated here will already be passed.
What about using not mutex but CAS on the `restartOrShutdownFuture` and each new operation will be performed after the previous one is completed and will thus be linearized and with the correct checks.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
```diff
@@ -82,10 +86,33 @@ public class IgniteServerImpl implements IgniteServer {
...
     private final ClassLoader classLoader;
 
-    private volatile @Nullable IgniteImpl instance;
+    private final Executor asyncContinuationExecutor;
+
+    /** Current Ignite instance. This field is not volatile to make hot path accesses from IgniteReference and other references
+     * faster (they always happen under a read lock, which guarantees visibility of changes to this field). So we access
+     * this field in this object under synchronization ({@link #igniteMonitor} serves as the monitor).
+     */
+    private @Nullable IgniteImpl ignite;
```
tkalkirill, line 95: I noticed that we only read and write in this field, why not replace it with `volatile`?

modules/runner/src/main/java/org/apache/ignite/internal/restart/RefCache.java
```diff
@@ -0,0 +1,55 @@
...
+     * Returns an instance corresponding to the given {@link Ignite} instance.
+     *
+     * <p>This method must ALWAYS be called under the attachment lock.
+     *
+     * @param ignite Ignite instance for which to obtain an object.
+     */
+    public T actualFor(Ignite ignite) {
+        IgniteAware<T> igniteAware = this.value;
+
+        if (igniteAware == null || igniteAware.ignite() != ignite) {
```
tkalkirill, line 48: How `igniteAware` can be `null` ?

modules/runner/src/main/java/org/apache/ignite/internal/restart/RestartProofApiObject.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import org.apache.ignite.Ignite;
+
+/**
+ * Base for references to API objects under a swappable {@link Ignite}.
+ */
+abstract class RestartProofApiObject<T> {
```
tkalkirill, line 28: What does `Proof` mean in the class name?

modules/runner/src/integrationTest/java/org/apache/ignite/internal/app/Record.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.apache.ignite.table.Tuple;
+
+/**
+ * Simple record for tests.
+ */
+@SuppressWarnings({"FieldCanBeLocal", "unused"})
+class Record {
+    private int id;
+    private String val;
```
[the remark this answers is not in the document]
rpuch: Mappers require the fields to be non-final to be able to create the object using a no-arg constructor and then fill it.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/app/SyncApiOperation.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Map;
+import java.util.function.Consumer;
+import org.apache.ignite.table.mapper.Mapper;
+
+/**
+ * Synchronous API operation.
+ */
+@SuppressWarnings("resource")
+enum SyncApiOperation {
+    IGNITE_NAME(refs -> refs.ignite.name()),
```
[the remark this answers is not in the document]
rpuch: All operations have names built using the same pattern: $REF_$OPERATION-ON-REF. Here, $REF is Ignite and $OPERATION-ON-REF is 'name', so it looks consistent.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/app/SyncApiOperation.java
```diff
@@ -0,0 +1,117 @@
...
+    TABLE_NAME(refs -> refs.table.name()),
+    TABLE_KVVIEW(refs -> refs.table.keyValueView()),
+    TABLE_TYPED_KVVIEW(refs -> refs.table.keyValueView(Integer.class, String.class)),
+    TABLE_MAPPED_KVVIEW(refs -> refs.table.keyValueView(Mapper.of(Integer.class), Mapper.of(String.class))),
+    TABLE_RECORDVIEW(refs -> refs.table.recordView()),
+    TABLE_TYPED_RECORDVIEW(refs -> refs.table.recordView(Record.class)),
+    TABLE_MAPPED_RECORDVIEW(refs -> refs.table.recordView(Mapper.of(Record.class))),
+    TABLE_PARTITION_MANAGER(refs -> refs.table.partitionManager()),
+
+    TABLE_FROM_TABLE_ASYNC_PUT(refs -> refs.tableFromTableAsync.keyValueView().put(null, KEY_TUPLE, VALUE_TUPLE)),
```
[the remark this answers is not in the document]
rpuch: This is a 'put called on tableFromTableAsync', that is, on a `Table` instance obtained from `IgniteTables#tableAsync(String)`. This 'async' is important here.

modules/runner/src/integrationTest/java/org/apache/ignite/internal/app/ItShutDownServerApiReferencesTest.java
rpuch: What's the advantage?

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
rpuch: `join()` is uninterruptible, hence IMO we should avoid it

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
rpuch: I tried to explain this in the javadoc. When executing user operations, we always read this under a read lock, so we don't need it to be `volatile` for visibility. If we make it `volatile`, we will add an extra volatile read, which might hurt the performance.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
rpuch: I fixed it a bit differently: I added `shutDown` flag that is rechecked on restart after acquiring the mutex.
The future chaining is already implemented.
CAS logic seems to be a bit more involved compared to simple and rude `synchronized`

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
rpuch: Renamed it to `Mutex`. `Mux` is funny, but it's a mutex, after all :)

modules/runner/src/main/java/org/apache/ignite/internal/restart/RestartProofApiObject.java
tkalkirill: Ok

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
tkalkirill: If I understood the documentation correctly, it says that reading from the synchronization block is cheaper in terms of performance than one volatile read.
Is there any proof of this? If we have two parallel threads reading this field, they will step on each other's toes and park someone, which will lead to performance degradation.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
tkalkirill: It looks more complicated to me, but it's up to you.

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
tkalkirill: Ok

tkalkirill (approved):

modules/runner/src/main/java/org/apache/ignite/internal/app/IgniteServerImpl.java
rpuch: I tried making the field volatile and rerun the benchmarks. Volatility effect is not noticable at all. So I just simplified everything by making the field volatile and removing `syncrhonized`.

PR #4242, opened 2024-08-16
IGNITE-22267 Improve Table API javadocs

alievmirza:
1. Remove "started table" term, "table" is enough
2. Remove "binary object" mentions
3. Explain how KV view projects row data into key-value pairs
4. Replace "Transaction or null to auto-commit" with "Transaction or null for implicit transaction"
5. Remove "@throws IgniteException If an unspecified platform exception has occurred internally", it is not useful
6. Describe the order of tables in tables() and tablesAsync()

https://issues.apache.org/jira/browse/IGNITE-22267


ptupitsyn:
@alievmirza I've pushed some more changes, please review them.

alievmirza:
@ptupitsyn Thank you for help, your part looks good!

PR #4244, opened 2024-08-16
IGNITE-22482 Add topmost root to node configuration

valepakh:
https://issues.apache.org/jira/browse/IGNITE-22482

This patch introduces "ignite" configuration root for the node configuration by moving roots to the extensions of the new `NodeConfigurationSchema` root.


docs/_docs/developers-guide/clients/cpp.adoc
```diff
@@ -147,24 +147,27 @@ Here is how the client connector configuration looks like:
...
-    enabled: true,
-    clientAuth: "require",
-    keyStore: {
-      path: "KEYSTORE_PATH",
-      password: "SSL_STORE_PASS"
-    },
-    trustStore: {
-      path: "TRUSTSTORE_PATH",
-      password: "SSL_STORE_PASS"
+{
```
ibessonov, line 150: Since we're fixing it anyway, we could remove quotations from the example, making it a proper HOCON rather than a JSON. Maybe it's worth doing as a separate JIRA, to make merge conflicts simpler. What do you think?

modules/cli/src/integrationTest/java/org/apache/ignite/internal/NodeConfig.java
```diff
@@ -44,27 +44,25 @@ public class NodeConfig {
...
-                + "  },\n"
-                + "  clientConnector.port: {} ,\n"
-                + "  rest: {\n"
+        return "ignite {\n"
+                + "  network {\n"
+                + "    port: {}\n"
+                + "    nodeFinder.netClusterNodes: [ {} ]\n"
+                + "  }\n"
+                + "  clientConnector.port: {}\n"
+                + "  rest {\n"
```
ibessonov, line 53: Changing the rest of these lines wasn't really necessary, what motivated you to do more changes than you really needed?
It's everywhere in this PR, I bet it can be 3 times smaller if you remove all unnecessary changes

modules/configuration-api/src/main/java/org/apache/ignite/configuration/RootKey.java
```diff
@@ -28,7 +28,7 @@
  * @param <T>    Type of the configuration tree described by the root key.
  * @param <VIEWT> Type of the immutable snapshot view associated with the tree.
  */
-public class RootKey<T extends ConfigurationTree<VIEWT, ?>, VIEWT> {
+public class RootKey<T extends ConfigurationTree<? super VIEWT, ?>, VIEWT> {
```
ibessonov, line 31: I think we don't need this change

modules/configuration-api/src/main/java/org/apache/ignite/configuration/SuperRootChange.java
```diff
@@ -17,20 +17,17 @@
...
-
 /**
  * Interface that represent a "change" for the conjunction of all roots in the configuration.
  */
 public interface SuperRootChange {
     /**
      * Returns a root view for the root key.
      */
-    <V> V viewRoot(RootKey<? extends ConfigurationTree<V, ?>, V> rootKey);
+    <V, C, T extends ConfigurationTree<? super V, ? super C>> V viewRoot(RootKey<T, V> rootKey);
```
ibessonov, line 27: What is this for?

modules/configuration/src/main/java/org/apache/ignite/internal/configuration/ConfigurationRegistry.java
```diff
@@ -136,7 +136,7 @@ public void initializeConfigurationWith(ConfigurationSource configurationSource)
      * @param <T> Configuration tree type.
      * @return Public configuration tree.
      */
-    public <V, C, T extends ConfigurationTree<V, C>> T getConfiguration(RootKey<T, V> rootKey) {
+    public <V, C, T extends ConfigurationTree<? super V, C>> T getConfiguration(RootKey<T, V> rootKey) {
```
ibessonov, line 139: I also doubt that this is required, as well as all the rest of `? super`, there should be an easier way

modules/client-handler/src/main/java/org/apache/ignite/client/handler/configuration/ClientHandlerConfigurationModule.java
```diff
@@ -35,7 +34,7 @@ public ConfigurationType type() {
     }
 
     @Override
-    public Collection<RootKey<?, ?>> rootKeys() {
-        return Collections.singleton(ClientConnectorConfiguration.KEY);
+    public Collection<Class<?>> schemaExtensions() {
+        return List.of(ClientConnectorExtensionConfigurationSchema.class);
```
ptupitsyn, line 38: Is it necessary to create an almost-empty `ClientConnectorExtensionConfigurationSchema` class? Can we somehow pass `ClientConnectorConfigurationSchema` directly?

ptupitsyn: It just looks like there must be a shortcut to avoid manually creating an empty class, but I might be wrong.

ptupitsyn: Ok, let's keep it if there is no easy way to simplify.

modules/cli/src/integrationTest/java/org/apache/ignite/internal/NodeConfig.java
```diff
@@ -44,27 +44,25 @@ public class NodeConfig {
...
-        return "{\n"
-                + "  network: {\n"
-                + "    port: {},\n"
-                + "    nodeFinder: {\n"
-                + "      netClusterNodes: [ {} ]\n"
-                + "    },\n"
-                + "  },\n"
-                + "  clientConnector.port: {} ,\n"
-                + "  rest: {\n"
+        return "ignite {\n"
```
ibessonov, line 47: You forgot to rollback this one

modules/runner/src/integrationTest/java/org/apache/ignite/internal/configuration/ItNodeConfigurationFileTest.java
```diff
@@ -57,27 +57,27 @@ public void wrongConfigurationFormat(TestInfo testInfo, @WorkDirectory Path work
...
-                + "      enabled: true,\n"
-                + "      clientAuth: none,\n"
-                + "      keyStore: {\n"
-                + "        path: 123\n"
-                + "      }\n"
-                + "    }\n"
-                + "  }\n"
-                + "}";
+                "ignite {\n"
+                        + "  rest: {\n"
```
ibessonov, line 61: Something's wrong with padding, please make it as it was

ibessonov (approved):

